PR #674, opened 2020-10-14
Remove 'TODO' in code example

jescalan:
I'm not sure if we want to ship this out with the TODO, but if its not implemented I feel like we still want a note right?


malnick (approved):
Thanks!

PR #678, opened 2020-10-14
docs: Fix a typo & broken link

radeksimko:
The "Edit this page" link actually still links to 'master' branch
but that seems to be a harder problem to solve as the template
seems to be shared.

### DIFF
--- a/website/pages/docs/[[...page]].jsx
+++ b/website/pages/docs/[[...page]].jsx
@@ -12,7 +12,7 @@ function DocsLayout(props) {
   return (
     <DocsPage
       productName={productName}
-      productSlug="red"
+      productSlug="boundary"
       subpath={subpath}
       order={order}
       staticProps={props}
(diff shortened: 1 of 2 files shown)


website/pages/docs/[[...page]].jsx
```diff
@@ -12,7 +12,7 @@ function DocsLayout(props) {
   return (
     <DocsPage
       productName={productName}
-      productSlug="red"
+      productSlug="boundary"
```
Comment (BrandonRomano) [inline website/pages/docs/[[...page]].jsx:15]: Actually, keep this as `red` for now!   This was our product codename prior to launch, and styles are still dependent on this.

We'll make this change in the near future, but first we'll have to do some component upgrades.

```suggestion
      productSlug="red"
```

BrandonRomano (changes requested):

website/pages/docs/[[...page]].jsx
Comment (radeksimko) [reply website/pages/docs/[[...page]].jsx:15]: Fair, I wasn't aware this affects styles too.

I suppose you know already, but in case it wasn't clear from the PR - it breaks the `Edit this page` URL on all pages.

jefferai (approved):

jefferai (approved):
Thanks!

PR #683, opened 2020-10-14
ensure make tools executes before build-ui

mootpt:
`make dev` requires `build-ui` which executes a script that requires `go-bindata` binary that isn't installed without executing `make tools`. This pr ensures the `tools` make target is executed first.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ cleangen:
 
 dev: BUILD_TAGS+=dev
 dev: BUILD_TAGS+=ui
-dev: tools build-ui-ifne
+dev: build-ui-ifne
 	@echo "==> Building Boundary with dev and UI features enabled"
 	@CGO_ENABLED=$(CGO_ENABLED) BUILD_TAGS='$(BUILD_TAGS)' BOUNDARY_DEV_BUILD=1 sh -c "'$(CURDIR)/scripts/build.sh'"
 
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ cleangen:
 
 dev: BUILD_TAGS+=dev
 dev: BUILD_TAGS+=ui
-dev: build-ui-ifne
+dev: tools build-ui-ifne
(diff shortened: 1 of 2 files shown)


malnick:
Thanks @mootpt - this looks great.

malnick (approved):

malnick (changes requested):
@mootpt - Second pass on this, I thought you added it to the `dev` step. Since this isn't UI specific, I think it's more appropriate to add it to the `dev` target, which inevitably runs the `build-ui` target.

mootpt:
got it. fixing

mootpt:
@malnick should be all set.

mootpt:
@malnick actually you sure you don't want it in build-ui step? That step executes `go-bindata` which is unavailable unless tools are installed

jefferai (changes requested):
We specifically chose not to run `tools` for `dev` because it installs binaries and may overwrite (or hide, via being further up in PATH) versions of binaries that you want to stay installed.

A better approach might be to separate out the step that runs go-bindata on the assets that result after building the UI, and then if that fails, print out something suggesting running `make tools`, but with a warning that this may replace binaries they already have on the system.

I realize this is a chicken and egg issue because then you'd have to go through the UI build again. Open to better suggestions!

jefferai:
Also, hi @mootpt !!

jefferai:
Perhaps if go-bindata is the only tool we need after the build step, we can simply attempt to find or invoke it before build-ui and if it doesn't exist, we prompt the user to install it or run `make tools` with appropriate caveats?

jack1902:
https://github.com/hashicorp/boundary/blob/f008dff9eba362e3ea7e54c93c853f3b15411c06/scripts/build.sh#L52 

`gox` is also required as part of `make dev`. Prehaps just mentioning you _might_ need to run `make tools`  in the README might help?

mootpt:
yup. I'll just add to README

jefferai (approved):
I wordsmithed it a little, but looks good -- thanks!

PR #687, opened 2020-10-14
internal/servers/controller: fix dropped test error

alrs:
This fixes a dropped test error in `internal/servers/controller/`.


jefferai:
Thanks!

jefferai (approved):

PR #689, opened 2020-10-15
Fix hyphenation of resource subcommand output

### LINKED ISSUE #686: Usage output for plural objects (host-catalogs and host-sets) not correct, missing dash
**Describe the bug**
The help function isn't showing the correct syntax of the command itself

**To Reproduce**
1. Run `boundary host-catalogs list -help` or `boundary host-sets list -help`
2. Returns: 
```
Usage: boundary host catalogs list [options] [args]

  List host catalogs within an enclosing scope or resource. Example:

    $ boundary host catalogs list
```

which is wrong (missing dash between host and catalogs in the usage example)

**Expected behavior**
Should output:

```
Usage: boundary host-catalogs list [options] [args]

  List host catalogs within an enclosing scope or resource. Example:

    $ boundary host-catalogs list
```

**Additional context**
Might be due to this plural settings:

https://github.com/hashicorp/boundary/blob/main/internal/cmd/commands/hostcatalogs/hostcatalog.go#L170

jefferai:
Fixes #686

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 
 ## vNext
 
+### What's New
+
+### Improvements
+
+### Bug Fixes
+
 * cli: Fix hyphenation in help output for resources with compound names
   ([Issue](https://github.com/hashicorp/boundary/issues/686))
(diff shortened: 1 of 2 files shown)


malnick (approved):
👍 Thanks for the changelog additions!

Issue #697, opened 2020-10-15
Issue with secrets storage on Ubuntu


saiyam1814:
@jefferai taking conversation here with Katakoda issue as well 
I am in process of writing a blog on neat installation steps for Ubuntu/Centos 
There are few things that do not seem to work and if I can get them resolved then we can have a neat install step which can be helpful for others in the community as well I believe 

I am planning to tell on 3 environments 
bare centos 7 VM , Ubuntu 18.04 Vm and katakoda ubuntu playground 

For Ubuntu I have bellow error s:
```
boundary authenticate password -auth-method-id=ampw_1234567890 \
>     -login-name=admin -password=password
Error reading auth token from system credential store: exec: "dbus-launch": executable file not found in $PATH

Authentication information:
  Account ID:      apw_EJK7s6DcCQ
  Auth Method ID:  ampw_1234567890
  Expiration Time: Thu, 22 Oct 2020 10:17:30 BST
  Token:
  at_h1qSNMdB1c_s125wLmVW7LgzNKf1TFeNrFDRUZNcR2qeWkNyjZ7Qd6E2DWGsyPT9KmPKZkxBaQps7JKkbeeoJEwt7xXMyR6YKjEkqbrFWcsCdXm8rYgqjsFJwUYS2WBFbNh2
  User ID:         u_1234567890
Error saving auth token to system credential store: exec: "dbus-launch": executable file not found in $PATH

```

Installed dbus 
```
apt install dbus-x11
```

```
boundary authenticate password -auth-method-id=ampw_1234567890     -login-name=admin -password=password
Error reading auth token from system credential store: The name org.freedesktop.secrets was not provided by any .service files

Authentication information:
  Account ID:      apw_EJK7s6DcCQ
  Auth Method ID:  ampw_1234567890
  Expiration Time: Thu, 22 Oct 2020 10:19:18 BST
  Token:
  at_VqoR4tBhxy_s18pG3buSH7fMcBVL1YP14NXipTgeg4Jm9aVd3WXL2WCpUvqs8xnCwnNC7D41zejxUfRYhJPGMvCcp1mbN1q9cPyM5AAJRvakFGQUNLC5ZjiyyoTuic4ZpoQUmu9XXtC2YY2u9BTdJDg8
  User ID:         u_1234567890
Error saving auth token to system credential store: The name org.freedesktop.secrets was not provided by any .service files
```

Next installed 
```
sudo apt-get install -y gnome-keyring
```

```
[...]

PR #698, opened 2020-10-15
docs: add details on not saving token to system key manager

malnick:
Also adds tabs to cleanup the CLI/UI/TF flows for getting started.


malnick:
Addresses https://github.com/hashicorp/boundary/issues/697

PR #702, opened 2020-10-15
website: fix edit this page links on /docs

kylemac:
previously, we were using `red` as a codename for the project. this fixes an issue where the "edit this page" links were pointing to a repo that didn't exist based on said codename.


BrandonRomano:
Closing this out in favor of https://github.com/hashicorp/boundary/pull/708;

We still need to use the code name for the styling (until soon, when we update our styling).  The PR above that I link to handles this in a way that keeps both styling and leads us to link to the right place.

PR #703, opened 2020-10-15
Fix IPv6 binding

### LINKED ISSUE #701: Unable to specify listen port with any IPv6 address
**Describe the bug**
Boundary will throw an error with any listen directive, when an IPv6 address and port are specified.
This happens with runtime options as well as when using the config file.

> Error initializing listener of type tcp: listen tcp: address :::8080: too many colons in address

**To Reproduce**
```
boundary dev -api-listen-address="[2001:db8::1:cee:c0de:b0b]:8080"
```
OR
```
listener "tcp" {
  address = "[2001:db8::1:cee:c0de:b0b]:8080"
  purpose = "api"
}
```
both result in:
> Error initializing listener of type tcp: listen tcp: address 2001:db8::1:cee:c0de:b0b:8080: too many colons in address

**Expected behavior**
Boundary should open a TCP socket on "2001:db8::1:cee:c0de:b0b" with port 8080.

jefferai:
Annoyingly, I somehow managed to test net.SplitHostPort and account for
IPv6 but then reassemble it for listening in the wrong way, such that it
flat out doesn't work. No idea how. Anyways, fixed.

Fixes #701

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 * cli: Fix hyphenation in help output for resources with compound names
   ([Issue](https://github.com/hashicorp/boundary/issues/686))
   ([PR](https://github.com/hashicorp/boundary/pull/689))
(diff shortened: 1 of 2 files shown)


jefferai:
Note: I'm folding this change into #705. As I was working on that I found other places where the address is being Sprinted instead of JoinHostPorted so rather than keep this as an outlier with that series of changes, I'm simply going to combine them.

PR #705, opened 2020-10-15
Support for Unix domain socket listeners

### LINKED ISSUE #699: Ability to use a Unix Domain Socket with listeners in addition to localhost
**Is your feature request related to a problem? Please describe.**
Many use cases will have some sort of proxy in front of Boundary, only requiring direct connections to Boundary from processes on the same host / the same namespace. 
While opening a TCP socket on localhost is an option, it carries unnecessary overhead for this use case.
A Unix Domain Socket would be much simpler and more efficient for inter-process communication. 

**Describe the solution you'd like**
Either a different runtime option for unix domain sockets eg : `-api-listen-socket="/var/run/boundary/api.sock"` 
Or the option to specify a socket location through the existing options eg: `-api-listen-address="unix://var/run/boundary/api.sock"`

Configuring a UDS listener could look like this:
```
listener "unix" {
  purpose = "api"
  socket = "var/run/boundary/api.sock"
}
```

jefferai:
The client already supports protocol-agnostic transport so handles unix:/// schemes.

Fixes #699

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,18 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 
 ### Improvements
 
+* controller: Allow API/Cluster listeners to be Unix domain sockets
+  ([Issue](https://github.com/hashicorp/boundary/pull/699))
+  ([PR](https://github.com/hashicorp/boundary/pull/705))
+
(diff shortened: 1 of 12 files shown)


malnick (approved):
👍

PR #713, opened 2020-10-16
docs: how to install without generated resources

malnick:


website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -0,0 +1,484 @@
+---
+layout: docs
+page_title: Installing Boundary
```
jefferai, line 3:
```suggestion
page_title: Installing Boundary Without Generated Resources
```

website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -0,0 +1,484 @@
...
+sidebar_title: Without Generated Resources 
+description: |-
+  How to install boundary without generated resources.
+---
+
+# Installing Boundary Without Generated Resources 
+
+Initializing Boundary without Generated resources for the first time can be a little confusing. How do you
```
jefferai, line 11:
```suggestion
Initializing Boundary without generated resources for the first time can be a little confusing. How do you 
```

website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -0,0 +1,484 @@
...
+From this command, you can see the flags available to skip the creation of auto-generated resources. These
+resources are useful when running in development environments but in a non development environment may not
+be desired. You can think of them as Boundary's default username and password, and if used in non dev workflows
+should be overidden or skipped to avoid undermining Boundary's security model.
+
+To initialize the Boundary database without generated resources:
+
+```
+$ boundary database init \
+  -skip-auth-method-creation \
```
jefferai, line 120: All you need is `-skip-initial-login-role`. If you look at the help output for that command, it does generation in-order and skipping anything higher up the ordering skips everything below it. For completeness you could leave all of them in there, but then you're missing `-skip-initial-login-role` :-)

website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -0,0 +1,484 @@
...
+  -skip-host-resources-creation \
+  -skip-scopes-creation \
+  -skip-target-creation \ 
+  -config /etc/boundary.hcl
+```
+
+When you start Boundary, you will effectively have a blank sheet to work against. From here, it's required that
```
jefferai, line 127:
```suggestion
When you start Boundary, you will effectively have a blank sheet to work against. The initial migrations in the database have been run (note that this includes creating special users like `u_anon` and the `global` scope) and the internal keyrings have been initialized.. From here, it's required that
```

website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -0,0 +1,484 @@
...
+you use the KMS recovery workflow described above to create at a minimum an auth method, a user, an account, and a 
+role with sufficient grants. Otherwise, you need to continue to use the recovery workflow for management. Though 
+continuing with the recovery workflow is not discouraged it's important to realize that this is effectively a
```
jefferai, line 130: I think saying it's not discouraged is a little _too_ encouraging. Maybe just state the warning and that in some circumstances it may be desirable to continue to use the workflow, but it must be done with caution.

website/content/docs/installing/no-gen-resources.mdx
malnick: 👍

website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -0,0 +1,484 @@
...
+$ export BOUNDARY_RECOVERY_CONFIG=<<EOT
+kms "aead" {
+	purpose   = "recovery"
+	aead_type = "aes-gcm"
+	key       = "8fZBjCUfN0TzjEGLQldGY4+iE9AkOvCfjh7+p0GtRBQ="
+	key_id    = "global_recovery"
+}
+EOT
+
+$ boundary connect ssh -target-id ttcp_873964g581
```
jefferai, line 74: You specifically _cannot_ use the recovery workflow to authorize a session! :-)

website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -0,0 +1,478 @@
...
+Before you can start Boundary, the database must be initialized. To do this, run:
+
+```
+$ boundary database init -h
+...
+```
+
+From this command, you can see the flags available to skip the creation of auto-generated resources. These
+resources are useful when running in development environments but in a non development environment may not
+be desired. You can think of them as Boundary's default username and password, and if used in non dev workflows
```
jefferai, line 113: I don't really understand this sentence. What about the generated resources undermines Boundary's security model?

website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -0,0 +1,478 @@
...
+}
+```
+
+</Tab>
+</Tabs>
+
+
+## Initialize the Database
+
+Before you can start Boundary, the database must be initialized. To do this, run:
```
jefferai, line 104: This doesn't _actually_ initialize the database. Maybe say:

```suggestion
Before you can start Boundary, the database must be initialized. It's useful to look at the help output for the init command:
```

website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -0,0 +1,478 @@
...
+be desired. You can think of them as Boundary's default username and password, and if used in non dev workflows
+should be overidden or skipped to avoid undermining Boundary's security model.
+
+To initialize the Boundary database without generated resources:
+
+```
+$ boundary database init -skip-initial-login-role -config /etc/boundary.hcl
+```
+
+When you start Boundary, you will effectively have a blank sheet to work against. The initial migrations in the database have been run (note that this includes creating special users like `u_anon` and the `global` scope) and the internal keyrings have been initialized.. From here, it's required that
```
jefferai, line 122:
```suggestion
When you start Boundary, you will effectively have a blank sheet to work against. The initial migrations in the database have been run (note that this includes creating special users like `u_anon` and the `global` scope) and the internal keyrings have been initialized. From here, it's required that
```

website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -0,0 +1,478 @@
...
+you use the KMS recovery workflow described above to create at a minimum an auth method, a user, an account, and a 
+role with sufficient grants. Otherwise, you need to continue to use the recovery workflow for management. t's important
```
jefferai, line 124:
```suggestion
role with sufficient grants. Otherwise, you need to continue to use the recovery workflow for management. It's important 
```

website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -0,0 +1,478 @@
...
+</Tab>
+<Tab heading="Terraform">
+
+```hcl
+resource "boundary_scope" "org" {
+  scope_id    = "global"
+  name        = "organization"
+  description = "Organization scope"
+
+  auto_create_admin_role   = false
```
jefferai, line 171: These are the defaults, you don't need to specify them. Not doing so makes it more clear in the other examples when you _are_ specifying them that that is not the default behavior.

website/content/docs/installing/no-gen-resources.mdx
malnick: For example, if you don't change the default login account's password, it's hardcoded.

website/content/docs/installing/no-gen-resources.mdx
malnick: I went back and forth on this, I wanted to explicitly call them out since the logic on the CLI is other way around: in TF, it's the default but on the CLI, you have to explicitly call the flag. If someone was looking at the TF and CLI examples side by side, it would be more clear.

website/content/docs/installing/no-gen-resources.mdx
jefferai: Only in dev mode. Everything is fully generated when doing `boundary database init` and you _cannot_ pick the values.

website/content/docs/installing/no-gen-resources.mdx
jefferai: Fair enough

website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -0,0 +1,466 @@
...
+  scope_id    = boundary_scope.org.id
+}
+```
+
+</Tab>
+</Tabs>
+
+
+#### Create Roles to Manage Scopes
+
```
jefferai, line 278: It might be worth some text here explaining to the user that these four roles will end up granting the same access as if role autogeneration had been enabled, but with Terraform in full control?

website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -0,0 +1,466 @@
...
+Assumes recovery key export from above steps is still set:
+
+```bash
+# Create global anonymous listing role
+$ boundary roles create -name 'global_anon_listing' \
+  -recovery-config /tmp/recovery.hcl \
+  -scope-id 'global'
+
+$ boundary add-grants -id <global_anon_listing_id> \
```
jefferai, line 298: you can specify `-grant` multiple times so this can be one call instead of three.

website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -0,0 +1,466 @@
...
+  -recovery-config /tmp/recovery.hcl \
+  -grant 'id=*;type=auth-method;actions=list,authenticate'
+
+$ boundary add-grants -id <global_anon_listing_id> \
+  -recovery-config /tmp/recovery.hcl \
+  -grant 'type=scope;actions=list'
+
+$ boundary add-grants -id <global_anon_listing_id> \
+  -recovery-config /tmp/recovery.hcl \
+  -grant 'id=*;actions=read,change-password'
```
jefferai, line 308: I think you meant for this to be `'id={{account.id}};actions=read,change-password'`

website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -0,0 +1,466 @@
...
+</Tab>
+<Tab heading="Terraform">
+
+```hcl
+resource "boundary_role" "global_anon_listing" {
+  scope_id = "global"
+  grant_strings = [
+    "id=*;type=auth-method;actions=list,authenticate",
+    "type=scope;actions=list",
+    "id=*;actions=read,change-password"
```
jefferai, line 324: Same here

website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -0,0 +1,466 @@
...
+
+2. Create anonymous listing role for org scope
+
+<Tabs>
+<Tab heading="CLI">
+
+Assumes recovery key export from above steps is still set:
+
+```bash
+$ boundary roles create -name 'global_anon_listing' \
```
jefferai, line 341: Did you mean for global_anon_listing here and below to be something like "org_anon_listing"?

website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -0,0 +1,456 @@
...
+```
+
+</Tab>
+</Tabs>
+
+
+### Create Roles to Manage Scopes
+
+The following describes the four baseline roles you'll need to create to manage resources within the org and project 
+scopes created above. These roles are similar to the roles created for you in dev mode but declaring them explicitly
```
jefferai, line 277:
```suggestion
scopes created above. These roles are similar to the roles created for you in if generation had not been skipped during `boundary database init`, but declaring them explicitly
```

wobuhuisuanmin1 (approved):

twisterin (approved):

twisterin (approved):

Issue #711, opened 2020-10-16
Errors in systemd example docs


pentiumoverdrive:
**Describe the bug**
So reading this one, I found lots of errors regarding the service file: 
https://www.boundaryproject.io/docs/installing/systemd

There should not be a ${TYPE} as flag to boundary:
`ExecStart=/usr/local/bin/${NAME} ${TYPE} -config /etc/${NAME}-${TYPE}.hcl`

It should instead say:
`ExecStart=/usr/local/bin/${NAME} server -config /etc/${NAME}-${TYPE}.hcl`

In CentOS 8, this seems deprecated, you get lots of warning if you don't comment out:
`Capabilities=CAP_IPC_LOCK+ep`

This doesn't work since there is no group created in the script:
`sudo adduser --system --group boundary || true`

So would work better with:

`sudo  groupadd boundary`
`sudo adduser --system --group boundary boundary || true`

Using the RPM package, this is created flawlessly.

malnick:
Thanks for raising this @pentiumoverdrive. 

The installing guide is specifically focused on Debian style linux distributions and we purposefully don't want to make it cross distro to avoid an over complicated example. We believe our operators know best how to administer services on their operating system of choice and are able to use this as a jumping off point to managing boundary in a long running environment they choose. 

I'll look into that deprecated capability flag, but CentOS might be slightly different here. Again, not sure since I made this as an example for Debian and didn't verify on other platforms. 

Regarding your comment on the `ExecStart` command, I added https://github.com/hashicorp/boundary/pull/717 to fix this. Thanks for catching that and let me know if you have any other questions.

PR #717, opened 2020-10-16
docs: fix ExecStart command in systemd docs

malnick:


malnick:
Addresses documentation bug in https://github.com/hashicorp/boundary/issues/711

Issue #727, opened 2020-10-18
Boundary tries to connect to 127.0.0.1:9201 even when -cluster-listen-address is set


tcassaert:
**Describe the bug**
When trying to run `boundary` in dev mode, with a different `cluster-listen-address`, boundary tries to reach `127.0.0.1:9201` instead of the provided `cluster-listen-address`.

**To Reproduce**
1. Run `boundary dev -cluster-listen-address=10.69.0.24`
2. Boundary starts, but shows following error:
```
2020-10-18T18:35:48.201Z [ERROR] worker: error making status request to controller: error="rpc error: code = Unavailable desc = last connection error: connection error: desc = "transport: Error while dialing unable to dial to controller: dial tcp 127.0.0.1:9201: connect: connection refused""
```

**Expected behavior**
Boundary should try to connect to the provided `cluster-listen-address`.

I realize there's not much reason to change the cluster-listen-address in dev mode, but when trying the options, it was confusing to see it fail.

malnick:
Thanks @tcassaert - I was able to repro this and will work on getting a fix in soon!

malnick:
@tcassaert We fixed this on #705 and I've added a changelog update to reflect it: https://github.com/hashicorp/boundary/pull/730

Feel free to build and test from `main`. Thanks!

tcassaert:
Thanks!
I wanted to try it out, but it looks like there is a problem with building it:
```
make tools
go generate -tags tools tools/tools.go
go: github.com/golang-migrate/migrate/v4@v4.13.0 requires
        modernc.org/b@v1.0.0: reading modernc.org/b/go.mod at revision v1.0.0: git fetch -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /root/go/pkg/mod/cache/vcs/e4d822a6e5117a941675151e51943e806b98394310c31351bf5d1078e126484b: exit status 128:
        error: RPC failed; result=22, HTTP code = 404
        fatal: The remote end hung up unexpectedly
make: *** [tools] Error 1
```

I don't know if this is a known issue?

jefferai:
This looks like a (hopefully temporary) issue with a downstream dep of a dep.

I just ran `go clean modcache` followed by `make tools` and it worked for me, so give it another go!

tcassaert:
Works fine!

PR #730, opened 2020-10-19
docs: update changelog with worker controllers setting bug fix

malnick:
Updates changelog to reflect fix for https://github.com/hashicorp/boundary/issues/727 from https://github.com/hashicorp/boundary/pull/705.


CHANGELOG.md
```diff
@@ -20,7 +20,10 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 * controller, worker: Fix listening on IPv6 addresses
   ([Issue](https://github.com/hashicorp/boundary/issues/701))
   ([PR](https://github.com/hashicorp/boundary/pull/703))
+* dev mode worker: Fix setting controller address for worker in dev mode
```
jefferai, line 23:
```suggestion
* worker: Fix setting controller address for worker in dev mode
```

Mostly this is to try to keep the number of categories down since we already have `worker`. I'm not sure what level of specificity we want here. Thoughts?

PR #733, opened 2020-10-20
internal/servers/controller/handlers/accounts: fix dropped test errors

alrs:
This fixes two dropped `err` variables.


jefferai (approved):
Thanks!

Issue #729, opened 2020-10-19
Database init error when locale is not english


TanguyBaudrin:
Hi,

When testing Boundary, I faced an issue during the init database.

After digging a while the script told me that the database was already initialized, so I started looking at the code, and found this :

https://github.com/hashicorp/boundary/blob/b8c527c6cda63c37e96de6cd4c8fbdb9fe87b7e1/internal/cmd/commands/database/init.go#L257

My postgresql installation was not installed with an english locale (a french one), so the error message doesn't have the `does not exist` but the french equivalent `n'existe pas`

Maybe a good way to handle this would be to use an error code, not dependant of the locale ?

PR #736, opened 2020-10-20
Fix database init when locale isn't english 

jimlambrt:
Fixes https://github.com/hashicorp/boundary/issues/729


jefferai (approved):
Needs a Changelog entry, other than that looks good!

jimlambrt:
Added Changelog entry

PR #737, opened 2020-10-20
Allow authorize-session to be invoked with target name

jefferai:
This requires a combination of target name and either scope ID or scope name. Sets the stage for potential future relaxations around name as well.


jimlambrt (dismissed):
Looks great (needs tests).   The only thing that I got stuck on was the name "RoundTripValue" since it's not really round tripped to and from authResult(), but rather just a special return value in the result set.... it's a minor point.  Maybe something like "LookupValue"?

PR #738, opened 2020-10-21
Don't use error string comparison for schema_migrations table initialization

enmand:
The check for schema_migrations uses a database/sql.QueryContext func, which returns
errors directly from the driver. In this case, check the error code from Postgres,
and if it matches 42P01 (the error code for table not found), run the initialization.

Downside here, is it's now checking a concrete error type from the database driver
in the command code. This could probably be abstracted away, in the case for other
database servers.

resolves hashicorp/boundary#729


enmand:
Oh... well, looks like I was too slow, and it's already been fixed 😅

PR #744, opened 2020-10-22
docs: add KMS section to controller configuration overview

malnick:


jefferai (commented):
I think this looks good, do you want to also mention the optional block for config file decryption?

malnick:
Thanks @jefferai - added the `config` KMS block to both examples.

PR #749, opened 2020-10-26
docs: specify address for worker listener configuration

malnick:


website/content/docs/configuration/worker.mdx
```diff
@@ -23,6 +24,9 @@ cluster.
 
 - `description` - Specifies a friendly description of this worker.
 
+- `address` - Specifies the IP address for the worker listener to bind on. Defaults to
+127.0.0.1 if not specified.
```
jefferai, line 28:
```suggestion
127.0.0.1:9202 if not specified. If address is specified but only an IP address is given, the port will default to 9202.
```

jefferai (approved):

PR #755, opened 2020-10-28
feat: add official Dockerfile for Boundary

malnick:


docker/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+RUN addgroup boundary && \
+    adduser -s /bin/sh -S -G boundary boundary
+
+ADD https://releases.hashicorp.com/boundary/${VERSION}/boundary_${VERSION}_linux_amd64.zip /tmp/
+ADD https://releases.hashicorp.com/boundary/${VERSION}/boundary_${VERSION}_SHA256SUMS /tmp/
+ADD https://releases.hashicorp.com/boundary/${VERSION}/boundary_${VERSION}_SHA256SUMS.sig /tmp/ 
+
+RUN apk add --no-cache ca-certificates gnupg openssl libcap su-exec dumb-init tzdata 
+RUN cd /tmp/ && \
```
malnick, line 13: Need to add the pub GPG key here.

docker/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+    BUILD_GPGKEY=91A6E7F85D05C65630BEF18951852D87348FFC4C; \
+    found=''; \
+    for server in \
+        hkp://p80.pool.sks-keyservers.net:80 \
+        hkp://keyserver.ubuntu.com:80 \
+        hkp://pgp.mit.edu:80 \
+    ; do \
```
samsalisbury, line 28: I would probably just hardcode the key in this repo somewhere and import it, rather than be exposed to the internet to download it, since its fingerprint is already hardcoded. Not blocking though.

scripts/uigen.sh
```diff
@@ -13,8 +13,7 @@ fi
 (
 	cd "$UI_CLONE_DIR"
 	if ! docker-compose -f docker-compose-embedding.yml run build; then
-		echo "==> UI build failed."
-		exit 1
+    yarn install && yarn build
```
samsalisbury, line 16: I can't see this script referenced anywhere else in the PR, it seems like a reasonable fallback though. Note that for release builds, the UI compilation is [defined here](https://github.com/hashicorp/boundary/blob/main/packages-oss.yml#L75-L92), in order to not rely on Docker. We should probably have that call this script.

samsalisbury (approved):
No blockers from me, but see inline comments.

docker/Dockerfile
malnick: Totally hear that, I was torn between having consistency between our other Dockerfiles though.

docker/config.hcl
```diff
@@ -0,0 +1,65 @@
+disable_mlock = true
+
+controller {
+  name = "demo-controller-1"
+  description = "A controller for a demo!"
+
+  database {
+    url = "postgresql://postgres:postgres@0.0.0.0:5432/postgres?sslmode=disable"
```
malnick, line 8: Going to update this with the `env://` pattern to allow easy overrides from `-e` flag in docker run.

docker/Dockerfile
```diff
@@ -0,0 +1,46 @@
+FROM alpine:3.10
```
mdeggies, line 1: All we need to do is update this to `docker.mirror.hashicorp.services/alpine:3.10` to avoid the dockerhub rate limits.

docker/config.hcl
malnick: Updated

docker/Dockerfile
malnick: Updated, thanks!

PR #756, opened 2020-10-28
Add buf breaking change detection to makefile

talanknight:


jefferai:
Seems fine to me, but do we want to bump `buf` at the same time? We're on 0.24, 0.29 is current.

I can also include it in a more comprehensive dep upgrade but feels like it would fit in here too.

jefferai (approved):

PR #760, opened 2020-10-30
New domain errors based on the ICU-019 RFC

jimlambrt:
Once merged, there will be a fast following PR that moves the existing internal/db error functions and sentinels to their new package (as outlined in the RFC).  Once the two PRs are merged, we can start refactoring the Boundary to use these new domain error types.

Also, the values in the errorCodeInfo map and the const Codes include in this PR are just my idea of how we might might the existing internal/db sentinel errors to new domain errors.


internal/errors/error.go
```diff
@@ -0,0 +1,153 @@
...
+	return e.Wrapped
+}
+
+// Is the equivalent of the std errors.Is, but allows Devs to only import
+// this package for the capability.
+func Is(err, target error) bool {
+	return errors.Is(err, target)
+}
+
+// Ass is the equivalent of the std errors.As, and allows devs to only import
```
louisruch, line 149: Typo in comment: `s/Ass/As`

internal/errors/match.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// Info about the Template, which is useful when matching a Template's Kind with
+// an Err's Kind.
+func (t *Template) Info() Info {
+	if t == nil {
+		return errorCodeInfo[Unknown]
+	}
+	switch {
+	case t == nil:
```
louisruch, line 44: This case has already been checked in the `if t == nil` above, we should probably drop the `case` or the `if`

internal/errors/match.go
```diff
@@ -0,0 +1,99 @@
...
+			Kind:    t.Kind,
+		}
+	default:
+		return errorCodeInfo[Unknown]
+	}
+}
+
+// Error satisfies the error interface but we intentionally don't return
+// anything of value, in an effort to stop users from substituting Templates in
+// place of Errs, when creating domain errors.
```
louisruch, line 60: 👍

louisruch (commented):
A couple of comments but otherwise this LGTM

talanknight (commented):
This looks great!

internal/errors/error.go
jimlambrt: ty!

internal/errors/match.go
jimlambrt: ty!

internal/errors/error.go
```diff
@@ -0,0 +1,153 @@
...
+	// unfortunately, we can't help.
+	return nil
+}
+
+// Info about the Err
+func (e *Err) Info() Info {
+	if e == nil {
+		return errorCodeInfo[Unknown]
+	}
+	if info, ok := errorCodeInfo[e.Code]; ok {
```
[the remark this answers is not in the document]
jimlambrt: ty!

internal/errors/info.go
```diff
@@ -0,0 +1,47 @@
+package errors
+
+type Info struct {
+	Kind    Kind
+	Message string
+}
+
+// errorCodeInfo provides a map of unique Codes (IDs) to their
```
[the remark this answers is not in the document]
jimlambrt: I sort of didn't know the best place to put errCodeInfo since it's tightly coupled to Kind and Code... so I decide to just separate it from both of them.  We can always combine them if there seems to be a problem with drift.

internal/errors/error.go
```diff
@@ -0,0 +1,153 @@
...
+				return New(CheckConstraint, WithMsg(msg), WithWrap(ErrCheckConstraint)).(*Err)
+			default:
+				return New(NotSpecificIntegrity, WithMsg(pqError.Message)).(*Err)
+			}
+		}
+		if pqError.Code == "42P01" {
+			return New(MissingTable, WithMsg(pqError.Message)).(*Err)
+		}
+	}
+	// unfortunately, we can't help.
```
mgaffney, line 81: This comment cracked me up 🤣 !

internal/errors/error.go
```diff
@@ -0,0 +1,153 @@
...
+	if e == nil {
+		return nil
+	}
+	return e.Wrapped
+}
+
+// Is the equivalent of the std errors.Is, but allows Devs to only import
+// this package for the capability.
+func Is(err, target error) bool {
+	return errors.Is(err, target)
```
mgaffney, line 146: I noticed there is no test for this. Not sure it matters but thought I would mention it.

internal/errors/error.go
mgaffney: I'm still seeing this error. Have you pushed an update yet?

mgaffney (approved):

jimlambrt:
@mgaffney pushed the commits.  ty!

internal/errors/error.go
jimlambrt: Not sure what value a tests would bring to a pass-through to the stdlib, so I didn't add them

internal/errors/error.go
jimlambrt: pushed commits.  sorry for the delay.

internal/errors/error.go
jimlambrt: I'm here all week.  Try the buffet!

mgaffney (approved):

talanknight (approved):

fkarakas:
Hello, what is this RFC ? ICU-019 cant find anywhere

PR #761, opened 2020-10-30
Add ability to set public cluster address

### LINKED ISSUE #758: Worker is unable to mantain connection with Controller located in another network
**Describe the bug**

We have deployed an infrastructure based on: [boundary-reference-architecture](https://github.com/hashicorp/boundary-reference-architecture) but with several modifications including GCP infrastructure:

In AWS:

- Two public subnets and one private subnet
- One controller and one worker in a public subnet
- One target in the private subnet
- The LoadBalancer has one new target group pointing to 9201 port to controller

In GCP:

- One public subnet and one private subnet in GCP
- One worker in public subnet in GCP
- One target in the private subnet

We have zero problems when dealing with the AWS Infra, but the worker in GCP is unable to mantain connection to the controller in AWS. Configuration in GCP Instances is replicated from the AWS EC2.

In the following images, we can see that the GCP worker reach the AWS controller in the first place through the Load Balancer, but once the communication has been established, it is lost because the controller tells the worker to reach it using its private address (10.0.0.11 in this case). As both instances are in different network, this can't be accomplished (Journalctl is executed in the GCP worker instance):

journalctl

We have make it work executing the following command in the GCP worker instance:

```bash
iptables -t nat -A OUTPUT -d 10.0.0.11 -j DNAT --to-destination 54.239.32.132

## 10.0.0.11 is the private address of the controller, the IP that controller sends to worker
## 54.239.32.132 was the public IP address of the load balancer
```

Doing this, boundary works as expected, allowing us to ssh to the GCP target instance using boundary connect ssh.

We suppose that this is caused by the controller sending its private IP address. We think that a possible solution might be to separate private IP address to expose boundary service and external IP address to send to workers in different configuration parameters.

**To Reproduce**
Run boundary with configuration stated in [boundary-reference-architecture](https://github.com/hashicorp/boundary-reference-architecture) and add a worker in another network.

**Expected behavior**
Boundary Worker should be able to mantain the connection with the Controller.

jefferai:
Resolves #758

### DIFF
--- a/internal/cmd/commands/dev/dev.go
+++ b/internal/cmd/commands/dev/dev.go
@@ -48,6 +48,7 @@ type Command struct {
 	flagTargetSessionConnectionLimit int
 	flagControllerAPIListenAddr      string
 	flagControllerClusterListenAddr  string
+	flagControllerPublicClusterAddr  string
 	flagWorkerProxyListenAddr        string
 	flagWorkerPublicAddr             string
 	flagPassthroughDirectory         string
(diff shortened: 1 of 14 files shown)


internal/cmd/commands/dev/dev.go
```diff
@@ -163,6 +164,13 @@ func (c *Command) Flags() *base.FlagSets {
 		Usage:  "Address to bind to for controller \"cluster\" purpose. If this begins with a forward slash, it will be assumed to be a Unix domain socket path.",
 	})
 
+	f.StringVar(&base.StringVar{
+		Name:   "controller-public-cluster-address",
```
malnick, line 168: I know it makes it long, but do we want this to be `dev-controller-public-cluster-address` to mirror it with the env var?

malnick (approved):
One comment on the flag name for the dev controller public cluster addr, but other than that, LGTM.

internal/cmd/commands/dev/dev.go
jefferai: No, none of these flags have dev in front because they're only valid for the dev command. The reason the env vars have DEV in front is because they are otherwise not indicated which command they are related to. Similar to e.g. BOUNDARY_CONNECT_XYZ.

PR #763, opened 2020-10-31
internal/servers/controller/handlers/groups: fix dropped test error

alrs:
This fixes a dropped test error.


jimlambrt (approved):
Thank you!

PR #768, opened 2020-11-02
Use ProductDownloader component, remove duplicate code

ztanner:
Uses the latest ProductDownloader component ( @ v6) to render the downloads page.


website/pages/downloads/style.module.css
```diff
@@ -1,61 +1,7 @@
...
-    text-decoration: underline;
-  }
-
-  & h1 {
-    text-align: center;
-  }
-
-  --brand: var(--red);
-  --brand-d1: #c03e43;
+  --red-d1: #c03e43;
```
BrandonRomano, line 2: Good call here. 

This is semi-fragile in terms of the component usage for the Boundary theme (as the component is dependent on this external declaration) -- but not really a real problem, because this is the only place where this will exist.

I have a task on my plate to clean up this theming, so I'll get to it eventually.

BrandonRomano (approved):

malnick:
@BrandonRomano Do you want me to merge this?

BrandonRomano (approved):

PR #769, opened 2020-11-03
docs: update installation steps with templated account.id grant string

xntrik:
My nvim setup appears to have removed a bunch of trailing spaces as well. The only change I was interested in addressing was at line 345.


xntrik:
I'm also unsure if this should be merged into `stable-website` or `main`

PR #774, opened 2020-11-04
Refactor existing sentinel errors 

jimlambrt:
This PR refactors Boundary to use the sentinel errors under internal/errors.  These sentinel errors used to be under internal/db.

Apologies for the messy commit history (it will all get squashed, so I didn't worry too much about it).


internal/kms/kms.go
```diff
@@ -2,7 +2,7 @@ package kms
 
 import (
 	"context"
-	"errors"
```
talanknight, line 5: Very small nit: Since this errors package import doesn't conflict yet, should we not put an import alias?

talanknight (approved):
LGTM.  Just 1 super small nit that you can ignore if you want. :)

internal/kms/kms.go
jimlambrt: I was just trying to be consistent, so devs could find the std lib errors pkg imports.

mgaffney (approved):

PR #777, opened 2020-11-05
feat: make auth token staleness and duration configurable

malnick:


internal/cmd/base/servers.go
```diff
@@ -223,6 +223,16 @@ func (b *Server) PrintInfo(ui cli.Ui) {
 		b.Info["cgo"] = "enabled"
 	}
 
+	var (
+		tokDurKey = "auth token max duration"
```
jefferai, line 227: I'm not sure we want to add these values -- at least not yet. Or maybe we do and will want to wordsmith them later. I'm basically thinking once this is configurable per-auth-method (assuming) this might end up being misleading, especially since we haven't decided the rules on what would override what.

I could go either way, we can leave this in now if you want, but wanted to raise that.

internal/cmd/base/servers.go
```diff
@@ -223,6 +223,16 @@ func (b *Server) PrintInfo(ui cli.Ui) {
...
+		tokStaKey = "auth token max staleness"
+	)
+
+	b.InfoKeys = append(b.InfoKeys, tokDurKey)
+	b.Info[tokDurKey] = fmt.Sprintf("%gs", globals.DefaultAuthTokenMaxDuration.Seconds())
```
jefferai, line 232: Please use `%ds`, or use `>String()` instead if you want the Go-style duration string.

internal/cmd/commands/dev/dev.go
```diff
@@ -378,6 +380,25 @@ func (c *Command) Run(args []string) int {
 	c.InfoKeys = append(c.InfoKeys, "worker public addr")
 	c.Info["worker public addr"] = c.Config.Worker.PublicAddr
 
+	// Check for auth token overrides
+	if c.Config.Controller.AuthTokenMaxDuration != "" {
```
jefferai, line 384: We don't actually parse a config in dev mode so this code will never actually do anything. You may want to instead add flags to control this behavior and have the flag parsing happen here. Or you can just remove this entirely for the moment if you don't want flags at this time.

internal/cmd/commands/server/server.go
```diff
@@ -159,6 +160,27 @@ func (c *Command) Run(args []string) int {
 		globals.DefaultMaxRequestDuration = c.Config.DefaultMaxRequestDuration
 	}
 
+	// Perform controller configuration overrides for auth token settings
+	if c.Config.Controller != nil {
```
jefferai, line 164: Rather than put this code here, please instead update the configuration parsing (which then means any given HCL config will always work for this purpose even without having to go through the block here in the server command). The general way to do this is to have a raw interface{} value for the actual HCL to be decoded into, then call parseutil.ParseDurationSecond to convert what came in (which can be string or integer) into an integer number of seconds. Then you know that after config parsing the number will always either be zero or an integer number of seconds.

internal/authtoken/repository.go
```diff
@@ -91,10 +89,9 @@ func (r *Repository) CreateAuthToken(ctx context.Context, withIamUser *iam.User,
 		return nil, fmt.Errorf("create: unable to get database wrapper: %w", err)
 	}
 
-	// TODO: Allow the caller to specify something different than the default duration.
 	// We truncate the expiration time to the nearest second to make testing in different platforms with
 	// different time resolutions easier.
-	expiration, err := ptypes.TimestampProto(time.Now().Add(maxTokenDuration).Truncate(time.Second))
+	expiration, err := ptypes.TimestampProto(time.Now().Add(globals.DefaultAuthTokenMaxDuration).Truncate(time.Second))
```
jefferai, line 94: For testing purposes it's much better if this is simply a property of the controller (via its RawConfig) that can then be passed through to any auth methods via the auth service handler. So rather than have globals, when the controller constructs the auth method service handler, it should provide the values from its configuration.

This does mean updating quite a bit more code elsewhere, although one way around this (and likely better overall) is to add an Option capability to the constructors for the service handler and the repositories that allow this to be passed through, or not.

internal/cmd/config/config.go
```diff
@@ -95,10 +97,12 @@ type Config struct {
...
-	Name              string    `hcl:"name"`
-	Description       string    `hcl:"description"`
-	Database          *Database `hcl:"database"`
-	PublicClusterAddr string    `hcl:"public_cluster_addr"`
+	Name                  string    `hcl:"name"`
+	Description           string    `hcl:"description"`
+	Database              *Database `hcl:"database"`
+	PublicClusterAddr     string    `hcl:"public_cluster_addr"`
+	AuthTokenMaxDuration  string    `hcl:"auth_token_max_duration"`
+	AuthTokenMaxStaleness string    `hcl:"auth_token_max_staleness"`
```
jefferai, line 105: This isn't really a "max" staleness value -- or if it is, I don't think that's something that's been discussed, because it implies overriding from auth methods isn't ever possible. That also applies to the MaxDuration above.

Likely these need to be something like `default_auth_token_ttl` and `default_auth_token_staleness_duration` or something like that.

internal/authtoken/repository.go
malnick: This was something I wasn't sure of: use a global or pass it in as an option. I went with the former because of brevity, but I think the appropriate pattern to set is using a functional option on the auth token type which is what _I think_ you're pointing to. A bit more code wrangling but a better long term pattern, does that sound right?

internal/cmd/config/config.go
malnick: Yeah the wording here was a little tricky. I based the current implementation off the existing wording for the vars we formerly used in the repo code, but I'm happy to swap these out for something more agreeable. 
 
Do we want to use the word `default` in the attribute name? I was thinking the setting would _override the default_ we set in the code. I like:
- `auth_token_ttl`
- `auth_token_staleness_duration`

internal/authtoken/repository.go
jefferai: Yep, exactly -- more code now but better long term pattern. As the comment in globals.go says, globals are bad, and we can definitely avoid them in this case. However, globals can provide the default! So feel free to keep values there, just use that as a default for if config values aren't specified -- or maybe even better, move them to the config package so they live next to the parsing code. But that way you can always be sure that a config's value will end up with some concrete >0 value regardless of it it was the default or not.

internal/cmd/base/servers.go
malnick: I removed it for now because it wasn't totally clear how to plumb through the setting from the configuration file or default to the setting we use in `authtoken.defaultOptions`. I think it would be nice to see it on startup at some point.

internal/cmd/base/servers.go
malnick: Fixed.

internal/cmd/commands/server/server.go
malnick: Added this to `config.Parse()`, thanks!

internal/cmd/config/config.go
```diff
@@ -31,6 +32,8 @@ telemetry {
 controller {
 	name = "dev-controller"
 	description = "A default controller created in dev mode"
+	auth_token_max_duration = "2h"
+	auth_token_max_staleness = "1h"
```
malnick, line 36: These were for testing and need to be removed.

internal/authtoken/repository_test.go
```diff
@@ -63,10 +65,12 @@ func TestRepository_New(t *testing.T) {
...
-				reader:       rw,
-				writer:       rw,
-				kms:          kmsCache,
-				defaultLimit: 5,
+				reader:              rw,
+				writer:              rw,
+				kms:                 kmsCache,
+				defaultLimit:        5,
+				timeToLiveDuration:  7 * 24 * time.Hour,
+				timeToStaleDuration: 24 * time.Hour,
```
malnick, line 73: Should add a non-exported constant for these default values in options.go and then use that constant here.

internal/cmd/config/config_test.go
```diff
@@ -64,8 +64,10 @@ func TestDevController(t *testing.T) {
 			},
 		},
 		Controller: &Controller{
-			Name:        "dev-controller",
-			Description: "A default controller created in dev mode",
+			Name:                 "dev-controller",
+			Description:          "A default controller created in dev mode",
+			AuthTokenTimeToLive:  "2h",
+			AuthTokenTimeToStale: "1h",
```
malnick, line 70: These values should be deleted with the above testing code being removed.

internal/cmd/config/config_test.go
```diff
@@ -78,6 +80,8 @@ func TestDevController(t *testing.T) {
 	exp.DevControllerKey = actual.Seals[0].Config["key"]
 	exp.DevWorkerAuthKey = actual.Seals[1].Config["key"]
 	exp.DevRecoveryKey = actual.Seals[2].Config["key"]
+	exp.Controller.AuthTokenTimeToLive = actual.Controller.AuthTokenTimeToLive
+	exp.Controller.AuthTokenTimeToStale = actual.Controller.AuthTokenTimeToStale
```
malnick, line 84: Same here.

internal/cmd/commands/dev/dev.go
malnick: Fixed

internal/authtoken/repository.go
malnick: Fixed

internal/authtoken/repository.go
```diff
@@ -45,15 +49,27 @@ func NewRepository(r db.Reader, w db.Writer, kms *kms.Kms, opt ...Option) (*Repo
 	}
 
 	opts := getOpts(opt...)
+
 	if opts.withLimit == 0 {
 		// zero signals the boundary defaults should be used.
 		opts.withLimit = db.DefaultLimit
 	}
+
+	if opts.withTokenTimeToLiveDuration == 0 {
```
jefferai, line 58: Don't do this here. Have the `getDefaultOptions` function do what it's meant to do, as we do in other packages.

internal/authtoken/repository.go
```diff
@@ -30,6 +29,11 @@ type Repository struct {
 	kms    *kms.Kms
 	// defaultLimit provides a default for limiting the number of results returned from the repo
 	defaultLimit int
+
+	// defaultTimeToLive and defaultTimeToStale durations ensure we do not accidentally set TTLs
+	// of zero value
+	timeToLiveDuration  time.Duration
```
jefferai, line 35: Don't add these here -- any function that needs to make use of this should instead take in options, which would provide the defaults if nothing is explicitly set.

internal/authtoken/repository.go
```diff
@@ -91,10 +107,9 @@ func (r *Repository) CreateAuthToken(ctx context.Context, withIamUser *iam.User,
 		return nil, fmt.Errorf("create: unable to get database wrapper: %w", err)
 	}
 
-	// TODO: Allow the caller to specify something different than the default duration.
 	// We truncate the expiration time to the nearest second to make testing in different platforms with
 	// different time resolutions easier.
-	expiration, err := ptypes.TimestampProto(time.Now().Add(maxTokenDuration).Truncate(time.Second))
+	expiration, err := ptypes.TimestampProto(time.Now().Add(r.timeToLiveDuration).Truncate(time.Second))
```
jefferai, line 112: This should come from options, not a repo var.

internal/cmd/config/config.go
```diff
@@ -99,6 +100,15 @@ type Controller struct {
 	Description       string    `hcl:"description"`
 	Database          *Database `hcl:"database"`
 	PublicClusterAddr string    `hcl:"public_cluster_addr"`
+
+	// AuthTokenTimeToLive is the total valid lifetime of a token denoted by time.Duration
+	AuthTokenTimeToLive         string `hcl:"auth_token_time_to_live"`
```
jefferai, line 105: This shouldn't be a string, it should be an `interface{}` and use ParseDuration, like we do elsewhere.

internal/authtoken/repository.go
malnick: @talanknight and I went back and forth on this a bit. The reason we chose this route was that TTL isn't optional, it's required. So in the end, we decided to set it on the repo the same way we do for level because we couldn't think of a reason you'd ever want a 0 TTL.

jefferai: You don't need to store those in the repository itself --  you only need to override the local package var if the value is set. That's in theory racy but in practice you should only be set at startup, and only if the option coming in is non-zero. That allows you to remove a bunch of this boilerplate here and in tests.

malnick: cleaned this up

internal/authtoken/options.go
```diff
@@ -14,12 +25,18 @@ type Option func(*options)
...
+	withTokenValue               bool
+	withTokenTimeToLiveDuration  time.Duration
+	withTokenTimeToStaleDuration time.Duration
+	withLimit                    int
 }
 
 func getDefaultOptions() options {
-	return options{}
+	return options{
+		withLimit:                    db.DefaultLimit,
```
jefferai, line 36: I don't think we need or want to do this. If we don't have this option set, the DB will already set this to the default. There's no reason to also have this codified here.

internal/authtoken/options.go
```diff
@@ -1,5 +1,16 @@
...
 
+import (
+	"time"
+
+	"github.com/hashicorp/boundary/internal/db"
+)
+
+var (
+	defaultTokenTimeToLiveDuration  = 7 * 24 * time.Hour
+	defaultTokenTimeToStaleDuration = 7 * time.Hour
```
jefferai, line 11: Note that you have changed this value from 24 hours to 7 hours. That's quite a drastic change. Do we actually want this?

internal/authtoken/options.go
jefferai: Oh I see, this was already in NewRepository. I think probably that code in and of itself was already not correct. @jimlambrt thoughts? Is there a reason to set to db.DefaultLimit explicitly here or in the repo code, or will that automatically be used by the underlying db code?

internal/authtoken/repository.go
jefferai: This is still coming from the repo var.

internal/authtoken/repository.go
```diff
@@ -211,7 +207,7 @@ func (r *Repository) ValidateToken(ctx context.Context, id, token string, opt ..
 	sinceLastAccessed := now.Sub(lastAccessed) + timeSkew
 	// TODO (jimlambrt 9/2020) - investigate the need for the timeSkew and see
 	// if it can be eliminated.
-	if now.After(exp.Add(-timeSkew)) || sinceLastAccessed >= maxStaleness {
+	if now.After(exp.Add(-timeSkew)) || sinceLastAccessed >= r.timeToStaleDuration {
```
jefferai, line 210: This should come from options.

internal/cmd/commands/connect/connect.go
```diff
@@ -257,12 +257,15 @@ func (c *Command) Run(args []string) (retCode int) {
 	}
 
 	switch {
-	case c.flagAuthzToken != "" && c.flagTargetId != "":
-		c.UI.Error(`-target-id and -authz-token cannot both be specified`)
-		return 1
-	case c.flagAuthzToken != "" && c.flagTargetName != "":
-		c.UI.Error(`-target-name and -authz-token cannot both be specified`)
-		return 1
+	case c.flagAuthzToken != "":
```
jefferai, line 260: I think you were planning on merging main into this branch so this wouldn't be here

internal/authtoken/options.go
malnick: Previously, when you called NewRepo(), that would set this default if the value was zero. We agreed to move that logic into options.go getDefaultOptions(). If we don't set this here, we will have a 0 value.

internal/authtoken/options.go
malnick: Good catch, fixed

internal/cmd/commands/connect/connect.go
malnick: Fixed

internal/authtoken/options.go
jefferai: Right -- see my second comment above as to whether we actually wanted that behavior in the first place. Could use Jim's opinion here.

website/content/docs/configuration/controller.mdx
```diff
@@ -40,6 +40,13 @@ used by workers after initial connection to controllers via the worker's
 bind a publicly accessible IP to a NIC on the host directly, such as an Amazon
 EIP.
 
+- `auth_token_time_to_live` - Maximum time to live (TTL) for all auth tokens globally (pertains
+to all tokens from all auth methods). Valid time units are anything specified by Golang's 
+[ParseDuration()](https://golang.org/pkg/time/#ParseDuration) method. Default is 7 days.
+- `auth_token_time_to_stale` - Maximum time of inactivity for all auth tokens globally (pertains
+to all tokens from all auth methods). Valid time units are anything specified by Golang's 
+[ParseDuration()](https://golang.org/pkg/time/#ParseDuration) method. Default is 1 day.
+
```
jefferai, line 49:
```suggestion
- `auth_token_time_to_live` - Maximum time to live (TTL) for all auth tokens globally (pertains
to all tokens from all auth methods). Valid time units are an integer number of seconds or a string specified by Golang's 
[ParseDuration()](https://golang.org/pkg/time/#ParseDuration) method. Default is 7 days.
- `auth_token_time_to_stale` - Maximum time of inactivity for all auth tokens globally (pertains
to all tokens from all auth methods). Valid time units are an integer number of seconds, or a string specified by Golang's 
[ParseDuration()](https://golang.org/pkg/time/#ParseDuration) method. Default is 1 day.

```

jefferai (approved):

Issue #781, opened 2020-11-06
Commands add-grants and add-principals do not exist


rbrayner:
**Describe the bug**
Commands add-grants and add-principals do not exist, when following the docs: https://www.boundaryproject.io/docs/installing/no-gen-resources

**To Reproduce**
Run the command `boundary`

<img width="1249" alt="image" src="https://user-images.githubusercontent.com/11684945/98324469-1a009500-1fcb-11eb-9d37-709df550347e.png">

**Expected behavior**
The help menu must describe the add-grants and add-principals commands.

**Additional context**
Version information:
  Git Revision:        eccd68d73c3edf14863ecfd31f9023063b809d5a
  Version Number:      0.1.1

jimlambrt:
grants and principals are a subcommand of the roles command:

```
❯ boundary roles help
Usage: boundary roles [sub command] [options] [args]

  This command allows operations on Boundary role resources. Example:

    Create a role:

      $ boundary roles create -name prodops -description "For ProdOps usage"

  Please see the roles subcommand help for detailed usage information.

Subcommands:
    add-grants           Add grants to a role
    add-principals       Add principals (users, groups) to a role
    create               Create a role
    delete               Delete a role
    list                 List a role
    read                 Read a role
    remove-grants        Remove grants from a role
    remove-principals    Remove principals (users, groups) from a role
    set-grants           Set the full contents of the grants on a role
    set-principals       Set the full contents of the principals (users, groups) on a role
    update               Update a role
```

jimlambrt:
I believe these subcommand have examples on the page you referenced from the docs: https://www.boundaryproject.io/docs/installing/no-gen-resources#anonymous-listing-role-for-global-scope

jimlambrt:
Hum... just re-read the docs and you're correct, the command examples aren't quite right.

PR #782, opened 2020-11-06
fix docs for adding roles and principals

jimlambrt:
fixes https://github.com/hashicorp/boundary/issues/781


malnick (approved):
Thank you!

PR #784, opened 2020-11-09
Change Format of API Error

talanknight:
This changes the format of the returned API errors by removing a few redundant fields.
* The http status code was removed in favor of just looking at the HTTP response
* Moved the request_id and TraceId out of the error to be read through a header eventually.  
* The error_id was removed in favor of a TraceId header message (currently only returned for 500s if the request doesn't include one).  This header is subject to change as OpenTelemetry is integrated with Boundary.

This is proto wire breaking change but since the API error proto is only used for json generation this should be ok.

Since there is quite a few auto-generated files in this PR please pay special attention to the following files:
* api/apierror.go
* internal/api/genapi/templates.go
* internal/cmd/base/format.go
* internal/proto/local/controller/api/v1/error.proto
* internal/servers/controller/handler.go
* internal/servers/controller/handlers/errors.go


api/response.go
```diff
@@ -36,15 +36,10 @@ func (r *Response) Decode(inStruct interface{}) (*Error, error) {
...
-			Status:  http.StatusForbidden,
-			Message: "Forbidden",
-		}, nil
+		return ErrPermissionDenied, nil
 	}
 
-	apiErr := &Error{
-		Status: int32(r.resp.StatusCode),
-	}
+	apiErr := &Error{}
```
louisruch, line 42: Small nit: it seems the need for this variable falls away with these changes.  It took me some time to figure out exactly what we are doing with it here and perhaps a slightly more readable option:
1) on line 54 `instruct = &Error{}` // perhpas with a comment, but ok if not
2) on line 70 `return nil, &Error{response: r}`

louisruch (approved):
On small nit otherwise this LGTM

louisruch (approved):

louisruch (approved):
9a9263f Changes LGTM

internal/servers/controller/handlers/errors.go
```diff
@@ -23,11 +22,12 @@ const (
...
 
 type apiError struct {
-	inner *pb.Error
+	status int32
+	inner  *pb.Error
 }
 
 func (e *apiError) Error() string {
-	res := fmt.Sprintf("Status: %d, Code: %q, Error: %q", e.inner.GetStatus(), e.inner.GetCode(), e.inner.GetMessage())
+	res := fmt.Sprintf("Status: %d, Code: %q, Error: %q", e.status, e.inner.GetKind(), e.inner.GetMessage())
```
jefferai, line 30:
```suggestion
	res := fmt.Sprintf("Status: %d, Kind: %q, Error: %q", e.status, e.inner.GetKind(), e.inner.GetMessage())
```

internal/servers/controller/handlers/errors.go
```diff
@@ -175,14 +174,16 @@ func ErrorHandler(logger hclog.Logger) runtime.ErrorHandlerFunc {
...
-			errId, err := internalErrorId()
-			if err != nil {
-				logger.Error("unable to generate internal error id", "error", err)
-				errId = "failed_to_generate_error_id"
-			}
-			logger.Error("internal error returned", "error id", errId, "error", inErr)
-			apiErr = getInternalError(errId)
+		traceId := r.Header.Get("TraceId")
+		if traceId == "" {
+			traceId = "failed_to_get_trace_id"
```
jefferai, line 179: If there's no trace ID, this should be blank, not a non-valid trace ID value

talanknight: I removed the trace id generation and instead we return the full error message with 500 responses.

louisruch (dismissed):

internal/servers/controller/handler.go
```diff
@@ -175,6 +176,16 @@ func handleGrpcGateway(c *Controller, props HandlerProperties) (http.Handler, er
 	return mux, nil
 }
 
+// generatedTraceId returns a boundary generated TraceId or "" if an error occurs when generating
+// the id.
+func generatedTraceId() string {
```
jefferai, line 181: This can probably be removed for now

jefferai (approved):

PR #785, opened 2020-11-09
Refactor CreateHost -> domain errors

louisruch:
### What does this PR do ### 

Refactors host repo, currently only the create host path, to the new boundary domain errors.

### Testing ### 

Looking for initial feedback at this point to make sure I am on the right path, will add unit tests and include manual testing results shortly

### Follow-up work ###

Testing (as described above)
Refactor remaining host paths, and verify full create host refactor

### DIFF
--- a/internal/errors/option.go
+++ b/internal/errors/option.go
@@ -16,7 +16,7 @@ type Option func(*Options)
 type Options struct {
 	withErrWrapped error
 	withErrMsg     string
-	withOp         Op
+	withOp         *Op
 }
 
 func getDefaultOptions() Options {
@@ -43,6 +43,6 @@ func WithMsg(msg string) Option {
 // the error.
 func WithOp(op Op) Option {
 	return func(o *Options) {
-		o.withOp = op
+		o.withOp = &op
 	}
 }
--- a/internal/host/static/repository_host.go
+++ b/internal/host/static/repository_host.go
@@ -23,23 +23,23 @@ import (
 // unique within h.CatalogId.
 func (r *Repository) CreateHost(ctx context.Context, scopeId string, h *Host, opt ...Option) (*Host, error) {
 	if h == nil {
-		return nil, fmt.Errorf("create: static host: %w", errors.ErrInvalidParameter)
+		return nil, errors.New(errors.InvalidParameter, errors.WithMsg("nil host"))
 	}
 	if h.Host == nil {
-		return nil, fmt.Errorf("create: static host: embedded Host: %w", errors.ErrInvalidParameter)
+		return nil, errors.New(errors.InvalidParameter, errors.WithMsg("embedded Host"))
 	}
 	if h.CatalogId == "" {
-		return nil, fmt.Errorf("create: static host: no catalog id: %w", errors.ErrInvalidParameter)
+		return nil, errors.New(errors.InvalidParameter, errors.WithMsg("no catalog id"))
 	}
 	if h.PublicId != "" {
-		return nil, fmt.Errorf("create: static host: public id not empty: %w", errors.ErrInvalidParameter)
+		return nil, errors.New(errors.InvalidParameter, errors.WithMsg("public id not empty"))
 	}
 	if scopeId == "" {
-		return nil, fmt.Errorf("create: static host: no scopeId: %w", errors.ErrInvalidParameter)
+		return nil, errors.New(errors.InvalidParameter, errors.WithMsg("no scopeId"))
 	}
 	h.Address = strings.TrimSpace(h.Address)
 	if len(h.Address) < MinHostAddressLength || len(h.Address) > MaxHostAddressLength {
-		return nil, fmt.Errorf("create: static host: bad address: %w", ErrInvalidAddress)
+		return nil, errors.New(errors.InvalidParameter, errors.WithMsg("bad address"))
 	}
 	h = h.clone()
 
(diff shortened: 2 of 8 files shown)


internal/errors/code.go
```diff
@@ -22,6 +22,7 @@ const (
 
 	// General function errors are reserved Codes 100-999
 	InvalidParameter Code = 100 // InvalidParameter represents and invalid parameter for an operation.
+	Internal         Code = 101 // Internal represents internal errors
```
louisruch, line 25: There were a number of errors in the host path that I felt where not specific errors and could not be addressed by the end user, so I felt we should create some sort of internal error when it truely is an internal error.

internal/errors/info.go
```diff
@@ -12,6 +12,10 @@ var errorCodeInfo = map[Code]Info{
 		Message: "unknown",
 		Kind:    Other,
 	},
+	Internal: {
+		Message: "internal error",
+		Kind:    Parameter,
```
louisruch, line 17: I just selected `Parameter` but we should probably add a new Kind for these

internal/errors/is.go
```diff
@@ -6,6 +6,16 @@ import (
 	"github.com/lib/pq"
 )
 
+// IsDomainError returns a boolean indicating whether the error is
+// a boundary domain error
+func IsDomainError(err error) bool {
```
louisruch, line 11: Helper function to avoid needing to do the four lines each time, open to alternatives

internal/errors/option.go
```diff
@@ -46,3 +52,22 @@ func WithOp(op Op) Option {
 		o.withOp = op
 	}
 }
+
+// WithCallerOp sets the `withOp` field to the caller's functions name; formatted package.func.
+// Optionally, with fullpath=true, it will include the full package path.
+// If caller lookup fails this will return an empty string
+func WithCallerOp(fullpath bool) Option {
```
louisruch, line 59: Helper function so we can have CallersFrames do the lifting for us here

internal/errors/is.go
talanknight: If you look at Convert inside error.go it will return either the domain error with the proper type or it will return nil so instead of using errors.As you could use

if dErr := errors.Convert(err); err != nil { ... }

Maybe we should update the IsX() methods below to use Convert.

internal/errors/option.go
louisruch: Some idea of the output of this with these changes and forced errors:
fullpath=true
```
Error from controller when performing create on static-type host: 
Error information:
  Code:                InvalidArgument
  Message:             github.com/hashicorp/boundary/internal/host/static.(*Repository).CreateHost:(54): create: static host: parameter violation: error #100
  Status:              400
```

fullpath=false
```
Error from controller when performing create on static-type host: 
Error information:
  Code:                InvalidArgument
  Message:             static.(*Repository).CreateHost (54): create: static host: parameter violation: error #100
  Status:              400
```

internal/host/static/repository_host.go
```diff
@@ -23,44 +23,46 @@ import (
 // unique within h.CatalogId.
 func (r *Repository) CreateHost(ctx context.Context, scopeId string, h *Host, opt ...Option) (*Host, error) {
 	if h == nil {
-		return nil, fmt.Errorf("create: static host: %w", errors.ErrInvalidParameter)
+		return nil, errors.New(errors.Internal, errors.WithMsg("nil host"), errors.WithCallerOp(false))
```
talanknight, line 26: Why make this error an Internal error while the others are InvalidParameters?  

It is likely most of these should be treated as Internal errors by the API (for example: if we are not populating the catalog id for a host it should be caught prior to this point and to call the repository with a host without a catalog id defined is an error in our code).  I don't think we want to wrap all errors that will appear as internal in the API as internal for the domain api.  It feels to early to remove that information since it still needs to be returned and possibly reacted to by the caller.  Thoughts?

internal/host/static/repository_host.go
```diff
@@ -73,14 +75,18 @@ func (r *Repository) CreateHost(ctx context.Context, scopeId string, h *Host, op
...
-			return nil, fmt.Errorf("create: static host: in catalog: %s: name %s already exists: %w",
-				h.CatalogId, h.Name, errors.ErrNotUnique)
+			return nil, errors.New(errors.NotUnique,
+				errors.WithMsg(fmt.Sprintf("in catalog: %s: name %s already exists", h.CatalogId, h.Name)),
+				errors.WithCallerOp(false))
 		}
 		if errors.IsCheckConstraintError(err) || errors.IsNotNullError(err) {
-			return nil, fmt.Errorf("create: static host: in catalog: %s: %q: %w",
-				h.CatalogId, h.Address, ErrInvalidAddress)
+			return nil, errors.New(errors.NotNull,
```
talanknight, line 83: There are a few potential constraints that could be failing here, one of which would be the length of the address. (constraints address_must_be_more_than_2_characters and address_must_be_less_than_256_characters are the names of those constraints)  I wonder if we want to have different errors for the length validation checks and other validation checks (like not null)

internal/servers/controller/handlers/hosts/host_service.go
```diff
@@ -176,7 +176,7 @@ func (s Service) createInRepo(ctx context.Context, scopeId, catalogId string, it
 	}
 	out, err := repo.CreateHost(ctx, scopeId, h)
 	if err != nil {
-		if errors.IsUniqueError(err) || errors.Is(err, errors.ErrNotUnique) {
+		if errors.IsDomainError(err) || errors.IsUniqueError(err) || errors.Is(err, errors.ErrNotUnique) {
```
jimlambrt, line 179: do we still need to check for IsUniqueError() and Is errors.ErrNotUnique?  I think, you can just make sure it's errors.Convert(err) != nil

jimlambrt (commented):
I think, db.NewPublicId needs to return a domain error (if you're doing a complete vertical slice of the implementation).  I didn't look at other internal/db calls used by host/static but this one sort of popped up during my review.

jimlambrt:
I didn't see that any of the host/static tests were updated... I thought we might see some updates to use things like errors.Match

jimlambrt (changes requested):
Nice work!   

I didn't go through and comment on every instance of internal error...

louisruch:
> 
> 
> I didn't see that any of the host/static tests were updated... I thought we might see some updates to use things like errors.Match

Yeah I specifically left out test updates/additions for now as I wanted to get some initial feedback before going through that

internal/errors/option.go
```diff
@@ -46,3 +52,22 @@ func WithOp(op Op) Option {
...
+	pc, _, line, ok := runtime.Caller(1)
```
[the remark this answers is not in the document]
jimlambrt: Hum... I wonder if "with caller" should be the default and we should just have and option for the times you don't want the Op to equal the caller.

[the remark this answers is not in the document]
louisruch: We could update `WithOp` to have it do the `withCaller` workflow when it receives an empty string.  The dev can still create an empty Op error by just not calling `WithOp`

internal/errors/is.go
jimlambrt: Hum... I may change my mind about using errors.Convert in this use-case.  Convert will only work if the type == errors.Err, and errors.As will work for this use-case and for wrapped domain errors.  I guess, it depends on how you want this to work

talanknight: Isn't a wrapped errors.Err still a errors.Err?  Maybe Convert() should handle unwrapping it for you.

jimlambrt: Convert() could easily unwrap it and that's probably the right solution.  Right now, it's just attempting a cast to *Err

louisruch:
Still a WIP will ping when ready for review

louisruch:
Closing this to create PR of boundary branch instead of my forked branch until the CircleCI fixes have been decided upon

PR #786, opened 2020-11-12
Refactor host_repo -> domain errors

louisruch:
### What does this PR do  

Updates domain errors to not print redundant error code info, i.e. only prints error code info if the wrapped error does not exist or is a different code
Removes new line from domain error stringer
Refactors create host path to domain errors
Refactors errors.New -> errors.E
Adds a New and Wrap helper functions

### Testing  

Updated unit tests and did a number of manual tests with errors injected

### Follow-up work 

Refactor remaining host repo

### DIFF
--- a/internal/db/id.go
+++ b/internal/db/id.go
@@ -18,11 +18,11 @@ func NewPublicId(prefix string) (string, error) {
 
 func newId(prefix string) (string, error) {
 	if prefix == "" {
-		return "", fmt.Errorf("missing prefix %w", errors.ErrInvalidParameter)
+		return "", errors.New(errors.InvalidParameter, errors.WithMsg("missing prefix"))
 	}
 	publicId, err := base62.Random(10)
 	if err != nil {
-		return "", fmt.Errorf("unable to generate id: %w", err)
+		return "", errors.Wrap(err, errors.WithMsg("unable to generate id"))
 	}
 	return fmt.Sprintf("%s_%s", prefix, publicId), nil
 }
--- a/internal/db/id.go
+++ b/internal/db/id.go
@@ -17,12 +17,13 @@ func NewPublicId(prefix string) (string, error) {
 }
 
 func newId(prefix string) (string, error) {
+	const op = errors.Op("db.newId")
 	if prefix == "" {
-		return "", errors.New(errors.InvalidParameter, errors.WithMsg("missing prefix"))
+		return "", errors.New(errors.InvalidParameter, errors.WithOp(op), errors.WithMsg("missing prefix"))
 	}
 	publicId, err := base62.Random(10)
 	if err != nil {
-		return "", errors.Wrap(err, errors.WithMsg("unable to generate id"))
+		return "", errors.Wrap(err, errors.WithOp(op), errors.WithMsg("unable to generate id"))
 	}
 	return fmt.Sprintf("%s_%s", prefix, publicId), nil
 }
--- a/internal/db/id.go
+++ b/internal/db/id.go
@@ -17,7 +17,7 @@ func NewPublicId(prefix string) (string, error) {
 }
 
 func newId(prefix string) (string, error) {
-	const op = errors.Op("db.newId")
+	const op = "db.newId"
 	if prefix == "" {
 		return "", errors.New(errors.InvalidParameter, errors.WithOp(op), errors.WithMsg("missing prefix"))
 	}
--- a/internal/db/id.go
+++ b/internal/db/id.go
@@ -19,11 +19,11 @@ func NewPublicId(prefix string) (string, error) {
 func newId(prefix string) (string, error) {
 	const op = "db.newId"
 	if prefix == "" {
-		return "", errors.New(errors.InvalidParameter, errors.WithOp(op), errors.WithMsg("missing prefix"))
+		return "", errors.New(errors.InvalidParameter, op, "missing prefix")
 	}
 	publicId, err := base62.Random(10)
 	if err != nil {
-		return "", errors.Wrap(err, errors.WithOp(op), errors.WithMsg("unable to generate id"))
+		return "", errors.Wrap(err, op, errors.WithMsg("unable to generate id"))
 	}
 	return fmt.Sprintf("%s_%s", prefix, publicId), nil
 }
--- a/internal/errors/code.go
+++ b/internal/errors/code.go
@@ -25,7 +25,7 @@ const (
 	InvalidAddress   Code = 101 // InvalidAddress represents an invalid host address for an operation
 	InvalidPublicId  Code = 102
 	InvalidFieldMask Code = 103
-	MissingFieldMask Code = 104
+	EmptyFieldMask   Code = 104
 
 	// DB errors are reserved Codes from 1000-1999
 	CheckConstraint      Code = 1000 // CheckConstraint represents a check constraint error
--- a/internal/errors/code.go
+++ b/internal/errors/code.go
@@ -22,6 +22,8 @@ const (
 
 	// General function errors are reserved Codes 100-999
 	InvalidParameter Code = 100 // InvalidParameter represents and invalid parameter for an operation.
+	InvalidAddress   Code = 101
+	InvalidPublicId  Code = 102
 
 	// DB errors are resevered Codes from 1000-1999
 	CheckConstraint      Code = 1000 // CheckConstraint represents a check constraint error
--- a/internal/errors/code.go
+++ b/internal/errors/code.go
@@ -24,6 +24,8 @@ const (
 	InvalidParameter Code = 100 // InvalidParameter represents an invalid parameter for an operation.
 	InvalidAddress   Code = 101 // InvalidAddress represents an invalid host address for an operation
 	InvalidPublicId  Code = 102
+	InvalidFieldMask Code = 103
+	MissingFieldMask Code = 104
 
 	// DB errors are reserved Codes from 1000-1999
 	CheckConstraint      Code = 1000 // CheckConstraint represents a check constraint error
--- a/internal/errors/code.go
+++ b/internal/errors/code.go
@@ -21,15 +21,15 @@ const (
 	Unknown Code = 0 // Unknown will be equal to a zero value for Codes
 
 	// General function errors are reserved Codes 100-999
-	InvalidParameter Code = 100 // InvalidParameter represents and invalid parameter for an operation.
-	InvalidAddress   Code = 101
+	InvalidParameter Code = 100 // InvalidParameter represents an invalid parameter for an operation.
+	InvalidAddress   Code = 101 // InvalidAddress represents an invalid host address for an operation
 	InvalidPublicId  Code = 102
 
-	// DB errors are resevered Codes from 1000-1999
+	// DB errors are reserved Codes from 1000-1999
 	CheckConstraint      Code = 1000 // CheckConstraint represents a check constraint error
 	NotNull              Code = 1001 // NotNull represents a value must not be null error
 	NotUnique            Code = 1002 // NotUnique represents a value must be unique error
-	NotSpecificIntegrity Code = 1003 // NotSpecificIntegrity represents an integrity error that has no specificy domain error code
+	NotSpecificIntegrity Code = 1003 // NotSpecificIntegrity represents an integrity error that has no specific domain error code
 	MissingTable         Code = 1004 // Missing table represents an undefined table error
 	RecordNotFound       Code = 1100 // RecordNotFound represents that a record/row was not found matching the criteria
 	MultipleRecords      Code = 1101 // MultipleRecords represents that multiple records/rows were found matching the criteria
--- a/internal/errors/code.go
+++ b/internal/errors/code.go
@@ -23,9 +23,9 @@ const (
 	// General function errors are reserved Codes 100-999
 	InvalidParameter Code = 100 // InvalidParameter represents an invalid parameter for an operation.
 	InvalidAddress   Code = 101 // InvalidAddress represents an invalid host address for an operation
-	InvalidPublicId  Code = 102
-	InvalidFieldMask Code = 103
-	EmptyFieldMask   Code = 104
+	InvalidPublicId  Code = 102 // InvalidPublicId represents an invalid public Id for an operation
+	InvalidFieldMask Code = 103 // InvalidFieldMask represents an invalid field mast for an operation
+	EmptyFieldMask   Code = 104 // EmptyFieldMask represents an empty field mask for an operation
 
 	// DB errors are reserved Codes from 1000-1999
 	CheckConstraint      Code = 1000 // CheckConstraint represents a check constraint error
--- a/internal/errors/error.go
+++ b/internal/errors/error.go
@@ -33,11 +33,13 @@ type Err struct {
 }
 
 // E creates a new Err with provided code and supports the options of:
-// WithOp - allows you to specify an optional Op (operation)
-// WithMsg() - allows you to specify an optional error msg, if the default
+//
+// * WithOp() - allows you to specify an optional Op (operation)
+//
+// * WithMsg() - allows you to specify an optional error msg, if the default
 // msg for the error Code is not sufficient.
-// WithWrap() - allows you to specify
-// an error to wrap
+//
+// * WithWrap() - allows you to specify an error to wrap
 func E(c Code, opt ...Option) error {
 	opts := GetOpts(opt...)
 	return &Err{
@@ -50,7 +52,8 @@ func E(c Code, opt ...Option) error {
 
 // New creates a new Err with provided code, op and msg
 // It supports the options of:
-// WithWrap() - allows you to specify an error to wrap
+//
+// * WithWrap() - allows you to specify an error to wrap
 func New(c Code, op Op, msg string, opt ...Option) error {
 	if op != "" {
 		opt = append(opt, WithOp(op))
@@ -65,7 +68,8 @@ func New(c Code, op Op, msg string, opt ...Option) error {
 // Wrap creates a new Err from the provided err and op,
 // preserving the code from the originating error.
 // It supports the options of:
-// WithMsg() - allows you to specify an optional error msg, if the default
+//
+// * WithMsg() - allows you to specify an optional error msg, if the default
 // msg for the error Code is not sufficient.
 func Wrap(e error, op Op, opt ...Option) error {
 	if op != "" {
--- a/internal/errors/error.go
+++ b/internal/errors/error.go
@@ -32,13 +32,13 @@ type Err struct {
 	Wrapped error
 }
 
-// New creates a new Err and supports the options of:
+// E creates a new Err with provided code and supports the options of:
 // WithOp - allows you to specify an optional Op (operation)
 // WithMsg() - allows you to specify an optional error msg, if the default
 // msg for the error Code is not sufficient.
 // WithWrap() - allows you to specify
 // an error to wrap
-func New(c Code, opt ...Option) error {
+func E(c Code, opt ...Option) error {
 	opts := GetOpts(opt...)
 	return &Err{
 		Code:    c,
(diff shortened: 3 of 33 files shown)


louisruch:
@talanknight @jimlambrt @mgaffney PTAL
I added a couple of error codes which give a bit more clear idea of the type of error that occurred let me know thoughts on naming and alternative ideas on this.  
I moved away from using sentials as much as I could, this also means I needed to update a number of tests to do an `error.Match` instead of an `error.Is`
Finally, I made a helper function to strip the line number off the Op in `error_test.go` as I did not see a need to verify the line number in the test as the code is just using stdlib here to get that line number

internal/host/static/repository_host.go
```diff
@@ -23,44 +23,45 @@ import (
 // unique within h.CatalogId.
 func (r *Repository) CreateHost(ctx context.Context, scopeId string, h *Host, opt ...Option) (*Host, error) {
 	if h == nil {
-		return nil, fmt.Errorf("create: static host: %w", errors.ErrInvalidParameter)
+		return nil, errors.New(errors.InvalidParameter, errors.WithMsg("nil host"))
```
louisruch, line 26: So the open question here is do we want to move away at this point from this generic `InvalidParameter` and return a specific `NilHost` error with this PR.  This would be a step to having a clear error code/ID for each possible error returned.

Simialarly below we would then have `NoCatalogId`, `NoScopeId` etc

internal/errors/code.go
```diff
@@ -22,12 +22,15 @@ const (
 
 	// General function errors are reserved Codes 100-999
 	InvalidParameter Code = 100 // InvalidParameter represents and invalid parameter for an operation.
+	GenerateId       Code = 101
```
louisruch, line 25: New codes to avoid a generic `internal` or similar error and to avoid using `unknown`

internal/servers/controller/handlers/errors.go
```diff
@@ -182,7 +183,13 @@ func ErrorHandler(logger hclog.Logger) runtime.ErrorHandlerFunc {
 				errId = "failed_to_generate_error_id"
 			}
 			logger.Error("internal error returned", "error id", errId, "error", inErr)
-			apiErr = getInternalError(errId)
+			var msg string
+			if dErr := errors.Convert(inErr, false); dErr != nil {
+				// This is a domain error, get message to return on API
+				// Do not return the full dErr.Error() as it may contain un-scrubbed info
+				msg = dErr.Msg
```
louisruch, line 190: Note to reviewers: are we ok exposing this message to end-users as it is not necessarily as well-formatted as the other messages we return https://github.com/hashicorp/boundary/blob/main/internal/servers/controller/handlers/errors.go#L21

internal/host/static/repository_host.go
mgaffney: I think returning `InvalidParameter` is fine for all of the checks except for the `ErrInvalidAddress` check.

internal/host/static/repository_host.go
```diff
@@ -23,44 +23,45 @@ import (
...
+		return nil, errors.New(errors.InvalidParameter, errors.WithMsg("public id not empty"))
 	}
 	if scopeId == "" {
-		return nil, fmt.Errorf("create: static host: no scopeId: %w", errors.ErrInvalidParameter)
+		return nil, errors.New(errors.InvalidParameter, errors.WithMsg("no scopeId"))
 	}
 	h.Address = strings.TrimSpace(h.Address)
 	if len(h.Address) < MinHostAddressLength || len(h.Address) > MaxHostAddressLength {
-		return nil, fmt.Errorf("create: static host: bad address: %w", ErrInvalidAddress)
+		return nil, errors.New(errors.InvalidParameter, errors.WithMsg("bad address"))
```
mgaffney, line 42: I think a specific error code/ID should returned for an invalid address since it could be a user input error or a programming error.

internal/host/static/repository_host.go
```diff
@@ -23,44 +23,45 @@ import (
...
 	}
 	h = h.clone()
 
 	opts := getOpts(opt...)
 
 	if opts.withPublicId != "" {
 		if !strings.HasPrefix(opts.withPublicId, HostPrefix+"_") {
-			return nil, fmt.Errorf("create: static host: passed-in public ID %q has wrong prefix, should be %q: %w", opts.withPublicId, HostPrefix, errors.ErrInvalidPublicId)
+			return nil, errors.New(errors.InvalidParameter,
```
mgaffney, line 50: I think `InvalidPublicId` (or something similar) should be a specific error code/ID

internal/servers/controller/handlers/hosts/host_service.go
```diff
@@ -178,8 +178,8 @@ func (s Service) createInRepo(ctx context.Context, scopeId, catalogId string, it
 	}
 	out, err := repo.CreateHost(ctx, scopeId, h)
 	if err != nil {
-		if errors.IsUniqueError(err) || errors.Is(err, errors.ErrNotUnique) {
-			// Push this error through so the error interceptor can interpret it correctly.
+		if dErr := errors.Convert(err, false); dErr != nil {
```
jimlambrt, line 181: Most excellent.

jimlambrt (changes requested):
Great work!  I hope my comments are helpful.

internal/errors/error.go
```diff
@@ -51,33 +64,44 @@ func New(c Code, opt ...Option) error {
 // Convert will convert the error to a Boundary *Err (returning it as an error)
 // and attempt to add a helpful error msg as well. If that's not possible, it
 // will return nil
-func Convert(e error) *Err {
+func Convert(e error, withCallerOp bool) *Err {
```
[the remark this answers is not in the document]
louisruch: Yeah so this has changed a bit as withOp has dropped completely now in favor of the unique errorId, update to this PR will be pushed shortly

internal/errors/code.go
```diff
@@ -22,12 +22,15 @@ const (
...
+	LookupHost       Code = 102
```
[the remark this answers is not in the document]
louisruch: oops thanks!

internal/errors/option.go
```diff
@@ -43,6 +43,6 @@ func WithMsg(msg string) Option {
 // the error.
```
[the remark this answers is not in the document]
louisruch: Dropping this now

internal/host/static/public_ids.go
```diff
@@ -24,7 +25,7 @@ func newHostCatalogId() (string, error) {
 func newHostId() (string, error) {
```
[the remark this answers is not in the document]
louisruch: Yeah have that fixed up locally

louisruch:
@talanknight @jimlambrt @malnick PTAL
This now includes the entire host repo path not just `repository_host.go`
This makes use of the new errorId field
Work to still do (might be in a new PR) is to pass domain errors back from the handler so it is not overwritten with by `fmt.Errorf` at the last step

internal/servers/controller/handlers/errors_test.go
```diff
@@ -97,7 +96,7 @@ func TestApiErrorHandler(t *testing.T) {
 		},
 		{
 			name: "Db invalid public id error",
-			err:  fmt.Errorf("test error: %w", errors.ErrInvalidPublicId),
+			err:  errors.New(errors.MissingPublicId, "testid", errors.WithMsg("test msg")),
```
louisruch, line 99: @mgaffney @talanknight @jimlambrt should this be a 400 or 500 as it is?

internal/servers/controller/handlers/errors_test.go
```diff
@@ -153,7 +152,70 @@ func TestApiErrorHandler(t *testing.T) {
...
-			err:  fmt.Errorf("test error: %w", errors.ErrMultipleRecords),
+			err:  errors.New(errors.MultipleRecords, "testid", errors.WithMsg("test msg")),
+			expected: &pb.Error{
+				Status:  http.StatusInternalServerError,
+				Code:    "Internal",
+				Details: &pb.ErrorDetails{ErrorId: ""},
+			},
+		},
+		{
+			name: "Db invalid address",
```
louisruch, line 163: @mgaffney @talanknight @jimlambrt should this be a 400 or 500 as it is?

louisruch:
I pulled back some of the larger refactor work and just focused on the changes to the domain error and refactored just the `CreateHost` path so we can get consensus.  

Sample before/after error message based on these domain error changes:

Before these changes:
```
error="create: static host: parameter violation: error #100
new host id: parameter violation: error #100
missing prefix: parameter violation: error #100"
```

WIth these changes
```
error="create: static host: new host id: missing prefix: parameter violation: error #100"
```

internal/errors/code.go
```diff
@@ -22,6 +22,8 @@ const (
 
 	// General function errors are reserved Codes 100-999
 	InvalidParameter Code = 100 // InvalidParameter represents and invalid parameter for an operation.
+	InvalidAddress   Code = 101
```
louisruch, line 25: Note to reviewers I add these are non-generic InvalidParameters because I assume we want to return 400 in these two cases, please confirm?

talanknight (commented):
The error message looks a lot cleaner!  I'm excited to see how this ends up.

louisruch:
@jimlambrt @mgaffney after chatting with Mike I made a couple of changes PTAL when you get a chance

internal/db/id.go
```diff
@@ -17,12 +17,13 @@ func NewPublicId(prefix string) (string, error) {
 }
 
 func newId(prefix string) (string, error) {
+	const op = errors.Op("db.newId")
```
jimlambrt, line 20: I think you can just use `const op = "db.newId"`  The `errors.Op` isn't needed and you won't have to cast it when using it later.

internal/host/static/repository_host.go
```diff
@@ -22,45 +22,50 @@ import (
...
+		return nil, errors.New(errors.InvalidAddress, errors.WithOp(op))
 	}
 	h = h.clone()
 
 	opts := getOpts(opt...)
 
 	if opts.withPublicId != "" {
 		if !strings.HasPrefix(opts.withPublicId, HostPrefix+"_") {
-			return nil, fmt.Errorf("create: static host: passed-in public ID %q has wrong prefix, should be %q: %w", opts.withPublicId, HostPrefix, errors.ErrInvalidPublicId)
+			return nil, errors.New(
```
jimlambrt, line 51: Just an observation, that this change may require a change to the caller, if they used `errors.Is(err, errors.ErrInvalidPublicId)` to determine how to handle the error.   If you wrapped the sentinel `errors.ErrInvalidPublicId` then you'd be backward compatible.... if that's important.

internal/servers/controller/handlers/hosts/host_service.go
```diff
@@ -178,11 +177,7 @@ func (s Service) createInRepo(ctx context.Context, scopeId, catalogId string, it
 	}
 	out, err := repo.CreateHost(ctx, scopeId, h)
 	if err != nil {
```
jimlambrt, line 179: any chance we want to test that all the errors raised by the repo are infact domain errors?

jimlambrt (changes requested):
Just one tiny bit about wrapping an error that might need attention.  Overall, this is soooo much better!  ty!

internal/host/static/repository_host.go
louisruch: I could not find anywhere in the code that the sentinels removed in this PR are checked (other than some tests not included in this path), perhaps I missed something.

internal/servers/controller/handlers/hosts/host_service.go
louisruch: I updated the tests in this path to do an `errors.Match`-  I could add an `assertDomainErr` check before?

internal/errors/error.go
```diff
@@ -48,15 +48,63 @@ func New(c Code, opt ...Option) error {
...
+			e.Op = arg
+		case string:
+			e.Msg = arg
+		case *Err: // order is important, this match must before "case error:"
+			c := *arg
+			e.Wrapped = &c
+		case error:
+			e.Wrapped = arg
+		default:
+			// ignore it
```
louisruch, line 82: We should probably log this somewhere, I am not sure how our logging story works any pointers would be great @mgaffney

malnick:
@louisruch Make sure to also include a changelog update with this PR for the specific library we're improving error handling for - thank you!

internal/errors/error.go
```diff
@@ -32,13 +32,13 @@ type Err struct {
...
 
-// New creates a new Err and supports the options of:
+// E creates a new Err with provided code and supports the options of:
 // WithOp - allows you to specify an optional Op (operation)
 // WithMsg() - allows you to specify an optional error msg, if the default
 // msg for the error Code is not sufficient.
 // WithWrap() - allows you to specify
 // an error to wrap
-func New(c Code, opt ...Option) error {
+func E(c Code, opt ...Option) error {
```
jefferai, line 41: I think E shouldn't have a required Code parameter. If you are wanting to wrap an error (but with advanced options) then you may not have a new code to provide. If there isn't another error given to wrap, then this should simply fill in a suitable default.

internal/errors/error.go
```diff
@@ -104,18 +139,25 @@ func (e *Err) Error() string {
 		join(&s, ": ", e.Msg)
 	}
 
-	if info, ok := errorCodeInfo[e.Code]; ok {
+	var skipInfo bool
+	var wrapped *Err
+	if As(e.Wrapped, &wrapped) {
+		// if wrapped error code is the same as this error, don't print redundant info
```
jefferai, line 145: What if it's the same code but different message/kind/op? This makes an assumption that although it's _possible_ to specify the same code when wrapping an error (since we are supporting multiple codes) that only one place in the stack will actually do so with the same error code. I don't think that's a good assumption. If people can specify codes they very well might...

jefferai (commented):
Made a couple comments, I think it's looking pretty good!

internal/errors/error.go
louisruch: Would you imagine there being a new `errors.WithCode` option so the caller can provide a code when calling `E` or just have it be the wrapped error code if provided otherwise `Unknown`?

internal/errors/error.go
louisruch: `Message` and `Op` are always printed if not empty, `Code` was specifically left out, with this check, to avoid the following when the stringer is invoked:

```
"static.CreateHost: invalid parameter: parameter violation: error #100 static.newHostId: invalid parameter: parameter violation: error #100 db.newId: missing prefix: parameter violation: error #100"
```

Instead, we get:

```
static.CreateHost: static.newHostId: db.newId: missing prefix: parameter violation: error #100
```

The `parameter violation: error #100` part, which comes from https://github.com/hashicorp/boundary/blob/main/internal/errors/info.go#L15 is printed each time we recursively print the wrapped error

Kind is a field based on code within the info map, I am not sure how we can have the same code but a different Kind?

internal/errors/error.go
```diff
@@ -32,13 +32,13 @@ type Err struct {
 	Wrapped error
```
jimlambrt, line 32: The formatting of this comment doesn't quite render correctly as a list for the supported options.  Happy to help with suggested fixes, but it's not very readable at the moment when rendered as godoc.

internal/errors/info.go
```diff
@@ -16,6 +16,22 @@ var errorCodeInfo = map[Code]Info{
 		Message: "invalid parameter",
```
jimlambrt, line 16: Can you add a godoc comment for the Info struct above?  ty!

internal/servers/controller/handlers/hosts/host_service.go
jimlambrt: The tests are great, but what will happen if something is missed an a non-domain error is returned from the repo?

jimlambrt (changes requested):
Looks really good!  Ty for your patience with all of us as we aligned on something we wanted to live with for a long time.  Just a few minor comments...

internal/servers/controller/handlers/hosts/host_service.go
louisruch: Sorry this was more my misunderstanding of the request.  Are you suggesting I put a convert before the return, as I had it in a previous version of this refactor, and only return err directly if it’s a domain error and otherwise return as it is today?

internal/errors/is.go
```diff
@@ -89,3 +89,20 @@ func IsMissingTableError(err error) bool {
 	}
 	return false
 }
+
+// IsNotFoundError returns a boolean indicating whether the error is known to
+// report a not found violation.
+func IsNotFoundError(err error) bool {
```
jefferai, line 95: This function illustrates one of my concerns with the multi-Code model. Switching logic that switches on Code (via using functions like this) can have different behavior depending on the ordering in which the codes are tested in the switch. At the same time, this will return if any error in the chain is RecordNotFound, but not necessarily if the originating error is, or even the most recent error. That means behavior of a switch on Code is also dependent on changes to logic that might seem innocuous -- e.g. if one of the wrapped errors decides to add or modify the code it's wrapping, that can also affect the behavior of this function.

louisruch: The current implementation of this `IsX` functions only check if the most recent domain error's code is of a certain type not any in the chain, @mgaffney mentioned in the call yesterday that perhaps we should have the ability to check if any in the chain match which I did give some thought to but have not implemented yet

internal/errors/is.go
```diff
@@ -89,3 +89,20 @@ func IsMissingTableError(err error) bool {
...
+	if err == nil {
+		return false
+	}
+
+	var domainErr *Err
+	if errors.As(err, &domainErr) {
+		if domainErr.Code == RecordNotFound {
+			return true
+		}
+	}
```
mgaffney, line 105: Could this be simplified to:
```go
return errors.Is(err, ErrRecordNotFound)
```

louisruch: No, because not all `RecordNotFound` codes are the sentinel `ErrRecordNotFound`.  As I am doing this refactor I am also removing the dependence on the sentinels, at least I understood that to be one of the objectives...removing them does however cause this, @jimlambrt also previously brought up the benefit of wrapping the sentinel does allow for the `Is` check.  We have this comment in: https://github.com/hashicorp/boundary/blob/main/internal/errors/sentinels.go#L4
```
Creating new Sentinel type errors like these should be deprecated in favor of the new Err type that includes unique Codes and a Matching function.
```
If we are planning to not create new sentinels I see this as also wanting to remove dependence on old sentinels...

talanknight: Keep in mind also that As() will stop at the top level domain error while Is() will check every error down the wrapped list until it finds one that results in Is() being true or it reaches the end.

internal/errors/error.go
jefferai: Yeah basically I think either you'd need WithWrap, or you'd need WithCode. Since we don't want to error from an error function, the code, if not specified, could be a default. But it's not strictly a required parameter given that if you wrap an error with this function it already has a code.

internal/errors/is.go
jefferai: Or we could just not support multiple Codes at this time and only add it in if/when needed. There was a question a few days ago about whether or not we can think of a time when we'd want to supply more than one Code. Did we ever think of any?

louisruch: So the only time we currently do this is when updating/creating a Host and we get a db `NotNull` or `ConstraintViolation` for both these cases the error returned becomes one with a code of `InvalidAddress`
https://github.com/hashicorp/boundary/blob/e8cf4885e5450b74bbcfd13b4a191dd1d40953e5/internal/host/static/repository_host.go#L178

talanknight: We might also do something similar when authenticating (not found error for auth method or login name could be translated into an unauthenticated error potentially)

internal/errors/error.go
louisruch: 👍

internal/errors/is.go
jefferai: In that case don't we have a single Code anyways? This is the point where we're translating the DB errors to API errors. If two DB errors map to the same API Code, it's still a single Code right?

louisruch: Well the top level would have one code, and therefore the `IsX` function, as it is currently written, will only be satisfied if the top level is `X`.  But the wrapped errors could have a different code:
`{"Code": InvalidAddress, "Msg": "Top level message", "Wrapped": {"Code": "NotNull", "Msg": "Db level message"}}`

louisruch: Currently, the only time the wrapped errors are used are in the stringer

louisruch:
Most recent changes can be viewed in these commits: https://github.com/hashicorp/boundary/pull/786/files/3bdc6e958fc745b65d7cd0ead5150dfac0937fd8..13c2f62154d231d8349cf3feff74e4607e601645

louisruch:
@jefferai @jimlambrt @talanknight PTAL I am considering unless someone objects, to merge this PR as is and then do the handler refactor in a quick follow up.  I am trying to avoid having this PR never stop growing.  The current state keeps the existing behavior with the modifications in the error package we have discussed.  For a reference to changes since last reviewed:
https://github.com/hashicorp/boundary/pull/786/files/3bdc6e958fc745b65d7cd0ead5150dfac0937fd8..13c2f62154d231d8349cf3feff74e4607e601645

talanknight (approved):

internal/errors/code.go
```diff
@@ -21,13 +21,17 @@ const (
 	Unknown Code = 0 // Unknown will be equal to a zero value for Codes
 
 	// General function errors are reserved Codes 100-999
-	InvalidParameter Code = 100 // InvalidParameter represents and invalid parameter for an operation.
+	InvalidParameter Code = 100 // InvalidParameter represents an invalid parameter for an operation.
+	InvalidAddress   Code = 101 // InvalidAddress represents an invalid host address for an operation
```
jimlambrt, line 25: Nit: there are several undocumented exported errors here.

jimlambrt (dismissed):
Looks great.  One little nit that you can ignore if you wish.

PR #790, opened 2020-11-17
Refactor domain error to use errorId

louisruch:
### What does this PR do

Refactors boundary domain errors to include the new `errorId` field
Removes withOp option
Adds a new `errors.Wrap` function

### Testing

Updated and added new unit tests

### Follow-up work

Deprecate sentinel errors, with the addition of the `errorId` sentinel errors no longer provide adequate information to be uniquely identified

### DIFF
--- a/internal/errors/sentinels.go
+++ b/internal/errors/sentinels.go
@@ -1,44 +1,42 @@
 package errors
 
-// Errors returned from this package may be tested against these errors
-// with errors.Is. Creating new Sentinel type errors like these should be
-// deprecated in favor of the new Err type that includes unique Codes and a
-// Matching function.
+// Errors returned from this package may be tested against these errors with errors.Is.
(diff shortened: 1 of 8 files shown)


internal/errors/sentinels.go
```diff
@@ -1,44 +1,42 @@
 package errors
 
-// Errors returned from this package may be tested against these errors
-// with errors.Is. Creating new Sentinel type errors like these should be
-// deprecated in favor of the new Err type that includes unique Codes and a
-// Matching function.
+// Errors returned from this package may be tested against these errors with errors.Is.
+// TODO: Deprecate sentinel errors in favor of the new Err type that includes unique ErrorId and Code.
```
louisruch, line 4: Note to reviewers:  I plan to deprecate all of these sentinel errors as I work for the refactor of all errors to be the new domain errors.  For now, I added a unique `errorId` to each to explicitly show the error was created via `sentinels.go`

talanknight (approved):

PR #796, opened 2020-11-19
docs: add workflow for transparent usage of Boundary by SSH client

malnick:


website/data/docs-navigation.js
```diff
@@ -90,6 +90,7 @@ export default [
       'manage-users-groups',
       'manage-roles',
       'manage-sessions',
+			'manage-ssh',
```
jefferai, line 93: I don't like the title "manage ssh" -- it's not managing SSH at all, it's making it easy to do a very specific ssh client workflow.

malnick: Agree. I was trying my hardest to find a way to make it work with the existing `manage-` titles but alas, this one is too different. How about `workflow-ssh`?

jorhett:
That's a horrible amount of setup, and leaves all terminations on nodes in `known_hosts` under the localhost entry FYI

This is considerably cleaner and easier for mere mortals to use ;-) 
```
   ProxyCommand  boundary connect -target-name=%h -exec nc -- {{boundary.ip}} {{boundary.port}}
```

malnick:
Thanks for the feedback @jorhett - the `localhost` example is meant to be an example. I assume most if not all users will want that to be set differently in a real-world scenario.

website/content/docs/common-workflows/workflow-ssh-proxycommand.mdx
```diff
@@ -0,0 +1,111 @@
...
+page_title: SSH ProxyCommand Workflow 
+sidebar_title: SSH ProxyCommand Workflow
+description: How to manage SSH proxy configuration with Boundary
+---
+
+# SSH ProxyCommand Workflow 
+
+The most common pattern for using Boundary to SSH is with the built-in `boundary connect ssh` command. However,
+there are more idomatic approaches that can be employed to make Boundary transparent to users, and at the same
+time simplify common developer and operator workflows. Using `ProxyCommand` to execute a proxy when envoking the
```
jefferai, line 12:
```suggestion
time simplify common developer and operator workflows. Using `ProxyCommand` to execute a proxy when invoking the
```

website/content/docs/common-workflows/workflow-ssh-proxycommand.mdx
```diff
@@ -0,0 +1,111 @@
...
+```bash
+Host localhost
+    SendEnv LANG LC_*
+    PasswordAuthentication yes
+
+    # requires BOUNDARY_CONNECT_AUTHZ_TOKEN to be set from authorize-session command
+    ProxyCommand sh -c "boundary connect -exec nc -- {{boundary.ip}} {{boundary.port}}"
+```
+
+The `ProxyCommand` tells the SSH client to envoke `boundary connect`. We are passing the `-exec nc` flag to
```
jefferai, line 30:
```suggestion
The `ProxyCommand` tells the SSH client to invoke `boundary connect`. We are passing the `-exec nc` flag to 
```

website/content/docs/common-workflows/workflow-ssh-proxycommand.mdx
```diff
@@ -0,0 +1,111 @@
...
+    # requires BOUNDARY_CONNECT_AUTHZ_TOKEN to be set from authorize-session command
+    ProxyCommand sh -c "boundary connect -exec nc -- {{boundary.ip}} {{boundary.port}}"
+```
+
+The `ProxyCommand` tells the SSH client to envoke `boundary connect`. We are passing the `-exec nc` flag to 
+`boundary connect` to wrap [netcat](http://netcat.sourceforge.net/), and then pass the `boundary.ip` and `boundary.port`
+varibles as arguments to `nc`. This allows us to proxy our SSH connection through a local netcat tunnel that's
+managed as a Boundary session.
+
+Now, when you run `ssh <user>@localhost`, SSH will envoke `boundary connect`, and will tunnel the traffic through
```
jefferai, line 35:
```suggestion
Now, when you run `ssh <user>@localhost`, SSH will invoke `boundary connect`, and will tunnel the traffic through 
```

website/content/docs/common-workflows/workflow-ssh-proxycommand.mdx
```diff
@@ -0,0 +1,111 @@
...
+`boundary connect` to wrap [netcat](http://netcat.sourceforge.net/), and then pass the `boundary.ip` and `boundary.port`
+varibles as arguments to `nc`. This allows us to proxy our SSH connection through a local netcat tunnel that's
+managed as a Boundary session.
+
+Now, when you run `ssh <user>@localhost`, SSH will envoke `boundary connect`, and will tunnel the traffic through 
+the local Boundary proxy in a wrapped netcat session.
+
+## Authenticate to Boundary
+
+Before you can envoke `boundary connect` you need to authenticate. In this example, we're using `boundary dev` with the
```
jefferai, line 40:
```suggestion
Before you can invoke `boundary connect` you need to authenticate. In this example, we're using `boundary dev` with the 
```

website/content/docs/common-workflows/workflow-ssh-proxycommand.mdx
```diff
@@ -0,0 +1,111 @@
...
+```
+
+## Authorize the Session
+
+Running Boundary in [dev mode](https://www.boundaryproject.io/docs/getting-started#what-is-dev-mode) creates other 
+default resources such as a target and auth method. The default target (ttcp_1234567890) is already setup with a host
+at `localhost` and a default port of `22`. 
+
+Authorize a session to the default target and store the token in the `BOUNDARY_CONNECT_AUTHZ_TOKEN` environment variable. 
+This will be used by Boundary when it's envoked by SSH:
```
jefferai, line 54:
```suggestion
This will be used by Boundary when it's invoked by SSH:
```

website/content/docs/common-workflows/workflow-ssh-proxycommand.mdx
```diff
@@ -0,0 +1,111 @@
...
+The most common pattern for using Boundary to SSH is with the built-in `boundary connect ssh` command. However,
+there are more idomatic approaches that can be employed to make Boundary transparent to users, and at the same
+time simplify common developer and operator workflows. Using `ProxyCommand` to execute a proxy when invoking the
+SSH client is a common practice. In this workflow, we'll cover configuring your SSH client to execute the 
+`boundary` command, enabling a simplified SSH workflow that leverages Boundary's authenticated proxy for accessing
+targets.
+
+## SSH Configuration
+
+Start by configuring a `Host` entry in `.ssh/ssh_config` for `localhost`:
```
jefferai, line 19: To me this doesn't feel like the right pattern. Rather than have to have people do the manual "authorize the session" stuff below, why not create an SSH entry for each service where you specify the Hostname, and have the target info baked in? E.g.
```
Host boundary-web
    Hostname localhost
    ProxyCommand sh -c "boundary connect -target-id ttcp_XXXXXXXXXX -exec nc -- {{boundary.ip}} {{boundary.port}}"
```

malnick: We can do that to, but you'll still need to authenticate and I didn't initially like the idea of hard coding the target ID. That being said, the target ID rarely if ever changes, so maybe it's not that bad - how about we show both?

jefferai: Another possibility is:

```
Host ttcp_*
  ProxyCommand sh -c "boundary connect -target-id %h -exec nc -- {{boundary.ip}} {{boundary.port}}"
```
This straddles the line between the comment above and the things we discussed -- it takes a host that almost certainly is a boundary target and uses the ProxyCommand without it taking over non-Boundary targets.

jorhett: > Thanks for the feedback @jorhett - the `localhost` example is meant to be an example. I assume most if not all users will want that to be set differently in a real-world scenario.

As I'm sure you well know from the support queries, the vast majority of people don't understand the layers involved. This leads me to defensively try and make all examples self-describing in a way that I can see what they are doing when they give me their example, whereas with everything under `localhost` you're gonna see a lot of things that look the same, and many of your users won't grasp those differences.

...and I also find target-oriented examples tend to be easier for people to grasp than ones that focus on the implementation rather than the goal     (speaking as a tech author)

jorhett: > it takes a host that almost certainly is a boundary target

given underscores aren't valid in a domain name, it's either that or an SSH config entry, sure. I think this is a totally useful and valid config if implemented in a wrapper that looks up the target id. But in testing with users, it would appear that directly using target IDs isn't going to be accepted at all.  And they're competent engineers.

jefferai (approved):

PR #808, opened 2020-12-02
Generate postgres.gen.go using schema version directory structure

talanknight:
Changes the way postgres.gen.go is generated by expecting the directory tree to contain the schema release version and potentially a /dev directory.


internal/db/migrations/genmigrations/generate.go
```diff
@@ -20,35 +21,77 @@ func generate(dialect string) {
...
+	sort.Strings(versions)
 
-	for _, name := range names {
-		if !strings.HasSuffix(name, ".sql") {
-			continue
+	largestVer := 0
+	for _, ver := range versions {
+		verVal, err := strconv.Atoi(ver)
+		if err != nil {
+			if ver != "dev" {
```
jefferai, line 38: Nit, but rather than get an error on `dev` and then check if it's `dev`, I think it's nicer if you just put in a switch statement on ver, with a `dev` case and a `default` case. That way if you hit the default case and get an error from atoi, it's always known to be and treated as an error.

jefferai (approved):
Neato. One minor comment, looks good though!

internal/cmd/commands/database/init.go
```diff
@@ -116,6 +118,12 @@ func (c *InitCommand) Flags() *base.FlagSets {
 
 	f = set.NewFlagSet("Init Options")
 
+	f.BoolVar(&base.BoolVar{
+		Name:   "allow-development-migration",
+		Target: &c.flagAllowDevMigration,
+		Usage:  "If set the init will continue even if the schema includes unsafe database update steps that have not been finalized.",
```
jefferai, line 124: It might be worth adding another sentence here indicating that there is no rollback mechanism, so a backup should be taken independently if needed.

jefferai (dismissed):

internal/cmd/commands/database/init.go
```diff
@@ -176,6 +184,15 @@ func (c *InitCommand) Run(args []string) (retCode int) {
 		}()
 	}
 
+	if migrations.DevMigration != c.flagAllowDevMigration {
+		if migrations.DevMigration {
+			c.UI.Error("This version of the binary has unsafe dev database schema updates.  To proceed anyways please use the 'allow-development-migration' flag.")
```
jefferai:
```suggestion
			c.UI.Error("This version of the binary has dev database schema updates which may not be supported in the next official release. To proceed anyways, please use the 'allow-development-migrations' flag.")
```
Note one of the changes in the text -- `allow-development-migrations` -- I think making it plural is probably correct since we always talk about them in terms of migrations, which covers both the single and plural cases.

PR #815, opened 2020-12-09
ICU-738/Refactor internal/db to domain errors

louisruch:
### What does this PR do

Refactors internal/db package to return domain errors

### Testing

Updated unit tests


internal/errors/info.go
```diff
@@ -56,6 +56,10 @@ var errorCodeInfo = map[Code]Info{
 		Message: "missing table",
 		Kind:    Integrity,
 	},
+	ColumnNotFound: {
+		Message: "column not found",
+		Kind:    Search,
```
louisruch, line 61: I am not sure what Kind `ColumnNotFound` should be @mgaffney @jimlambrt?

internal/errors/info.go
```diff
@@ -64,4 +68,8 @@ var errorCodeInfo = map[Code]Info{
 		Message: "multiple records",
 		Kind:    Search,
 	},
+	MaxRetries: {
+		Message: "too many retries",
+		Kind:    Search,
```
louisruch, line 73: I am not sure what Kind `ColumnNotFound` should be @mgaffney @jimlambrt?

louisruch:
@jimlambrt @mgaffney This is the first run at the db refactor, I did not pull the trigger on moving convert to within DB.  If you think we should do that now let us have a call to discuss some questions I have on how that would look.

internal/errors/error.go
```diff
@@ -117,16 +117,15 @@ func Convert(e error) *Err {
 	if e == nil {
 		return nil
 	}
-	var err *Err
-	if As(e, &err) {
+	if err, ok := e.(*Err); ok {
```
louisruch, line 120: I went back to doing this as a direct cast instead of an `As` because many places in the code do a `fmt.Errorf("context %w`.  The convert function then losses that context when wrapping the error higher up.  Once all code paths have been converted to domain errors this problem will go away

louisruch:
@mgaffney @jimlambrt PTAL
Since this refactor was focused on the db package, the final messages have mixed format and therefore sometimes look like:
`delete session: failed db.LookupById: record not found` instead of `package.func: message: db.LookupById: record not found`

This will eventually become fully migrated to the new format once all packages have been refactored

louisruch:
> 
> 
> @mgaffney @jimlambrt PTAL
> Since this refactor was focused on the db package, the final messages have mixed format which means a higher up package does a `fmt.Errorf("context %w)` of the domain error and therefore sometimes look like:
> `delete session: failed db.LookupById: record not found` instead of `package.func: message: db.LookupById: record not found`
> 
> This will eventually become fully migrated to the new format once all packages have been refactored

The op code being duplicated in a recursive call could be a problem that we should perhaps tackle in the stringer; however, this is not a problem - the handler function being passed to `doTx` also does a `doTx` internally - once all code paths have been migrated to domain errors - this error will have additional context between the duplicate op codes here.

internal/errors/error.go
talanknight: Would you mind adding a comment describing when the TODO can be accomplished or why it isn't done now?  Something like "When all code paths have been converted to domain errors change this to using As instead of casting the error."

internal/errors/code.go
```diff
@@ -27,6 +27,21 @@ const (
 	InvalidFieldMask Code = 103 // InvalidFieldMask represents an invalid field mast for an operation
 	EmptyFieldMask   Code = 104 // EmptyFieldMask represents an empty field mask for an operation
 
+	// PasswordTooShort results from attempting to set a password which is to short.
+	PasswordTooShort Code = 200
+
+	// PasswordUnsupportedConfiguration results from attempting to perform an
```
talanknight, line 33: Now might be a good time to expand on what it means for a password config to be unsupported vs. invalid.

Unsupported means boundary is unable to support the password config type (right now any that is not argon2)
Invalid means any supported configuration that does not have parameters which boundary is able to use.

internal/errors/info.go
talanknight: Column not found indicates the database is not in a state it should be which can only be due to internal failure. 
 Maybe this is one of the rare  instances where Internal makes sense.

talanknight (commented):
This looks great!  It's nice to see how the message usefulness improves (possibly with the exception of the example that prints "db.DoTx: db.DoTx: db.Update: db.lookupAfterWrite: db.LookupById:" but even that is helpful in that it provides a readable trace.)

jimlambrt (changes requested):
Amazing work!  The only change I'd like to see is the MaxRetries Kind.

jimlambrt (approved):
Looks great.  I like the changes to use Match(...) for the implementation of the Is* functions.

PR #816, opened 2020-12-10
Add `kube` connect subcommand

jefferai:


internal/cmd/commands.go
```diff
@@ -257,16 +257,16 @@ func initCommands(ui, serverCmdUi cli.Ui, runOpts *RunOptions) {
...
-		"connect ssh": func() (cli.Command, error) {
+		"connect kube": func() (cli.Command, error) {
 			return &connect.Command{
 				Command: base.NewCommand(ui),
-				Func:    "ssh",
+				Func:    "kube",
 			}, nil
 		},
-		"connect rdp": func() (cli.Command, error) {
+		"connect k8s": func() (cli.Command, error) {
```
malnick, line 266: Wondering if the auto complete will be confusing with both commands that start with `k`? You'll get two possible sub commands, and if you're unfamiliar, may not know they do the same thing. Wondering if you think we should start with one only and grow to two once people start using it?

malnick (commented):
One small comment, otherwise this looks great!

internal/cmd/commands.go
jefferai: Will scope to just `kube` for now. Many people prefer `k8s` but `kube` matches `kubectl` -- it's not `k8sctl`.

PR #831, opened 2020-12-15
Attempt canceling sessions by the client

jefferai:
In some circumstances (like a Ctrl-C), the client knows it's shutting
down and the TOFU token will be lost. In this case it makes sense to
cancel the session rather than have it time out. This introduces a
command enum to the handshake message which can be used for the client
to send commands to the worker -- in this case the only specified
command is to cancel the session. It doesn't occur if the session is
already timed out or if we can't authorize connections as this generally
means the session is already invalid.


internal/servers/worker/session.go
```diff
@@ -152,6 +152,28 @@ func (w *Worker) activateSession(ctx context.Context, sessionId, tofuToken strin
...
+func (w *Worker) cancelSession(ctx context.Context, sessionId string) (pbs.SESSIONSTATUS, error) {
+	rawConn := w.controllerSessionConn.Load()
+	if rawConn == nil {
+		return pbs.SESSIONSTATUS_SESSIONSTATUS_UNSPECIFIED, errors.New("could not get a controller client")
+	}
+	conn, ok := rawConn.(pbs.SessionServiceClient)
+	if !ok {
+		return pbs.SESSIONSTATUS_SESSIONSTATUS_UNSPECIFIED, errors.New("could not cast atomic controller client to the real thing")
+	}
+	if conn == nil {
```
louisruch, line 164: Can this ever be nil, given the above two checks?

internal/cmd/commands/connect/connect.go
```diff
@@ -588,22 +593,38 @@ func (c *Command) Run(args []string) (retCode int) {
...
+			sendSessionCancel = true
 		} else {
 			if c.connectionsLeft.Load() == 0 {
 				termInfo.Reason = "No connections left in session"
 			}
 		}
 	}
 
+	if sendSessionCancel {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
louisruch, line 616: minor nit: could use a `const` to define the timeout

internal/servers/worker/tcp_proxy.go
```diff
@@ -65,13 +65,16 @@ func (w *Worker) handleTcpProxyV1(connCtx context.Context, clientAddr *net.TCPAd
 	go func() {
 		defer connWg.Done()
 		_, err := io.Copy(netConn, tcpRemoteConn)
+		netConn.Close()
```
louisruch, line 68: I am not sure but seems like a potential race between the `Close` calls and the `io.Copy` calls in the two `go func()`.  Would we not want to rather do the two closes after the `congWg.Wait()`, therefore giving both anon funcs time to perform the `io.Copy()` as needed?

internal/servers/worker/tcp_proxy.go
```diff
@@ -65,13 +65,16 @@ func (w *Worker) handleTcpProxyV1(connCtx context.Context, clientAddr *net.TCPAd
...
+		tcpRemoteConn.Close()
 		w.logger.Debug("copy from client to endpoint done", "error", err)
 	}()
 	go func() {
 		defer connWg.Done()
 		_, err := io.Copy(tcpRemoteConn, netConn)
+		tcpRemoteConn.Close()
+		netConn.Close()
 		w.logger.Debug("copy from endpoint to client done", "error", err)
```
louisruch, line 77: nit: It would be easier to scan logs for a clear error message, rather than having to parse them for an occurrence of a non `nil` error logline.
I am aware this is not a change you made in this specific PR so feel free to ignore it for now...

louisruch (approved):
Some questions left inline, one potential bug but feel free to merge as is depending on your response

internal/cmd/commands/connect/connect.go
```diff
@@ -539,12 +539,17 @@ func (c *Command) Run(args []string) (retCode int) {
 			c.connWg.Add(1)
 			go func() {
 				defer listeningConn.Close()
-				if err := c.handleConnection(
-					listeningConn,
+				defer c.connWg.Done()
```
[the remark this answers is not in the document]
jefferai: 👍

internal/cmd/commands/connect/connect.go
jefferai: Good point, updated

internal/servers/worker/session.go
jefferai: Yep! One of those tricksy Go things. The value back from the atomic.Value is an `interface{}` which can be nil, that is, `interface{}(nil)`. But it can also be a non-empty reference to a different interface variable, but where that interface variable itself is `nil`, that is, `interface{}(pbs.SessionServiceClient(nil))`.

internal/servers/worker/tcp_proxy.go
jefferai: Yeah I thought about gating this on `nil`. But from a debug perspective I thought it might be useful to keep as is. We can always change later.

internal/servers/worker/tcp_proxy.go
jefferai: Copy blocks until EOF. So what this is doing is saying, if either end EOFs, ensure that the other end also exits. This is how things work on the client side. I added it here too because I have a sneaking suspicion that this might be the cause of a couple instances we've seen where a client disconnects but the session doesn't end -- the EOF on one side doesn't end up resulting in the other side being closed properly. I have to test to see if it resolves that, but in the meantime it seemed like a good thing to add anyways.

louisruch: Aha I see what you are doing here now, thanks!

louisruch (approved):

jimlambrt (approved):
This looks great and I learned a bit about how the workers and controllers communicate.

internal/cmd/commands/connect/connect.go
```diff
@@ -642,31 +662,53 @@ func (c *Command) handleConnection(
 	if err != nil {
 		switch {
 		case strings.Contains(err.Error(), "tls: internal error"):
```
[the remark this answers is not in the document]
jefferai: Yeah there are open issues on Go where people are begging them to do this for `net` and after years and years they have finally become amenable to it but have not actually implemented this yet.

PR #834, opened 2020-12-17
Move listener address in worker config example

simonfrey:
This fixes: The address wrongly being documented as part of the `worker` block, despite belonging into the listener

### DIFF
--- a/website/content/docs/configuration/worker.mdx
+++ b/website/content/docs/configuration/worker.mdx
@@ -14,7 +14,6 @@ The `worker` stanza configures Boundary worker-specific parameters.
 worker {
   name = "example-worker"
   description = "An example worker"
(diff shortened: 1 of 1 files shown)


website/content/docs/configuration/worker.mdx
```diff
@@ -74,13 +70,13 @@ KMS types, see our [KMS documentation](/docs/configuration/kms).
 listener "tcp" {
 	purpose = "proxy"
 	tls_disable = true
+  address = 127.0.0.1
```
jefferai, line 73:
```suggestion
	address = 127.0.0.1
```

jefferai (approved):

jefferai:
Thanks!

PR #839, opened 2020-12-18
Fix IDs being lowercased in role grants

### LINKED ISSUE #794: Boundary roles add-grant lowercases IDs in the grant statement
**Describe the bug**

`boundary role add-grants` lowercases IDs supplied in the grant text, causing the grant to not match on mixed-case IDs.

**To Reproduce**

1. Run `boundary roles create -name=foo  ...`
2. Run `boundary roles add-grants id=(id#-from-step-1) grant='id=hcst_MixedCase;type=*;actions=*`
3. Run `boundary roles read id=(id#-from-step-1) 
4. See that the grant now shows as shown below, and does not succeed in granting the rights

```
Canonical Grants:
    id=hcst_mixedcase;type=*;actions=*
```

**Expected behavior**

Mixed-case IDs should be maintained as-is in the grant

**Additional context**

Seen using both the boundary server and client v0.1.1 on the following platforms: (the only ones available to me)
* the Ubuntu/Debian package in the Hashicorp apt repo
* HomeBrew

jefferai:
Fixes #794

### DIFF
--- a/internal/servers/controller/handlers/roles/role_service_test.go
+++ b/internal/servers/controller/handlers/roles/role_service_test.go
@@ -1423,7 +1423,7 @@ func checkEqualGrants(t *testing.T, expected []string, got *pb.Role) {
 		require.NoError(err)
 		assert.Equal(expected[i], got.GrantStrings[i])
 		assert.Equal(expected[i], got.Grants[i].GetRaw())
-		assert.Equal(parsed.CanonicalString(), got.Grants[i].GetCanonical())
+		assert.Equal(v, got.Grants[i].GetCanonical())
 		j := got.Grants[i].GetJson()
 		require.NotNil(j)
 		assert.Equal(parsed.Id(), j.GetId())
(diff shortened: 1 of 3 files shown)


internal/servers/controller/handlers/roles/role_service_test.go
```diff
@@ -1423,7 +1423,7 @@ func checkEqualGrants(t *testing.T, expected []string, got *pb.Role) {
 		require.NoError(err)
 		assert.Equal(expected[i], got.GrantStrings[i])
 		assert.Equal(expected[i], got.Grants[i].GetRaw())
-		assert.Equal(parsed.CanonicalString(), got.Grants[i].GetCanonical())
+		assert.Equal(v, got.Grants[i].GetCanonical())
```
louisruch, line 1426: Curious about the need for this change, shouldn't `parsed.CanonicalString()` still equal `got.Grants[i].GetCanonical()` after this change...

louisruch (approved):
One question on a test update but otherwise LGTM

internal/servers/controller/handlers/roles/role_service_test.go
jefferai: Yes, but I did it this way so that I could ensure that it failed _without_ the change -- the test being like this is why this wasn't caught before. This way it will also catch a regression.

jimlambrt (approved):
Looks great!  Ty!

PR #841, opened 2020-12-21
ICU-739/iam domain error refactor

louisruch:
### What does this PR do

Refactors internal/iam package to return domain errors
Fixes wrapped issues in doTx anon functions in host and db packages

### Testing

Updated unit tests

### DIFF
--- a/internal/host/static/repository_host.go
+++ b/internal/host/static/repository_host.go
@@ -133,7 +133,7 @@ func (r *Repository) UpdateHost(ctx context.Context, scopeId string, h *Host, ve
 				return nil, db.NoRowsAffected, errors.New(errors.InvalidAddress, op, "invalid address")
 			}
 		default:
-			return nil, db.NoRowsAffected, errors.New(errors.InvalidFieldMask, op, fmt.Sprintf("field: %s", f))
+			return nil, db.NoRowsAffected, errors.New(errors.InvalidFieldMask, op, fmt.Sprintf("invalid field mask: %s", f))
 		}
 	}
 	var dbMask, nullFields []string
@@ -164,10 +164,13 @@ func (r *Repository) UpdateHost(ctx context.Context, scopeId string, h *Host, ve
 			rowsUpdated, err = w.Update(ctx, returnedHost, dbMask, nullFields,
 				db.WithOplog(oplogWrapper, h.oplog(oplog.OpType_OP_TYPE_UPDATE)),
 				db.WithVersion(&version))
-			if err == nil && rowsUpdated > 1 {
+			if err != nil {
+				return errors.Wrap(err, op)
+			}
+			if rowsUpdated > 1 {
 				return errors.E(errors.WithCode(errors.MultipleRecords))
 			}
-			return err
+			return nil
 		},
 	)
 
@@ -250,10 +253,13 @@ func (r *Repository) DeleteHost(ctx context.Context, scopeId string, publicId st
 		func(_ db.Reader, w db.Writer) (err error) {
 			dh := h.clone()
 			rowsDeleted, err = w.Delete(ctx, dh, db.WithOplog(oplogWrapper, h.oplog(oplog.OpType_OP_TYPE_DELETE)))
-			if err == nil && rowsDeleted > 1 {
+			if err != nil {
+				return errors.Wrap(err, op)
+			}
+			if rowsDeleted > 1 {
 				return errors.E(errors.WithCode(errors.MultipleRecords))
 			}
-			return err
+			return nil
 		},
 	)
 
--- a/internal/iam/repository.go
+++ b/internal/iam/repository.go
@@ -72,7 +72,7 @@ func (r *Repository) create(ctx context.Context, resource Resource, _ ...Option)
 	}
 	resourceCloner, ok := resource.(Cloneable)
 	if !ok {
-		return nil, errors.New(errors.InvalidParameter, op, "resource is not Cloneable for create")
+		return nil, errors.New(errors.InvalidParameter, op, "resource is not Cloneable")
 	}
 	metadata, err := r.stdMetadata(ctx, resource)
 	if err != nil {
@@ -120,7 +120,7 @@ func (r *Repository) update(ctx context.Context, resource Resource, version uint
 	}
 	resourceCloner, ok := resource.(Cloneable)
 	if !ok {
-		return nil, db.NoRowsAffected, errors.New(errors.InvalidParameter, op, "resource is not Cloneable for create")
+		return nil, db.NoRowsAffected, errors.New(errors.InvalidParameter, op, "resource is not Cloneable")
 	}
 	metadata, err := r.stdMetadata(ctx, resource)
 	if err != nil {
@@ -191,7 +191,7 @@ func (r *Repository) delete(ctx context.Context, resource Resource, _ ...Option)
 	}
 	resourceCloner, ok := resource.(Cloneable)
 	if !ok {
-		return db.NoRowsAffected, errors.New(errors.InvalidParameter, op, "resource is not Cloneable for create")
+		return db.NoRowsAffected, errors.New(errors.InvalidParameter, op, "resource is not Cloneable")
 	}
 	metadata, err := r.stdMetadata(ctx, resource)
 	if err != nil {
--- a/internal/iam/repository.go
+++ b/internal/iam/repository.go
@@ -2,9 +2,10 @@ package iam
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"strings"
+
+	"github.com/hashicorp/boundary/internal/errors"
 
 	"github.com/hashicorp/boundary/internal/db"
 	"github.com/hashicorp/boundary/internal/kms"
@@ -13,7 +14,7 @@ import (
 )
 
 var (
-	ErrMetadataScopeNotFound = errors.New("scope not found for metadata")
+	ErrMetadataScopeNotFound = errors.New(errors.RecordNotFound, "iam", "scope not found for metadata")
 )
 
 // Repository is the iam database repository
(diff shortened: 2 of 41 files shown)


internal/iam/resource.go
```diff
@@ -49,28 +48,29 @@ type ResourceWithScope interface {
...
-		return nil, stderrors.New("error reader is nil for LookupScope")
+		return nil, errors.New(errors.InvalidParameter, op, "missing reader")
 	}
 	if resource == nil {
-		return nil, stderrors.New("error resource is nil for LookupScope")
+		return nil, errors.New(errors.InvalidParameter, op, "missing resource")
 	}
 	if resource.GetPublicId() == "" {
-		return nil, fmt.Errorf("LookupScope: scope id is unset %w", errors.ErrInvalidParameter)
+		return nil, errors.New(errors.InvalidParameter, op, "missing public id")
```
louisruch, line 59: The old error message said scope id is unset but this was after doing a `resource.GetPublicId() == ""` so I updated this to `missing public id` to align with other errors messages, let me know if I should keep it as `missing scope id`

louisruch:
@jimlambrt @talanknight @mgaffney PTAL

internal/host/static/repository_host.go
```diff
@@ -133,7 +133,7 @@ func (r *Repository) UpdateHost(ctx context.Context, scopeId string, h *Host, ve
 				return nil, db.NoRowsAffected, errors.New(errors.InvalidAddress, op, "invalid address")
 			}
 		default:
-			return nil, db.NoRowsAffected, errors.New(errors.InvalidFieldMask, op, fmt.Sprintf("field: %s", f))
+			return nil, db.NoRowsAffected, errors.New(errors.InvalidFieldMask, op, fmt.Sprintf("invalid field mask: %s", f))
```
mgaffney, line 136: What will the final error string for this be?

louisruch: So I made this change because the default message gets dropped when a custom message is supplied so this change makes it go from:
```static.UpdateHost: field: Bilbo: parameter violation: error #103`
to
```static.UpdateHost: invalid field mask: Bilbo: parameter violation: error #103```

Comment (louisruch) [reply internal/host/static/repository_host.go:136]: @mgaffney I felt having the text `invalid field mask` in the string rather than needing to rely on the `#103` was useful

**Review by mgaffney — APPROVED** [review]

File: internal/iam/repository.go [hunk 3]
```diff
@@ -101,24 +103,28 @@ func (r *Repository) create(ctx context.Context, resource Resource, opt ...Optio
...
+		return nil, db.NoRowsAffected, errors.New(errors.InvalidParameter, op, "missing version")
 	}
 	if resource == nil {
-		return nil, db.NoRowsAffected, errors.New("error updating resource that is nil")
+		return nil, db.NoRowsAffected, errors.New(errors.InvalidParameter, op, "missing resource")
 	}
 	resourceCloner, ok := resource.(Cloneable)
 	if !ok {
-		return nil, db.NoRowsAffected, errors.New("error resource is not Cloneable for update")
+		return nil, db.NoRowsAffected, errors.New(errors.InvalidParameter, op, "resource is not Cloneable for create")
```
jimlambrt, line 123: create should be update.

internal/iam/repository.go
```diff
@@ -161,38 +167,45 @@ func (r *Repository) update(ctx context.Context, resource Resource, version uint
...
+func (r *Repository) delete(ctx context.Context, resource Resource, _ ...Option) (int, error) {
+	const op = "iam.(Repository).delete"
 	if resource == nil {
-		return db.NoRowsAffected, errors.New("error deleting resource that is nil")
+		return db.NoRowsAffected, errors.New(errors.InvalidParameter, op, "missing resource")
 	}
 	resourceCloner, ok := resource.(Cloneable)
 	if !ok {
-		return db.NoRowsAffected, errors.New("error resource is not Cloneable for delete")
+		return db.NoRowsAffected, errors.New(errors.InvalidParameter, op, "resource is not Cloneable for create")
```
jimlambrt, line 194: create should be "delete"

internal/iam/repository_group.go
```diff
@@ -90,33 +92,34 @@ func (r *Repository) UpdateGroup(ctx context.Context, group *Group, version uint
 			g := group.Clone().(*Group)
 			resource, rowsUpdated, err = r.update(ctx, g, version, dbMask, nullFields)
 			if err != nil {
-				return err
+				return errors.Wrap(err, op)
 			}
 			repo, err := NewRepository(read, w, r.kms)
 			if err != nil {
-				return fmt.Errorf("update group: failed creating inner repo: %w for %s", err, group.PublicId)
+				return errors.Wrap(err, op, errors.WithMsg(fmt.Sprintf("for group %s", group.PublicId)))
```
jimlambrt, line 99: you've lost traceability between this error and the next one.. it might be ambiguous which error happen

internal/iam/repository_group.go
```diff
@@ -90,33 +92,34 @@ func (r *Repository) UpdateGroup(ctx context.Context, group *Group, version uint
...
 			return nil
 		},
 	)
 	if err != nil {
 		if errors.IsUniqueError(err) {
-			return nil, nil, db.NoRowsAffected, fmt.Errorf("update group: group %s already exists in org %s: %w", group.Name, group.ScopeId, errors.ErrNotUnique)
+			return nil, nil, db.NoRowsAffected, errors.New(errors.NotUnique, op, fmt.Sprintf("group %s already exists in scope %s", group.Name, group.ScopeId))
 		}
-		return nil, nil, db.NoRowsAffected, fmt.Errorf("update group: %w for %s", err, group.PublicId)
+		return nil, nil, db.NoRowsAffected, errors.Wrap(err, op, errors.WithMsg(fmt.Sprintf("for group %s", group.PublicId)))
```
jimlambrt, line 112: another error that matches others from a traceability point of view.

internal/iam/repository_group.go
```diff
@@ -127,15 +130,15 @@ func (r *Repository) LookupGroup(ctx context.Context, withPublicId string, opt .
 		db.ExpBackoff{},
 		func(read db.Reader, w db.Writer) error {
 			if err := read.LookupByPublicId(ctx, &g); err != nil {
-				return fmt.Errorf("lookup group: failed %w for %s", err, withPublicId)
+				return errors.Wrap(err, op, errors.WithMsg(fmt.Sprintf("for group %s", withPublicId)))
```
jimlambrt, line 133: I'm not a fan of this pattern.. I think, you're losing traceability.

jimlambrt: it's repeated quite a bit in the PR... you're dropping context from the original errors.

jimlambrt: once we decide what to do (if anything) I'll leave it to you to chase down all the instances of this pattern.

internal/iam/resource.go
```diff
@@ -49,28 +48,29 @@ type ResourceWithScope interface {
...
 	}
 	if resource.GetScopeId() == "" {
 		// try to retrieve it from db with it's scope id
 		if err := reader.LookupById(ctx, resource); err != nil {
-			return nil, fmt.Errorf("unable to get resource by public id: %w", err)
+			return nil, errors.Wrap(err, op)
```
jimlambrt, line 64: hum.. not sure what this error will look like... what's the error string?

internal/iam/repository_user.go
```diff
@@ -125,12 +126,12 @@ func (r *Repository) UpdateUser(ctx context.Context, user *User, version uint32,
...
 				dbOpts...,
 			)
-			if err == nil && rowsUpdated > 1 {
-				// return err, which will result in a rollback of the update
-				return stdErrors.New("error more than 1 resource would have been updated ")
-			}
 			if err != nil {
-				return err
+				return errors.Wrap(err, op)
+			}
```
jimlambrt, line 131: easier to read.  ty!

jimlambrt (changes requested):
Looks great.  Couple questions.  ty!

internal/iam/resource.go
louisruch: This would become something like `iam.LookupScope: db.LookupById: error we got in lookup`, if I left the context in it would be: `iam.LookupScope: unable to get resource by public id: db.LookupById: error we got in lookup`  I figured here the `unable to get resource by public id` was equivalent to db.LookupById`

internal/iam/repository_group.go
louisruch: I agree that you won't be able to do a search for say `lookup group: failed` to find the line of code in repository_group that is failing,  Which might not be ideal, However, I am not sure we really want to keep all the redundant message text in all these errors.  You can still pretty easily find the `op` and then search upwards, also using the wrapped errors op you should be able to find the call that is returning in the error.

You might have noticed in other instances I actually dropped the `for group` completely in the `doTx` and just had it wrapped once after the `doTx` when the error is actually returned.  I made a commit now that brings this inline.

internal/iam/repository_group.go
louisruch: Yeah so the issue here really is around having the redundant op from the wrapped error. as I discuss in a different thread.  So in this example it's between:
```iam.(Repository).UpdateGroup: failed creating inner repo: iam.NewRepository: org error```
and
```iam.(Repository).UpdateGroup: iam.NewRepository: org error```
I felt the iam.NewRepository already clarifies the msg `failed creating inner repo`.  I do however agree this does make a simple search for a specific string not as easy

internal/iam/repository_group.go
louisruch: IMO `errors.New` is really where you want the unique error message to be traceable, as this is the location where you would most likely `"fix"` the error issue.  When wrapping the errors the `op` should be sufficient to trace the path, sometimes you do want additional context in the message but not always...
In this example, we would most likely be addressing any issue within `LookupByPublicId` and at this point within `LookupGroup`

jimlambrt (approved):

PR #842, opened 2020-12-22
Organize DB Schema Migration Code and DB Init Checks

### LINKED ISSUE #805: Migrations fail without PgCrypto and Postgres permissions but Boundary doesn't know
**Describe the bug**
If you are missing `pgcrypto` when you run the initial database migrations and don't have the permissions to install it, Boundary will report an error and fail to properly initialise the database, but will still record that migrations were successful, preventing use of Boundary.

**To Reproduce**
1. Install `postgresql12-server`.
2. Make sure your current user does not exist in Postgres. On CentOS 8, this is the default for the `root` user.
3. Run ` boundary database init -config /etc/boundary.hcl`
4. Observe the error message: 
```sql
Error running database migrations: 1 error occurred:
        * error running migrations: 2 errors occurred:
        * migration failed: permission denied to create extension "pgcrypto" in line 0:
begin;

  create extension if not exists "pgcrypto";

  create domain wh_inet_port as integer
  check(
    value > 0
    and
    value <= 65535
  );
  comment on domain wh_inet_port is
  'An ordinal number between 1 and 65535 representing a network port';

  create domain wh_bytes_transmitted as bigint
  check(
    value >= 0
  );
  comment on domain wh_bytes_transmitted is
  'A non-negative integer representing the number of bytes transmitted';

  -- wh_dim_id generates and returns a random ID which should be considered as
  -- universally unique.
  create or replace function wh_dim_id()
    returns text
  as $$
    select encode(digest(gen_random_bytes(16), 'sha256'), 'base64');
  $$ language sql;

  create domain wh_dim_id as text
  check(
    length(trim(value)) > 0
  );
  comment on domain wh_dim_id is
  'Random ID generated with pgcrypto';

  create domain wh_public_id as text
  check(
    value = 'None'
    or
    length(trim(value)) > 10
  );
  comment on domain wh_public_id is
  'Equivalent to wt_public_id but also allows the value to be ''None''';

  create domain wh_timestamp as timestamp with time zone not null;
  comment on domain wh_timestamp is
  'Timestamp used in warehouse tables';

  create domain wh_dim_text as text not null
  check(
    length(trim(value)) > 0
  );
  comment on domain wh_dim_text is
  'Text fields in dimension tables are always not null and always not empty strings';

  -- wh_date_id returns the wh_date_dimension id for ts.
  create or replace function wh_date_id(ts wh_timestamp)
    returns integer
  as $$
    select to_char(ts, 'YYYYMMDD')::integer;
  $$ language sql;

  -- wh_time_id returns the wh_time_of_day_dimension id for ts.
  create or replace function wh_time_id(ts wh_timestamp)
    returns integer
  as $$
    select to_char(ts, 'SSSS')::integer;
  $$ language sql;

  -- wh_date_id returns the wh_date_dimension id for current_timestamp.
  create or replace function wh_current_date_id()
    returns integer
  as $$
    select wh_date_id(current_timestamp);
  $$ language sql;

  -- wh_time_id returns the wh_time_of_day_dimension id for current_timestamp.
  create or replace function wh_current_time_id()
    returns integer
  as $$
    select wh_time_id(current_timestamp);
  $$ language sql;

commit;

 (details: pq: permission denied to create extension "pgcrypto")
        * pq: current transaction is aborted, commands ignored until end of transaction block in line 0: SELECT pg_advisory_unlock($1)
```
5. Try and run the command again but be told `Database already initialized.`.

Boundary is _technically_ setup, but it isn't functional (the web ui returns a blank page), and it hasn't given the credentials for the user to log in with, rendering it useless.

**Expected behavior**
In my opinion, Boundary should abandon the transaction if `pgcrypto` can't be installed, and record as such, allowing for the user to install the extension and re-run migrations. The docs should probably also mention that manually installing `pgcrypto` may be required.

**Additional context**
Testing was performed on a CentOS 8.2004 system with Postgres 12.5.

talanknight:
Reorganizes the code which applies schema changes in Boundary and sets up the ability (in future PRs) to migrate to newer versions of the database schema when needed.

Added checks when `boundary server -config` starts a controller to ensure that the database was initialized correctly.  Also checks in `boundary database init -config` to ensure the database is in a state that can be initialized.

Fixes #805


jimlambrt:
Just a general question: do you intend to use the new errors package?

internal/db/schema/postgres.go
```diff
@@ -0,0 +1,352 @@
...
+func (p *postgres) run(ctx context.Context, migration io.Reader) error {
+	op := errors.Op("schema.run")
+	migr, err := ioutil.ReadAll(migration)
+	if err != nil {
+		return errors.Wrap(err, op)
+	}
+	// run migration
+	query := string(migr[:])
+	if _, err := p.conn.ExecContext(ctx, query); err != nil {
+		if pgErr, ok := err.(*pq.Error); ok {
```
jimlambrt, line 159: do we want tests for this code branch?

internal/db/schema/postgres.go
```diff
@@ -0,0 +1,352 @@
...
+func (p *postgres) version(ctx context.Context) (version int, dirty bool, err error) {
+	op := errors.Op("schema.version")
+	query := `SELECT version, dirty FROM ` + pq.QuoteIdentifier(defaultMigrationsTable) + ` LIMIT 1`
+	err = p.conn.QueryRowContext(ctx, query).Scan(&version, &dirty)
+	switch {
+	case err == sql.ErrNoRows:
+		return nilVersion, false, nil
+
+	case err != nil:
+		if e, ok := err.(*pq.Error); ok {
```
jimlambrt, line 264: do we want tests for this code branch?  I don't think, we can get here if `ensureVersionTable(...)` is correct.. but if there's ever a bug in that function...

internal/db/schema/state.go
```diff
@@ -0,0 +1,57 @@
+package schema
+
+import (
+	"context"
+)
+
+const nilVersion = -1
+
+type migrationState struct {
```
jimlambrt, line 9: would you mind adding a godoc here about this type?  I know, it's not exported but it would be great.

internal/db/schema/state.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// binarySchemaVersion provides the database schema version supported by
+	// this binary.
+	binarySchemaVersion int
+
+	upMigrations   map[int][]byte
+	downMigrations map[int][]byte
+}
+
+var migrationStates = make(map[string]migrationState)
```
jimlambrt, line 23: a comment here about how this is initialized via the generator might be nice.

internal/db/schema/state.go
```diff
@@ -0,0 +1,57 @@
...
+
+// State contains information regarding the current state of a boundary database's schema.
+type State struct {
+	InitializationStarted bool
```
jimlambrt, line 27: some docs for these exported fields might be nice.

internal/db/schema/state.go
```diff
@@ -0,0 +1,57 @@
...
+	Dirty                 bool
+	CurrentSchemaVersion  int
+	BinarySchemaVersion   int
+}
+
+// State provides the state of the boundary schema contained in the backing database.
+func (b *Manager) State(ctx context.Context) (*State, error) {
```
jimlambrt, line 34: why have this receiver function here and not in manager.go where I'd sort of expect it?

internal/db/schema/statement.go
```diff
@@ -0,0 +1,59 @@
...
+type statementProvider struct {
+	pos      int
+	versions []int
+	up, down map[int][]byte
+}
+
+func newStatementProvider(dialect string, curVer int) (*statementProvider, error) {
+	op := errors.Op("schema.newStatementProvider")
+	qp := statementProvider{pos: -1}
+	qp.up, qp.down = migrationStates[dialect].upMigrations, migrationStates[dialect].downMigrations
```
jimlambrt, line 22: should there be a function to get these from the package var?  Sort of like DevMigration(...) or BinarySchemaVersion(...)

internal/db/schema/state.go
```diff
@@ -0,0 +1,57 @@
...
+
+// State contains information regarding the current state of a boundary database's schema.
+type State struct {
+	InitializationStarted bool
+	Dirty                 bool
+	CurrentSchemaVersion  int
+	BinarySchemaVersion   int
+}
+
+// State provides the state of the boundary schema contained in the backing database.
```
jimlambrt, line 33: Perhaps a better name might be: CurrentState(...)???

internal/db/schema/state.go
```diff
@@ -0,0 +1,57 @@
...
+		BinarySchemaVersion: BinarySchemaVersion(b.dialect),
+	}
+	v, dirty, err := b.driver.version(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if v == nilVersion {
+		return &dbS, nil
+	}
+	dbS.InitializationStarted = true
```
jimlambrt, line 45: why not make this side affect more explicit? Perhaps an option?

internal/db/schema/statement.go
```diff
@@ -0,0 +1,59 @@
+package schema
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/hashicorp/boundary/internal/errors"
+)
+
+// statementProvider provides the migration statements in order.
```
jimlambrt, line 10: ty for the docs.

internal/db/schema/state.go
```diff
@@ -0,0 +1,57 @@
...
+	dbS.CurrentSchemaVersion = v
+	dbS.Dirty = dirty
+	return &dbS, nil
+}
+
+func DevMigration(dialect string) bool {
+	return migrationStates[dialect].devMigration
```
jimlambrt, line 52: do we want a test for this function?

jimlambrt (changes requested):
Great work!  I did have some questions/suggestions and I'm happy to discuss them in real-time.

Also, I did not review the tests.  I'm happy to do so, if you'd like me to.

internal/db/schema/state.go
talanknight: What side affect are you referring to?

jimlambrt: it appears to start the initialization.  perhaps I misunderstood.

talanknight: No, it records that the initialization has started if there is some version specified in the DB, otherwise it assumes the db initialization has never been run.

internal/db/schema/postgres.go
```diff
@@ -0,0 +1,352 @@
...
+
+	query := `CREATE TABLE IF NOT EXISTS ` + pq.QuoteIdentifier(defaultMigrationsTable) + ` (version bigint not null primary key, dirty boolean not null)`
+	if _, err = p.conn.ExecContext(ctx, query); err != nil {
+		return errors.Wrap(err, op)
+	}
+
+	return nil
+}
+
+const advisoryLockIDSalt uint = 1486364155
```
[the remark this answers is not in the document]
talanknight: Removed all this other lock id stuff.

internal/db/schema/postgres.go
```diff
@@ -0,0 +1,352 @@
...
+
+// GenerateAdvisoryLockId inspired by rails migrations, see https://goo.gl/8o9bCT
+func generateAdvisoryLockId(databaseName string, additionalNames ...string) string {
```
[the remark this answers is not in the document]
talanknight: Removed this existing lock id stuff.

internal/db/schema/statement.go
talanknight: Added functions for both up and down.

internal/db/schema/postgres.go
```diff
@@ -0,0 +1,352 @@
...
+
+	px, err := newPostgres(ctx, db)
+	if err != nil {
+		return nil, errors.Wrap(err, op)
+	}
+
+	return px, nil
+}
+
+func (p *postgres) Close() error {
```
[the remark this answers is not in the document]
talanknight: Moved to the testing file.

internal/db/schema/postgres.go
```diff
@@ -0,0 +1,352 @@
...
+	connErr := p.conn.Close()
+	dbErr := p.db.Close()
+	if connErr != nil || dbErr != nil {
+		return fmt.Errorf("conn: %v, db: %v", connErr, dbErr)
```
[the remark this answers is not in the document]
talanknight: Removed error handling since this was moved too testing and we can now assert error behavior.

internal/db/schema/postgres.go
```diff
@@ -0,0 +1,352 @@
...
+	}
+
+	if err := px.ensureVersionTable(ctx); err != nil {
+		return nil, errors.Wrap(err, op)
+	}
+
+	return px, nil
+}
+
+func (p *postgres) open(ctx context.Context, u string) (*postgres, error) {
```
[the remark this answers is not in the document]
talanknight: Yes.  Moved.

internal/db/schema/manager.go
```diff
@@ -0,0 +1,130 @@
...
+	if err := r.Scan(&gotLock); err != nil || !gotLock {
+		return lockErr
+	}
+
+	return nil
+}
+
+// ExclusiveLock attempts to obtain an exclusive lock on the database.  If the
+// lock can be obtained an error is returned.
+func (b *Manager) ExclusiveLock(ctx context.Context, k SchemaLockKey) error {
```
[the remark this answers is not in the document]
talanknight: Moved to the driver.

internal/db/schema/manager.go
```diff
@@ -0,0 +1,130 @@
...
+type SchemaLockKey uint
+
+const SchemaAccessLockId SchemaLockKey = 3865661975
+
+// Manager provides a way to run operations and retrieve information regarding
+// the underlying boundary database schema.
+// Manager is not thread safe.
+type Manager struct {
+	db      *sql.DB
+	driver  *postgres
```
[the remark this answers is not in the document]
talanknight: Yeah, I wanted to limit the amount that we exported but... I've moved this to its own package and created a driver interface.

talanknight (commented):
Thank you for the review!

internal/db/schema/state.go
jimlambrt: the doc made me chuckle... "an interface that..."  Don't leave me hanging like that :)

internal/db/schema/state.go
jimlambrt: yeah or neah on the comment?

internal/db/schema/state.go
jimlambrt: I think, you missed my comment about comments/docs.

jimlambrt (dismissed):
Ty for all the updates.  Just a few minor nits that are yours to act upon or ignore.  Great work!

internal/db/schema/state.go
```diff
@@ -0,0 +1,68 @@
+package schema
+
+import (
+	"context"
+)
+
+const nilVersion = -1
+
+// migrationState is an interface that
```
jefferai, line 9: an interface that...?

internal/cmd/commands/server/server.go
```diff
@@ -341,6 +343,42 @@ func (c *Command) Run(args []string) int {
 			c.UI.Error(fmt.Errorf("Error connecting to database: %w", err).Error())
 			return 1
 		}
+
+		ctx := context.TODO()
```
jefferai, line 347: Use the command's context via `c.Context`

internal/cmd/commands/server/server.go
```diff
@@ -341,6 +343,42 @@ func (c *Command) Run(args []string) int {
...
+		ckState, err := sMan.CurrentState(ctx)
+		if err != nil {
+			c.UI.Error(fmt.Errorf("Error checking schema state: %w", err).Error())
+			return 1
+		}
+		if !ckState.InitializationStarted {
+			c.UI.Error("Database has not been initialized. Please run `boundary database init`.")
+			return 1
+		}
+		if ckState.Dirty {
```
jefferai, line 367: Just curious, when would this happen, vs. the transaction containing the migration rolling back?

internal/cmd/commands/database/init.go
```diff
@@ -263,34 +263,58 @@ func (c *InitCommand) Run(args []string) (retCode int) {
...
-		ran, err := db.InitStore("postgres", nil, c.srv.DatabaseUrl)
+		if st.InitializationStarted {
+			// TODO: Separate from the "dirty" bit maintained by the schema
+			//  manager maintain a bit which indicates that this full command
+			//  was completed successfully (with all default resources being created).
+			//  Use that bit to determine if a previous init was completed
+			//  successfully or not.
+			c.UI.Error(base.WrapAtLength("Database was already been " +
+				"initialized. If the initialization did not complete successfully " +
+				"please revert the database to it's fresh state."))
```
jefferai, line 297:
```suggestion
				"please revert the database to its fresh state."))
```

internal/cmd/commands/database/init.go
```diff
@@ -263,34 +263,58 @@ func (c *InitCommand) Run(args []string) (retCode int) {
...
 			return 1
 		}
-		ran, err := db.InitStore("postgres", nil, c.srv.DatabaseUrl)
+		if st.InitializationStarted {
+			// TODO: Separate from the "dirty" bit maintained by the schema
+			//  manager maintain a bit which indicates that this full command
+			//  was completed successfully (with all default resources being created).
+			//  Use that bit to determine if a previous init was completed
+			//  successfully or not.
+			c.UI.Error(base.WrapAtLength("Database was already been " +
```
jefferai, line 295:
```suggestion
			c.UI.Error(base.WrapAtLength("Database has already been " +
```

internal/db/schema/state.go
talanknight: Moved CurrentState to manager.go along with the test.

internal/db/schema/state.go
talanknight: I did.  Sorry.

internal/db/schema/state.go
talanknight: Added comments. PTAL.

internal/db/schema/postgres.go
talanknight: Added a test that exercises that code branch.

internal/db/schema/postgres.go
talanknight: Added a test which does.

internal/cmd/commands/server/server.go
talanknight: We currently never rollback.  If there is an error we stop where it failed and leave the bit dirty.

internal/cmd/commands/server/server.go
talanknight: Woops, thanks for finding my TODO.  c.Context is never set with the way we initialize the server.Command.  

I've changed that now.  PTAL and let me know if instead of setting our own context the server and dev commands should use base.NewCommand like the others.  I am assuming there was a reason why we didn't want to cancel all context when the Shutdown channel is called (like is set up when NewCommand() is called, but I couldn't figure out why that would be the case.

jefferai: Ah, yeah, you're right. We haven't needed it before (there's only one other place that uses it and that's for something we don't have yet). Ctrl-C is tied to the general shutdown channel -- generally speaking this is because we want to shut down in an orderly fashion instead of having everything shut down/cancel at once. But in this case I think it's a good reason to do more or less the same thing as with NewCommand and populate c.Context, tied to the shutdown channel.

jefferai: Once you do that you can pass that into CreateDevDatabase too.

internal/cmd/commands/server/server.go
jefferai: Wouldn't the transaction failing cause the changes from that migration to roll back? Isn't each migration its own transaction?

talanknight: That's correct, but that is only because the content of the migration issued the rollback and not because the migration library itself enforces it.   This does get close to a topic I plan on trying to fix soon.  When we fail after the the schema update has completed successfully but before all the default resources are created.  IMO we should still mark the migration as being in a dirty state.  Maybe the solution is pulling the dirty bit out to be managed through an interface in the schema manager so the caller can decide if the migration is `dirty` or not.  I'm happy to talk more about this face to face or on slack.

internal/cmd/commands/server/server.go
talanknight: woops.  I had missed a whole handful of context.TODOs.  Thanks again.  I've made the changes and now the `server` and `dev` commands use the base.NewCommand call.

jimlambrt (approved):

jimlambrt (dismissed):
looks great.

internal/docker/supported.go
```diff
@@ -30,7 +30,7 @@ func startDbInDockerSupported(dialect string) (cleanup func() error, retURL, con
 		resource, err = pool.Run("postgres", "12", []string{"POSTGRES_PASSWORD=password", "POSTGRES_DB=boundary"})
 		url = "postgres://postgres:password@localhost:%s?sslmode=disable"
 		if err == nil {
-			url = fmt.Sprintf("postgres://postgres:password@%s?sslmode=disable", resource.GetHostPort("5432/tcp"))
+			url = fmt.Sprintf("postgres://postgres:password@%s/boundary?sslmode=disable", resource.GetHostPort("5432/tcp"))
```
mgaffney, line 33: Why specify a `boundary` database for the connection?

mgaffney (changes requested):
Overall it looks great! Just a few questions and suggestions.

internal/db/schema/manager.go
```diff
@@ -0,0 +1,164 @@
...
+// Manager is not thread safe.
+type Manager struct {
+	db      *sql.DB
+	driver  driver
+	dialect string
+}
+
+// NewManager creates a new schema manager. An error is returned
+// if the provided dialect is unrecognized or if the passed in db is unreachable.
+func NewManager(ctx context.Context, dialect string, db *sql.DB) (*Manager, error) {
```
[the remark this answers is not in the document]
talanknight: It would actually be schema.NewManager which doesn't stutter.  schema.New would be confusing since we aren't returning a new schema.

internal/db/schema/manager.go
```diff
@@ -0,0 +1,164 @@
...
+	// Dirty is set to true if the database failed in a previous migration/initialization.
+	Dirty bool
+	// CurrentSchemaVersion is the schema version that is currently running in the database.
+	CurrentSchemaVersion int
+	// BinarySchemaVersion is the schema version which this boundary binary supports.
+	BinarySchemaVersion int
+}
+
+// CurrentState provides the state of the boundary schema contained in the backing database.
+func (b *Manager) CurrentState(ctx context.Context) (*State, error) {
```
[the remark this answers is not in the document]
talanknight: It was at one point but I changed it to CurrentState from a recommendation from @jimlambrt .  I think CurrentState differentiates it from the migrationState struct and map that are in state.go.

internal/db/schema/postgres/postgres.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	return px, nil
+}
+
+// TrySharedLock attempts to capture a shared lock. If it is not successful it returns an error.
+// https://www.postgresql.org/docs/9.6/static/explicit-locking.html#ADVISORY-LOCKS
+func (p *Postgres) TrySharedLock(ctx context.Context) error {
+	const op = "postgres.(Postgres).TrySharedLock"
+	r := p.db.QueryRowContext(ctx, "SELECT pg_try_advisory_lock_shared($1)", schemaAccessLockId)
```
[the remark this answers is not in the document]
talanknight: Just because we do not have a use for it yet.

internal/db/schema/postgres/postgres.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	return nil
+}
+
+// Unlock calls pg_advisory_unlock and returns an error if one is returned.
+func (p *Postgres) Unlock(ctx context.Context) error {
+	const op = "postgres.(Postgres).Unlock"
+
+	query := `SELECT pg_advisory_unlock($1)`
```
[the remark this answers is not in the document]
talanknight: This was a copy over from the golang-migrate code.  However, from my understanding the boolean only indicates if the lock was held previously or not.  I don't think that information is important in this code.

internal/db/schema/postgres/postgres.go
```diff
@@ -0,0 +1,341 @@
...
+		if pgErr, ok := err.(*pq.Error); ok {
+			var line uint
+			var col uint
+			var lineColOK bool
+			if pgErr.Position != "" {
+				if pos, err := strconv.ParseUint(pgErr.Position, 10, 64); err == nil {
+					line, col, lineColOK = computeLineFromPos(query, int(pos))
+				}
+			}
+			message := fmt.Sprintf("migration failed")
```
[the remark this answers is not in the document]
talanknight: Yes, but why?  Keep in mind this code will only ever be executed out of band from normal serving traffic and an error here would probably cause the whole binary to exit anyways.

internal/docker/supported.go
talanknight: Because the existing url points to the default DB on the server which results in the db named `postgres` having all the values initialized on it.

internal/db/schema/manager.go
```diff
@@ -0,0 +1,164 @@
...
+
+// State contains information regarding the current state of a boundary database's schema.
+type State struct {
+	// InitializationStarted indicates if the current database has already been initialized
+	// (successfully or not) at least once.
+	InitializationStarted bool
+	// Dirty is set to true if the database failed in a previous migration/initialization.
+	Dirty bool
+	// CurrentSchemaVersion is the schema version that is currently running in the database.
+	CurrentSchemaVersion int
```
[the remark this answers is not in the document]
talanknight: That is much better. Thank you.

internal/db/schema/manager.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// NewManager creates a new schema manager. An error is returned
+// if the provided dialect is unrecognized or if the passed in db is unreachable.
+func NewManager(ctx context.Context, dialect string, db *sql.DB) (*Manager, error) {
+	const op = "schema.NewManager"
+	dbM := Manager{db: db, dialect: dialect}
+	var err error
+	switch dialect {
+	case "postgres", "postgresql":
```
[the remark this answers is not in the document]
talanknight: Yes. Thank you for catching that.

internal/db/schema/postgres/testing.go
```diff
@@ -0,0 +1,173 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Package testing has the database tests.
+// All database drivers must pass the Test function.
+// This lives in it's own package so it stays a test dependency.
```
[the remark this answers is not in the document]
talanknight: Removed. Thanks.

internal/db/schema/postgres/postgres.go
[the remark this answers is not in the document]
talanknight: Yes. Thank you for catching that!

internal/db/schema/manager.go
```diff
@@ -0,0 +1,164 @@
...
+func (b *Manager) SharedLock(ctx context.Context) error {
+	const op = "schema.(Manager).SharedLock"
+	if err := b.driver.TrySharedLock(ctx); err != nil {
+		return errors.Wrap(err, op)
+	}
+	return nil
+}
+
+// ExclusiveLock attempts to obtain an exclusive lock on the database.  If the
+// lock can be obtained an error is returned.
```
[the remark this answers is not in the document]
talanknight: That was a typo.  This should be fixed now.

internal/db/schema/postgres/postgres.go
```diff
@@ -0,0 +1,341 @@
...
+	if err := r.Scan(&gotLock); err != nil {
+		return errors.Wrap(err, op)
+	}
+	if !gotLock {
+		return errors.New(errors.MigrationLock, op, "Lock failed")
+	}
+	return nil
+}
+
+// Lock calls pg_advisory_lock with the provided context and returns an error if one is returned.
```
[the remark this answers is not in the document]
talanknight: I rewrote it a little.  Thanks

internal/db/schema/postgres_migration.gen.go
```diff
@@ -1,39 +1,12 @@
 // Code generated by "make migrations"; DO NOT EDIT.
-package migrations
```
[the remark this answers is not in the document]
talanknight: Moved the "DO NOT EDIT" warning under the package declaration.

internal/db/schema/postgres/testing.go
```diff
@@ -0,0 +1,173 @@
...
+
+	TestNilVersion(t, d) // test first
+	TestLockAndUnlock(t, d)
+	TestRun(t, d, bytes.NewReader(migration))
+	TestSetVersion(t, d) // also tests Version()
+	// drop breaks the driver, so test it last.
+	TestDrop(t, d)
+}
+
+func TestNilVersion(t *testing.T, d *Postgres) {
```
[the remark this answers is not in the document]
talanknight: stopped exporting this and the other test helpers.

internal/db/schema/manager.go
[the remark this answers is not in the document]
mgaffney: Oops! You are right. Sorry about that.

internal/db/schema/postgres/postgres.go
```diff
@@ -0,0 +1,345 @@
...
+	}
+
+	return px, nil
+}
+
+// TrySharedLock attempts to capture a shared lock. If it is not successful it returns an error.
+// https://www.postgresql.org/docs/9.6/static/explicit-locking.html#ADVISORY-LOCKS
+func (p *Postgres) TrySharedLock(ctx context.Context) error {
+	const op = "postgres.(Postgres).TrySharedLock"
+	r := p.conn.QueryRowContext(ctx, "SELECT pg_try_advisory_lock_shared($1)", schemaAccessLockId)
```
mgaffney, line 89: Previous comments are not showing up for some reason. 

[Me](https://github.com/hashicorp/boundary/pull/842#discussion_r558595089):
> Why is there no equivalent unlock method for this (`select pg_advisory_unlock_shared($1)`)?

[Todd](https://github.com/hashicorp/boundary/pull/842#discussion_r558631304):
> Just because we do not have a use for it yet.

Shouldn't we be releasing the lock by calling `select pg_advisory_unlock_shared($1)`? Or are we just letting Postgres do the cleanup when the connection closes? If it is the latter, are there any dangers on relying on that?

mgaffney (approved):
Great job!

jimlambrt (approved):
Looks great.

mgaffney (approved):

PR #843, opened 2020-12-23
Reverse server's incorrect test for globalUnicast IPs

jorhett:
Workers will error out on valid IPs shown here: https://www.boundaryproject.io/docs/configuration/worker#complete-configuration-example

```
Dec 23 21:13:25 boundary-worker-1 boundary[377038]: Controller address "10.10.10.1" is invalid: cannot be a global unicast address
Dec 23 21:13:25 boundary-worker-1 systemd[1]: boundary-worker.service: Main process exited, code=exited, status=1/FAILURE
```

The test implemented in e6c98f1454e64b29e5744efa0e4f4090be0d6257 is simply wrong: a global unicast address is required, not a failure. Any valid IP address it can reach must be a global unicast address

### DIFF
--- a/internal/cmd/commands/server/server.go
+++ b/internal/cmd/commands/server/server.go
@@ -264,8 +264,8 @@ func (c *Command) Run(args []string) int {
 					errMsg = "an unspecified"
 				case ip.IsMulticast():
 					errMsg = "a multicast"
-				case ip.IsGlobalUnicast():
-					errMsg = "a global unicast"
+				case !ip.IsGlobalUnicast():
+					errMsg = "is not a global unicast"
 				}
 				if errMsg != "" {
 					c.UI.Error(fmt.Sprintf("Controller address %q is invalid: cannot be %s address", controller, errMsg))


jefferai:
Closing as this has the wrong effect -- it will go from disallowing a global unicast address to only allowing a global unicast address, leaving private addresses in the dust.

malnick:
Superseded by https://github.com/hashicorp/boundary/pull/845

jorhett:
> Closing as this has the wrong effect -- it will go from disallowing a global unicast address to only allowing a global unicast address, leaving private addresses in the dust.

No problem with the other fix, but private addresses are valid global unicast address FWIW

jefferai:
Ah, interesting. I didn't notice that caveat in the docs, as the function naming would seem to suggest that it must be an actual globally routed address. The fact that they have "It returns true even if ip is in IPv4 private address space or local IPv6 unicast address space." as part of the docs indicates that the Go authors are aware of this nomenclature issue. :-)

jefferai:
However -- dug into the source code and while IsGlobalUnicast is allowed for private addresses, it [fails for loopback](https://golang.org/src/net/ip.go?s=4811:4846#L172) so for now I'm going to keep it as-is.

PR #849, opened 2021-01-05
Correct assert.Equal actual/expected

louisruch:
### What does this PR do

Updates a number of usages of `assert.Equal` where error checks incorrectly had `actual` as the first paramter and `expected` as the second parameter.  This avoids confusing error messages in failed tests such as:
```
Not equal: 
     expected: "actual error string"
     actual  : "expected error string"
```


louisruch:
@jimlambrt @mgaffney @jefferai @talanknight this was a slight annoyance I have come accross during the error refactors, I was fixing them inline while updating tests, but since I have already missed some I decided to make a quick PR based on some loose greping

talanknight (approved):

PR #851, opened 2021-01-06
ICU-934/Verify controller config block

### LINKED ISSUE #819: If not all options are set in the configuration file, then we will get the following error like panic
If not all options are set in the configuration file, then we will get the following error like panic: runtime error: invalid memory address or nil pointer dereference
If we set not all parameters in the boundary configuration file, then we will get the following error:
```
sudo boundary database init -config /etc/boundary/boundary.hcl
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x126af3f]
goroutine 1 [running]:
github.com/hashicorp/boundary/internal/cmd/commands/database.(*InitCommand).Run(0xc0005401e0, 0xc00010c030, 0x2, 0x2, 0x0)
	/go/internal/cmd/commands/database/init.go:208 +0x2ff
github.com/mitchellh/cli.(*CLI).Run(0xc0003683c0, 0xc0003683c0, 0xc00052aa00, 0xc000074260)
	/root/go/pkg/mod/github.com/mitchellh/cli@v1.1.2/cli.go:262 +0x1cf
github.com/hashicorp/boundary/internal/cmd.RunCustom(0xc00010c010, 0x4, 0x4, 0xc000841e60, 0xc000080058)
	/go/internal/cmd/main.go:186 +0x846
github.com/hashicorp/boundary/internal/cmd.Run(...)
	/go/internal/cmd/main.go:92
main.main()
	/go/cmd/boundary/main.go:13 +0xda
```

```
boundary version
Version information:
  Git Revision:        d8020842ae8b6c742b94538baada313d7eb52809
  Version Number:      0.1.2
```

Expected result: error message with a small clarification
Current result: go traceback

Could you please add in the command boundary “check config” option? Also, it would be better to add a sample boundary config file in documentation.

louisruch:
### What does this PR do

Verifies the controller config block has been defined
Fixes #819 

### Testing

Repro'ed issue on main:
```
boundary database init --config test.hcl
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x1e42b4e]

goroutine 1 [running]:
github.com/hashicorp/boundary/internal/cmd/commands/database.(*InitCommand).Run(0xc000306280, 0xc00003a0d0, 0x2, 0x2, 0x0)
	/Users/louisruch/boundary/internal/cmd/commands/database/init.go:230 +0x3ee
github.com/mitchellh/cli.(*CLI).Run(0xc000438dc0, 0xc000438dc0, 0xc000332e40, 0xc000505ee0)
	/Users/louisruch/go/pkg/mod/github.com/mitchellh/cli@v1.1.2/cli.go:262 +0x1cf
github.com/hashicorp/boundary/internal/cmd.RunCustom(0xc00003a0b0, 0x4, 0x4, 0xc000427e60, 0xc0000480b8)
	/Users/louisruch/boundary/internal/cmd/main.go:186 +0x846
github.com/hashicorp/boundary/internal/cmd.Run(...)
	/Users/louisruch/boundary/internal/cmd/main.go:92
main.main()
	/Users/louisruch/boundary/cmd/boundary/main.go:13 +0xda
```

Same config with fix implemented:
```
boundary database init --config test.hcl
"controller" config block not found
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 Canonical reference for changes, improvements, and bugfixes for Boundary.
 
-## vNext
+## Next
 
 ### Bug Fixes
 
@@ -10,7 +10,7 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
   ([Issue](https://github.com/hashicorp/boundary/issues/819)) 
   ([PR](https://github.com/hashicorp/boundary/pull/851))
 
-## v0.1.4 (2021/01/05)
+## 0.1.4 (2021/01/05)
 
 ### New and Improved
 
@@ -28,7 +28,7 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
   ([Issue](https://github.com/hashicorp/boundary/issues/830))
   ([PR](https://github.com/hashicorp/boundary/pull/846))
   
-## v0.1.3 (2020/12/18)
+## 0.1.3 (2020/12/18)
 
 ### Changes/Deprecations
 
@@ -70,7 +70,7 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
   ([Issue](https://github.com/hashicorp/boundary/issues/794))
   ([PR](https://github.com/hashicorp/boundary/pull/839))
 
-## v0.1.2 (2020/11/17)
+## 0.1.2 (2020/11/17)
 
 ### New and Improved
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -20,7 +28,7 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
   ([Issue](https://github.com/hashicorp/boundary/issues/830))
   ([PR](https://github.com/hashicorp/boundary/pull/846))
   
-## 0.1.3 (2020/12/18)
+## v0.1.3 (2020/12/18)
```
jefferai, line 31: Good catch on this not being consistent, but we should actually do the inverse: remove the `v` prefix on the earlier entries. All of the other HC tools do not use a `v` prefix in the Changelog. (They also use more verbose dates but I felt like this was a nicer paradigm.)

jefferai (dismissed):
Approved pending changelog update

PR #855, opened 2021-01-08
Fix worker example config

louisruch:
### What does this PR do

Adds quotes around the address value in the worker listener stanze example config


louisruch:
@malnick @jefferai found this error while setting up my environment to test the config controller nil panic the other day

jefferai (approved):

PR #858, opened 2021-01-08
Skip curl output when using automatic versioning and reading

### LINKED ISSUE #856: Outputting a curl string returns the wrong call on PATCH if version is not supplied
If version isn't supplied on a CLI call, the `-output-curl-string` function shows the GET call to get the version instead of the final call. Generally speaking the final call is the one we want.

jefferai:
This makes it so that the actual output is the update/delete.

Fixes #856

### DIFF
--- a/api/accounts/account.gen.go
+++ b/api/accounts/account.gen.go
@@ -57,10 +57,8 @@ func (n AccountReadResult) GetResponseMap() map[string]interface{} {
 	return n.response.Map
 }
 
-type (
-	AccountCreateResult = AccountReadResult
-	AccountUpdateResult = AccountReadResult
-)
+type AccountCreateResult = AccountReadResult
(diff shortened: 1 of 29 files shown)


api/accounts/account.gen.go
```diff
@@ -57,10 +57,8 @@ func (n AccountReadResult) GetResponseMap() map[string]interface{} {
 	return n.response.Map
 }
 
-type (
```
talanknight, line 60: This appears to revert some gofumpt changes like this one.

talanknight (dismissed):
Looks good! (Please see my comment about go fumpt)

api/accounts/account.gen.go
jefferai: Ah I'll add running fmt at the end of api.

talanknight (approved):

PR #860, opened 2021-01-13
Groundwork for returning authorized actions

jefferai:
This lays out the initial groundwork for returning authorized actions within the acl/auth system, including a test, and applies it to target resources.

Once this is approved, I'll sweep through the other resources in this model.


internal/perms/acl.go
```diff
@@ -123,6 +123,21 @@ func (a ACL) Allowed(r Resource, aType action.Type) (results ACLResults) {
 	return
 }
 
+// AllowedActions returns the set of actions out of a given possible set that
+// are allowed given the scope/id/type/pin. Returns nil if no match.
+func (a ACL) AllowedActions(r Resource, aTypes action.Actions) action.Actions {
```
louisruch, line 128: Curious about the naming choice of `aTypes`, why not `available` or `actions`?

louisruch (approved):
LGTM

internal/types/action/action.go
```diff
@@ -107,3 +107,16 @@ func (a Type) String() string {
 		"remove-accounts",
 	}[a]
 }
+
+type Actions []Type
```
jimlambrt, line 111: mind adding godocs for the new exported type and its Strings() receiver function?

internal/servers/controller/handlers/targets/target_service.go
```diff
@@ -34,7 +35,18 @@ import (
 	"google.golang.org/protobuf/types/known/wrapperspb"
 )
 
-var maskManager handlers.MaskManager
+var (
+	maskManager     handlers.MaskManager
+	TargetIdActions = action.Actions{
```
jimlambrt, line 40: mind adding godoc comment for new exported var?

internal/types/action/action.go
```diff
@@ -107,3 +107,16 @@ func (a Type) String() string {
...
+
+func (a Actions) Strings() []string {
```
jimlambrt, line 113: Nice and simple.

internal/auth/option.go
```diff
@@ -87,3 +89,9 @@ func WithAnonymousUserNotAllowed(notAllowed bool) Option {
 		o.withAnonymousUserNotAllowed = notAllowed
 	}
 }
+
+func WithResource(resource *perms.Resource) Option {
```
jimlambrt, line 93: Mind adding docs to this exported func?   Also, it appears there aren't unit tests for these options.  Can you start a new pattern and add one for this new option, which will align with what the other repos are doing?

internal/perms/acl.go
```diff
@@ -123,6 +123,21 @@ func (a ACL) Allowed(r Resource, aType action.Type) (results ACLResults) {
 	return
 }
 
+// AllowedActions returns the set of actions out of a given possible set that
+// are allowed given the scope/id/type/pin. Returns nil if no match.
```
jimlambrt, line 127: Mind adding a unit test for this new func?

internal/types/action/action.go
jimlambrt: Mind adding a simple unit test?

jimlambrt (changes requested):
Looks great.  Just a couple suggestions.

internal/perms/acl.go
jefferai: No particular reason. I changed it to `avail`.

internal/types/action/action.go
jefferai: Done.

internal/servers/controller/handlers/targets/target_service.go
jefferai: It's only exported because the test is in a `_test` package so I don't intend for others to use it, but I can.

internal/auth/option.go
jefferai: Done.

internal/perms/acl.go
jefferai: Moved this logic into FetchActionsForId which has tests.

jimlambrt (approved):
Looks great!

PR #865, opened 2021-01-15
ICU-744/Refactor internal/target to new domain errors

louisruch:
### What does this PR do

Refactor internal/target to new domain errors


internal/target/repository.go
```diff
@@ -576,24 +595,24 @@ func (r *Repository) SetTargetHostSets(ctx context.Context, targetId string, tar
 			msgs := make([]*oplog.Message, 0, 2)
 			targetTicket, err := w.GetTicket(target)
 			if err != nil {
-				return fmt.Errorf("set target host sets: unable to get ticket: %w", err)
+				return errors.Wrap(err, op)
```
jimlambrt, line 598: Other ticket errors included:  errors.WithMsg("unable to get ticket")

jimlambrt (changes requested):
Just one small nit.

internal/target/repository.go
louisruch: Oh good catch thanks!

internal/target/repository.go
```diff
@@ -32,14 +29,15 @@ type Repository struct {
 // NewRepository creates a new target Repository. Supports the options: WithLimit
 // which sets a default limit on results returned by repo operations.
 func NewRepository(r db.Reader, w db.Writer, kms *kms.Kms, opt ...Option) (*Repository, error) {
+	const op = "target.NewRepository"
 	if r == nil {
-		return nil, stderrors.New("error creating db repository with nil reader")
+		return nil, errors.New(errors.InvalidParameter, op, "nil reader")
```
s-christoff, line 34: Is there a nil error code we can return instead? Feels like it is used a lot 🤔

internal/target/repository.go
```diff
@@ -58,10 +56,11 @@ func NewRepository(r db.Reader, w db.Writer, kms *kms.Kms, opt ...Option) (*Repo
 // with its host set ids.  If the target is not found, it will return nil, nil, nil.
 // No options are currently supported.
 func (r *Repository) LookupTarget(ctx context.Context, publicIdOrName string, opt ...Option) (Target, []*TargetSet, error) {
+	const op = "target.(Repository).LookupTarget"
 	opts := getOpts(opt...)
 
 	if publicIdOrName == "" {
-		return nil, nil, fmt.Errorf("lookup target: missing private id: %w", errors.ErrInvalidParameter)
+		return nil, nil, errors.New(errors.InvalidParameter, op, "missing public id")
```
s-christoff, line 63: Was the prior `private` a typo? I am assuming it is because of `if publicIdOrName`

internal/target/repository.go
```diff
@@ -71,27 +70,27 @@ func (r *Repository) LookupTarget(ctx context.Context, publicIdOrName string, op
...
+			return nil, nil, errors.New(errors.InvalidParameter, op, "unknown combination of parameters")
 		}
 	} else {
 		switch {
 		case !scopeIdEmpty:
-			return nil, nil, fmt.Errorf("lookup target: passed in scope ID when using target ID for lookup: %w", errors.ErrInvalidParameter)
+			return nil, nil, errors.New(errors.InvalidParameter, op, "passed in scope ID when using target ID for lookup")
 		case !scopeNameEmpty:
-			return nil, nil, fmt.Errorf("lookup target: passed in scope name when using target ID for lookup: %w", errors.ErrInvalidParameter)
+			return nil, nil, errors.New(errors.InvalidParameter, op, "passed in scope name when using target ID for lookup")
```
s-christoff, line 93: Weird question - would it be more beneficial to add what we want in these errors? 
Like, "passed in scope name wanted target ID for lookup"?

s-christoff (approved):

internal/target/repository.go
louisruch: I have been trying, except for obvious mistakes, to keep these as pure refactors.  This means I am keeping the error message the same as before.  Changes to behavior, or message in this case, could easily go unnoticed in a review.

For this specific question, the use of `wanted` could perhaps be more clear, but the current message is clear enough IMO.

internal/target/repository.go
louisruch: Yes we are checking `publicIdOrName` being empty here so I updated the error message returned

internal/target/repository.go
louisruch: `InvalidParameter` is the generic type used for all these parameter checks, we don't really care at the `Code` level if it was nil, empty, set when it should not have been set, or some other state.  What we care about is that the parameter sent to the function was invalid.

PR #872, opened 2021-01-19
Adding Migrate Command

talanknight:
Add a `boundary database migrate` command.  Set the dirty bit for the entire initialization and migration phase, not just for the parts that mutate the db schema.


internal/cmd/commands/database/init.go
```diff
@@ -257,15 +257,15 @@ func (c *InitCommand) Run(args []string) (retCode int) {
 		migrationUrlToParse = urlToParse
 	}
 
-	dbaseUrl, err := config.ParseAddress(urlToParse)
+	migrationUrl, err := config.ParseAddress(migrationUrlToParse)
```
jefferai, line 260: Good catch

internal/cmd/commands/database/init.go
```diff
@@ -275,6 +275,17 @@ func (c *InitCommand) Run(args []string) (retCode int) {
 		c.UI.Error(fmt.Errorf("Error setting up schema manager for locking: %w", err).Error())
 		return 1
 	}
+	// This is an advisory locks on the DB which is released when the db session ends.
```
jefferai, line 278:
```suggestion
	// This is an advisory lock on the DB which is released when the DB session ends.
```
(I don't particularly care about DB/db but it ought to be consistent within the same comment line :-D )

internal/cmd/commands/database/init.go
jefferai: Oh I see we were using it later on before.

internal/cmd/commands/database/init.go
```diff
@@ -546,3 +548,12 @@ func (c *InitCommand) ParseFlagsAndConfig(args []string) int {
 
 	return 0
 }
+
+func (c *InitCommand) unsetDirty(man *schema.Manager) int {
+	if err := man.UnsetDirty(c.Context); err != nil {
+		c.UI.Error(base.WrapAtLength("Database could not be marked undirty after" +
```
jefferai, line 554: Note that since you don't have a space after `after` this will end up as `...undirty afterinitialization`. Might as well just put it on one line and let WrapAtLength do its thing.

internal/cmd/commands/database/init.go
```diff
@@ -287,54 +298,45 @@ func (c *InitCommand) Run(args []string) (retCode int) {
...
+		c.UI.Error(base.WrapAtLength("Database could not be marked as dirty in " +
+			"preperation for it's initialization."))
 		return 1
 	}
 
 	// Core migrations using the migration URL
 	{
 		migrationUrl = strings.TrimSpace(migrationUrl)
-		ran, err := schema.InitStore(c.Context, dialect, migrationUrl)
+		ran, err := schema.MigrateStore(c.Context, dialect, migrationUrl, schema.WithSkipSetDirty(true), schema.WithSkipUnsetDirty(true))
```
jefferai, line 318: Can you add a comment here and explain why you're skipping setting dirty and setting undirty? It's not clear to me especially given that you do call unsetDirty later, so if you're skipping setDirty...

internal/cmd/commands/database/migrate.go
```diff
@@ -0,0 +1,311 @@
...
+
+func (c *MigrateCommand) Help() string {
+	return base.WrapForHelpText([]string{
+		"Usage: boundary database migrate [options]",
+		"",
+		"  Migrate Boundary's database:",
+		"",
+		"    $ boundary database migrate -config=/etc/boundary/controller.hcl",
+		"",
+		"",
```
jefferai, line 55: Can you remove this second extra line?

internal/cmd/commands/database/migrate.go
```diff
@@ -0,0 +1,311 @@
...
+	})
+
+	f.StringVar(&base.StringVar{
+		Name:       "log-format",
+		Target:     &c.flagLogFormat,
+		Completion: complete.PredictSet("standard", "json"),
+		Usage:      `Log format. Supported values are "standard" and "json".`,
+	})
+
+	f = set.NewFlagSet("Init options")
```
jefferai, line 101:
```suggestion
	f = set.NewFlagSet("Migration options")
```

internal/cmd/commands/database/migrate.go
```diff
@@ -0,0 +1,311 @@
...
+
+	if schema.DevMigration(dialect) != c.flagAllowDevMigrations {
+		if schema.DevMigration(dialect) {
+			c.UI.Error(base.WrapAtLength("This version of the binary has " +
+				"dev database schema updates which may not be supported in the " +
+				"next official release. To proceed anyways please use the " +
+				"'-allow-development-migrations' flag."))
+			return 2
+		} else {
+			c.UI.Error(base.WrapAtLength("The '-allow-development-migrations' " +
```
jefferai, line 149: I don't think this should be an error. A user running dev mode may always want to specify that flag just in case, and we shouldn't error out just because there happen to be none at this time.

internal/cmd/commands/database/migrate.go
```diff
@@ -0,0 +1,311 @@
...
+		return 1
+	}
+
+	if c.Config.Controller.Database == nil {
+		c.UI.Error(`"controller.database" config block not found`)
+		return 1
+	}
+
+	urlToParse := c.Config.Controller.Database.Url
+	if urlToParse == "" {
```
jefferai, line 185: We actually don't need this if a migration URL is set. (Technically in `init` if a migration URL is set, we don't need this either if they also skip auto creation of all initial resources.)

We could leave it in as an error, but I could imagine a use-case where someone has a config that _only_ has a migration URL so they can do upgrades but don't have access to credentials that let them modify actual data outside of the migrations. Feels nicer to me to take it out.

internal/db/schema/manager.go
```diff
@@ -127,8 +156,9 @@ func (b *Manager) ExclusiveUnlock(ctx context.Context) error {
 }
 
 // RollForward updates the database schema to match the latest version known by
-// the boundary binary.  An error is not returned if the database is already at
-// the most recent version.
+// the boundary binary. It is not an error for the database to be at the
+// most recent version.
+// An error is returned if the dirty bit has not been set.
```
jefferai, line 161: Can you explain this change? Just not following, why would we _want_ the dirty bit to be set after doing migrations?

internal/cmd/commands/database/migrate.go
talanknight: I worry that this flag would end up seen as a "just-make-it-work" flag which ends up being placed everywhere, including in production, where it doesn't belong.  I agree that when there isn't a dev schema change and they've built from head it could get annoying to have to unset this flag, but I'm weighing it against the potential for data loss in production.

internal/db/schema/manager.go
talanknight: The intent is to fix the behavior of the init command.  Currently if we initialize the schema but fail when setting the default resources the database is considered initialized and not dirty which is not correct IMO.

internal/cmd/commands/database/migrate.go
talanknight: Fixed, Thanks.

internal/cmd/commands/database/migrate.go
jefferai: Wouldn't you not have development migrations in production though? I agree it might just end up being used all over, however, it _ought_ to be a no-op in prod. And in many cases people will indeed want it every time they are running in dev cause they'll start with a known base DB.

internal/cmd/commands/database/init.go
```diff
@@ -287,54 +298,48 @@ func (c *InitCommand) Run(args []string) (retCode int) {
...
-		}
-	}()
-
-	migrationUrl, err := config.ParseAddress(migrationUrlToParse)
-	if err != nil && err != config.ErrNotAUrl {
-		c.UI.Error(fmt.Errorf("Error parsing migration url: %w", err).Error())
+	// The initialization is dirty as long as it doesn't complete successfully.
+	if err := man.SetDirty(c.Context); err != nil {
+		c.UI.Error(base.WrapAtLength("Database could not be marked as dirty in " +
+			"preparation for it's initialization."))
```
jefferai, line 311:
```suggestion
			"preparation for its initialization."))
```

internal/cmd/commands/database/init.go
```diff
@@ -287,54 +298,48 @@ func (c *InitCommand) Run(args []string) (retCode int) {
...
 		return 1
 	}
 
 	// Core migrations using the migration URL
 	{
 		migrationUrl = strings.TrimSpace(migrationUrl)
-		ran, err := schema.InitStore(c.Context, dialect, migrationUrl)
+		// We skip setting and unsetting dirty in MigrateStore so this command can manage it instead
+		// since this command does more than simply mutate the schema and erroring out prior to adding
+		// the initial resources would leave the initialization in a dirty state.
```
jefferai, line 320:
```suggestion
		// the initial resources would leave the initialization in a dirty state. Note that we have
		// already set it dirty just prior to this block.
```

talanknight:
I've made quite a few changes here now including checking for KMS at controller startup, checking oplog being populated at init, share migrate schema logic between init and migrate command, and remove roll back from generated code.

PTAL

internal/cmd/commands/database/migrate.go
talanknight: We discussed over slack that this can be a warning instead.

internal/cmd/commands/server/server.go
```diff
@@ -630,3 +633,23 @@ func (c *Command) Reload() error {
...
+	kmsRepo, err := kms.NewRepository(rw, rw)
+	if err != nil {
+		return fmt.Errorf("error creating kms repository: %w", err)
+	}
+	rks, err := kmsRepo.ListRootKeys(c.Context, kms.WithLimit(1))
+	if err != nil {
+		return err
+	}
+	for _, rk := range rks {
+		if rk.GetScopeId() == scope.Global.String() {
```
jefferai, line 650: @jimlambrt Do you think we should list scopes as well, and then verify that we have a root key for each scope? It could also be a future enhancement. At some point we could get all the way down the rabbit hole and verify that for each known type of key we have a key for each scope, but I don't know where the point of diminishing returns is here.

internal/db/schema/postgres/postgres.go
```diff
@@ -52,6 +52,8 @@ const (
 var defaultMigrationsTable = "boundary_schema_version"
 
 // Postgres is a driver usable by a boundary schema manager.
+// Postgres is not thread safe to be thread safe it should separate the schema access lock id used for internal
+// operations and the one exposed externally.
```
jefferai, line 56: Is this a comment on the code or on postgres itself? If the former is there a way to (eventually) fix it?

jefferai:
I think it's looking pretty good, just a couple of questions/comments.

I assume wrapping migrations in a transaction will come in a different PR(?) as that seems like it'd be easier/cleaner as it's an enhancement onto this base.

internal/db/schema/postgres/postgres.go
talanknight: This is a comment on this code.  I had intended for the comment after to indicate 1 way that this code could be fixed if we wanted it to.  Since we use it only inside the manager, which should capture a lock before running it might not make sense to make this thread safe.  I'll simplify the comment to just be "This struct is not thread safe."

talanknight:
Wrapping migrations in a tx is planned for a follow up PR, which I'll send out as soon as this PR is merged.

internal/cmd/commands/database/init.go
```diff
@@ -538,3 +490,17 @@ func (c *InitCommand) ParseFlagsAndConfig(args []string) int {
 
 	return 0
 }
+
+func (c *InitCommand) verifyOplogIsEmpty() error {
+	const op = "database.(InitCommand).verifyOplogIsEmpty"
+	r := c.srv.Database.DB().QueryRowContext(c.Context, "select not exists(select 1 from oplog_entry limit 1)")
```
mgaffney, line 496: Nice query!

internal/cmd/commands/database/migrate.go
```diff
@@ -0,0 +1,253 @@
...
+	flagConfig             string
+	flagConfigKms          string
+	flagLogLevel           string
+	flagLogFormat          string
+	flagMigrationUrl       string
+	flagAllowDevMigrations bool
+}
+
+func (c *MigrateCommand) Synopsis() string {
+	return "Initialize Boundary's database"
```
mgaffney, line 42: This is the same synopsis as the `database init` command. Should it be different?

internal/cmd/commands/server/server.go
```diff
@@ -363,17 +367,12 @@ func (c *Command) Run(args []string) int {
...
-			return 1
-		}
 		if ckState.Dirty {
 			c.UI.Error(base.WrapAtLength("Database is in a bad state. Please revert the database into the last known good state."))
 			return 1
 		}
 		if ckState.BinarySchemaVersion > ckState.DatabaseSchemaVersion {
-			// TODO: Add the command to migrate up the schema version once that command exists.
-			c.UI.Error("Older schema version is than is expected from this binary.")
+			c.UI.Error(base.WrapAtLength("Older schema version is than is expected from this binary. Run 'boundary database migrate' to update the database."))
```
mgaffney, line 375:
```suggestion
			c.UI.Error(base.WrapAtLength("Database schema must be updated to use this version. Run 'boundary database migrate' to update the database."))
```

mgaffney (approved):

jefferai (approved):

internal/cmd/commands/database/migrate.go
talanknight: Changed it. Thanks!

mgaffney (approved):

internal/cmd/commands/server/server.go
jimlambrt: For now, I think this is probably sufficient.  A scope's creation along with it's associated root key and deks are all in one transaction.  I don't think, the API provides anyway for an operator to delete keys on their own... so at this point, I don't think the db can get into a state where there's a scope without it's root and deks.

internal/cmd/commands/database/funcs.go
```diff
@@ -1,11 +1,71 @@
...
+		ui.Error(fmt.Errorf("Error establishing db connection: %w", err).Error())
+		return noop, 1
+	}
+	if err := dBase.PingContext(ctx); err != nil {
+		ui.Error(fmt.Sprintf("Unable to connect to the database at %q", u))
+		return noop, 1
+	}
+	man, err := schema.NewManager(ctx, dialect, dBase)
+	if err != nil {
+		if errors.Match(errors.T(errors.MigrationLock), err) {
```
jimlambrt, line 33: Happy to see Match used here.

jimlambrt (approved):
Looks great.  Nice to not see the "down"s any longer.

PR #875, opened 2021-01-21
Fix lockup when initalizing database

talanknight:


internal/cmd/commands/database/init.go
```diff
@@ -257,15 +257,16 @@ func (c *InitCommand) Run(args []string) (retCode int) {
...
 	}
 
-	dbaseUrl, err := config.ParseAddress(urlToParse)
+	migrationUrl, err := config.ParseAddress(migrationUrlToParse)
 	if err != nil && err != config.ErrNotAUrl {
-		c.UI.Error(fmt.Errorf("Error parsing database url: %w", err).Error())
+		c.UI.Error(fmt.Errorf("Error parsing migration url: %w", err).Error())
 		return 1
 	}
+	migrationUrl = strings.TrimSpace(migrationUrl)
```
louisruch, line 265: hmm, I would imagine `ParseAddress` should be returning an address with spaces already trimmed?  I see this is used elsewhere so it's fine just more curiosity I guess

louisruch (approved):
LGTM
Only one question, probably not even needed in this review, but it came up so :)

internal/cmd/commands/database/init.go
jefferai: Agree, if it's not trimming it probably should.

talanknight: Since this is an existing issue I'll fix it in my next PR which adds the migration command.

PR #877, opened 2021-01-21
ICU-745/Refactor internal/session to new domain errors

louisruch:
### What does this PR do

Refactor internal/session to new domain errors


internal/iam/repository_user.go
```diff
@@ -131,7 +131,7 @@ func (r *Repository) UpdateUser(ctx context.Context, user *User, version uint32,
 			}
 			if rowsUpdated > 1 {
 				// return err, which will result in a rollback of the update
-				return errors.New(errors.MultipleRecords, op, "error more than 1 resource would have been updated")
+				return errors.New(errors.MultipleRecords, op, "more than 1 resource would have been updated")
```
louisruch, line 134: This is just bringing the error message inline with all other of this type

internal/session/repository_session.go
```diff
@@ -306,28 +314,28 @@ func (r *Repository) TerminateSession(ctx context.Context, sessionId string, ses
 		func(reader db.Reader, w db.Writer) error {
 			rowsAffected, err := w.Exec(ctx, terminateSessionCte, []interface{}{sessionId, sessionVersion})
 			if err != nil {
-				return fmt.Errorf("unable to terminate session %s: %w", sessionId, err)
+				return errors.Wrap(err, op, errors.WithMsg(fmt.Sprintf("unable to terminate session %s", sessionId)))
 			}
 			if rowsAffected == 0 {
-				return fmt.Errorf("unable to terminate session %s", sessionId)
+				return errors.New(errors.SessionNotFound, op, fmt.Sprintf("unable to terminate session %s", sessionId))
```
louisruch, line 320: I made this session not found, but I am not convinced that is always the case for a non-errored failure to update?

internal/session/session.go
```diff
@@ -209,50 +209,51 @@ func (s *Session) Clone() interface{} {
...
 		case contains(opts.WithFieldMaskPaths, "ConnectionLimit"):
-			return fmt.Errorf("session vet for write: connection limit is immutable: %w", errors.ErrInvalidParameter)
+			return errors.New(errors.InvalidParameter, op, "connection limit is immutable")
 		case contains(opts.WithFieldMaskPaths, "WorkerFilter"):
-			return fmt.Errorf("session vet for write: worker filter is immutable: %w", errors.ErrInvalidParameter)
+			return errors.New(errors.InvalidParameter, op, "worker filter is immutable")
 		case contains(opts.WithFieldMaskPaths, "TerminationReason"):
 			if _, err := convertToReason(s.TerminationReason); err != nil {
-				return fmt.Errorf("session vet for write: termination reason '%s' is invalid: %w", s.TerminationReason, errors.ErrInvalidParameter)
+				return errors.Wrap(err, op)
```
louisruch, line 256: `convertToReason` is already creating the exact same error message as created here, so I just wrapped the error returned instead of making a new error here

louisruch:
@jimlambrt @s-christoff PTAL

internal/session/repository_session.go
s-christoff: Would `InvalidSessionState` be a better err here?

louisruch:
ping @jim-hashicorp

louisruch:
ping @jimlambrt

internal/session/repository_session.go
```diff
@@ -559,20 +572,21 @@ func (r *Repository) CloseConnections(ctx context.Context, closeWith []CloseWith
 // ErrSessionNotPending error if a connection cannot be made because the session
```
jimlambrt, line 572: ErrSessionNotPending is no longer a valid error... can you update the docs?

internal/session/error.go
```diff
@@ -8,12 +8,4 @@ var (
 	// ErrInvalidStateForOperation indicates the session's state does not allow the
 	// operation.
 	ErrInvalidStateForOperation = errors.New("state is invalid for operation")
```
jimlambrt, line 10: Why not refactor this and the dependent code in repository_session.go to use the new errors?

internal/session/repository_session.go
```diff
@@ -210,34 +212,35 @@ func (r *Repository) ListSessions(ctx context.Context, opt ...Option) ([]*Sessio
...
+		return nil, errors.Wrap(err, op)
 	}
 	defer rows.Close()
 
 	var sessionsWithState []*sessionView
 	for rows.Next() {
 		var s sessionView
 		if err := r.reader.ScanRows(rows, &s); err != nil {
-			return nil, fmt.Errorf("changes: scan row failed: %w", err)
+			return nil, errors.Wrap(err, op)
```
jimlambrt, line 223: consider adding a contextual msg here.

internal/session/repository_session.go
jimlambrt: bad state or unknown error.

internal/session/repository_session.go
```diff
@@ -417,30 +427,32 @@ type ConnectionAuthzSummary struct {
...
 	var info *ConnectionAuthzSummary
 	for rows.Next() {
 		if info != nil {
-			return nil, fmt.Errorf("session summary: query returned more than one row")
+			return nil, errors.New(errors.MultipleRecords, op, "query returned more than one row")
 		}
 		info = &ConnectionAuthzSummary{}
 		if err := r.reader.ScanRows(rows, info); err != nil {
-			return nil, fmt.Errorf("session summary: scan row failed: %w", err)
+			return nil, errors.Wrap(err, op)
```
jimlambrt, line 444: consider addint a contextual msg

internal/session/repository_session.go
```diff
@@ -367,6 +376,7 @@ func (r *Repository) TerminateCompletedSessions(ctx context.Context) (int, error
 // and returns it, along with it's states.  If the authorization fails, it
 // an error of ErrInvalidStateForOperation.
```
jimlambrt, line 377: I wish we were not raising gRPC status errors in the domain... such a bad pattern for the domain layer to be protocol aware.

internal/session/error.go
jimlambrt: Ah.. figured it out.. it's the protocol aware anti-pattern. Would be nice to fix this somehow.

jimlambrt (changes requested):
Excellent work.  Just a few questions.

internal/session/error.go
louisruch: Yes, I am going to have a quick follow-up PR, I chatted with a couple of people and I think we can just nuke the GRPC status returns here, however, I did not want to sneak it in this PR.

louisruch:
@jimlambrt address feedback, I did not want to rip out the gRPC status returns in this PR as I wanted to hit a wider audience with the change so thought it would be best to have it in it's own smaller PR.

s-christoff (dismissed):
a few nits but i don't see anything bad

jimlambrt (dismissed):
Looks good.

PR #879, opened 2021-01-22
Add scope repo function to recursively list

jefferai:
A simple function that returns an appropriate set of scopes based on the
incoming scope ID. Includes a test.


jimlambrt (approved):
Looks great!

PR #881, opened 2021-01-22
Add recursive listing to roles

jefferai:
Notes:

* Nearly all of the logic in role_service.go can (and will) be refactored into a shared function when I sweep through the rest of the resource types
* I removed the additional verification logic because it's not used anywhere and the anticipated need for it has so far been satisfied by the later ActionSet functions. So I'm considering it dead code. It can always be revived from the repo history.
* There are some remaining tests to be written


jefferai:
Taking out of draft so it can be commented on, but see notes above

internal/iam/repository_role.go
```diff
@@ -182,13 +182,13 @@ func (r *Repository) DeleteRole(ctx context.Context, withPublicId string, _ ...O
...
+func (r *Repository) ListRoles(ctx context.Context, withScopeIds []string, opt ...Option) ([]*Role, error) {
 	const op = "iam.(Repository).ListRoles"
-	if withScopeId == "" {
-		return nil, errors.New(errors.InvalidParameter, op, "missing scope id")
+	if len(withScopeIds) == 0 {
+		return nil, errors.New(errors.InvalidParameter, op, "missing scope ids")
 	}
 	var roles []*Role
-	err := r.list(ctx, &roles, "scope_id = ?", []interface{}{withScopeId}, opt...)
+	err := r.list(ctx, &roles, "scope_id in (?)", []interface{}{withScopeIds}, opt...)
```
jimlambrt, line 191: I'm interested to see how this bit o'magic works in gorm v2.  I have a suspicion that we may need to expand the "?" to match the number of paramters, but that's just a suspicion.. we'll see.  With that said, this is so deceptively simple.

jimlambrt (approved):
Looks great!  Lots of plumbing to do now.  Excellent work.  (PS. don't forget to add the needed tests at some point).

internal/iam/repository_role.go
```diff
@@ -182,13 +182,13 @@ func (r *Repository) DeleteRole(ctx context.Context, withPublicId string, _ ...O
 }
 
 // ListRoles in a scope and supports WithLimit option.
-func (r *Repository) ListRoles(ctx context.Context, withScopeId string, opt ...Option) ([]*Role, error) {
+func (r *Repository) ListRoles(ctx context.Context, withScopeIds []string, opt ...Option) ([]*Role, error) {
```
mgaffney, line 185: Can you update the godoc for the method to reflect the changes in the signature?

mgaffney (changes requested):
Overall looks good. Just one minor comment about updating the godoc.

internal/iam/repository_role.go
jefferai: Sure, done.

jefferai:
Added test and merged

PR #887, opened 2021-01-25
Wrap postgres up migrations into a single transaction

talanknight:
This strips out transactions from the .up.sql files and has the schema manager control the migration during roll forward.


internal/db/schema/migrations/generate/generate.go
```diff
@@ -96,9 +90,23 @@ func generate(dialect string) {
...
 				largestSchemaVersion = fullV
 			}
+
+			cbts, err := ioutil.ReadFile(fmt.Sprintf("%s/%s/%s/%s", srcDir, dialect, ver, name))
+			if err != nil {
+				fmt.Printf("error opening file %s with dialect %s: %v", name, dialect, err)
+				os.Exit(1)
+			}
+
+			contents := strings.TrimSpace(string(cbts))
```
jefferai, line 100: An alternate way to do this which might be a tiny bit "safer" (in case anyone does "Begin" or something) would be to split the lines, then strip out the first line if its lower-case equivalent contains "begin", and do the same thing with end, then reconstruct it with newlines.

Either way works, I think it's unlikely anyone will do mixed case and it will be caught immediately if they do.

jefferai (approved):
I think it looks good to me, pending fixing the failing tests

talanknight:
Fixed the failing test.  @mgaffney Please take a look.

internal/db/schema/migrations/generate/generate.go
talanknight: Please take a look at my new approach where I just compare the lowercase version of the first len("begin;") characters and the last len("commit;") characters.

internal/db/schema/postgres/postgres.go
```diff
@@ -228,16 +278,24 @@ func runesLastIndex(input []rune, target rune) int {
 
 // SetVersion sets the version number, and whether the database is in a dirty state.
 // A version value of -1 indicates no version is set.
-func (p *Postgres) SetVersion(ctx context.Context, version int, dirty bool) error {
+func (p *Postgres) setVersion(ctx context.Context, version int, dirty bool) error {
```
mgaffney, line 281: It would be helpful to update the first word of the function doc from `SetVersion` to `setVersion` to reflect the name.

mgaffney (approved):
Looks good. I left one minor comment.

mgaffney (approved):

PR #888, opened 2021-01-27
Create Sessions List Self Action

talanknight:
Allows listing all sessions for the user currently making a request.


internal/session/repository_session.go
```diff
@@ -170,8 +170,7 @@ func (r *Repository) ListSessions(ctx context.Context, opt ...Option) ([]*Sessio
 	var args []interface{}
 
 	inClauseCnt := 0
-	switch {
-	case len(opts.withScopeIds) != 0:
+	if len(opts.withScopeIds) != 0 {
```
jefferai, line 173: It might be worth checking with @jimlambrt to see if he had a reason for this to be either/or instead of and -- I think what you changed it to will work, but not sure offhand why it was like that in the first place.

jefferai: (BTW I fully realize this _must_ be changed for this to work, but it seems like it should have been this way to begin with, hence the question)

internal/servers/controller/handlers/sessions/session_service.go
```diff
@@ -79,35 +79,55 @@ func (s Service) ListSessions(ctx context.Context, req *pbs.ListSessionsRequest)
...
-
 	finalItems := make([]*pb.Session, 0, len(seslist))
 	res := &perms.Resource{
-		Type: resource.Session,
+		ScopeId: authResults.Scope.Id,
+		Type:    resource.Session,
 	}
 	for _, item := range seslist {
 		item.Scope = scopeInfoMap[item.GetScopeId()]
-		res.ScopeId = item.Scope.Id
```
jefferai, line 104: This change needs to be reverted -- if it's recursive, the session list can have values from different scopes. So the scope for authorized actions must come from each item as it's tested, not be set by the auth scope.

internal/servers/controller/handlers/sessions/session_service.go
```diff
@@ -79,35 +79,55 @@ func (s Service) ListSessions(ctx context.Context, req *pbs.ListSessionsRequest)
...
 		item.AuthorizedActions = authResults.FetchActionSetForId(ctx, item.Id, IdActions, auth.WithResource(res)).Strings()
```
jefferai, line 125: Not a comment on this line but I can't comment above -- CollectionActions should have ListSelf added to it, at the top of the file.

internal/servers/controller/handlers/sessions/session_service.go
```diff
@@ -196,6 +216,8 @@ func (s Service) authResult(ctx context.Context, id string, a action.Type) auth.
 			res.Error = handlers.NotFoundError()
 			return res
 		}
+	case action.ListSelf:
```
jefferai, line 219: I think this is just `case action.List, action.ListSelf` above?

internal/servers/controller/handlers/sessions/session_service_test.go
```diff
@@ -292,6 +292,119 @@ func TestList(t *testing.T) {
...
+		{
+			name: "List Many Sessions",
+			req:  &pbs.ListSelfSessionsRequest{ScopeId: pWithSessions.GetPublicId()},
+			res:  &pbs.ListSelfSessionsResponse{Items: wantSession},
+		},
+		{
+			name: "List No Sessions",
+			req:  &pbs.ListSelfSessionsRequest{ScopeId: pNoSessions.GetPublicId()},
+			res:  &pbs.ListSelfSessionsResponse{},
+		},
```
jefferai, line 385: Can you make user parameterized, and then add a self lookup for user 2? You create a test session using `at2` above but then it's not actually used in the test.

internal/servers/controller/handlers/sessions/session_service.go
jefferai: Oh also -- we should add list:self to default role permissions for authenticated users for global/org scopes. We don't need to upgrade existing roles, that's up to users to decide if they want to add it.

internal/session/repository_session.go
jimlambrt: Hum... looks as if I believed you'd want to search for user sessions across scopes or just list sessions by scope.  Honestly, don't remember why I went down this path.

internal/servers/controller/handlers/sessions/session_service.go
talanknight: Added.  Thanks!

internal/servers/controller/handlers/sessions/session_service.go
talanknight: Yes.  I thought I had fixed this before sending it out.  It is fixed now. Sorry for the noise.

internal/servers/controller/handlers/sessions/session_service.go
talanknight: Yes, I don't know what happened to this file. I thought I had fixed it up beyond this prior to sending out the PR.  Sorry.  I've fixed this and the other places in this file where things are obviously wrong.  PTAL.

internal/iam/repository_scope.go
```diff
@@ -286,11 +286,18 @@ func (r *Repository) CreateScope(ctx context.Context, s *Scope, userId string, o
 					}
 					grants = append(grants, roleGrant)
 
+					roleGrant, err = NewRoleGrant(defaultRolePublicId, "type=scope;actions=list:self")
```
jefferai, line 289:
```suggestion
					roleGrant, err = NewRoleGrant(defaultRolePublicId, "id=*;type=scope;actions=list:self")
```

jefferai (approved):
LGTM!

PR #889, opened 2021-01-27
Support upgrading from old migration table name

talanknight:
Checking the CurrentState now looks for both the old and new table and also reports if it has been initialized previously. 
Before we run the migration we update the old table's name, if present, to the new name.


internal/cmd/commands/database/init.go
```diff
@@ -491,6 +493,64 @@ func (c *InitCommand) ParseFlagsAndConfig(args []string) int {
 	return 0
 }
 
+// initDatabase initializes the schema to the binary's currently suppoorted version.
+// If init previously ran and left any indication that it ran an error is returned.
```
jefferai, line 497: Then.........

Don't leave us hanging :-D

internal/cmd/commands/database/init.go
```diff
@@ -491,6 +493,64 @@ func (c *InitCommand) ParseFlagsAndConfig(args []string) int {
...
+// This might be the case for early versions of boundary where the migrations
+// weren't completely executed in a transaction and a version table might have
+// been left despite a rollback.
+// Returns a cleanup function which must be called even if an error is returned and
+// an error code where a non-zero value indicates an error happened.
+func initDatabase(ctx context.Context, ui cli.Ui, dialect, u string) (func(), int) {
```
jefferai, line 503: There are only two tiny blocks that differ in this and migrateDatabase. I'm not sure why you pulled it out into two functions that are essentially the same, but could it be put back and simply use a bool to control if it's init vs. migrate behavior?

talanknight: Fixed this.

internal/db/schema/manager.go
```diff
@@ -156,22 +156,32 @@ func (b *Manager) RollForward(ctx context.Context) error {
...
+	}
+	if err := b.driver.EnsureVersionTable(ctx); err != nil {
+		return errors.Wrap(err, op)
 	}
 	for qp.Next() {
 		select {
 		case <-ctx.Done():
-			return errors.Wrap(ctx.Err(), op)
+			err := ctx.Err()
+			if commitErr := b.driver.CommitRun(); commitErr != nil {
```
talanknight, line 181: I'm not sure if this is the right thing to do if we have timed out.  I didn't really want to expose a rollback interface since it wouldn't be usable by non-transaction based drivers. @mgaffney

mgaffney: Committing the transaction if there is a time out seems like a bad option. You could add a `Rollbacker` interface with `Rollback` being the only method and then call it here if the driver supports it.

internal/db/schema/postgres/postgres.go
```diff
@@ -77,19 +77,14 @@ func New(ctx context.Context, instance *sql.DB) (*Postgres, error) {
...
-
 	return px, nil
 }
 
 // TrySharedLock attempts to capture a shared lock. If it is not successful it returns an error.
 // https://www.postgresql.org/docs/9.6/static/explicit-locking.html#ADVISORY-LOCKS
 func (p *Postgres) TrySharedLock(ctx context.Context) error {
 	const op = "postgres.(Postgres).TrySharedLock"
-	const query = "SELECT pg_try_advisory_lock_shared($1)"
+	const query = "select pg_try_advisory_lock_shared($1)"
```
mgaffney, line 87: Thank you!

internal/db/schema/postgres/postgres.go
```diff
@@ -323,33 +320,44 @@ func (p *Postgres) setVersion(ctx context.Context, version int, dirty bool) erro
...
+// CurrentState returns the version, if the database was ever initialized
+// previously, if it is currently in a dirty state, and any error. A version
+// value of -1 indicates no version is set.
+func (p *Postgres) CurrentState(ctx context.Context) (version int, previouslyRan, dirty bool, err error) {
 	const op = "postgres.(Postgres).CurrentState"
-	query := `SELECT Version, dirty FROM ` + pq.QuoteIdentifier(defaultMigrationsTable) + ` LIMIT 1`
+
+	tableQuery := `select table_name from information_schema.tables where table_schema=(select current_schema()) and table_name in ('schema_migrations', '` + defaultMigrationsTable + `')`
+	tableName := defaultMigrationsTable
+	if err = p.conn.QueryRowContext(ctx, tableQuery).Scan(&tableName); err != nil {
```
mgaffney, line 331: This hides if both tables exist and if so, which one would be selected. Should we guard against this?

internal/db/schema/postgres/postgres.go
```diff
@@ -323,33 +320,44 @@ func (p *Postgres) setVersion(ctx context.Context, version int, dirty bool) erro
...
+		if err == sql.ErrNoRows {
+			return nilVersion, false, false, nil
+		}
+		return nilVersion, false, false, errors.Wrap(err, op)
+	}
+
+	query := `select version, dirty from ` + pq.QuoteIdentifier(tableName) + ` limit 1`
```
mgaffney, line 338: This hides if more than one row exists. I think we should guard against this just to be safe.

mgaffney (changes requested):

mgaffney (approved):

PR #895, opened 2021-01-29
chore: bump ui commitish to latest from boundary-ui/main

randallmorey:
While there are no user-facing changes in admin, let's bump the commitish anyway to stay up-to-date.


gsusmi (approved):
Looks good. Ty!

gsusmi (approved):

PR #896, opened 2021-02-01
Basic types required for the OIDC auth method backend.

jimlambrt:
This PR adds the basic Go types (AuthMethod, Account, AudClaim, CallbackUrl, Certificate and SigningAlg) required for the backend, based on the recent OIDC schema changes.   These type implementations follow the existing backend pattern established in other features (password auth, iam, sessions, etc).   

The PR also includes a few minor fixes to the schema. 

@jefferai - This PR includes the new derived reader we discussed last week, which will be used to HMAC OIDC auth method client secrets.

I'm sorry for the size of the PR.  Since all the types follow the same pattern, it doesn't feel like breaking them all up into separate PRs would really make it any easier to review them.  Luckily, the PR copies an established pattern and the vast majority (70%) of it is either tests or generated code.

Also: notice that this PR is against the jimlambrt-odic branch.  This longer lived branch is where we'll do the backend dev until the OIDC auth method is complete and ready to merge to main


internal/auth/oidc/aud_claim.go
```diff
@@ -0,0 +1,96 @@
...
+
+// NewAudClaim creates a new in memory audience claim assigned to an OIDC
+// AuthMethod. It supports no options.  If an AuthMethod as assigned AudClaims,
+// then ID tokens issued from the provider must contain one of the assigned
+// audiences to be valid.
+//
+// For more info on oidc aud claims, see the oidc spec:
+// https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
+func NewAudClaim(authMethodId string, audClaim string) (*AudClaim, error) {
+	const op = "oidc.NewCallbackUrl"
```
talanknight, line 32:
```suggestion
	const op = "oidc.NewAudClaim"
```

internal/auth/oidc/auth_method.go
```diff
@@ -0,0 +1,193 @@
...
+//
+// ClientSecret equals the client's secret which will be encrypted when stored
+// in the database and an hmac representation will also be stored when every the
+// secret changes.  The secret is not returned via the API, the hmac is returned
+// so callers can determine if it's been updated.
+//
+// MaxAge equals the Maximum Authentication Age. Specifies the allowable elapsed
+// time in seconds since the last time the End-User was actively authenticated
+// by the OP. If the elapsed time is greater than this value, the OP MUST
+// attempt to actively re-authenticate the End-User. If MaxAge == -1, then it
```
talanknight, line 57: "If MaxAge == -1 ..."   should this be <=0  or does "0" also have a special meaning?  My expectation would have been "0" means the user should always be re-authenticated and "-1" means never.

internal/auth/oidc/auth_method.go
```diff
@@ -0,0 +1,193 @@
...
+	a.KeyId = cipher.KeyID()
+	if err := a.hmacClientSecret(cipher); err != nil {
+		return errors.Wrap(err, op)
+	}
+	return nil
+}
+
+// decrypt the auth method after reading it from the db
+func (a *AuthMethod) decrypt(ctx context.Context, cipher wrapping.Wrapper) error {
+	const op = "oidc.(AuthMethod).encrypt"
```
talanknight, line 171:
```suggestion
	const op = "oidc.(AuthMethod).decrypt"
```

internal/auth/oidc/callback_url.go
```diff
@@ -0,0 +1,98 @@
...
+	*store.CallbackUrl
+	tableName string
+}
+
+// NewCallbackUrl creates a new in memory callback assigned to an OIDC
+// AuthMethod.  It supports no options.
+//
+// For more info on oidc callbacks, see redirect_uri in the oidc spec:
+// https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
+func NewCallbackUrl(authMethodId string, callback *url.URL) (*CallbackUrl, error) {
```
talanknight, line 31: What is the point of having a callback url?  I had expected this to be hard coded in the API and not something stored in the database.  For example, if I have auth method `oidcam_123` then the callback url would always be `<whatever the domain is>/v1/auth-methods/oidc_123:callback`.

internal/auth/oidc/certificate.go
```diff
@@ -0,0 +1,101 @@
...
+// Certificates.
+type Certificate struct {
+	*store.Certificate
+	tableName string
+}
+
+// NewCertificate creates a new in memory certificate assigned to and OIDC auth
+// method.
+func NewCertificate(authMethodId string, certificatePem string) (*Certificate, error) {
+	const op = "oidc.NewCallbackUrl"
```
talanknight, line 29:
```suggestion
	const op = "oidc.NewCertificate"
```

internal/auth/oidc/client_secret.go
```diff
@@ -0,0 +1,22 @@
...
+// RedactedClientSecret is the redacted string or json for an oidc id_token.
+const redactedClientSecret = "[REDACTED: OIDC client_secret]"
+
+// String will redact the client_secret.
+func (s ClientSecret) String() string {
+	return redactedClientSecret
+}
+
+// MarshalJSON will redact the client_secret.
+func (s ClientSecret) MarshalJSON() ([]byte, error) {
```
talanknight, line 20: For my own education, when does our code marshal these values?

internal/auth/oidc/ids.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/errors"
+)
+
+const (
+	// AuthMethodPrefix defines the prefix for AuthMethod public ids.
+	AuthMethodPrefix = "amoidc"
+	// AccountPrefix defines the prefix for Account public ids.
+	AccountPrefix = "acctoidc"
```
talanknight, line 12: password accounts are `apw`.  This should match by using `aoidc`.

internal/auth/oidc/auth_method.go
talanknight: Now I see you also documented the option itself.  Consider just referencing the Option's godoc.

talanknight (changes requested):
Got some of the way through and wanted to send some comments out.  I'll continue tonight or tomorrow.

internal/auth/oidc/client_secret.go
jimlambrt: We currently don't but if someone adds a future log msg that does, this will redact the secret appropriately.

internal/auth/oidc/account.go
```diff
@@ -0,0 +1,131 @@
...
+// SetTableName sets the table name.
+func (a *Account) SetTableName(n string) {
+	a.tableName = n
+}
+
+// oplog will create oplog metadata for the Account.
+func (c *Account) oplog(op oplog.OpType, authMethodScopeId string) oplog.Metadata {
+	metadata := oplog.Metadata{
+		"resource-public-id": []string{c.AuthMethodId}, // the auth method is the root aggregate
+		"resource-type":      []string{"oidc auth aud claim"},
```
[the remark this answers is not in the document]
jimlambrt: ty!

internal/auth/oidc/callback_url.go
jimlambrt: After an earlier discussion with @mgaffney, I believe we needed to support setting the domain portion for the callback and that the pathing would be statically set by Boundary.

internal/auth/oidc/ids.go
jimlambrt: I actually debated this and I think a single `a` is a less than ideal prefix for a prefix, so I went with the less ambiguous `acct` for oidc.  We can't change password accounts, but it's not a pattern we have to go forward with.

internal/auth/oidc/auth_method.go
jimlambrt: it's never going to change, so there's no synchronization problem.   max age is documented in the sql, proto, auth method, and option.

internal/auth/oidc/callback_url.go
talanknight: How would the API look? I think the answer to that determines what this method looks like.

Would it allow the full url to be provided and just error if the suffix was wrong?  Would it accept just the part before `/v1/auth-methods/amoidc_123` but append that before storage?  Would it accept/store/return only the first part and maybe have a different readonly field which appends the `/v1/auth-methods/amoidc_123` portion to what is entered?

internal/auth/oidc/ids.go
talanknight: We could also update the password account prefix then and support both prefixes (`apw` and `acctpw`) for the pw accounts.  I think there is benefit in consistency here. @jefferai Do you have thoughts?

internal/auth/oidc/callback_url.go
jimlambrt: I sorted of "punted" on what portion would be stored in the db.  I know for certain that it must be a valid URL, so that's what I've implemented so far.  I'm not sure the repo should enforce what portion is stored vs what is "read-only" and appended for display.  The service layer would seem to own that responsibility in our implementation.   How would the repo know what the API pathing is?

talanknight: I don't think the repo should know what the API pathing is.  The name `NewCallbackUrl` to me suggested that this is the complete URL used for the callback, not that it is just a part of it and the documentation doesn't say either way as well.

internal/auth/oidc/signing_alg_test.go
```diff
@@ -0,0 +1,268 @@
...
+			name: "empty-auth-method",
+			args: args{
+				authMethodId: "",
+				alg:          RS256,
+			},
+			wantErr:   true,
+			wantIsErr: errors.InvalidParameter,
+		},
+		{
+			name: "empty-alg",
```
talanknight, line 83: Add a test for unsupported algorithm.

internal/auth/oidc/auth_method.go
```diff
@@ -0,0 +1,193 @@
...
+//
+// ClientId equals an OAuth 2.0 Client Identifier valid at the Authorization
+// Server.
+//
+// ClientSecret equals the client's secret which will be encrypted when stored
+// in the database and an hmac representation will also be stored when every the
+// secret changes.  The secret is not returned via the API, the hmac is returned
+// so callers can determine if it's been updated.
+//
+// MaxAge equals the Maximum Authentication Age. Specifies the allowable elapsed
```
talanknight, line 54: What does it mean for the auth method's MaxAge to "force the IpD to re-authenticate the End-User"?  Does that mean the auth token is never generated or is generated but already expired?  Maybe this would be useful to test out the auth flow... but is there another time when that would be wanted?

talanknight (commented):
Looks good! I didn't review the KMS or Derivation utilities at the bottom closely since I don't know those very well.

internal/auth/oidc/callback_url.go
jimlambrt: It may very well be the whole path... that's not been decided.  Once we decide what portion will be stored, I can update the godocs and add validation in the basic CallbackUrl type.

internal/auth/oidc/signing_alg_test.go
jimlambrt: ty!

internal/auth/oidc/auth_method.go
jimlambrt: In short, the IdP must re-authenticate the user before issuing an auth token.  The token issued with have a "normal" TTL for the RP (defined by the IdP).  The issued auth token will have an `auth_time` claim, so the RP can verify the last time the end-user was authenticated. 

OIDC spec definition: _OPTIONAL: Maximum Authentication Age. Specifies the allowable elapsed time in seconds since the last time the end-user was actively authenticated by the OP. If the elapsed time is greater than this value, the OP must attempt to actively re-authenticate the end-user. (The max_age request parameter corresponds to the OpenID 2.0 PAPE max_auth_age request parameter.) When max_age is used, the ID token returned must include an auth_time claim value._

jimlambrt: By supporting `max_age`, we allow admins to support two things.

1. To enforce a minimum session freshness: If an app has a requirement that users must re-authenticate once per day, this can be enforced in the context of a much longer SSO session by providing max_age with a value. These are defined in seconds.
2. To force an immediate re-authentication: If an app requires that a user re-authenticate prior to access.

internal/auth/oidc/auth_method.go
```diff
@@ -0,0 +1,193 @@
...
+// DiscoveryUrl equals a URL where the OIDC provider's configuration can be
+// retrieved. The OIDC Discovery Specification requires the URL end in
+// /.well-known/openid-configuration. However, Boundary will strip the URI and
+// only store the scheme and domain.
+//
+// ClientId equals an OAuth 2.0 Client Identifier valid at the Authorization
+// Server.
+//
+// ClientSecret equals the client's secret which will be encrypted when stored
+// in the database and an hmac representation will also be stored when every the
```
mgaffney, line 50:
```suggestion
// in the database and an hmac representation will also be stored when ever the
```

internal/auth/oidc/auth_method.go
```diff
@@ -0,0 +1,193 @@
...
+// other options are ignored.
+//
+// State equals the state of the OIDC auth method.  State is not a supported
+// parameter when creating new AuthMethod's since it must be Inactive for all
+// new AuthMethods.
+//
+// DiscoveryUrl equals a URL where the OIDC provider's configuration can be
+// retrieved. The OIDC Discovery Specification requires the URL end in
+// /.well-known/openid-configuration. However, Boundary will strip the URI and
+// only store the scheme and domain.
```
mgaffney, line 44: The spec only says the URL must **end** in `/.well-known/openid-configuration`. If we remove everything after the scheme and domain, won't that break all of these URLs:
- `https://bigauth.com/foo/bar/baz/.well-known/openid-configuration`
- `https://bigauth.com/bar/baz/foo/.well-known/openid-configuration`
- `https://bigauth.com/baz/foo/bar/.well-known/openid-configuration`

internal/auth/oidc/auth_method.go
```diff
@@ -0,0 +1,193 @@
...
+// ClientSecret equals the client's secret which will be encrypted when stored
+// in the database and an hmac representation will also be stored when every the
+// secret changes.  The secret is not returned via the API, the hmac is returned
+// so callers can determine if it's been updated.
+//
+// MaxAge equals the Maximum Authentication Age. Specifies the allowable elapsed
+// time in seconds since the last time the End-User was actively authenticated
+// by the OP. If the elapsed time is greater than this value, the OP MUST
+// attempt to actively re-authenticate the End-User. A value -1 basically
+// forces the IdP to re-authenticate the End-User.  Zero is not a valid value.
```
mgaffney, line 58: Where do we protect against a zero being set?

internal/auth/oidc/auth_method.go
```diff
@@ -0,0 +1,193 @@
...
+			ClientSecret: string(clientSecret),
+			MaxAge:       int32(opts.withMaxAge),
+		},
+	}
+	if err := a.validate(op); err != nil {
+		return nil, err // intentionally not wrapped.
+	}
+	if a.ClientSecretHmac != "" {
+		return nil, errors.New(errors.InvalidParameter, op, "client secret hmac should be empty")
+	}
```
mgaffney, line 88: Is this necessary since `ClientSecretHmac` is never set in the above code?

internal/auth/oidc/auth_method.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+// encrypt the auth method before writing it to the db
+func (a *AuthMethod) encrypt(ctx context.Context, cipher wrapping.Wrapper) error {
+	const op = "oidc.(AuthMethod).encrypt"
+	if err := structwrapping.WrapStruct(ctx, cipher, a.AuthMethod, nil); err != nil {
+		return errors.Wrap(err, op, errors.WithCode(errors.Encrypt))
+	}
+	a.KeyId = cipher.KeyID()
+	if err := a.hmacClientSecret(cipher); err != nil {
```
mgaffney, line 163: What will happen when the client secret is an empty string?

mgaffney (changes requested):
I haven't finished reviewing but I left comments for parts I have reviewed.

internal/db/schema/migrations/postgres/2/84_oidc.up.sql
```diff
@@ -16,18 +16,23 @@ create table auth_oidc_method (
...
     constraint client_id_not_empty
     check(length(trim(client_id)) > 0), 
-  client_secret bytea not null, -- encrypted oidc client secret issued by the oidc provider.
+  client_secret bytea, -- encrypted oidc client secret issued by the oidc provider.
+  client_secret_hmac text 
+    constraint client_secret_hmac_not_empty
+    check(length(trim(client_secret_hmac)) > 0),
   key_id wt_private_id not null -- key used to encrypt entries via wrapping wrapper. 
-    references kms_oidc_key_version(private_id) 
+    references kms_database_key_version(private_id)
```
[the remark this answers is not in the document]
jimlambrt: Nope.  The OIDC DEK isn't for database encryption/decryption.   We're deriving a key from the OIDC DEK, and using the derived key for things that are being returned by the API like: the client secret HMAC, the token_request_id and the OIDC AuthURL.

internal/auth/oidc/account.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/hashicorp/boundary/internal/auth/oidc/store"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/oplog"
+	"google.golang.org/protobuf/proto"
+)
+
+// DefaultAccountTableName defines the default table name for an Account
+const DefaultAccountTableName = "auth_oidc_account"
+
+// Account contains an OIDC auth method configuration. It is assigned to an OIDC
```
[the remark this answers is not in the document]
jimlambrt: chuckle.  ty!

internal/auth/oidc/account.go
```diff
@@ -0,0 +1,131 @@
...
+	"net/url"
+
+	"github.com/hashicorp/boundary/internal/auth/oidc/store"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/oplog"
+	"google.golang.org/protobuf/proto"
+)
+
+// DefaultAccountTableName defines the default table name for an Account
+const DefaultAccountTableName = "auth_oidc_account"
```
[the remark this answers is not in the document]
jimlambrt: ty!

internal/auth/oidc/auth_method.go
```diff
@@ -0,0 +1,193 @@
...
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/kms"
+	"github.com/hashicorp/boundary/internal/oplog"
+	wrapping "github.com/hashicorp/go-kms-wrapping"
+	"github.com/hashicorp/go-kms-wrapping/structwrapping"
+	"google.golang.org/protobuf/proto"
+)
+
+// DefaultAuthMethodTableName defines the default table name for an AuthMethod
+const DefaultAuthMethodTableName = "auth_oidc_method"
```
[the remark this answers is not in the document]
jimlambrt: ty!

internal/auth/oidc/auth_method.go
jimlambrt: There's a constraint on the column, but I'll add it to the `validate` function.  ty!

jimlambrt: I added a comment on about this for the validate function.

internal/auth/oidc/auth_method.go
jimlambrt: Depends... I know it should never be set.  If a future dev doesn't understand that and sets it, then there will be a defect.  If I add this here, then future dev (aka. Me in a few months) will surely hesitate before setting it above.

internal/auth/oidc/auth_method.go
jimlambrt: An hmac of the empty string will be generated and callers can use it to determine if the secret has changed in the future.

internal/auth/oidc/auth_method.go
jimlambrt: ty!

internal/auth/oidc/aud_claim.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	"github.com/hashicorp/boundary/internal/auth/oidc/store"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/oplog"
+	"google.golang.org/protobuf/proto"
+)
+
+// DefaultAudClaimTableName defines the default table name for an AudClaim
+const DefaultAudClaimTableName = "auth_oidc_aud_claim"
```
mgaffney, line 11: Does this need to be exported?

internal/auth/oidc/callback_url.go
```diff
@@ -0,0 +1,98 @@
...
+	"net/url"
+
+	"github.com/hashicorp/boundary/internal/auth/oidc/store"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/oplog"
+	"google.golang.org/protobuf/proto"
+)
+
+// DefaultCallbackUrlTableName defines the default table name for a CallbackUrl
+const DefaultCallbackUrlTableName = "auth_oidc_callback_url"
```
mgaffney, line 13: Does this need to be exported?

internal/auth/oidc/certificate.go
```diff
@@ -0,0 +1,101 @@
...
+	"fmt"
+
+	"github.com/hashicorp/boundary/internal/auth/oidc/store"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/oplog"
+	"google.golang.org/protobuf/proto"
+)
+
+// DefaultCertificateTableName defines the default table name for a certificate
+const DefaultCertificateTableName = "auth_oidc_certificate"
```
mgaffney, line 15: Does this need to be exported?

internal/auth/oidc/signing_alg.go
```diff
@@ -0,0 +1,121 @@
...
+	ES384: true,
+	ES512: true,
+	PS256: true,
+	PS384: true,
+	PS512: true,
+	EdDSA: true,
+}
+
+// DefaultSigningAlgTableName defines the default table name for a SigningAlg
+const DefaultSigningAlgTableName = "auth_oidc_signing_alg"
```
mgaffney, line 45: Does this need to be exported?

mgaffney (changes requested):
Looks good. Some minor comments about exported constants.

internal/auth/oidc/auth_method.go
```diff
@@ -0,0 +1,199 @@
...
+// time in seconds since the last time the End-User was actively authenticated
+// by the OP. If the elapsed time is greater than this value, the OP MUST
+// attempt to actively re-authenticate the End-User. A value -1 basically
+// forces the IdP to re-authenticate the End-User.  Zero is not a valid value.
+//
+// See: https://openid.net/specs/openid-connect-core-1_0.html
+func NewAuthMethod(scopeId string, discoveryUrl *url.URL, clientId string, clientSecret ClientSecret, opt ...Option) (*AuthMethod, error) {
+	const op = "oidc.NewAuthMethod"
+
+	var u string
```
s-christoff, line 65: Could this var name be updated so the code that creates it doesn't have to be read? 
On the other hand, you only have to look down a couple lines and figure it out. 🤷🏻  I dunno

internal/kms/kms.go
```diff
@@ -352,3 +355,38 @@ func (k *Kms) loadDek(ctx context.Context, scopeId string, purpose KeyPurpose, r
 
 	return multi, nil
 }
+
+// DerivedReaderReader returns a reader from which keys can be read, using the
```
s-christoff, line 359: `NewDerivedReader`

internal/kms/kms.go
```diff
@@ -352,3 +355,38 @@ func (k *Kms) loadDek(ctx context.Context, scopeId string, purpose KeyPurpose, r
...
+	case *multiwrapper.MultiWrapper:
+		raw := w.WrapperForKeyID("__base__")
+		var ok bool
+		if aeadWrapper, ok = raw.(*aead.Wrapper); !ok {
+			return nil, errors.New(errors.InvalidParameter, op, "unexpected wrapper type from multiwrapper base")
+		}
+	case *aead.Wrapper:
+		aeadWrapper = w
+	default:
+		return nil, errors.New(errors.InvalidParameter, op, "unknown wrapper type")
```
s-christoff, line 385: Should this return an encrypt error?

internal/auth/oidc/aud_claim.go
jimlambrt: ty

internal/auth/oidc/callback_url.go
jimlambrt: ty!

internal/auth/oidc/certificate.go
jimlambrt: ty!

internal/auth/oidc/signing_alg.go
jimlambrt: ty!

s-christoff (commented):
None of these are major - just minor nits I saw whilst reading. Great job :)

internal/auth/oidc/account.go
```diff
@@ -0,0 +1,131 @@
...
+type Account struct {
+	*store.Account
+	tableName string
+}
+
+// NewAccount creates a new in memory Account assigned to OIDC AuthMethod.
+// WithFullName, WithEmail, WithName and WithDescription are the only valid
+// options. All other options are ignored.
+//
+// Issuer_id equals the Verifiable Identifier for an Issuer. An Issuer
```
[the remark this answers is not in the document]
jimlambrt: ty!

internal/auth/oidc/account.go
```diff
@@ -0,0 +1,131 @@
...
+// Email equals the End-User's preferred e-mail address. Its value MUST conform
+// to the RFC 5322 [RFC5322] addr-spec syntax. The RP MUST NOT rely upon this
+// value being unique
+//
+// See: https://openid.net/specs/openid-connect-core-1_0.html
+func NewAccount(authMethodId string, issuerId *url.URL, subjectId string, opt ...Option) (*Account, error) {
+	const op = "oidc.NewAccount"
+
+	if issuerId == nil {
+		return nil, errors.New(errors.InvalidParameter, op, "nil issuer id")
```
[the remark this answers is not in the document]
jimlambrt: ty!

internal/auth/oidc/aud_claim.go
```diff
@@ -0,0 +1,96 @@
...
+	return c, nil
+}
+
+// validate the AudClaim.  On success, it will return nil.
+func (a *AudClaim) validate(caller errors.Op) error {
+	if a.OidcMethodId == "" {
+		return errors.New(errors.InvalidParameter, caller, "missing oidc auth method id")
+	}
+	if a.Aud == "" {
+		return errors.New(errors.InvalidParameter, caller, "empty aud claim")
```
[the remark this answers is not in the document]
jimlambrt: because I'm not paying attention.  ty!

internal/kms/kms.go
jimlambrt: ty!

internal/kms/kms.go
jimlambrt: I think, this is an invalid parameter, since we don't even get to attempt encryption.

internal/auth/oidc/auth_method.go
jimlambrt: if they were separated a bit more in the code, then maybe.

internal/kms/kms.go
louisruch: Yes agreed this should be InvalidParam

mgaffney (approved):

s-christoff (approved):
🤙

talanknight (approved):

PR #898, opened 2021-02-01
Add initial filtering docs

jefferai:


website/content/docs/concepts/filtering/worker-tags.mdx
```diff
@@ -0,0 +1,87 @@
...
+# Worker Tags
+
+Starting in Boundary 0.1.5, a worker can be configured with a set of key/value
+tags in its configuration file. The keys and values can be any lower-cased
+printable value. Each key can have more than one value:
+
+```hcl
+worker {
+    name = "web-prod-us-east-1"
+	tags {
```
malnick, line 23: The whitespace is a little off here.

website/content/docs/concepts/filtering/worker-tags.mdx
```diff
@@ -0,0 +1,87 @@
...
+```
+
+Note that this is the canonical format, as it maps closely to the filter
+structure. However, for compatibility with some other systems, it is also
+possible to specify the tags in a pure key=value style:
+
+```hcl
+worker {
+    name = "web-prod-us-east-1"
+	tags = ["region=us-east-1", "type=prod", "type=webservers"]
```
malnick, line 51: Whitespace is off a bit here.

malnick (commented):
A couple whitespace comments and also wondering if we can update the configuration/worker section to include tagging?

malnick (approved):

PR #899, opened 2021-02-01
Add worker filter to create target tests

louisruch:
### What does this PR do

Adds a unit test to the create target test suite to include worker filter


louisruch:
@jefferai while trying to add worker filters to the TF provider I was getting an error.  The easiest way for me to figure out what I was doing wrong was via this unit test so I decided to just push it.

jefferai (approved):
Looks good, thanks!

PR #900, opened 2021-02-02
Don't error worker filters if a key isn't found, treat it as a non-match

jefferai:


website/content/docs/concepts/filtering/worker-tags.mdx
```diff
@@ -74,6 +74,13 @@ nested under `tags` and include a `name` value:
 }
 ```
 
+!> If an expression fails due to a key not being found within the input data,
```
louisruch, line 77: hmm not sure what `!>` is

louisruch (approved):
LGTM you just have a merge conflict in the changelog

website/content/docs/concepts/filtering/worker-tags.mdx
louisruch: NVM I found this https://github.com/hashicorp/remark-plugins/tree/main/plugins/paragraph-custom-alerts#paragraph-custom-alerts very cool

malnick (dismissed):

PR #901, opened 2021-02-02
Fix nil TLSClientConfig

kmaris:
The client from go-cleanhttp does not set TLSClientConfig. This can be a possible fix or go-cleanhttp can create the config at client creation.


jefferai (approved):
Thanks!

jefferai:
I'll circle back around to this later, need to ensure I get a changelog entry added, but looks good.

jefferai:
Merging, will add Changelog after.

Issue #903, opened 2021-02-02
Boundary 0.1.5 Docker Container Fails to Start


malnick:
In Boundary 0.1.5 we disallow setting the worker `controllers` configuration to all interfaces (0.0.0.0). The configuration in the docker container still has this set, and needs to be updated with the latest release of Boundary.

PR #906, opened 2021-02-03
update docker config for boundary to remove the controllers setting

malnick:
- Removes the `controllers` setting from the worker configuration because in docker we're running in single server mode
- Updates the bind address for all listeners to localhost

Fixes issue #903

Important caveats:
- When running in docker this requires `--hostname boundary` to be passed in on `docker run`

We will be adding the ability to use env:// in the listener address field in 0.1.6.

### DIFF
--- a/docker/config.hcl
+++ b/docker/config.hcl
@@ -5,34 +5,42 @@ controller {
   description = "A controller for a demo!"
 
   database {
+    # This configuration setting requires the user to execute the container with the URL as an env var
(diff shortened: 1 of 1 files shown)


jefferai:
Seems fine to me, although I haven't tested locally -- if it works, go for it!

jefferai (approved):
Seems fine to me, although I haven't tested locally -- if it works, go for it!

PR #907, opened 2021-02-04
Allow public_cluster_addr to use an env var or file

jefferai:
This makes Docker scenarios easier because it passes the hostname in as `HOSTNAME`, so it can be set to `env://HOSTNAME`


malnick (dismissed):
Woot!

malnick (dismissed):
🥳

malnick:
Fixes https://github.com/hashicorp/boundary/issues/903

PR #912, opened 2021-02-04
Add a duplicate authenticate:login API verb

jefferai:
In a release or two we'll deprecate the original `authenticate` verb.

Note @talanknight : This also requires a bump of grpc-gateway to 2.1.0 (and really pre-2.1.1) so it'd be good to take a look at whether anything there seems problematic.


internal/servers/controller/handlers/authmethods/authmethod_service.go
```diff
@@ -260,6 +262,28 @@ func (s Service) Authenticate(ctx context.Context, req *pbs.AuthenticateRequest)
 	return &pbs.AuthenticateResponse{Item: tok, TokenType: req.GetTokenType()}, nil
 }
 
+// AuthenticateLogin implements the interface pbs.AuthenticationServiceServer.
+func (s Service) AuthenticateLogin(ctx context.Context, req *pbs.AuthenticateLoginRequest) (*pbs.AuthenticateLoginResponse, error) {
+	if err := validateAuthenticateLoginRequest(req); err != nil {
+		return nil, err
+	}
+	authResults := s.authResult(ctx, req.GetAuthMethodId(), action.Authenticate)
```
talanknight, line 270: Should this be a subaction now or do we want to wait to add subactions when we expect people to actually want to grant permissions based on those subactions?

internal/servers/controller/handlers/authmethods/authmethod_service.go
```diff
@@ -239,6 +239,8 @@ func (s Service) DeleteAuthMethod(ctx context.Context, req *pbs.DeleteAuthMethod
 }
 
 // Authenticate implements the interface pbs.AuthenticationServiceServer.
+//
+// Deprecated in favor of AuthenticateLogin
 func (s Service) Authenticate(ctx context.Context, req *pbs.AuthenticateRequest) (*pbs.AuthenticateResponse, error) {
```
talanknight, line 244: Instead of duplicating the logic consider replacing the body with:
```
r, err := s.AuthenticateLogin(ctx, &pbs.AuthenticateLoginRequest{req.blabla})
// checks, etc...
return &pbs.AuthenticateResponse{r.blablabla}, err
```
Or since the AuthenticateRequest/Response is JSON equivalent to AuthenticateLoginRequest/Response consider replacing the former with the latter in the proto (and thus here as well so it can just be a passthrough)

internal/servers/controller/handlers/authmethods/authmethod_service.go
```diff
@@ -614,3 +639,31 @@ func validateAuthenticateRequest(req *pbs.AuthenticateRequest) error {
...
 }
+
+func validateAuthenticateLoginRequest(req *pbs.AuthenticateLoginRequest) error {
+	badFields := make(map[string]string)
+	if strings.TrimSpace(req.GetAuthMethodId()) == "" {
+		badFields["auth_method_id"] = "This is a required field."
+	} else if !handlers.ValidId(password.AuthMethodPrefix, req.GetAuthMethodId()) {
+		badFields["auth_method_id"] = "Invalid formatted identifier."
+	}
+	// TODO: Update this when we enable different auth method types.
```
talanknight, line 650: If each auth method type uses a different subaction this TODO is no longer relevant so you can remove it.

talanknight (changes requested):
Also update outgoing_interceptor.go please.

internal/servers/controller/handlers/authmethods/authmethod_service.go
jefferai: We don't currently plant to make these separately permissioned subactions.

internal/servers/controller/handlers/authmethods/authmethod_service.go
jefferai: If you re-use a message in the proto it won't compile, that's why there are different messages to begin with.

jefferai: I'm going to leave as-is as we're going to be removing the original function in a couple of releases. Not worth combining them to reduce duplication and then having to pull it apart again.

internal/servers/controller/handlers/authmethods/authmethod_service.go
jefferai: Done.

jefferai:
I've updated outgoing_interceptor.go

internal/api/genapi/templates.go
```diff
@@ -68,6 +69,7 @@ func fillTemplates() {
 			TypeOnCreate:        in.typeOnCreate,
 			CreateResponseTypes: in.createResponseTypes,
 			RecursiveListing:    in.recursiveListing,
+			SubAction:           in.subAction,
```
talanknight, line 72: Are we sure we want to approach sub actions this way (through setting subactions as a string)?  I personally would take the approach of making the pathing changes in this PR be transparent to the API to buy us some time to think about and discuss the best way to support sub actions once that becomes necessary in the API (maybe when we add OIDC).

talanknight (approved):
Besides the concern about updating the API before we need to, also please update handler_test.go which ensures each path is registered.  This is probably a good idea given the recent changes to the pathing logic in grpc-gateway for multi-":" verbs.

internal/api/genapi/templates.go
jefferai: I'm happy to have a conversation about it, but given we already know what OIDC will look like, I think we can have that conversation with the team now, rather than later.

jefferai: Per discussion, removed the API changes

jefferai:
@talanknight Tests all pass, you good with this now?

talanknight (approved):

PR #915, opened 2021-02-05
Remove list:self from main

jefferai:
As per futher discussions, we won't be supporting this, in favor of
simply read:self and cancel:self (since listing now already filters out
items on which you have no permissions).

This doesn't remove list:self entirely from the code base as much of
that will simply be converted to read:self and cancel:self -- that will
be in a later PR. However, with this removed if we want to release we
won't expose an API publicly that isn't going to stick around.


jefferai:
Will merge if tests are happy, as this simply removes code without changing behavior.

jefferai:
Final test issue was fixed by #917 so merging now that that's in.

PR #918, opened 2021-02-05
Release the shared lock using a non cancelled context

### LINKED ISSUE #909: Controller shutdown error: Unable to release shared lock to the database
**Describe the bug**
When shutting down a controller an error is returned "Unable to release shared lock to the database: schema.(Manager).SharedUnlock: postgres.(Postgres).UnlockShared: unknown, unknown: error #0: driver: bad connection"

**To Reproduce**
Steps to reproduce the behavior, In boundary version 0.1.5:
1. Make sure that config.hcl has a `controller` section with a database url specified.
1. Run `boundary database init -config config.hcl`
1. Run `boundary server -config config.hcl`
1. Wait for controller to finish starting and hit ctrl-C
1. See error "Unable to release shared lock to the database: schema.(Manager).SharedUnlock: postgres.(Postgres).UnlockShared: unknown, unknown: error #0: driver: bad connection"

**Expected behavior**
No error returned

**Additional context**
The error indicates that the connection to the database was not in a state that allowed releasing the lock.  Since this is a postgres advisory lock it should be getting released when the connection to the database is closed.  If this happens while the controller is still running it could make it possible for `database migrate` to happen while the controller is still running.

talanknight:
Release the controller's shared advisory lock using a non cancelled context.
Also releases the context resources by always calling cancel in a few places.

Fixes #909


jefferai (approved):
Looks good, nice catch on those other places!

PR #919, opened 2021-02-05
Ensure the controller maintains the shared lock

talanknight:
This ensures that the DB connection remains open between the schema manager and the database.  Since this lock is used to ensure the migrate and init commands don't run at the same time as the controller we don't want the connection to silently fail.

### DIFF
--- a/internal/cmd/commands/server/server.go
+++ b/internal/cmd/commands/server/server.go
@@ -36,9 +36,9 @@ var (
 type Command struct {
 	*base.Server
 
-	SighupCh      chan struct{}
-	ReloadedCh    chan struct{}
-	SigUSR2Ch     chan struct{}
+	SighupCh   chan struct{}
+	ReloadedCh chan struct{}
+	SigUSR2Ch  chan struct{}
 
 	Config     *config.Config
 	controller *controller.Controller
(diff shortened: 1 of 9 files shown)


internal/cmd/commands/server/server.go
```diff
@@ -653,6 +613,97 @@ func (c *Command) Reload() error {
...
+				c.UI.Error("The Schema Manager lost connection with the DB and cannot ensure it's integrity.")
+				c.CancelFn()
+				return
+			}
+
+			if err := c.SchemaManager.SharedLock(c.Context); err == nil {
+				continue
+			}
+
+			c.UI.Error("The Schema Manager lost connection with the DB and cannot ensure it's integrity.")
```
jefferai, line 692:
```suggestion
			c.UI.Error("The Schema Manager lost connection with the DB and cannot ensure its integrity.")
```

internal/cmd/commands/server/server.go
```diff
@@ -124,6 +126,7 @@ func (c *Command) AutocompleteFlags() complete.Flags {
 }
 
 func (c *Command) Run(args []string) int {
+	defer c.CancelFn()
```
jefferai, line 129: Note that this might be racy. If it's important that we ensure that any cleanup happens before the process exits, then we need to wait for the cleanup to be verified finished before exiting. Based on the comment below it sounds like this is best-effort anyways, but want to raise the issue just in case.

jefferai (commented):
Leaving a neutral review as I want a more experienced DB person checking the DB side but I think it looks good -- see my one comment regarding cleanup ordering, although I believe from your comments that it likely doesn't matter.

talanknight:
I'm going to close this while we figure a more generally applicable solution for doing distributed R/W mutexes.

PR #920, opened 2021-02-08
a small refactor, so we can easily support options when creating new auth tokens.

jimlambrt:


internal/authtoken/authtoken.go
```diff
@@ -97,14 +105,20 @@ func newAuthTokenId() (string, error) {
...
 	token, err := base62.Random(tokenLength)
 	if err != nil {
-		return "", errors.Wrap(err, op, errors.WithCode(errors.Io))
+		return nil, errors.Wrap(err, op, errors.WithCode(errors.Io))
 	}
-	return fmt.Sprintf("%s%s", TokenValueVersionPrefix, token), nil
+
+	return &AuthToken{
+		AuthToken: &store.AuthToken{
+			Token: token,
```
talanknight, line 119: Is there a reason why you removed the auth token version prefix?

talanknight (commented):
Looks great!

internal/authtoken/authtoken.go
jimlambrt: That was just an oversight during the refactor.  Ty for catching this!

talanknight (approved):

PR #923, opened 2021-02-10
Add read:self and cancel:self to sessions; add dev unpriv user

jefferai:
This adds:

* Sessions support for read:self and cancel:self permissions. This is automatic with fallback; no API change is needed.
* Strips out values from listing if only self actions are allowed on the resource and the user ID is different. This prevents a default ACL of `id=*;type=session;actions=read:self,cancel:self` from causing all values to be visible when listing.
* Adds an unprivileged user to development/testing instances. This user has only default permissions plus the ability to authorize sessions in the default target.


jimlambrt:
Just an observation that iam.CreateScope(...) has sort of gotten a bit long in the tooth.  Perhaps, we should break it up a bit in a future PR, which might help it's readability.

internal/cmd/base/initial_resources.go
```diff
@@ -125,78 +125,110 @@ func (b *Server) CreateInitialAuthMethod(ctx context.Context) (*password.AuthMet
 	b.InfoKeys = append(b.InfoKeys, "generated auth method id")
 	b.Info["generated auth method id"] = b.DevAuthMethodId
 
-	// Create the dev user
-	if b.DevLoginName == "" {
-		b.DevLoginName, err = base62.Random(10)
+	createUser := func(loginName, loginPassword, userId string, admin bool) (*iam.User, error) {
```
jimlambrt, line 128: that's a really long closure... easy to read though.

jimlambrt (approved):
Looks great.

internal/cmd/base/initial_resources.go
jefferai: I agree, it's long, but it does a related chunk of work...I found myself duplicating all of that, which meant that the unit of work was...all of that. :-)

jefferai:
* Sessions support for read:self and cancel:self permissions. This is automatic with fallback; no API change is needed.

* Strips out values from listing if only self actions are allowed on the resource and the user ID is different. This prevents a default ACL of id=*;type=session;actions=read:self,cancel:self from causing all values to be visible when listing.

* Adds an unprivileged user to development/testing instances. This user has only default permissions plus the ability to authorize sessions in the default target.

PR #925, opened 2021-02-11
repo signatures for OIDC

jimlambrt:


internal/auth/oidc/repository_auth_method.go
```diff
@@ -0,0 +1,147 @@
...
+//
+// No options are currently supported.
+func (r *Repository) TestAuthMethod(ctx context.Context, m *AuthMethod, opt ...Option) error {
+	panic("to-do")
+}
+
+// MakeInactive will transision an OIDC auth method from either the
+// ActivePrivateState or the ActivePublicState into the temporary StoppingState
+// and then, after a small amount of time, to the InactiveState.
+func (r *Repository) MakeInactive(ctx context.Context, authMethodId string, _ ...Option) error {
```
mgaffney, line 73: Should these methods return the state?

internal/auth/oidc/repository_auth_method.go
```diff
@@ -0,0 +1,147 @@
...
+	panic("to-do")
+}
+
+// StartAuth accepts a request to start an OIDC authentication/authorization
+// attempt containing the client_type, the client_version, and a map of client
+// round-tripper k/v pairs.  It returns two URLs.  authUrl is an OIDC
+// authorization request URL.  tokenUrl is the URL the client can use to retrieve
+// the results of the user's OIDC authentication attempt.  No options are
+// currently supported.
+func (r *Repository) StartAuth(ctx context.Context, authMethodId string, clientType, clientVersion string, clientRoundTripKVs map[string]string) (authUrl *url.URL, tokenUrl *url.URL, e error) {
```
mgaffney, line 101: I think the `clientType`, `clientVersion`, and `clientRoundTripKVs` parameters should be combined into a single struct since they are related. Same thing with the `authUrl` and `tokenUrl` return values since they are tightly related.

internal/auth/oidc/repository_auth_method.go
```diff
@@ -0,0 +1,147 @@
...
+// final_redirect_url.
+//
+// * Exchange the callbackCodeParameter for provider tokens and validate the
+// tokens.  Call UserInfo endpoint using access token.
+//
+// * Create/update account using "sub" claim as external ID and setting email
+// and full name for the account.
+//
+// * Return the Account and final redirect URL.
+func (r *Repository) ExchangeAuthResponse(ctx context.Context, authMethodId string, callbackStateParameter string, callbackCodeParameter string) (*Account, string, error) {
```
mgaffney, line 133: `callbackStateParameter` and `callbackCodeParameter` seem overly verbose

internal/auth/oidc/repository_auth_method.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// DecryptTokenRequestId will decrypt the tokenRequestId. The tokenRequestId was
+// created and encrypted by the repo's StartAuth(...) function.  If decryption
+// fails, an empty string and error is returned.
+//
+// This function is typically when the auth method service is handling
+// "authenticate:token".  If decryption is successful, then the service would
+// retrieve the pending token associated with the request id, update the token's
+// status to "issued" and return it to the caller.
```
mgaffney, line 144: Successfully decrypting the token does not indicate the user has successfully been authenticated. If the token can be decrypted, the repository should retrieve the current status of the token. A new token should only be issued if the status is 'auth token pending' and the request has not expired.

internal/authtoken/repository.go
```diff
@@ -337,3 +338,19 @@ func (r *Repository) DeleteAuthToken(ctx context.Context, id string, opt ...Opti
...
+// options are ignored.
+func (r *Repository) CreatePendingAuthToken(ctx context.Context, tokenRequestId string, iamUser *iam.User, withAuthAccountId string, opt ...Option) error {
+	panic("to-do")
+}
+
+// IssueAuthToken will retrieve the "pending" token and update it's status to
+// "issued".  If the token has already been issued, an error is returned with a
+// nil token.  If no token is found for the tokenRequestId an error is returned
+// with a nil token.
+func (r *Repository) IssueAuthToken(ctx context.Context, tokenRequestId string) (*AuthToken, error) {
```
mgaffney, line 354: Can we discuss the flow of these methods?

mgaffney (changes requested):

internal/auth/oidc/repository.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+// Repository is the oidc repository
+type Repository struct {
+	reader db.Reader
+	writer db.Writer
+	kms    *kms.Kms
+
+	providers  map[string]*oidc.Provider
+	providerMu *sync.Mutex
```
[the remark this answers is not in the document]
jimlambrt: Ty!

internal/auth/oidc/repository.go
```diff
@@ -0,0 +1,70 @@
...
+	r.providerMu.Lock()
+	defer r.providerMu.Unlock()
+	r.providers[authMethodId] = p
+}
+
+func (r *Repository) delProvider(ctx context.Context, authMethodId string, p *oidc.Provider) {
+	r.providerMu.Lock()
+	defer r.providerMu.Unlock()
+	delete(r.providers, authMethodId)
+}
```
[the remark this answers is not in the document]
jimlambrt: yep.  I had already done this work on the `jimlambrt-oidc-wip` branch.

ty!

internal/auth/oidc/repository.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/hashicorp/cap/oidc"
+)
+
+// Repository is the oidc repository
+type Repository struct {
+	reader db.Reader
+	writer db.Writer
+	kms    *kms.Kms
+
+	providers  map[string]*oidc.Provider
```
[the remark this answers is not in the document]
jimlambrt: ty!  already moved on the `jimlambrt-oidc-wip` branch.

internal/auth/oidc/repository_auth_method.go
```diff
@@ -0,0 +1,147 @@
...
+// If the AuthMethod contains a DiscoveryUrl for an OIDC provider, TestAuthMethod
+// retrieves the OpenID Configuration document. The values in the AuthMethod
+// (and associated data) are validated with the retrieved document. The issuer and
+// id token signing algorithm in the configuration are validated with the
+// retrieved document. TestAuthMethod also verifies the authorization, token,
+// and user_info endpoints by connecting to each and uses any certificates in the
+// configuration as trust anchors to confirm connectivity.
+//
+// No options are currently supported.
+func (r *Repository) TestAuthMethod(ctx context.Context, m *AuthMethod, opt ...Option) error {
```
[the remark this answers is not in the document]
jimlambrt: I thought we had discussed that basic types don't do out of process calls for validation.  If we still believe that's a valid pattern, then it can't really move to the basic AuthMethod type.

internal/auth/oidc/repository_auth_method.go
jimlambrt: In my mind, the domain function either succeeds or fails, so returning State isn't necessary.

internal/auth/oidc/repository_auth_method.go
jimlambrt: Agreed that they are related, but I'm not sure they should be combined into a struct unless there's some ambiguity across the existing function parameters.  Embedding them within a struct doesn't make the function signature more readable.  I might argue it makes it less readable, especially in the godocs.   All this applies to the return values as well.  There's no ambiguity about the purpose/intent of authUrl and tokenUrl.  Happy and interested to discuss more about this.

internal/auth/oidc/repository_auth_method.go
jimlambrt: Yeah, I'm not in love with the names as well.  Happy to hear suggestions!  I want to make sure there's no ambiguity about what's expected, without having to delve into the godocs if possible.

internal/auth/oidc/repository_auth_method.go
jimlambrt: Agreed.  This comment is not intended to be a specification for how to implement the service handler.  It's more of just a hint on how the function fits into the service handler's flow.

internal/authtoken/repository.go
jimlambrt: Happy to.

internal/auth/oidc/repository_auth_method.go
```diff
@@ -0,0 +1,147 @@
...
+func (r *Repository) LookupAuthMethod(ctx context.Context, publicId string, _ ...Option) (*AuthMethod, error) {
+	panic("to-do")
+}
+
+// ListAuthMethods returns a slice of AuthMethods for the scopeId. WithLimit is the only option supported.
+func (r *Repository) ListAuthMethods(ctx context.Context, scopeIds []string, opt ...Option) ([]*AuthMethod, error) {
+	panic("to-do")
+}
+
+// DeleteAuthMethod will delete the auth method from the repository.  No options
```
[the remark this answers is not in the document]
jimlambrt: This is true of any operation that writes to an OIDC auth method (update, delete, and explicit state transitions).  I have included this explicitly in the docs for the operations that explicitly change state, but not the rest of them.  I will be updating the godocs for all of the signatures related to side-affects and important operational details when they are implemented.

mgaffney (approved):

PR #926, opened 2021-02-11
Use uint for internal types

daneharrigan:
The internal types `action.Type`, `resource.Type`, and `scope.Type` all start at zero, but aren't the same integer type. This is to align them all on `uint`.


jefferai:
Thanks!

PR #927, opened 2021-02-11
Add testing section for individual/user testing 

s-christoff:
Howdy,
Some other Hashicorp products ([Waypoint](https://github.com/hashicorp/waypoint#running-the-unit-tests),[Consul](https://github.com/hashicorp/consul/blob/master/.github/CONTRIBUTING.md#testing),[Terraform](https://github.com/hashicorp/terraform/blob/master/.github/CONTRIBUTING.md#acceptance-tests-testing-interactions-with-external-services),[Nomad](https://github.com/hashicorp/nomad#testing),[Vault](https://github.com/hashicorp/vault#developing-vault), [Packer](https://github.com/hashicorp/packer/blob/master/.github/CONTRIBUTING.md#running-unit-tests),[Vagrant](https://github.com/hashicorp/vagrant/blob/main/.github/CONTRIBUTING.md#running-tests)) provide testing documentation in either their main readme or contributing documentation to help both operators and developers alike. We should do this too! 
I'm open to feedback on how to add a `make test` to the `Makefile` or something similar.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,6 +148,9 @@ website-start:
 test-ci: install-go
 	~/.go/bin/go test ./... -v $(TESTARGS) -timeout 120m
 
+test: 
+	~/.go/bin/go test ./... -timeout 120m
+
 install-go:
 	./ci/goinstall.sh
 
--- a/Makefile
+++ b/Makefile
@@ -149,7 +149,7 @@ test-ci: install-go
 	~/.go/bin/go test ./... -v $(TESTARGS) -timeout 120m
 
 test: 
-	~/.go/bin/go test ./... -timeout 120m
+	~/.go/bin/go test ./... -timeout 30m
 
 install-go:
 	./ci/goinstall.sh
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -148,6 +148,9 @@ website-start:
 test-ci: install-go
 	~/.go/bin/go test ./... -v $(TESTARGS) -timeout 120m
 
+test: 
+	~/.go/bin/go test ./... -timeout 120m
```
jimlambrt, line 152: that's a REALLY long timeout :)

jimlambrt (approved):
Looks great!  Ty!

Makefile
s-christoff: I said the same thing - lol, seems on avg in CI our tests go about 30minutes. I can get it to 45minutes and see what happens?

jimlambrt: I'm not sure I believe the CI numbers... it looks like they're done long before CI reports that they are done.   Can we try something like 30 min and see how it goes.   I really think, my average is more like 5 min when I run them without any caching.

PR #930, opened 2021-02-12
Add initial Desktop docs

malnick:


gsusmi:
Looks great @malnick! Just a note to make sure we don't forget - two images are missing: boundary-desktop-active.png & boundary-desktop-terminated.png

PR #933, opened 2021-02-12
fix 'make update-ui-version'

samsalisbury:
- When using a branch name rather than a commit, the branch was not being
  updated to the latest version.


mdeggies:
Thanks, Sam!

mdeggies (approved):

PR #935, opened 2021-02-15
support a testing postgres URL

jimlambrt:
Adding support for a testing postgres URL via an env var of BOUNDARY_TESTING_PG_URL would give developers the ability to run one unit test at time and examine the state of their test postgres instance.


mgaffney (dismissed):

talanknight (approved):

jefferai (changes requested):
I think in this case you would need to not set the `cleanup` func to destroying a Docker image. But you also don't want to set it too early. So you may need a bool to indicate if a container was created, and use that to gate setting the cleanup function on line 42 (50) below.

jimlambrt:
@jefferai 
won’t the return handle the cleanup properly?  https://github.com/hashicorp/boundary/blob/e0e5b2a0ebb1140879349d1eecb55c2901965caf/internal/docker/supported.go#L34

jefferai:
Lol I missed that, so sorry. Can you update the docker bit to "11" as discussed today, and then I think merge?

jefferai (approved):

PR #940, opened 2021-02-16
Fix migration for Postgres 11

jefferai:
This fixes an issue with running migrations (likely in 0.1.5 and 0.1.6) against Postgres 11.

It also changes our dev environment to be using the PG 11 Docker image rather than PG 12.


mgaffney (approved):

malnick (approved):
Changelog looks good!

jimlambrt (commented):
Looks good.

PR #948, opened 2021-02-18
internal/db/schema: fix dropped test error

alrs:
This fixes a dropped test error in `internal/db/schema`.


louisruch (approved):
LGTM thanks!

PR #952, opened 2021-02-19
Add ability to filter List*Requests

talanknight:
Adds a `filter` query param which must match the syntax expected by "https://github.com/hashicorp/go-bexpr".


internal/proto/local/controller/api/services/v1/account_service.proto
```diff
@@ -124,6 +124,7 @@ message GetAccountResponse {
 
 message ListAccountsRequest {
   string auth_method_id = 1 [json_name="auth_method_id"];
+  string filter = 2;
```
jefferai, line 127: Since filter is 30 in a lot of the files below, could we just standardize on it as being 30 everywhere? It's 1 here and 20 in a couple of places. Just seems nicer, although I know it doesn't matter in a technical sense.

internal/proto/local/controller/api/services/v1/session_service.proto
```diff
@@ -63,6 +63,7 @@ message GetSessionResponse {
 message ListSessionsRequest {
 	string scope_id = 1;
 	bool recursive = 20 [json_name="recursive"];
+	string filter = 30 [json_name="filter"];
```
jefferai, line 66: Is there a reason you put `json_name` here and not in other places?

louisruch (approved):
LGTM great work!

internal/servers/controller/handlers/sessions/session_service.go
```diff
@@ -120,19 +124,25 @@ func (s Service) ListSessions(ctx context.Context, req *pbs.ListSessionsRequest)
...
+		}
+		onlySelf := true
+		for _, v := range authorizedActions {
+			if v != action.ReadSelf && v != action.CancelSelf {
+				onlySelf = false
+				break
 			}
 		}
+		if onlySelf && item.GetUserId() == authResults.UserId {
+			continue
```
jefferai, line 138: I don't think this logic is right...if you're going to flip the logic from what it was before, you need `!=` here.

The item.AuthorizedActions line below should probably also go in the `if filter.Match` clause too.

talanknight: Woops, thanks for catching that! I'll update it any place i missed up the flipping of the logic.

Re: putting item.AuthorizedActions inside the filter.Match clause, if we perform this outside of the filter.Match clause we allow the filter to use the `authorized_actions` as part of the filter.

internal/proto/local/controller/api/services/v1/account_service.proto
talanknight: I've updated the protos to try to be consistent with making filter be 10s based field with a json_name option specified.  Like you said, it doesn't matter technically but consistency is important.

internal/proto/local/controller/api/services/v1/session_service.proto
talanknight: Nope.  I've added the option in all the .proto files.

internal/servers/controller/handlers/sessions/session_service.go
```diff
@@ -120,19 +124,25 @@ func (s Service) ListSessions(ctx context.Context, req *pbs.ListSessionsRequest)
...
+			if v != action.ReadSelf && v != action.CancelSelf {
+				onlySelf = false
+				break
 			}
 		}
+		if onlySelf && item.GetUserId() != authResults.UserId {
+			continue
+		}
+
+		item.AuthorizedActions = authorizedActions.Strings()
```
jefferai, line 141: Can you move this into the `if` below? Might as well not generate the strings if it's filtered.

jefferai (commented):
Looks generally good, pending truing up the proto numbers and fixing the tests around the self actions on session.

internal/servers/controller/handlers/sessions/session_service.go
jefferai: Discussed in Slack; it's here so it can be used in the filter

jefferai (approved):
Looks good!

PR #956, opened 2021-02-22
Dep updates (pgx, dktest)

jefferai:
* pgx
* dktest


talanknight (changes requested):
It doesn't look like pgx is needed anymore, as discussed in slack.  Please update internal/db/domains_test.go to remove the import.

talanknight (approved):

PR #963, opened 2021-02-25
Ongoing OIDC work:  repo method for updating an auth method

jimlambrt:
This PR adds:

1. repo.UpdateAuthMethod(...) (with supporting un-exported functions) and repo.ValidateAuthMethod(...) 

2. new options required for update functions: WithForce, WithDryRun, WithAuthMethod, WithPublicId

3. new VersionMismatch error code that represents a version mismatch between the persisted resource and version parameter of the update function

4. new un-exported test function testFreePort(...) which returns an available free localhost port

@jefferai , @s-christoff, @louisruch: I added you as optional reviewers


internal/auth/oidc/options.go
```diff
@@ -132,3 +136,34 @@ func WithUnauthenticatedUser(enabled bool) Option {
 		o.withUnauthenticatedUser = enabled
 	}
 }
+
+// WithForce provides an option to force the write operation, regardless of
+// whether or not it's pre-verification succeeds.
+func WithForce() Option {
```
jefferai, line 142: The new and improved Boundary...now with _force_

internal/auth/oidc/testing.go
```diff
@@ -237,3 +237,16 @@ func testProvider(t *testing.T, clientId, clientSecret, allowedRedirectURL strin
 	require.NoError(err)
 	return p1
 }
+
+// testFreePort just returns an available free localhost port
+func testFreePort(t *testing.T) int {
```
jefferai, line 242: This is actually a great helper function -- can you put it in `internal/testing` so it's more broadly available?

jimlambrt: sounds good.  I'll open a PR after this to move it under the sdk, and once it's merged, I'll open an oidc PR to start using it here and remove it from this package.

internal/auth/oidc/repository_auth_method_update.go
```diff
@@ -17,32 +31,561 @@ import "context"
...
 //
 // * WithForce: when this option is provided, the auth method is persistented in
 // the repository without testing it fo validity with Repository.TestAuthMethod.
 //
 // Successful updates must invalidate (delete) the Repository's cache of the
 // oidc.Provider for the AuthMethod.
-func (r *Repository) UpdateAuthMethod(ctx context.Context, m *AuthMethod, version uint32, fieldMaskPaths []string, _ ...Option) (*AuthMethod, error) {
+func (r *Repository) UpdateAuthMethod(ctx context.Context, am *AuthMethod, version uint32, fieldMaskPaths []string, opt ...Option) (*AuthMethod, int, error) {
 	const op = "oidc.(Repository).UpdateAuthMethod"
-	panic("to-do")
```
jefferai, line 30: Maybe leave this in?

internal/auth/oidc/repository_auth_method_update.go
```diff
@@ -17,32 +31,561 @@ import "context"
...
+		am = opts.withAuthMethod
+	default:
+		return errors.New(errors.InvalidParameter, op, "neither WithPublicId(...) nor WithAuthMethod(...) options were provided")
+	}
+
+	if err := am.isComplete(); err != nil {
+		return errors.Wrap(err, op)
+	}
+
+	// FYI: once converted to an oidc.Provider, any certs configured will be used as trust anchors for all HTTP requests
```
jefferai, line 516: Can you explain this? To me this sounds like you are using (and modifying) `http.DefaultClient`/`http.DefaultTransport`. If so -- don't do that. Use `go-cleanhttp` instead. If it just means "for the requests in this function" then that's fine.

internal/auth/oidc/repository_auth_method_update.go
```diff
@@ -17,32 +31,561 @@ import "context"
...
+	}
+
+	// we need to prevent redirects during these tests... we don't want to have
+	// redirects going to the controller's callback (aka the configured provider's callback)
+	providerClient.CheckRedirect = func(req *http.Request, via []*http.Request) error {
+		return http.ErrUseLastResponse
+	}
+
+	// test JWKs URL
+	statusCode, err := pingEndpoint(ctx, providerClient, "JWKs", "GET", info.JWKSURL)
```
jefferai, line 551: Any reason it's `JWKSURL` instead of `JwksUrl`? I'm guessing it's for legacy compatibility with Vault -- although that can easily be handled (and likely already is) with JSON tags -- but just checking.

internal/auth/oidc/repository_auth_method_update.go
```diff
@@ -17,32 +31,561 @@ import "context"
...
+	if err != nil {
+		result = multierror.Append(result, errors.New(errors.Unknown, op, fmt.Sprintf("unable to verify JWKs endpoint: %s", info.JWKSURL), errors.WithWrap(err)))
+	}
+	if statusCode != 200 {
+		result = multierror.Append(result, errors.New(errors.Unknown, op, fmt.Sprintf("non-200 status (%d) from JWKs endpoint: %s", statusCode, info.JWKSURL), errors.WithWrap(err)))
+	}
+
+	// test Auth URL
+	if _, err := pingEndpoint(ctx, providerClient, "AuthURL", "GET", info.AuthURL); err != nil {
```
jefferai, line 560: Same question around AuthUrl and TokenUrl.

internal/auth/oidc/repository_auth_method_update.go
```diff
@@ -17,32 +31,561 @@ import "context"
...
+
+	return result.ErrorOrNil()
+}
+
+type HTTPClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
+// pingEndpoint will make an attempted http request, return status code and errors
+// (non 2xx status codes are not an error)
```
jefferai, line 579: FWIW, my immediate read on this comment was "wouldn't 4xx be an error" but because actually having a comment about HTTP semantics made me think that the error for this function had to do with HTTP semantics. I'd just remove this line -- saying it returns the status code and method error seems fine. :-)

internal/auth/oidc/repository_auth_method_update.go
jimlambrt: `convertToProvider(...)` will convert the AuthMethod to a CAP `oidc.Provider`.  Internally, the cap provider uses go-cleanhttp to get a pooled transport and then if the provider was configured with certs, it creates a cert pool for the http.Client.  Here's the code in cap: https://github.com/hashicorp/cap/blob/4b74380f71bda210a64705240c21c6e309f17039/oidc/provider.go#L576

jimlambrt: All this results in `provider.DiscoveryInfo(ctx)` will use the certs configured for the auth method.

internal/auth/oidc/repository_auth_method_update.go
jimlambrt: the `info` struct is imported from CAP and it the initialisms standard from  https://github.com/golang/go/wiki/CodeReviewComments#initialisms

internal/auth/oidc/repository_auth_method_update.go
jimlambrt: same answer :)

internal/auth/oidc/repository_auth_method_update.go
jimlambrt: done.

internal/auth/oidc/repository_auth_method_update.go
jimlambrt: I panic about all the things these days.

internal/auth/oidc/repository_auth_method_update.go
```diff
@@ -17,32 +31,559 @@ import "context"
...
+		return errors.Wrap(err, op)
+	}
+
+	var result *multierror.Error
+	if info.Issuer != am.DiscoveryUrl {
+		result = multierror.Append(result, errors.New(errors.InvalidParameter, op,
+			fmt.Sprintf("auth method issuer doesn't match discovery issuer: expected %s and got %s", am.DiscoveryUrl, info.Issuer)))
+	}
+	for _, a := range am.SigningAlgs {
+		if !strutil.StrListContains(info.IdTokenSigningAlgsSupported, a) {
```
talanknight, line 532: The user must enter the correct capitalization for the signing algorithm to be used?

internal/auth/oidc/repository_auth_method_update.go
```diff
@@ -17,32 +31,559 @@ import "context"
...
+	if publicId == "" {
+		return nil, nil, errors.New(errors.InvalidParameter, op, "missing public id")
+	}
+	if !validVoName(valueObjectName) {
+		return nil, nil, errors.New(errors.InvalidParameter, op, fmt.Sprintf("invalid value object name: %s", valueObjectName))
+	}
+	if !strutil.StrListContains(dbMask, string(valueObjectName)) && !strutil.StrListContains(nullFields, string(valueObjectName)) {
+		return nil, nil, nil
+	}
+	if len(strutil.RemoveDuplicates(newVOs, false)) != len(newVOs) {
```
talanknight, line 341: Some value objects don't care about removing case insensitive duplicates while others do.  Should this function allow the caller to specify?

jimlambrt: All value objects (Callbacks, Certs, Algs, and Auds must be unique or the database will reject them based on the primary keys of their storage tables.  That's this function raises an error if they are not.

internal/auth/oidc/repository_auth_method_update.go
jimlambrt: Yes.  The spec defines the valid values: https://tools.ietf.org/html/rfc7518#section-3.1

internal/auth/oidc/repository_auth_method_update.go
```diff
@@ -1,14 +1,28 @@
...
+)
 
 // UpdateAuthMethod will retrieve the auth method from the repository,
 // update it based on the field masks provided, and then validate it using
 // Repository.TestAuthMethod(...).  If the test succeeds, the auth method
 // is persisted in the repository and the written auth method is returned.
 // fieldMaskPaths provides field_mask.proto paths for fields that should
 // be updated.  Fields will be set to NULL if the field is a
-// zero value and included in fieldMask. Name, Description, State, DiscoveryUrl,
+// zero value and included in fieldMask. Name, Description, OperationalState, DiscoveryUrl,
```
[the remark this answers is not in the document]
jimlambrt: line #100 calls ValidateAuthMethod(..) to the applied changes request in the update.  If that fails, an error is returned and the changes are not persisted. 

Changing other values does not have any side-affects on the Auth Method's state.  For instance, if the caller sets the SigningAlg to ES52 (which is not published in the method's discovery info) on an AuthMethod that's ActivePublic, and provides the WithForce() option, the update will update the DiscoveryUrl and have no affect on the state which will remain ActivePublic.    This is by design and intentional.  The API needs to support this use-case.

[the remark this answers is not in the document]
jimlambrt: We need to check the validity of an AuthMethod for several operations: Updates and OperationalState changes.  At this point, I'm not sure why we would force a caller to make two calls when they want to update an AuthMethod's values and then it's OperationalState, since we need to check the validity of the AuthMethod for both operations.

I do see some value in providing explicit OperationalState change endpoints... and I've implemented these narrowly defined update operations in the jimlambrt-oidc-wip branch for a the next PR.  They still need unit tests.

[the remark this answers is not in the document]
jimlambrt: Hum.. I just noticed that the RFC doesn't specify a way to transition to ActivePublic or ActivePrivate unless the Auth Method is successfully validated against the IdP's discovery document.  I think, that's a poor choice.  We should allow some sort of WithForce(...) option for these operations as well.  If the operator is sure the config will work and the IdP's discovery info is out of date, they have to have a way to enable the auth method for users.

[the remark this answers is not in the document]
jefferai: >There is custom actions dedicated to /v1/auth-methods/amoidc_1234567890:change-state:make-inactive, /v1/auth-methods/amoidc_1234567890:change-state:make-private, and /v1/auth-methods/amoidc_1234567890:change-state:make-public

I realize most people will use SDK or UI, but -- doesn't this seem overly wordy? Why not just `change-state` with the desired state as a parameter?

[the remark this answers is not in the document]
jimlambrt: @jefferai @talanknight - if there are no objections, I'm going to add a TODO to `UpdateAuthMethod` about the need to align on supporting the update of OperationalState or not, so I can proceed with a merge.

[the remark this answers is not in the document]
jefferai: Sure - my question isn't really about the code here, it's about the API aspects, so that's fine by me.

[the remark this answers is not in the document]
mgaffney: In the RFC, the `OperationalState` is calculated and set by the system not a user. This is a usability feature. It prevents end users from seeing or selecting an AuthMethod that is provably misconfigured which in turn protects an administrator from exposing an error and sparing them some potential embarrassment.

The calculation happens when any of the non authentication operations are called. The calculation uses the configuration of the AuthMethod and the operation invoked to determine the OperationalState. An AuthMethod that has a provably wrong configuration is always in the inactive state. 

I believe a configuration missing any of the following items would qualify as provably wrong:
- discovery url
- client id
- client secret
- callback url
- signing algorithm

internal/auth/oidc/repository_auth_method_update.go
```diff
@@ -17,32 +31,565 @@ import "context"
...
+
+	if err := validateFieldMask(fieldMaskPaths); err != nil {
+		return nil, db.NoRowsAffected, errors.Wrap(err, op)
+	}
+
+	dbMask, nullFields := dbcommon.BuildUpdatePaths(
+		map[string]interface{}{
+			"Name":             am.Name,
+			"Description":      am.Description,
+			"OperationalState": am.OperationalState,
```
mgaffney, line 63: This should be a calculated field (see my comment above).

mgaffney (changes requested):

internal/auth/oidc/repository_auth_method_update.go
jimlambrt: I have concerns/questions which we should discuss.  

With that said, I think the UpdateAuthMethod(...) follows the spirt of the RFC as I read it.  

Unless an admin uses the WithForce option, the update operation determines if the auth method is valid, via AuthMethod.IsComplete and Repository.ValidateAuthMethod. IsComplete ensures the required configuration fields are populated and ValidateAuthMethod checks the configuration against the IdP's published discovery info.   

This implementation would appear to support the usability feature you've cited, by preventing an administrator from an embarrassing outage by updating auth method fields without ample warning that the auth method may be incomplete or misconfigured.   

Additionally, the implementation allows for one transaction/operation which can safely update an incomplete configuration and set the operational state to whatever the admin desires.  If admins prefer, they can split this workflow into separate operations: update then a 2nd transaction to change the operational state.    Admins can choose the workflow they wish (aka whatever supports their needs the best) vs a single workflow imposed by us.   Why not support both workflows, as long as both workflows provide administrators ample notification of possible configuration issues before inadvertent changes  are visible to users of the auth method?

mgaffney: Sorry to disagree, but I don't think the UpdateAuthMethod(...) follows the spirit of the RFC, because one of the main goals I had when I wrote it was to prevent a provably misconfigured OIDC Auth Method from being exposed to end users.

What are the advantages of allowing end users to see and select an OIDC auth method that cannot work? What will be an end user's experience when 10 OIDC auth methods are listed but only 1 will work and they have no way of knowing which one? I understand the advantages of eliminating unnecessary steps in any workflow. However, I don't think this step is unnecessary since it prevents bad things from happening.

I'm happy to discuss any concerns you have. Let me know when you want to chat.

jimlambrt: Ty for the response and explanation!   I hope we can catch up in the morning before HEX kicks off.  I really appreciate the vision you have for protecting the end-user's experience.   

Unfortunately, I think the path we're headed down may make updates very difficult to reason about for the operator in some situations.  Imagine the scenario where I've created an auth method and "made it public", now I update the auth method using the WithForce option (because my provider's discovery info was and is never correct).  The update succeeds, but I'm blissfully unaware that my update had the side-affect of changing the auth method's operational state from public to inactive without raising an error.  

Perhaps we need some way to disambiguate an "incomplete" auth method from an auth method which doesn't pass the "Validate" tests we've prescribed.   This is the sort of ambiguous behavior which could be removed completely from the system, by simply making the update of OperationalStatus explicit.  You either update it directly, or you call one of the explicit domain functions to change the OperationalStatus (MakeInactive, MakePrivate, or MakePublic).  In my experience, system side-affects typically lead to system behavior which are difficult to reason about and properly disambiguate.  

As always, I'm looking forward to discussing this in real-time.  As always, our collaboration makes me a better engineer and I'm very grateful.

internal/auth/oidc/repository_auth_method_update.go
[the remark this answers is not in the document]
jimlambrt: see :https://github.com/hashicorp/boundary/pull/963/commits/2b110c144f910dee53f6d9e0b4bb091a11102b83

internal/auth/oidc/repository_auth_method_update.go
jimlambrt: see: https://github.com/hashicorp/boundary/pull/963/commits/2b110c144f910dee53f6d9e0b4bb091a11102b83

internal/auth/oidc/repository_auth_method_update.go
```diff
@@ -41,6 +41,23 @@ import (
...
+	// TODO(jimlambrt) 3/2021: after some discussion, we've aligned on a need to
+	// disambiguate an error raised because the auth method is not "complete" vs
+	// "warnings" raised from Repository.ValidateAuthMethod(...). when an auth
+	// method and the discovery info published by the OIDC provider seem to
+	// indicate that there's a possible configuration issue.  A future PR, will
+	// distinguish between these two different auth method concerns (one an error
+	// and the other being a warning about a possible error/concern).  In that
+	// future PR, callers will not be allowed to "force" and update to an auth
+	// method which is active (private or public), if that updated would result
+	// in an incomplete auth method which would be useable by users of the
```
mgaffney, line 56:
```suggestion
	// in an incomplete auth method which would be unusable by users of the
```

mgaffney (commented):
One important suggestion in the TODO

mgaffney (approved):

PR #964, opened 2021-02-25
Add TestFreePort(...) which just returns an available free localhost port

jimlambrt:


jefferai (approved):
Approved; however, I wonder if it would be useful to either return the listener instead of the port, or to make that optional (so you return a listener and port, but if the bool is false, the listener is nil and is closed by the defer call). Either way works for me though.

PR #966, opened 2021-02-26
Inclusion of create target error description

pedrorsantana:
**Is your feature request related to a problem? Please describe.**
During a failure in creating the target, the exception caused did not come with a description capable of bringing a quick understanding of the problem, just the stacktrace.

**Describe the solution you'd like**
Use the existing module to include messages.

### DIFF
--- a/internal/target/repository_tcp_target.go
+++ b/internal/target/repository_tcp_target.go
@@ -79,7 +79,7 @@ func (r *Repository) CreateTcpTarget(ctx context.Context, target *TcpTarget, opt
 			var targetOplogMsg oplog.Message
 			returnedTarget = t.Clone()
 			if err := w.Create(ctx, returnedTarget, db.NewOplogMsg(&targetOplogMsg)); err != nil {
-				return errors.Wrap(err, op)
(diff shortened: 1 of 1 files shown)


internal/target/repository_tcp_target.go
```diff
@@ -79,7 +79,7 @@ func (r *Repository) CreateTcpTarget(ctx context.Context, target *TcpTarget, opt
 			var targetOplogMsg oplog.Message
 			returnedTarget = t.Clone()
 			if err := w.Create(ctx, returnedTarget, db.NewOplogMsg(&targetOplogMsg)); err != nil {
-				return errors.Wrap(err, op)
+				return errors.Wrap(err, op, errors.WithMsg("Unable to create target"))
```
louisruch, line 82:
```suggestion
				return errors.Wrap(err, op, errors.WithMsg("unable to create target"))
```
We have been following the standard of not starting with a capital in error messages https://github.com/golang/go/wiki/CodeReviewComments#error-strings

louisruch (changes requested):

louisruch (approved):
Thank you

PR #968, opened 2021-02-26
Remove sentinel errors

louisruch:
### What does this PR do ###

Removes sentinel errors that should no longer be used now that all repos have been refactored
Removes errors.go files in password and sessions that are no longer needed
Cleans up some un wrapped errors that were being returned

### Follow up work ###

After this PR merges only the server repo needs to be refactored


louisruch:
@s-christoff @jimlambrt PTAL

s-christoff (approved):
🤙

jimlambrt (approved):
Looks great!

PR #970, opened 2021-03-01
Fix WithKeyId option

jefferai:
Currently this would validate that a KeyId was in the returned wrapper
but then return the base wrapper; and if it required a database lookup
it wouldn't validate it anyways.

All uses of this were for decryption so it didn't actually affect
anything to this point (because of the multiwrapper) but for any
case where we need to _encrypt_ against a specific version or derive a key
against a specific version this would be broken because you'd end up using the base of the multiwrapper -- unless you then casted to a multiwrapper, pulled out the wrapper you wanted, and did it that way. That's a lot for implementers to know so it makes sense to just return that wrapper if the option is given.

This also properly sends the Key ID around sessions. This worked
currently for the same reason as above but would have broken for any
session in flight once a key was rotated.

### DIFF
--- a/internal/kms/kms.go
+++ b/internal/kms/kms.go
@@ -139,10 +139,6 @@ func (k *Kms) GetExternalWrappers() *ExternalWrappers {
 	return ret
 }
 
-func generateKeyId(scopeId string, purpose KeyPurpose, version uint32) string {
-	return fmt.Sprintf("%s_%s_%d", scopeId, purpose, version)
-}
-
 // GetWrapper returns a wrapper for the given scope and purpose. When a keyId is
 // passed, it will ensure that the returning wrapper has that key ID in the
 // multiwrapper. This is not necesary for encryption but should be supplied for
@@ -168,8 +164,11 @@ func (k *Kms) GetWrapper(ctx context.Context, scopeId string, purpose KeyPurpose
 	val, ok := k.scopePurposeCache.Load(scopeId + purpose.String())
 	if ok {
 		wrapper := val.(*multiwrapper.MultiWrapper)
-		if opts.withKeyId == "" || wrapper.WrapperForKeyID(opts.withKeyId) != nil {
+		if opts.withKeyId == "" {
 			return wrapper, nil
+		}
+		if keyIdWrapper := wrapper.WrapperForKeyID(opts.withKeyId); keyIdWrapper != nil {
+			return keyIdWrapper, nil
 		}
 		// Fall through to refresh our multiwrapper for this scope/purpose from the DB
 	}
@@ -191,6 +190,13 @@ func (k *Kms) GetWrapper(ctx context.Context, scopeId string, purpose KeyPurpose
 		return nil, errors.Wrap(err, op, errors.WithMsg(fmt.Sprintf("error loading %s for scope %s", purpose.String(), scopeId)))
 	}
 	k.scopePurposeCache.Store(scopeId+purpose.String(), wrapper)
+
+	if opts.withKeyId != "" {
+		if keyIdWrapper := wrapper.WrapperForKeyID(opts.withKeyId); keyIdWrapper != nil {
+			return keyIdWrapper, nil
+		}
+		return nil, errors.New(errors.KeyNotFound, op, "unable to find specified key ID")
+	}
 
 	return wrapper, nil
 }
--- a/internal/kms/kms.go
+++ b/internal/kms/kms.go
@@ -305,6 +305,8 @@ func (k *Kms) loadDek(ctx context.Context, scopeId string, purpose KeyPurpose, r
 		keys, err = repo.ListTokenKeys(ctx)
 	case KeyPurposeSessions:
 		keys, err = repo.ListSessionKeys(ctx)
+	default:
+		return nil, errors.New(errors.InvalidParameter, op, "unknown or invalid DEK purpose specified")
 	}
 	if err != nil {
 		return nil, errors.Wrap(err, op, errors.WithMsg("error listing root keys"))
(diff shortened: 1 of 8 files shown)


internal/kms/kms.go
```diff
@@ -139,10 +139,6 @@ func (k *Kms) GetExternalWrappers() *ExternalWrappers {
 	return ret
 }
 
-func generateKeyId(scopeId string, purpose KeyPurpose, version uint32) string {
```
jimlambrt, line 142: ty for removing this!

internal/kms/kms.go
```diff
@@ -168,9 +164,12 @@ func (k *Kms) GetWrapper(ctx context.Context, scopeId string, purpose KeyPurpose
...
 	if ok {
 		wrapper := val.(*multiwrapper.MultiWrapper)
-		if opts.withKeyId == "" || wrapper.WrapperForKeyID(opts.withKeyId) != nil {
+		if opts.withKeyId == "" {
 			return wrapper, nil
 		}
+		if keyIdWrapper := wrapper.WrapperForKeyID(opts.withKeyId); keyIdWrapper != nil {
+			return keyIdWrapper, nil
+		}
 		// Fall through to refresh our multiwrapper for this scope/purpose from the DB
```
jimlambrt, line 173: ty for the comment!

internal/kms/kms_ext_test.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+// NOTE: This is a sequential test that relies on the actions that have come
+// before. Please see the comments for details.
+func TestKms(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+	conn, _ := db.TestSetup(t, "postgres")
+	rw := db.New(conn)
+	wrapper := db.TestWrapper(t)
```
jimlambrt, line 27: nit: I've taken to naming this `rootWrapper` in my newer tests just to be more explicit about its purpose.

internal/kms/kms_ext_test.go
```diff
@@ -0,0 +1,127 @@
...
+	scopePurposeMap := map[string]interface{}{}
+	for i := 1; i < 3; i++ {
+		// This iterates through wrappers for all three scopes and four purposes,
+		// ensuring that the key bytes and IDs are different for each of them,
+		// simulating calling the KMS object from different scopes for different
+		// purposes and ensuring the keys are different when that happens.
+		t.Run(fmt.Sprintf("verify wrappers different x %d", i), func(t *testing.T) {
+			assert, require := assert.New(t), require.New(t)
+			for _, scopeId := range []string{scope.Global.String(), org.GetPublicId(), proj.GetPublicId()} {
+				for _, purpose := range []kms.KeyPurpose{kms.KeyPurposeUnknown, kms.KeyPurposeOplog, kms.KeyPurposeDatabase, kms.KeyPurposeSessions, kms.KeyPurposeTokens} {
```
jimlambrt, line 83: I'm curious, if we should declare a list/dict of known KeyPurposes?  Just last night I discovered that I didn't add the OIDC dek to `Kms.loadDek(...)` properly.  (I've fixed it in my jimlambrt-oidc-wip branch).  Having a defined list might prevent this sort of issues in the future and would be useful in tests like this as well.

internal/kms/kms_ext_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/hashicorp/boundary/internal/iam"
+	"github.com/hashicorp/boundary/internal/kms"
+	"github.com/hashicorp/boundary/internal/types/scope"
+	"github.com/hashicorp/go-kms-wrapping/wrappers/aead"
+	"github.com/hashicorp/go-kms-wrapping/wrappers/multiwrapper"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// NOTE: This is a sequential test that relies on the actions that have come
```
jimlambrt, line 20: Great addition to our unit tests!  ty!

internal/kms/kms_test.go
```diff
@@ -1,127 +1,68 @@
...
 	ctx := context.Background()
 	conn, _ := db.TestSetup(t, "postgres")
 	rw := db.New(conn)
-	wrapper := db.TestWrapper(t)
-	badWrapper := db.TestWrapper(t)
-	repo, err := kms.NewRepository(rw, rw)
-	require.NoError(t, err)
-	kmsCache := kms.TestKms(t, conn, wrapper)
-	org, proj := iam.TestScopes(t, iam.TestRepo(t, conn, wrapper))
+	rootWrapper := db.TestWrapper(t)
```
jimlambrt, line 20: loving `rootWrapper`

jimlambrt (approved):
Just a couple of comments.  Looks great!

internal/kms/kms_ext_test.go
jefferai: Probably it should be `extWrapper`. I actually ended up confusing myself a bit when writing the test because I was originally using `rootWrapper` as you mentioned but that wrapper is actually the external wrapper, not one of the internal keys.

internal/kms/kms_ext_test.go
jefferai: This is actually an unchanged test, I just renamed the file :-D

internal/kms/kms_test.go
jefferai: Updated to `extWrapper` :-)

internal/kms/kms_ext_test.go
jefferai: I was in process of doing this and I think it's good for tests but wouldn't actually have helped with the `loadDek` issue. The issue there really is that the switch case needs a `default`.

PR #972, opened 2021-03-02
Ongoing OIDC work: repo methods for transitioning operational states

jimlambrt:
This PR adds the ability to transition an OIDC auth method's operational states (inactive, private and public)

@jefferai , @s-christoff, @louisruch: I added you as optional reviewers


internal/auth/oidc/repository_auth_method_operational_state.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/kms"
+	"github.com/hashicorp/boundary/internal/oplog"
+)
+
+// MakeInactive will transision an OIDC auth method from either the
+// ActivePrivateState or the ActivePublicState to the InactiveState.
+// No options are supported.
+func (r *Repository) MakeInactive(ctx context.Context, authMethodId string, _ ...Option) error {
```
talanknight, line 16: Since this is modifying the state of the auth method, I think it should take the version number, right?

Also in other methods which change a single field the pattern has been to return the full resource with the updated version number.  See SetPassword for password accounts as an example or Add|Set|Remove Group members for groups, or CancelSession in the session repo.

jimlambrt: I actually didn't think any of your assertions would be true, given that it's not an "updatable" field but rather it's a calculated field.  With that said, I have no strong preference and I'm trying to stay focused on moving forward.

internal/auth/oidc/repository_auth_method_operational_state.go
```diff
@@ -0,0 +1,103 @@
...
+	var updatedAm *AuthMethod
+	_, err = r.writer.DoTx(
+		ctx,
+		db.StdRetryCnt,
+		db.ExpBackoff{},
+		func(_ db.Reader, w db.Writer) error {
+			updatedAm = am.Clone()
+			updatedAm.OperationalState = string(desiredState)
+			dbMask := []string{"OperationalState"}
+			rowsUpdated, err := w.Update(ctx, updatedAm, dbMask, nil, db.WithOplog(oplogWrapper, updatedAm.oplog(oplog.OpType_OP_TYPE_UPDATE)))
```
talanknight, line 89: This doesn't look like it actually uses the version.    Add a TODO?

jimlambrt: sorry about that.  ty!

internal/auth/oidc/repository_auth_method_operational_state.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, errors.New(errors.InvalidParameter, op, fmt.Sprintf("%s is not a valid auth method state", desiredState))
+	}
+	am, err := r.lookupAuthMethod(ctx, authMethodId)
+	if err != nil {
+		return nil, errors.Wrap(err, op)
+	}
+	if am == nil {
+		return nil, errors.New(errors.RecordNotFound, op, fmt.Sprintf("%s auth method not found", authMethodId))
+	}
+	if am.OperationalState == string(desiredState) {
```
talanknight, line 66: Just a question that maybe should be ignored... but should it be an error if the passed in version doesn't match the am.Version, even if the desired state matches the existing operational state?

talanknight (approved):
Looks good! Thanks for adding the version.

internal/auth/oidc/repository_auth_method_operational_state.go
jimlambrt: Hum... the auth method is at the desired state so getting an error might just be odd.

internal/auth/oidc/repository_auth_method_operational_state.go
```diff
@@ -0,0 +1,100 @@
...
+	const op = "oidc.(Repository).MakePublic"
+	updated, err := r.transitionAuthMethodTo(ctx, authMethodId, ActivePublicState, version)
+	if err != nil {
+		return nil, errors.Wrap(err, op)
+	}
+	return updated, nil
+}
+
+func (r *Repository) transitionAuthMethodTo(ctx context.Context, authMethodId string, desiredState AuthMethodState, version uint32, _ ...Option) (*AuthMethod, error) {
+	const op = "oidc.(Repository).updateOperationalState"
```
mgaffney, line 52: The op const does not match the method name.

internal/auth/oidc/repository_auth_method_operational_state.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, errors.Wrap(err, op)
+	}
+	if am == nil {
+		return nil, errors.New(errors.RecordNotFound, op, fmt.Sprintf("%s auth method not found", authMethodId))
+	}
+	if am.OperationalState == string(desiredState) {
+		return am, nil
+	}
+	if am.OperationalState == string(InactiveState) {
+		if err := am.isComplete(); err != nil {
```
mgaffney, line 70: Shouldn't this be a call to `ValidateAuthMethod`?

mgaffney (changes requested):
The transition out of Inactive only checks if the auth method is complete not if it is valid.

internal/auth/oidc/repository_auth_method_operational_state.go
jimlambrt: ty!

internal/auth/oidc/repository_auth_method_operational_state.go
jimlambrt: I don't believe so... unless we want to support some sort of WithForce option for these transition functions.

PR #974, opened 2021-03-03
Ongoing OIDC work: the things required to kick off an OIDC authentication flow

jimlambrt:
This PR adds:

1. oidc.StartAuth(...) which accepts a request to start an OIDC authentication/authorization
2. new options: WithRoundtripPayload and WithKeyId
3. new protos for encoding/encrypting both a request.Token and request.State inside of with a common request.Wrapper, which are returned from StartAuth(...) within a tokenUrl and authUrl 
4. functions to encrypt, decrypt and wrap the request.Token and request.State, along with some bits to derive keys from the oidc DEK.
5. Exporting authtoken.NewAuthTokenId() so StartAuth(...) can create a proper token_request_id
6. define a few new error codes along with their infos.
7. Fix kms.GetWrapper(...) to properly load the oidc DEK.
8. Add a "derived" purpose wrapper cache in the kms package.

@mgaffney, @talanknight: I would appreciate a complete review from both of you, with some special attention to StartAuth(...) 

@jefferai: I would appreciate your review of the kms changes, along with the crypto stuff for the request.Token and request.State

 @s-christoff, @louisruch: I added you as optional reviewers

Thank you!


internal/auth/oidc/service.go
```diff
@@ -2,88 +2,229 @@ package oidc
...
+	if ok {
+		if err := v.Validate(); err != nil {
+			return "", errors.Wrap(err, op)
+		}
+	}
+	marshaled, err := proto.Marshal(m)
+	if err != nil {
+		return "", errors.Wrap(err, op, errors.WithMsg("unable to marshal message"), errors.WithCode(errors.Encode))
+	}
+	blobInfo, err := wrapper.Encrypt(ctx, marshaled, nil)
```
jefferai, line 87: It wouldn't be a bad idea to add some AAD here. Some combo of scope ID/public ID might be a good choice.

internal/auth/oidc/service.go
```diff
@@ -2,88 +2,229 @@ package oidc
...
+}
+
+// derivedKeyPurpose represents the purpose of the derived key.
 //
-// * Use the authtoken.(Repository).IssueAuthToken to issue the request id's
-// token and mark it as issued in the repo.  If the token is already issue, an
-// error is returned.
-func TokenRequest(ctx context.Context, kms *kms.Kms, atRepoFn AuthTokenRepFactory, tokenRequestId string) (string, error) {
-	panic("to-do")
+// TODO (jimlambrt 3/2021) ask @jefferai if this should be moved to the kms package
```
jefferai, line 213: I'm fine with that if it's easier. I don't see any other needs _at the moment_ unless we refactor session establishment key derivation to use it (which we could do but I don't see a benefit right now). So also happy to have this stay here and we can unify later if needed.

internal/auth/oidc/service.go
jimlambrt: ty!

jefferai (approved):

internal/auth/oidc/service_start_auth.go
```diff
@@ -0,0 +1,143 @@
...
+		return nil, nil, errors.New(errors.Unknown, op, "unable to generate nonce", errors.WithWrap(err))
+	}
+	hash, err := provider.ConfigHash()
+	if err != nil {
+		return nil, nil, errors.New(errors.Unknown, op, "unable to get provider config hash", errors.WithWrap(err))
+	}
+	st := &request.State{
+		TokenRequestId:     tokenRequestId,
+		CreateTime:         &timestamp.Timestamp{Timestamp: createTime},
+		ExpirationTime:     &timestamp.Timestamp{Timestamp: exp},
```
talanknight, line 94: I think the appropriate time to decide when to expire is at the time the callback or polling request are made and not at the time the start authenticate returns.  Consider only including the create time and leaving the calculation of the expiration time for the endpoints servicing these requests.

internal/auth/oidc/service_callback.go
```diff
@@ -0,0 +1,42 @@
...
+// for the account.
+//
+// * Use iam.(Repository).LookupUserWithLogin(...) look up the iam.User matching
+// the Account.
+//
+// * Use the authtoken.(Repository).CreateAuthToken(...) to create a pending
+// auth token for the authenticated user.
+func Callback(
+	ctx context.Context,
+	kms *kms.Kms,
```
talanknight, line 36: start auth used the KMS associated with OIDC, Shouldn't we do the same here?

internal/auth/oidc/service.go
```diff
@@ -2,88 +2,230 @@ package oidc
...
+	}
+
+	decryptedMsg, err := wrappingWrapper.Decrypt(ctx, &blobInfo, []byte(fmt.Sprintf("%s%s", wrappedRequest.AuthMethodId, wrappedRequest.ScopeId)))
+	if err != nil {
+		return nil, errors.New(errors.Encrypt, op, "unable to decrypt message", errors.WithWrap(err))
+	}
+	return decryptedMsg, nil
+}
+
+// unwrapMessage does just that, it unwrappes the encoded reqquest.Wrapper proto message
```
talanknight, line 134:
```suggestion
// unwrapMessage does just that, it unwraps the encoded request.Wrapper proto message
```

internal/auth/oidc/service_start_auth.go
```diff
@@ -0,0 +1,143 @@
...
+
+	t := &request.Token{
+		RequestId:      tokenRequestId,
+		ExpirationTime: &timestamp.Timestamp{Timestamp: exp},
+	}
+	encodedEncryptedTk, err := encryptMessage(ctx, requestWrapper, am, t)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, op)
+	}
+	tokenUrl, err = url.Parse(fmt.Sprintf(TokenEndpoint, apiAddr, authMethodId, encodedEncryptedTk))
```
talanknight, line 137: We do not need to return the token url as the client should have access to this.  What the requester needs is an encrypted payload containing token request id and the auth start time.  The request for the token will likely be a `POST` with that value as part of the request body.

talanknight (commented):
Looks good!

internal/auth/oidc/request/request.go
```diff
@@ -0,0 +1,67 @@
+package request
```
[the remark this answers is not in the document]
jimlambrt: ty for the suggestion!

internal/auth/oidc/service_callback.go
jimlambrt: the function signature has already been updated in the "wip" branch: https://github.com/hashicorp/boundary/blob/jimlambrt-oidc-wip/internal/auth/oidc/service_callback.go

internal/auth/oidc/service_start_auth.go
jimlambrt: The RFC specifies the return of the `authorization_request_URL`.  When did the return value change?  Also, why would the client need the auth start time?

internal/auth/oidc/service_start_auth.go
jimlambrt: I disagree.  The TTL will not change in-flight, since it's immutable based on this point in the authentication process.  Of course the callback and token request endpoints are free apply some sort of leeway if they deem it appropriate to the exp, but that doesn't change the time of expiration.

jimlambrt: BTW, CreateTime is purely informational and not required by any downstream actor, so you could argue for not including it in the payload, but I see it sort of like the `iat` claim.  It's just a bit of helpful info for downstream participants.

internal/auth/oidc/service_start_auth.go
jimlambrt: BTW, not generating a full URL for the return is just one less thing to keep track of in the controller, but I am curious when the spec changed... and why the need for start time.  ty!

jimlambrt: I was thinking more about this... how will clients other than our UIs and CLI have "know" what the complete URL should be?  It would seem to be a better UX and inclusive for non-HashiCorp clients to include the complete URL.

talanknight: "When did the return value change?"
I'm confused by the question since the method has not been written yet.  Maybe the question is highlighting another inconsistency in the RFC between the API section for `:authenticate:start` and the Authentication Operations section (which is fine IMO, discussing now seems like an okay time to resolve these inconsistencies. Lets chat on slack and zoom).

Regarding start time: This is related to our other comment thread regarding when is the right time to determine what the ttl of an `:authenticate:token` request should be (`:authenticate:start` vs. `:authenticate:token`)

All endpoints related to OIDC are part of the same open API spec.  Clients would know the `:authenticate:token` since it is published at the same place that `:authenticate:start` is published.

jimlambrt: We discussed this via zoom.  I'm going to return 4 things from StartAuth(...) -> `authUrl *url.URL, tokenUrl *url.URL, tokenId string, e error`

talanknight (approved):

PR #976, opened 2021-03-04
Update CLI error codes to be consistent with `connect`

jefferai:


internal/cmd/commands/database/init.go
```diff
@@ -253,20 +253,20 @@ func (c *InitCommand) Run(args []string) (retCode int) {
 	c.srv.DatabaseUrl, err = config.ParseAddress(urlToParse)
 	if err != nil && err != config.ErrNotAUrl {
 		c.UI.Error(fmt.Errorf("Error parsing database url: %w", err).Error())
-		return 1
+		return 2
```
louisruch, line 256: 3?

louisruch (commented):
nit: given we are standardizing return codes here, it would be nice if they could be made consts instead of magic numbers within the code.  This will just help future coding and debugging to quickly know what each means.  

Feel free to add this as a backlog task as well

internal/cmd/commands/database/init.go
```diff
@@ -181,17 +181,17 @@ func (c *InitCommand) Run(args []string) (retCode int) {
...
 	if err := c.srv.SetupKMSes(c.UI, c.Config); err != nil {
 		c.UI.Error(err.Error())
-		return 1
+		return 2
 	}
 
 	if c.srv.RootKms == nil {
 		c.UI.Error("Root KMS not found after parsing KMS blocks")
-		return 1
+		return 2
```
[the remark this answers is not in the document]
jefferai: Good catch

internal/cmd/commands/database/init.go
jefferai: Here too

jefferai:
@louisruch great idea on the consts; I will make that change now.

jefferai:
@louisruch done, PTAL

louisruch (approved):
LGTM

louisruch (approved):

PR #978, opened 2021-03-05
Fix WithKeyId after recent PR

jefferai:
Fixing the session passing around its key ID (which had previously been
blank) exposed another issue which was caught during bats tests:
WithKeyId statements using the wrong key, which prior to now had been
ignored because the argument was empty, resulting in an ineffective
option. This fixes those situations.

It also updates a bats test that didn't get updated in the CLI return
status PR.


jimlambrt (commented):
Looks great!

PR #985, opened 2021-03-05
Add migration hooks to allow Go functions to be run

jefferai:
These can be run before or after migrations. This should be seen as
anticipatory. There isn't really much added logic, it mostly reorganizes
what's there.


internal/db/schema/manager.go
```diff
@@ -191,9 +193,20 @@ func (b *Manager) runMigrations(ctx context.Context, qp *statementProvider) erro
 		default:
 			// context is not done yet. Continue on to the next query to execute.
 		}
-		if err := b.driver.Run(ctx, bytes.NewReader(qp.ReadUp()), qp.Version()); err != nil {
+		upVer := qp.ReadUp()
+		if upVer.PreHook != nil {
```
talanknight, line 197: Instead of Pre/Post hooks, what if we decoupled them completely from a statement version?  What I am envisioning would be each version could be either an SQL statement that is passed to driver.Run(ctx, ...) or a goFunction that is passed to something like driver.ExecuteFn(ctx, migrateFunction) but not both.

The benefit of this approach, in my opinion, is that it allows the sql.Tx to not have to be exported since there may be a db in the future that we rely on dirty bits instead of transactions.  It also makes the migration steps a little easier to reason about since you can just step through each migration version sequentially.

internal/db/schema/migrations/generate/generate.go
```diff
@@ -118,23 +118,28 @@ func generate(dialect string) {
...
 	}
 }
 
 var migrationsTemplate = template.Must(template.Must(template.New("Content").Parse(
-	`{{ .Name }}: []byte(` + "`\n{{ .Content }}\n`" + `),
-`)).New("MainPage").Parse(`package schema
+	`{{ .Name }}: {
+		Statements: []byte(` + "`\n{{ .Content }}\n`" + `),
+	},
+`)).New("MainPage").Parse(`// Code generated by "make migrations"; DO NOT EDIT.
```
talanknight, line 132: Having this comment above the package makes it a package comment, but this is generated to a package where not every file is auto generated.

talanknight (commented):
This looks like it'll be very helpful for allowing go logic to modify db state in the current migration process.  It is a little difficult to see how it'll look and how easy it'll be to maintain in practice without an example.  I've added just a few comments to spur discussion.

internal/db/schema/migrations/generate/generate.go
talanknight: Oh, sorry. I see now you moved it to its own package.

internal/db/schema/manager.go
jefferai: So to make sure I'm understanding, you are suggesting that rather than pass statements, that we encapsulate the entirety of a version in a Go function? That way a "basic" update function would just execute the statements, but then rather than pre/post hooks we can just mix in logic as needed?

PR #986, opened 2021-03-05
Ongoing OIDC: update fix null field updates and validation

jimlambrt:


internal/auth/oidc/repository_auth_method_update_test.go
```diff
@@ -129,6 +129,40 @@ func Test_UpdateAuthMethod(t *testing.T) {
 				return am
 			},
 		},
+		{
+			name: "inactive-not-complete-no-with-force",
```
talanknight, line 133: Can you add a test which unsets the name and/or description?

jimlambrt: ty!

internal/auth/oidc/repository_auth_method_update_test.go
```diff
@@ -129,6 +129,75 @@ func Test_UpdateAuthMethod(t *testing.T) {
 				return am
 			},
 		},
+		{
+			name: "null-name-description",
```
talanknight, line 133: Please add a test that just unsets a single valueObject like SigningAlg.  I think there is a bug where Update will fail since both the mask and null masks are empty in such a case.

jimlambrt: ty!

PR #988, opened 2021-03-08
fix tests and call to ValidateDiscoveryInfo(...)

jimlambrt:


jimlambrt:
Ooops... too many dependencies on stuff still in jimlambrt-oidc-wip

PR #998, opened 2021-03-15
Create a handlers.Id type for first ValidId parameter.

talanknight:
This tries to ensure ValidId arguments are passed in the right order


jimlambrt (approved):
Looks greet.  ty!

PR #1010, opened 2021-03-17
Packagespec 0.2.2

99:
This PR upgrades to the packagespec 0.2.2


samsalisbury (dismissed):
:shipit:

samsalisbury (commented):
Hm, there seem to be a lot of whitespace differences in the config.yml file here. I wonder why that is, I can't recreate them locally (when I perform this upgrade I see only a single line changed in the `packages-oss.lock/pkgs.yml`).

Let's sync up later @99 and figure out what happened before merging.

samsalisbury (approved):

PR #1011, opened 2021-03-17
docs: clarify CORS configuration options

### LINKED ISSUE #995: CORS frontend behavior. Origin HTTP header
**Describe the bug**
The front-end is attaching the `protocol` while requesting the `authentication` endpoint in the `Origin` header.

Example: 

```
Origin: https://boundary.example.com
```

Meaning that the cors configuration inside the `controller`/`api` must include the protocol otherwise it can not access the web UI.

```
listener "tcp" {
  address = "0.0.0.0:9200"
  purpose = "api"
  tls_disable = true
  cors_enabled = true
  cors_allowed_origins = ["https://${domain}"]
}
```
`tls_disable = true` because the boundary server is behind a load balancer handling the TLS.

**To Reproduce**
Spin up the dev environment, check the request headers for url:

```
https://boundary.example.com/v1/auth-methods/ampw_69ZOdUZiqA:authenticate
```
Header: 
```
Origin: https://boundary.example.com
```

**Expected behavior**
Maybe it could be interesting to just add the hostname in the `Origin` HTTP header.
Another possibility is to modify the `cors` docs to specify the behavior. Currently, it only specifies the `hostname`.

**Additional context**
Pretty simple setup in AWS.

443:(ELB) LoadBalancer:ACM -> 9200:EC2 -> RDS and KMS

randallmorey:
This closes #995

### DIFF
--- a/website/content/docs/configuration/controller.mdx
+++ b/website/content/docs/configuration/controller.mdx
@@ -136,7 +136,7 @@ listener "tcp" {
   # Uncomment to enable CORS for the Admin UI. Be sure to set the allowed origin(s)
   # to appropriate values.
   #cors_enabled = true
-  #cors_allowed_origins = ["yourcorp.yourdomain.com"]
+  #cors_allowed_origins = ["https://yourcorp.yourdomain.com", "serve://boundary"]
 }
 
 # Data-plane listener configuration block (used for worker coordination)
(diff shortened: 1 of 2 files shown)


angelbarrera92:
LGTM :)

jefferai (approved):
Approved, but please add into the `unix` listener page as well. I should probably ask the web team if there is a way to make most of these two an include...

malnick (approved):

PR #1016, opened 2021-03-18
Honor child scope list permissions when recursing

jefferai:
Current behavior is that you need list grant on the parent scope when
recursively listing, and this is sufficient. This is good for
administrative workflows but can be unintuitive for resources that only
exist at a project level. This change allows permissions on the child
scope to also be taken into account, so that you can start a recursive
list from a parent scope without direct list support on that scope but
still get back child scopes for which list access is granted.

This also fixes a bug where visibility of a scope for recursive listing
would be based on whether the user had list resources for roles within
that scope instead of for the resource type in question.


jimlambrt (changes requested):
just a few suggestions.

internal/servers/controller/common/scopeids/scope_ids.go
```diff
@@ -29,8 +30,14 @@ func GetScopeIds(
...
 	rootScopeId string,
+	// The type of resource we are listing
+	typ resource.Type,
 	// Whether or not the search should be recursive
 	recursive bool) ([]string, map[string]*scopes.ScopeInfo, error) {
+	const op = "GetScopeIds"
+	if typ == resource.Unknown {
+		return nil, nil, errors.New(errors.InvalidParameter, op, "unknown resource")
+	}
 	// Get the list of scope IDs. If it's recursive, check ACLs.
```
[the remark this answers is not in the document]
jefferai: Sure, I'll add that.

internal/servers/controller/handlers/authmethods/authmethod_service.go
```diff
@@ -128,14 +128,26 @@ func (s Service) ListAuthMethods(ctx context.Context, req *pbs.ListAuthMethodsRe
 	}
 	authResults := s.authResult(ctx, req.GetScopeId(), action.List)
 	if authResults.Error != nil {
-		return nil, authResults.Error
+		// If it's forbidden, and it's a recursive request, and they're
+		// successfully authenticated but just not authorized, keep going as we
+		// may have authorization on downstream scopes.
+		if authResults.Error == handlers.ForbiddenError() &&
```
jimlambrt, line 134: is there a anyway to provide a func that combines authResults(...) and GetScopeIds(...) that could be used in all the List functions?  It seems like this block is going to be copied quite a bit.

internal/servers/controller/common/scopeids/scope_ids.go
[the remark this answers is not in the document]
jefferai: Done.

internal/servers/controller/handlers/authmethods/authmethod_service.go
jefferai: I'll do that as a follow after this PR. I wanted to keep changes minimal (as opposed to additions) for easy reviewing.

jimlambrt (changes requested):
In this PR the changes to: Internal/common/scopeids/scope_ids.go  have no unit tests.

There are quite a few missing unit tests in the packages touched by this PR.  I’m not suggesting that all these missing tests be addressed in this PR, but we should note them and address them in future PRs:

- internal/auth/auth.go could use some better coverage for things like the recovery KMS code branches.
- internal/perms/options.go -> it might be nice to have unit tests for these options
- internal/servers/controller/handlers/errors.go -> no unit tests for NotFoundError(), Is(..), ApiErrorWithCode(…), ForbiddenError(…), UnauthenticatedError()
- internal/servers/controller/handlers/mask_manager.go -> no units for MaskContains(…)
- internal/servers/controller/handlers/outgoing_interceptor.go -> OutgoingInterceptor(…) missing tests for the case where *pbs.AuthenticateLoginResponse
- internal/servers/controller/handlers/authmethods/authmethods_service.go -> currently no unit tests for Service.AuthenticateLogin(…)
- internal/servers/controller/handlers/targets/target_service.go -> currently no unit tests for Service.AuthorizeSession(…)

jefferai:
@jimlambrt As discussed, updated the tests. They can be simplified more -- right now I haven't changed the logic to copy part of the group service code -- but at this point that's merely cleanup as the test just goes further than it needs to. So it's ready for another look.

internal/servers/controller/handlers/groups/group_service_test.go
```diff
@@ -1356,3 +1359,68 @@ func TestRemoveMember(t *testing.T) {
...
+	require.Error(err)
+	require.Nil(out)
+	require.Equal(err.Error(), handlers.ForbiddenError().Error())
+
+	// Add list on the project
+	r := iam.TestRole(t, conn, p.GetPublicId())
+	iam.TestRoleGrant(t, conn, r.GetPublicId(), "id=*;type=group;actions=list,read")
+	iam.TestUserRole(t, conn, r.GetPublicId(), "u_auth")
+
+	// Attempt a recursive list at the org level with the given user. It should
```
talanknight, line 1416: Can you add 1 more case where you attempt a recursive list on global?

Along those same lines, there are resources that validate that list requests's scopes are for the scopes that the resource can be created in.  For example host_catalog_service.go validateListRequest validates the scope id as being a valid project scope.  There may be others.

talanknight (commented):
Looks good.  Just wanting to figure out my 1 question regarding resources that aren't allowed in a specific scope, otherwise LGTM.

internal/servers/controller/handlers/groups/group_service_test.go
jefferai: > Can you add 1 more case where you attempt a recursive list on global?

Actually I was likely going to remove this test as it's a much much much more stripped down version of what's in scope_ids_test.go. :-)

That's good feedback about validating list requests. I can skip that logic if it's recursive.

jefferai: Ah -- that's because it's not new here in this PR. Host catalogs already skips that validation if it's recursive! Same with targets.

talanknight (approved):

internal/servers/controller/handlers/groups/group_service_test.go
jefferai: Removed this test in favor of the one that ended up in scope_ids_test.go.

jimlambrt (approved):
Looks great!

PR #1017, opened 2021-03-19
Bump proto/grpc deps

jefferai:


mgaffney (approved):

talanknight (commented):
There are a number of changes in these dependency updates including to the field masks and optional field.  None of them appear to affect us currently however.

Could you please run `make tools` and `make gen` with these updates?  I think the protobuf version update will change some generated .pb.go files.

jefferai:
I did run make tools/make gen before submitting this! There are no changes.

PR #1019, opened 2021-03-19
New error codes.

jimlambrt:


louisruch:
@talanknight PTAL at the changes to API interceptor and new errors codes as well

talanknight (approved):

PR #1026, opened 2021-03-24
docs: Clarify note on VAULT_TOKEN env variable

mocofound:
A slight update for clarity regarding the wording of the note regarding the recommendation of using the 'VAULT_TOKEN' environment variable as opposed to using the 'token' parameter in the configuration file.


jefferai (approved):
Thanks!

PR #1029, opened 2021-03-25
Fix test scope errors

jefferai:
When using a disabled verifier context we were faking scope information. Due to other recent changes, this was now causing test errors. This adjusts that flow to actually query the DB, fixing tests and making them more like real code.


jefferai:
@jimlambrt @talanknight All tests passing for me. As previously discussed, merging.

PR #1039, opened 2021-03-28
Ongoing OIDC: add support for migration logging to db and primary auth method migration

jimlambrt:
This PR adds:

- Logging to the db during migrations which requires a new table (with appropriate triggers) and changes to the `schema` package.
- The ability to override the global migrationStates (primarily in support of testing the migrations themselves) when running migrations.  this required adding support for options to a few of the types in the `schema` package.
-  Sets address primary_auth_method during migration for any scope with only one auth method and logs all scopes which have more than one auth method.

@mgaffney - I would appreciate your review of the schema changes for logging and the primary auth method update and reporting statements.

@jefferai, @talanknight - Currently, I'm outputting the migrations logs only for the `database migrate`, not the `init` subcommand.  I'd appreciate your review of the CLI changes.

@talanknight - I'm appreciate a complete review of this PR, especially the changes in the

### DIFF
--- a/internal/db/schema/manager.go
+++ b/internal/db/schema/manager.go
@@ -6,6 +6,7 @@ import (
 	"database/sql"
 	"fmt"
 	"io"
+	"time"
 
 	"github.com/hashicorp/boundary/internal/db/schema/postgres"
 	"github.com/hashicorp/boundary/internal/errors"
@@ -38,16 +39,25 @@ type driver interface {
 // the underlying boundary database schema.
 // Manager is not thread safe.
 type Manager struct {
-	db      *sql.DB
-	driver  driver
-	dialect string
+	db              *sql.DB
+	driver          driver
+	dialect         string
+	migrationStates map[string]migrationState
 }
 
 // NewManager creates a new schema manager. An error is returned
 // if the provided dialect is unrecognized or if the passed in db is unreachable.
-func NewManager(ctx context.Context, dialect string, db *sql.DB) (*Manager, error) {
+func NewManager(ctx context.Context, dialect string, db *sql.DB, opt ...Option) (*Manager, error) {
 	const op = "schema.NewManager"
 	dbM := Manager{db: db, dialect: dialect}
+	opts := getOpts(opt...)
+	if opts.withMigrationStates != nil {
+		dbM.migrationStates = opts.withMigrationStates
+	} else {
+		// intentionally set it to the reference, so changes to the global var
+		// will be reflected in the manager instance
+		dbM.migrationStates = migrationStates
+	}
 	switch dialect {
 	case "postgres":
 		var err error
@@ -156,7 +166,7 @@ func (b *Manager) RollForward(ctx context.Context) error {
 		return errors.New(errors.NotSpecificIntegrity, op, fmt.Sprintf("schema is dirty with version %d", curVersion))
 	}
 
-	if err = b.runMigrations(ctx, newStatementProvider(b.dialect, curVersion)); err != nil {
+	if err = b.runMigrations(ctx, newStatementProvider(b.dialect, curVersion, WithMigrationStates(b.migrationStates))); err != nil {
 		return errors.Wrap(err, op)
 	}
 	return nil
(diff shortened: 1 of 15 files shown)


jimlambrt:
@talanknight - I added a `panic` in the deferred unlock of an advisory lock during migrations in `RollForward(...)`.  I'm not sure this ideal, but the existing implementation didn't catch the error which seemed even worse.  ty!

see: https://github.com/hashicorp/boundary/blob/5647d92e30ef866fd815eb9d60bf9645567bde88/internal/db/schema/manager.go#L157

internal/db/schema/options.go
```diff
@@ -0,0 +1,37 @@
...
+	return opts
+}
+
+// Option - how Options are passed as arguments.
+type Option func(*options)
+
+// options = how options are represented
+type options struct {
+	withMigrationStates map[string]migrationState
+	withDeleteLog       bool
```
jefferai, line 18: It doesn't seem like this is ever used outside of tests...?

jimlambrt: that's correct... and without it doesn't appear you can write a test that only runs a subset of migrations.  Since my migration includes both DML and DDL, I need a way to run up migrations up to a specific migrations, then seed the database with test state, and then run my DML migration, and finally assert that the state is valid.

jefferai (approved):

internal/db/schema/migrations/postgres/2/80_log.up.sql
```diff
@@ -0,0 +1,48 @@
...
+  end;
+$$ language plpgsql;
+comment on function log_migration_version() is
+'log_migration_version will set the log_migration entries to the current migration version';    
+
+create trigger
+  migration_version_column
+before
+insert on log_migration
+  for each row execute procedure log_migration_version();
```
mgaffney, line 48: Missing a `commit;` statement at the end of the file which causes an error when I run the sql files using the shell.

Command from the root directory:
```zsh
for f in $(ls internal/**/*.up.sql); do psql -d $PG_DB -f "$f"; done
```

Output:
```
psql:internal/db/schema/migrations/postgres/2/87_iam.up.sql:67: ERROR:  relation "log_migration" does not exist
LINE 17: insert into log_migration(entry)
                     ^
Time: 0.305 ms
ROLLBACK
```

mgaffney (changes requested):

internal/db/schema/migrations/postgres/2/80_log.up.sql
jimlambrt: ty!

mgaffney (approved):

internal/db/schema/manager.go
```diff
@@ -200,3 +214,48 @@ func (b *Manager) runMigrations(ctx context.Context, qp *statementProvider) erro
 	}
 	return nil
 }
+
+// LogEntry represents a log entry generated during migrations.
+type LogEntry struct {
```
talanknight, line 219: This is great!

internal/db/schema/options.go
```diff
@@ -0,0 +1,37 @@
...
+func getDefaultOptions() options {
+	return options{}
+}
+
+// WithMigrationStates provides an optional migration states.
+func WithMigrationStates(states map[string]migrationState) Option {
+	return func(o *options) {
+		o.withMigrationStates = states
+	}
+}
```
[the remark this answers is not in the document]
jimlambrt: While I appreciate the option is currently only needed for tests, I don't see some inherently dangerous about the option should someone decide to use it to run a specific set of migrations in prod.

internal/db/schema/migrations/postgres/2/87_iam_test.go
```diff
@@ -0,0 +1,110 @@
...
+		m, err := schema.NewManager(ctx, dialect, d, schema.WithMigrationStates(oState))
+		require.NoError(err)
+
+		assert.NoError(m.RollForward(ctx))
+		state, err := m.CurrentState(ctx)
+		require.NoError(err)
+		assert.Equal(priorMigration, state.DatabaseSchemaVersion)
+		assert.False(state.Dirty)
+
+		// okay, now we can seed the database with test data
```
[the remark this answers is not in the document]
jimlambrt: Isn't your assertion true for all tests?  I'm not seeing how this test is any different that all the other unit tests that use test functions from other packages.

talanknight (approved):

internal/db/schema/migrations/postgres/2/87_iam_test.go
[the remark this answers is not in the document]
talanknight: Just to document our conversation from off thread, The best long term fix for this type of concern is to have a better solution for integration tests which would allow us to test version to version migration steps.  Regarding this specific situation having it as a unit test like this makes sense for now since the alternative would be mutating the sql directly which would not necessarily put the data in a state that accurately represents what the code would/could put it in.

Thanks for taking the time to discuss this with me!

PR #1040, opened 2021-03-29
Job Repo

louisruch:
### What does this PR do

Add Job and JobRun packages and related repos


louisruch:
@mgaffney PTAL
@s-christoff @jimlambrt Adding you as optional reviewers of this PR as well

jimlambrt:
Broadly, why are we using generated Ids over private IDs?  Is there some reason we can't just generate a private ID for every JobRun?  

Also, why are we not oplogging Job and JobRun operations?

internal/job/repository_job_run.go
```diff
@@ -0,0 +1,199 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/db/common"
+	"github.com/hashicorp/boundary/internal/errors"
+)
+
+// CreateJobRun inserts run into the repository and returns a new JobRun
+// containing the run's id, run is not changed. run must not contain
```
jimlambrt, line 14: why not just use a PrivateId vs an "identity" id?

internal/job/repository_job_run.go
```diff
@@ -0,0 +1,199 @@
...
+	if id == 0 {
+		return db.NoRowsAffected, errors.New(errors.InvalidParameter, op, "missing id")
+	}
+
+	run := allocJobRun()
+	var rowsDeleted int
+	_, err := r.writer.DoTx(
+		ctx, db.StdRetryCnt, db.ExpBackoff{},
+		func(_ db.Reader, w db.Writer) (err error) {
+			rowsDeleted, err = w.Delete(ctx, run, db.WithWhere("id = ?", id))
```
jimlambrt, line 184: I'm not sure this where clause is needed, if `id` is tagged as the primary key.

internal/job/testing.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/hashicorp/boundary/internal/servers"
+	wrapping "github.com/hashicorp/go-kms-wrapping"
+	"github.com/hashicorp/go-uuid"
+	"github.com/jinzhu/gorm"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+var (
+	zeroTime   = &timestamp.Timestamp{Timestamp: &timestamppb.Timestamp{}}
```
jimlambrt, line 20: perhaps name these package vars with a prefix of `test`?

jimlambrt (changes requested):
Great job (pun intended)!  Just a few questions.

louisruch:
> Broadly, why are we using generated Ids over private IDs? Is there some reason we can't just generate a private ID for every JobRun? I'm not seeing the requirement/use-case for "identity" IDs here.
> 
> Also, why are we not oplogging Job and JobRun operations?

Darn!  Thanks I was meant to add oplog before sending out review request will do that today!

internal/job/doc.go
```diff
@@ -0,0 +1,32 @@
+// Package job provides a Job and a JobRun struct suitable for persisting to
+// the repository.
+//
+// The uniqueness of a job is determined by the name and code input parameters, this
+// uniqueness is enforced by the database.  The private_id is generated by the
```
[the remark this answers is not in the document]
louisruch: I added a section in the RFC about how the job registration and running works.  TLDR: the lookup for getting the Job to run from the map on the controller will be keyed off the id.

internal/job/doc.go
```diff
@@ -0,0 +1,32 @@
+// Package job provides a Job and a JobRun struct suitable for persisting to
```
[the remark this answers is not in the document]
louisruch: Currently no, at some point in the future maybe.  It depends on how we expose jobs to plugins and if we ever want to have users create jobs.

internal/job/repository_job_run.go
louisruch: ty

internal/job/testing.go
louisruch: ty

louisruch:
@jimlambrt @mgaffney this should be ready for review again, I was having some strange git issues I ended up rebasing at one point sorry if that messes with you checking it out locally.

internal/job/repository_job.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	return j, rowsUpdated, nil
+}
+
+// LookupJob will look up a job in the repository using the privateId. If the job is not
+// found, it will return nil, nil.
+//
+// All options are ignored.
+func (r *Repository) LookupJob(ctx context.Context, privateId string, _ ...Option) (*Job, error) {
```
mgaffney, line 154: Will a `ListJobs` method be added?

internal/job/repository_job_run.go
```diff
@@ -0,0 +1,325 @@
...
+)
+
+// FetchWork queries the job repository for available work and returns a JobRun
+// of a job that needs to be run, if there are not jobs that need to be run `nil, nil`
+// will be returned.
+//
+// • serverId is required and is the private_id of the server that will run the job.
+//
+// All options are ignored.
+func (r *Repository) FetchWork(ctx context.Context, serverId string, _ ...Option) (*JobRun, error) {
```
mgaffney, line 25: Is a controller only allowed to run one job at a time?

mgaffney: Also, the name `FetchWork` seems odd to me for two reasons. First, calling this method results in changes to the database and I wouldn't expect anything named "Fetch" to make changes. And second, "Work" is not used in any of the types in this package.

mgaffney (changes requested):
Looks good so far but I wasn't able to finish reviewing all of the code. I left some comments and questions for the parts I did review.

internal/job/query.go
```diff
@@ -0,0 +1,43 @@
+package job
+
+const fetchWorkQuery = `
+	SELECT
```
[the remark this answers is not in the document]
louisruch: sure! Thanks

internal/job/job.go
```diff
@@ -0,0 +1,93 @@
...
+import (
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/job/store"
+	"github.com/hashicorp/boundary/internal/oplog"
+	"google.golang.org/protobuf/proto"
+)
+
+const (
+	// DefaultJobTableName is the default name of the database table that is used to persist a Job
+	DefaultJobTableName = "job"
```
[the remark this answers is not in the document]
louisruch: So I used `internal/target` as the base to make the job repo from.  It was exported there even though it was not used outside the library.  I thought about why and my guess was that the godoc will show this and perhaps the Default table name would be good to show.  Looking now I see this is not a consistent pattern as I thought and there is a mix across the code between:

- defaultXTableName
- DefaultXTableName
- and a third option that does not have the default as a const and just returns it in the `TableName` method

So long story no it does not unless we do want it in godocs

internal/job/job.go
```diff
@@ -0,0 +1,93 @@
...
+//
+// • Code is not user facing and should be used to distinguish unique jobs of
+// the same type that can run in parallel.
+//
+// • Description is the human-friendly description of the job.
+//
+// WithNextScheduledRun() is the only valid option.  If this option is not
+// provided the NextScheduledRun of the job will default to zero time, and be available
+// to run immediately.
+func NewJob(name, code, description string, opt ...Option) (*Job, error) {
```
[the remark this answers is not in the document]
louisruch: Sure but lets chat Monday about the overall structure of this first

internal/job/job_run_status.go
```diff
@@ -0,0 +1,32 @@
...
+	// Completed represents that the job run has successfully finished
+	Completed = "completed"
+
+	// Failed represent that the job run had an error during execution
+	Failed = "failed"
+
+	// Interrupted represents that the job run was interrupted by a server
+	// other than the server running the job.
+	Interrupted = "interrupted"
+)
```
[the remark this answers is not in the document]
louisruch: Sure

internal/job/options.go
```diff
@@ -0,0 +1,45 @@
...
+
+func getDefaultOptions() options {
+	return options{
+		withNextScheduledRun: &timestamp.Timestamp{Timestamp: &timestamppb.Timestamp{}},
+		withJobRunStatus:     Running,
+	}
+}
+
+// WithNextScheduledRun provides an option to provide the next scheduled run time for a job.
+func WithNextScheduledRun(ts *timestamp.Timestamp) Option {
```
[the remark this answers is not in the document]
louisruch: Yeah I agree having the make the protobuf each time was a bit annoying even in testing

internal/job/options.go
```diff
@@ -0,0 +1,45 @@
...
+	return func(o *options) {
+		o.withNextScheduledRun = ts
+	}
+}
+
+// WithJobRunStatus provides an option to provide the run status for the job run.
+func WithJobRunStatus(s string) Option {
```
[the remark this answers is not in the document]
louisruch: sure

internal/job/job.go
```diff
@@ -0,0 +1,93 @@
...
+		return nil, errors.New(errors.InvalidParameter, op, "missing description")
+	}
+
+	opts := getOpts(opt...)
+	job := &Job{
+		Job: &store.Job{
+			Name:             name,
+			Description:      description,
+			Code:             code,
+			NextScheduledRun: opts.withNextScheduledRun,
```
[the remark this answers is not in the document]
louisruch: Yeah I saw that in your latest PR.  I was originally going down this route, but then changed it to having each job determine the time to run the next job based on a time input which would be the database current_time. `NextRun(currentTime time.Time) time.Time`.  This was done to allow for the next run to be calculated based on cron expression or similar rather than just every 15 minutes.

However as I am coding I am not loving this as much and thinking of just limiting jobs to returning a durtaion for now, lets chat Monday

internal/job/repository_job.go
louisruch: I am not sure, I had a simple `ListJob` but I didnt really have a use case for it, so I pulled it out

internal/job/repository_job_run.go
louisruch: No, since there is no denylisting currently being created I wanted to limit the work allocation to a single job to give all controllers a chance to get work.  I was considering making this a input or a similar `WithLimit` option but haven't yet.

I went around in circles on this, I didn't want to name it `GetJobRun` because that seems even more like its just getting a JobRun from the database.  I didn't want to name it `CreateJobRun` because we might not actually create a `JobRun` 

Lets chat Monday

mgaffney (approved):
Looks great! Nice work!

PR #1041, opened 2021-03-29
Ongoing OIDC: ensure auth method names are unique across all subtypes

jimlambrt:


internal/db/schema/migrations/postgres/2/90_auth.up.sql
```diff
@@ -0,0 +1,87 @@
...
+alter table auth_method
+  add column name wt_name;
+
+alter table auth_method
+  add constraint auth_method_scope_id_name_uq
+    unique (scope_id, name);
+
+
+-- insert_auth_method_subtype() is a replacement of the function definition in
+-- migraiton 07_auth.up.sql  This new defintion also inserts the sub type's name
```
jefferai, line 14:
```suggestion
-- migration 07_auth.up.sql  This new defintion also inserts the sub type's name
```

internal/db/schema/migrations/postgres/2/90_auth.up.sql
```diff
@@ -0,0 +1,87 @@
...
+alter table auth_method
+  add column name wt_name;
+
+alter table auth_method
+  add constraint auth_method_scope_id_name_uq
+    unique (scope_id, name);
+
+
+-- insert_auth_method_subtype() is a replacement of the function definition in
+-- migration 07_auth.up.sql  This new defintion also inserts the sub type's name
```
jefferai, line 14:
```suggestion
-- migration 07_auth.up.sql  This new definition also inserts the sub type's name
```

internal/db/schema/migrations/postgres/2/90_auth.up.sql
```diff
@@ -0,0 +1,87 @@
...
+begin
+  insert into auth_method
+    (public_id, scope_id, name)
+  values
+    (new.public_id, new.scope_id, new.name);
+  return new;
+end;
+$$ language plpgsql;
+comment on function insert_auth_method_subtype() is
+'insert_auth_method_subtype() inserts sub type date into the base type auth method table';
```
jefferai, line 30:
```suggestion
'insert_auth_method_subtype() inserts sub type name into the base type auth method table';
```

internal/db/schema/migrations/postgres/2/90_auth.up.sql
```diff
@@ -0,0 +1,87 @@
...
+-- ensure that auth method names are unique across all sub types.
+alter table auth_method
+  add column name wt_name;
+
+alter table auth_method
+  add constraint auth_method_scope_id_name_uq
+    unique (scope_id, name);
+
+
+-- insert_auth_method_subtype() is a replacement of the function definition in
```
jefferai, line 13: Should this migration also drop the name column from the subtype so that we don't have duplicate data?

jefferai:
This creates the triggers but don't we need a migration to populate the data into `auth_method` for existing auth methods at the time the migration is run?

Also I'm not entirely following the delete triggers -- are those specifically related to this change or were they simply missing before?

internal/db/schema/migrations/postgres/2/90_auth.up.sql
jimlambrt: I don't think so, `name` is similar to the rest of the columns that are inserted into the base type's table.

jimlambrt:
> This creates the triggers but don't we need a migration to populate the data into `auth_method` for existing auth methods at the time the migration is run?
> 

Ty!  I had overlooked this requirement.  great catch.

> Also I'm not entirely following the delete triggers -- are those specifically related to this change or were they simply missing before?

The delete triggers were missing.

jimlambrt:
@jefferai - I'm now populating the auth_method.name column in the migration.  it's ready for your re-review.  ty!

internal/db/schema/migrations/postgres/2/90_auth.up.sql
```diff
@@ -0,0 +1,97 @@
...
+where 
+  auth_method.public_id = pw.public_id and
+  pw.name is not null;
+
+-- insert_auth_method_subtype() is a replacement of the function definition in
+-- migration 07_auth.up.sql  This new definition also inserts the sub type's name
+-- into the base type. The name column must be on the base type, so the database
+-- can ensure that auth method names are unique across all sub types.
+create or replace function
+  insert_auth_method_subtype()
```
mgaffney, line 28: Should there be `create trigger` calls for this function for `auth_password_method` and `auth_oidc_method`?

mgaffney (changes requested):

internal/db/schema/migrations/postgres/2/90_auth.up.sql
jimlambrt: there were already triggers for this function on those tables.  this just replaces the function's definition.

mgaffney (approved):

jefferai (approved):

PR #1045, opened 2021-03-29
OIDC Account Creation and Account go SDK

talanknight:
This adds all the functionality for creating OIDC accounts from the go sdk level all the way down to the repo.  Since the go SDK can now create OIDC accounts I've also added the other account related go SDK tests for oidc.


jimlambrt (changes requested):
Just a couple minor things.

talanknight:
I've made quite a few changes since @jimlambrt 's initial review.  This now allows the auth method to be created without a discovery_url set and only lets an account be created on an auth method which has that value set and with the issuer matching that value.   Any renaming of fields was focused on API side, but we probably will want to come back and update the storage to match in a future PR.

internal/auth/oidc/account.go
```diff
@@ -40,25 +40,22 @@ type Account struct {
 // value being unique
 //
 // See: https://openid.net/specs/openid-connect-core-1_0.html
-func NewAccount(authMethodId string, issuerId *url.URL, subjectId string, opt ...Option) (*Account, error) {
+func NewAccount(authMethodId string, subjectId string, opt ...Option) (*Account, error) {
```
jimlambrt, line 43: mind adding WithIssuer to the list of supported options in the godoc?

internal/auth/oidc/repository_account.go
```diff
@@ -12,6 +12,84 @@ import (
 	"github.com/hashicorp/boundary/internal/oplog"
 )
 
+// CreateAccount inserts a into the repository and returns a new Account
```
jimlambrt, line 15: "inserts a" is correct but it's not that readable.

internal/auth/oidc/repository_account.go
```diff
@@ -12,6 +12,84 @@ import (
...
+	if a.IssuerId != "" {
+		return nil, errors.New(errors.InvalidParameter, op, "issuer id must be empty")
+	}
+	if scopeId == "" {
+		return nil, errors.New(errors.InvalidParameter, op, "missing scope id")
+	}
+
+	a = a.Clone()
+
+	am, err := r.LookupAuthMethod(ctx, a.AuthMethodId)
```
jimlambrt, line 53: hum... I'm sort of rethinking this.  If we only allow operators to provision accounts for the current auth_oidc_method.issuer then how will operators provision accounts if they need to prepare for an upcoming issuer change?  I'm not sure how real this use-case is but it's yet again an example of Boundary forcing a specific workflow on the operator which may be incorrect.

jimlambrt: I'm fine with having the current auth method's issuer be the default, but I"m not sure we should force that.

internal/db/schema/migrations/postgres/2/84_oidc.up.sql
```diff
@@ -363,6 +363,34 @@ create trigger
 before insert on auth_oidc_account
   for each row execute procedure insert_auth_oidc_account_subtype();
```
jimlambrt, line 365: I just added a lenghty comment on the changes to repository_account.go basically questioning the premise of enforcing this workflow on the operators.

internal/auth/oidc/repository_account.go
jimlambrt: If we do decide to move forward with either providing a default issuer or enforcing this constraint on operators, then perhaps a comment about how looking up this now (and the race condition is creates) is acceptable?

internal/auth/oidc/repository_account.go
```diff
@@ -12,6 +12,84 @@ import (
...
+			if err := w.Create(ctx, newAccount, db.WithOplog(oplogWrapper, a.oplog(oplog.OpType_OP_TYPE_CREATE, scopeId))); err != nil {
+				return errors.Wrap(err, op)
+			}
+			return nil
+		},
+	)
+
+	if err != nil {
+		if errors.IsUniqueError(err) {
+			return nil, errors.New(errors.NotUnique, op, fmt.Sprintf("in auth method %s: name %q already exists or subject %q already exists for issuer %q",
```
jimlambrt, line 85: "exists for issuer %q in scope %s"

internal/servers/controller/handlers/accounts/account_service.go
```diff
@@ -312,6 +316,71 @@ func (s Service) createInRepo(ctx context.Context, authMethodId, scopeId string,
...
+		return nil, errors.Wrap(err, op, errors.WithMsg("unable to create user"))
+	}
+	if out == nil {
+		return nil, handlers.ApiErrorWithCodeAndMessage(codes.Internal, "Unable to create user but no error returned from repository.")
+	}
+	return out, nil
+}
+
+func (s Service) createInRepo(ctx context.Context, authMethodId, scopeId string, item *pb.Account) (*pb.Account, error) {
+	const op = "account_service.(Service).createInRepo"
```
jimlambrt, line 359: Just an observation that when you call createInRepo you have an AuthMethod and if you passed that, you wouldn't need the scopeId and then you wouldn't have back to back string parameters in the signature.

internal/servers/controller/handlers/accounts/account_service.go
```diff
@@ -312,6 +316,71 @@ func (s Service) createInRepo(ctx context.Context, authMethodId, scopeId string,
...
+	if item == nil {
+		return nil, errors.New(errors.InvalidParameter, op, "nil item")
+	}
+	var out auth.Account
+	switch auth.SubtypeFromId(authMethodId) {
+	case auth.PasswordSubtype:
+		am, err := s.createPwInRepo(ctx, authMethodId, scopeId, item)
```
jimlambrt, line 366: same observation as about about the AuthMethod and the scopeId parameter.

internal/servers/controller/handlers/accounts/account_service.go
```diff
@@ -312,6 +316,71 @@ func (s Service) createInRepo(ctx context.Context, authMethodId, scopeId string,
...
+		if err != nil {
+			return nil, err
+		}
+		if am == nil {
+			return nil, handlers.ApiErrorWithCodeAndMessage(codes.Internal, "Unable to create auth method but no error returned from repository.")
+		}
+		out = am
+	case auth.OidcSubtype:
+		am, err := s.createOidcInRepo(ctx, authMethodId, scopeId, item)
```
jimlambrt, line 375: same observation as about about the AuthMethod and the scopeId parameter.

jimlambrt (changes requested):
Looks great... a few suggestions and sort of one bigger question about only allowing provisioned accounts to have an issuer that equals the auth methods issuer.

internal/auth/oidc/ids.go
```diff
@@ -21,12 +21,9 @@ func newAuthMethodId() (string, error) {
 	return id, nil
 }
 
-func newAccountId(am *AuthMethod, issuer, sub string) (string, error) {
+func newAccountId(authMethodId, issuer, sub string) (string, error) {
```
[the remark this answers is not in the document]
talanknight: If we do not look up the auth method when creating an account (which would be the result of implementing your comment regarding providing an optional `issuer`) then we don't have the auth method.  This function doesn't need an auth method, it just needs the auth method's public id to perform its task so I think it makes sense to narrow the function's parameters down to only what it needs.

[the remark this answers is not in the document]
jimlambrt: does our Slack discussion change your opinion?

internal/auth/oidc/repository_account.go
talanknight: I updated the PR to reflect this.  It is now optional to provide the issuer when creating an account.

internal/servers/controller/handlers/authmethods/authmethod_service_test.go
```diff
@@ -907,1715 +907,3 @@ func TestCreate(t *testing.T) {
 		})
 	}
 }
-
-
-// TODO: remove everything under this.
```
jimlambrt: ty!

internal/servers/controller/handlers/authmethods/authmethod_service_test.go
```diff
@@ -908,6 +908,8 @@ func TestCreate(t *testing.T) {
 	}
 }
 
+
```
jimlambrt: ty!

jimlambrt (approved):
Looks great!

PR #1052, opened 2021-03-31
Ongoing OIDC: store DisableDiscoveredConfigValidation

jimlambrt:


internal/auth/oidc/repository_auth_method_update.go
```diff
@@ -208,14 +212,15 @@ func (r *Repository) UpdateAuthMethod(ctx context.Context, am *AuthMethod, versi
 				// method's version.
 				updatedAm = am.Clone()
 				updatedAm.Version = uint32(version) + 1
-				rowsUpdated, err = w.Update(ctx, updatedAm, []string{"Version"}, nil, db.NewOplogMsg(&authMethodOplogMsg), db.WithVersion(&version))
+				rowsUpdated, err = w.Update(ctx, updatedAm, []string{"Version", "DisableDiscoveredConfigValidation"}, nil, db.NewOplogMsg(&authMethodOplogMsg), db.WithVersion(&version))
```
talanknight, line 215: Could you have a test which:
1. Updates with force
2. Verifies a lookup on the authmethod has DisableDiscoveryConfigValidation set
3. Updating again without setting force
4. a lookup on the authmethod no longer has that value set.

talanknight (approved):

internal/proto/local/controller/storage/auth/oidc/store/v1/oidc.proto
```diff
@@ -41,21 +41,28 @@ message AuthMethod {
...
+  // indicates the AuthMethod config was not validated against the IdP's
+  // discovery info document when it was updated or its operational_state was
+  // changed.
+  // @inject_tag: `gorm:"not_null"`
+  bool disable_discovered_config_validation = 89;
+
   // discovery_url is the OIDC Discovery URL without any .well-known component
   // @inject_tag: `gorm:"not_null"`
-  string discovery_url = 90 [(custom_options.v1.mask_mapping) = {this:"DiscoveryUrl" that:"attributes.issuer"}];
+  string discovery_url = 90 [(custom_options.v1.mask_mapping) = { this: "DiscoveryUrl" that: "attributes.issuer" }];
```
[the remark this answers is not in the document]
jefferai: I don't have a strong opinion as to spaces or not. What I was changing (opportunistically) in 1051 is that `this` didn't have a space after the colon but `that` did. The inconsistency within the same block bugged me :-D

jefferai:
Looks good to me, although I think Todd's test suggestion is a good one.

I _believe_ based on the changes here that there are actually no further updates in the handler, so long as it sets WithForce when appropriate -- that is, that I don't need to explicitly set the update mask if it was not manually set in the `false` case, because your changes in this PR will have that same effect further in the DB layer. Confirm/deny?

internal/auth/oidc/repository_auth_method_update.go
jimlambrt: done and ty!  While adding the test I found a few other issues.  The biggest one: proto.Clone(...) doesn't appear to properly clone booleans.  See the new commits.

internal/auth/oidc/repository_auth_method_operational_state.go
```diff
@@ -94,13 +94,44 @@ func (r *Repository) transitionAuthMethodTo(ctx context.Context, authMethodId st
...
+			case err == nil && rowsUpdated > 1:
 				return errors.New(errors.MultipleRecords, op, fmt.Sprintf("updated auth method and %d rows updated", rowsUpdated))
+			case err == nil && rowsUpdated == 0:
+				// this is different than how "no rows updated" is handled in
+				// the typical update pattern since we are not returning the
+				// number of rows updated, we need to raise an error here.
+				return errors.New(errors.RecordNotFound, op, fmt.Sprintf("updated auth method and %d rows updated", rowsUpdated))
+			default:
+			}
+			// if err != nil {
```
jefferai, line 115: Dead code?

jimlambrt: yes. ty!

PR #1053, opened 2021-03-31
Separate out password/oidc code into separate files, and update some validation slightly

jefferai:


jefferai:
@talanknight Checked with @jimlambrt and merging this as it's nearly entirely moving code blocks around to make it easier to find password/oidc specific stuff.

PR #1054, opened 2021-03-31
Update fields in Authenticate

jefferai:
This changes AuthenticateResponse to contain a command field and
attributes field similar to AuthenticateRequest. This is in support of
authenticate responses other than for the `login` command.

It then updates:

* The service handler to properly encode token values for password login
  commands
* The intercept handler to ensure cookies are handled properly
* The Authenticate SDK command to return the command output instead of
  the output of a token read. In order to make decoding of the result
  easier into concrete types it also caches the raw JSON attributes
  field so it can be unmarshaled into the correct type; see the
  adjustements to tests and the `password` authenticate command.
* Various tests


internal/proto/local/controller/api/services/v1/auth_method_service.proto
```diff
@@ -211,8 +210,12 @@ message AuthenticateRequest {
 }
 
 message AuthenticateResponse {
-  resources.authtokens.v1.AuthToken item = 1;
-  string token_type = 2 [json_name="token_type"];
+  reserved 1, 2; // Old item and token_type
+  reserved "item", "token_type";
+  // Valid keys and values depend on the type of Auth Method as well as the command.
+  google.protobuf.Struct attributes = 4 [json_name="attributes"];
```
talanknight, line 216: I wonder if you want to do a similar thing you did with PasswordCredentials for `PasswordResponseAttributes` to document what the structure of a password specific AuthenticateResponse attributes is.  This will be helpful with tests, the outgoing interceptor, and for anyone looking at the code trying to figure out what should be in the response.  It might also be helpful for future plugins to have a way to document what the expected request and response look like.

internal/proto/local/controller/api/services/v1/auth_method_service.proto
```diff
@@ -211,8 +210,12 @@ message AuthenticateRequest {
 }
 
 message AuthenticateResponse {
-  resources.authtokens.v1.AuthToken item = 1;
-  string token_type = 2 [json_name="token_type"];
+  reserved 1, 2; // Old item and token_type
```
talanknight, line 213: nit: Since we only expose this API through JSON, I don't think we need to reserve the numbers yet, but not really important.

talanknight (approved):

internal/servers/controller/handlers/authmethods/password.go
```diff
@@ -140,6 +166,34 @@ func (s Service) authenticateWithPwRepo(ctx context.Context, scopeId, authMethod
 	return prot, nil
 }
 
+func validateAuthenticatePasswordRequest(req *pbs.AuthenticateRequest) error {
```
jimlambrt, line 169: perhaps check for nil req?

internal/servers/controller/handlers/authmethods/password.go
```diff
@@ -140,6 +166,34 @@ func (s Service) authenticateWithPwRepo(ctx context.Context, scopeId, authMethod
...
+	badFields := make(map[string]string)
+
+	attrs := req.GetAttributes().GetFields()
+	if _, ok := attrs[loginNameField]; !ok {
```
jimlambrt, line 173: it might be nice to define consts for these keys.

internal/servers/controller/handlers/outgoing_interceptor.go
```diff
@@ -18,9 +20,25 @@ func OutgoingInterceptor(ctx context.Context, w http.ResponseWriter, m proto.Mes
...
+			// do if there are no attributes
+			return nil
+		}
+		aToken := &authtokenpb.AuthToken{}
+		// We may have "token_type" if it's a token, or it may not be a token at
+		// all, so ignore unknown fields
+		if err := StructToProto(m.GetAttributes(), aToken, WithDiscardUnknownFields(true)); err != nil {
+			return err
+		}
+		tokenType := m.GetAttributes().GetFields()["token_type"].GetStringValue()
```
jimlambrt, line 34: perhaps a const for the field name?

internal/servers/controller/handlers/outgoing_interceptor.go
```diff
@@ -18,9 +20,25 @@ func OutgoingInterceptor(ctx context.Context, w http.ResponseWriter, m proto.Mes
 	m = m.ProtoReflect().Interface()
```
jimlambrt, line 20: nil checks?

internal/servers/controller/handlers/subtypes.go
```diff
@@ -6,12 +6,18 @@ import (
...
-func StructToProto(fields *structpb.Struct, p proto.Message) error {
+func StructToProto(fields *structpb.Struct, p proto.Message, opt ...Option) error {
 	js, err := fields.MarshalJSON()
 	if err != nil {
 		return err
 	}
-	if err := protojson.Unmarshal(js, p); err != nil {
+	opts := getOpts(opt...)
+	if opts.withDiscardUnknownFields {
+		err = (protojson.UnmarshalOptions{DiscardUnknown: true}.Unmarshal(js, p))
```
jimlambrt, line 16: TIL.  ty!

internal/servers/controller/handlers/subtypes.go
```diff
@@ -6,12 +6,18 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
-func StructToProto(fields *structpb.Struct, p proto.Message) error {
+func StructToProto(fields *structpb.Struct, p proto.Message, opt ...Option) error {
```
jimlambrt, line 9: nil checks?

internal/servers/controller/testing.go
```diff
@@ -128,7 +129,7 @@ func (tc *TestController) Token() *authtokens.AuthToken {
 		tc.t.Error("no default auth method ID configured")
```
jimlambrt, line 129: consider calling tc.t.Helper() in the receiver functions.

api/authmethods/authenticate.go
```diff
@@ -2,12 +2,62 @@ package authmethods
...
+func (a *AuthenticateResult) UnmarshalJSON(inBytes []byte) error {
+	type in struct {
+		Command    string          `json:"command"`
+		Attributes json.RawMessage `json:"attributes"`
+	}
+	i := new(in)
+	if err := json.Unmarshal(inBytes, i); err != nil {
+		return err
+	}
+	a.Command = i.Command
```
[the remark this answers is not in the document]
jefferai: Maybe, but not in the UnmarshalJSON function. Empty in should be empty out.

api/authmethods/authenticate.go
```diff
@@ -2,12 +2,62 @@ package authmethods
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 
-	"github.com/hashicorp/boundary/api/authtokens"
+	"github.com/hashicorp/boundary/api"
 )
```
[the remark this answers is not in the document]
jefferai: Adding it here would make it an outlier as we don't have godoc comments on any of the types that 1:1 match our API. Whether we should is another question, but not something I want to add to all types in this PR.

api/authmethods/authenticate.go
```diff
@@ -2,12 +2,62 @@ package authmethods
...
 
-func (c *Client) Authenticate(ctx context.Context, authMethodId, command string, attributes map[string]interface{}, opt ...Option) (*authtokens.AuthTokenReadResult, error) {
+type AuthenticateResult struct {
+	Command       string                 `json:"-"`
+	Attributes    map[string]interface{} `json:"-"`
+	attributesRaw json.RawMessage        `json:"-"`
```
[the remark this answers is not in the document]
jefferai: Fixed

api/authmethods/authenticate.go
```diff
@@ -29,15 +79,15 @@ func (c *Client) Authenticate(ctx context.Context, authMethodId, command string,
 		return nil, fmt.Errorf("error performing client request during Authenticate call: %w", err)
 	}
 
-	target := new(authtokens.AuthTokenReadResult)
-	target.Item = new(authtokens.AuthToken)
-	apiErr, err := resp.Decode(target.Item)
+	target := new(AuthenticateResult)
```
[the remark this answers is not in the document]
jefferai: Maybe, but this is the current pattern used everywhere in the API.

api/authmethods/authenticate.go
```diff
@@ -2,12 +2,62 @@ package authmethods
...
+func (n AuthenticateResult) GetRawAttributes() json.RawMessage {
+	return n.attributesRaw
+}
+
+func (n AuthenticateResult) GetResponse() *api.Response {
+	return n.response
+}
+
+// Authenticate is a generic authenticate API call that returns a generic result.
+func (c *Client) Authenticate(ctx context.Context, authMethodId, command string, attributes map[string]interface{}, opt ...Option) (*AuthenticateResult, error) {
```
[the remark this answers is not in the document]
jefferai: Done

internal/cmd/commands/authenticate/password.go
```diff
@@ -91,6 +90,19 @@ func (c *PasswordCommand) AutocompleteFlags() complete.Flags {
 	return c.Flags().Completions()
 }
 
+type dummyGenericResponse struct {
+	item     interface{}
+	response *api.Response
+}
+
```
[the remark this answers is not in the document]
jefferai: Done.

internal/proto/local/controller/api/services/v1/auth_method_service.proto
jefferai: `buf` complains if these aren't marked as reserved since they were previously used.

internal/proto/local/controller/api/services/v1/auth_method_service.proto
jefferai: The structure of the password specific authenticate response attributes is already defined -- it's an AuthToken.

I think we could potentially add more of these in the future but I'd prefer to defer for now.

internal/servers/controller/handlers/authmethods/authmethod_service.go
```diff
@@ -279,23 +280,23 @@ func (s Service) Authenticate(ctx context.Context, req *pbs.AuthenticateRequest)
...
-			ScopeId: authResults.Scope.Id,
-			Type:    resource.AuthToken,
-		}
-		tok.AuthorizedActions = authResults.FetchActionSetForId(ctx, tok.Id, authtokens.IdActions, auth.WithResource(res)).Strings()
-		return &pbs.AuthenticateResponse{Item: tok, TokenType: req.GetTokenType()}, nil
+		return s.authenticatePassword(ctx, req, &authResults)
+
 	case auth.OidcSubtype:
 		return &pbs.AuthenticateResponse{}, nil
 	}
```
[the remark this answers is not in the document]
jefferai: Which message/error?

internal/servers/controller/handlers/authmethods/authmethod_service.go
```diff
@@ -893,16 +894,17 @@ func validateAuthenticateRequest(req *pbs.AuthenticateRequest) error {
...
 
 	if strings.TrimSpace(req.GetAuthMethodId()) == "" {
 		badFields["auth_method_id"] = "This is a required field."
-		return handlers.InvalidArgumentErrorf("Invalid fields provided in request.", badFields)
+	} else {
+		st := auth.SubtypeFromId(req.GetAuthMethodId())
+		switch st {
+		case auth.PasswordSubtype, auth.OidcSubtype:
+		default:
+			badFields["auth_method_id"] = "Unknown auth method type."
```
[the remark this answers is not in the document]
jefferai: Done.

internal/servers/controller/handlers/authmethods/password.go
```diff
@@ -85,6 +91,26 @@ func (s Service) updatePwInRepo(ctx context.Context, scopeId, id string, mask []
 	return out, nil
 }
 
+// Authenticate implements the interface pbs.AuthenticationServiceServer.
+func (s Service) authenticatePassword(ctx context.Context, req *pbs.AuthenticateRequest, authResults *auth.VerifyResults) (*pbs.AuthenticateResponse, error) {
+	reqAttrs := req.GetAttributes().GetFields()
```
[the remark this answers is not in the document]
jefferai: Those are already handled by the validation functions.

internal/servers/controller/handlers/authmethods/password.go
jefferai: Going to defer to some other time

internal/servers/controller/handlers/authmethods/password.go
jefferai: I don't think grpc can ever give us a nil proto. Regardless, it's validated by the first non-password-specific call.

internal/servers/controller/handlers/outgoing_interceptor.go
jefferai: Not sure what the contract is here, so not modifying what's there.

PR #1057, opened 2021-04-01
Add OIDC Start handler function

jefferai:
There are a lot of changes but most of them are reorganizing, so I'll call out what to look at and what to ignore:

* In `service_start_auth.go`, `apiAddr` was removed from the function signature and the appropriate callback URLs were swapped in. In addition I updated some deprecated timestamp calls, and I updated the roundtrip payload functionality to perform urlencoding. It seemed better to let the caller pass in a marshaled JSON string but not have to deal with the urlencoding since they're not generating the url. Associated tests were updated.
* In `common.go` I made the Oidc repo factory function a type alias. This allows using either one to pass into the service handler new function without running afoul of cyclical deps.
* In `authmethod_service.go` I fixed a few bugs(?) where we were calling error functions but not returning them, and I removed a validation check that has been moved into `password.go`'s validation but isn't needed (and is in fact unwanted) for oidc.
* Ignore everything in `authmethod_service_test.go` and `password_test.go` -- these are purely reorganization!
* In `oidc.go` there is added functionality to check
* In `oidc_test.go` the new functions are `getSetup` at the top (I didn't refactor existing tests yet, but this is meant to serve as a way to just get a lot of stuff set up and running for service-level tests without copying boilerplate everywhere...it will also serve as a kind of rubric for when I add it into dev mode for API-level tests); and `TestAuthenticate_OIDC_Start` at the bottom. Everything else was simply moved there as-is and can be ignored.


internal/auth/oidc/service_start_auth.go
```diff
@@ -56,27 +53,29 @@ func StartAuth(ctx context.Context, oidcRepoFn OidcRepoFactory, apiAddr string,
...
 	now := time.Now()
-	createTime, err := ptypes.TimestampProto(now.Truncate(time.Second))
-	if err != nil {
-		return nil, nil, "", errors.New(errors.Unknown, op, "create timestamp failed", errors.WithWrap(err))
-	}
-	exp, err := ptypes.TimestampProto(now.Add(AttemptExpiration).Truncate(time.Second))
-	if err != nil {
-		return nil, nil, "", errors.New(errors.Unknown, op, "exp timestamp failed", errors.WithWrap(err))
-	}
+	createTime := timestamppb.New(now.Truncate(time.Second))
```
jimlambrt, line 77: ty for updating these!

internal/servers/controller/handlers/authmethods/password.go
```diff
@@ -169,6 +168,16 @@ func (s Service) authenticateWithPwRepo(ctx context.Context, scopeId, authMethod
 func validateAuthenticatePasswordRequest(req *pbs.AuthenticateRequest) error {
 	badFields := make(map[string]string)
 
+	if req.GetAttributes() == nil && req.GetCredentials() != nil {
+		// TODO: Eventually, remove this
+		req.Attributes = req.Credentials
+	}
+	if req.GetAttributes() == nil || req.GetAttributes().GetFields() == nil {
+		badFields["attributes"] = "This is a required field."
```
jimlambrt, line 176: might be nice to use const for these repeatedly used keys

jimlambrt (approved):
Looks great!  (Just one tiny suggestion)

internal/servers/controller/handlers/authmethods/oidc.go
```diff
@@ -123,6 +139,99 @@ func (s Service) updateOidcInRepo(ctx context.Context, scopeId string, req *pbs.
...
+			}
+		}
+	case callbackCommand:
+	case tokenCommand:
+		tType := strings.ToLower(strings.TrimSpace(req.GetTokenType()))
+		if tType != "" && tType != "token" && tType != "cookie" {
+			badFields[tokenTypeField] = `The only accepted types are "token" and "cookie".`
+		}
+	default:
+		badFields["command"] = "Invalid command for this auth method type."
```
talanknight, line 226: extract the "command" string into a const.

internal/servers/controller/handlers/authmethods/oidc.go
```diff
@@ -123,6 +139,99 @@ func (s Service) updateOidcInRepo(ctx context.Context, scopeId string, req *pbs.
...
+
+	return &pbs.AuthenticateResponse{Command: req.GetCommand(), Attributes: attrs}, nil
+}
+
+func validateAuthenticateOidcRequest(req *pbs.AuthenticateRequest) error {
+	badFields := make(map[string]string)
+
+	switch req.GetCommand() {
+	case startCommand:
+		if req.GetAttributes() != nil && req.GetAttributes().GetFields() != nil {
```
talanknight, line 202: I would imagine a good validation of an authenticate request is that the attributes are able to be converted to the proto representation (which validates that all the fields passed in are recognizable).  That would make it easier to validate the received values.

talanknight (commented):
LGTM.  Using the handlers.StructToProto helper function is the main change I would make to both verify the fields in the request are expected and to do verification w/o having to use string based keys into the fields map.

internal/servers/controller/handlers/authmethods/oidc.go
jefferai: Done.

internal/servers/controller/handlers/authmethods/oidc.go
jefferai: There isn't a proto for this. We could make one, but I'm not sure that's the right thing to do, because this field is defined as an arbitrary struct. So you'd just convert the struct into a proto containing.....a struct, and you'd have the same logic one level down.

It's also not clear to me that ensuring all fields passed in are recognizable is a good thing. It's nice to not accept fields if we don't know about them. On the other hand, not doing so can create complex upgrading issues later on due to breaking clients that pass parameters that older servers don't know about.

PR #1058, opened 2021-04-02
refactor to use WithOrderByCreateTime(...)

jimlambrt:


internal/session/options.go
```diff
@@ -18,14 +18,15 @@ type Option func(*options)
...
-	withLimit          int
-	withOrder          string
-	withScopeIds       []string
-	withUserId         string
-	withExpirationTime *timestamp.Timestamp
-	withTestTofu       []byte
-	withListingConvert bool
-	withSessionIds     []string
+	withLimit             int
+	withOrderByCreateTime bool
```
talanknight, line 22: Could we convert these 2 fields into enums?  That could allow multiple columns to order by and also make it explicit if it is in a descending or ascending order.  For example if someone currently passed in WithOrderByCreateTime(true), WithOrderByVersion(false), when checking the set options both withOrderByCreateTime would be set as well as withOrderByVersion and the value would be false.  Changing this to be a single field in the options struct with the OrderBy enum type the ambiguity goes away.

jimlambrt: there is no ambiguity yet…. WithOrderByVersion(…) and WithOrderByCreateTime(…) are in different packages.  With that said, I'm pondering if I should make a change based on your suggestion.

jimlambrt: updated to use db.WithOrderBy enum.  Ty!

talanknight (approved):

PR #1062, opened 2021-04-02
OIDC dev auth method + callback handler

jefferai:


internal/cmd/commands/dev/dev.go
```diff
@@ -532,3 +539,27 @@ func (c *Command) Run(args []string) int {
...
+	testProvider                *capoidc.TestProvider
+	testProviderAlg             capoidc.Alg
+	testProviderCaCert          []*x509.Certificate
+	testController              *httptest.Server
+	authMethod                  *oidc.AuthMethod
+	testProviderAllowedRedirect string
+}
+
+func (c *Command) startOidcTestProvider() error {
+	/*
```
jimlambrt, line 553: for an example, see: https://github.com/hashicorp/cap/blob/935c659d2806460da893885239f1257d41d911fe/oidc/examples/cli/main.go#L125

jimlambrt (commented):
This all looks great as an approach.

internal/servers/controller/handlers/authmethods/oidc.go
```diff
@@ -143,14 +138,16 @@ func (s Service) updateOidcInRepo(ctx context.Context, scopeId string, req *pbs.
 func (s Service) authenticateOidc(ctx context.Context, req *pbs.AuthenticateRequest, authResults *auth.VerifyResults) (*pbs.AuthenticateResponse, error) {
 	const op = "authmethod_service.(Service).authenticateOidc"
 	if req == nil {
-		return nil, errors.New(errors.InvalidParameter, op, "nil request")
```
jefferai, line 146: @jimlambrt I think I am confused about when errors are "human friendly" vs. not...I think I had an idea that these new error types/schemas were all human friendly but that isn't necessarily the case. In this function though we pass back what comes back from the function directly to the response handler so I think here it's correct to use human-readable ones.

internal/auth/oidc/repository_auth_method_create.go
```diff
@@ -32,11 +33,19 @@ func (r *Repository) CreateAuthMethod(ctx context.Context, am *AuthMethod, _ ...
...
+	opts := getOpts(opt...)
+	am.PublicId = opts.withPublicId
+	if am.PublicId == "" {
+		id, err := newAuthMethodId()
+		if err != nil {
+			return nil, errors.Wrap(err, op)
+		}
+		am.PublicId = id
+	} else {
+		if !handlers.ValidId(handlers.Id(am.PublicId), AuthMethodPrefix) {
```
jimlambrt, line 45: this violates the layers of our system... you're calling an upper layer from the repo.  we need to refactor this func into a more util type thing or just dup the code here.

internal/auth/oidc/auth_method.go
```diff
@@ -77,7 +77,7 @@ func NewAuthMethod(scopeId string, clientId string, clientSecret ClientSecret, o
 			ScopeId:          scopeId,
 			Name:             opts.withName,
 			Description:      opts.withDescription,
-			OperationalState: string(InactiveState),
+			OperationalState: string(opts.withOperationalState),
```
jimlambrt, line 80: I had intentionally not allowed this... we allow "incomplete" authmethods to be created and these incomplete methods should stay inactive until they are complete.   If we want to support setting the state to anything but "inactive" then we need to ensure it isCompete() and if not raise an error.

internal/cmd/commands/authenticate/password.go
```diff
@@ -228,11 +227,20 @@ func (c *PasswordCommand) Run(args []string) int {
 					break
 				}
 			}
+
+			if !gotErr {
+				c.UI.Output("\nThe token was successfully stored in the chosen keyring and is not displayed here.")
```
jimlambrt, line 232: I love this.

jimlambrt (changes requested):
Looks great.  Couple minor things.

internal/cmd/commands/authenticate/password.go
jefferai: :-)

website/content/docs/concepts/security/permissions.mdx
```diff
@@ -249,86 +245,62 @@ wildcard or templated grant strings.
...
-                id=&lt;pin&gt;;type=&lt;type&gt;;actions=change-password
-              </code>
-            </li>
-          </ul>
+            <ul>
+              <li><code>id=&lt;id&gt;;actions=delete</code></li>
+              <li><code>id=&lt;pin&gt;;type=&lt;type&gt;;actions=delete</code></li>
+            </ul>
+          <li>
+            <code>set-password</code>: Set a password on an account, without requring the current password
```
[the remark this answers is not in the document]
jefferai: Heh, whitespace changes for the win. Fixed.

internal/auth/oidc/repository_auth_method_create.go
jefferai: Done.

internal/auth/oidc/auth_method.go
jefferai: Fixed.

jimlambrt (approved):

PR #1064, opened 2021-04-02
CallbackUrl  value object changed to ApiUrl field, Remove ID suffix to storage fields

talanknight:
Renamed issuer_id to issuer, subject_id to subject and changed the callback url value object to a single field on an auth method.


internal/auth/oidc/repository_auth_method_update.go
```diff
@@ -200,7 +192,7 @@ func (r *Repository) UpdateAuthMethod(ctx context.Context, am *AuthMethod, versi
 		db.StdRetryCnt,
 		db.ExpBackoff{},
 		func(reader db.Reader, w db.Writer) error {
-			msgs := make([]*oplog.Message, 0, 9) // AuthMethod, Algs*2, Certs*2, Callbacks*2, Audiences*2
+			msgs := make([]*oplog.Message, 0, 9) // AuthMethod, Algs*2, Certs*2, ApiUrl*2, Audiences*2
```
jimlambrt, line 195: I think this is 7 now... the ApiUrl is part of the auth method.

internal/db/schema/migrations/postgres/2/84_oidc.up.sql
```diff
@@ -217,27 +202,26 @@ create or replace function
...
 as $$
   begin
-    -- validate callback and signing alg
+    -- validate signing alg
     if old.state = 'inactive' and new.state != 'inactive' then
       perform 
       from 
         auth_oidc_method am
-       join auth_oidc_callback_url  cb    on am.public_id = cb.oidc_method_id 
        join auth_oidc_signing_alg   alg   on am.public_id = alg.oidc_method_id
```
jimlambrt, line 210: odd spacing now that there's only one join table there's no need to align them with spaces.

internal/db/schema/migrations/postgres/2/84_oidc.up.sql
```diff
@@ -364,33 +348,30 @@ before insert on auth_oidc_account
...
-declare cb_cnt int;
   begin
     select 
       am.state,
-      count(alg.oidc_method_id) as alg_cnt,
-      count(cb.oidc_method_id) as cb_cnt
+      count(alg.oidc_method_id) as alg_cnt
     from 
       auth_oidc_method am
       left outer join auth_oidc_signing_alg   alg   on am.public_id = alg.oidc_method_id
```
jimlambrt, line 370: same odd spacing as above.

internal/servers/controller/handlers/authmethods/authmethod_service.go
```diff
@@ -734,10 +734,12 @@ func validateCreateRequest(req *pbs.CreateAuthMethodRequest) error {
 						}
 					}
 				}
-				if attrs.GetApiUrlPrefix() != nil {
+				if strings.TrimSpace(attrs.GetApiUrlPrefix().GetValue()) == "" {
+					// TODO: When we start accepting the address used in the request make this an optional field.
```
jimlambrt, line 738: ty for the TODO

jimlambrt (approved):
Just a few small suggestions.

internal/servers/controller/handlers/authmethods/authmethod_service.go
talanknight: Your welcome.  Do you think we need to have this check in more places than just the API level?

PR #1065, opened 2021-04-02
Vault integration: add CredentialLibrary resources

mgaffney:
Structs added:
* `vault.CredentialLibrary`
* `vault.Lease`

### DIFF
--- a/internal/credential/vault/vault_lease.go
+++ b/internal/credential/vault/vault_lease.go
@@ -1,0 +1,102 @@
+package vault
+
+import (
+	"time"
+
+	"github.com/hashicorp/boundary/internal/credential/vault/store"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/oplog"
+	"google.golang.org/protobuf/proto"
+)
+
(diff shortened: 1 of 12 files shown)


internal/credential/vault/vault_lease.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// SetTableName sets the table name.
+func (l *Lease) SetTableName(n string) {
+	l.tableName = n
+}
+
+func (l *Lease) oplog(op oplog.OpType) oplog.Metadata {
+	metadata := oplog.Metadata{
+		"resource-type": []string{"credential-vault-lease"},
```
louisruch, line 78: Should you not have the `resource-public-id` added to the metadata?

louisruch (approved):
One question about the lease oplog metadata otherwise LGTM

internal/credential/vault/vault_lease.go
mgaffney: Which `resource-public-id`?

mgaffney: Never mind. I see what you mean. Ty!

PR #1066, opened 2021-04-03
fix scopes double import

alrs:
This fixes a double import in `internal/servers/controller/handlers/scopes`.


jefferai (approved):
Thanks!

PR #1068, opened 2021-04-05
Bump to Go 1.15.11

jefferai:
This has some important fixes, including for a potential SQL transaction
deadlock.


jefferai:
Merging as per internal approval

PR #1069, opened 2021-04-05
Add OIDC Auth Method and Account CLI commands

talanknight:
Note: This is a PR against my other open renaming PR's branch since that PR changes a few request requirements (requires api url to be defined at auth method request, issuer to be set on the auth method or account time, etc...)

- Added the already written oidc auth method command into the list of boundary cli commands. 
- Renamed the ca-cert field to `idp-ca-cert` to avoid collision with another flag.
- Made an `attributes` not being included in the request not be an error but be treated as the default subtype specific attribute proto message.


jimlambrt:
I just noticed that we're not allowing the operators to set the FullName and Email for oidc accounts.  Can we add that?

jimlambrt (changes requested):
Looks great, except for the inability to provide a full name and email on the odic accounts.

jimlambrt (approved):

PR #1070, opened 2021-04-05
Refactor  internal/servers/controller/handlers errors

s-christoff:


internal/servers/controller/handlers/authtokens/authtoken_service.go
```diff
@@ -143,34 +144,36 @@ func (s Service) DeleteAuthToken(ctx context.Context, req *pbs.DeleteAuthTokenRe
...
 	repo, err := s.repoFn()
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, op)
 	}
 	u, err := repo.LookupAuthToken(ctx, id)
 	if err != nil {
 		if errors.IsNotFoundError(err) {
-			return nil, handlers.NotFoundErrorf("AuthToken %q doesn't exist.", id)
+			return nil, nil
```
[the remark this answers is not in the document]
louisruch: Hmm i see we are checking the `u == nil` case below as well.  We could refactor this into something like:
```
u, err := repo.LookupAuthToken(ctx, id)
if err != nil && !errors.IsNotFoundError(err) {
     return nil, errors.Wrap(err, op)
}
if u == nil {
...
}
```

internal/servers/controller/handlers/groups/group_service.go
```diff
@@ -247,24 +248,26 @@ func (s Service) RemoveGroupMembers(ctx context.Context, req *pbs.RemoveGroupMem
...
 	repo, err := s.repoFn()
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, op)
 	}
 	g, m, err := repo.LookupGroup(ctx, id)
 	if err != nil {
 		if errors.IsNotFoundError(err) {
-			return nil, handlers.NotFoundErrorf("Group %q doesn't exist.", id)
+			return nil, nil
```
louisruch, line 259: same comment as above

internal/servers/controller/handlers/groups/group_service.go
```diff
@@ -247,24 +248,26 @@ func (s Service) RemoveGroupMembers(ctx context.Context, req *pbs.RemoveGroupMem
...
 		}
-		return nil, fmt.Errorf("unable to get group: %w", err)
+		return nil, errors.Wrap(err, op)
 	}
 	if g == nil {
-		return nil, handlers.NotFoundErrorf("Group %q doesn't exist.", id)
+		return nil, errors.New(errors.InvalidParameter, op, fmt.Sprintf("group %q not found", id))
```
louisruch, line 264: In this PR https://github.com/hashicorp/boundary/pull/1077 we upper cased error messages that would end up as the first message to the user

internal/servers/controller/handlers/groups/group_service.go
```diff
@@ -291,6 +294,7 @@ func (s Service) createInRepo(ctx context.Context, scopeId string, item *pb.Grou
 }
 
 func (s Service) updateInRepo(ctx context.Context, scopeId, id string, mask []string, item *pb.Group) (*pb.Group, error) {
+	const op = "groups.(Service).createInRepo"
```
louisruch, line 297: `s/createInRepo/updateInRepo`

internal/servers/controller/handlers/host_catalogs/host_catalog_service.go
```diff
@@ -285,23 +287,15 @@ func (s Service) createInRepo(ctx context.Context, projId string, item *pb.HostC
...
+		return nil, errors.Wrap(err, op)
 	}
 	out, err := repo.CreateCatalog(ctx, h)
 	if err != nil {
-		if e := errors.Convert(err); e != nil {
-			// This is a domain error, push this error through so the error interceptor can interpret it correctly.
-			return nil, e
-		}
-		return nil, fmt.Errorf("unable to create host catalog: %w", err)
+		return nil, errors.Wrap(err, op, errors.WithMsg("unable to create host catalog"))
```
louisruch, line 298: Hmm I see we were not very consistent with uppercase previously

internal/servers/controller/handlers/host_sets/host_set_service.go
```diff
@@ -61,8 +61,9 @@ var _ pbs.HostSetServiceServer = Service{}
 // NewService returns a host set Service which handles host set related requests to boundary and uses the provided
 // repositories for storage and retrieval.
 func NewService(repoFn common.StaticRepoFactory) (Service, error) {
+	const op = "host_sets.NewService"
 	if repoFn == nil {
-		return Service{}, fmt.Errorf("nil static repository provided")
+		return Service{}, errors.New(errors.InvalidParameter, op, "missing statis repository")
```
louisruch, line 66: `s/statis/static`

internal/servers/controller/handlers/hosts/host_service.go
```diff
@@ -189,6 +190,7 @@ func (s Service) getFromRepo(ctx context.Context, id string) (*pb.Host, error) {
 }
 
 func (s Service) createInRepo(ctx context.Context, scopeId, catalogId string, item *pb.Host) (*pb.Host, error) {
+	const op = "host_service.(Service).createInRepo"
```
louisruch, line 193: `s/host_service/hosts`

internal/servers/controller/handlers/mask_manager.go
```diff
@@ -59,7 +63,8 @@ func mapFromProto(ps ...protoreflect.ProtoMessage) (map[string]string, error) {
 			opts := f.Options().(*descriptorpb.FieldOptions)
 			if nameMap := proto.GetExtension(opts, pb.E_MaskMapping).(*pb.MaskMapping); !proto.Equal(nameMap, &pb.MaskMapping{}) && nameMap != nil {
 				if _, ok := mapping[nameMap.GetThis()]; ok {
-					return nil, fmt.Errorf("duplicate mapping from field %q with the mapping key %q", f.Name(), nameMap.GetThis())
+					//todo(schristoff): I'm not in love with this error either
+					return nil, errors.New(errors.InvalidParameter, op, fmt.Sprintf("duplicate mapping from field %q with the mapping key %q", f.Name(), nameMap.GetThis()))
```
louisruch, line 67: I am ok with this error message

internal/servers/controller/handlers/mask_manager.go
```diff
@@ -30,26 +32,28 @@ func NewMaskManager(dest protoreflect.ProtoMessage, src ...protoreflect.ProtoMes
 	for k, v := range srcToDest {
 		ov, ok := destToSrc[v]
 		if !ok || ov != k {
-			return nil, fmt.Errorf("mapping src field %q maps to %q, dest %q maps to %q", k, v, v, ov)
+			return nil, errors.New(errors.Encrypt, op, fmt.Sprintf("mapping src field %q maps to %q, dest %q maps to %q", k, v, v, ov))
```
louisruch, line 35: `Encode` might be more appropriate than `Encrypt` but its a strange one...

internal/servers/controller/handlers/mask_manager.go
```diff
@@ -30,26 +32,28 @@ func NewMaskManager(dest protoreflect.ProtoMessage, src ...protoreflect.ProtoMes
...
 		}
 		result[k] = v
 	}
 
 	// Now check to make sure there aren't any dangling dest mappings.
 	for k, v := range destToSrc {
 		if ov, ok := srcToDest[v]; !ok || ov != k {
-			return nil, fmt.Errorf("mapping dest field %q maps to %q, src %q maps to %q", k, v, v, ov)
+			//todo(schristoff): I'm not in love with this error
+			return nil, errors.New(errors.Encrypt, op, fmt.Sprintf("mapping src field %q maps to %q, dest %q maps to %q", k, v, v, ov))
```
louisruch, line 44: So I agree, but not sure I have a great alternative.  I have had to read this a couple of times to figure it out.  

Also I am not a fan of the `ov` naming here as it creates confusion around `ok` - `'ok` here is the map lookup bool and meant to be`okay` not `o-k` like `o-v` is trying to be.  My first read of this was that `ok` was the second key like `ov` is the second value...

louisruch (commented):
This looks good, I have some comments and fixes.  I am not sure what we should do re: capitalizing first letters of error messages.  It seems we have been fairly inconsistent with this, and its hard to know if this is truly the final step before the end user sees the message at this point in the stack...

internal/servers/controller/handlers/authtokens/authtoken_service.go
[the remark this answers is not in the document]
s-christoff: tbh, some of these I saw and **do not** remember changing but I left them in because 🤷🏻‍♀️ 
This is one of those ones (also the one above this comment)

internal/iam/repository_group.go
```diff
@@ -230,7 +230,7 @@ func (r *Repository) AddGroupMembers(ctx context.Context, groupId string, groupV
 
 	oplogWrapper, err := r.kms.GetWrapper(ctx, scope.GetPublicId(), kms.KeyPurposeOplog)
 	if err != nil {
-		return nil, errors.Wrap(err, op, errors.WithMsg("unable to get oplog wrapper"))
```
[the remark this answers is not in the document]
s-christoff: oh ok, was following form with previous ways I did this (which means maybe I should go find those not-so-good ways) :(

internal/servers/controller/handlers/mask_manager.go
s-christoff: yes I had much trouble reading this as well

louisruch (approved):
LGTM 🥇

PR #1071, opened 2021-04-06
Ongoing OIDC: authenticate token request handler

jimlambrt:
@jefferai - I'm unsure about the error handling from the oidc Repository.TokenRequest(...) call.


internal/servers/controller/handlers/authmethods/oidc.go
```diff
@@ -196,6 +201,54 @@ func (s Service) authenticateOidcStart(ctx context.Context, req *pbs.Authenticat
...
+	if err := handlers.StructToProto(req.GetAttributes(), attrs); err != nil {
+		return nil, errors.New(errors.InvalidParameter, op, "error parsing request attributes")
+	}
+	if attrs.TokenId == "" {
+		return nil, errors.New(errors.InvalidParameter, op, "empty token id request attributes")
+	}
+
+	token, err := oidc.TokenRequest(ctx, s.kms, s.atRepoFn, attrs.TokenId)
+	if err != nil {
+		fmt.Println(err)
```
talanknight, line 226: Left over debug line?  Delete.

internal/servers/controller/handlers/authmethods/oidc.go
```diff
@@ -196,6 +201,54 @@ func (s Service) authenticateOidcStart(ctx context.Context, req *pbs.Authenticat
...
+		// TODO: Log something so we don't lose the error's context and entire msg...
+		switch {
+		case errors.Match(errors.T(errors.Forbidden), err):
+			return nil, errors.Wrap(err, op, errors.WithMsg("Forbidden"))
+		case errors.Match(errors.T(errors.AuthAttemptExpired), err):
+			return nil, errors.Wrap(err, op, errors.WithMsg("Forbidden"))
+		case errors.Match(errors.T(errors.InvalidParameter), err):
+			return nil, errors.New(errors.Internal, op, "Error invalid parameters for OIDC token request.")
+		default:
+			return nil, errors.New(errors.Internal, op, "Error during OIDC token request.")
```
talanknight, line 236: Why don't we want to wrap the errors for InvalidParam or this case?

internal/servers/controller/handlers/authmethods/oidc.go
```diff
@@ -196,6 +201,54 @@ func (s Service) authenticateOidcStart(ctx context.Context, req *pbs.Authenticat
...
+		}
+	}
+
+	attrsMap := map[string]interface{}{
```
talanknight, line 240: With password auth methods we can return the tokens as either a token on the response or as a split cookie.  This is determined based on the `token_type` value in the initial authentication request.

This is done by populating `token_type` in the response and letting internal/servers/controller/handlers/outgoing_interceptor.go intercept the response.

talanknight (changes requested):

internal/servers/controller/handlers/authmethods/oidc.go
jimlambrt: ty!

internal/servers/controller/handlers/authmethods/oidc.go
jimlambrt: changed to wrapping the error. ty!

internal/servers/controller/handlers/authmethods/oidc.go
jimlambrt: Please let me know if I've done this correctly.

talanknight: @randallmorey Is it in the final `token` step of oidc authentication that FE is expected to provide the token type field?  If so, @jimlambrt , this looks right.  Thanks! :)

internal/proto/local/controller/api/services/v1/auth_method_service.proto
```diff
@@ -190,52 +189,65 @@ message ChangeStateResponse {
...
 message OidcStartAttributes {
   // An object which will be marshaled as JSON and roundtripped in the token command call.
-  google.protobuf.Struct roundtrip_payload = 1 [json_name="roundtrip_payload"];
+  google.protobuf.Struct roundtrip_payload = 1 [json_name = "roundtrip_payload"];
   // Cached marshaled payload. This is not ingressed from the client; anything found will be thrown out.
   string cached_roundtrip_payload = 2;
 }
 
+// The layout of the struct for "attributes" field in AuthenticateRequest for a
+// oidc type's token command. This message isn't directly referenced anywhere
```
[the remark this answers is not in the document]
randallmorey: I can see the value of prefixing the command with type.  Might help developers reading the code.

internal/servers/controller/handlers/authmethods/oidc.go
randallmorey: Yes `token_type` is planned for the `token` step.

talanknight (approved):

PR #1075, opened 2021-04-06
Check option is not nil before applying it

talanknight:


jimlambrt:
Just checking: we never want to support the ability to override a default that's not nil to nil?

jimlambrt (approved):

PR #1076, opened 2021-04-07
Add is_primary field to auth methods

talanknight:


internal/servers/controller/handlers/authmethods/authmethod_service.go
```diff
@@ -40,6 +40,7 @@ const (
 	attributesField   = "attributes"
 	authMethodIdField = "auth_method_id"
 	tokenTypeField    = "token_type"
+	isPrimaryField = "is_primary"
```
jimlambrt, line 43: odd formatting.

jimlambrt (approved):
Just one little formatting suggestion.  I assume the next PR will include the CLI bits for this new field?

PR #1085, opened 2021-04-08
Add API/CLI support for fetching token results

jefferai:


internal/cmd/commands/authenticate/oidc.go
```diff
@@ -103,87 +106,49 @@ func (c *OidcCommand) Run(args []string) int {
...
+	go func() {
+		defer wg.Done()
+		for {
+			select {
+			case <-c.Context.Done():
+				c.PrintCliError(errors.New("Command canceled."))
+				watchCode = base.CommandCliError
+				return
+
+			case <-time.After(time.Second):
```
jimlambrt, line 121: Hum... give what it will take to authen at the IdP, every second seems a bit aggressive.  Hum... but I get the user experience is degraded if we wait too long.  How about 2s?

jimlambrt (approved):
Looks great and it works with Okta.  One small question

internal/cmd/commands/authenticate/oidc.go
```diff
@@ -103,87 +106,49 @@ func (c *OidcCommand) Run(args []string) int {
...
+				result, err = authmethods.NewClient(client).Authenticate(c.Context, c.FlagAuthMethodId, "token", map[string]interface{}{
```
talanknight, line 122: Would there be any benefit to creating the auth method client once and sharing it across the different Authenticate calls?

internal/cmd/commands/authenticate/oidc.go
```diff
@@ -103,87 +106,49 @@ func (c *OidcCommand) Run(args []string) int {
 		return base.CommandCliError
```
talanknight, line 106: Not directly relevant to this PR but in the case that the user is trying to authenticate where a browser isn't available on the same platform could we print out the URL and allow them to open it manually?

talanknight (approved):
Looks good!

internal/cmd/commands/authenticate/oidc.go
jefferai: Good point. If we get an error back from the open I think we should print it out. I'll fix that.

internal/cmd/commands/authenticate/oidc.go
jefferai: I thought about that and wasn't sure. But I don't think it will hurt so I'll make that change.

internal/cmd/commands/authenticate/funcs.go
```diff
@@ -0,0 +1,105 @@
...
+		keyringType != "" &&
+		tokenName != "" {
+		marshaled, err := json.Marshal(token)
+		if err != nil {
+			c.UI.Error(fmt.Sprintf("Error marshaling auth token to save to keyring: %s", err))
+			gotErr = true
+		} else {
+			switch keyringType {
+			case "wincred", "keychain":
+				if err := zkeyring.Set("HashiCorp Boundary Auth Token", tokenName, base64.RawStdEncoding.EncodeToString(marshaled)); err != nil {
```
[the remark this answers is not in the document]
jefferai: Done.

internal/cmd/commands/authenticate/oidc.go
jefferai: 2 seconds feels like it might be a bit too long. You could close a browser and then stare at the CLI wondering what it's doing. For now I'll split the difference at 1500ms :-)

PR #1088, opened 2021-04-08
Add Dry Run Update OIDC Auth Method in API and CLI

talanknight:
Missing dryrun tests for API but manually tested for CLI.


internal/servers/controller/handlers/authmethods/authmethod_service.go
```diff
@@ -441,16 +443,33 @@ func (s Service) updateInRepo(ctx context.Context, scopeId string, req *pbs.Upda
...
 		}
+		if dryRun {
+			storageToWire = func(in auth.AuthMethod) (*pb.AuthMethod, error) {
+				am, err := toAuthMethodProto(in)
+				if err != nil {
+					return nil, errors.Wrap(err, op)
+				}
+				attrs := &pb.OidcAuthMethodAttributes{}
+				if err := handlers.StructToProto(am.GetAttributes(), attrs); err != nil {
+					return nil, errors.Wrap(err, op, errors.WithMsg("cant convert from attribute struct to proto"))
```
jefferai, line 461:
```suggestion
					return nil, errors.Wrap(err, op, errors.WithMsg("can't convert from attribute struct to proto"))
```

internal/servers/controller/handlers/authmethods/authmethod_service.go
```diff
@@ -441,16 +443,33 @@ func (s Service) updateInRepo(ctx context.Context, scopeId string, req *pbs.Upda
...
+				}
+				attrs.DryRun = true
+				if am.Attributes, err = handlers.ProtoToStruct(attrs); err != nil {
+					return nil, errors.Wrap(err, op, errors.WithMsg("cant convert from attribute proto to struct"))
```
jefferai, line 465:
```suggestion
					return nil, errors.Wrap(err, op, errors.WithMsg("can't convert from attribute proto to struct"))
```

jefferai (approved):

PR #1089, opened 2021-04-08
issuer and api url Field condition update

talanknight:
Issuers and API Url both require a schema to be set (either http or https).
Modification on the issuer value now only strips off everything after and including ".well-known/".
Include the `disable_discovered_config_validation` on the auth method resource returned by the API.


jimlambrt (approved):
Looks great.  ty!

PR #1091, opened 2021-04-08
Vault integration: test helpers and token renewal

mgaffney:


go.mod
```diff
@@ -26,9 +26,11 @@ require (
 	github.com/hashicorp/go-kms-wrapping v0.6.1
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/hashicorp/go-retryablehttp v0.6.8
+	github.com/hashicorp/go-rootcerts v1.0.2 // indirect
```
louisruch, line 29: `go mod tidy`

internal/credential/vault/vault.go
```diff
@@ -0,0 +1,108 @@
...
+func newClient(c *clientConfig) (*client, error) {
+	const op = "vault.newClient"
+	if !c.IsValid() {
+		return nil, errors.New(errors.InvalidParameter, op, "invalid configuration")
+	}
+	vc := vault.DefaultConfig()
+	vc.Address = c.Addr
+	if len(c.CaCert) > 0 {
+		rootConfig := &rootcerts.Config{
+			CACertificate: []byte(c.CaCert),
```
louisruch, line 51: unnecessary type conversion `[]byte`

internal/credential/vault/vault.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+		tlsConfig := vc.HttpClient.Transport.(*http.Transport).TLSClientConfig
```
louisruch, line 53: Since you have the `TlsSkipVerify` bool in the config would you not want to set`InsecureSkipVerify` here?

louisruch (approved):
Some comments but otherwise LGTM with a caveat: I did not do a extremely thorough review of the `vault_testing.go` changes it seems good and well tested, however I am not super familiar with setting up docker especially for vault, so I would not have caught any non-obvious mistakes...

go.mod
mgaffney: ty!

internal/credential/vault/vault.go
mgaffney: ty!

internal/credential/vault/vault.go
mgaffney: Great catch! Ty!

PR #1099, opened 2021-04-09
Refactor internal/servers/controller/handlers errors v2 

s-christoff:
Reverted original due to merge conflicts with OIDC.


internal/servers/controller/handlers/mask_manager.go
```diff
@@ -30,26 +32,27 @@ func NewMaskManager(dest protoreflect.ProtoMessage, src ...protoreflect.ProtoMes
...
+			return nil, errors.New(errors.Encode, op, fmt.Sprintf("mapping src field %q maps to %q, dest %q maps to %q", k, v, v, ov))
 		}
 		result[k] = v
 	}
 
 	// Now check to make sure there aren't any dangling dest mappings.
 	for k, v := range destToSrc {
 		if ov, ok := srcToDest[v]; !ok || ov != k {
-			return nil, fmt.Errorf("mapping dest field %q maps to %q, src %q maps to %q", k, v, v, ov)
+			return nil, errors.New(errors.Encrypt, op, fmt.Sprintf("mapping src field %q maps to %q, dest %q maps to %q", k, v, v, ov))
```
louisruch, line 43: Should be `Encode` instead of `Encrypt` also I believe the dest and src order was correct.  The ordering of the map range is swapped from the previous range

louisruch (approved):
One comment but otherwise this looks great!!

s-christoff:
Multiple verbal approvals internally

PR #1102, opened 2021-04-10
Add Job Scheduler

louisruch:
### What does this PR do

- Moves job repo into scheduler: internal/job -> internal/scheduler/job
- Adds Job interface
- Adds scheduler package


louisruch:
@mgaffney ptal

louisruch:
@mgaffney PTAL

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,248 @@
...
+	_, err = repo.UpdateJobNextRun(ctx, string(jobId), nextRunIn)
+	if err != nil {
+		return errors.Wrap(err, op)
+	}
+	return nil
+}
+
+// Start begins the scheduling loop that will query the repository for jobs to run and
+// run them in a goroutine.
+func (s *Scheduler) Start() {
```
mgaffney, line 127: Why not accept a `context.Context` parameter?

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,248 @@
...
+	if s.started.Load() {
```
mgaffney, line 128: Shouldn't this be an atomic [compare-and-swap](https://pkg.go.dev/go.uber.org/atomic#Bool.CAS) operation?

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,248 @@
...
+	if !ok {
+		return fmt.Errorf("job (%v: %v) not registered on scheduler", jobId, job.Name())
+	}
+
+	s.l.Lock()
+	if _, ok := s.runningJobs[jobId]; ok {
+		s.l.Unlock()
+		return fmt.Errorf("job (%v: %v) is already running", jobId, job.Name())
+	}
+	jobContext, jobCancel := context.WithCancel(context.Background())
```
mgaffney, line 194: Why not derive the job's context from `s.baseContext` instead of `context.Background()`?

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,248 @@
...
+	go func() {
+		runErr := job.Run(jobContext)
+		repo, err := s.jobRepoFn()
+		if err != nil {
+			// Failed to create repo needed update run.  Best option is to log error,
+			// delete running job and return.
+			s.logger.Error("error creating job repo after job run", "error", err)
+			s.l.Lock()
+			delete(s.runningJobs, jobId)
+			s.l.Unlock()
```
mgaffney, line 207: This doesn't call `jobCancel()` and doesn't update the job status in the database.

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,248 @@
...
+		return fmt.Errorf("job (%v: %v) is already running", jobId, job.Name())
+	}
+	jobContext, jobCancel := context.WithCancel(context.Background())
+	s.runningJobs[jobId] = runningJob{runId: r.PrivateId, cancelCtx: jobCancel}
+	s.l.Unlock()
+
+	go func() {
+		runErr := job.Run(jobContext)
+		repo, err := s.jobRepoFn()
+		if err != nil {
```
mgaffney, line 201: Why not make this call before starting the goroutine so the error could be returned and handled by the caller?

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,248 @@
...
+			}
+		default:
+			s.logger.Debug(fmt.Sprintf("job run %q for job (%v: %v) failed: %v", r.PrivateId, jobId, job.Name(), runErr))
+			if _, inner := repo.FailRun(jobContext, r.PrivateId); inner != nil {
+				s.logger.Error("error updating failed job run", "error", inner)
+			}
+		}
+
+		s.l.Lock()
+		delete(s.runningJobs, jobId)
```
mgaffney, line 225: The cancel function for the job's context is never called which will leak resources.

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	s.l.Lock()
+	for jId, j := range s.runningJobs {
+		j.cancelCtx()
+		delete(s.runningJobs, jId)
+	}
+	s.l.Unlock()
+
+	s.baseCancel()
```
mgaffney, line 246: Should this be before the line 240 to prevent additional jobs from being run?

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,248 @@
...
+	s.started.Store(false)
```
mgaffney, line 247: Multiple concurrent calls could happen between line 234 and this line.

mgaffney (changes requested):

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,248 @@
...
+	if code == "" {
+		return "", errors.New(errors.InvalidParameter, op, "missing code")
+	}
+
+	repo, err := s.jobRepoFn()
+	if err != nil {
+		return "", errors.Wrap(err, op)
+	}
+
+	j, err := repo.CreateJob(ctx, job.Name(), code, job.Description())
```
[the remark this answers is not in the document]
louisruch: Yeah I caught this in my manual testing, I need to dedupe at this point

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,248 @@
...
+
+// Scheduler is used to register and run recurring jobs on the controller.
+type Scheduler struct {
+	serverId  string
+	jobRepoFn jobRepo.JobRepoFactory
+	logger    hclog.Logger
+
+	registeredJobs *sync.Map
+	runningJobs    map[JobId]runningJob
+	l              sync.RWMutex
```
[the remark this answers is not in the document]
louisruch: The goroutine that is updating progress will be grabing an rlock

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,248 @@
+package scheduler
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/hashicorp/boundary/internal/errors"
+	jobRepo "github.com/hashicorp/boundary/internal/scheduler/job"
```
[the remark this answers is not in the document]
louisruch: ty

internal/scheduler/scheduler.go
louisruch: ty

internal/scheduler/scheduler.go
louisruch: I wanted to discuss this with you, I was worried about the `repo` sitting around while the job runs as the `job.Run` is a blocking call that blocks for an unknown time

internal/scheduler/scheduler.go
louisruch: ty re cancel, at this point i cannot update the database because I could not create the repo

mgaffney (approved):
LGTM. One thing I noticed is the pb.go files are generated with protoc 3.12.4. The rest of the code base is using 3.15.8 so you might want to install 3.15.8 and run `make proto` again.

internal/scheduler/job.go
```diff
@@ -0,0 +1,63 @@
...
+
+// Job defines an interface for jobs that can be invoked by the scheduler.
+type Job interface {
+	// Status reports the job’s current status.  The status is periodically persisted by
+	// the scheduler when a job is running, and will be used to verify a job is making progress.
+	Status() JobStatus
+
+	// Run performs the required work depending on the implementation.
+	// The context is used to notify the job that it should exit early.
+	Run(ctx context.Context) error
```
jefferai, line 23: Placeholder comment as I have a suggestion here but need to see implementation first.

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,264 @@
...
+//
+// • jobId must be provided and is the id of the job returned by RegisterJob
+//
+// All options are ignored.
+func (s *Scheduler) UpdateJobNextRun(ctx context.Context, jobId JobId, nextRunIn time.Duration, _ ...Option) error {
+	const op = "scheduler.(Scheduler).UpdateJobNextRun"
+	if jobId == "" {
+		return errors.New(errors.InvalidParameter, op, "missing job id")
+	}
+	_, ok := s.registeredJobs.Load(jobId)
```
jefferai, line 126: Is it possible that a jobId could be passed in here that was sourced in the DB elsewhere but has not yet been loaded into the cache?

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,264 @@
...
+
+// Start begins the scheduling loop that will query the repository for jobs to run and
+// run them in a goroutine.
+func (s *Scheduler) Start(ctx context.Context) {
+	if !s.started.CAS(s.started.Load(), true) {
+		s.logger.Debug("scheduler already started, skipping")
+		return
+	}
+
+	s.baseContext, s.baseCancel = context.WithCancel(ctx)
```
jefferai, line 150: Why not use the controller's base context? It seems like this cancel function is only ever called when Shutdown is called; if this is only called when the controller is itself shutdown, you'd already get a canceled base context, so this just becomes superfluous.

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,264 @@
...
+		s.logger.Debug("scheduling loop going back to sleep")
+		timer.Reset(s.runJobsInterval)
+	}
+}
+
+func (s *Scheduler) runJob(r *job.Run) error {
+	jobId := JobId(r.JobId)
+	j, ok := s.registeredJobs.Load(jobId)
+	job := j.(Job)
+	s.logger.Debug(fmt.Sprintf("starting job run %q for job (%v: %v)", r.PrivateId, jobId, job.Name()))
```
jefferai, line 200: Don't use printf and friends with the logger; use the logger's arguments. This would be something like:

`s.logger.Debug("starting job run", "id", r.PrivateId, "job_id", jobId, "name", job.Name())`

Also what's the difference between the private ID and the jobId here? Shouldn't they be the same?

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,264 @@
...
+		}
+
+		s.logger.Debug("scheduling loop going back to sleep")
+		timer.Reset(s.runJobsInterval)
+	}
+}
+
+func (s *Scheduler) runJob(r *job.Run) error {
+	jobId := JobId(r.JobId)
+	j, ok := s.registeredJobs.Load(jobId)
```
jefferai, line 198: After looking at the various places that the `registeredJobs` map is used I'm not really seeing the value. Here it is used to get the name (which you have from the incoming `job.Run` anyways), and then `Run` is called, but that could be a function on the `job.Run`. Is there more that is planned for this cache? As-is it saves a round trip to the DB on registration but this is unlikely to be very dynamic at all, and it is used to check whether it's cached in the update function but nothing is done with the value, and if the job ID is invalid the database layer will error anyways.

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,264 @@
...
+	jobContext, jobCancel := context.WithCancel(s.baseContext)
+	s.runningJobs[jobId] = runningJob{runId: r.PrivateId, cancelCtx: jobCancel}
+	s.l.Unlock()
+
+	go func() {
+		defer jobCancel()
+		runErr := job.Run(jobContext)
+		switch runErr {
+		case nil:
+			s.logger.Debug(fmt.Sprintf("job run %q for job (%v: %v) complete", r.PrivateId, jobId, job.Name()))
```
jefferai, line 224: Same comment here and below about logging arguments.

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,264 @@
...
+}
+
+// Scheduler is used to register and run recurring jobs on the controller.
+type Scheduler struct {
+	serverId  string
+	jobRepoFn jobRepoFactory
+	logger    hclog.Logger
+
+	registeredJobs *sync.Map
+	runningJobs    map[JobId]runningJob
```
jefferai, line 29: Any reason not to use a sync.Map here? I know below that there is a spot that locks the mutex and looks for existing value in the map but that can be handled with LoadOrStore on a sync.Map.

louisruch: Hmm, I could.  I had some more complex locking at first which I removed after some tweaks which means I could probably do this move as well...

internal/scheduler/scheduler.go
louisruch: It shouldn't with the current setup but I guess it doesn't have to be a requirement here, lets chat tomorrow to walk through the process...

internal/scheduler/scheduler.go
louisruch: My thought was to be able to spin the scheduler up and down without needing to cancel the controllers base context.  Not a strict requirement just didn't necessarily want to tie them so strictly together.

internal/scheduler/scheduler.go
louisruch: Lets chat in person, the main reason for `registeredJobs` is the `Run()` I have no other way of getting it...  The idea here is that the scheduler needs to be initialized with all jobs to be able to run any job - without this step we would need to serialize the job struct in order perform the `Run`

internal/scheduler/scheduler.go
louisruch: ty

internal/scheduler/scheduler.go
jefferai: We can discuss today -- my gut says that the scheduler likely should always be spun up when the controller is active, and it's individual jobs that may or may not need to be spun up.

internal/scheduler/scheduler.go
jefferai: Generally speaking the mutex+lock is a good approach if you either have a bad use case for sync.Map (as they outline in the docs) or if you specifically want to stop execution until you get the lock. I believe in your cases you don't need to stop execution if it's already there (or not), just take different paths. If I'm correct then I do think it would simplify things a bit to use it. It's not complex now, but anytime you can remove dealing with a separate lock is usually for the better :-)

louisruch:
Spoke with @jefferai going to merge this into the long live `louis-job` branch and make the changes he suggested in my follow up PR

PR #1103, opened 2021-04-12
Use 202 for token polling endpoint instead of 204

jefferai:


randallmorey (dismissed):

internal/servers/controller/handlers/authmethods/oidc.go
```diff
@@ -269,7 +270,16 @@ func (s Service) authenticateOidcToken(ctx context.Context, req *pbs.Authenticat
 		}
 	}
 	if token == nil {
-		return nil, nil
+		attrs, err := structpb.NewStruct(map[string]interface{}{
+			"status": "unknown",
```
jimlambrt, line 274: perhaps define const for the field name?

internal/servers/controller/handlers/outgoing_interceptor.go
```diff
@@ -34,6 +35,15 @@ func OutgoingInterceptor(ctx context.Context, w http.ResponseWriter, m proto.Mes
 			return nil
 		}
 		fields := m.GetAttributes().GetFields()
+		if m.GetCommand() == "token" {
```
jimlambrt, line 38: perhaps define a const for "token" which is used a few times.

jimlambrt (dismissed):
Just a few minor suggestions.

internal/servers/controller/handlers/outgoing_interceptor.go
jefferai: It's used a few times but in different ways. In one case it's a command, in the other it's a field name. So it could be e.g. `tokenString` but `tokenField` and `tokenCommand` don't fit...which makes me wonder if there is then value since it's purely to get rid of a string and not to convey useful info.

internal/servers/controller/handlers/authmethods/oidc.go
jefferai: Done.

PR #1105, opened 2021-04-12
Make oidc auth callback redirect on most errors

talanknight:
This also removes the token url field in the authenticate start response.
Big open question is how we want to format the error response.  Happy to iterate on the formatting in this PR or a follow up PR if prefered.


internal/auth/oidc/repository_auth_method_read.go
```diff
@@ -61,7 +61,7 @@ func (r *Repository) lookupAuthMethod(ctx context.Context, authMethodId string,
 
 // getAuthMethods allows the caller to either lookup a specific AuthMethod via
 // its id or search for a set AuthMethods within a set of scopes.  Passing both
-// scopeIds and a authMethodId is an error. The WithUnauthenticatedUser,
+// scopeIds and a authMethod is an error. The WithUnauthenticatedUser,
```
jimlambrt, line 64: I don't think you're passing the entire auth method.. just its id.

internal/auth/oidc/service_callback.go
```diff
@@ -46,7 +46,7 @@ func Callback(
 	oidcRepoFn OidcRepoFactory,
 	iamRepoFn IamRepoFactory,
 	atRepoFn AuthTokenRepoFactory,
-	authMethodId,
+	am *AuthMethod,
```
jimlambrt, line 49: why pass the entire auth method, when just its public id is needed?

jimlambrt: it would appear all the changes in this method and the tests aren't needed if we keep the more precise parameter of an auth method id.

internal/servers/controller/handlers/authmethods/oidc.go
```diff
@@ -150,94 +150,124 @@ func (s Service) authenticateOidc(ctx context.Context, req *pbs.AuthenticateRequ
...
+	if err != nil {
+		return nil, errors.Wrap(err, op)
+	}
+	if am == nil {
+		return nil, errors.New(errors.RecordNotFound, op, fmt.Sprintf("auth method %s not found", req.GetAuthMethodId()))
+	}
+	if am.GetApiUrl() == "" {
+		return nil, errors.New(errors.InvalidParameter, op, "auth method doesn't have api url defined")
 	}
+	errResponse := errorResponseBuilder(am.GetApiUrl())
```
jimlambrt, line 221: why use a function pointer here?

internal/servers/controller/handlers/authmethods/oidc.go
```diff
@@ -150,94 +150,124 @@ func (s Service) authenticateOidc(ctx context.Context, req *pbs.AuthenticateRequ
...
+		// TODO: Package the OIDC error into the redirectUrl
+		return errResponse(fmt.Errorf("Error: %q, Details: %q", attrs.GetError(), attrs.GetErrorDescription())), nil
+	}
+	finalRedirectUrl, err = oidc.Callback(
 		ctx,
 		s.oidcRepoFn,
 		oidc.IamRepoFactory(s.iamRepoFn),
 		s.atRepoFn,
-		req.GetAuthMethodId(),
+		am,
```
jimlambrt, line 239: why not just pass the am.PublicId?

jimlambrt (changes requested):

internal/servers/controller/handlers/authmethods/oidc.go
talanknight: The auth method is needed in this calling method to redirect for errors in this method.  Instead of doing the lookup twice I passed it through.

talanknight: Just to follow up, relooking at the code the called method does not need the full auth method.  I must have confused my initial iteration on this fix, where the called method also generates the redirectUrl with what is actually there.  I'll fix this and re-request review.  Thanks!

jimlambrt (dismissed):

internal/servers/controller/handlers/authmethods/oidc.go
talanknight: follow up to the follow up: The auth method is needed for looking up the provider, getting scope, state, max age, etc...  Leaving it as is.

PR #1106, opened 2021-04-13
Fix curl string building

jefferai:
I'm not sure why it was doing it the other way with assuming the string
being built would always have a space at the end, but, this way is
easier to understand and fixes a bug where there wasn't a space being
left between some fields.


s-christoff (approved):
🤙

PR #1108, opened 2021-04-13
Add more OIDC error output.

jefferai:
Some of this is in the form of logging, and some via returning error
messages to users.

Also cleans a couple of linter complaints.

### DIFF
--- a/internal/servers/controller/handlers/authmethods/oidc.go
+++ b/internal/servers/controller/handlers/authmethods/oidc.go
@@ -26,7 +26,6 @@ const (
 	tokenCommand    = "token"
 
 	// token request/response fields
-	tokenField  = "token"
 	statusField = "status"
 
 	// field names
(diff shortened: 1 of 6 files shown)


internal/servers/controller/handlers/authmethods/oidc.go
```diff
@@ -259,7 +259,7 @@ func (s Service) authenticateOidcCallback(ctx context.Context, req *pbs.Authenti
 		FinalRedirectUrl: finalRedirectUrl,
 	})
 	if err != nil {
-		return errResponse(errors.New(errors.Internal, op, "Error marshaling parameters")), nil
+		return errResponse(errors.New(errors.Internal, op, "Error marshaling parameters after successful callback", errors.WithWrap(err))), nil
```
talanknight, line 262: I thought you wanted to log these errors.  Currently this would not be logged because it is sent as a redirect url instead of through the normal way we return error messages.

talanknight (approved):
Looks good.  The only place I think it might make sense to add some more logging would be oidc.go line 238 since that seems like it would be internal error, but the validate call should have caught that already anyways.

jefferai:
Yeah, that's why I left that comment.

PR #1111, opened 2021-04-13
Update misleading Authenticated bool with new name and explanation.

jefferai:


internal/auth/auth.go
```diff
@@ -63,11 +63,31 @@ type RequestInfo struct {
...
-	AuthTokenId   string
-	Error         error
-	Scope         *scopes.ScopeInfo
-	Authenticated bool
+	UserId      string
+	AuthTokenId string
+	Error       error
+	Scope       *scopes.ScopeInfo
+
+	// AuthenticatedFinished means that the request has been passed through the
```
jefferai, line 71: ``
```suggestion
	// AuthenticatedFinished means that the request has passed through the
```

talanknight (approved):

PR #1112, opened 2021-04-13
Add -container-image flag 

s-christoff:
Allows users to input their own container image. 
Defaults to `postgres:latest` if none is supplied.


internal/docker/supported.go
```diff
@@ -25,23 +25,26 @@ func startDbInDockerSupported(dialect string) (cleanup func() error, retURL, con
...
+			resource, err = pool.Run(dialect, tag, []string{"POSTGRES_PASSWORD=password", "POSTGRES_DB=boundary"})
 			url = "postgres://postgres:password@localhost:%s?sslmode=disable"
 			if err == nil {
 				url = fmt.Sprintf("postgres://postgres:password@%s/boundary?sslmode=disable", resource.GetHostPort("5432/tcp"))
 			}
 		}
 
 	default:
-		panic(fmt.Sprintf("unknown dialect %q", dialect))
+		fmt.Errorf("unknown dialect %q", dialect)
```
jefferai, line 47: Did you mean to assign to `err` here? As it is this will basically be a no-op and won't trigger exiting in the next block.

internal/docker/supported.go
```diff
@@ -83,3 +86,19 @@ func cleanupDockerResource(pool *dockertest.Pool, resource *dockertest.Resource)
 	}
 	return fmt.Errorf("Failed to cleanup local container: %s", err)
 }
+
+// splitImage takes the WithDatabaseImage option and separates
+// it into repo + tag. If a tag is not found, it sets the tag to latest
+func splitImage(opt ...Option) (string, string) {
+	opts := GetOpts(opt...)
+	if opts.withContainerImage == "" {
+		return "postgres", "latest"
```
jefferai, line 95: Our current version we target is `11`, please use that!

internal/docker/supported.go
```diff
@@ -83,3 +86,19 @@ func cleanupDockerResource(pool *dockertest.Pool, resource *dockertest.Resource)
...
+	}
+	separated := strings.Split(opts.withContainerImage, ":")
+	if len(separated) > 1 {
```
jefferai, line 98: Docker image tags cannot themselves contain colons:
```
$ docker tag postgres:11 mypostgres:11:foo
Error parsing reference: "mypostgres:11:foo" is not a valid repository/tag: invalid reference format
```
So this function should really return an error as well, and you could put this logic into a `switch` on the length of the split strings, with cases for 1, 2, and a default that errors.

internal/docker/option.go
```diff
@@ -0,0 +1,29 @@
...
+
+// Options - how Options are represented.
+type Options struct {
+	withContainerImage string
+}
+
+func getDefaultOptions() Options {
+	return Options{}
+}
+
```
jefferai, line 23: The double newline here is a clue to me that you don't have your editor formatting code for you on save 😸

internal/db/testing.go
```diff
@@ -30,7 +31,8 @@ func TestSetup(t *testing.T, dialect string, opt ...TestOption) (*gorm.DB, strin
 
 	switch opts.withTestDatabaseUrl {
 	case "":
-		cleanup, url, _, err = StartDbInDocker(dialect)
+		opts := docker.WithContainerImage(dialect)
```
jefferai, line 34: Note that our "dialect" is actually `"postgres"`. This means that the net effect of passing `dialect` straight through will be that it will not have any colons, so will split into a db and tag of `postgres:latest` instead of what we want, which is `postgres:11`.

The disconnect here is that the dialect is referring to the SQL language understood, not a particular version of a server that speaks it. (You could for instance have a dialect of `postgres` and actually use a `cockroachdb` image.) In fact in `manager_test.go` above (yes, I'm reviewing upwards), you can see that issue -- you're creating a dialect of `postgres` which is correct for `sql.Open` but will result in the wrong image (`postgres:latest`).

I think you need to re-add dialect as a parameter in the places where you replaced it with the options, and use them in the appropriate and distinct ways. They're separate things.

internal/docker/supported.go
s-christoff: 🤦🏻  I meant to revert this, my b

internal/cmd/base/option.go
```diff
@@ -83,3 +84,9 @@ func WithSkipTargetCreation() Option {
 		o.withSkipTargetCreation = true
 	}
 }
+
+func WithContainerImage(name string) Option {
```
louisruch, line 88: Can you add godoc comment

internal/cmd/base/servers.go
```diff
@@ -462,16 +462,34 @@ func (b *Server) ConnectToDatabase(dialect string) error {
...
+	// assign the repo to dialect
+	switch opts.withContainerImage {
+	case "postgres":
+		dialect = "postgres"
+	default:
+		separated := strings.Split(opts.withContainerImage, ":")
+		dialect = separated[0]
+	}
+
+	if opts.withContainerImage == "" {
```
louisruch, line 483: Seems we can drop this check and just have the above switch is we add `withContainerImage: "postgres"` to `getDefaultOptions`

internal/docker/option.go
```diff
@@ -0,0 +1,28 @@
+package docker
+
+//GetOpts - iterate the inbound Options and return a struct.
```
louisruch, line 3:
```suggestion
// GetOpts - iterate the inbound Options and return a struct.
```

internal/docker/option.go
```diff
@@ -0,0 +1,28 @@
...
+// Options - how Options are represented.
+type Options struct {
+	withContainerImage string
+}
+
+func getDefaultOptions() Options {
+	return Options{}
+}
+
+func WithContainerImage(image string) Option {
```
louisruch, line 24: godoc comment please

internal/docker/option.go
```diff
@@ -0,0 +1,28 @@
...
+// Option - how Options are passed as arguments.
+type Option func(*Options)
+
+// Options - how Options are represented.
+type Options struct {
+	withContainerImage string
+}
+
+func getDefaultOptions() Options {
+	return Options{}
```
louisruch, line 21: Similar question as above to defaulting the container image to `postgres` this will allow you to do: 
```
if err := c.CreateDevDatabase(c.Context, opts...); err != nil {...}
```

instead of needing to append an option like:
```
opts = append(opts, base.WithContainerImage("postgres"))
if err := c.CreateDevDatabase(c.Context, opts...); err != nil {....}
```

internal/docker/supported.go
```diff
@@ -81,5 +91,26 @@ func cleanupDockerResource(pool *dockertest.Pool, resource *dockertest.Resource)
...
+// it into repo + tag. If a tag is not found, it sets the tag to latest
+func splitImage(opt ...Option) (string, string, error) {
+	opts := GetOpts(opt...)
+
+	separated := strings.Split(opts.withContainerImage, ":")
+	separatedlen := len(separated)
+
+	switch separatedlen {
+	case 1:
+		return separated[0], "11", nil
```
louisruch, line 107: The comment mentioned this sets tag to latest, however it seems you set it to `11` here.  I believe this is the version we support for postgres right now but it is not the latest. We should also perhaps error if `separated[0] != "postgres"`

internal/cmd/base/servers.go
s-christoff: 🤦‍♀️🤦‍♀️🤦‍♀️🤦‍♀️

internal/docker/supported.go
```diff
@@ -81,5 +91,31 @@ func cleanupDockerResource(pool *dockertest.Pool, resource *dockertest.Resource)
...
+	opts := GetOpts(opt...)
+
+	separated := strings.Split(opts.withContainerImage, ":")
+	separatedlen := len(separated)
+
+	switch separatedlen {
+	case 1:
+		if separated[0] == "postgres" {
+			return separated[0], "11", nil
+		} else {
```
louisruch, line 109: you can drop the `else` since the `if` has a return inside of it

internal/docker/supported.go
```diff
@@ -25,15 +25,25 @@ func startDbInDockerSupported(dialect string) (cleanup func() error, retURL, con
 	}
 
 	var resource *dockertest.Resource
-	var url string
+	var url, tag string
+
+	opts := GetOpts(opt...)
+	if opts.withContainerImage != "" {
+		dialect, tag, err = splitImage(opt...)
```
louisruch, line 32: I am a little confused with what is going on in this path, we parse the `withContainerImage` then call this function with the `dialect` as well as the the option.  Here we parse it again setting the `dialect` parameter we passed in?

s-christoff: I think this makes more sense *now* (as of recent commit), but I'm still thinking about it.

internal/cmd/base/servers.go
```diff
@@ -462,16 +462,27 @@ func (b *Server) ConnectToDatabase(dialect string) error {
...
+	opts := getOpts(opt...)
+
+	// opts.WithContainerImage could have repo:tag, so split it and only
+	// assign the repo to dialect
+	switch opts.withContainerImage {
+	case "postgres":
+		dialect = "postgres"
+	default:
+		separated := strings.Split(opts.withContainerImage, ":")
+		dialect = separated[0]
```
jefferai, line 480: I defer to @jimlambrt on this but I think we currently always want the dialect to be "postgres". The dialect is passed through to the SQL driver and is related to the actual SQL that we have. I'm not sure it would actually work if we set the dialect to anything else, even if the docker container itself is launching something that _speaks_ the postgres protocol.

jefferai: I think for now the best way to handle this would be to have a WithDialect option that defaults to `postgres`.

s-christoff: I actually don't think we want this anymore but I'm looking

jefferai (approved):
😹

internal/cmd/base/option.go
```diff
@@ -83,3 +88,17 @@ func WithSkipTargetCreation() Option {
...
+
+// WithContainerImage tells the command which container image
+// to start a dev database with
+func WithContainerImage(name string) Option {
+	return func(o *Options) {
+		o.withContainerImage = name
+	}
+}
+
+func withDialect(dialect string) Option {
```
louisruch, line 100: Is there a reason you do not want to export this option?  If you do export please add godoc, otherwise this is fine

internal/cmd/base/servers.go
```diff
@@ -462,16 +462,25 @@ func (b *Server) ConnectToDatabase(dialect string) error {
...
+func (b *Server) CreateDevDatabase(ctx context.Context, opt ...Option) error {
 
-	var container, url string
+	var container, url, dialect string
 	var err error
 	var c func() error
 
+	opts := getOpts(opt...)
+
+	//We should only get back postgres for now, but laying the foundation for non-postgres
```
louisruch, line 473:
```suggestion
	// We should only get back postgres for now, but laying the foundation for non-postgres
```

internal/docker/supported.go
```diff
@@ -81,5 +96,30 @@ func cleanupDockerResource(pool *dockertest.Pool, resource *dockertest.Resource)
...
 		return nil
 	}
-	return fmt.Errorf("Failed to cleanup local container: %s", err)
+	return fmt.Errorf("failed to cleanup local container: %s", err)
+}
+
+// splitImage takes the WithDatabaseImage option and separates
+// it into repo + tag. If a tag is not found, it sets the tag to latest
+func splitImage(opts Options) (string, string, error) {
+
```
louisruch, line 105: nit: remove white space

internal/docker/supported.go
```diff
@@ -81,5 +96,30 @@ func cleanupDockerResource(pool *dockertest.Pool, resource *dockertest.Resource)
...
+	separated := strings.Split(opts.withContainerImage, ":")
+	separatedlen := len(separated)
+
+	switch separatedlen {
+	case 1:
+		if separated[0] == "postgres" {
+			return separated[0], "11", nil
+		}
+		return "", "", fmt.Errorf("valid reference format is repo:tag, if"+
```
louisruch, line 114: I am not exactly sure what you want this to do, it seems like this is the correct behavior where `postgres` becomes `postgres:11` otherwise it fails.  However, an alternative is to do what the comment says it will do here:
```
if separated[0] == "postgres" {
	return separated[0], "11", nil
}
return separated[0], "latest", nil
```
It is fine to leave as is as well, mainly pointing out discrepancy between the comment and code...

louisruch: Or 
```
version := "latest"
if separated[0] == "postgres" {
	version = "11"
}
return separated[0], version, nil
```

louisruch (approved):
A couple of small comments and nit otherwise LGTM

PR #1122, opened 2021-04-14
Restrict auth method listing for u_anon

jefferai:
This keeps only essential information in list output for auth methods
when the user is u_anon. Boundary's default is to allow listing of
auth methods for anonymous users to allow discovery of authentication
points, but there's no need to return more information about internal
configuration than necessary.

### DIFF
--- a/internal/servers/controller/handlers/options.go
+++ b/internal/servers/controller/handlers/options.go
@@ -20,6 +20,7 @@ type Option func(*options)
 type options struct {
 	withDiscardUnknownFields bool
 	WithLogger               hclog.Logger
+	WithAnonymousListing     bool
 }
 
 func getDefaultOptions() options {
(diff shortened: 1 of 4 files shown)


internal/servers/controller/handlers/options.go
```diff
@@ -41,3 +42,11 @@ func WithLogger(logger hclog.Logger) Option {
 		o.WithLogger = logger
 	}
 }
+
+// WithAnonymousListing provides an option when creating responses to only include those
+// desired for listing to anonymous users.
+func WithAnonymousListing(anonListing bool) Option {
```
jimlambrt, line 48: mind adding a test?

talanknight (dismissed):

internal/servers/controller/handlers/options.go
jimlambrt: Not sure it needs to be exported.

jimlambrt (approved):
minor bits.

internal/servers/controller/handlers/options.go
jefferai: It does, because this is in `handlers`, and it's being used in actual handler implementations.

jefferai: Test added.

PR #1123, opened 2021-04-14
website: add support for tip.boundaryproject.io

kylemac:
adds proper configuration for tip.boundaryproject.io which will be a subdomain to showcase upcoming/unreleased documentation being prepared for the latest version. this documentation will be pulled from `origin/main`

this should be a simple review. i have not configured the actual domain DNS yet because i want this logic in place which to prevent indexing before the DNS change makes the actual subdomain propagate. In effect, you're just making sure this change doesn't break the site build and that i got the config option name correct


BRKalow (approved):
👍

PR #1129, opened 2021-04-15
fix: remove extra "s" from auth method authorized collections listing

malnick:
When reading an auth method's authorized collections we have an extra 's': 
```
Authorized Actions on Auth Method's Collections:
    accountss:
      create
      list
```

I think the plurality was intentional. If it is, we can supersede this with another larger PR that fixes this across our authorized actions listings.


jefferai:
This probably needs to be updated in host catalogs too. Our generator also has `%ss` but not sure if it is producing wrong output everywhere or if it's usually right, as host catalogs and auth methods have extra output.

jefferai:
I've closed this in 46d7613e with this change and host catalogs, which had the same exact problem.

PR #1130, opened 2021-04-15
internal/servers/controller/handlers/users: fix dropped test error

alrs:
This fixes a dropped test error in `internal/servers/controller/handlers/users`.


jefferai:
Thanks!

PR #1132, opened 2021-04-15
website: move tip config obj to proper place

kylemac:
the config value wasn't being caught until moved to the callback


BRKalow (approved):
Huh, I see in the implementation why this is the case, but it seems like the original spot would be the proper spot for custom config attributes we declare.

kylemac:
@BRKalow yeah, i'm a bit confused too. This does the trick for now but let's maybe figure out what's amiss here

PR #1137, opened 2021-04-15
Bump PQ dep

jefferai:


jefferai:
https://github.com/lib/pq/releases/tag/v1.10.0

jefferai:
Just a single change to support inline SSL certs, so merged.

PR #1138, opened 2021-04-15
Add "no-op" action and use it as default for scope permissions

jefferai:
Because of listing visibility, if no actions are assigned to a user for
an individual resource, it's excluded from the list. This is fine and
good. However, when we made a change in 0.2.0 to restrict the
information returned from listing for the anonymous user for scopes and
auth methods, it exposed the fact that because of the way this
visibility works we had to add `read` to scopes by default -- thereby
rather defeating the purpose of restricting what goes back to `u_anon`
requests.

The purpose of "no-op" is to never be used for real actions within the
API, but as something that can be granted to specifically enable
visibility for the targeted user. This makes the controls over list
visibility much more granular.

Why is it called "no-op" instead of e.g. "visible"? Both because grants
provide actions, not nouns, and because if it's something that conveys
real capability people might assume it can be used to do something
real :-)


louisruch (approved):
This looks great, it did feel strange needing read access to list this resolve this perfectly!

jimlambrt (approved):

PR #1148, opened 2021-04-19
schema changes.

jimlambrt:


internal/db/schema/migrations/postgres/3/01_oidc.up.sql
```diff
@@ -0,0 +1,28 @@
...
+    references auth_oidc_method(public_id)
+    on delete cascade
+    on update cascade,
+  scope text not null
+    constraint scope_must_not_be_empty
+       check(length(trim(scope)) > 0) 
+    constraint scope_must_be_less_than_1024_chars
+      check(length(trim(scope)) < 1024)
+    constraint scope_must_not_be_openid -- the default scope is not allowed, since it's redundant
+      check(trim(scope) != 'openid'),
```
mgaffney, line 21: Does case sensitivity matter?

internal/db/schema/migrations/postgres/3/01_oidc.up.sql
```diff
@@ -0,0 +1,28 @@
...
+  primary key(oidc_method_id, scope)
+);
+comment on table auth_oidc_aud_claim is
```
mgaffney, line 24:
```suggestion
comment on table auth_oidc_scope is
```
Wrong table :)

internal/db/schema/migrations/postgres/3/01_oidc.up.sql
```diff
@@ -0,0 +1,28 @@
+begin;
+
+
+-- auth_oidc_scope entries are the optional scopes for a specific oidc auth
+-- method.  There can be 0 or more for each parent oidc auth method.  If an auth
+-- method has any scopes, they will be added to provider requests along with the
+-- default of "openid". 
+create table auth_oidc_scope (
+  create_time wt_timestamp,
```
mgaffney, line 9: Should there be a trigger to handle the create_time?

internal/db/schema/migrations/postgres/3/01_oidc.up.sql
```diff
@@ -0,0 +1,28 @@
...
+       check(length(trim(scope)) > 0) 
+    constraint scope_must_be_less_than_1024_chars
+      check(length(trim(scope)) < 1024)
+    constraint scope_must_not_be_openid -- the default scope is not allowed, since it's redundant
+      check(trim(scope) != 'openid'),
+  primary key(oidc_method_id, scope)
+);
+comment on table auth_oidc_aud_claim is
+'auth_oidc_scope entries are the optional scopes for a specific oidc auth method.  There can be 0 or more for each parent oidc auth method.  If an auth method has any scopes, they will be added to provider requests along with the openid default.';
+
```
mgaffney, line 26: Are there any immutable columns?

mgaffney (changes requested):

internal/db/schema/migrations/postgres/3/01_oidc.up.sql
jimlambrt: I don't think so... I doubt the API will support updates, but I don't see any integrity issues with the value objects in this table  if it does.

internal/db/schema/migrations/postgres/3/01_oidc.up.sql
jimlambrt: ty!

internal/db/schema/migrations/postgres/3/01_oidc.up.sql
jimlambrt: yes, they are: `case sensitive list of ASCII scope values.`. ty!

mgaffney (approved):

PR #1150, opened 2021-04-19
Update other resources' anon listing to match auth-methods and scopes pattern

jefferai:
This updates the various resource types to match the behavior we introduced in 0.2.0 for auth-methods and scopes, that is, only returning some specific information when the listing user is the anonymous user.


talanknight (commented):
Let's discuss this in slack.

talanknight (approved):

PR #1151, opened 2021-04-20
Use Go 1.16 native embedding and remove go-bindata

jefferai:


samsalisbury (approved):

internal/ui/assets.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+// Sadly we can't embed this into the embed line, but we use it elsewhere
+const contentDir = ".tmp/boundary-ui/ui/admin/dist"
+
+// content is our static web server content.
+//go:embed .tmp/boundary-ui/ui/admin/dist
+var content embed.FS
+
+func Handler() http.Handler {
```
jimlambrt, line 18: Any desire to make this more testable via a func that just returns the desired embed FS?    That would allow us to assert at least some of the assets we believe should be available from the `dist` directory.

jefferai: Sure, I can do that.

randallmorey (approved):
Worked for me locally!

PR #1154, opened 2021-04-21
fix primary account data returned for users when a passwd auth method is the primary auth method

jimlambrt:


jefferai (approved):
Makes sense to me!

PR #1155, opened 2021-04-21
Update delete action behavior and JSON format output

jefferai:
This PR changes delete behavior:

* On success, a `204` is returned rather than a `200` with an empty JSON object
* The CLI now passes through the `404` error if encountered instead of swallowing it

It also updates formatting for JSON to use the response body, which means that the `item` (or `items`) in the JSON format are actually the format across the wire, instead of being reinterpreted by unmarshaling into the Go SDK and then remarshaling.


jimlambrt (approved):
Looks great!

randallmorey (approved):

talanknight (approved):

PR #1160, opened 2021-04-22
Ongoing OIDC: incorporate ClaimsScopes into the oidc repository 

jimlambrt:


jimlambrt:
This PR is against main and it should be against `jimlambrt-oidc-scopes`.  I'll reopen it against the right branch

PR #1164, opened 2021-04-23
Vault integration: add repository create credential library method

mgaffney:
This also adds the `http_method` and `http_request_body` attributes to the Vault credential library.


internal/credential/vault/repository_credential_library.go
```diff
@@ -0,0 +1,74 @@
...
+	if l.VaultPath == "" {
+		return nil, errors.New(errors.InvalidParameter, op, "no vault path")
+	}
+	if l.PublicId != "" {
+		return nil, errors.New(errors.InvalidParameter, op, "public id not empty")
+	}
+	if scopeId == "" {
+		return nil, errors.New(errors.InvalidParameter, op, "no scope id")
+	}
+	l = l.clone()
```
louisruch, line 42: You are calling clone here and then again on line 58 one of them could be dropped

internal/db/schema/migrations/postgres/5/02_vault_credential.up.sql
```diff
@@ -179,6 +198,22 @@ begin;
...
       constraint vault_path_must_not_be_empty
         check(length(trim(vault_path)) > 0),
+    http_method text not null
+      constraint credential_vault_http_method_enm_fkey
+        references credential_vault_http_method_enm (name)
+        on delete restrict
+        on update cascade,
+    http_request_body text
+      constraint http_request_body_only_allowed_with_post_method
+        check(
```
louisruch, line 208: neat will make a note of this for future!

louisruch (approved):
LGTM, one comment re cloning twice

internal/credential/vault/repository_credential_library.go
mgaffney: We do [this](https://github.com/hashicorp/boundary/blob/93d81157758398255baf9715a38fa0b381d975cb/internal/host/static/repository_host_catalog.go#L37) in [other](https://github.com/hashicorp/boundary/blob/93d81157758398255baf9715a38fa0b381d975cb/internal/host/static/repository_host_catalog.go#L71) places too. If I remember correctly, the first clone prevents changes to the object passed in and the second clone is needed for the `w.Create` method.

PR #1166, opened 2021-04-23
Add monitor job loop

louisruch:
### What does this PR do

- Adds interrupt run repo function and query
- Adds job monitor loop to update progress of running jobs and interrupt defunct job runs
- Replaces running job with a sync.Map
- Removes scheduler shutdown


louisruch:
@mgaffney @jefferai PTAL 
This PR adds the interrupt and update progress logic as well as addresses some comments from @jefferai review of my last PR.  This does not yet address to comments around `code` or how jobs are registered.  Will discuss with you in our next sync up and then open a PR accordingly.

internal/scheduler/job/repository_run.go
```diff
@@ -86,33 +86,37 @@ func (r *Repository) UpdateProgress(ctx context.Context, runId string, completed
...
 			rows, err := r.Query(ctx, updateProgressQuery, []interface{}{completed, total, runId})
 			if err != nil {
 				return errors.Wrap(err, op)
 			}
-			defer rows.Close()
 
 			var rowCnt int
 			for rows.Next() {
 				if rowCnt > 0 {
+					_ = rows.Close()
```
mgaffney, line 99: Why the switch away from `defer rows.Close()`?

internal/scheduler/scheduler.go
```diff
@@ -140,17 +138,33 @@ func (s *Scheduler) UpdateJobNextRun(ctx context.Context, jobId JobId, nextRunIn
...
+// The scheduler cannot be started again once the ctx is cancelled, a new scheduler will
+// need to be instantiated in order to begin scheduling again.
+func (s *Scheduler) Start(ctx context.Context) error {
+	const op = "scheduler.(Scheduler).Start"
 	if !s.started.CAS(s.started.Load(), true) {
 		s.logger.Debug("scheduler already started, skipping")
-		return
+		return nil
+	}
+
```
mgaffney, line 151: Should a check be done to make sure `ctx` has not been canceled/closed before work starts? For example:

```go
if err := ctx.Err(); err != nil {
  return err
}
```

internal/scheduler/scheduler.go
```diff
@@ -167,22 +181,22 @@ func (s *Scheduler) start() {
 
 			repo, err := s.jobRepoFn()
 			if err != nil {
-				s.logger.Error("error creating job repo", "error", err)
+				s.logger.Debug("error creating job repo", "error", err)
```
mgaffney, line 184: Why are we logging the errors at Debug level?

internal/scheduler/scheduler.go
```diff
@@ -195,70 +209,96 @@ func (s *Scheduler) start() {
 
 func (s *Scheduler) runJob(r *job.Run) error {
 	jobId := JobId(r.JobId)
-	j, ok := s.registeredJobs.Load(jobId)
-	job := j.(Job)
-	s.logger.Debug(fmt.Sprintf("starting job run %q for job (%v: %v)", r.PrivateId, jobId, job.Name()))
+	regJob, ok := s.registeredJobs.Load(jobId)
+	j := regJob.(Job)
+	s.logger.Debug("starting job run", "run id", r.PrivateId, "job id", jobId, "name", j.Name())
```
mgaffney, line 214: Will this panic if `ok` is false?

internal/scheduler/scheduler.go
```diff
@@ -195,70 +209,96 @@ func (s *Scheduler) start() {
...
-	s.l.Lock()
-	if _, ok := s.runningJobs[jobId]; ok {
-		s.l.Unlock()
-		return fmt.Errorf("job (%v: %v) is already running", jobId, job.Name())
-	}
 	jobContext, jobCancel := context.WithCancel(s.baseContext)
-	s.runningJobs[jobId] = runningJob{runId: r.PrivateId, cancelCtx: jobCancel}
-	s.l.Unlock()
+	_, loaded := s.runningJobs.LoadOrStore(jobId, runningJob{runId: r.PrivateId, cancelCtx: jobCancel, status: j.Status})
+	if loaded {
```
mgaffney, line 226: Should jobCancel be called here?

internal/scheduler/job/repository_run.go
```diff
@@ -293,6 +302,68 @@ func (r *Repository) FailRun(ctx context.Context, runId string, _ ...Option) (*R
...
+				if err != nil {
+					_ = rows.Close()
+					return errors.Wrap(err, op, errors.WithMsg("unable to scan rows for job run"))
+				}
+				runs = append(runs, run)
+			}
+
+			// Write Run update to oplog
+			for _, run := range runs {
+				if err = upsertRunOplog(ctx, w, oplogWrapper, oplog.OpType_OP_TYPE_UPDATE, []string{"Status", "EndTime"}, run); err != nil {
```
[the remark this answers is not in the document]
louisruch: Yeah happy to discuss this, I originally thought of trying to create multiple messages, but the metadata for each message is different looking at [WriteOplogEntryWith](https://github.com/hashicorp/boundary/blob/main/internal/db/read_writer.go#L136) I figured the idea here was if the metadata is the same but we have multiple events...

internal/scheduler/scheduler.go
```diff
@@ -60,13 +60,15 @@ func New(serverId string, jobRepoFn jobRepoFactory, logger hclog.Logger, opt ...
...
-		runJobsInterval: opts.withRunJobInterval,
+		serverId:           serverId,
+		jobRepoFn:          jobRepoFn,
+		logger:             logger,
+		registeredJobs:     new(sync.Map),
+		runningJobs:        new(sync.Map),
+		runJobsLimit:       opts.withRunJobsLimit,
+		runJobsInterval:    opts.withRunJobInterval,
+		monitorInterval:    opts.withMonitorInterval,
+		interruptThreshold: opts.withInterruptThreshold,
```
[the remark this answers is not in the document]
louisruch: ty

internal/scheduler/job/repository_run.go
louisruch: I need to close the rows for the reader to do `LookupById` below so I could not wait for defer

internal/scheduler/scheduler.go
```diff
@@ -140,17 +138,33 @@ func (s *Scheduler) UpdateJobNextRun(ctx context.Context, jobId JobId, nextRunIn
...
+	}
+
+	// Interrupt all runs previously allocated to this server
+	_, err = repo.InterruptRuns(ctx, 0, job.WithServerId(s.serverId))
+	if err != nil {
+		return errors.Wrap(err, op)
 	}
 
-	s.baseContext, s.baseCancel = context.WithCancel(ctx)
+	s.baseContext = ctx
```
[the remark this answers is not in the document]
louisruch: This was a change from having a next context created here and it was a simple change to just do `s.baseContext = ctx` so went with it.  Now that you bring it up I could just pass this context into the goroutines that need it instead of pulling it from the struct! ty

internal/scheduler/scheduler.go
louisruch: ty, I made all the logs `Error` because I was not sure how to set debug level in unit tests so I could see the logs to make sure it was all working.  I must have made a grep error :/

internal/scheduler/scheduler.go
louisruch: It will, however I was not worried about it because the map is private and controlled by the loop so it will never not be `ok`.  I guess it doesn't hurt to add the check though...

mgaffney: The check is already on line 215

internal/scheduler/scheduler.go
louisruch: Good question, I was not sure if I wanted to cancel the running job at this point.  The only reason this would happen is if the job has been finished (failed, completed or interrupted)  So in theory the only way we can get here is a race or another server interrupted this.  So perhaps it should be cancelled yes!

internal/scheduler/scheduler.go
louisruch: Oh I see what you are saying, my bad!  I thought you where taking about `j, ok := regJob.(Job)` vs `j := regJob.(Job)`

internal/scheduler/scheduler.go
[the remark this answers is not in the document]
louisruch: Just did this, so clean TY!

jimlambrt (changes requested):
just one suggestion.

internal/scheduler/job/repository_run.go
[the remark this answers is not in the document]
louisruch: hmm this doesn't seem to be the problem I thought it was.  I was initially hitting a `db in bad state` error if I did not close the rows before trying to use the reader again, i.e:

```
rows, err := r.Query(...)
defer rows.Close()
//use rows
err = r.LookupById(...)
// this returned the bad state error because defer was not called and rows were not closed yet
```

I am not hitting this any more so defer seems to work well.

mgaffney (approved):
LGTM!

PR #1168, opened 2021-04-25
Ongoing OIDC: incorporate ClaimsScopes into the oidc repository

jimlambrt:


internal/auth/oidc/auth_method.go
```diff
@@ -106,6 +106,10 @@ func NewAuthMethod(scopeId string, clientId string, clientSecret ClientSecret, o
 			a.SigningAlgs = append(a.SigningAlgs, string(alg))
 		}
 	}
+	if len(opts.withClaimsScopes) > 0 {
+		a.ClaimsScopes = make([]string, 0, len(opts.withClaimsScopes))
```
jefferai, line 110: I realize it gets round-tripped back through the auth method but I guess I'd assume that what went into the options is unlikely to be used further...do you think it's safe to avoid this copy?

jefferai (approved):
Looks great!

internal/auth/oidc/auth_method.go
jimlambrt: I agree, that it's not needed.  ty!

internal/auth/oidc/auth_method.go
```diff
@@ -270,9 +270,10 @@ func (am *AuthMethod) isComplete() error {
 }
 
 type convertedValues struct {
-	Algs  []interface{}
-	Auds  []interface{}
-	Certs []interface{}
+	Algs   []interface{}
+	Auds   []interface{}
+	Certs  []interface{}
+	Scopes []interface{}
```
talanknight, line 276: Consider renaming Scopes to ClaimScopes since the name Scope collides with a concept in boundary and the convertedValue struct does not only capture claim values.

internal/auth/oidc/repository_auth_method_read.go
```diff
@@ -187,6 +190,7 @@ type authMethodAgg struct {
 	ApiUrl                            string
 	Auds                              string
 	Certs                             string
+	Scopes                            string
```
talanknight, line 193: nit: consider renaming to ClaimScopes since there is the naming collision with boundary scopes.

internal/auth/oidc/service_token_request_test.go
```diff
@@ -234,7 +234,7 @@ func Test_TokenRequest(t *testing.T) {
 			atRepoFn:     atRepoFn,
 			authMethodId: testAuthMethod.PublicId,
 			tokenRequest: func() string {
-				exp, err := ptypes.TimestampProto(time.Now().Add(AttemptExpiration).Truncate(time.Second))
+				exp := timestamppb.New(time.Now().Add(AttemptExpiration).Truncate(time.Second))
 				require.NoError(t, err)
```
talanknight, line 238: You can remove the NoError check following both of these timestamppb.New() calls here and elsewhere in this PR.

internal/db/schema/migrations/postgres/5/01_oidc.up.sql
```diff
@@ -0,0 +1,85 @@
+begin;
```
talanknight, line 1: Should we have waited to merge `/4/` until this PR was in since the user wouldn't have been able to show Email or FullName before this anyways?

talanknight (approved):
Looks good!

internal/auth/oidc/auth_method.go
jimlambrt: ty!

internal/auth/oidc/repository_auth_method_read.go
jimlambrt: ty!

internal/auth/oidc/service_token_request_test.go
jimlambrt: ty

internal/db/schema/migrations/postgres/5/01_oidc.up.sql
jimlambrt: It's a good question.  Whether or not Email/FullName showed would depend on the Provider implementation.  For Okta and Auth0 they wouldn't have shown, but you could have implemented Dex as a broker and returned whatever you'd like for claims with just the default `openid` scope.

PR #1172, opened 2021-04-26
packagespec upgrade -version 0.2.4

samsalisbury:
Also upgrades circleci-multi-file-config meaning that `make ci-config` will be much more reliable in future, and will no longer incorrectly report "nothing to be done" when the config needs updating.


claire-labry (approved):
LGTM

PR #1174, opened 2021-04-27
OIDC Callback Error Redirect

talanknight:


internal/servers/controller/handlers/marshler.go
```diff
@@ -0,0 +1,19 @@
+package handlers
+
+import (
+	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+var JSONMarshaler = &runtime.JSONPb{
```
jimlambrt, line 8: why not just provide a function that returns the needed runtime.JSONPb?  Is there some reason all of them need to share the same instance?

jimlambrt (changes requested):
looks good.  just a few suggestions.

internal/errors/code.go
```diff
@@ -85,4 +85,7 @@ const (
 
 	// External system errors are reserved codes 3000-3999
 	Unavailable Code = 3000 // Unavailable represents that an external system is unavailable
+
+	// OIDC authentication provided errors
+	OidcProviderCallbackError Code = 4000
```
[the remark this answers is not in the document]
talanknight: Thank you!

internal/servers/controller/handlers/marshler.go
talanknight: Thank you!

jimlambrt (approved):

PR #1175, opened 2021-04-27
Ongoing OIDC: support to request additional OIDC scopes from the IdP

jimlambrt:


internal/db/schema/migrations/postgres/3/01_oidc.up.sql
```diff
@@ -0,0 +1,36 @@
+begin;
+
+
+-- auth_oidc_scope entries are the optional scopes for a specific oidc auth
+-- method.  There can be 0 or more for each parent oidc auth method.  If an auth
+-- method has any scopes, they will be added to provider requests along with the
+-- default of "openid". 
+create table auth_oidc_scope (
```
mgaffney, line 8: Why are we updating the `migrations/postgres/3` directory with another `01` migration? Also, isn't the "3" already in main?

jimlambrt: ty!  I noticed this as well.  Looks like I moved 01_oidc.up.sql to `5` but failed to delete it in `3` somehow.  I'm backtracking through the branches to figure this out.

jimlambrt: Yep.. the original schema changes PR up the migration in `3` and I moved it in a subsequent PR to `5` and added more to it.

PR #1179, opened 2021-04-29
Initial Output Fields Work

jefferai:
This doesn't update all handlers, it does just enough to show the output fields paradigm soup to nuts -- acls/perms, auth, handlers, CLI. Specifically to the handlers, it shows:

* A non-top-level resource
* A top level resource including ensuring output fields are respected for recursive listing
* A resource that does a Self pivot, e.g. ReadSelf -> Read

One important thing to note: if we don't always include `id` and (if applicable) `scope_id`, then we actually cannot look up authorized actions. So I made the decision to simply have the handlers always include those two values. It seems like a decent tradeoff.

Everything that will be changed after this will be in the same model (CLI, handlers). Also documentation will come later. There are also some opportunities for cleanup opened up here that I'm not going to tackle right now.


internal/auth/testing.go
```diff
@@ -3,6 +3,7 @@ package auth
 import (
 	"context"
 
+	"github.com/hashicorp/boundary/internal/requests"
 	"github.com/hashicorp/boundary/internal/servers/controller/common"
 )
```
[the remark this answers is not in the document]
jimlambrt: it might be nice have some tests for this func... I know, it's executed a lot, so feel free to disregard.

internal/perms/output_fields.go
```diff
@@ -0,0 +1,94 @@
...
+	for f := range o {
+		ret = append(ret, f)
+	}
+	sort.Strings(ret)
+	return
+}
+
+// SelfOrDefaults returns either the fields map itself or the defaults for the
+// given user
+func (o OutputFieldsMap) SelfOrDefaults(userId string) OutputFieldsMap {
```
jimlambrt, line 58: fyi, there's no coverage for this in the perms package unit tests.

internal/requests/requests.go
```diff
@@ -0,0 +1,63 @@
...
+
+// ContextRequestInforation is a type used solely for context keys -- see the
+// variable below
+type ContextRequestInformation struct{}
+
+// ContextRequestInformationKey is a value to keep linters from complaining
+// about clashing identifiers
+var ContextRequestInformationKey ContextRequestInformation
+
+// RequestContext is used to propoagate request information. It can be updated
```
jimlambrt, line 17: sp of propoagate?

internal/requests/requests.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// UserId contains the final discovered user ID
+	UserId string
+
+	// OutputFields is the set of fields authorized for output for the
+	// authorized action, if not the default
+	OutputFields perms.OutputFieldsMap
+}
+
+func NewRequestContext(parent context.Context) context.Context {
```
jimlambrt, line 40: Do we want unit tests for NewRequestContext(...), RequestContextFromCtx(...) and OutputFields(...)?

internal/requests/requests.go
```diff
@@ -0,0 +1,63 @@
...
+	return context.WithValue(parent, ContextRequestInformationKey, &RequestContext{})
+}
+
+// RequestContextFromCtx pulls out RequestContext and returns it. It will always
```
jimlambrt, line 44: The downside to this design is: the caller may believe that the data they add to RequestContext is propagated and there's no way to tell if it is or isn't.

jimlambrt: you could keep it very similar and just return a bool if the RequestContext was found.  Then it's up to the caller to determine the proper thing to do.

internal/requests/requests.go
```diff
@@ -0,0 +1,63 @@
...
+		return &RequestContext{}
+	}
+	reqCtx, ok := reqCtxRaw.(*RequestContext)
+	if !ok {
+		return &RequestContext{}
+	}
+	return reqCtx
+}
+
+func OutputFields(ctx context.Context) perms.OutputFieldsMap {
```
jimlambrt, line 59: sort of the same comment here... do we want to return a bool too?

internal/servers/controller/handlers/authmethods/authmethod_service.go
```diff
@@ -63,29 +65,30 @@ var (
```
jimlambrt, line 65: Out of scope of this PR, but why not wrap these package vars (IdActions, CollectionActions, collectionTypeMap) in funcs?  I'm happy to take this work on.  pkg maps just seem to be a data race or panic waiting to happen in the future.

internal/servers/controller/handler.go
```diff
@@ -227,6 +228,18 @@ func wrapHandlerWithCommonFuncs(h http.Handler, c *Controller, props HandlerProp
...
 		ctx = auth.NewVerifierContext(ctx, c.logger, c.IamRepoFn, c.AuthTokenRepoFn, c.ServersRepoFn, c.kms, requestInfo)
 
+		// Add general request information to the context. The information from
+		// the auth verifier context is pretty specifically curated to
+		// authentication/authorization verification so this is more
+		// general-purpose.
+		//
+		// We could use requests.NewRequestContext but this saves an immediate
+		// lookup.
+		ctx = context.WithValue(ctx, requests.ContextRequestInformationKey, &requests.RequestContext{
```
jimlambrt, line 238: see my comment about using an option to NewRequestContext(...)

internal/auth/auth.go
```diff
@@ -158,6 +159,12 @@ func Verify(ctx context.Context, opt ...Option) (ret VerifyResults) {
 
 	ret.Scope = new(scopes.ScopeInfo)
 
+	var reqInfo *requests.RequestContext
+	reqInfoRaw := ctx.Value(requests.ContextRequestInformationKey)
```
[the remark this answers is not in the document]
jefferai: Done.

internal/auth/testing.go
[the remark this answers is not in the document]
jefferai: Added supported options. As for testing for the func, I'll leave it for now. This exists basically only for tests so the tests are themselves testing this a _lot_.

internal/requests/requests.go
jefferai: Erp. Fixed.

internal/perms/output_fields.go
jefferai: Good catch. Added.

internal/servers/controller/handlers/accounts/account_service.go
```diff
@@ -152,13 +175,24 @@ func (s Service) GetAccount(ctx context.Context, req *pbs.GetAccountRequest) (*p
...
+
+	outputFields := requests.OutputFields(ctx)
+	item, err := toProto(ctx, acct, handlers.WithOutputFields(&outputFields))
 	if err != nil {
 		return nil, err
 	}
-	u.Scope = authResults.Scope
-	u.AuthorizedActions = authResults.FetchActionSetForId(ctx, u.Id, IdActions[auth.SubtypeFromId(u.Id)]).Strings()
-	return &pbs.GetAccountResponse{Item: u}, nil
+	if outputFields.Has(globals.ScopeField) {
```
talanknight, line 188: It would be nice to have all the outgoing proto building logic in one place instead of partially inside toProto and partially inside the caller.  Consider passing the values for these fields into the toProto (maybe using options, or maybe some other way).

talanknight (commented):
Still looking at the PR (just started the authmethod_service now), but wanted to get some of my comments out there before I continue on.

internal/requests/requests.go
jefferai: Will do.

globals/fields.go
```diff
@@ -0,0 +1,23 @@
+package globals
+
+const (
```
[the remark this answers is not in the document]
jefferai: We already use globals. This is just in a package called "globals" so that we acknowledge that fact. :-)

internal/servers/controller/handlers/authmethods/authmethod_service.go
jefferai: We could. Honestly they should be const, but you can't declare maps as const :-/ Won't do it in this PR though.

internal/perms/output_fields.go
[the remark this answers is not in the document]
jefferai: There already are

internal/servers/controller/handlers/authmethods/authmethod_service.go
```diff
@@ -446,51 +533,53 @@ func (s Service) createInRepo(ctx context.Context, scopeId string, item *pb.Auth
...
-				attrs := &pb.OidcAuthMethodAttributes{}
-				if err := handlers.StructToProto(am.GetAttributes(), attrs); err != nil {
-					return nil, errors.Wrap(err, op, errors.WithMsg("can't convert from attribute struct to proto"))
-				}
-				attrs.DryRun = true
-				if am.Attributes, err = handlers.ProtoToStruct(attrs); err != nil {
-					return nil, errors.Wrap(err, op, errors.WithMsg("can't convert from attribute proto to struct"))
-				}
-				return am, nil
+		if dr {
```
talanknight, line 591: This whole code block doesn't seem relevant anymore since we do not use any of the built wire proto.

internal/servers/controller/handlers/authmethods/authmethod_service.go
```diff
@@ -246,12 +302,32 @@ func (s Service) UpdateAuthMethod(ctx context.Context, req *pbs.UpdateAuthMethod
...
+	}
+	if outputFields.Has(globals.AuthorizedCollectionActionsField) {
+		collectionActions, err := populateCollectionAuthorizedActions(ctx, authResults, am)
+		if err != nil {
+			return nil, err
+		}
+		item.AuthorizedCollectionActions = collectionActions
+	}
+
+	if item.GetAttributes() != nil && dryRun {
```
talanknight, line 344: The more I look at this the more I think it would be better if it could be handled in the toProto method.  That is the whole purpose of that, and now that it accepts a variadic option param it probably isn't hard to add these and the authorized action stuff.

internal/perms/output_fields.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/hashicorp/boundary/globals"
+)
+
+// OutputFieldsMap is used to store information about allowed output fields in
+// grants
+type OutputFieldsMap map[string]bool
+
+// AddStrings adds the given fields and returns the map and whether or not the
+// input included "*", which is used to shortcut some logic when checking ACL.
+func (o OutputFieldsMap) AddFields(input []string) (ret OutputFieldsMap, starField bool) {
```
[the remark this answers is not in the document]
jefferai: Donezo

internal/perms/output_fields.go
```diff
@@ -0,0 +1,94 @@
...
+	"sort"
+
+	"github.com/hashicorp/boundary/globals"
+)
+
+// OutputFieldsMap is used to store information about allowed output fields in
+// grants
+type OutputFieldsMap map[string]bool
+
+// AddStrings adds the given fields and returns the map and whether or not the
```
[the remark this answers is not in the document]
jefferai: It thinks this is updated so can't apply it, but I will make this change manually.

internal/auth/additional_verification_test.go
```diff
@@ -106,3 +113,150 @@ func TestFetchActionSetForId(t *testing.T) {
...
+
+	amClient := authmethods.NewClient(tc.Client())
+	resp, err := amClient.List(tc.Context(), scope.Global.String(), authmethods.WithRecursive(true))
+	require.NoError(err)
+	require.NotNil(resp)
+	require.NotNil(resp.GetItems())
+	assert.Len(resp.GetItems().([]*authmethods.AuthMethod), 5)
+	items := resp.GetResponse().Map["items"].([]interface{})
+	require.NotNil(items)
+	for _, item := range items {
```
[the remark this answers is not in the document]
jefferai: Will do.

internal/servers/controller/handlers/accounts/account_service.go
jefferai: We were already doing this (toProto was just being called in a helper function) so it's not a new paradigm. I agree that it could be made better, but one reason I didn't do this now is I'm not sure `handlers` is the right place for options here and so I didn't want to unilaterally expand options there without getting some feedback. Once we figure out where it should all live it's easy enough to sweep through and update all of this.

globals/fields.go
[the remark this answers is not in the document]
talanknight: Understood, but most of our previous globals were internal and were kept relatively close to where they were actually used (maybe not always... but hopefully more often than not).  Since these are exported values in a non-internal directory I would prefer better documentation on these fields.

internal/servers/controller/handlers/authmethods/authmethod_service.go
jefferai: I am honestly not sure what the dry run stuff was so I'm happy to take it out if you and Jim concur but for the moment it works in its updated form.

internal/servers/controller/handlers/authmethods/authmethod_service.go
jefferai: I'll address this in the next PR.

internal/servers/controller/handlers/authmethods/authmethod_service.go
jefferai: Removed.

internal/servers/controller/handlers/accounts/account_service.go
jefferai: I'll address this in the next PR.

internal/auth/additional_verification_test.go
```diff
@@ -106,3 +113,150 @@ func TestFetchActionSetForId(t *testing.T) {
 		})
 	}
 }
+
+func TestRecursiveListingDifferentOutputFields(t *testing.T) {
```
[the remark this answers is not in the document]
jefferai: Adjusted tests to verify default visibility behavior on the global scope.

PR #1182, opened 2021-04-30
Drop job code, add job plugin

louisruch:
### What does this PR do

- Drops the concept of a job code
- Adds a new `job_plugin` table.  
- `job_plugin` has not been flushed out yet, it is still unclear exactly what needs to be added to the table so it only has a `publicId` for now
- `jpi_internal` is the default `plugin_id` and is the only valid plugin id.  It represents internal jobs run by the controller.
- Once plugins are flushed out, we will add a `WithPluginId` to job repo and scheduler


louisruch:
@mgaffney PTAL, you only need to review c92dbed as the rest is branched from #1166

mgaffney (approved):
LGTM!

PR #1185, opened 2021-05-01
Incorrect Grants when setting anonymous read permissions

Theragus:
when setting the grants for the anonymous listing role it's only setting is 'type=scope;actions=list'. But I think it should be 'id=*;type=scope;actions=list,read'

With 'type=scope;actions=list', it's not possible to read the orgs in the global scope unauthenticated.

when initializing a server without skipping the default roles and in the dev server these permission are also set to 'id=*;type=scope;actions=list,read'

### DIFF
--- a/website/content/docs/installing/no-gen-resources.mdx
+++ b/website/content/docs/installing/no-gen-resources.mdx
@@ -318,7 +318,7 @@ resource "boundary_role" "global_anon_listing" {
   scope_id = "global"
   grant_strings = [
     "id=*;type=auth-method;actions=list,authenticate",
-    "id=*;type=scope;actions=list,read",
+    "id=*;type=scope;actions=list,no-op",
     "id={{account.id}};actions=read,change-password"
   ]
   principal_ids = ["u_anon"]
--- a/website/content/docs/installing/no-gen-resources.mdx
+++ b/website/content/docs/installing/no-gen-resources.mdx
@@ -302,7 +302,7 @@ $ boundary roles create -name 'global_anon_listing' \
 $ boundary roles add-grants -id <global_anon_listing_id> \
   -recovery-config /tmp/recovery.hcl \
   -grant 'id=*;type=auth-method;actions=list,authenticate' \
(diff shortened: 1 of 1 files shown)


randallmorey:
Hi there and thanks for your pull request.  You correctly noticed that a `list` grant alone will not result in the resource being listed.  In order to be listed, a grant must exist on at least one of the resource instances.  Based on existing permissions, you might add `read` to achieve listings.  But this may sometimes be more permission than you wish to grant.

We'll be introducing a new grant `no-op` very soon that helps solve this problem.  This grant provides no meaningful permissions except in the context of `list`.  Resources may thus be listed _without_ being individually readable.

We expect this to land in our next release.  Let's wait on this PR until then.  Thank you!

website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -302,7 +302,7 @@ $ boundary roles create -name 'global_anon_listing' \
 $ boundary roles add-grants -id <global_anon_listing_id> \
   -recovery-config /tmp/recovery.hcl \
   -grant 'id=*;type=auth-method;actions=list,authenticate' \
-  -grant 'type=scope;actions=list' \
+  -grant 'id=*;type=scope;actions=list,read' \
```
jefferai, line 305:
```suggestion
  -grant 'id=*;type=scope;actions=list,no-op' \
```

website/content/docs/installing/no-gen-resources.mdx
```diff
@@ -318,7 +318,7 @@ resource "boundary_role" "global_anon_listing" {
   scope_id = "global"
   grant_strings = [
     "id=*;type=auth-method;actions=list,authenticate",
-    "type=scope;actions=list",
+    "id=*;type=scope;actions=list,read",
```
jefferai, line 321:
```suggestion
    "id=*;type=scope;actions=list,no-op",
```

jefferai:
I'm going to make that change and commit as we're close to 0.2.1; this will get deployed at that time. Thanks!

PR #1186, opened 2021-05-01
OIDC: add support for account claim maps

jimlambrt:
Account claim maps are optional claim maps from custom claims to the standard claims of sub, name and email.  These maps are represented as key=value where the key equals the from_claim and the value equals the to_claim.  For example "oid=sub".

We've added this new feature to support Providers like Azure, which have non-deterministic/discoverable subjects claims.  Azure for instance, provides no method to determine a user's subject within a relying party unless you have the user's id_token.  This makes it impossible to pre-provision Azure OIDC accounts within Boundary unless you support mapping some other discoverable claim to the account's subject.  Here is where account claim maps come to play.

One important note:  Boundary admins are free create new OIDC auth methods with an account claim map for the to_claim of `sub`.  But once an auth-method has been created, Boundary will restrict any updates that would change the account claim maps for the to_claim of `sub`.  If you need to change which inbound claims are mapped to the `sub` claim, you'll need to recreate the auth-method with the mapping required for the claim `sub`.   The other maps to `name` and `email` are unrestricted and can be updated anytime.


internal/db/schema/migrations/postgres/6/01_oidc.up.sql
```diff
@@ -0,0 +1,92 @@
...
+    constraint auth_oidc_method_fkey
+    references auth_oidc_method(public_id)
+    on delete cascade
+    on update cascade,
+  from_claim text not null
+    constraint from_claim_must_not_be_empty
+       check(length(trim(from_claim)) > 0) 
+    constraint from_claim_must_be_less_than_1024_chars
+      check(length(trim(from_claim)) < 1024),
+  to_claim text not null constraint to_claim_valid_values CHECK (to_claim IN ('sub', 'name', 'email')), -- intentionally case-sensitive matching
```
mgaffney, line 18: nit: can you format this to match the other check constraints?

jimlambrt: happy too.  ty!

jimlambrt:
tagging @stellarsquall @randallmorey for the UI and documentation impacts.

internal/auth/oidc/repository_auth_method_test.go
```diff
@@ -130,7 +130,7 @@ func Test_upsertAccount(t *testing.T) {
 			idClaims:        map[string]interface{}{"iss": "https://alice.com"},
 			atClaims:        map[string]interface{}{},
 			wantErrMatch:    errors.T(errors.Unknown),
-			wantErrContains: "subject is not present",
+			wantErrContains: "to account subject and it is not present in ID Token",
```
talanknight, line 133: Can you add a test which performs the actual mapping and verify things work as expected?

internal/cmd/commands/authmethodscmd/oidc_funcs.go
```diff
@@ -129,6 +132,12 @@ func extraOidcFlagsFuncImpl(c *OidcCommand, set *base.FlagSets, _ *base.FlagSet)
 				Target: &c.flagClaimsScopes,
 				Usage:  `The optional claims scope requested. May be specified multiple times.`,
 			})
+		case accountClaimMaps:
+			f.StringSliceVar(&base.StringSliceVar{
+				Name:   accountClaimMaps,
+				Target: &c.flagAccountClaimMaps,
+				Usage:  `The optional account claim maps from custom claims to the standard claims of sub, name and email.  These maps are represented as key=value where the key equals the from-claim and the value equals the to-claim.  For example "oid=sub". May be specified multiple times.`,
```
talanknight, line 139:
```suggestion
				Usage:  `The optional account claim maps from custom claims to the standard claims of sub, name and email.  These maps are represented as key=value where the key equals the from-claim and the value equals the to-claim.  For example "oid=sub". May be specified multiple times for different to-claims.`,
```

talanknight: I'm not sure what good wording would be but consider calling out that the `from-claim` is the claim from the provider and the `to-claim` is what boundary treats the claim as.

internal/cmd/commands/authmethodscmd/oidc_funcs.go
```diff
@@ -42,6 +43,7 @@ const (
 	caCertFlagName                            = "idp-ca-cert"
 	allowedAudienceFlagName                   = "allowed-audience"
 	claimsScopes                              = "claims-scopes"
+	accountClaimMaps                          = "account-claim-maps"
```
talanknight, line 46: Honest question since I'm not as familiar with oidc as you,  Is the `account-` prefix needed and useful here?

internal/servers/controller/handlers/authmethods/authmethod_service.go
```diff
@@ -814,6 +815,12 @@ func validateCreateRequest(req *pbs.CreateAuthMethodRequest) error {
 						}
 					}
 				}
+				if len(attrs.GetAccountClaimMaps()) > 0 {
+					_, err := oidc.ParseAccountClaimMaps(attrs.GetAccountClaimMaps()...)
```
talanknight, line 819: Please confirm that multiple kv pairs do not map to the same ToClaim since the errors returned here can be nicer than when returned by the DB.  Both here and in update.

api/authmethods/option.gen.go
```diff
@@ -87,6 +87,30 @@ func WithRecursive(recurse bool) Option {
 	}
 }
 
+func WithOidcAuthMethodAccountClaimMaps(inAccountClaimMaps []string) Option {
```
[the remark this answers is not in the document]
jimlambrt: sticking with the kvbuilder from the shared lib for this.

internal/auth/oidc/account_claim_map.go
```diff
@@ -0,0 +1,103 @@
...
+	if cs.OidcMethodId == "" {
+		return errors.New(errors.InvalidParameter, caller, "missing oidc auth method id")
+	}
+	if cs.FromClaim == "" {
+		return errors.New(errors.InvalidParameter, caller, "missing from claim")
+	}
+	switch cs.ToClaim {
+	case "":
+		return errors.New(errors.InvalidParameter, caller, "missing to claim")
+	case "sub", "email", "profile":
```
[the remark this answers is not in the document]
jimlambrt: there are better ways of doing this.  ty!

internal/auth/oidc/repository_auth_method.go
```diff
@@ -32,13 +32,39 @@ func (r *Repository) upsertAccount(ctx context.Context, am *AuthMethod, IdTokenC
 	if AccessTokenClaims == nil {
 		return nil, errors.New(errors.InvalidParameter, op, "missing Access Token claims")
 	}
+
+	fromSub, fromName, fromEmail := "sub", "name", "email"
```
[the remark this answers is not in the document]
jimlambrt: ty!

internal/cmd/commands/authmethodscmd/oidc_funcs.go
jimlambrt: not really but I'm trying to be clear that it's the claim maps that will be used for the account.

internal/cmd/commands/authmethodscmd/oidc_funcs.go
jimlambrt: ty!

internal/auth/oidc/repository_auth_method_test.go
jimlambrt: ty!

internal/servers/controller/handlers/authmethods/authmethod_service.go
jimlambrt: ty!

internal/db/schema/migrations/postgres/6/01_oidc.up.sql
```diff
@@ -0,0 +1,94 @@
...
+
+
+create trigger
+  immutable_columns
+before
+update on auth_oidc_account_claim_map
+  for each row execute procedure immutable_columns('oidc_method_id', 'from_claim', 'to_claim', 'create_time');
+  
+
+  -- we will drop the oidc_auth_method_with_value_obj view, so we can recreate it
```
talanknight, line 41:
```suggestion
-- we will drop the oidc_auth_method_with_value_obj view, so we can recreate it
```

internal/db/schema/migrations/postgres/6/01_oidc.up.sql
```diff
@@ -0,0 +1,94 @@
...
+    on delete cascade
+    on update cascade,
+  from_claim text not null
+    constraint from_claim_must_not_be_empty
+       check(length(trim(from_claim)) > 0) 
+    constraint from_claim_must_be_less_than_1024_chars
+      check(length(trim(from_claim)) < 1024),
+  to_claim text not null 
+    constraint to_claim_valid_values 
+      CHECK (to_claim IN ('sub', 'name', 'email')), -- intentionally case-sensitive matching
```
talanknight, line 20:
```suggestion
      check (to_claim in ('sub', 'name', 'email')), -- intentionally case-sensitive matching
```

talanknight (dismissed):
Looks good! Just 2 small changes to the sql files.

PR #1188, opened 2021-05-02
OIDC: make callback URL paths deterministic. 

jimlambrt:
All ODIC callbacks will now have the same predictable PATH:  `/v1/auth-methods/oidc:authenticate:callback`

@stellarsquall @randallmorey -> this has positive impacts on the admin UI and docs.


jefferai (approved):
Looks great!

PR #1197, opened 2021-05-04
Louis job scheduler

louisruch:
### What does this PR do

- Adds job repo
- Adds scheduler 
- The `job_plugin` table is a place holder for future expansion of the scheduler.  Currently all jobs that are created are linked to the `jpi_internal` plugin which are jobs that are held in a local `registeredJobs` map populated at controller startup.


louisruch:
@jimlambrt @mgaffney @jefferai PTAL, this will merge all the job repo and scheduler work into main, which currently does not run any jobs.  Work will start shortly to create a vault renewal job.

internal/servers/controller/controller.go
```diff
@@ -146,6 +158,9 @@ func (c *Controller) Start() error {
 		return nil
 	}
 	c.baseContext, c.baseCancel = context.WithCancel(context.Background())
+	if err := c.scheduler.Start(c.baseContext); err != nil {
```
louisruch, line 161: Since the scheduler will currently not run any jobs (non exist). Should I replace this with a `TODO` so we don't have the goroutine running for now?

internal/db/schema/migrations/postgres/7/02_job.up.sql
```diff
@@ -0,0 +1,124 @@
...
+
+    create table job (
+         plugin_id wt_public_id not null
+             constraint job_plugin_fk
+                 references job_plugin(public_id)
+                 on delete cascade
+                 on update cascade,
+         name wt_name not null,
+         description wt_description not null,
+         next_scheduled_run timestamp with time zone not null,
```
jefferai, line 22: I think this should be `wt_timestamp`

mgaffney: @louisruch , correct me if I'm wrong, but I think this is not a `wt_timestamp` because `wt_timestamp` uses `current_timestamp` as a [default value](https://github.com/hashicorp/boundary/blob/3c994f66f877224fd1d75e6ce3ff4efb3aa9a0ad/internal/db/schema/migrations/postgres/0/01_domain_types.up.sql#L43) which is not the behavior wanted here.

mgaffney (approved):
Nice work!

internal/db/schema/migrations/postgres/7/02_job.up.sql
louisruch: Oh sorry I never saw this, yeah @mgaffney is correct we wanted to avoid the default behavior.

louisruch:
@mgaffney @jefferai as discuss re oplog ebb8774

internal/db/schema/migrations/postgres/7/02_job.up.sql
jefferai: Ah OK, sounds good. I saw wt_timestamp was with time zone not null and forgot about the default.

internal/scheduler/job/options.go
```diff
@@ -0,0 +1,80 @@
+package job
+
+import (
+	"time"
+)
+
+const (
+	defaultRunJobsLimit = 1
+	defaultPluginId     = "jpi_internal"
```
jefferai, line 9: At the risk of bikeshedding -- if you think about this as "the thing that is registering a job", I think `system` is a better choice -- the system itself is registering the job, as opposed to `internal` which could indicate "an internal registration" but not _what_ registered it.

Separately, if this is likely tied to a plugin, `jpi` ties it rather closely to the job system as opposed to the ID of the plugin itself. If that's not intentional, probably just `pi` is a better choice.

louisruch: Sure pi_system works, given the rest of your comment I assume I should also rename the job_plugin table to just plugin then.

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,293 @@
...
+	s.logger.Debug("starting job run", "run id", r.PrivateId, "job name", r.JobName)
+
+	repo, err := s.jobRepoFn()
+	if err != nil {
+		return fmt.Errorf("failed to create job repository: %w", err)
+	}
+
+	j := regJob.(Job)
+	jobContext, jobCancel := context.WithCancel(ctx)
+	_, loaded := s.runningJobs.LoadOrStore(r.JobName, runningJob{runId: r.PrivateId, cancelCtx: jobCancel, status: j.Status})
```
jefferai, line 213: Is there a reason you're not storing a pointer to the runningJob struct? It doesn't strictly need to be one, but you could then modify the values after the fact if you did so. For instance, you could create the context only if loaded, so not have to ensure that it's canceled in all subsequent paths.

internal/servers/controller/controller.go
jefferai: I think it's fine as-is.

internal/scheduler/job/options.go
jefferai: Yeah I think so

jefferai:
Just those few comments...looking great! I didn't look deeply at the tests but they seem quite comprehensive :-D

louisruch:
@jefferai 43f6661

internal/db/schema/migrations/postgres/7/02_job.up.sql
```diff
@@ -1,20 +1,28 @@
...
+
+    create table plugin (
+        public_id wt_plugin_id primary key
     );
 
-    insert into job_plugin (public_id)
+    insert into plugin (public_id)
     values
-        ('jpi_internal');
+        ('pi_system');
```
mgaffney, line 17: 1. Should a trigger be added to prevent `pi_system` from being deleted?
2. And should `plugin_id` be an immutable column?

mgaffney (changes requested):

internal/db/schema/migrations/postgres/7/02_job.up.sql
jefferai: I think so, for both

louisruch:
@mgaffney 0d84a5d

mgaffney (dismissed):

PR #1199, opened 2021-05-04
Create Credential Store Resource in the API

talanknight:
Adding the credential store resource in the API along with mask mapping and pathing.


talanknight:
@mgaffney, I created this as a draft PR so we can discuss naming and structure of the resource.

internal/proto/local/controller/storage/credential/vault/store/v1/vault.proto
```diff
@@ -42,30 +42,30 @@ message CredentialStore {
...
   // @inject_tag: `gorm:"default:null"`
-  string tls_server_name = 11;
+  string tls_server_name = 11 [(custom_options.v1.mask_mapping) = {this:"TlsServerName" that: "attributes.tls_server_name"}];
 
   // tls_skip_verify disables verification of TLS certificates. Using this
   // option is highly discouraged as it decreases the security of data
   // transmissions to and from the Vault server.
   // @inject_tag: `gorm:"default:false"`
-  bool tls_skip_verify = 12;
+  bool tls_skip_verify = 12 [(custom_options.v1.mask_mapping) = {this:"TlsSkipVerify" that: "attributes.disable_tls_verification"}];
```
mgaffney, line 68: `tls_skip_verify` was chosen to be consistent with the name [Vault](https://www.vaultproject.io/docs/agent#tls_skip_verify), [Consul](https://www.consul.io/docs/connect/ca/vault#tlsskipverify), [Nomad](https://www.nomadproject.io/docs/configuration/vault#tls_skip_verify), and [Terraform](https://registry.terraform.io/providers/hashicorp/vault/latest/docs#skip_tls_verify) use for this parameter. Why change it here?

internal/proto/local/controller/api/resources/credentialstores/v1/credential_store.proto
```diff
@@ -0,0 +1,69 @@
...
+  // Output only. The available actions on this resource for this user.
+  repeated string authorized_actions = 300 [json_name = "authorized_actions"];
+
+  // Output only. The authorized actions for the scope's collections.
+  map<string, google.protobuf.ListValue> authorized_collection_actions = 310 [json_name = "authorized_collection_actions"];
+}
+
+// The attributes of a vault typed Credential Store.
+message VaultCredentialStoreAttributes {
+  // The complete url address of vault.
```
mgaffney, line 55: What about the `token` and `client_certificate` attributes?

mgaffney (commented):
I left a few questions

internal/proto/local/controller/storage/credential/vault/store/v1/vault.proto
talanknight: That is a good enough reason for me to keep it as `tls_skip_verify` :).  I wasn't aware of those existing fields, so I changed it to be similar to the flag we use to disable oidc config validation since there seemed to be comparable `disable_discovered_config_validation`. `

internal/proto/local/controller/api/resources/credentialstores/v1/credential_store.proto
talanknight: I've added these fields.

talanknight:
I've done a few iterations of the protos (and fixed some an issue with mask manager).  PTAL.

talanknight:
Since the discussion on this PR has died I'm going to close this PR and move our conversation to https://github.com/hashicorp/boundary/pull/1215

PR #1204, opened 2021-05-05
Fix `boundary database init` respecting format flag

jefferai:
The output was ready for json format but the flag wasn't included.


jimlambrt (approved):
ty!

PR #1210, opened 2021-05-07
Credential Library SAD repo functions

louisruch:
### What does this PR do

- Implements Set, Delete and Add target credential library repository functions
- Refactors SDA repository functions for host sets out of `repository.go` into `repository_host_set.go`


louisruch:
@mgaffney PTAL 

I decided to create a new `repository_credential_library.go` file instead of expanding `repository.go`, it felt strange having the host set functions not in its own file so I refactored them fb1bf21

The new credential library functions are in this commit c0db439

I wanted to check with you before implementing but it seems I should update `LookupTarget` and `UpdateTcpTarget` to also return a slice of credential libraries.  Perhaps we should also add a `withCredentialLibraries` option for `CreateTcpTarget`?

louisruch:
@mgaffney This should be good for review now.  @talanknight Going to start tagging you in this PRs as well.

internal/servers/controller/handlers/targets/target_service.go
```diff
@@ -204,7 +204,7 @@ func (s Service) GetTarget(ctx context.Context, req *pbs.GetTargetRequest) (*pbs
 	if authResults.Error != nil {
 		return nil, authResults.Error
 	}
-	t, ts, err := s.getFromRepo(ctx, req.GetId())
+	t, ts, _, err := s.getFromRepo(ctx, req.GetId())
```
louisruch, line 207: @talanknight I returned the credential library up to the service handler, but I didn't do anything from this point on.  I was not sure if I should update the [proto](https://github.com/hashicorp/boundary/blob/main/internal/proto/local/controller/api/resources/targets/v1/target.proto#L22) to include the cred libs or if you would be doing that as part of the API work.

louisruch:
ping @mgaffney @talanknight

internal/servers/controller/handlers/targets/target_service.go
talanknight: I can update that as part of the API work.

internal/target/repository_credential_library.go
```diff
@@ -0,0 +1,349 @@
...
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/kms"
+	"github.com/hashicorp/boundary/internal/oplog"
+)
+
+// AddTargetCredentialLibraries adds the clIds to the targetId in the repository. The target
+// and the list pf credential libraries attached to the target, after clIds are added,
+// will be returned on success.
+// The targetVersion must match the current version of the targetId in the repository.
+func (r *Repository) AddTargetCredentialLibraries(ctx context.Context, targetId string, targetVersion uint32, clIds []string, _ ...Option) (Target, []*CredentialLibrary, error) {
```
talanknight, line 18: The successful result of any operation by the API is to return the resource itself (including the host sets).  Can you add the []TargetSets as a return value (and add []*CredentialLibrary for the SAD HostSet operations?)

internal/target/repository.go
```diff
@@ -53,14 +53,15 @@ func NewRepository(r db.Reader, w db.Writer, kms *kms.Kms, opt ...Option) (*Repo
 }
 
 // LookupTarget will look up a target in the repository and return the target
-// with its host set ids.  If the target is not found, it will return nil, nil, nil.
+// with its host set ids and credential library ids.  If the target is not found,
+// it will return nil, nil, nil, nil.
 // No options are currently supported.
-func (r *Repository) LookupTarget(ctx context.Context, publicIdOrName string, opt ...Option) (Target, []*TargetSet, error) {
+func (r *Repository) LookupTarget(ctx context.Context, publicIdOrName string, opt ...Option) (Target, []*TargetSet, []*CredentialLibrary, error) {
```
talanknight, line 59: The naming convention between TargetSet and CredentialLibrary are inconsistent IMO.  To make them consistent CredentialLibrary should be renamed TargetLibrary, TargetSet should be renamed HostSet, or rename both to Library and Set.

internal/target/repository_host_set.go
```diff
@@ -0,0 +1,349 @@
...
+	"github.com/hashicorp/boundary/internal/kms"
+	"github.com/hashicorp/boundary/internal/oplog"
+)
+
+// AddTargetHostSets provides the ability to add host sets (hostSetIds) to a
+// target (targetId).  The target's current db version must match the
+// targetVersion or an error will be returned.   The target and a list of
+// current host set ids will be returned on success. Zero is not a valid value
+// for the WithVersion option and will return an error.
+func (r *Repository) AddTargetHostSets(ctx context.Context, targetId string, targetVersion uint32, hostSetIds []string, _ ...Option) (Target, []*TargetSet, error) {
```
talanknight, line 18: As discussed for the TargetLibrary SAD operations, please also return []*CredentialLibrary here so the API can return the full Target resource.

talanknight (changes requested):
Looks Good!

internal/target/repository.go
louisruch: This was originally named `TargetCredentialLibrary` but @mgaffney changed it to `CredentialLibrary` to remove stutter.  I will probably update `TargetSet` to `HostSet`

internal/servers/controller/handlers/targets/target_service.go
louisruch: Sounds good

internal/target/repository_credential_library.go
louisruch: Great point!

internal/target/repository.go
louisruch: Lets chat with Mike tomorrow about best naming here, TargetSet is also the name of the view

mgaffney (commented):
This looks really good. I left just one question.

internal/target/repository_credential_library.go
```diff
@@ -204,52 +210,46 @@ func (r *Repository) DeleteTargetCredentialLibraries(ctx context.Context, target
...
-		cl, err := NewCredentialLibrary(targetId, id)
+	if len(changes) == 0 {
+		// Nothing needs to be changed, return early
+		credLibs, err := fetchLibraries(ctx, r.reader, targetId)
 		if err != nil {
-			return nil, db.NoRowsAffected, errors.Wrap(err, op, errors.WithMsg("unable to create in memory target credential library"))
+			return nil, nil, db.NoRowsAffected, errors.Wrap(err, op)
 		}
-		addCredLibs = append(addCredLibs, cl)
+		return nil, credLibs, db.NoRowsAffected, nil
```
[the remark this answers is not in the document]
louisruch: oh weird I updated the return signature but never actually returned hostSets, ty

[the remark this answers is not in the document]
louisruch: Oh I see this is the return early path my bad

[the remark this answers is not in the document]
louisruch: good catch!

mgaffney (approved):
Nice work!

internal/target/repository_credential_library_test.go
```diff
@@ -0,0 +1,482 @@
...
+			args: args{
+				targetVersion: 1000,
+				clIds:         []string{lib1.PublicId, lib2.PublicId},
+				addToOrigLibs: true,
+			},
+			wantErr:     true,
+			wantErrCode: errors.VersionMismatch,
+		},
+		{
+			name:  "remove existing and add users and grps",
```
talanknight, line 399: I think a few of these test names are left w/ copy/paste errors (referencing sets and users and groups).

internal/target/repository_host_set_test.go
```diff
@@ -0,0 +1,493 @@
...
+	"github.com/hashicorp/boundary/internal/host/static"
+	"github.com/hashicorp/boundary/internal/iam"
+	"github.com/hashicorp/boundary/internal/kms"
+	"github.com/hashicorp/boundary/internal/oplog"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/proto"
+)
+
+func TestRepository_AddTargetHostSets(t *testing.T) {
```
talanknight, line 20: I realize this was just a refactoring (move to it's own file) but consider adding the corresponding tests I called out in the repository_credential_library_test.go file regarding adding host sets that are already attached and removing sets which aren't associated yet at all.  This will help ensure all SAD operation behaviors are consistent with each other.

talanknight (changes requested):

internal/target/repository_credential_library_test.go
louisruch: :/

talanknight (approved):

PR #1213, opened 2021-05-07
Remove text in oidc command on json format

### LINKED ISSUE #1193: Authenticate with openid connect formatted as json returns disturbing text
**Describe the bug**

Authenticate with openid connect authentication method. And return response as json.
It will return following text, before json response:
```
Opening returned authentication URL in your browser...
```

That is a problem if you will further process the json response i.e. with `jq`

**To Reproduce**
1. `boundary authenticate oidc -auth-method-id amoidc_123456 -format=json | jq -r '.item.token'`

denniskniep:
Fixes #1193

Fixes that the openid connect authenticate command formatted as json returns
disturbing text. Therefore it can not be piped and parsed by another tool

### DIFF
--- a/internal/cmd/commands/authenticate/oidc.go
+++ b/internal/cmd/commands/authenticate/oidc.go
@@ -101,7 +101,9 @@ func (c *OidcCommand) Run(args []string) int {
 		return base.CommandCliError
 	}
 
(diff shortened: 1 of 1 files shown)


jefferai:
Thanks!

PR #1214, opened 2021-05-10
Migrate the server upsert query to use database time

jefferai:
This ensures that we are using consistent ideas of time regardless of
the local time on each worker (for this query).


louisruch (dismissed):
LGTM, I saw this and wanted to update it as we are using DB time for all other queries but clearly forgot!

jefferai:
Note: test failure is unrelated (https://hashicorp.atlassian.net/browse/ICU-1490)

PR #1215, opened 2021-05-10
Add Credential Store CRUDL API and handlers

talanknight:
This PR defines and implements the interfaces for CRUDL operations on the Credential Store.

* This PR does not add payload validation for the CA Cert or client certs. This should be resolved with ICU-1479.
* This PR does not allow for updating the client certificate/key since in the repo the cert and cert key pair is treated as a singular value object while in the API they are treated as separate fields with distinct field masks. ICU-1488 was created to resolve this issue.
* This PR does not return the vault token hmac except for in Create and Get requests. This should be resolved in ICU-1480 and ICU-1483.
* This PR forces the client credential and the client credential private key to be sent in different fields. (task ICU-1497 will fix this)


internal/servers/controller/common/common.go
```diff
@@ -11,12 +12,13 @@ import (
...
-	AuthTokenRepoFactory    = oidc.AuthTokenRepoFactory
-	IamRepoFactory          func() (*iam.Repository, error)
-	OidcAuthRepoFactory     = oidc.OidcRepoFactory
-	PasswordAuthRepoFactory func() (*password.Repository, error)
-	ServersRepoFactory      func() (*servers.Repository, error)
-	StaticRepoFactory       func() (*static.Repository, error)
-	SessionRepoFactory      func() (*session.Repository, error)
-	TargetRepoFactory       func() (*target.Repository, error)
+	AuthTokenRepoFactory       = oidc.AuthTokenRepoFactory
+	VaultCredentialRepoFactory func() (*vault.Repository, error)
```
louisruch, line 16: Can you add this definition as it is done above for `oidc.AuthTokenRepoFactory` I will need the repo factory within the vault package for the renewal job work I am doing

internal/types/resource/resource.go
```diff
@@ -4,22 +4,24 @@ package resource
...
+	AuthToken         Type = 8
+	HostCatalog       Type = 9
+	HostSet           Type = 10
+	Host              Type = 11
+	Target            Type = 12
+	Controller        Type = 13
+	Worker            Type = 14
+	Session           Type = 15
+	CredentialStore   Type = 16
+	CredentialLibrary Type = 16
```
louisruch, line 24:
```suggestion
	CredentialLibrary Type = 17
```

internal/credential/vault/credential_store.go
```diff
@@ -25,17 +24,6 @@ type CredentialStore struct {
 // client cert, namespace, TLS server name, and TLS skip verify are the
 // only valid options. All other options are ignored.
 func NewCredentialStore(scopeId string, vaultAddress string, token []byte, opt ...Option) (*CredentialStore, error) {
-	const op = "vault.NewCredentialStore"
```
louisruch, line 28: Why are these checks dropped?

talanknight: The checks are dropped because they are only required when persisting in storage (In the Create* call) but not when creating a credential store in memory (for Update for example).  The Create call already checks these values for being set.

internal/types/resource/resource.go
talanknight: Hmmm, I thought it was because they were somehow stored in the DB but after looking closer I don't think that is true.  I thought I had asked about it before but maybe I'm thinking of a different enum.  I've updated these to use iota. @jefferai let me know if there is a reason to not use iota here.

internal/credential/vault/repository.go
```diff
@@ -6,6 +6,9 @@ import (
 	"github.com/hashicorp/boundary/internal/kms"
 )
 
+// OidcRepoFactory is used by "service functions" to create a new oidc repo
+type CredentialRepoFactory func() (*Repository, error)
```
talanknight, line 10: Copy/paste error for the comment.
"Credential" prefix may be redundant, consider removing.

PR #1216, opened 2021-05-10
Quickstart with Docker-Compose

denniskniep:
Starts a preprovisioned Boundary environment with an IdentityProvider
for OIDC flow and a ssh server for testing the connect feature.

Following setup is started:
* Using Postgres as database for boundary
* Preprovision Boundary with Terraform and CLI
* Using Keycloak as IdentityProvider with ready to go config and user
* SSH Server waiting for connection through boundary

I think this setup is useful for developers as well as for users who wants to quickly test scenarios. Especially scenarios involving  authentication via OIDC and then connecting to ssh server. This basic use case needs no further manual configuration. Setup is automatically done via docker-compose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,10 @@ number of connections per session or setting a maximum time limit; try canceling
 an active session from the sessions page or via `boundary sessions`, make your
 own commands with `boundary connect -exec`, and so on.
 
+### Quickstart with Docker-Compose
+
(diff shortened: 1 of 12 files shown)


denniskniep:
Hi,
is everything fine with this PR?

jimlambrt:
Thanks for submitting this. We are working on an OIDC example for our reference architecture (https://github.com/hashicorp/boundary-reference-architecture), and though this closely aligns with that work, we are going to make the first version of it as simple as possible. We are open to future growth on that example.

PR #1218, opened 2021-05-12
Define base types for the credential package

mgaffney:


internal/credential/credential.go
```diff
@@ -0,0 +1,60 @@
...
+type Credential interface {
+	Entity
+	Secret() SecretData
+}
+
+// Dynamic is a credential generated by a library for a specific session.
+type Dynamic interface {
+	Credential
+	SessionId() string
+	Library() Library
```
louisruch, line 59: Does `Library() Library` not belong to the `Credential`

louisruch (approved):
LGTM
This was actually great to read, nice summary of how it all links together

talanknight (approved):
That's great!  I think everything up through Resource could be put into a common package shareable across the internal packages since the definition applies for all resources across boundary's domain.

This'll replace the interfaces defined the subtype.go nicely

Issue #1055, opened 2021-03-31
Attempting to delete a stale active session throws a duplicate key violation


dylanrhysscott:
**Describe the bug**

When deleting old stale but still active sessions Boundary reports the following error from the controller

```
2021-03-31T22:21:52.261Z [ERROR] controller: error performing termination of completed sessions: error="session.(Repository).TerminateCompletedSessions: db.DoTx: session.(Repository).TerminateCompletedSessions: db.Exec: duplicate key value violates unique constraint "session_state_session_id_end_time_key": unique constraint violation: 
```

**To Reproduce**

1. Open several sessions to the same target 
2. I think let the session timeout due to OS sleep or similar - I left a session or two open and my Mac went to sleep
3. Session apparently remains active even though the terminal process is no long running but unable to be deleted from 

**Expected behavior**

Sessions should be cancellable without a dependency violation

**Additional context**

I attempt to remove the session via the Boundary desktop application and CLI with same result

<img width="1298" alt="Screenshot 2021-03-31 at 23 33 56" src="https://user-images.githubusercontent.com/4922401/113219650-b167cc80-9279-11eb-9715-bc64f7c62c06.png">

The session I tried to cancel has long since been closed from the client but has remained stuck here

From the CLI same error is presented

```
➜  ~ boundary sessions cancel -id s_tURbHPlAvP
Error from controller when performing cancel on session:
Error information:
  Kind:                InvalidArgument
  Message:             Invalid request.  Request attempted to make second resource with the same field value that must be unique.
  Status:              400
➜  ~
```

Boundary CLI version

```
Version information:
  Git Revision:        bc565922fbd3a18c9f6a22cd2e80a93df0d7cd45
  Version Number:      0.1.7
```

Boundary Server Version

```
/ $ boundary version

Version information:
  Git Revision:        c0f33f982c87c0eb4127cb16cf06b03a37b91dbd
  Version Number:      0.1.8

/ $
```

I am running Boundary server in Docker on Kubernetes backed by Postgres

rodrigocastrillon:
Hello. 

Same happens to me when I try to use a browser to connect to a target host. I also noticed that Firefox is not able to connect to the mapped port (connection timeout), while Chrome and Edge can (I see packets with tcpdump from worker to destination host), but connection is reset at some point. I am trying to connect to a Solar Winds https URL behind the worker.

After some testing I have several stale sessions that I am not able to cancel:
```
[...]

Issue #894, opened 2021-01-29
Sessions are not closing/terminated [ICU-1257]


Akisaji:
**Describe the bug**
Sessions are displayed as active in the webui and cli and stay active after cancelling the connection.

**To Reproduce**
1. Create a connection
2. Shutdown boundary with connection still active
3. Start boundary(usually the next day)
4. Connection is still displayed in the view way beyond it's TTL

**Expected behavior**
For the connection to be closed when boundary is shut downed and no longer displayed as active. And be able to cancel the connecting without it popping back up.

**Additional context**
The way we have boundary setup currently, it's shutdown at the end of day so my guess is this is what is causing the issues.

So I'm having this issue currently with multiple sessions and have seen the issue come back several times. Only way i've found to "fix" this is to reinitialize the db which i would like to avoid.

```
boundary sessions read -id s_0P4gIy6IG6

Session information:
  Auth Token ID:        at_7c30sq3trl
  Created Time:         Thu, 21 Jan 2021 16:03:07 CET
  Endpoint:             tcp://url
  Expiration Time:      Fri, 22 Jan 2021 00:03:07 CET
  Host ID:              hst_GojaoqThFC
  Host Set ID:          hsst_3snDcq1TsV
  ID:                   s_0P4gIy6IG6
  Status:               terminated
  Target ID:            ttcp_p2xUpl4ZHU
  Termination Reason:   canceled
  Type:                 tcp
  Updated Time:         Fri, 29 Jan 2021 12:14:40 CET
  User ID:              u_TWZ6052pWZ
  Version:              31

  Scope:
    ID:                 p_j3z93Fl705
    Name:               test
    Parent Scope ID:    o_c4HqJ42HzW
    Type:               project

  States:
    Start Time:         Thu, 21 Jan 2021 16:04:17 CET
    Status:             terminated

    End Time:           Thu, 21 Jan 2021 16:04:17 CET
    Start Time:         Thu, 21 Jan 2021 16:03:11 CET
    Status:             canceling

    End Time:           Thu, 21 Jan 2021 16:03:11 CET
    Start Time:         Thu, 21 Jan 2021 16:03:08 CET
    Status:             active

    End Time:           Thu, 21 Jan 2021 16:03:08 CET
    Start Time:         Thu, 21 Jan 2021 16:03:07 CET
    Status:             pending

```
And for a list in the scope i get the following output

```
[...]

PR #1220, opened 2021-05-13
Add cleanup of dead connections no longer reported by a worker

jefferai:
When a worker no longer reports a connection (e.g. after a restart), the controller will now mark those connections as closed so that sessions can be terminated instead of sitting in canceling state due to stuck connections.

Related to #894 
Related to #1055


internal/session/repository_connection.go
```diff
@@ -48,7 +49,13 @@ func (r *Repository) ListConnections(ctx context.Context, sessionId string, opt
 	const op = "session.(Repository).ListConnections"
 	opts := getOpts(opt...)
 	var connections []*Connection
-	err := r.list(ctx, &connections, "session_id = ?", []interface{}{sessionId}, opts) // pass options, so WithLimit and WithOrder are supported
+	var args []interface{}
+	if sessionId != "" {
+		args = append(args, sessionId)
+	}
+	// The where clause will be ignored if args is empty, if we don't want to
```
mgaffney, line 56: This essentially makes the `sessionId` parameter optional and allows all of the rows in the `session_connection` table to be returned. Why do we want to do that?

internal/session/query.go
```diff
@@ -232,4 +234,49 @@ where
...
+  -- Find connections that are not closed so we can reference those IDs
+  unclosed_connections as (
+    select connection_id
+      from session_connection_state
+    where
+      -- It's the current state
+      end_time is null
+        and
+      -- Current state isn't closed state
+      (state = 'authorized' or state = 'connected')
```
mgaffney, line 255:
```suggestion
      state in ('authorized', 'connected')
```

internal/session/query.go
```diff
@@ -232,4 +234,49 @@ where
...
+        and
+      -- These are connection IDs that just got reported to us by the given
+      -- worker, so they should not be considered closed.
+        %s
+      -- Only unclosed ones
+      public_id in (select connection_id from unclosed_connections)
+  )
+  update session_connection
+    set
+      closed_reason = 'system error'
```
mgaffney, line 277: "system error" seems to generic to me. Should we create a new `closed_reason` for this? Maybe "worker terminated"?

mgaffney (changes requested):
This looks really good. Just a few comments and questions.

internal/db/option.go
```diff
@@ -35,7 +35,7 @@ type Options struct {
 	// WithVersion must be accessible from other packages.
 	WithVersion *uint32
 
-	withSkipVetForWrite bool
+	WithSkipVetForWrite bool
```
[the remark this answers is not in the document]
jefferai: Yeah, let me validate that change you suggested works and if so I can revert this.

internal/db/schema/migrations/postgres/8/01_connection.up.sql
```diff
@@ -0,0 +1,25 @@
+begin;
+
+alter table session_connection
+  add column server_id text;
+
+-- Note: here, and in the session table, we should add a trigger ensuring that
+-- if server_id goes to null, we mark connections as closed. See
+-- https://hashicorp.atlassian.net/browse/ICU-1495
+alter table session_connection
+  add constraint session_connection_server_id_fkey
```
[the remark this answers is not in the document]
jefferai: Ah, I think I copied this from somewhere else so likely a different place that hadn't been updated :-)

internal/session/repository_connection.go
jefferai: The behavior of the where clause being ignored if args are empty is _already_ how it behaves. The existing code doesn't check whether session ID is empty or not. What my change does is make it so that if you pass in an empty session ID, instead of getting nothing back (because it puts an empty string in as the session ID, you get all connections back. In my mind, this matches the comment at the top -- this function lists connections. Session ID therefore is a constraint.

In support of that point, it will be useful at some point in the near future to be able to filter by server ID instead of by session ID. That puts us in the realm of being able to specify one constraint or the other or both.

internal/session/query.go
jefferai: I don't know that we can say for sure that the worker was terminated. We just know that the worker is no longer advertising it.

I think we should keep this as is, and I'll add another subtask to have the worker, when given a shutdown signal, attempt to gracefully terminate its connections and tell the controller that the reason is `worker terminating` or so.

internal/session/repository_connection.go
mgaffney: The current comment at the top is "ListConnections will sessions.  Supports the WithLimit and WithOrder options." which doesn't make any sense let alone describe the expected behavior. In general, I think we use Option for optional parameters and treat the other parameters as required. Your change makes `sessionId` an optional parameter and changes the meaning of the method from "lookup connections for session id" to "get all connections".

jefferai: As discussed on Slack I actually think it _should_ be optional and the current behavior (and API signature) is incorrect. However for now I've changed it so that passing `*` explicitly lists all connections, rather than passing no value.

internal/db/timestamp/scanners.go
```diff
@@ -3,21 +3,16 @@ package timestamp
 import (
 	"database/sql/driver"
 	"errors"
-	"fmt"
 	"time"
 
-	"github.com/golang/protobuf/ptypes"
+	timestamppb "google.golang.org/protobuf/types/known/timestamppb"
```
louisruch, line 8: nit: redundant alias

louisruch (approved):
Two minor nits otherwise this looks great

internal/session/repository_connection.go
mgaffney: I think we should split this out to a new method instead of using a `*`.

mgaffney: Meaning, we should create a `GetAllConnections` method and if it is only needed for testing, add it should be in a `_test.go` file and unexported.

jefferai: Didn't see this comment at the time of merge; addressed by https://github.com/hashicorp/boundary/pull/1224

cten:
After upgrading to 0.11.2 I am seeing all the `pending` sessions in the UI but get this when trying to Cancel them 
Screenshot 2023-01-24 at 10 25 08 AM

PR #1221, opened 2021-05-13
Add CRUDL operations for vault credential libraries

talanknight:
In addition to adding API endpoints for library CRUDL operations, this PR also:
* Moves setting default values for HttpMethod to CreateCredentialLibrary and UpdateCredentialLibrary.
* Allows the creation of in memory CredentialLibraries with no restrictions on HttpMethod and HttpRequestBody


internal/credential/vault/repository_credential_library.go
```diff
@@ -10,6 +10,15 @@ import (
...
 	"github.com/hashicorp/boundary/internal/oplog"
+	"github.com/hashicorp/boundary/sdk/strutil"
+)
+
+const (
+	nameField = "Name"
+	descriptionField = "Description"
+	vaultPathField = "VaultPath"
+	httpMethodField = "HttpMethod"
+	httpRequestBodyField = "HttpRequestBody"
```
mgaffney, line 21: This feels wrong to me. Creating string `const` values for fields means the fields are now declared in the protobuf, the struct the protobuf generator generates, and now as here. What value does this add?

mgaffney (changes requested):

internal/credential/vault/repository_credential_library.go
talanknight: This helps avoid typo errors since some of these are used up to 5 times in this file.

mgaffney: True. And avoiding typos is a good thing. It still feels wrong to me that we are hand rolling field names when we have structs with the names but I don't have a better solution right now.

mgaffney (approved):

louisruch (approved):

internal/credential/vault/repository_credential_library.go
```diff
@@ -103,30 +119,38 @@ func (r *Repository) UpdateCredentialLibrary(ctx context.Context, scopeId string
...
+			descriptionField:     l.Description,
+			vaultPathField:       l.VaultPath,
+			httpMethodField:      l.HttpMethod,
+			httpRequestBodyField: l.HttpRequestBody,
 		},
 		fieldMaskPaths,
 		nil,
 	)
+
+	if strutil.StrListContains(nullFields, httpMethodField) {
```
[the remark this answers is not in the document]
louisruch: I guess there is a comment on the method. So maybe no merge as is...

internal/servers/controller/handlers/credentiallibraries/credentiallibrary_service.go
```diff
@@ -0,0 +1,602 @@
...
+				badFields[globals.AttributesField] = "Attribute fields do not match the expected format."
+				break
+			}
+			if handlers.MaskContains(req.GetUpdateMask().GetPaths(), vaultPathField) && attrs.VaultPath == "" {
+				badFields[vaultPathField] = "This is a required field and cannot be set to empty."
+			}
+			if m := attrs.GetHttpMethod(); handlers.MaskContains(req.GetUpdateMask().GetPaths(), httpMethodField) && m != nil && !strutil.StrListContains([]string{"GET", "POST"}, strings.ToUpper(m.GetValue())) {
+				badFields[httpMethodField] = "If set, value must be 'GET' or 'POST'."
+			}
+			if b := attrs.GetHttpRequestBody(); b != nil && strings.ToUpper(attrs.GetHttpMethod().GetValue()) == "GET" {
```
jefferai, line 578: There are a couple of options here...

As-is, if you are taking in a raw request body, you should validate that it can be parsed successfully as JSON. This may not be possible anymore at some point if we allow templating, which...we probably will...but even in that case likely it'd be specific values that are templated and they could just be properly JSON-escaped.

The other option here is to take in a map[string]string, then do that formatting into JSON ourselves at request time. This has the advantage that it's more UX-friendly -- you're defining the parameters of the request body without worrying about the JSON formatting and escaping. So it's easier on the CLI and in the UI. Generally speaking Vault's API is designed to be able to convert from string to necessary other types depending on context (e.g. if an endpoint needs an int for a param, it will do that translation for you) but there are a few places here and there where a raw value might be expected/needed, in which case you'd need a raw request body parameter anyways.

Of course, a third option is do the nicer UX thing first, then add a raw request body parameter later _if needed_ and make those mutually exclusive.

Probably worth roping in opinions from FE and PM.

PR #1223, opened 2021-05-14
Add missing API-level state checks during session cancel

jefferai:
This validates it's not already canceled or in terminated state.


louisruch (approved):
LGTM

mgaffney (approved):

PR #1225, opened 2021-05-14
Create Target API for SAD operations on credential libraries.

talanknight:
Please note that with this change I added a view for the target/credential library relationship and added a new struct in the credential package named TargetLibrary.  This is done so we can attach the store id with the library info on the target.


internal/db/schema/migrations/postgres/10/03_target.up.sql
```diff
@@ -50,4 +50,15 @@ begin;
   create trigger immutable_columns before update on target_credential_library
     for each row execute procedure immutable_columns('target_id', 'credential_library_id', 'credential_purpose', 'create_time');
 
+create view target_library
+as
+select
+    tcl.*,
```
jefferai, line 56: I believe that the current practice for views is to explicitly state the fields, which among other things ensures that if columns are added to `tcl` that the view isn't suddenly scanning wrong to the application layer structs. Mike/Jim would know best here though.

internal/proto/local/controller/api/resources/targets/v1/target.proto
```diff
@@ -18,6 +18,16 @@ message HostSet{
...
 }
 
+message CredentialLibrary{
+	// Output only. The ID of the Credential Library.
+	string id = 10;
+
+	// Output only. The Credential Store to which this Credential Library belongs.
+	string credential_store_id = 20 [json_name="credential_store_id"];
+
+	string purpose = 30;
```
jefferai, line 28: What's `purpose`?

jefferai:
Looking pretty good, just those couple of comments.

jefferai:
I merged `mgaffney-vault` into here to see if it fixes the build issues.

jefferai (approved):

internal/servers/controller/handlers/targets/target_service.go
```diff
@@ -909,6 +1024,75 @@ func (s Service) removeInRepo(ctx context.Context, targetId string, hostSetIds [
...
+	for _, l := range libraries {
+		lids = append(lids, l.GetId())
+	}
+	_, _, _, err = repo.SetTargetCredentialLibraries(ctx, targetId, version, strutil.RemoveDuplicates(lids, false))
+	if err != nil {
+		// TODO: Figure out a way to surface more helpful error info beyond the Internal error.
+		return nil, nil, nil, handlers.ApiErrorWithCodeAndMessage(codes.Internal, "Unable to set credential libraries in target: %v.", err)
+	}
+
+	out, hs, cl, err := repo.LookupTarget(ctx, targetId)
```
louisruch, line 1065: Why do a lookup here?  Doesn't the `SetTargetCredentialLibraries` return the same info?

internal/servers/controller/handlers/targets/target_service.go
```diff
@@ -1194,20 +1392,106 @@ func validateSetRequest(req *pbs.SetTargetHostSetsRequest) error {
...
+			break
+		}
+	}
+	for _, cl := range req.GetCredentialLibraries() {
+		if !handlers.ValidId(handlers.Id(cl.GetId()), vault.CredentialLibraryPrefix) {
+			badFields[globals.CredentialLibrariesField] = fmt.Sprintf("Incorrectly formatted credential library identifier %q.", cl.GetId())
+			break
+		}
+		if cl.GetPurpose() != "" && cl.GetPurpose() != "application" {
+			badFields[globals.CredentialLibrariesField] = fmt.Sprintf("Unrecognized purpose  %q for credential library.", cl.GetPurpose())
```
louisruch, line 1441:
```suggestion
			badFields[globals.CredentialLibrariesField] = fmt.Sprintf("Unrecognized purpose %q for credential library.", cl.GetPurpose())
```

internal/servers/controller/handlers/targets/target_service.go
```diff
@@ -1194,20 +1392,106 @@ func validateSetRequest(req *pbs.SetTargetHostSetsRequest) error {
...
```
louisruch, line 1471:
```suggestion
			badFields[globals.CredentialLibrariesField] = fmt.Sprintf("Unrecognized purpose %q for credential library.", cl.GetPurpose())
```

internal/types/action/action.go
```diff
@@ -8,42 +8,45 @@ type Type uint
...
-	Cancel           Type = 27
-	AddAccounts      Type = 28
-	SetAccounts      Type = 29
-	RemoveAccounts   Type = 30
-	ReadSelf         Type = 31
-	CancelSelf       Type = 32
-	ChangeState      Type = 33
-	DeleteSelf       Type = 34
-	NoOp             Type = 35
+	Unknown                   Type = 0
```
louisruch, line 11: Same comment re using `iota`

internal/servers/controller/handlers/targets/target_service_test.go
```diff
@@ -1264,3 +1266,479 @@ func TestRemoveTargetHostSets(t *testing.T) {
...
+				CredentialLibraryIds: tc.addLibraryIds,
+				CredentialLibraries:  tc.addLibraries,
+			}
+
+			got, err := s.AddTargetCredentialLibraries(auth.DisabledAuthTestContext(iamRepoFn, proj.GetPublicId()), req)
+			require.NoError(t, err, "Got error: %v", s)
+
+			assert.ElementsMatch(t, tc.resultLibraryIds, got.GetItem().GetCredentialLibraryIds())
+
+			if len(tc.resultLibraries) != 0 {
```
louisruch, line 1355: All test cases have an empty `resultLibraries`, so this if is never entered

louisruch (commented):
Some questions and small comments but overall this looks great

internal/servers/controller/handlers/targets/target_service.go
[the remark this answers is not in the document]
talanknight: It does not return the same info as AddTargetCredentialLibraries.  Set and Delete is consistent across all resource's SAD repo operations. Target's Add seems to be a little different than group's.

internal/types/action/action.go
talanknight: I followed the comment above this const block which said they are not using iota intentially since the values are stored in the db.

internal/servers/controller/handlers/targets/target_service_test.go
talanknight: Thank you!

internal/types/action/action.go
louisruch: I guess its easier to see a change...

louisruch (approved):
LGTM

internal/target/credential_library.go
```diff
@@ -71,3 +71,13 @@ func (t *CredentialLibrary) oplog(op oplog.OpType) oplog.Metadata {
 	}
 	return metadata
 }
+
+type TargetLibrary struct {
```
mgaffney, line 75: Can you add doc?

mgaffney (commented):
Small request

internal/target/credential_library.go
talanknight: Added. Thanks!

mgaffney (approved):

PR #1226, opened 2021-05-14
Include the Token HMAC in the Update or List responses

talanknight:


internal/credential/vault/repository_credential_store.go
```diff
@@ -820,4 +834,4 @@ func (r *Repository) DeleteCredentialStore(ctx context.Context, publicId string,
 	}
 
 	return rowsDeleted, nil
-}
+}
```
louisruch, line 837: github wants a newline

internal/credential/vault/repository_credential_store.go
```diff
@@ -690,6 +690,15 @@ func (r *Repository) UpdateCredentialStore(ctx context.Context, cs *CredentialSt
 			}
 			msgs = append(msgs, &csOplogMsg)
 
+			ir, err := NewRepository(reader, w, r.kms)
```
louisruch, line 693: Since you are doing a lookup anyway is there a reason we don't do the lookup at the end of this method and return that with all necessarily fields blanked out

talanknight: Changed.  Thanks!

louisruch (approved):
LGTM

internal/credential/vault/repository_credential_store.go
```diff
@@ -737,6 +736,15 @@ func (r *Repository) UpdateCredentialStore(ctx context.Context, cs *CredentialSt
 				msgs = append(msgs, returnedClientCert.oplogMessage(db.CreateOp))
 			}
 
+			ir, err := NewRepository(reader, w, r.kms)
```
mgaffney, line 739: Why create a new repository when this call is in a transaction?

mgaffney (approved):
Looks good! I just had 1 question

internal/credential/vault/repository_credential_store.go
talanknight: Fixed this to instead use the same mechanism that Lookup uses inlined in the update call.

PR #1230, opened 2021-05-17
Add a certificate key hmac field for credential store.

talanknight:
This also adds the client certificate data into the credential store resource on the API.


jefferai (approved):
Looks good!

mgaffney (approved):
Looks good!

PR #1237, opened 2021-05-18
Upgrade build node version & packagespec

mdeggies:
This PR 1) upgrades boundary to the latest version of packagespec, v0.2.6, 2) installs the 12.x series of node which is needed to build the UI -- the default debian 10.x installation no longer works with the boundary-ui dependencies, and 3) includes the latest UI version from boundary-ui.


randallmorey (approved):

randallmorey (changes requested):
Can we use node 14?

randallmorey (approved):

PR #1241, opened 2021-05-19
Vault integration: run tests requiring docker only on supported platforms

mgaffney:


internal/credential/vault/supported.go
```diff
@@ -0,0 +1,11 @@
+// +build linux darwin windows
```
talanknight, line 1: This only runs the code that uses NewTestVaultServer when built on these 3 platforms, but what we want is for the dockertest to only be included when built using these 3 platforms since we are seeing build errors, not runtime errors, right?

louisruch: +1 
All unit tests are run on Linux AMD 64 it seems.  The errors we are seeing are from the binary build on the `bsd` platforms we are only seeing them now because the API work caused the cred package to be built and hence the dockertest

mgaffney: I didn't realize that was the problem. I'll work on that.

mgaffney: @talanknight can you take another look?

louisruch (approved):

talanknight (approved):

PR #1243, opened 2021-05-19
Already-canceled sessions return 200

jefferai:
If a session is already canceled or terminated, attempting to cancel it
will return the current session state and a 200, instead of an error.


louisruch (approved):
LGTM

PR #1244, opened 2021-05-19
Vault token renewal job

louisruch:
### What does this PR do

Adds vault token renewal job


talanknight (changes requested):

internal/credential/vault/jobs.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+
+		r.numProcessed++
+	}
+
+	return nil
+}
+
+// NextRunIn queries the vault credential repo to determine when the next token renewal job should run.
+func (r *TokenRenewalJob) NextRunIn() time.Duration {
```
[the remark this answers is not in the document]
louisruch: So the interface was created with the idea that the implementer would do something like `return time.Minute` so no error was originally envisioned here.  I will update though makes sense...

internal/credential/vault/jobs.go
```diff
@@ -0,0 +1,200 @@
...
+			// or malformed.  Set status to "expired" so credentials created with token can be
+			// cleaned up.
+			numRows, err := repo.writer.Exec(ctx, updateTokenStatusQuery, []interface{}{StatusExpired, s.Store.TokenHmac})
+			if err != nil {
+				return errors.Wrap(err, op)
+			}
+			if numRows != 1 {
+				return errors.New(errors.Unknown, op, "token expired but failed to update repo")
+			}
+			r.logger.Info("Vault credential store token has expired", "credential store id", s.Store.StoreId, "previous status", s.Store.TokenStatus)
```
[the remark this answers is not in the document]
louisruch: Perhaps, I might consider only logging if the token that expires is `current`.  I added this log because if a `current` token becomes invalid for whatever reason the end user will need to update the credential store with a new token.

internal/credential/vault/jobs.go
```diff
@@ -0,0 +1,200 @@
...
+		if err != nil {
+			return errors.Wrap(err, op, errors.WithMsg("unable to get database wrapper"))
+		}
+
+		err = s.Store.decrypt(ctx, databaseWrapper)
+		if err != nil {
+			return errors.Wrap(err, op)
+		}
+
+		vc, err := s.Store.client()
```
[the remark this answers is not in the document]
louisruch: Ill think about this a bit more, currently we do have a limit in number of tokens we query from the `db`

internal/credential/vault/jobs.go
```diff
@@ -0,0 +1,200 @@
...
+		return errors.Wrap(err, op)
+	}
+
+	repo, err := r.repoFn()
+	if err != nil {
+		return errors.Wrap(err, op)
+	}
+
+	var rps []*renewablePrivateStore
+	err = repo.reader.SearchWhere(ctx, &rps, `now() + (?||'min')::interval > renewal_time`, []interface{}{tokenRenewalWindow}, db.WithLimit(r.limit))
```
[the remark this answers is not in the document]
louisruch: Ill chat with @mgaffney tomorrow about alternatives here

internal/credential/vault/jobs.go
```diff
@@ -0,0 +1,200 @@
...
+	vault "github.com/hashicorp/vault/api"
+)
+
+const (
+	defaultTokenRenewalInterval = time.Minute * 5
+	tokenRenewalWindow          = 10 // in minutes
+)
+
+type TokenRenewalJob struct {
+	repoFn RepositoryFactory
```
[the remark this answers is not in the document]
louisruch: Ill chat with @mgaffney about this tomorrow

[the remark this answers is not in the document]
louisruch: Since the repo can be used concurrently (its running while another service looks at `NextRunTime`) its considered not safe and therefore we do need to use repoFn

internal/credential/vault/jobs.go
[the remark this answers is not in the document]
louisruch: @talanknight @mgaffney thoughts on added a rate limiter separate from the query limit?  Not sure what a good limit for vault here would be?

louisruch:
@mgaffney @talanknight This should be ready for re-review

internal/credential/vault/repository_credential_store.go
```diff
@@ -166,6 +166,12 @@ func (r *Repository) CreateCredentialStore(ctx context.Context, cs *CredentialSt
 			}
 			msgs = append(msgs, newToken.oplogMessage(db.CreateOp))
 
+			// Set next run time for token renewal task to at least the mid point of this token
+			tokenRenewalIn := token.expiration / 2
```
mgaffney, line 170: Suggest moving this to a method on token.

internal/credential/vault/repository_credential_store_test.go
```diff
@@ -27,11 +29,17 @@ func TestRepository_CreateCredentialStoreResource(t *testing.T) {
...
+		repo, err := NewRepository(rw, rw, kms, sche)
 		require.NoError(err)
 		require.NotNil(repo)
 		_, prj := iam.TestScopes(t, iam.TestRepo(t, conn, wrapper))
 
+		trj, err := NewTokenRenewalJob(rw, rw, kms, hclog.L())
+		require.NoError(err)
+		err = sche.RegisterJob(context.Background(), trj)
+		require.NoError(err)
+
```
mgaffney, line 42: Why add these to this test?

internal/db/timestamp/timestamp.go
```diff
@@ -22,5 +22,5 @@ func Now() *Timestamp {
 
 // AsTime converts x to a time.Time.
 func (x *Timestamp) AsTime() time.Time {
-	return x.AsTime()
+	return x.Timestamp.AsTime()
```
mgaffney, line 25: Why the change?

internal/scheduler/job.go
```diff
@@ -20,9 +20,9 @@ type Job interface {
 	// NextRunIn returns the duration until the next job run should be scheduled.  This
 	// method is invoked after a run has successfully completed and the next run time
 	// is being persisted by the scheduler.
-	NextRunIn() time.Duration
+	NextRunIn() (time.Duration, error)
```
mgaffney, line 23: Can you update the comment to explain the semantics of returning an error?

mgaffney (changes requested):
The looks really good. Nice work! 

Per our conversation, here are the various pieces of SQL we discussed:

### New index:
```sql
    create index credential_vault_token_expiration_time_ix
              on credential_vault_token(expiration_time);
comment on index credential_vault_token_expiration_time_ix is
      'the credential_vault_token_expiration_time_ix is used by the token renewal job';
```

### View with additional columns:
```sql
   create view credential_vault_job_renewable_tokens as
        select token_hmac,
               token, -- encrypted
               store_id,
               status,
               last_renewal_time,
               expiration_time,
               -- renewal time is the midpoint between the last renewal time and the expiration time
               last_renewal_time + (expiration_time - last_renewal_time) / 2 as renewal_time
          from credential_vault_token
         where status in ('current', 'maintaining');
```

### Query for getting the token renewal job's next run time:
```sql
select extract(epoch from renewal_time - now())::int as renewal_in
  from credential_vault_job_renewable_tokens
 where expiration_time = (
         select min(expiration_time)
           from credential_vault_job_renewable_tokens
       )
```

internal/credential/vault/jobs.go
```diff
@@ -0,0 +1,218 @@
...
+		return errors.Wrap(err, op)
+	}
+
+	var respErr *vault.ResponseError
+	renewedToken, err := vc.renewToken()
+	if ok := errors.As(err, &respErr); ok && respErr.StatusCode == http.StatusForbidden {
+		// Vault returned a 403 when attempting a renew self, the token is either expired
+		// or malformed.  Set status to "expired" so credentials created with token can be
+		// cleaned up.
+		numRows, err := r.writer.Exec(ctx, updateTokenStatusQuery, []interface{}{StatusExpired, s.Store.TokenHmac})
```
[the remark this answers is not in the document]
louisruch: Yeah I have noticed that in your most recent PRs and agree, will update

internal/credential/vault/jobs_test.go
```diff
@@ -0,0 +1,427 @@
...
+			assert.Equal(tt.args.kms, got.kms)
+			require.NotNil(got.logger)
+			assert.Equal(tt.wantLimit, got.limit)
+		})
+	}
+}
+
+func TestTokenRenewal_RunLimits(t *testing.T) {
+	t.Parallel()
+	assert, require := assert.New(t), require.New(t)
```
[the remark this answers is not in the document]
louisruch: Interesting i was not aware of this ty

internal/db/timestamp/timestamp.go
louisruch: `x.AsTime()` keeps calling this same function iteratively leading to stack overflow

internal/credential/vault/repository_credential_store_test.go
louisruch: `repo.CreateCredentialStore` attempts to update the next runtime of the TokenRenewalJob and errors out if the job does not exist.

internal/credential/vault/repository_credential_store.go
```diff
@@ -166,6 +166,12 @@ func (r *Repository) CreateCredentialStore(ctx context.Context, cs *CredentialSt
...
+			if err = r.scheduler.UpdateJobNextRunInAtLeast(ctx, TokenRenewalJobName, tokenRenewalIn); err != nil {
+				return errors.Wrap(err, op, errors.WithMsg("error updating next run time for vault token renewal job"))
+			}
+
```
mgaffney, line 174: Two questions: 
1. Should this be called after the transaction completes?
2. Why fail the call to `CreateCredentialStore` if the scheduler has an error here?

internal/credential/vault/repository_credential_store.go
```diff
@@ -714,6 +720,12 @@ func (r *Repository) UpdateCredentialStore(ctx context.Context, cs *CredentialSt
 				}
 				msgs = append(msgs, returnedToken.oplogMessage(db.CreateOp))
 
+				// Set next run time for token renewal task to at least the mid point of this token
+				tokenRenewalIn := token.expiration / 2
+				if err = r.scheduler.UpdateJobNextRunInAtLeast(ctx, TokenRenewalJobName, tokenRenewalIn); err != nil {
```
mgaffney, line 725: Same two questions.

internal/db/schema/migrations/postgres/10/03_vault_credential.up.sql
```diff
@@ -315,51 +320,55 @@ begin;
...
+            cert.key_id               as client_key_id
        from credential_vault_store store
-  left join current_tokens token
+  left join active_tokens token
          on store.public_id = token.store_id
   left join credential_vault_client_certificate cert
          on store.public_id = cert.store_id;
   comment on view credential_vault_store_client_private is
     'credential_vault_store_client_private is a view where each row contains a credential store and the credential store''s data needed to connect to Vault. '
+    'The view returns a separate row for each current and maintaining token, maintaining tokens should only be used for token/credential renewal and revocation.'
```
mgaffney, line 371:
```suggestion
    'The view returns a separate row for each current and maintaining token, maintaining tokens should only be used for token/credential renewal and revocation. '
```
Postgres concatenates all of the lines in the comment so each line (except for the last one) needs to end with a space.

internal/credential/vault/query.go
```diff
@@ -88,4 +88,34 @@ select *
...
+		  last_renewal_time + (expiration_time - last_renewal_time) / 2 as renewal_time
+		from credential_vault_token
+		where status in ('current', 'maintaining')
+	)
+	select extract(epoch from renewal_time - now())::int as renewal_in
+  	  from renewable_tokens
+     where expiration_time = (
+       select min(expiration_time)
+         from renewable_tokens
+     );
```
mgaffney, line 119: Can this `where` clause be moved to the `with renewable_tokens` part of the CTE?

mgaffney (approved):
This looks great. I just have a question about the `tokenRenewalNextRunInQuery` query.

talanknight (approved):

PR #1247, opened 2021-05-20
internal/cmd/config: allow explicit origin list w/implicit CORS toggle

vancluever:
This ensures that cors_allowed_origins can be set when cors_enabled is
not explicitly specified and allowed to be its default (currently true).

The current behavior is to clobber this value when cors_enabled is not
explicitly set, ultimately setting it to the default of
"serve://boundary" only.

### DIFF
--- a/internal/cmd/config/config_test.go
+++ b/internal/cmd/config/config_test.go
@@ -116,6 +116,34 @@ func TestDevController(t *testing.T) {
 		assert.Equal(t, []string{"*"}, l0.CorsAllowedOrigins)
 		assert.Nil(t, l0.CorsDisableDefaultAllowedOriginValues)
 
+		// Implicitly with explicit wildcard
+		conf = `
+		listener "tcp" {
+			purpose = "api"
+			cors_allowed_origins = ["*"]
+		}
+		`
+		actual, err = Parse(conf)
+		assert.NoError(t, err)
+		l0 = actual.Listeners[0]
+		assert.True(t, *l0.CorsEnabled)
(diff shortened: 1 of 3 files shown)


internal/cmd/config/config_test.go
```diff
@@ -116,6 +116,20 @@ func TestDevController(t *testing.T) {
 		assert.Equal(t, []string{"*"}, l0.CorsAllowedOrigins)
 		assert.Nil(t, l0.CorsDisableDefaultAllowedOriginValues)
 
+		// Implicitly with explicit wildcard
```
louisruch, line 119: It might be worthwhile adding a test to verify the implicit `cors_enabled` with explicit non wildcard `cors_allowed_origins` scenario, then ensure `desktopCorsOrigin` is appended to the `cors_allowed_origins` list as expected.

The only test of the appending functionality is when `cors_allowed_origins` is not specified

louisruch (dismissed):
One suggestion on adding an additional test, but you can have my unsolicited approval

internal/cmd/config/config_test.go
vancluever: Done!

vancluever:
Thanks for the review @louisruch!

vancluever:
(PS: Just also noticed this needs to be re-reviewed 🙂 )

louisruch (approved):
LGTM
Good call on adding the change log message there!

vancluever:
Thanks again 🙂

PR #1250, opened 2021-05-21
Create go SDK resources/methods for credential store and library

talanknight:
Separating this from the CLI work to unblock the TF provider work.


internal/api/genapi/templates.go
```diff
@@ -24,7 +24,11 @@ func getArgsAndPaths(in []string, parentTypeName, action string) (colArg, resArg
 		strToReplace = in[len(in)-2]
 	}
 	colArg = fmt.Sprintf("%sId", strcase.ToLowerCamel(strings.ReplaceAll(strToReplace, "-", "_")))
-	colPath = fmt.Sprintf("%ss", in[len(in)-1])
+	colPath = in[len(in)-1]
+	// append s at the end only if it isn't already present
+	if strings.LastIndex(colPath, "s") != len(colPath)-1 {
```
louisruch, line 29: Seems strange to use a search we can just compare the last char in the string.  
```
if colPath[len(colPath)-1] != 's' {...}
```
If we don't have a guarantee the `in[len(in)-1]` is not empty we can add `len(colPath) > 0` in the front

internal/credential/vault/repository_credential_store.go
```diff
@@ -680,15 +680,23 @@ func (r *Repository) UpdateCredentialStore(ctx context.Context, cs *CredentialSt
...
 			default:
 				rowsUpdated, err = w.Update(ctx, returnedCredentialStore, filteredDbMask, filteredNullFields, db.NewOplogMsg(&csOplogMsg), db.WithVersion(&version))
 				if err != nil {
 					return errors.Wrap(err, op, errors.WithMsg("unable to update credential store"))
 				}
-				if rowsUpdated != 1 {
+				switch rowsUpdated {
+				case 1:
+				case 0:
+					return nil
```
louisruch, line 698: This will be `0` if the store was deleted in a racing request or a version mismatch, should this really return nil here?

internal/tests/api/credentiallibraries/credentiallibrary_test.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestList(t *testing.T) {
+	assert, require := assert.New(t), require.New(t)
+	tc := controller.NewTestController(t, nil)
+	defer tc.Shutdown()
+
+	vaultServ := vault.NewTestVaultServer(t, vault.WithTestVaultTLS(vault.TestNoTLS))
```
louisruch, line 23: `TestNoTLS` is default so `vault.WithTestVaultTLS(vault.TestNoTLS)` does not need to be provided, but feel free to keep this if you feel it reads better

api/credentiallibraries/option.gen.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func DefaultVaultCredentialLibraryHttpMethod() Option {
+	return func(o *options) {
+		raw, ok := o.postMap["attributes"]
+		if !ok {
+			raw = interface{}(map[string]interface{}{})
+		}
+		val := raw.(map[string]interface{})
+		val["http_method"] = nil
```
louisruch, line 120: Why not `delete(val, "http_method")`?  
If you do delete you can also simplify this a bit by returning early in the `if !ok` instead of making `raw` just to set nil

louisruch (commented):
I have some question but otherwise this looks great

internal/credential/vault/repository_credential_store.go
talanknight: Yes.  This is how all other resources work with versioned changes.  If the version mismatches or the resource cannot be found it returns nil, no rows updated,  and no error.

internal/tests/api/credentiallibraries/credentiallibrary_test.go
talanknight: Removed. Thanks.

internal/api/genapi/templates.go
talanknight: Thanks.

api/credentiallibraries/option.gen.go
talanknight: The reason for this is we want the value to be included in the request and set to Null.  Doing this populates the field mask and tells the API that we want to clear whatever value was previously set for this value.  This is different from just not including the key in that if the key is not included the field mask is not included and the value is not updated.

internal/credential/vault/repository_credential_store.go
louisruch: 👍

api/credentiallibraries/option.gen.go
louisruch: aha, kk thks

louisruch (approved):
LGTM

PR #1253, opened 2021-05-25
Add initial SQL for managed groups

jefferai:
Note: all of the changes in `dev.go`, `testing.go`, `servers.go` are me moving the dev oidc setup outside of the `boundary dev` command and into a shared function that can be called by the test controller, as we require a valid OIDC auth method for this SQL. There are no real changes other than updates due to the new location and those changes can be ignored.


jimlambrt (changes requested):
one small change.  looks great and like the tests!

internal/db/schema/migrations/postgres/9/02_oidc_managed_group.up.sql
```diff
@@ -0,0 +1,98 @@
+begin;
+
+create table auth_oidc_managed_group (
+  public_id wt_public_id
+    primary key,
+  auth_method_id wt_public_id
+    not null,
+  scope_id wt_scope_id
+    not null,
+  "name" wt_name,
```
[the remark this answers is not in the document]
jefferai: I saw that pattern before. I assumed it was because syntax highlighting was highlighting it and making things confusing. I can take it out though.

jimlambrt (approved):
Looks great!

internal/db/schema/migrations/postgres/9/01_managed_groups.up.sql
```diff
@@ -0,0 +1,27 @@
...
+    primary key,
+  auth_method_id wt_public_id
+    not null,
+  -- Ensure that if the auth method is deleted (which will also happen if the
+  -- scope is deleted) this is deleted too
+  constraint auth_method_fkey
+    foreign key (auth_method_id) -- fk1
+      references auth_method(public_id)
+      on delete cascade
+      on update cascade
```
mgaffney, line 15: There should be a unique constraint for auth_method_id + public_id which should be the foreign key used by the subtypes. ([credential_store example](https://github.com/hashicorp/boundary/blob/mgaffney-vault/internal/db/schema/migrations/postgres/10/02_credential.up.sql#L12-L16))

internal/db/schema/migrations/postgres/9/02_oidc_managed_group.up.sql
```diff
@@ -0,0 +1,97 @@
...
+    foreign key (auth_method_id) -- fk1
+      references auth_oidc_method (public_id)
+      on delete cascade
+      on update cascade,
+  -- Ensure it relates to an abstract managed group
+  constraint auth_managed_group_fkey
+    foreign key (public_id) -- fk2
+      references auth_managed_group (public_id)
+      on delete cascade
+      on update cascade,
```
mgaffney, line 27: This should be a foreign key to `auth_managed_group (auth_method_id, public_id)` ([credential_vault_store example](https://github.com/hashicorp/boundary/blob/mgaffney-vault/internal/db/schema/migrations/postgres/10/03_vault_credential.up.sql#L28-L32))

internal/db/schema/migrations/postgres/9/02_oidc_managed_group.up.sql
```diff
@@ -0,0 +1,97 @@
...
+  constraint auth_oidc_managed_group_auth_method_id_name_uq
+    unique(auth_method_id, name),
+  -- QUESTION: is this necessary? It isn't for uniqueness, but maybe for an index to look up via auth method ID?
+  constraint auth_oidc_managed_group_auth_method_id_public_id_uq
+    unique(auth_method_id, public_id)
```
mgaffney, line 32: This is not necessary here. It should be on the base table instead.

mgaffney (changes requested):

internal/db/schema/migrations/postgres/9/02_oidc_managed_group.up.sql
```diff
@@ -0,0 +1,97 @@
...
+
+  insert into auth_managed_group
+    (public_id, auth_method_id)
+  values
+    (new.public_id, new.auth_method_id);
+
+  return new;
+
+end;
+$$ language plpgsql;
```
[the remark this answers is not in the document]
jefferai: Ah, I figured since the oidc subtype was responsible for the insertion that it would go in that file. I'm happy either way -- I can move it.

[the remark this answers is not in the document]
jefferai: Oh nm, I see oidc doesn't really factor in at all. Moved.

internal/db/schema/migrations/postgres/9/01_managed_groups.up.sql
jefferai: I'm assuming the reason for this is so that the subtype can't link only to the public ID but with a different auth method ID?

Added.

internal/db/schema/migrations/postgres/9/02_oidc_managed_group.up.sql
jefferai: Removed.

jefferai: Or rather, moved.

internal/db/schema/migrations/postgres/9/02_oidc_managed_group.up.sql
jefferai: Updated.

internal/db/schema/migrations/postgres/9/02_oidc_managed_group.up.sql
```diff
@@ -0,0 +1,97 @@
...
+
+-- Add into the base table when inserting into the concrete table
+create trigger
+  insert_managed_group_subtype
```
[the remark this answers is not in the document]
jefferai: Added.

mgaffney (approved):
LGTM!

PR #1257, opened 2021-05-27
internal/servers/worker: Controller failure connection cleanup

vancluever:


internal/servers/worker/status.go
```diff
@@ -115,6 +116,35 @@ func (w *Worker) sendWorkerStatus(cancelCtx context.Context) {
...
+			w.logger.Warn("status error grace period has expired, canceling all sessions on worker",
+				"last_status_time", lastStatus.StatusTime.String(),
+				"grace_period", statusFailGrace,
+			)
+
+			// Iterate through all sessions. Transition any sessions with state
+			// <SESSIONSTATUS_SESSIONSTATUS_CANCELING to that state.
+			w.sessionInfoMap.Range(func(sessionId, siRaw interface{}) bool {
+				si := siRaw.(*sessionInfo)
+				if si.status < pbs.SESSIONSTATUS_SESSIONSTATUS_CANCELING {
```
jefferai, line 138: Please use a switch on the status rather than depending on particular enum values never to change via `<`.

internal/servers/worker/status.go
```diff
@@ -115,6 +116,35 @@ func (w *Worker) sendWorkerStatus(cancelCtx context.Context) {
...
+					w.logger.Warn("canceling session due to worker communication failure", "session_id", sessionId)
+					si.Lock()
+					si.status = pbs.SESSIONSTATUS_SESSIONSTATUS_CANCELING
```
jefferai, line 141: I realize that setting it to canceling here allows the logic below to end up tearing down the session, but the issue here is that the session isn't necessarily canceled -- so if the status starts being sent/received properly again, we'd have a mismatch on session state from the controller.

Instead, you should terminate and mark closed the various connections within the sessions. Users won't be able to create new connections because if the communication with the controller is down it can't be authorized, and if the communication is up and the session was in fact canceled then it also won't be authorized.

vancluever: I was assuming that the controller would advance the status appropriately, but it seems not to do so and there's other connection-specific logic that may need to be followed. Will update accordingly.

internal/servers/worker/status.go
vancluever: :+1: I had a feeling that this might be correct. Will update.

internal/servers/worker/status.go
vancluever: Some notes for myself and anyone else looking:

My assumption, from a purely status perspective, was that the worker could advance the state of the session appropriately once the connection came back online (_if_ it came back online in time, before it got reaped), but that does not seem to be the case. Based on this, purely closing the connections should be fine.

The question then is, how does the _connection's_ state get updated properly? Looks like we clean connections appropriately in the controller's [`Status`](https://github.com/hashicorp/boundary/blob/e443bd4bb448f0351096503d2388fe2e80102064/internal/servers/controller/handlers/workers/worker_service.go#L54-L166) handler, but the connection will still be accounted for if the status is not properly advanced in the worker. Nothing seems to do so at this time; synchronizing the connection status on close seems to be [dependent on the response from the controller](https://github.com/hashicorp/boundary/blob/e443bd4bb448f0351096503d2388fe2e80102064/internal/servers/worker/session.go#L288-L294).

The logical answer seems to be advancing the connection state if it can't be properly advanced via the controller.

vancluever: I still wonder if this is the right approach; unless the session should be able to be re-used after we've terminated all connections on it, this _seems_ like the cleanest approach - advancing the state allows already established tear-down logic to take place and for the session to eventually be removed on the worker.

All that seems to be necessary here is:

* Ensuring that the status gets advanced [here](https://github.com/hashicorp/boundary/blob/e443bd4bb448f0351096503d2388fe2e80102064/internal/servers/controller/handlers/workers/worker_service.go#L110-L115) on appropriate update. 
* Dead connection cleanup will handle canceling the connections as appropriate.
* Cleaning up _orphaned sessions_ if they are not reported by the worker (this seems to be something that could possibly be an issue, independent of any work here currently).

For completeness, we might also want to manually force connection states on the worker if [`CloseConnection`](https://github.com/hashicorp/boundary/blob/e443bd4bb448f0351096503d2388fe2e80102064/internal/servers/worker/session.go#L243) fails too, but it does not seem 100% necessary as the worker is likely to reap it from the session on the next `Status` call.

jefferai: The problem is that sessions can live cross-worker; connections for a session can be handled by any worker so long as the client provides the right TOFU token. So unless a client has specifically indicated to a worker that it wants the session canceled, all a worker can know is that its connections related to a session should be closed. There may be other workers still connected to controllers handling other connections for that session that aren't hitting this failure condition.

vancluever: :+1: gotcha; this makes sense. Will update.

vancluever:
Moving to draft for now while I add the remaining tests.

Logic has been moved to closing connections instead of marking session statuses. We re-use as much code as possible, and ensure that connections are locally marked as closed if we can't reach the controller. This ensures that if a status heartbeat is sent before the whole session has a chance to be cleaned up, the correct status will be sent and the connections will be marked as closed in the database.

vancluever:
Marking this as ready for review to keep it going. Optimally I'd like to get an acceptance test going for this, but I'm still groking what is necessary to do this - it doesn't appear we have good precedent anywhere in E2E testing yet (other than possibly in bats tests, which I don't really want to do here, given what we need to simulate). I've located the bits in `boundary connect` that will help in doing this, but obviously working on it will take a bit of time, and I'm not too sure if it needs to hold up the work.

vancluever:
Moving this back to draft for now while the E2E test is WIP.

Got the authorization/connect/send-recv test working, but need to figure out the best way to introduce failure now.

vancluever:
Marking as ready for review again!

There are now 2 E2E tests, one that tests a single controller and another that tests multi-controller connections.

Both of these use a tool I wrote: https://github.com/vancluever/dwadle. I'll be making some updates to it over the next little bit, but it's functional for the E2E tests and nothing that's being used right now should be affect by changes.

vancluever:
I think the `tests/cluster` package now runs longer than 10 minutes with no output, which is why the Circle job has been failing. I've updated the (output) timeout for acceptance tests for output to 30m.

vancluever:
Closing in favor of #1330.

PR #1265, opened 2021-06-02
Add API/CLI for managed groups

jefferai:
This required a change to the template to avoid a double-"filter" flag,
hence various other commands being regenerated. It also adds support for
showing the member IDs of a managed group on read of the group, and the
managed group IDs when reading an account.


jimlambrt (approved):
Looks great!

vancluever (approved):
LGTM here! Review was mostly observational but everything generally made sense!

PR #1267, opened 2021-06-02
protobuf: remove unused imports

mgaffney:
This removes unused imports in the protobuf files which are reported as warnings by the `protoc` compiler.


talanknight (approved):
Thank you!

PR #1269, opened 2021-06-03
Add a Docker entrypoint script and modify Dockerfiles to mimic Vault, to enable mlock inside Docker containers

omkensey:
* Add entrypoint script with logic borrowed from Vault's.  This script adds
IPC_LOCK capability at runtime and then executes Boundary under its own user.

* Modify Dockerfiles to use the entrypoint script.

* Remove USER directives from the Boundary Dockerfiles.  This is required to
allow the entrypoint to run setcap.

Because Boundary itself is no longer the entrypoint, the behavior of the
container with respect to passing or overriding arguments is different now.
The Vault script's logic around handling this was not copied entirely -- only
to the extent required to allow the `server` subcommand to run properly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 
 ## Next
 
+### New and Improved
+
+* mlock: Add a Docker entrypoint script and modify Dockerfiles to mimic Vault
+  ([PR](https://github.com/hashicorp/boundary/pull/1269))
+
 ## 0.2.3 (2021/05/21)
 
 ### Deprecations/Changes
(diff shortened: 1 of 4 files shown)


malnick (approved):
👍 Thanks Joe!

omkensey:
> I'm running the Docker container for Boundary v0.7.1, but I'm running into the IPC lock error message when my Boundary container starts up.
> 
> `Couldn't start Boundary with IPC_LOCK. Disabling IPC_LOCK, please use --privileged or --cap-add IPC_LOCK`

You'll still get that if you run it as an ordinary `docker run`, because Docker doesn't add that capability by default.  But if you add `--cap-add IPC_LOCK` (or add that capability in the Boundary container config in whatever orchestrator you're using) then `mlock` should be enabled and you won't see that error any more.

PR #1272, opened 2021-06-03
Vault integration: Refactor: rename database views and associated structs

mgaffney:
This change renames two database views and associated go structs to use a consistent naming pattern. Views ending in 'private' contain encrypted credentials needed for connecting to Vault. Views ending in 'public' mirror the 'private' views but do not include any encrypted values.

This change does not include any changes to functionality or behavior.


talanknight (approved):
One thing I would like to see regarding the go structs is identifying when the secrets themselves are present in a decrypted state.  For example lookupPrivateStore has the decrypted secret in it,  the result of toCredentialStore() from a private store does _not_ have any secrets.

louisruch (approved):
LGTM

mgaffney:
@talanknight could you give an example of what you would like to see?

PR #1273, opened 2021-06-03
Adding SAD target operations for credential libraries to the SDK.

talanknight:
Not included in here is the ability to include the purpose along with the credential library id.  All purposes default to "application" as the API calls out in it's documentation.


louisruch (approved):
LGTM this gen work is really nice

mgaffney (approved):

PR #1275, opened 2021-06-03
base types for events package

jimlambrt:
This is first of 3 PRs to add the events package and they should be reviewed in order.  After this one, then #1276 and #1277.

This PR will be merged into main after the last one in the series (#1277) has been approved.

BTW, none of these 3 PRs will affect any of the rest of Boundary.  They are stand-alone infrastructure for eventing audits, errors, and observations.  Subsequent PRs may change how some of this infrastructure works and we work to integrating the events package with the rest of Boundary and we continue to receive feedback from the Event Logging RFCs.


internal/observability/events/event_type.go
```diff
@@ -0,0 +1,10 @@
+package event
+
+type Type string
+
+const (
+	EveryType       Type = "*"
```
s-christoff, line 6: Opinions on changing EveryType to AllType(s)?

s-christoff (dismissed):
Some minor nits but nothing huge - thanks for laying this foundation :)

internal/observability/events/event.go
```diff
@@ -0,0 +1,51 @@
...
+
+// RequestInfo defines the fields captured about a Boundary request.
+type RequestInfo struct {
+	Id       string `json:"id,omitempty"`
+	Method   string `json:"method,omitempty"`
+	Path     string `json:"path,omitempty"`
+	PublicId string `json:"public_id,omitempty"`
+}
+
+// UserInfo defines the fields captured about a user for a Boundary request.
```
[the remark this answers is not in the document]
jimlambrt: Agreed. I was sort of holding off a bit on documenting them until your RFC was more ratified (which includes these structs).  Once we have the "royal" blessing, we can open a PR to spruce up some godoc comments for these.

internal/observability/events/event_audit.go
```diff
@@ -0,0 +1,142 @@
...
+	Timestamp      time.Time    `json:"timestamp"`              // std audit field
+	RequestInfo    *RequestInfo `json:"request_info,omitempty"` // boundary field
+	Auth           *Auth        `json:"auth,omitempty"`         // std audit field
+	Request        *Request     `json:"request,omitempty"`      // std audit field
+	Response       *Response    `json:"response,omitempty"`     // std audit field
+	SerializedHMAC string       `json:"serialized_hmac"`        // boundary field
+	Flush          bool         `json:"-"`
+}
+
+func newAudit(fromOperation Op, opt ...Option) (*Audit, error) {
```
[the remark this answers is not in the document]
jimlambrt: The entry point for writing Audit events will be: event.WriteAudit(ctx context.Context, a *Audit, opt...Option)  The Audit struct has to be exported since it's referenced in the exported WriteAudit(...), but newAudit(...) should never be called by anyone outside the event package.

internal/observability/events/event_audit.go
```diff
@@ -0,0 +1,142 @@
...
+		return nil, errors.New(errors.InvalidParameter, op, "missing from operation")
+	}
+	opts := getOpts(opt...)
+	if opts.withId == "" {
+		var err error
+		opts.withId, err = newId(string(AuditType))
+		if err != nil {
+			return nil, errors.Wrap(err, op)
+		}
+	}
```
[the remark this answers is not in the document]
jimlambrt: datetime

internal/observability/events/event.go
```diff
@@ -0,0 +1,51 @@
+package event
```
mgaffney, line 1: The package name and directory name do not match.

internal/observability/events/event_type.go
jimlambrt: I think EveryType (singular) reads better than AllType (singular) and all the type definitions are singulars.

internal/observability/events/eventer.go
```diff
@@ -0,0 +1,12 @@
+package event
+
+const (
+	OpField          = "op"           // OpField in an event.
```
mgaffney, line 4: Are the comments for each `const` field correct?

internal/observability/events/options.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+}
+
+func WithRequestInfo(i *RequestInfo) Option {
+	return func(o *options) {
+		o.withRequestInfo = i
+	}
+}
+
+func WithNow(now time.Time) Option {
```
mgaffney, line 66: `event.WithNow` reads odd to me.

mgaffney (changes requested):
Many (most?) of the exported types, functions, constant, etc are missing docs.

internal/observability/events/eventer.go
mgaffney: I misread the comment. Please ignore.

internal/observability/events/options.go
s-christoff: +1 to this

jimlambrt: Hum... perhaps.  BTW, it's only used in tests, but had to be exported since some of those tests will not be in the event package but rather in event_test

jimlambrt:
> Many (most?) of the exported types, functions, constant, etc are missing docs.
this will be addressed in a future PR. See my previous comment: https://github.com/hashicorp/boundary/pull/1275#discussion_r645734666

internal/observability/events/event_audit.go
```diff
@@ -0,0 +1,142 @@
...
+	var validId string
+	var initValidId sync.Once
+	payload := Audit{}
+	for i, v := range events {
+		gated, ok := v.Payload.(*Audit)
+		if !ok {
+			return "", nil, errors.New(errors.InvalidParameter, op, fmt.Sprintf("event %d is not an audit payload", i))
+		}
+		initValidId.Do(func() {
```
vancluever: Curious what the intent here is for using `sync.Once` to initialize this - only thing I can think of is to protect against concurrency, but I don't see anything here that would need that? Would a simple empty check/non-empty assertion on `gated.Id` be better?

```go
if gated.Id == "" {
  return "", nil, errors.New(errors.InvalidParameter, op, fmt.Sprintf("event %d: Id is required", i))
}

if validId == {
  validId = gated.Id
}
```

internal/observability/events/options.go
jimlambrt: `event` is the package name, so not sure that can be changed.  The prefix `With` is our standard for options and `Now` aligns with the std lib `time.Now()` and every other instance of a time.Time that represents now.

internal/observability/events/event_audit_test.go
```diff
@@ -0,0 +1,278 @@
...
+		{
+			name:         "missing-op",
+			wantErrMatch: errors.T(errors.InvalidParameter),
+		},
+		{
+			name:   "valid-no-opts",
+			fromOp: "valid-no-opts",
+			want: &Audit{
+				Version: AuditVersion,
+				Type:    string(ApiRequest),
```
vancluever: Should we type out `Audit.Type` so that this kind of type conversion is not necessary? Assuming this field will always be non-arbitrary.

jimlambrt:
@mgaffney @s-christoff added a commit with godoc comments for exported fields except for the unratified audit related fields and/or structs.  Ty for the suggestion!

internal/observability/events/event.go
jimlambrt: fixed. ty!

internal/observability/events/event.go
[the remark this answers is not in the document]
jimlambrt: @s-christoff I've pushed a commit with additional godoc comments for the exported bits with the exception of unratified audit fields and/or structs.

internal/observability/events/event_error_test.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+}
+
+func Test_errvalidate(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name            string
+		id              string
+		op              Op
+		wantErr         error
```
vancluever: _Minor_ just noting that I had to actually check here to see that `wantErr` was actually the valid case here, and not error validation. Might be worth just changing this to `want`?

internal/observability/events/event_observation_test.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+}
+
+func Test_observationvalidate(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name            string
+		id              string
+		op              Op
+		wantErr         error
```
vancluever: This field is never referenced in any of the tests, so the `valid` test is not being asserted, for example.

internal/observability/events/options.go
vancluever: I think `WithTime` would be better? Given that it takes a `time.Time` and does not generate one via `time.Now()`.

vancluever (changes requested):
Agreeing with @mgaffney's points for sure re: comments and package naming (although some of these could be addressed now as I have not hit "refresh" yet 😜 )

Also some other things that need fixing including a test where an assertion seems to be getting missed.

internal/observability/event/options.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	return opts
+}
+
+// Option - how Options are passed as arguments.
+type Option func(*options)
+
+// options = how options are represented
+type options struct {
+	withId          string
```
louisruch, line 21: Perhaps `withId Id`

louisruch (commented):
This looks great, I have a couple questions and comments, I see these PR's are getting some traction on reviews - I read through previous comments sorry in advance if I doubled up on any comment/question

internal/observability/event/id.go
```diff
@@ -0,0 +1,15 @@
+package event
+
+import (
+	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/errors"
+)
+
+func newId(prefix string) (string, error) {
```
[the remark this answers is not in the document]
louisruch: I see you are using `newId` in follow up PRs outside the scope of Type it seems

internal/observability/events/event_audit_test.go
jimlambrt: Audit.Type is string based on an effort to adhere to the Audit schema put forth on ENG-030 and CSL-121.  I used a defined type internally because it afford some advantages... but we should adhere to the previous work by the Nomad and Consul teams when it comes to the outbound audit event data.

internal/observability/events/event_error_test.go
jimlambrt: ty!

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/hashicorp/cap v0.0.0-20210518163718-e72205e8eaae
 	github.com/hashicorp/dbassert v0.0.0-20200930125617-6218396928df
 	github.com/hashicorp/errwrap v1.1.0
+	github.com/hashicorp/eventlogger v0.0.0-20210523164657-c216620e1746 // indirect
```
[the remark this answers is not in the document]
jimlambrt: ty!

internal/observability/event/event_audit.go
```diff
@@ -0,0 +1,145 @@
...
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/eventlogger"
+)
+
+// AuditEventType defines the version of audit events
```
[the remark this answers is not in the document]
jimlambrt: ty!

internal/observability/events/event_observation_test.go
jimlambrt: yep... I went with wantErrMatch and forgot to remove it.

ty!

internal/observability/event/event_audit.go
```diff
@@ -0,0 +1,145 @@
...
+			return "", nil, errors.New(errors.InvalidParameter, op, fmt.Sprintf("event %d has an invalid id: %s != %s", i, gated.Id, validId))
+		}
+		if gated.Version != AuditVersion {
+			return "", nil, errors.New(errors.InvalidParameter, op, fmt.Sprintf("event %d has an invalid version: %s != %s", i, gated.Version, AuditVersion))
+		}
+		if gated.Type != string(ApiRequest) {
+			return "", nil, errors.New(errors.InvalidParameter, op, fmt.Sprintf("event %d has an invalid type: %s != %s", i, gated.Type, string(AuditType)))
+		}
+		if gated.RequestInfo != nil {
+			payload.RequestInfo = gated.RequestInfo
```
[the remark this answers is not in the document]
jimlambrt: the last event will win by design.

internal/observability/events/event_audit.go
jimlambrt: ty!

louisruch (dismissed):
lgtm

jimlambrt:
I've rolled back the directory name change for now... it will be reapplied before merging but it was causing all sorts of conflicts with the upstream PRs

vancluever (approved):
:shipit: 

Just want to bring up as a minor point https://github.com/hashicorp/boundary/pull/1275/files#r645785137 (the thread over `WithNow` and how it might be better as `WithTime` since this option is not actually deriving or enforcing a current time.

PR #1281, opened 2021-06-05
Fix comment in template for CLI

tiqwab:
Codes in `internal/cmd/commands` are generated by `make cli` not `make api`?


louisruch (approved):
Thank you

PR #1286, opened 2021-06-07
A drier version of `NewEventer(...)`  

jimlambrt:
I'm not sure the factory is more readable. It does have a  bit more complexity like new mutex since the new funcs which are used in aggregate for the factory could be called in a concurrent scenario by future devs.  

While not my preferred way to structure a factory func both @s-christoff and @louisruch expressed concern over the length of the original `NewEventer(...)` so I'm happy to propose an alternative decomposition.

### DIFF
--- a/internal/observability/event/eventer.go
+++ b/internal/observability/event/eventer.go
@@ -51,6 +51,8 @@ type Eventer struct {
 	flushableNodes []flushable
 	conf           EventerConfig
 	logger         hclog.Logger
+
+	l sync.RWMutex
 }
 
 var (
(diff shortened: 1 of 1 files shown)


internal/observability/event/eventer.go
```diff
@@ -121,15 +146,31 @@ func NewEventer(log hclog.Logger, c EventerConfig, opt ...Option) (*Eventer, err
...
+	// this is simply a way to reconcile the timestamps for "now" that the
+	// broker adds to the events with a known value.
 	if !opts.withNow.IsZero() {
-		e.broker.StopTimeAt(opts.withNow)
+		b.StopTimeAt(opts.withNow)
 	}
+	return b
+}
+
+type pipeline struct {
```
s-christoff, line 157: If we go forward this PR I would really like to update `pipeline` and `pipelines` to be a little bit more.. differently named?

jimlambrt:
After a bit of discussion with @s-christoff and @louisruch, we're not moving forward with this PR right now.  Perhaps in the future, new requirements will require further decomposition, but for now it doesn't appear it's helpful.

PR #1289, opened 2021-06-07
Update GrantsForUser to handle managed groups

jefferai:
This adds support to GrantsForUser to handle looking up grants conveyed via managed groups, and updates associated tests as well.


jimlambrt (approved):
Excellent!

vancluever (approved):

PR #1291, opened 2021-06-07
Add Vault Credential Store CLI Commands

talanknight:
This PR adds a template param that prefixes returned field errors with the subaction prefix. This allows flags to be prefixed with the subtype and decreases the chance that a flag will collide with a more general flag.

The code in this PR accepts all string values as inline fields.  Reading fields from files will come in a later PR.


internal/cmd/commands/credentialstorescmd/vault_funcs.go
```diff
@@ -0,0 +1,186 @@
...
+			f.StringVar(&base.StringVar{
+				Name:   namespaceFlagName,
+				Target: &c.flagNamespace,
+				Usage:  "The vault namespace the store should use.",
+			})
+		case vaultCaCertFlagName:
+			f.StringVar(&base.StringVar{
+				Name:   vaultCaCertFlagName,
+				Target: &c.flagCaCert,
+				Usage:  "The CA Cert to use when connecting with vault.",
```
mgaffney, line 74:
```suggestion
				Usage:  "The CA Cert to use when connecting to vault.",
```

internal/cmd/commands/credentialstorescmd/vault_funcs.go
```diff
@@ -0,0 +1,186 @@
...
+			})
+		case tlsServerNameFlagName:
+			f.StringVar(&base.StringVar{
+				Name:   tlsServerNameFlagName,
+				Target: &c.flagTlsServerName,
+				Usage:  `The TLS server name used when connecting to vault.`,
```
mgaffney, line 80:
```suggestion
				Usage:  `Name to use as the SNI host when connecting via TLS.`,
```
This is [the wording](https://www.vaultproject.io/docs/agent#tls_server_name) from the Vault agent configuration page.

internal/cmd/commands/credentialstorescmd/vault_funcs.go
```diff
@@ -0,0 +1,186 @@
...
+func extraVaultFlagsFuncImpl(c *VaultCommand, set *base.FlagSets, _ *base.FlagSet) {
+	f := set.NewFlagSet("Vault Credential Store Options")
+
+	for _, name := range flagsVaultMap[c.Func] {
+		switch name {
+		case addressFlagName:
+			f.StringVar(&base.StringVar{
+				Name:   addressFlagName,
+				Target: &c.flagAddress,
+				Usage:  "The vault address.",
```
mgaffney, line 62:
```suggestion
				Usage:  "The address of the Vault server. This should be a complete URL such as https://127.0.0.1:8200",
```
Taken from the Vault Agent [configuration page](https://www.vaultproject.io/docs/agent#address).

mgaffney (approved):

PR #1294, opened 2021-06-08
docs: update user groups workflow with new pw account prefix

malnick:
Updates docs to use the new prefix for password accounts `acctpw`


malnick:
This was already present in https://github.com/hashicorp/boundary/pull/1290/files - closing.

PR #1304, opened 2021-06-09
Add token revocation job and cred renewal/revocation job

louisruch:
### What does this PR do

Adds token revocation job 
Add credential renewal and revocation job


louisruch:
@talanknight @mgaffney PTAL
This PR is rather large, I was working simultaneously on all three while Mike was finishing the revoke work and view.  But it is pretty repetitive, not as boiler platey as CLI/API work though.  Let me know if you would prefer this to be broken up into three separate PRs

mgaffney (changes requested):
Looks great! Just some questions about the changes to the test helpers.

internal/credential/vault/jobs_test.go
```diff
@@ -2,22 +2,152 @@ package vault
...
+	expiration time.Duration) *Token {
+	t.Helper()
+	assert, require := assert.New(t), require.New(t)
+	kmsCache := kms.TestKms(t, conn, wrapper)
+	rw := db.New(conn)
+
+	secret, _ := v.CreateToken(t)
+	acc, err := secret.TokenAccessor()
+	require.NoError(err)
+	inToken, err := newToken(cs.PublicId, []byte(secret.Auth.ClientToken), []byte(acc), expiration)
```
talanknight, line 50: nit: This seems like it could be made a few lines shorter.  Since you are using secret.Auth to get ClientToken, consider doing the same to get `sec.Auth.Accessor` and bypass the require.NoError and the TokenAccessor() call.

internal/credential/vault/testing.go
```diff
@@ -590,28 +590,22 @@ func (v *TestVaultServer) CreateToken(t *testing.T, opt ...TestOption) (*vault.S
 
 // LookupToken calls /auth/token/lookup on v for token. See
 // https://www.vaultproject.io/api-docs/auth/token#lookup-a-token.
-func (v *TestVaultServer) LookupToken(t *testing.T, token string) *vault.Secret {
+func (v *TestVaultServer) LookupToken(t *testing.T, token string) (*vault.Secret, error) {
```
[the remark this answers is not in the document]
louisruch: 👍

internal/credential/vault/jobs.go
```diff
@@ -222,3 +241,478 @@ func (r *TokenRenewalJob) Name() string {
...
+		return errors.Wrap(err, op)
+	}
+
+	token := s.token()
+	if token == nil {
+		// Store has no token to revoke
+		return nil
+	}
+
+	vc, err := s.client()
```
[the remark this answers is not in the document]
louisruch: The default timeout is 60 seconds, when creating a new client we first get the default here:
https://github.com/hashicorp/boundary/blob/520f21ca96774cb1f1c69f4fbdcc4a8b98e1c1aa/internal/credential/vault/vault.go#L51
Which calls the vault api default which includes the timeout:
https://github.com/hashicorp/vault/blob/master/api/client.go#L158

internal/credential/vault/jobs.go
```diff
@@ -222,3 +241,478 @@ func (r *TokenRenewalJob) Name() string {
...
+	if err != nil {
+		return errors.Wrap(err, op)
+	}
+
+	var respErr *vault.ResponseError
+	err = vc.revokeToken()
+	if ok := errors.As(err, &respErr); ok && respErr.StatusCode == http.StatusForbidden {
+		// Vault returned a 403 when attempting a revoke self, the token is already expired.
```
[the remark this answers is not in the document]
louisruch: A 403 indicates the token is not valid on the Vault server we are trying to authenticate on.  I don't think the address change should be a concern for this job, but you are correct that if the user changes the address, and that address is not just a DNS change but a different server we will then set this token to `expired` while it is still potentially a valid token on the original server.  Lets discuss in a call or over slack tomorrow.

internal/credential/vault/jobs_test.go
louisruch: ty!

internal/credential/vault/jobs.go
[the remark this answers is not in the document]
louisruch: created https://hashicorp.atlassian.net/browse/ICU-1765 to track

internal/credential/vault/testing_test.go
```diff
@@ -414,6 +414,7 @@ func TestTestVaultServer_LookupLease(t *testing.T) {
 	time.Sleep(time.Second)
 
 	leaseLookup := v.LookupLease(t, cred.LeaseID)
+	require.NotNil(leaseLookup)
```
mgaffney, line 417: Why add this check when `LookupLease` already does the same check?

internal/credential/vault/testing.go
```diff
@@ -634,6 +634,19 @@ func (v *TestVaultServer) LookupToken(t *testing.T, token string) *vault.Secret
 	return secret
 }
 
+// LookupTokenError calls /auth/token/lookup on v for the token. It expects the lookup to fail,
+// and returns the error received.  See
+// https://www.vaultproject.io/api-docs/auth/token#lookup-a-token.
+func (v *TestVaultServer) LookupTokenError(t *testing.T, token string) error {
```
mgaffney, line 640: Why return the error when this is asserting there is an error?

internal/credential/vault/vault_test.go
```diff
@@ -31,16 +31,18 @@ func TestClient_RenewToken(t *testing.T) {
 
 	_, token := v.CreateToken(t)
 	secretLookup := v.LookupToken(t, token)
+	require.NotNil(t, secretLookup)
```
mgaffney, line 34: Why add this check when `LookupToken` already does it?

internal/credential/vault/vault_test.go
```diff
@@ -68,6 +70,7 @@ func TestClient_LookupToken(t *testing.T) {
 
 	_, token := v.CreateToken(t)
 	secretLookup := v.LookupToken(t, token)
+	require.NotNil(secretLookup)
```
mgaffney, line 73: Same as above

mgaffney (changes requested):

internal/credential/vault/testing.go
louisruch: My thought was that the caller could assert it is a specific error that test is looking for.

louisruch:
@mgaffney PTAL I updated the lookupTokenError to be more clear and handle all checks as we discussed

internal/credential/vault/testing.go
```diff
@@ -644,17 +647,21 @@ func (v *TestVaultServer) LookupToken(t *testing.T, token string) *vault.Secret
 	return secret
 }
 
-// LookupTokenError calls /auth/token/lookup on v for the token. It expects the lookup to fail,
-// and returns the error received.  See
+// VerifyTokenInvalid calls /auth/token/lookup on v for the token. It expects the lookup to
+// fail with a StatusForbidden.  See
 // https://www.vaultproject.io/api-docs/auth/token#lookup-a-token.
-func (v *TestVaultServer) LookupTokenError(t *testing.T, token string) error {
+func (v *TestVaultServer) VerifyTokenInvalid(t *testing.T, token string) {
```
mgaffney, line 653: Nice name! ty!

mgaffney (approved):

talanknight (approved):

PR #1308, opened 2021-06-10
Integrate with Vault to retrieve and manage per session credentials

mgaffney:
- Add migrations for credential base types
- Name unique and foreign key constraints
- Add tables for Vault Credential subsystem
- Add tables for Target subsystem
- Add tables for Session subsystem
- Add a comment to each new table
- Run migrations
- Update schema based on feedback in the RFCs
- Changes table names for vault subsystem to follow naming convention
- Rename 'credential type' to 'credential purpose'
- Remove unnecessary oplog tickets
- Make vault_address a wt_url type
- Add client certs used for mTLS to table comment
- Add foreign keys to KMS table
- Run make migrations
- Store expiration timestamps not durations
- Refactor: move declarations
- Updates to vault token table
- Add missing foreign keys from subtypes to credential
- Fix the vault lease table
- Run 'make migrations'
- Add protobufs for credential and vault packages
- Fix spelling error magnified by copy/paste development
- Add vault token status enumeration table
- Use standard suffix for foreign key constraint names
- Run 'make migrations'
- Add space after period for readability of concatenated comments
- WIP
- Add options for the vault credential store
- Add options to credential store
- Rename token status column
- Update protobufs
- Remove stutter: rename vaultToken to token
- Schema: stores can have multiple tokens but only one current token
- Add vault.Token resource
- Use []byte for certificates and certificate keys
- Add vault.ClientCertificate resource
- Add additional CredentialStore tests
- Remove dead code
- Make value for oplog resource-type key more descriptive
- Add vault.CredentialLibrary resource
- Remove unused test parameter
- Add TestCredentialLibraries function
- Refactor: rename const for insert token query
- Refactor: move token method from test file to production file
- Add test helper for creating vault tokens
- Fix gorm tags in proto file
- Add vault.Lease resource
- Add 'resource-public-id' to the lease's oplog
- Add vault test helpers
- Add support for renewing a Vault token
- Updates based on PR feedback
- Fix for tests failing in CircleCI
- Attempt #2 to fix tests failing in CircleCI
- Attempt 3 to fix tests failing in CircleCI
- Refactor: move migrations to higher number to prevent conflict with main
- Add vault repository
- Allow a client certificate to be created without a Store ID
- Add WithClientCert option
- Support token and client certificate in credential store
- Change type of CaCert to []byte
- Add domain errors for Vault tokens
- Add repository CreateCredentialStore
- Remove extra whitespace in imports
- Remove unnecessary call
- Run migrations
- Regenerate pb.go files with updated protoc version
- godoc: Add Vault specific details to CreateCredentialStore
- Add client cert and token to test credential store
- Remove version field from vault token
- Add repository LookupCredentialStore
- Fix tests
- Add comment to exported method
- Consolidate test helpers into a single file
- Add repository UpdateCredentialStore
- WIP: UpdateCredentialStore
- WIP: delete CA cert fixed
- WIP: enable all tests for UpdateCredentialStore
- Create database view for per credential store vault connection data
- Add repository lookupPrivateCredentialStore
- Add repository lookupPrivateCredentialStore
- Refactor: rename test field
- WIP: support token updates in UpdateCredentialStore
- Fix tests
- Add tests for updating a token
- Support updating the vault address
- Only the store_id column is immutable for a client certificate
- Regenerate protobufs after merge
- Complete implementation for UpdateCredentialStore
- Add repository ListCredentialStores
- Add repository DeleteCredentialStore
- Orphaned -> Orphan
- Fix c&p token -> key
- Add repository CreateCredentialLibrary
- Updates from PR feedback
- Add options for HTTP method and request body
- Add support for POST and request body in vault credential library
- Add repository UpdateCredentialLibrary
- Remove redundant parentheses
- Add repository LookupCredentialLibrary
- Add repository DeleteCredentialLibrary
- Add repository ListCredentialLibraries
- Refactor: bump up migrations directory number
- Run make migrations
- Fix Method const block
- Replace token_sha256 with token_hmac
- Add protobufs for target credential library
- Add target.TargetCredentialLibrary
- Run tests in parallel
- Assert the test got what it wanted
- Creating the credential store resource in the API along with mask mapping.
- Refactor: rename to eliminate stutter
- Add test helper for creating target credential libraries
- Changing vault token to string and adding token and client certificate to api.
- Add session dynamic credential
- Add test helper for creating vault leases
- Add test for writing a dynamic credential to the database
- Reverting string back to []byte.  Allow multiple proto destinations for mask mapping.
- Updating comments and regen after updating json_name field.
- Updating comments distinguishing the client certificate field as a write only field and the details of the private key behavior.
- Updating mask mapping.
- Adding subtype checks, toProto, handler setup.
- Vault token is now input only, other modifiable or optional fields are now proto well known types.
- Add List to handler.
- Refactor target host set
- Add target cred libraries
- Add create credential store to request handler
- Update lookupm update and create target methods
- Adding Get Method
- Remove nolonger needed debug logic.
- Adding delete request handler.
- Add update credential store
- Adding additional checks and tests for client certs, vault ca certs, and tokens.
- make fmt
- Update internal/types/resource/resource.go
- Move CredentialRepoFactory to the vault package.
- Change resource enum to use iota.  Add credential store and library to resource test.
- Separate test expectations for NewCredentialStore vs creating a credential store in DB.
- Creating Cred Library API and handlers
- Rename certificate_key to client_certificate_key.
- calc changes in db
- Update internal/target/options.go
- make gen after merge.
- Feedback
- Define base types for the credential package
- Add hostSet to return early setcredlib
- Add additional add/del tests
- Fix up the tests for library stuff.
- Fixing the tests for credential libraries.
- Move to using a newly defined credential.Store instead of the subtype defined CredentialStore.
- Use the credential defined Library interface instead of the one defined in subtype.
- Add test for unsetting http method and request body.
- Remove Http Method and Body requirements around in memory CredentialLibrary. Apply GET method default at Create and Update time.
- Register the library handler to the server muxer.
- Create Target API for SAD operations on credential libraries.
- Fixing comments and op const.
- Add the vault token hmac in Update and List of Credential Stores.
- Add test helper for mounting the Vault PKI secrets engine
- Refactor: convert Vault TLS settings into a testing option
- Add database secrets engine to Vault test server
- Refactor: rename internal Vault client methods
- Add wt_sentinel domain type
- Refactor: rename vault lease id to external id
- Refactor: renew table
- Refactor rename vault.Lease to vault.Credential
- Refactor: rename files
- Refactor: finish renaming vault_lease to credential
- Invert the dependency between the credential and vault packages
- Add database view for connecting to Vault via a library
- Reorder migrations so domain types are first
- Add Get and Post to internal Vault client
- Allow empty bodies for POST methods to Vault
- Make migrations
- Change library request body from a string to a []byte
- Refactor: renew privateCredentialStore to privateStore
- Add tests for http body in a library
- Add scope_id column to private credential library view
- Add comments for the target_library view and enumerated the columns pulled from the target_credential_library table.
- Add support for specifying purpose in the API (currently only allowed purpose is "application").
- Update comment for outgoing credential library proto.
- Add a certificate key hmac field for credential store.
- Implement private library
- Enable PostgreSQL infinity timestamp
- Fix wt_sentinel
- Fix credential interface
- Allow empty externalId for a vault credential
- Make the vault credential library a credential.Library
- Add convenience methods to db/timestamp.Timestamp
- Allow vault credential to have an infinite expiration
- Add polices to Vault test server
- Move lookup to the end for the returned resource.
- Add error for vault requests that fail
- Implement IssueCredentials
- Fix missing dependency after merge of main
- Fix spacing of error message.
- Rename SubType to Subtype
- Remove unused testing logic.
- Add documentation for exported TargetLibrary struct.
- No longer create a new repo and call it's interface from inside update transaction.
- Return the looked up cred store when updating.
- make gen after merge.
- Fix TLS settings for the Vault test server
- Run make fmt
- Panics are bad
- Fix comment in query
- Fix grammar in doc comment
- Clarify TODO comment about error handling when issuing credentials
- Add additional assertion
- Move error code closer to the source
- Change http request body type to a []byte
- Docker: only run tests requiring docker on supported platforms
- Fix test compilation failures
- Fix dockertest dependency
- Add additional test for credential subtypes
- Initial token renewal job
- Add jira ticket number to TODO comment
- Only assert unless it is required
- Non sentinel values only have to be greater than 0
- Fix down migrations
- Udpates due to merge
- Sentinel values only have to be greater than 0
- Create credential related sdk structs and methods.
- Post merge interface updating.
- Removing scratch code that was not cleaned up.
- updates
- Convert client certificate and client certificate keys into pem blocks and validate.
- Add newline at end of file.
- Faster check for "s" at the end of a resource name when generating from templates.
- make fmt
- updates
- Add dynamic credentials to a session
- Refactor: move contents of file
- updates
- Updates
- Add InvalidDynamicCredential error
- Assign credentials
- Replace application purpose string with enum value
- PR feedback / fix go test .
- Delete dead code
- Use common view
- Refactor: extract interface and rename method
- Refactor: move code to eliminate privPurpLibrary
- Add comment to requestMap
- Refactor: do not export methods on an unexported struct
- Update internal/db/schema/migrations/postgres/10/03_vault_credential.up.sql
- Add comments to credential Purpose constants
- Reorder switch statements to handle errors first
- PR feedback
- update
- fix merge
- Run 'make fmt'
- Refactor: rename database views and associated structs
- Adding SAD target operations for credential libraries to the SDK.
- Add status field and enumeration table for credential status
- Refactor: rename Status to TokenStatus
- Add status to vault credentials
- Add private database view for vault credentials
- Add scope_id to view and cleanup TODO comment
- Reformat embedded sql queries
- Add vault client helpers
- updates
- white space
- Add LookupLease test
- Test issue credentials with client TLS
- Add additional checks to vault ping
- Add an error for vault token capabilities and create a VaultToken error kind
- Add support for working with Vault policies
- Use Vault 1.7.2 for testing
- Add testing helper for creating a vault client with a non-root token
- Fix Vault renew lease test
- Add a var for the required capabilities of a boundary Vault token
- Enhance the test helper for adding a policy to Vault
- Allow the test role in postgresql to revoke credentials
- Remove out of date comments
- Enhance Vault test helpers for the database secrets engine
- Reformat comments
- Add a method for getting the capabilities of the current Vault token
- Add revoke lease to internal Vault client
- Reformat: combine err return line with check for nil line
- Remove redundant checks in test code
- Refactor CreateToken test helper to return the secret and the token
- Fix build failures caused by signature change in vault.CreateToken
- Fix Vault version and Vault API link in comments
- Make the capabilities String method more readable
- Refactor test to move declarations closer to first use
- Fix comment
- Implement revoke method on vault repository
- Add private credential
- Regen after merge.
- Fixing field mask related merge error.


internal/api/genapi/templates.go
```diff
@@ -24,7 +24,11 @@ func getArgsAndPaths(in []string, parentTypeName, action string) (colArg, resArg
 		strToReplace = in[len(in)-2]
 	}
 	colArg = fmt.Sprintf("%sId", strcase.ToLowerCamel(strings.ReplaceAll(strToReplace, "-", "_")))
-	colPath = fmt.Sprintf("%ss", in[len(in)-1])
+	colPath = in[len(in)-1]
```
jefferai, line 27: This change is unnecessary. The problem it's trying to fix is caused because the values in input.go are incorrect. It needs to be fixed there, not here.

jefferai: @talanknight

internal/proto/local/controller/api/resources/credentialstores/v1/credential_store.proto
```diff
@@ -0,0 +1,87 @@
...
+  // The PEM encoded CA cert to verify the vault server's ssl certificate
+  google.protobuf.StringValue ca_cert = 30 [json_name = "ca_cert", (custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = { this: "attributes.ca_cert" that: "CaCert" }];
+
+  // The value to use as the SNI host when connecting to vault via TLS.
+  google.protobuf.StringValue tls_server_name = 40 [json_name = "tls_server_name", (custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = { this: "attributes.tls_server_name" that: "TlsServerName" }];
+
+  // When set to true verification of the TLS certificate is disabled.
+  google.protobuf.BoolValue tls_skip_verify = 50 [json_name = "tls_skip_verify", (custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = { this: "attributes.tls_skip_verify" that: "TlsSkipVerify" }];
+
+  // Input only field. The current vault token used by this credential store for creating new credentials.
```
jefferai, line 70: Down here you switch to "Input only field" and "Output only field" instead of "Output only" in the message above. Will grpc-gateway interpret that correctly? Not sure if it must literally be `Output only.` or if it just looks for the existence of those two words regardless of what follows them. @talanknight

internal/api/genapi/templates.go
talanknight: But librarie and librarys aren't words.

internal/servers/controller/handlers/credentialstores/vault.go
```diff
@@ -0,0 +1,70 @@
...
+		return nil, nil, errors.New(errors.InvalidParameter, op, "private key payload contained multiple pem blocks")
+	}
+
+	bs, err := decodePemBlocks(cert)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, op, errors.WithMsg("failed to parse client certificate into pem blocks"))
+	}
+	pkIdx := -1
+	for i, b := range bs {
+		if !strings.Contains(strings.ToLower(b.Type), "certificate") {
```
jefferai, line 36: FWIW, the PEM "standard" uses upper case for block declarations so if something is not uppercase it's likely to be unaccepted in a lot of implementations. Even Go's PEM decode function requires `BEGIN` and `END`, case-sensitively, for the block declaration. So technically you're allowing `-----BEGIN certificate-----` although it's unlikely such a block would be very portable.

I don't think this _needs_ to be changed, but I think doing a case sensitive comparison with no ToLower/ToUpper here is perfectly acceptable as well.

jefferai: @talanknight

internal/api/genapi/templates.go
jefferai: Fair point. Can we instead allow providing a plural override into the input structs and use that directly in the template?

internal/cmd/commands/credentialstorescmd/funcs.go
```diff
@@ -0,0 +1,203 @@
...
+			"  Attributes:",
+			base.WrapMap(4, maxLength, item.Attributes),
+		)
+	}
+
+	return base.WrapForHelpText(ret)
+}
+
+var keySubstMap = map[string]string{
+	"address":                     "Address",
```
jefferai, line 195: The substitutions in these maps should be "human friendly" so things like "TLS Server Name", "CA Cert", etc. @talanknight

internal/cmd/commands/credentialstorescmd/vault_funcs.go
```diff
@@ -0,0 +1,190 @@
...
+				Name:   clientCertificateKeyFlagName,
+				Target: &c.flagClientCertKey,
+				Usage:  `The client certificate's private key to use when boundary connects to vault for this store. This can be the value itself, refer to a file on disk (file://) from which the value will be read, or an env var (env://) from which the value will be read.`,
+			})
+		}
+	}
+}
+
+func extraVaultFlagHandlingFuncImpl(c *VaultCommand, f *base.FlagSets, opts *[]credentialstores.Option) bool {
+	switch c.flagAddress {
```
jefferai, line 112: Can address and token not be provided? I just ask because for any values that can't be blank, we should skip generating a Default function for it. @talanknight

internal/api/genapi/templates.go
talanknight: PathArgs is actually an outdated field anyways which seems to date back to when we had multiple resource names in our path.  I'll rework this a bit to have Read/Update and custom actions use just a variable named `id` and replace this field to just be the plural resource name since it will only then be used for building the resource paths.

internal/proto/local/controller/api/services/v1/credential_library_service.proto
```diff
@@ -0,0 +1,125 @@
...
+
+  // ListCredentialLibraries returns a list of stored Credential Libraries which are in the
+  // provided scope. The request must include the Credential Store id and if missing,
+  // malformed, or referencing a non existing scope, an error is returned.
+  rpc ListCredentialLibraries(ListCredentialLibrariesRequest) returns (ListCredentialLibrariesResponse) {
+    option (google.api.http) = {
+      get: "/v1/credential-libraries"
+    };
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      summary: "Lists all Credential Library."
```
jefferai, line 34:
```suggestion
      summary: "Lists all Credential Libraries."
```

internal/credential/vault/vault.go
```diff
@@ -0,0 +1,242 @@
...
+	cl    *vault.Client
+	token TokenSecret
+}
+
+func newClient(c *clientConfig) (*client, error) {
+	const op = "vault.newClient"
+	if !c.isValid() {
+		return nil, errors.New(errors.InvalidParameter, op, "invalid configuration")
+	}
+	vc := vault.DefaultConfig()
```
jefferai, line 52: Is there a reason below not to call `ConfigureTLS` on the config object, instead of duplicating a bunch of its stuff below? Are you trying to avoid using env vars? @mgaffney

internal/credential/vault/vault.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+	return t, nil
+}
+
+// revokeToken calls the /auth/token/revoke-self Vault endpoint. This
+// endpoint is accessible with the default policy in Vault 1.7.2. See
+// https://www.vaultproject.io/api-docs/auth/token#revoke-a-token-self.
+func (c *client) revokeToken() error {
+	const op = "vault.(client).revokeToken"
+	// The `token` parameter  s kept for backwards compatibility but is ignored, so use ""
```
jefferai, line 125:
```suggestion
	// The `token` parameter is kept for backwards compatibility but is ignored, so use ""
```

internal/credential/vault/vault.go
mgaffney: `ConfigureTLS` reads the client cert and the client cert key from files which is a problem for us since we are getting those from the database.

internal/credential/vault/testing.go
```diff
@@ -0,0 +1,853 @@
...
+	return secret, token
+}
+
+// LookupToken calls /auth/token/lookup on v for the token. See
+// https://www.vaultproject.io/api-docs/auth/token#lookup-a-token.
+func (v *TestVaultServer) LookupToken(t *testing.T, token string) *vault.Secret {
+	t.Helper()
+	require := require.New(t)
+	vc := v.client(t).cl
+	secret, err := vc.Auth().Token().Lookup(token)
```
jefferai, line 647: I think this should be `LookupSelf()` as that's more likely to be available in the default policy (or any policy).

internal/credential/vault/testing.go
```diff
@@ -0,0 +1,853 @@
...
+}
+
+// VerifyTokenInvalid calls /auth/token/lookup on v for the token. It expects the lookup to
+// fail with a StatusForbidden.  See
+// https://www.vaultproject.io/api-docs/auth/token#lookup-a-token.
+func (v *TestVaultServer) VerifyTokenInvalid(t *testing.T, token string) {
+	t.Helper()
+	require, assert := require.New(t), assert.New(t)
+	vc := v.client(t).cl
+	secret, err := vc.Auth().Token().Lookup(token)
```
jefferai, line 660: Same comment.

internal/credential/vault/testing.go
mgaffney: The [production code](https://github.com/hashicorp/boundary/blob/cdd47dc1380a479082835ca0c405656ad8fd396d/internal/credential/vault/vault.go#L162) calls `LookupSelf()`. This call is part of the test harness. It's a method on the `TestVaultServer` and it uses the test server's root token for auth when it calls Vault to avoid any policy issues.

It is mainly used in tests that need to verify the token being used in the test actually has the properties expected.

internal/credential/vault/testing.go
mgaffney: This call is also part of the test harness and it is currently only used in tests that revoke tokens and need to verify the token has been revoked.

internal/credential/vault/vault.go
mgaffney: ty!

PR #1314, opened 2021-06-10
Add go token type

louisruch:
### What does this PR do

- Adds TokenSecret and KeySecret go types to be able to redact sensitive data when stringer methods are invoked


louisruch:
@mgaffney I also added a KeySecret since it was the other encrypted value

internal/credential/vault/secret.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// MarshalJSON will redact the TokenSecret.
+func (s TokenSecret) MarshalJSON() ([]byte, error) {
+	return json.Marshal([]byte(redactedTokenSecret))
+}
+
+// KeySecret equals a Vault client certificate key.  This type provides a
+// wrapper so the secret isn't inadvertently leaked into a log or error.
+type KeySecret []byte
```
talanknight, line 31: This isn't used in the NewClientCertificate signature.  Should it be?

internal/credential/vault/secret.go
```diff
@@ -0,0 +1,49 @@
...
+func (s TokenSecret) GoString() string {
+	return redactedTokenSecret
+}
+
+// MarshalJSON will redact the TokenSecret.
+func (s TokenSecret) MarshalJSON() ([]byte, error) {
+	return json.Marshal([]byte(redactedTokenSecret))
+}
+
+// KeySecret equals a Vault client certificate key.  This type provides a
```
mgaffney, line 29:
```suggestion
// KeySecret equals a Vault client certificate private key.  This type provides a
```

internal/credential/vault/secret.go
mgaffney: +1

mgaffney (commented):
I have the same question as @talanknight

PR #1321, opened 2021-06-12
Expose NewVaultTestServer

louisruch:


louisruch:
@mgaffney The controller package in `testing` did a similar method for exposing `NewTestController` I have made a number of changes to the tf provider to verify this is working as well

malnick (approved):
First pass on this and it looks good. ~If we could merge this to `mgaffney-vault`, then I would not need to manage my own compound branch that includes this and the former branch while doing provider development.~ 

Out of overwhelming laziness, I didn't test this, but it should be easy enough to merge in a patch if we need it.

louisruch:
> First pass on this and it looks good. ~If we could merge this to `mgaffney-vault`, then I would not need to manage my own compound branch that includes this and the former branch while doing provider development.~
> 
> Out of overwhelming laziness, I didn't test this, but it should be easy enough to merge in a patch if we need it.

Make sure to pull your TF provider branch I made a number of changes yesterday.

talanknight (approved):

PR #1324, opened 2021-06-15
Update some deps

jefferai:


jefferai:
These are deps I'm responsible for and I've checked them out -- merging.

PR #1325, opened 2021-06-15
Update grpc, grpc-gateway, and a few other deps

jefferai:


jefferai:
@talanknight 

grpc-gateway has a new feature that brings it to 2.4.0 so I feel good about that. For grpc, the two that I identified of interest are https://github.com/grpc/grpc-go/pull/4311 and https://github.com/grpc/grpc-go/pull/4270

The latter one I don't _think_ will be an issue but I think tests will catch it if it is.

talanknight (approved):

PR #1328, opened 2021-06-16
Vault integration: soft delete stores

mgaffney:
# What does this PR do?

Implements a soft delete of a credential store when `vault.(Repository).DeleteCredentialStore` is called.

- [x] Add new `revoke` to `credential_vault_token_status_enm`
- [x] Add `delete_time` column to `credential_vault_store` table

- [x] Change `credential_vault_store_private` view
  - [x] Add `delete_time` column
  - [x] Include tokens with status of `revoke`
- [x] Exclude soft deleted credential stores from the `credential_vault_store_public` view

- [x] Add before update trigger to `credential_vault_store`, to prevent `delete_time` from being changed once it is set
- [x] Add after update trigger to `credential_vault_store`, when `delete_time` set to `not null`:
  - [x] update `credential_vault_token` set status to revoke where `store_id` equals current store and status is `current` or `maintaining`
  - [x] delete all rows from `target_credential_library` where library is a child of store to prevent new sessions from getting credentials

- [x] Add before insert trigger to `target_credential_library` to prevent insertion if `store_id` is a soft deleted credential store

- [x] Change `DeleteCredentialStore` to set `delete_time` to `now()` instead of deleting the credential store
  - [x] ~~Schedule revoke job to run immediately.~~ (TODO comment placeholder added)

- [x] Change `UpdateCredentialStore` to return an error if `storeId` is a soft deleted credential store
- [x] Change `LookupCredentialStore` to return `nil, nil` if `storeId` is a soft deleted credential store which is the same behavior as an unknown `storeId`
- [x] Change `ListCredentialStores` to not return any soft deleted credential stores
- [x] Change `CreateCredentialLibrary` to return an error if `storeId` is a soft deleted credential store


internal/db/schema/migrations/postgres/12/03_vault_credential.up.sql
```diff
@@ -55,26 +57,90 @@ begin;
...
+      if old.delete_time is null then
+
+        -- mark current and maintaining tokens as revoke
+        update credential_vault_token
+           set status   = 'revoke'
+         where store_id = new.public_id
+           and status in ('current', 'maintaining');
+
+        -- delete the store's libraries
+        delete
```
louisruch, line 109: This will by-pass the cred lib delete oplog

louisruch (commented):
Ill take a proper look at the test tomorrow, the only question I had was around oplog on the cred library delete in the trigger

internal/db/schema/migrations/postgres/12/03_vault_credential.up.sql
mgaffney: I think that should be fine since the same trigger will execute when the oplog is replayed.

louisruch (approved):
LGTM apart for the merge conflict

internal/db/schema/migrations/postgres/12/03_vault_credential.up.sql
louisruch: Aha yes great point!

talanknight (approved):

PR #1330, opened 2021-06-17
internal/servers/worker: Controller failure connection cleanup

vancluever:


vancluever:
Closing #1257 in favor of this one so that I can start merging other controller-related changes into this one to take advantage of some of the acceptance testing bits, and we can make further enhancements against progress here.

internal/servers/worker/status.go
```diff
@@ -115,6 +159,33 @@ func (w *Worker) sendWorkerStatus(cancelCtx context.Context) {
...
+		// Check for last successful status. Ignore nil last status, this probably
+		// means that we've never connected to a controller, and as such probably
+		// don't have any sessions to worry about anyway.
+		//
+		// If a length of time has passed since we've been able to communicate, we
+		// want to start terminating all sessions as a "break glass" kind of
+		// scenario, as there will be no way we can really tell if these
+		// connections should continue to exist.
+
+		lastStatus := w.LastStatusSuccess()
```
jefferai, line 171: I think it's important to capture something more for when lastStatus is nil. It seems _unlikely_ that we'd never have sent status but have connections, but I think what we could do is on worker startup time capture the current time. If lastStatus is nil, use that instead as the comparison for the grace period.

internal/servers/worker/status.go
```diff
@@ -115,6 +159,33 @@ func (w *Worker) sendWorkerStatus(cancelCtx context.Context) {
...
+		gracePeriod := w.statusGracePeriod()
+		if lastStatus != nil && lastStatus.StatusTime.Before(time.Now().Add(gracePeriod*-1)) {
```
jefferai, line 173: Pedantry: the Before/Add stuff could probably be simplified a bit by using `time.Sub`

go.mod
```diff
@@ -55,6 +55,7 @@ require (
 	github.com/posener/complete v1.2.3
 	github.com/spf13/cobra v1.1.1 // indirect
 	github.com/stretchr/testify v1.7.0
+	github.com/vancluever/dawdle v0.3.0
```
jefferai, line 58: This does need to move into the HC org prior to merging.

internal/servers/worker/status.go
vancluever: Apparently I thought this didn't exist, because I swear I looked for it 🤦‍♂️ Will fix.

vancluever: Fixed!

internal/servers/worker/status.go
vancluever: Fixed!

internal/servers/worker/worker.go
```diff
@@ -28,6 +28,7 @@ type Worker struct {
 	started     *ua.Bool
 
 	controllerStatusConn *atomic.Value
+	workerStartTime      *atomic.Value
```
jefferai, line 31: This doesn't need to be an atomic value since it will never change -- it will be set on startup and only read thereafter. It can be a plain `time.Time`.

vancluever: Fixed!

internal/servers/worker/status.go
```diff
@@ -154,46 +225,98 @@ func (w *Worker) sendWorkerStatus(cancelCtx context.Context) {
...
+			pbs.SESSIONSTATUS_SESSIONSTATUS_CANCELING,
+			pbs.SESSIONSTATUS_SESSIONSTATUS_TERMINATED,
+		),
+		cleanupConnectionsConditionExpired,
+	)
+}
+
+type cleanupConnectionsCondition func(si *sessionInfo) bool
+
+func cleanupConnectionsConditionAnyStatus(statuses ...services.SESSIONSTATUS) cleanupConnectionsCondition {
```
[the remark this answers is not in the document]
vancluever: Fixed!

go.mod
vancluever: Done!

vancluever:
Addressed all current review comments, and as discussed, removed (for now) the unit testing done in `internal/servers/worker`. These tests may come back at a later time if they're deemed to be particularly useful outside of what we have in E2E tests and `TestWorker` already.

vancluever:
@jefferai this should be good to go now.

internal/tests/cluster/session_cleanup_test.go
```diff
@@ -0,0 +1,509 @@
+package cluster
```
jefferai, line 1: Overall (maybe nit) on the file: Convention for us is not to pass require objects around but rather testing.T objects. Partly it's due to Go convention, partly because the only option with require is to fail immediately -- it precludes you from deciding you want to assert without failing, and partly it's because you can't call `t.Log` and friends since you don't have access to the original testing.T.

jefferai (approved):
Approved pending addressing my one comment

PR #1331, opened 2021-06-17
Vault job credential and credential store cleanup

louisruch:
### What does this PR do

- Allows library_id and session_id to be null on a credential
- On delete set null on both library_id and session_id instead of cascade
- Adds a trigger to set status to revoke when moving to null session_id
- Adds job to delete cred stores that are soft deleted and have no active tokens
- Adds job to delete credentials that have been soft deleted and are not active

### DIFF
--- a/internal/credential/vault/jobs_test.go
+++ b/internal/credential/vault/jobs_test.go
@@ -338,8 +338,22 @@ func TestTokenRenewalJob_Run(t *testing.T) {
 	_, prj := iam.TestScopes(t, iam.TestRepo(t, conn, wrapper))
 	v := NewTestVaultServer(t)
 
-	_, ct := v.CreateToken(t)
-	in, err := NewCredentialStore(prj.GetPublicId(), v.Addr, []byte(ct))
+	// Create 24 hour token
+	req := &vault.TokenCreateRequest{
+		DisplayName: t.Name(),
+		NoParent:    true,
+		Period:      "24h",
+		Policies:    []string{"default", "boundary-controller"},
+	}
+	vc := v.client(t).cl
+	secret, err := vc.Auth().Token().Create(req)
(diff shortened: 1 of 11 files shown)


louisruch:
@mgaffney PTAL

internal/credential/vault/jobs_test.go
```diff
@@ -338,8 +338,22 @@ func TestTokenRenewalJob_Run(t *testing.T) {
...
+		Period:      "24h",
+		Policies:    []string{"default", "boundary-controller"},
+	}
+	vc := v.client(t).cl
+	secret, err := vc.Auth().Token().Create(req)
+	require.NoError(err)
+	require.NotNil(secret)
+	token, err := secret.TokenID()
+	require.NoError(err)
+	require.NotNil(token)
```
mgaffney, line 354: Hopefully #1332 can help with this

louisruch: This was perfect!

internal/db/schema/migrations/postgres/12/03_vault_credential.up.sql
```diff
@@ -443,6 +443,36 @@ begin;
   create trigger update_credential_status_column before update on credential_vault_credential
     for each row execute procedure update_credential_status_column();
 
+  -- not_null_columns() will make the column names not null which are passed as
+  -- parameters when the trigger is created. It raises error code 23502 which is a
+  -- class 23 integrity constraint violation: not null column
+  create function not_null_columns()
```
mgaffney, line 449: Can you add a test to `internal/db/domains_test.go` for this?

mgaffney (changes requested):
This looks great! Just needs one thing.

mgaffney (approved):
Looks great!

PR #1332, opened 2021-06-18
Add ability to set the token period in the vault test harness

mgaffney:
This PR would also [help cleanup one of the tests](https://github.com/hashicorp/boundary/pull/1331/files?file-filters%5B%5D=.go#diff-f3d69b2f4dd5a826401c6992cc8f273ec6806f3811778d5eff8bff3f993e5a72R341-R356) in #1331.


louisruch (approved):
Thanks this will work in a couple of tests for me!!

PR #1333, opened 2021-06-18
Allow where without args in searchWhere

louisruch:
### What does this PR do

Updates `SearchWhere` to allow a where clause with no provided args

### DIFF
--- a/internal/db/read_writer.go
+++ b/internal/db/read_writer.go
@@ -1014,9 +1014,12 @@ func (rw *Db) LookupWhere(_ context.Context, resource interface{}, where string,
 }
 
 // SearchWhere will search for all the resources it can find using a where
-// clause with parameters.  Supports the WithLimit option.  If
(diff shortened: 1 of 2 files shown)


louisruch:
@mgaffney I raised this with @jimlambrt your question and decided we should allow where without args PTAL

mgaffney (approved):

internal/db/read_writer.go
```diff
@@ -1023,6 +1023,9 @@ func (rw *Db) SearchWhere(_ context.Context, resources interface{}, where string
 	if rw.underlying == nil {
 		return errors.New(errors.InvalidParameter, op, "missing underlying db")
 	}
+	if where == "" && len(args) > 0 {
```
mgaffney, line 1026: Can you update the method doc?

jimlambrt (approved):

PR #1334, opened 2021-06-19
Validate current vault token on addr update

louisruch:
### What does this PR do

Validates current token when there is an address change
Validates client cert / key is valid on change
Adds WithClientKey TestOption, this option was needed to allow the test where we change a cert but not a key, since this is validated now it was failing.


louisruch:
@mgaffney @talanknight PTAL I ended up adding a `WithClientKey` test option because the test where updating the cert without updating the key was failing now that I create the client each time we call `UpdateCredentialStore`.  We could move the client creation back into the `if` but I felt creating a client to validate the cert/key pair was probably a good idea

mgaffney (approved):

talanknight (approved):
Awesome. Thanks for making this change!

PR #1338, opened 2021-06-21
ASD cmds for Target and returning credentials to AuthSession request

talanknight:
I've manually tested the CLI add/set/remove operations for targets as well as the auth session request.

The Secret returned by the API is now a base64 encoded string.
The Target resource now properly includes the add/set/remove credential library actions in its authorized action set.
Connect commands have not been updated in this PR.

Sample response:

```
% boundary targets authorize-session -id ttcp_ngiS0tSibN

Target information:
  Authorization Token:
  CQqaoLzmN8NNiwraJ6iPgt785t1uK1HMB1xAnQgixNAjwPFYxVcnNh5CiwA5XsWC3FKn36Tnrp5WDCPMc6HWS3iSuEQVzp9v4rhkwu3KTRLx2VQ5Jnt7J18qhxMLGPZmds5dcq7rYwR3NQqkrf57ZCY2UE4GYWnTpJnaPi2YnKUyspJ7kkQRVeeKiifNLoRzefj3Uo5PrBjinmQRR7bUsMJ3UaiM2SUTQGuDuMwh61L4R47HaD8RfM5XULH7R3QK9ejCfJzYFjUQC13R6Es14BcLzf2s2aCvdPdVMyPCdyLyzBUhJXhq6e8S9odUykXcsGvsx6ZkT2bAhxStXF8HojYGnpK59kY4M3x215Mvnq8DKvstaoqPENaZjydGyNDChXDZja8X2nMkSsiDndB2C5rkQS8FVNCnRoaeHwcnnnvtR6yDpNETuun69gkn2EAMJfdVfhhzRxLhwQgwKmugtYNAALYY752VwSFBDgLU4ZKdasyvzHnwJXxEbJEPXjtmbqsCh47CJyJFdqiRuetDKmLpSWCZzkrG65SSjUSP9iKx9gSpmfUNAz21ywCU5prYtywVZtKzQMnR6Vmwum2Afr3URCAutcVJGfUHRt8mYBY3m4zgpdj4ZrMKSh3opCVNfNaWiWmscACMmV6wZSMH4tixmh4zEdZJCwssfZ9AwheBmzx2GKa9Z4TwNg4bZ9koRZEN31J4bzZuiaypXzH6ae69WKCj7toKPq5hgSiz5aAf5QAzBAvac9oJPofASqoVi2KS9SV4yAHQo5Lrh3dshztAWZ39uT2ZzsJSwFxidhVmCSoRYMXoPPfPg9R6CUv1HG4AkViH
  Created Time:          Tue, 22 Jun 2021 13:21:29 EDT
  Endpoint:              tcp://localhost:22
  Host ID:               hst_ntVTxM7asT
  Scope ID:              p_pW0EqdliF3
  Session ID:            s_l935Z3nIzT
  Target ID:             ttcp_ngiS0tSibN
  Type:                  tcp
  User ID:               u_nEydvLPVag

  Credentials:
    Credential Store ID:            csvlt_E1V7XW8YHx
    Credential Library ID:          clvlt_56qyz3Za9D
    Credential Library Type:        vault
    Secret:
      {
      "certificate": "-----BEGIN
      CERTIFICATE-----\nMIIDTzCCAjegAwIBA<snip>",
      "expiration": 1624425690,
      "issuing_ca": "-----BEGIN
      CERTIFICATE-----\nMIIDTzCCAjegAwIBAg<snip>",
      "private_key": "-----BEGIN RSA PRIVATE
      KEY-----\nMIIEpQIBAA<snip>",
      "private_key_type": "rsa",
      "serial_number": "4a:9b:60:74:d0:98:a3:b0:84:7d:d7:83:30:18:03:38:35:d8:e2:9e"
      }

```


internal/proto/local/controller/api/resources/targets/v1/target.proto
```diff
@@ -43,8 +43,8 @@ message SessionCredential {
 	// Output only. The library which generated this credential.
 	CredentialLibrary credential_library = 1;
 
-	// Output only. The secret of this credential.
-	google.protobuf.Value secret = 2;
+	// Output only. The secret of this credential base64 encoded.
+	string secret = 2;
```
jefferai, line 47: Why not make this `bytes`? The result of nearly any marshaling function will be `[]byte` and then we get the base64 encoding for free, and on the other end we get the decoding for free.

jefferai: In my branch working on some enhancements I've made this change; see what you think when I PR.

jefferai: OK, I've come around to this; having it as a string means more easily replacing it on the client side without then having it auto marshaled to base64.

api/targets/credential_utils.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"strings"
+)
+
+// DecodeJsonSecret converts base64 encoded json strings into a go map.
+// Numerical values in the map are of type json.Number.
+func DecodeJsonSecret(in string) (map[string]interface{}, error) {
```
mgaffney, line 11: Can you add a test for this?

internal/cmd/commands/targetscmd/funcs.go
```diff
@@ -136,6 +155,42 @@ func (c *Command) extraHelpFunc(helpMap map[string]func() string) string {
...
 		})
+	case "add-credential-libraries":
+		helpStr = base.WrapForHelpText([]string{
+			"Usage: boundary target add-credential-libraries [options] [args]",
+			"",
+			"  This command allows adding credential-library resources to target resources. Example:",
+			"",
+			"    Add credential-library resources to a tcp-type target:",
+			"",
+			`      $ boundary targets add-credential-libraries -id ttcp_1234567890 -credential-library clvl_1234567890 -credential-library clvlt_0987654321`,
```
mgaffney, line 166:
```suggestion
			`      $ boundary targets add-credential-libraries -id ttcp_1234567890 -credential-library clvlt_1234567890 -credential-library clvlt_0987654321`,
```

api/targets/credential_utils.go
jefferai, line 11: This is only used in the `internal` package right now. Can you move it somewhere in there, or copy the code there, so that isn't an exported/public part of the API since the API doesn't need it yet (and maybe ever)?

jefferai (changes requested):
Approved pending my one request

api/targets/credential_utils.go
talanknight: moved it in with the one test that uses it.

jefferai:
Note: The test that failed is likely due to Circle CI having issues, which has been a theme today. I've validated that test manually, and am merging as per Todd's approval on Slack.

PR #1340, opened 2021-06-22
internal/servers/controller: Worker failure connection cleanup

vancluever:
This commit adds the support to do the following:

* Mark connections for non-reporting workers as closed. This is the
controller counterpart to the worker functionality (see #1330). This is
written as a scheduled job that does most of the work DB-side, save some
rudimentary checking of individual workers' last update times.

* Works to reconcile states if such a broken controller-worker
connection resumes and a worker reports a connection as connected that
should be disconnected. In this case, the controller will send an update
request, and the worker will honor it and terminate the connection.

* Further refinement of the grace period setting has been added here.
We have converged on the current server "liveness" setting as our
default here, which is half of the previous 30s (15 seconds, in other
words). Additionally, this is now configurable on the controller and
worker side, with the caveat that it's currently impossible to do so in
config as the setting has been untagged in HCL. This is exposed so that
we can run some sophisticated testing scenarios where we skew the grace
period to either the controller or worker to ensure the aforementioned
reconciliation works.

* Some repository functions have been added to support the new
functionality.

### DIFF
--- a/internal/cmd/base/servers.go
+++ b/internal/cmd/base/servers.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/armon/go-metrics"
 	berrors "github.com/hashicorp/boundary/internal/errors"
@@ -24,6 +25,7 @@ import (
 	"github.com/hashicorp/boundary/internal/db"
 	"github.com/hashicorp/boundary/internal/kms"
 	"github.com/hashicorp/boundary/internal/observability/event"
+	"github.com/hashicorp/boundary/internal/servers"
 	"github.com/hashicorp/boundary/internal/types/scope"
 	"github.com/hashicorp/boundary/version"
 	"github.com/hashicorp/go-hclog"
@@ -38,6 +40,23 @@ import (
 	"github.com/jinzhu/gorm"
 	"github.com/mitchellh/cli"
 	"google.golang.org/grpc/grpclog"
+)
+
+const (
+	// defaultStatusGracePeriod is the default status grace period, or the period
+	// of time that we will go without a status report before we start
+	// disconnecting and marking connections as closed. This is tied to the
+	// server default liveness setting, a related value. See the servers package
+	// for more details.
+	defaultStatusGracePeriod = servers.DefaultLiveness
+
+	// statusGracePeriodEnvVar is the environment variable that can be used to
+	// configure the status grace period. This setting is provided in seconds,
+	// and can never be lower than the default status grace period defined above.
+	//
+	// TODO: This value is temporary, it will be removed once we have a better
+	// story/direction on attributes and system defaults.
+	statusGracePeriodEnvVar = "BOUNDARY_STATUS_GRACE_PERIOD"
 )
 
 type Server struct {
@@ -101,6 +120,11 @@ type Server struct {
 	DevDatabaseCleanupFunc     func() error
 
 	Database *gorm.DB
+
+	// StatusGracePeriodDuration represents the period of time (as a
+	// duration) that the controller will wait before marking
+	// connections from a disconnected worker as invalid.
+	StatusGracePeriodDuration time.Duration
 }
 
 func NewServer(cmd *Command) *Server {
@@ -674,4 +698,52 @@ func MakeSighupCh() chan struct{} {
 		}
 	}()
 	return resultCh
+}
+
+// SetStatusGracePeriodDuration sets the value for
+// StatusGracePeriodDuration.
+//
+// The grace period is the length of time we allow connections to run
+// on a worker in the event of an error sending status updates. The
+// period is defined the length of time since the last successful
+// update.
+//
+// The setting is derived from one of the following, in order:
+//
+//   * Via the supplied value if non-zero.
+//   * BOUNDARY_STATUS_GRACE_PERIOD, if defined, can be set to an
+//   integer value to define the setting.
+//   * If either of these is missing, the default is used. See the
+//   defaultStatusGracePeriod value for the default value.
+//
+// The minimum setting for this value is the default setting. Values
+// below this will be reset to the default.
+func (s *Server) SetStatusGracePeriodDuration(value time.Duration) {
+	var result time.Duration
+	switch {
+	case value > 0:
+		result = value
+	case os.Getenv(statusGracePeriodEnvVar) != "":
+		// TODO: See the description of the constant for more details on
+		// this env var
+		v := os.Getenv(statusGracePeriodEnvVar)
+		n, err := strconv.Atoi(v)
+		if err != nil {
+			s.Logger.Error(fmt.Sprintf("could not read setting for %s", statusGracePeriodEnvVar),
+				"err", err,
+				"value", v,
+			)
+			break
+		}
+
+		result = time.Second * time.Duration(n)
+	}
+
+	if result < defaultStatusGracePeriod {
+		s.Logger.Debug("invalid grace period setting or none provided, using default", "value", result, "default", defaultStatusGracePeriod)
+		result = defaultStatusGracePeriod
+	}
+
+	s.Logger.Debug("session cleanup in effect, connections will be terminated if status reports cannot be made", "grace_period", result)
+	s.StatusGracePeriodDuration = result
 }
--- a/internal/cmd/config/config.go
+++ b/internal/cmd/config/config.go
@@ -119,6 +119,13 @@ type Controller struct {
 	// denoted by time.Duration
 	AuthTokenTimeToStale         interface{} `hcl:"auth_token_time_to_stale"`
 	AuthTokenTimeToStaleDuration time.Duration
+
+	// StatusGracePeriod represents the period of time (as a duration) that the
+	// controller will wait before marking connections from a disconnected worker
+	// as invalid.
+	//
+	// TODO: This field is currently internal.
+	StatusGracePeriodDuration time.Duration `hcl:"-"`
 }
 
 type Worker struct {
@@ -132,6 +139,13 @@ type Worker struct {
 	// key=value syntax. This is trued up in the Parse function below.
 	TagsRaw interface{}         `hcl:"tags"`
 	Tags    map[string][]string `hcl:"-"`
+
+	// StatusGracePeriod represents the period of time (as a duration) that the
+	// worker will wait before disconnecting connections if it cannot make a
+	// status report to a controller.
+	//
+	// TODO: This field is currently internal.
+	StatusGracePeriodDuration time.Duration `hcl:"-"`
 }
 
 type Database struct {
--- a/internal/servers/controller/controller.go
+++ b/internal/servers/controller/controller.go
@@ -38,7 +38,7 @@ type Controller struct {
 
 	workerAuthCache *cache.Cache
 
-	// Used for testing
+	// Used for testing and tracking worker health
 	workerStatusUpdateTimes *sync.Map
 
 	// Repo factory methods
@@ -117,7 +117,9 @@ func New(conf *Config) (*Controller, error) {
 	jobRepoFn := func() (*job.Repository, error) {
 		return job.NewRepository(dbase, dbase, c.kms)
 	}
-	c.scheduler, err = scheduler.New(c.conf.RawConfig.Controller.Name, jobRepoFn, c.logger)
+	// TODO: the RunJobsLimit is temporary until a better fix gets in. This
+	// currently caps the scheduler at running 10 jobs per interval.
+	c.scheduler, err = scheduler.New(c.conf.RawConfig.Controller.Name, jobRepoFn, c.logger, scheduler.WithRunJobsLimit(10))
 	if err != nil {
 		return nil, fmt.Errorf("error creating new scheduler: %w", err)
 	}
@@ -182,7 +184,29 @@ func (c *Controller) Start() error {
 
 func (c *Controller) registerJobs() error {
 	rw := db.New(c.conf.Database)
-	return vault.RegisterJobs(c.baseContext, c.scheduler, rw, rw, c.kms, c.logger)
+	if err := vault.RegisterJobs(c.baseContext, c.scheduler, rw, rw, c.kms, c.logger); err != nil {
+		return err
+	}
+
+	if err := c.registerSessionCleanupJob(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// registerSessionCleanupJob is a helper method to abstract
+// registering the session cleanup job specifically.
+func (c *Controller) registerSessionCleanupJob() error {
+	sessionCleanupJob, err := newSessionCleanupJob(c.logger, c.SessionRepoFn, int(c.conf.StatusGracePeriodDuration.Seconds()))
+	if err != nil {
+		return fmt.Errorf("error creating session cleanup job: %w", err)
+	}
+	if err = c.scheduler.RegisterJob(c.baseContext, sessionCleanupJob); err != nil {
+		return fmt.Errorf("error registering session cleanup job: %w", err)
+	}
+
+	return nil
 }
 
 func (c *Controller) Shutdown(serversOnly bool) error {
--- a/internal/servers/controller/handlers/workers/worker_service.go
+++ b/internal/servers/controller/handlers/workers/worker_service.go
@@ -76,12 +76,19 @@ func (ws *workerServiceServer) Status(ctx context.Context, req *pbs.StatusReques
 		Controllers: controllers,
 	}
 
-	// TODO (jeff, 05-2021): We should possibly list all connections here with
-	// their statuses, and check those against the found connections below. If
-	// any we think should be closed are marked open by the worker, the worker
-	// should be told to close them.
+	var (
+		// For tracking the reported open connections.
+		reportedOpenConns []string
+		// For tracking the session IDs we've already requested
+		// cancellation for. We won't need to add connection cancel
+		// requests for these because canceling the session terminates the
+		// connections.
+		requestedSessionCancelIds []string
+	)
 
-	var foundConns []string
+	// This is a map of all sessions and their statuses. We keep track of
+	// this for easy lookup if we need to make change requests.
+	sessionStatuses := make(map[string]pbs.SESSIONSTATUS)
 
 	for _, jobStatus := range req.GetJobs() {
 		switch jobStatus.Job.GetType() {
@@ -91,6 +98,9 @@ func (ws *workerServiceServer) Status(ctx context.Context, req *pbs.StatusReques
 			if si == nil {
 				return nil, status.Error(codes.Internal, "Error getting session info at status time")
 			}
+
+			// Record status.
+			sessionStatuses[si.GetSessionId()] = si.Status
 
 			// Check connections before potentially bypassing the rest of the
 			// logic in the switch on si.Status.
@@ -103,7 +113,7 @@ func (ws *workerServiceServer) Status(ctx context.Context, req *pbs.StatusReques
 					// report as found, so that we should attempt to close it.
 					// Note that unspecified is the default state for the enum
 					// but it's not ever explicitly set by us.
-					foundConns = append(foundConns, conn.GetConnectionId())
+					reportedOpenConns = append(reportedOpenConns, conn.GetConnectionId())
 				}
 			}
 
@@ -127,7 +137,7 @@ func (ws *workerServiceServer) Status(ctx context.Context, req *pbs.StatusReques
 			}
 			// If the session from the DB is in canceling status, and we're
 			// here, it means the job is in pending or active; cancel it. If
-			// it's in termianted status something went wrong and we're
+			// it's in terminated status something went wrong and we're
 			// mismatched, so ensure we cancel it also.
 			currState := sessionInfo.States[0].Status
 			if currState.ProtoVal() != si.Status {
@@ -148,13 +158,53 @@ func (ws *workerServiceServer) Status(ctx context.Context, req *pbs.StatusReques
 						},
 						RequestType: pbs.CHANGETYPE_CHANGETYPE_UPDATE_STATE,
 					})
+					// Log the session ID so we don't add a duplicate change
+					// request on connection normalization.
+					requestedSessionCancelIds = append(requestedSessionCancelIds, sessionId)
 				}
 			}
 		}
 	}
 
-	// Run our connection cleanup function
-	closedConns, err := sessRepo.CloseDeadConnectionsOnWorkerReport(ctx, req.Worker.PrivateId, foundConns)
+	// Normalize the current state of connections on the worker side
+	// with the data from the controller. In other words, if one of our
+	// found connections isn't supposed to be alive still, kill it.
+	//
+	// This is separate from the above session normalization and is
+	// additive to it, we don't add sessions that have already been
+	// added there as canceling sessions already closes the
+	// connections.
+	shouldCloseConnections, err := sessRepo.ShouldCloseConnectionsOnWorker(ctx, reportedOpenConns, requestedSessionCancelIds)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Error fetching connections that should be closed: %v", err)
+	}
+
+	for sessionId, connIds := range shouldCloseConnections {
+		var connChanges []*pbs.Connection
+		for _, connId := range connIds {
+			connChanges = append(connChanges, &pbs.Connection{
+				ConnectionId: connId,
+				Status:       session.StatusClosed.ProtoVal(),
+			})
+		}
+
+		ret.JobsRequests = append(ret.JobsRequests, &pbs.JobChangeRequest{
+			Job: &pbs.Job{
+				Type: pbs.JOBTYPE_JOBTYPE_SESSION,
+				JobInfo: &pbs.Job_SessionInfo{
+					SessionInfo: &pbs.SessionJobInfo{
+						SessionId:   sessionId,
+						Status:      sessionStatuses[sessionId],
+						Connections: connChanges,
+					},
+				},
+			},
+			RequestType: pbs.CHANGETYPE_CHANGETYPE_UPDATE_STATE,
+		})
+	}
+
+	// Run our controller-side cleanup function.
+	closedConns, err := sessRepo.CloseDeadConnectionsForWorker(ctx, req.Worker.PrivateId, reportedOpenConns)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Error closing dead conns for worker %s: %v", req.Worker.PrivateId, err)
 	}
--- a/internal/servers/controller/session_cleanup_job.go
+++ b/internal/servers/controller/session_cleanup_job.go
@@ -1,0 +1,138 @@
+package controller
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/scheduler"
+	"github.com/hashicorp/boundary/internal/servers/controller/common"
+	"github.com/hashicorp/boundary/internal/session"
+	"github.com/hashicorp/go-hclog"
+)
+
+// sessionCleanupJob defines a periodic job that monitors workers for
+// loss of connection and terminates connections on workers that have
+// not sent a heartbeat in a significant period of time.
+//
+// Relevant connections are simply marked as disconnected in the
+// database. Connections will be independently terminated by the
+// worker, or the event of a synchronization issue between the two,
+// the controller will win out and order that the connections be
+// closed on the worker.
+type sessionCleanupJob struct {
+	logger        hclog.Logger
+	serversRepoFn common.ServersRepoFactory
+	sessionRepoFn common.SessionRepoFactory
+
+	// The amount of time to give disconnected workers before marking
+	// their connections as closed.
+	gracePeriod int
+
+	// Whether or not the operation has completed. This is used to dictate the
+	// result to the Status call for this job as the call is atomic.
+	completed bool
+}
+
+// newSessionCleanupJob instantiates the session cleanup job.
+func newSessionCleanupJob(
+	logger hclog.Logger,
+	sessionRepoFn common.SessionRepoFactory,
+	gracePeriod int,
+) (*sessionCleanupJob, error) {
+	const op = "controller.newNewSessionCleanupJob"
+	switch {
+	case logger == nil:
+		return nil, errors.New(errors.InvalidParameter, op, "missing logger")
+	case sessionRepoFn == nil:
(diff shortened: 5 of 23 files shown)


internal/cmd/config/config.go
```diff
@@ -115,6 +115,13 @@ type Controller struct {
 	// denoted by time.Duration
 	AuthTokenTimeToStale         interface{} `hcl:"auth_token_time_to_stale"`
 	AuthTokenTimeToStaleDuration time.Duration
+
+	// StatusGracePeriod represents the period of time (as a duration) that the
+	// controller will wait before marking connections from a disconnected worker
+	// as invalid.
+	//
+	// TODO: This field is currently internal.
+	StatusGracePeriodDuration time.Duration `hcl:"-"`
```
jefferai, line 124: I don't think we'll ever want this value exposed via config, so rather than put it in here, can you simple set it as a field on the `base.Server` and then it can be toggled in tests from there?

internal/servers/controller/controller.go
```diff
@@ -27,6 +30,11 @@ import (
 	ua "go.uber.org/atomic"
 )
 
+const (
+	defaultStatusGracePeriod = 15 * time.Second
```
jefferai, line 34: Why did this get changed to 15 from 30? Seems very short.

internal/servers/controller/controller.go
```diff
@@ -151,6 +164,52 @@ func New(conf *Config) (*Controller, error) {
...
+//
+//   * Through configuration,
+//   * BOUNDARY_STATUS_GRACE_PERIOD, if defined, can be set to an
+//   integer value to define the setting.
+//   * If either of these is missing, the default (15 seconds) is
+//   used.
+//
+// The minimum setting for this value is the default setting. Values
+// below this will be reset to the default.
+func (c *Controller) setStatusGracePeriod() {
```
jefferai, line 185: Rather than copy this code between controllers and workers, please make it a function of `base.Server` that checks the env var and uses that or uses the default (similar to how we set up listeners, controllers, etc.), and sets it in the Server struct. Since both worker and controller have that in their local `Config` object they can share the logic.

internal/servers/controller/handlers/workers/worker_service.go
```diff
@@ -76,13 +76,27 @@ func (ws *workerServiceServer) Status(ctx context.Context, req *pbs.StatusReques
...
+	// reconciliation needs to happen.
+	workerConnectionStatuses := make(map[string]session.ConnectionStatus)
+	workerConnections, err := sessRepo.ListConnectionsByServerId(ctx, req.Worker.PrivateId)
+	if err != nil {
+		return &pbs.StatusResponse{}, status.Errorf(codes.Internal, "Error listing connections for worker: %v", err)
+	}
 
-	var foundConns []string
+	for _, conn := range workerConnections {
+		_, states, err := sessRepo.LookupConnection(ctx, conn.PublicId, nil)
```
jefferai, line 89: This could potentially be thousands or tens of thousands of DB calls. I think we need a different way to handle this. One might be to have an analogue of LookupSessions that allows a WithSessionIds but instead WithConnectionIds, so we just give that list as part of a single query. Another might be to return the current status for each connection along with the session in the list. Both have pros and cons and probably worth a deeper discussion with the DB experts.

jefferai: Or possibly the current state could be returned as part of ListConnectionsByServerId.

internal/servers/controller/controller.go
vancluever: This is to put it in line with the server liveness timeout (seen [here](https://github.com/hashicorp/boundary/blob/33929c27e956385fd3fe72b8f672eb67a9701ffc/internal/servers/repository.go#L16)). This is something I wanted to investigate after finishing the work here. Definitely flexible on changing it, but it made sense to me that if a server was not showing up as "alive" on things like `ListServers`, then it'd probably make sense to start disconnecting things from it at around the same time.

internal/servers/controller/handlers/workers/worker_service.go
vancluever: :+1: totally fair. Will look at a different way for sure!

vancluever:
@jefferai feedback all makes sense, will work, on applying it tomorrow!

vancluever:
@jefferai this is ready for review again. I'll check in a bit to see if tests passed. 🤞

internal/cmd/config/config.go
jefferai: I think this can now be removed here and in the Worker block right?

internal/servers/controller/testing.go
```diff
@@ -411,6 +416,9 @@ func NewTestController(t *testing.T, opts *TestControllerOpts) *TestController {
 		opts.Config.Controller.Name = opts.Name
 	}
 
+	if opts.StatusGracePeriodDuration > 0 {
+		opts.Config.Controller.StatusGracePeriodDuration = opts.StatusGracePeriodDuration
```
jefferai, line 420: I think this also isn't needed anymore?

internal/servers/controller/handlers/workers/worker_service.go
```diff
@@ -148,13 +158,53 @@ func (ws *workerServiceServer) Status(ctx context.Context, req *pbs.StatusReques
...
+		}
+
+		ret.JobsRequests = append(ret.JobsRequests, &pbs.JobChangeRequest{
+			Job: &pbs.Job{
+				Type: pbs.JOBTYPE_JOBTYPE_SESSION,
+				JobInfo: &pbs.Job_SessionInfo{
+					SessionInfo: &pbs.SessionJobInfo{
+						SessionId:   sessionId,
+						Status:      sessionStatuses[sessionId],
+						Connections: connChanges,
```
jefferai, line 198: If I'm right, for a session that isn't canceled but where individual connections should be canceled, this won't yet work as desired on the worker. The worker ranges through sessions and only takes effect if the (updated) session state is canceling or terminated. I think it needs another case that checks individual connections if the session is still active.

jefferai:
Meta comment -- in the future please don't use branch names with slashes -- our standard is a hyphen instead. This is due to https://github.com/golang/go/issues/38861 (and previous issue https://github.com/golang/go/issues/36902). It's not obvious that this should ever be a problem, and yet... 😀

internal/servers/worker/testing.go
```diff
@@ -147,6 +152,10 @@ func NewTestWorker(t *testing.T, opts *TestWorkerOpts) *TestWorker {
 		opts.Config.Worker.Name = opts.Name
 	}
 
+	if opts.StatusGracePeriodDuration > 0 {
+		opts.Config.Worker.StatusGracePeriodDuration = opts.StatusGracePeriodDuration
```
jefferai, line 156: Same comment, I think this should be using the server var now?

internal/session/query.go
```diff
@@ -278,5 +278,35 @@ with
...
+with
+  -- Find connections that are closed so we can reference those IDs
+  closed_connections as (
+    select connection_id
+      from session_connection_state
+    where
+      -- It's the current state
+      end_time is null
+        and
+      -- Current state isn't closed state
```
jefferai, line 300: I think this comment is describing the opposite of the below check

internal/servers/controller/handlers/workers/worker_service.go
vancluever: This is addressed and is actually addressed by the skew cases where we skew in favor of the controller versus the worker. I want to link the work now but will wait until after addressing the rest of the review feedback to ensure that the link is stable. 😛

vancluever: OK, I thought I added this but apparently it's not there. Taking a look now, also need to figure out why the tests are passing if this is indeed the case. The cases where the controller is skewed for should definitely close connections on the worker first, and then cut off the connection when the next status request is made and a change request goes through. Will review this and add the code.

vancluever:
Adding @louisruch to review the job code.

internal/servers/controller/controller.go
```diff
@@ -178,6 +178,24 @@ func (c *Controller) Start() error {
 }
 
 func (c *Controller) registerJobs() error {
+	if err := c.registerSessionCleanupJob(); err != nil {
```
louisruch, line 181: Might be good to merge `main` into `vancluever/session-cleanup` and then `vancluever/session-cleanup` into this branch as main now makes use of this method.

internal/servers/controller/session_cleanup_job.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// NextRunIn returns the next run time after a job is completed.
+//
+// The next run time is defined for sessionCleanupJob as one second.
+// This is because the job should run continuously to terminate
+// connections as soon as a worker has not reported in for a long
+// enough time. Only one job will ever run at once, so there is no
+// reason why it cannot run again immediately.
+func (j *sessionCleanupJob) NextRunIn() (time.Duration, error) { return time.Second, nil }
```
louisruch, line 78: While I don't necessarily disagree with the comment a second does seem aggressive.  Regardless of that please note that each scheduler only requests work once a minute, so you will only truly reach continuously running at scale when multiple controllers are running schedulers.

internal/servers/controller/session_cleanup_job.go
```diff
@@ -0,0 +1,191 @@
...
+				"num_processed", j.numProcessed,
+				"num_workers", j.numWorkers,
+			)
+			continue
+		}
+
+		// If the guard failed, that means that our worker has not
+		// reported within the acceptable grace period. Find all active
+		// connections for the worker and close them.
+		numClosed, err := sessionRepo.CloseDeadConnectionsForWorker(ctx, worker.PrivateId, nil)
```
louisruch, line 148: I am wondering if it might be beneficial to move this call outside of the worker range, and then create a modified query that instead of calling:
```
-- Related to the worker that just reported to us
      server_id = $1
```
does a `server_id in` with a list of worker_ids that we collected in the range which were outside the grace period.

louisruch (commented):
Did a first pass, I did not look at tests at all yet.  I will have another look over the weekend once I had some down time - left a couple of comments in the mean time

internal/servers/controller/session_cleanup_job.go
vancluever: @louisruch are you referring to having a single query versus more than one? If so, @jefferai and I discussed this and concluded that there would not be enough workers for this to matter.

internal/servers/controller/session_cleanup_job.go
vancluever: Yeah, I discovered that after setting this up, hah.

One thing that I'm worried about is that if I set this, and we add a next run time of say, like a minute, that it will possibly be two minutes before the job runs again, as the next scheduled time may not come up before the next job run happens.

Is there any other/better way to guarantee that this job runs once per minute? The actual grace period is 15 seconds, so we want to make absolute sure that it runs as soon as possible to ensure connections are closed as soon as they can be.

internal/servers/controller/session_cleanup_job.go
louisruch: That is what I was suggesting.  Are we confident with the claim there won’t be enough workers to worry about?  This is something we can revisit in the future though, just felt like a relatively easy optimization.

vancluever: @jefferai thoughts? Do we have any data on what our upper limit on scale right now is for workers?

@louisruch I think that in terms of orders of magnitude, running a query for every _worker_ is probably much safer than running a query for, say, every _connection_. Probably within the acceptable range, but it would be good to know what our expect scale is for workers before saying 100% for sure.

internal/servers/controller/session_cleanup_job.go
vancluever: Discussed this with @louisruch offline and agreed on the fact that the 1 second interval should be good for now to ensure that the job always runs within a particular controller's minimum scheduler interval (1 minute). This will be increasingly variable as the number of controllers go up (and assuming that they were _not_ all started at around the same time), but it helps cover cases where there will be a small amount of controllers where the possibility of the actual next run time being a full minute is higher.

internal/servers/controller/session_cleanup_job.go
jefferai: Given that the database holds the last updated time, I wonder if we even need to calculate each worker's grace individually. Think about a supermarket where they have a sign like, if the birth date is past today in 1974, that person can buy booze. I wonder if we really can and ought to use the DB for this. Basically, just tell the DB, if the last status time is before now minus 15 seconds, close all connections. I think this could generate the list and close things in a single CTE in a single transaction.

internal/cmd/base/servers.go
```diff
@@ -606,3 +617,49 @@ func MakeSighupCh() chan struct{} {
...
+//
+// The minimum setting for this value is the default setting. Values
+// below this will be reset to the default.
+func (s *Server) SetStatusGracePeriodDuration(value time.Duration) {
+	var result time.Duration
+	switch {
+	case value > 0:
+		result = value
+	case os.Getenv(statusGracePeriodEnvVar) != "":
+		v := os.Getenv(statusGracePeriodEnvVar)
```
jimlambrt, line 645: I'm curious about the need/choice to set this via an envvar.  It's not a typical pattern in Boundary.  Is there some specific use-case that requires envvar support?

internal/servers/controller/session_cleanup_job.go
jimlambrt: @jefferai you're suggestion would seem to scale better and have the benefit of eliminating some race conditions as well.

vancluever: Alright, will move this to a single query. Incidentally, this was the original thought (when I was looking at table changes to make this happen), but it was scrapped when we decided that multiple queries was fine.

I'm sure we can do it without additional table changes though, so don't take that as me saying that we need to do that again.

internal/cmd/base/servers.go
vancluever: @jefferai ^^^

internal/servers/controller/controller.go
vancluever: Going to rebase this to main post-release actually, as that "long-running" branch is not being used anymore.

internal/cmd/base/servers.go
jefferai: @jimlambrt The issue here is that really we want this to be a database value so that it's not a per-controller configuration value that might be different on different controllers. So I'm hesitant to add it to config as a stop-gap because then we'd have to remove it later and force users to switch over.

What this really needs is the previously discussed attributes/defaults system -- the same thing that default worker behavior on targets is waiting on, as well as synchronized user preferences. The idea for the moment is it's an env var that is unpublished/undocumented that acts as a relief valve for anyone having issues with the implementation, but that officially we only have a static value right now.

jimlambrt: @jefferai and I chatted about this... @vancluever can we add a comment that makes it explicit that this is a temporary stopgap measure?  This will make it clear to everyone that the envvar is not a permanent solution and we  will definitely deprecated it in the future in favor of a holistic approach to attributes/defaults system.   ty!

internal/servers/controller/controller.go
vancluever: This was done last week (as discussed with @louisruch offline so that we could resolve some job throttling issues).

internal/cmd/base/servers.go
vancluever: Done!

internal/servers/controller/session_cleanup_job.go
vancluever: This is now done!

internal/servers/controller/handlers/workers/worker_service.go
vancluever: Resolving old fixed conversation!

internal/servers/worker/status.go
```diff
@@ -99,6 +59,34 @@ func (w *Worker) LastStatusSuccess() *LastStatusInformation {
...
+//
+// The timeout is aligned with the worker's status grace period. A nil error
+// means the status was sent successfully.
+func (w *Worker) WaitForNextSuccessfulStatusUpdate() error {
+	w.logger.Debug("waiting for next status report to controller")
+	waitStatusStart := time.Now()
+	ctx, cancel := context.WithTimeout(w.baseContext, w.conf.StatusGracePeriodDuration)
+	defer cancel()
+	for {
+		if err := ctx.Err(); err != nil {
```
jefferai, line 75: Another way to do this which feels _to me_ to be more canonical but very well may not be is to have a `select` statement on `ctx.Done()` and `time.After(time.Second)`. This will work, but may be worth looking into.

internal/servers/controller/session_cleanup_job.go
```diff
@@ -0,0 +1,138 @@
...
+//
+// The next run time is defined for sessionCleanupJob as one second.
+// This is because the job should run continuously to terminate
+// connections as soon as a worker has not reported in for a long
+// enough time. Only one job will ever run at once, so there is no
+// reason why it cannot run again immediately.
+func (j *sessionCleanupJob) NextRunIn() (time.Duration, error) { return time.Second, nil }
+
+// Status returns the status of the running job.
+func (j *sessionCleanupJob) Status() scheduler.JobStatus {
```
louisruch, line 80: Its fine if you want to keep it like this, but with the Vault Credential Cleanup work I set both `Completed` and `Total` to the number of affected rows in the single query https://github.com/hashicorp/boundary/blob/main/internal/credential/vault/jobs.go#L930.  
This gives us a possibly helpful future record (outside of logs) of how many creds we to cleanup in a job.  We could similarly set a field to `len(results)` in `Run`

internal/servers/repository.go
```diff
@@ -67,18 +75,27 @@ func (r *Repository) listServersWithReader(ctx context.Context, reader db.Reader
 	opts := getOpts(opt...)
 	liveness := opts.withLiveness
 	if liveness == 0 {
-		liveness = defaultLiveness
+		liveness = DefaultLiveness
 	}
+
+	var where string
+	if liveness > 0 {
```
louisruch, line 82: I am not clear how `liveness` will ever `<= 0`.  Doesn't the `if` statement above prevent this?  Unless of course we change the `DefaultLiveness` to `0` at some point but that should never happen anyway...

internal/servers/repository.go
```diff
@@ -67,18 +75,27 @@ func (r *Repository) listServersWithReader(ctx context.Context, reader db.Reader
...
+		where = fmt.Sprintf("type = $1 and update_time > now() - interval '%d seconds'", uint32(liveness.Seconds()))
```
louisruch, line 83: we have a new domain helper function Mike added for Vault that you could use here as:
`update_time > wt_add_seconds_to_now(?)` and then add `-1 * liveness.Seconds()` to the args.

internal/servers/worker/status.go
```diff
@@ -99,6 +59,37 @@ func (w *Worker) LastStatusSuccess() *LastStatusInformation {
...
+//
+// The timeout is aligned with the worker's status grace period. A nil error
+// means the status was sent successfully.
+func (w *Worker) WaitForNextSuccessfulStatusUpdate() error {
+	w.logger.Debug("waiting for next status report to controller")
+	waitStatusStart := time.Now()
+	ctx, cancel := context.WithTimeout(w.baseContext, w.conf.StatusGracePeriodDuration)
+	defer cancel()
+	for {
+		select {
```
louisruch, line 75: Seems to be a good use-case for using a [Cond](https://golang.org/pkg/sync/#Cond), where the worker can just broadcast every time it performs an update to the `lastStatusSuccess` instead of polling every second.  Ok keeping it like this as well if that causes a larger refactor that seems.

internal/servers/worker/testing.go
```diff
@@ -207,6 +211,10 @@ func NewTestWorker(t *testing.T, opts *TestWorkerOpts) *TestWorker {
 		opts.Config.Worker.Name = opts.Name
 	}
 
+	if opts.StatusGracePeriodDuration > 0 {
```
louisruch, line 214: Is this needed?  Doesn't `SetStatusGracePeriodDuration` below handle both the zero and non-zero case?

internal/session/repository_connection.go
```diff
@@ -148,6 +153,128 @@ func (r *Repository) CloseDeadConnectionsOnWorkerReport(ctx context.Context, ser
...
+	_, err := r.writer.DoTx(
+		ctx,
+		db.StdRetryCnt,
+		db.ExpBackoff{},
+		func(reader db.Reader, w db.Writer) error {
+			rows, err := w.Query(ctx, closeConnectionsForDeadServersCte, []interface{}{gracePeriod})
+			if err != nil {
+				return errors.Wrap(err, op)
+			}
+
```
louisruch, line 185: We should add `defer rows.Close()` here

internal/session/repository_connection.go
```diff
@@ -148,6 +153,128 @@ func (r *Repository) CloseDeadConnectionsOnWorkerReport(ctx context.Context, ser
...
+		const sessionIdFmtStr = `and session_id not in (%s)`
+		sessionsStr = fmt.Sprintf(sessionIdFmtStr, strings.Join(sessionsParams, ","))
+	}
+
+	rows, err := r.reader.Query(
+		ctx,
+		fmt.Sprintf(shouldCloseConnectionsCte, connsStr, sessionsStr),
+		args,
+	)
+
```
louisruch, line 260: We should add defer rows.Close() here

internal/session/query.go
```diff
@@ -278,5 +278,100 @@ with
...
+	// The query returns the set of servers that have had connections closed
+	// along with their last update time and the number of connections closed on
+	// each.
+	closeConnectionsForDeadServersCte = `
+with
+  -- Get dead servers, parameterized off of grace period in seconds
+  dead_servers as (
+    select private_id, update_time
+    from server
+    where update_time < now() - interval '1 second' * $1
```
louisruch, line 300: Could also use the `wt_add_seconds_to_now` function here, I feel it reads better but as you like.  Could also create a new `wt_subtract_seconds_from_now` if you want to avoid sending the arg with a `-1*`

louisruch (commented):
Couple of comments but this is really looking great!

internal/servers/worker/status.go
vancluever: Let's leave this like it is right now.  I think adjusting the status check inter-communication in such a way would be a bit more impacting than just adjusting things here, in addition to me not necessarily understanding sync.Cond well (and I haven't seen many good examples of its use). Could be something we entertain in the future.

internal/servers/worker/testing.go
vancluever: Yeah, it's one or the other. Probably a relic of maintenance, will fix either way.

vancluever:
@louisruch comments addressed!

vancluever:
Also just rebased off of main to fix conflicts.

internal/servers/repository.go
vancluever: Reverted this as we actually use a negative number here to disable the default liveness setting.

vancluever: See https://github.com/hashicorp/boundary/blob/ebc0b96b0b7dc989202f8fba255de79e251f2164/internal/servers/options.go#L40-L47

louisruch: ty

internal/servers/controller/testing.go
```diff
@@ -629,3 +638,63 @@ func (tc *TestController) AddClusterControllerMember(t *testing.T, opts *TestCon
...
+// come in. If it does not come in within the default status grace
+// period, this function returns an error.
+func (tc *TestController) WaitForNextWorkerStatusUpdate(workerId string) error {
+	tc.Logger().Debug("waiting for next status report from worker", "worker", workerId)
+	waitStatusStart := time.Now()
+	ctx, cancel := context.WithTimeout(tc.ctx, tc.b.StatusGracePeriodDuration)
+	defer cancel()
+	var err error
+	for {
+		if err = ctx.Err(); err != nil {
```
louisruch, line 652: Not sure if you wanted to implement the same suggestion Jeff gave re turning this into a for select

louisruch (dismissed):
Not sure if @jefferai wanted to review this again before merging but this LGTM

internal/servers/controller/testing.go
vancluever: Should be done now!

louisruch (approved):
LGTM

internal/servers/controller/testing.go
```diff
@@ -629,3 +638,71 @@ func (tc *TestController) AddClusterControllerMember(t *testing.T, opts *TestCon
 	}
 	return NewTestController(t, nextOpts)
 }
+
+// WaitForNextWorkerStatusUpdate waits for the next status check from a worker to
+// come in. If it does not come in within the default status grace
+// period, this function returns an error.
+func (tc *TestController) WaitForNextWorkerStatusUpdate(workerId string) error {
```
jefferai, line 645: Why are these two functions separate? One just calls the other -- why not just have a single function?

I looked and don't see it used anywhere else -- even tests only use the exported version.

internal/servers/controller/testing.go
```diff
@@ -629,3 +638,71 @@ func (tc *TestController) AddClusterControllerMember(t *testing.T, opts *TestCon
...
+		case <-time.After(time.Second):
+			// pass
+		}
+
+		var waitStatusCurrent time.Time
+		var err error
+		tc.Controller().WorkerStatusUpdateTimes().Range(func(k, v interface{}) bool {
+			if k == nil || v == nil {
+				err = fmt.Errorf("nil key or value on entry: key=%#v value=%#v", k, v)
+				return false
```
jefferai, line 675: Don't you want to return `true` in all of these cases? Finding a bad value is obviously a bad thing, but it doesn't mean that the value for the worker ID passed in is bad. You could have perfectly legitimate data for the worker ID we care about but be ignoring it by cutting out early on an error for some other value.

internal/servers/controller/testing.go
```diff
@@ -629,3 +638,71 @@ func (tc *TestController) AddClusterControllerMember(t *testing.T, opts *TestCon
...
+
+			if workerStatusUpdateId == workerId {
+				waitStatusCurrent = workerStatusUpdateTime
+				return false
+			}
+
+			return true
+		})
+
+		if err != nil {
```
jefferai, line 698: Similar question -- if the data for the worker we're interested in is good, why return an error here? It's good to surface that something is wrong, but not ideal to not perform the intended function if that data is fine.

internal/servers/worker/status.go
```diff
@@ -204,12 +196,25 @@ func (w *Worker) sendWorkerStatus(cancelCtx context.Context) {
...
 						continue
 					}
 					si := siRaw.(*sessionInfo)
 					si.Lock()
 					si.status = sessInfo.GetStatus()
+					// Update connection state if there are any connections in
+					// the request.
+					for _, conn := range sessInfo.GetConnections() {
+						connId := conn.GetConnectionId()
+						connInfo, ok := si.connInfoMap[conn.GetConnectionId()]
```
jefferai, line 209: This could be simplified to use `connId` from the line above.

internal/servers/worker/status.go
```diff
@@ -240,22 +245,31 @@ func (w *Worker) cleanupConnections(cancelCtx context.Context, ignoreSessionStat
...
 			}
+
+		default:
+			// Cancel connections *with* regard to individual connection
+			// state (ie: only ones that the controller has requested be
+			// terminated).
+			closedIds := w.cancelConnections(si.connInfoMap, false)
+			for _, connId := range closedIds {
+				closeInfo[connId] = si.id
+				w.logClose(si.id, connId)
```
jefferai, line 271: Given it's only two places and only a single line for the function, my personal preference would be to just duplicate the logging lines here and below. To me it's a bit less mental energy to just read the line here than have to remember or jump down to the function to see what it's doing.

jefferai: Take it or leave it, just my $0.02.

internal/session/query.go
```diff
@@ -278,5 +278,100 @@ with
...
+  -- Find connections that are not closed so we can reference those IDs
+  unclosed_connections as (
+    select connection_id
+      from session_connection_state
+    where
+      -- It's the current state
+      end_time is null
+        and
+      -- Current state isn't closed state
+      state in ('authorized', 'connected')
```
jefferai, line 311: Would it be better to do something like `state not in ('closed')`? That way if we add more states prior to `closed` then it still works.

internal/session/repository_connection.go
```diff
@@ -148,6 +153,129 @@ func (r *Repository) CloseDeadConnectionsOnWorkerReport(ctx context.Context, ser
...
+			errors.InvalidParameter, op, fmt.Sprintf("gracePeriod must be at least %d seconds", DeadWorkerConnCloseMinGrace))
+	}
+
+	results := make([]CloseConnectionsForDeadWorkersResult, 0)
+	_, err := r.writer.DoTx(
+		ctx,
+		db.StdRetryCnt,
+		db.ExpBackoff{},
+		func(reader db.Reader, w db.Writer) error {
+			rows, err := w.Query(ctx, closeConnectionsForDeadServersCte, []interface{}{gracePeriod})
```
jefferai, line 181: It'd be nicer to use ScanRows below -- it's not in the writer interface right now but it could be added to the interface with no further change to code as the underlying db object already has that function.

internal/servers/controller/testing.go
vancluever: This was to avoid some issues with short-circuiting on context timeouts in the `select` at the top of the inner function's `for` loop. I've fixed this and by just wrapping the select in a function so we can `break` properly and the fix will be in the next PR.

internal/servers/controller/testing.go
vancluever: This is for type asserting that the key is a string and the value is a `time.Time`, just basically validating the data integrity in the `sync.Map`. It's more about ensuring that if there are unexpected values in `WorkerStatusUpdateTimes`, it probably means that there is a problem somewhere else and we should probably investigate it more.

I think the other thing we could do is just type assert without validation, given that this function is only designed for testing currently. Will leave for now, but let me know your thoughts.

internal/servers/controller/testing.go
vancluever: See the [above reply](https://github.com/hashicorp/boundary/pull/1340#discussion_r670884050). Again, fine with just removing it completely and letting it panic instead, but I don't think we just continue given that we expect the data in the map to be a very specific structure. That is, unless I'm missing something and there would be data in `WorkerStatusUpdateTimes` where the key/value would not be `string`/`time.Time`.

jefferai: I don't think we need to panic; you'd skip the bad data and move on to the next piece of data in the range.

That said this is in testing code so especially in that case assuming it will be good and erroring on very unexpected data is probably fine.

PR #1344, opened 2021-06-23
Update unique error

louisruch:
### What does this PR do

Updates update cred store unique error to correct distinguish between token and store unique violation


louisruch:
@mgaffney PTAL, this is a interim change while we discuss the token work further, I created this ticket to track https://hashicorp.atlassian.net/browse/ICU-1800

mgaffney (approved):

PR #1347, opened 2021-06-24
Add db sentinel

louisruch:


internal/db/sentinel.go
```diff
@@ -0,0 +1,36 @@
+package db
```
mgaffney, line 1: Can we move this to a `sentinel` package? I think `sentinel.Prefix()` reads better than `db.Prefix()` (likewise for the other functions)

mgaffney (changes requested):

louisruch:
Closing in favor of new PR

PR #1354, opened 2021-06-25
Close connections on worker on shutdown, add force shutdown

vancluever:
This adds functionality to ensure that connections are closed on a
worker when the server is shut down or reloaded.

Connections are closed, and then the shutdown process waits for the next
report to the controller to go through, waiting the grace period that
was added in #1330. The rest of shutdown is then proceeded with.
Listener shutdown has been moved to the start of the process to ensure
that new connections can't be made while we are shutting down.

Additionally, functionality has been added to force shut down servers
after a second interrupt has been received after the first one. This
takes the form of code in both the server and dev mode code paths that
just performs an os.Exit if another shutdown is received during the
normal shutdown process.

Finally, the commit also includes some refactoring to tests so that we
could use some of the new E2E testing code to test session connections
during shutdown and reload.


vancluever:
Screenshots for the forced shutdown:

## `boundary server`

image

## `boundary dev`

(With debug logging to demonstrate gap between initial CTRL-C and forced shutdown)

image

malnick:
Linking related issue: https://github.com/hashicorp/boundary/issues/894

vancluever:
rebased off of #1340

internal/cmd/commands/dev/dev.go
```diff
@@ -565,7 +568,23 @@ func (c *Command) Run(args []string) int {
...
 		case <-c.ShutdownCh:
-			c.UI.Output("==> Boundary dev environment shutdown triggered")
+			c.UI.Output("==> Boundary dev environment shutdown triggered, interrupt again to force")
+
+			// Add a force-shutdown goroutine to consume another interrupt
+			abortForceShutdownCh := make(chan struct{})
+			defer close(abortForceShutdownCh)
+			go func() {
+				shutdownCh := make(chan os.Signal, 4)
+				signal.Notify(shutdownCh, os.Interrupt, syscall.SIGTERM)
```
jefferai, line 578: I think this is missable if the second interrupt arrives prior to this being registered. I don't have a great idea for how to do it better, and it's unlikely to happen since the second would have to come _very_ quickly. Just calling it out in case you have other ideas.

vancluever: Yeah, none short of rewriting [`MakeShutdownCh`](https://github.com/hashicorp/boundary/blob/4ba7eb5ff44d9a1541adf42f70d85541ed82ce28/internal/cmd/base/base.go#L106-L119). I'll noodle on it for a few days.

vancluever:
Going to hold off on fixing conflicts for now until #1400 is merged. After that there will probably be a reabase and squash here so that we can track code that will be moving to the `helper` package.

vancluever:
As mentioned, this has been rebased and squashed and should be ready for review now!

internal/tests/helper/testing_helper.go
```diff
@@ -0,0 +1,490 @@
...
+	"github.com/hashicorp/boundary/api/targets"
+	"github.com/hashicorp/boundary/globals"
+	targetspb "github.com/hashicorp/boundary/internal/gen/controller/api/resources/targets"
+	"github.com/hashicorp/boundary/internal/proxy"
+	"github.com/hashicorp/boundary/internal/servers/controller/common"
+	"github.com/hashicorp/boundary/internal/servers/worker"
+	"github.com/hashicorp/boundary/internal/session"
+	"github.com/hashicorp/go-cleanhttp"
+	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/vault/sdk/helper/base62"
```
jefferai, line 28: Can you repoint this to `github.com/hashicorp/go-secure-stdlib/base62` and `go mod tidy` after?

internal/cmd/commands/dev/dev.go
jefferai: Resolving for now, I don't have a better idea and I think it's a very unlikely scenario.

internal/servers/worker/worker.go
```diff
@@ -132,22 +132,61 @@ func (w *Worker) Shutdown(skipListeners bool) error {
...
 		if err := w.conf.Eventer.FlushNodes(context.Background()); err != nil {
 			return fmt.Errorf("error flushing worker eventer nodes: %w", err)
 		}
 	}
+
+	w.logger.Debug("shutdown successful")
 	return nil
 }
 
+func (w *Worker) CloseConnectionsForShutdownOrReload() {
```
jefferai, line 185: This function is only used in one place, can this be inlined above?

jefferai:
Looking good! Just two minor comments.

vancluever:
@jefferai updated!

jefferai (approved):
Looks good pending fixing merge conflicts and running `go mod tidy`! Nice work!

jefferai (approved):

PR #1355, opened 2021-06-25
integrate events into the servers

jimlambrt:
This PR includes:

- Adding support for an envvar BOUNDARY_DEVELOPER_ENABLE_EVENTS
- Changes in the event package based on the work to integrate events into our workers/controllers.   This included:
    - Adding support for an envvar defined in globals.BOUNDARY_DEVELOPER_ENABLE_EVENTS and event.TestEnableEventing(..) so other devs can easily flip support on/off in their unit tests.
    - Adding a SerializedWriter to deal with concurrent writes made to stderr via both a hclog and an eventer.
    - Adding hcl tags to the EventerConfig
    - Adding the StderrSink type and removing the type for a stdout sink
    - Exporting a few Validate functions
- Changes to servers:
   - Added support for a serialization lock which is shared between hclog and eventing when writing to stderr
   - Added support to setup eventing for cmds that start servers
   - Added a new handler wrapper to start and flush with gated events for audit and observations
   - Added support to Flush gated nodes on shutdown

Reviewers might find this helpful when manually testing events:
`BOUNDARY_DEVELOPER_ENABLE_EVENTS=true boundary dev 2>&1 | jq -R -r '. as $line | try fromjson catch $line'`


internal/observability/event/eventer_config.go
```diff
@@ -6,24 +6,15 @@ import (
...
+	ObservationsEnabled bool         `hcl:"observations_enabled"` // ObservationsEnabled specifies if observation events should be emitted.
+	Sinks               []SinkConfig `hcl:"sinks"`                // Sinks are all the configured sinks
 }
 
-// validate will validate the config. BTW, a config isn't required to have any
+// Validate will Validate the config. BTW, a config isn't required to have any
 // sinks to be valid.
-func (c *EventerConfig) validate() error {
+func (c *EventerConfig) Validate() error {
 	const op = "event.(EventerConfig).validate"
```
s-christoff, line 17:
```suggestion
	const op = "event.(EventerConfig).Validate"
```

internal/observability/event/sink_format.go
```diff
@@ -10,7 +10,7 @@ const (
 
 type SinkFormat string // SinkFormat defines the formatting for a sink in a config file stanza (json)
 
-func (f SinkFormat) validate() error {
+func (f SinkFormat) Validate() error {
 	const op = "event.(SinkFormat).validate"
```
s-christoff, line 14:
```suggestion
	const op = "event.(SinkFormat).Validate"
```

internal/servers/common/handler_test.go
```diff
@@ -0,0 +1,570 @@
...
+}
+
+type testMockBroker struct {
+	errorOnSendAudit       bool
+	errorOnSendObservation bool
+	errorOnFlush           bool
+}
+
+func (b *testMockBroker) Send(ctx context.Context, t eventlogger.EventType, payload interface{}) (eventlogger.Status, error) {
+	const op = "common.(testBroker).Send"
```
s-christoff, line 468:
```suggestion
	const op = "common.(testMockBroker).Send"
```

internal/observability/event/eventer_config.go
```diff
@@ -6,24 +6,15 @@ import (
...
-	AuditEnabled        bool              // AuditEnabled specifies if audit events should be emitted.
-	ObservationsEnabled bool              // ObservationsEnabled specifies if observation events should be emitted.
-	Sinks               []SinkConfig      // Sinks are all the configured sinks
+	AuditEnabled        bool         `hcl:"audit_enabled"`        // AuditEnabled specifies if audit events should be emitted.
+	ObservationsEnabled bool         `hcl:"observations_enabled"` // ObservationsEnabled specifies if observation events should be emitted.
+	Sinks               []SinkConfig `hcl:"sinks"`                // Sinks are all the configured sinks
 }
 
-// validate will validate the config. BTW, a config isn't required to have any
+// Validate will Validate the config. BTW, a config isn't required to have any
```
s-christoff, line 14:
```suggestion
// Validate will Validate the config. A config isn't required to have any
```
I love the BTW but wondering if it should just be?

s-christoff (commented):
First run through with some comments, gonna take another look later

s-christoff (dismissed):
🤙 great work!

internal/cmd/base/servers.go
```diff
@@ -51,6 +54,9 @@ type Server struct {
 	CombineLogs bool
 	LogLevel    hclog.Level
 
+	SerializationLock *sync.Mutex
```
jefferai, line 57: I get after reading more of the PR that this is really "serialization to stderr" but initially it seemed like it was related to serialization of the config, e.g. if we get a SIGHUP. If you can think of a better name it might be clearer.

internal/observability/event/event_audit.go
```diff
@@ -14,7 +14,7 @@ const auditVersion = "v0.1"
 type auditEventType string
 
 const (
-	apiRequest auditEventType = "APIRequest" // ApiRequest defines an API request audit event type
+	ApiRequest auditEventType = "APIRequest" // ApiRequest defines an API request audit event type
```
jefferai, line 17: Is this used outside of Boundary? Just wondering if we want `APIRequest` vs. `API_Request` or `ApiRequest`

internal/servers/common/handler.go
```diff
@@ -0,0 +1,223 @@
...
+
+	"github.com/hashicorp/boundary/internal/auth"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/kms"
+	"github.com/hashicorp/boundary/internal/observability/event"
+	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/vault/sdk/helper/base62"
+)
+
+// UnknownStatusCode re-uses CloudFlare's popular extended 520 status code.
```
jefferai, line 17: This doesn't appear to be used anywhere

internal/observability/event/eventer.go
```diff
@@ -46,69 +46,93 @@ type broker interface {
...
+	defer sysEventerLock.Unlock()
+	sysEventer = e
 	return nil
 }
 
-// SysEventer returns the "system wide" eventer for Boundary.
+// SysEventer returns the "system wide" eventer for Boundary and can/will return
+// a nil Eventer
 func SysEventer() *Eventer {
 	return sysEventer
```
jefferai, line 115: Just a reminder to put an rlock here

internal/observability/event/eventer.go
```diff
@@ -142,21 +166,39 @@ func NewEventer(log hclog.Logger, c EventerConfig, opt ...Option) (*Eventer, err
 		return nil, fmt.Errorf("%s: failed to register json node: %w", op, err)
 	}
 
+	if opts.withSerializationLock == nil {
+		opts.withSerializationLock = new(sync.Mutex)
```
jefferai, line 170: Rather than do this I'd honestly just make it required. If we don't have a serialization lock we are at real risk of something going wonky with any legacy or fallback logger code.

internal/observability/event/serialized_writer.go
```diff
@@ -0,0 +1,39 @@
...
+
+	s.l.Lock()
+	defer s.l.Unlock()
+	reader := bytes.NewReader(p)
+
+	n, err := reader.WriteTo(s.w)
+	if err != nil {
+		return 0, err
+	}
+	return int(n), err
```
jefferai, line 38: Rather than cast this, why not return int64? All `io.Reader`-ish functions use that so it seems more standard.

internal/servers/common/handler.go
```diff
@@ -0,0 +1,223 @@
...
+		publicId, _, _ := auth.GetTokenFromRequest(logger, kms, r)
+
+		var err error
+		info := &event.RequestInfo{
+			Id:       generatedTraceId(ctx),
+			PublicId: publicId,
+			Method:   r.Method,
+			Path:     r.URL.RequestURI(),
+		}
+		ctx, err = event.NewRequestInfoContext(ctx, info)
```
jefferai, line 134: Does it make sense to combine the request info and eventer context functions? That way we're not deriving another layer of contexts.

jefferai:
Just a couple of minor comments. Not hitting approve as a couple of those need to be addressed, but looks great!

internal/cmd/base/servers.go
jimlambrt: Renamed to StderrLock.  ty!

internal/observability/event/event_audit.go
jimlambrt: Currently, It's slated to be in audit events and I was choosing to match Consul and Nomad's value.

internal/servers/common/handler.go
jimlambrt: ty!

internal/observability/event/eventer.go
jimlambrt: ty!

internal/observability/event/serialized_writer.go
jimlambrt: the interface requires an int
```
type Writer interface {
    Write(p []byte) (n int, err error)
}
```

internal/observability/event/eventer.go
jimlambrt: Agreed and this will make all the tests more explicit as well.  ty!

internal/servers/common/handler.go
jimlambrt: I kept them separate because I thought there'd be situations where you have an Eventer but no request.  Happy to discuss if that's really an edge case we shouldn't worry about.  We'd need to combine the Eventer and RequestInfo into a struct to have just one layer of context for them.

internal/observability/event/serialized_writer.go
jefferai: Wow. TIL that Writer uses int and Reader uses int64. I never really noticed it before. I guess they figure you're never going to manage to write more than 2^32 (except when it's 2^16 or 2^64) bytes in one call. Still...bizarre.

jefferai (approved):

PR #1357, opened 2021-06-25
[Merge to main] internal/servers/worker: Controller failure connection cleanup (#1330)

vancluever:
This PR is just a merge of `vancluever/session-cleanup` which was supposed to be a long-running feature branch created to address several session cleanup tasks, which possibly would have included a DB update. Its only commit is 4440f3c, previously approved in #1330.

---

Additional changes:

* Circle: Set acceptance test no_output_timeout to 30m


internal/tests/cluster/session_cleanup_test.go
```diff
@@ -0,0 +1,513 @@
...
+
+	// Resume the connection, and reconnect.
+	proxy.Resume()
+	time.Sleep(time.Second * 10)                          // Sleep to wait for worker to report back as healthy
+	sess = newTestSession(ctx, t, tcl, "ttcp_1234567890") // re-assign, other connection will close in t.Cleanup()
+	sConn = sess.Connect(ctx, t, logger)
+	sConn.TestSendRecvAll(t)
+}
+
+func TestWorkerSessionCleanupMultiController(t *testing.T) {
```
jimlambrt, line 125: I'm also confused about what non-test fixture is under test.

internal/tests/cluster/session_cleanup_test.go
```diff
@@ -0,0 +1,513 @@
...
+
+// newTestSession authorizes a session and creates all of the data
+// necessary to initialize.
+func newTestSession(
+	ctx context.Context,
+	t *testing.T,
+	tcl *targets.Client,
+	targetId string,
+) *testSession {
+	require := require.New(t)
```
jimlambrt, line 268: should we add a call to t.Helper() here and all the testSession receiver functions?

internal/tests/cluster/session_cleanup_test.go
```diff
@@ -0,0 +1,513 @@
...
+
+	return s
+}
+
+// connect returns a connected websocket for the stored session,
+// connecting to the stored workerAddr with the configured transport.
+//
+// The returned (wrapped) net.Conn should be ready for communication.
+func (s *testSession) connect(ctx context.Context, t *testing.T) net.Conn {
+	require := require.New(t)
```
jimlambrt, line 318: should we add a call to t.Helper() here?

jimlambrt (changes requested):
I'm sorry for the late review in this work.  Just a few questions.  ty!

internal/tests/cluster/session_cleanup_test.go
```diff
@@ -0,0 +1,513 @@
...
+	"github.com/mr-tron/base58"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/proto"
+	"nhooyr.io/websocket"
+	"nhooyr.io/websocket/wspb"
+)
+
+const testSendRecvSendMax = 60
+
+func TestWorkerSessionCleanup(t *testing.T) {
```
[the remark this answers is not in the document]
vancluever: Discussed this offline with @jimlambrt and the confusion was due to the use of the `TestConnection` and `TestWorker` helpers. This should be cleared up now with the discussion.

internal/tests/cluster/session_cleanup_test.go
vancluever: Discussed this offline with @jimlambrt and the confusion was due to the use of the `TestConnection` and `TestWorker` helpers. This should be cleared up now with the discussion.

internal/servers/worker/status.go
```diff
@@ -13,9 +15,48 @@ import (
...
+// period is defined the length of time since the last successful
+// update.
+//
+// The setting is derived from one of the following:
+//
+//   * BOUNDARY_STATUS_GRACE_PERIOD, if defined, can be set to an
+//   integer value to define the setting.
+//   * If this is missing, the default (30 seconds) is used.
+//
+func (w *Worker) statusGracePeriod() time.Duration {
```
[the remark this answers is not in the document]
vancluever: Discussed this with @jimlambrt - in light of this moving to `base.Server` in #1340, going to defer adding the extra tests until then.

internal/tests/cluster/session_cleanup_test.go
vancluever: This should be addressed now!

internal/tests/cluster/session_cleanup_test.go
vancluever: This should be addressed now!

internal/servers/worker/session.go
```diff
@@ -251,7 +251,7 @@ func (w *Worker) closeConnection(ctx context.Context, req *pbs.CloseConnectionRe
 	return resp, nil
 }
 
-func (w *Worker) closeConnections(ctx context.Context, closeMap map[string]string) error {
+func (w *Worker) closeConnections(ctx context.Context, closeMap map[string]string) {
```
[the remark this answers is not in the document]
vancluever: Assertions/tests have now been added.

internal/servers/worker/session.go
```diff
@@ -251,49 +252,132 @@ func (w *Worker) closeConnection(ctx context.Context, req *pbs.CloseConnectionRe
...
+// session.
+func (w *Worker) closeConnections(ctx context.Context, closeInfo map[string]string) {
+	if closeInfo == nil {
+		// This should not happen, but it's a no-op if it does. Just
+		// return.
+		return
+	}
+
+	w.logger.Trace("marking connections as closed", "session_and_connection_ids", fmt.Sprintf("%#v", closeInfo))
+	response, err := w.closeConnection(ctx, w.makeCloseConnectionRequest(closeInfo))
```
jimlambrt, line 270: This great, but you could just pull the required map keys via: `reflect.ValueOf(closeInfo).MapKeys()`  I'm not convinced it makes the code more readable though given the need to use the reflect package.  Just a thought... but feel free to ignore my ramblings here.

internal/servers/worker/session_test.go
```diff
@@ -0,0 +1,249 @@
+package worker
+
```
jimlambrt, line 2: these unit tests are great.  ty!

jimlambrt (approved):
Looks great.  I left a comment which is really nothing more than a "comment".  As we discussed via zoom, it would be great if Worker.closeConnections(...) returned an error but it's not required.

vancluever:
> Looks great. I left a comment which is really nothing more than a "comment". As we discussed via zoom, it would be great if Worker.closeConnections(...) returned an error but it's not required.

Yeah, and not against further refactoring in the future. The main reason why it's not so much right now is due to the dependency in the proxy handler, but that could probably be further refined for sure.

internal/servers/worker/session.go
vancluever: Yeah, I thought about the fact that we use the map without using the values (which is why I noted in the description), and given the narrowness of the scope here hopefully it's not a big deal. Just on a purely pedantic basis I'd be more into hand-rolling a function to collect the keys versus using reflect just to avoid the extra overhead of using the package in general. But that'd not really be relevant here.

jimlambrt (dismissed):

jimlambrt (approved):

PR #1358, opened 2021-06-26
Update primary error functions to take a context, deprecate old functions

s-christoff:


louisruch (commented):
This looks really good!  I have one comment about Wrap calling EDeprecated 
The only other general comment I would have is that all the test files only test EDeprecated, perhaps if we do add an unexported e that could be used instead...I will look again Monday great work so far!

internal/errors/error.go
```diff
@@ -109,7 +161,7 @@ func Wrap(e error, op Op, opt ...Option) error {
 		opt = append(opt, WithWrap(e))
 	}
 
-	return E(opt...)
+	return EDeprecated(opt...)
```
louisruch, line 164: Weird I came to edit my comment but couldn't find it...
I am not sure we should be calling `EDeprecated` from `Wrap` or `Convert` this would block us from eventually deprecating it.  I would suggest we add a new function `e` that performs what `EDeprecated` currently does.   Wrap and Convert can then call `e` instead of `EDeprecated`, `EDeprecated` can just return `e` and `NewDeprecated` can do its thing then call `e`.  This also removes the duplicate code between the `deprecated` functions and the new ones, it also means we could update all tests to just test `e` for now

internal/errors/error.go
```diff
@@ -62,19 +65,68 @@ func E(opt ...Option) error {
...
+	if msg != "" {
+		opt = append(opt, WithMsg(msg))
+	}
+
+	return E(ctx, opt...)
+}
+
+// EDeprecated is the legacy version of E which does not
+// create an event. Please refrain from using this.
+func EDeprecated(opt ...Option) error {
```
talanknight, line 103: Others may feel differently but instead of doing a phased deprecation of these internal functions, why not just make it a refactor and add the context at the caller site?  Where there is no context available we could use context.TODO().  That function's documentation says "Code should use context.TODO when
 it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter)."  which seems to fit our case.

jimlambrt: Hum... so there would seem to be a significant side effects from your suggested approach: none of the error events with context.TODO would have an associated request id which really is a broken state for observability and not something we'd likely want for 50% of our error events.  Also, this is an exported function and it's used in quite a few places outside the errors pkg.

talanknight: E and New should handle cases where ctx are passed to it w/o a request id present.  My suggested approach  doesn't seem any more broken than not calling the observability enabled error functions at all which is the approach this change implements.  This approach seems to ensure that EDeprecated and NewDeprecated will continue to exist forever given that it is used in so many places and there are places where requestId will never be present.

louisruch: I tend to agree with Todd; once we have fully migrated all errors, what do we want to do with errors generated outside the scope of a `RequestId`?  It seems that the event package foresaw the possibility of not having an eventer in the context and therefore uses the `SysEventer`.  It seems sane to use `context.TODO`

Another possibility is add the context to `New` and not `E`.  `E` was always meant to give the dev more granularity in error creation (i.e. errors that do not have a Message).  The usage of `E` is extremely sparse so we could have it be purely used for error creation.  

Something else to consider is there are a couple of usages of `New` and possibly `E` where we also call the `WithWrap` option, we could therefore have multiple `WriteError` events being called along the error path.

jimlambrt: We have accommodated errors that aren't associated with a request within WriteError(...) so there is no requirement that `NewDeprecated` and `EDeprecated` exist forever.    BTW, I wasn't try to suggest that every error must have a request id.  I believe what we're trying to do is making it explicitly clear in the code base everywhere we created errors without regard to their context.  In a few months, when someone looks at `errors.New` or `errors.E` with a context.TODO  they won't have to reason if it was intentional or just something missed when we tried to clean up all the bits during the refactor.

talanknight: Could we add the deprecation steps in the documentation for the new deprecated functions so we know 
1. what needs to be done to remove them and
2. we know when they are safe to remove?

Bonus points for creating jira tasks for the deprecation steps :)

jimlambrt: Happy to add some comments and Jira tasks about the steps needed to remove the deprecated functions.

jimlambrt:
> If we are planning to pass a context to Wrap in the future we should also make a WrapDeprecated function now; if we are not planning to pass the context to Wrap, we won’t be able to call E because we don’t have a context in the future. Similar comment on Convert.

+1 on this idea.

jimlambrt (changes requested):
Amazing work!  I just have a few comments/suggestions.

jefferai:
There are a large number of cases in this PR where there is a context in the function but EDeprecated is used instead of just using the context passed in. It would be good to go through and use the context when available.

internal/errors/error_test.go
```diff
@@ -15,7 +15,7 @@ import (
 
 func Test_ErrorE(t *testing.T) {
```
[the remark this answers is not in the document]
jimlambrt: fixed

internal/errors/option.go
```diff
@@ -55,3 +56,9 @@ func WithCode(code Code) Option {
 		o.withCode = code
 	}
 }
+
+func WithoutEvent(withoutEvent bool) Option {
```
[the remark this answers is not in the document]
jimlambrt: done.

internal/errors/error.go
```diff
@@ -46,7 +48,8 @@ type Err struct {
 //
```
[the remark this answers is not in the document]
jimlambrt: done.

internal/errors/error.go
```diff
@@ -46,7 +48,8 @@ type Err struct {
...
 // * WithCode() - allows you to specify an optional Code, this code will be prioritized
 // over a code used from WithWrap().
-func E(opt ...Option) error {
+func E(ctx context.Context, opt ...Option) error {
```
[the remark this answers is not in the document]
jimlambrt: added unit test for nil context and adding comment.

internal/errors/error.go
```diff
@@ -62,19 +65,26 @@ func E(opt ...Option) error {
...
 		Code:    code,
 		Op:      opts.withOp,
 		Wrapped: opts.withErrWrapped,
 		Msg:     opts.withErrMsg,
 	}
+	if opts.withoutEvent {
+		return err
+	}
+
+	event.WriteError(ctx, "", err)
```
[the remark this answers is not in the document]
jimlambrt: done.

internal/errors/error.go
```diff
@@ -62,19 +65,26 @@ func E(opt ...Option) error {
...
+
+	return err
 }
 
 // New creates a new Err with provided code, op and msg
 // It supports the options of:
 //
 // * WithWrap() - allows you to specify an error to wrap
-func New(c Code, op Op, msg string, opt ...Option) error {
+func New(ctx context.Context, c Code, op Op, msg string, opt ...Option) error {
```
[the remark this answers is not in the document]
jimlambrt: done

internal/errors/error.go
```diff
@@ -94,7 +104,75 @@ func New(c Code, op Op, msg string, opt ...Option) error {
 //
 // * WithMsg() - allows you to specify an optional error msg, if the default
 // msg for the error Code is not sufficient.
-func Wrap(e error, op Op, opt ...Option) error {
+func Wrap(ctx context.Context, e error, op Op, opt ...Option) error {
```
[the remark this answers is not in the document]
jimlambrt: done

internal/errors/error.go
jimlambrt: I've refactored a bit to resolve this issue.

louisruch (commented):
The `WithoutEvent` and using `context.Todo` works well and neatly resolves my concerns around tying us down to never being able to deprecate these functions.  ty

jimlambrt (approved):

PR #1359, opened 2021-06-28
internal/credential/vault: fix dropped error

alrs:
This fixes a dropped `err` variable in `internal/credential/vault`.


louisruch (approved):
Thank you!

PR #1360, opened 2021-06-28
Add WriteSysEvent to the events package.

jimlambrt:
Creating a new event type for SysEvents makes it very explicit that they are not associated with an API request.  It also will 
allow operators to easily manage these system events which will likely be far more chatty than any other type.

Also.. this is a precursor to deprecating all "traditional" logging in the Boundary system.


s-christoff (approved):
🤙

PR #1363, opened 2021-06-29
Allow set-credential-library when a single library id is provided.

talanknight:


internal/cmd/commands/targetscmd/funcs.go
```diff
@@ -330,12 +330,13 @@ func extraFlagsHandlingFuncImpl(c *Command, _ *base.FlagSets, opts *[]targets.Op
...
 		switch len(c.flagApplicationCredentialLibraries) {
 		case 0:
-		case 1:
-			if c.flagApplicationCredentialLibraries[0] == "null" {
-				*opts = append(*opts, targets.DefaultApplicationCredentialLibraryIds())
-			}
 		default:
 			*opts = append(*opts, targets.WithApplicationCredentialLibraryIds(c.flagApplicationCredentialLibraries))
+
+			if len(c.flagApplicationCredentialLibraries) == 1 && c.flagApplicationCredentialLibraries[0] == "null" {
```
louisruch, line 336: Since we do a len check `if len(c.flagApplicationCredentialLibraries) == 0 {` off the bat we can probably drop this switch entirely.  Or move the len check into the `case 0:` not sure what will look better when we add more purposes.  Also good ignoring this for now and just leaving it as is...

louisruch (approved):
Comment about adjusting the checks but LGTM either way

internal/cmd/commands/targetscmd/funcs.go
jefferai: Personally I'd prefer keeping the `case 1` state and simply doing an `else` there, and leaving the default to just always add whatever we have.

jefferai: To be clear, it's not that I think this will _not work_, I just think it's a little more clear as to the behavior at different lengths.

PR #1367, opened 2021-06-29
[WIP] feat: add bats tests for credential stores

malnick:


moduli:
I believe these tests were already added in https://github.com/hashicorp/boundary/pull/2174. I think we can close this PR.

PR #1369, opened 2021-06-30
worker: ensure connections are closed in local state on controller fail

vancluever:
This backports some test work that is being added in #1340 to assert
that the worker is actually marking connections closed in its *local*
state on a failure to contact the controller. This was missed in #1357
during some late-stage refactoring.

Summary of the changes:

* If we can't actually contact the controller during connection closure
to mark the connections as closed, we actually construct a "fake"
response that just simply includes a closed state for all connections
requested. This is then used to update the local connection state.

* Added a 3 second timeout to the CloseConnection request to the
controller to ensure the call to close connections completed in a
timely manner and did not hang indefinitely.

Testing enhancements:

* Added a new helper in TestHelper called LookupSession that allows the
lookup of a session from the worker's internal session state.

* This is used in the new ExpectConnectionStateOnWorker check in the E2E
test, which is part of a larger amount of enhancements to the E2E test
that are coming in #1340.

* Added a ConnectionStatusFromProtoVal helper function to the session
package to allow the conversion of protobuf connection status values to
their higher-level counterparts; essentially it is a reverse of
ConnectionStatus.ProtoVal. This is used in some of our tests (with more
coming in #1340) to just help with keeping things more idiomatic.

### DIFF
--- a/internal/servers/worker/session.go
+++ b/internal/servers/worker/session.go
@@ -15,10 +15,9 @@ import (
 	"github.com/hashicorp/boundary/internal/session"
 )
 
-const (
-	validateSessionTimeout              = 90 * time.Second
-	errMakeSessionCloseInfoNilCloseInfo = "nil closeInfo supplied to makeSessionCloseInfo, this is a bug, please report it"
-)
+const validateSessionTimeout = 90 * time.Second
+
+var errMakeSessionCloseInfoNilCloseInfo = errors.New("nil closeInfo supplied to makeSessionCloseInfo, this is a bug, please report it")
 
 type connInfo struct {
 	id         string
@@ -271,6 +270,7 @@ func (w *Worker) closeConnections(ctx context.Context, closeInfo map[string]stri
 	// How we handle close info depends on whether or not we succeeded with
 	// marking them closed on the controller.
 	var sessionCloseInfo map[string][]*pbs.CloseConnectionResponseData
+	var err error
 
 	// TODO: This, along with the status call to the controller, probably needs a
 	// bit of formalization in terms of how we handle timeouts. For now, this
@@ -287,11 +287,16 @@ func (w *Worker) closeConnections(ctx context.Context, closeInfo map[string]stri
 		)
 
 		// Since we could not reach the controller, we have to make a "fake" response set.
-		sessionCloseInfo = w.makeFakeSessionCloseInfo(closeInfo)
+		sessionCloseInfo, err = w.makeFakeSessionCloseInfo(closeInfo)
 	} else {
 		// Connection succeeded, so we can proceed with making the sessionCloseInfo
 		// off of the response data.
-		sessionCloseInfo = w.makeSessionCloseInfo(closeInfo, response)
+		sessionCloseInfo, err = w.makeSessionCloseInfo(closeInfo, response)
+	}
+
+	if err != nil {
+		w.logger.Error(err.Error())
+		w.logger.Error("serious error in processing return data from controller, aborting marking connections as closed")
 	}
 
 	// Mark connections as closed
@@ -334,11 +339,9 @@ func (w *Worker) makeCloseConnectionRequest(closeInfo map[string]string) *pbs.Cl
 func (w *Worker) makeSessionCloseInfo(
 	closeInfo map[string]string,
 	response *pbs.CloseConnectionResponse,
-) map[string][]*pbs.CloseConnectionResponseData {
+) (map[string][]*pbs.CloseConnectionResponseData, error) {
 	if closeInfo == nil {
-		// Should never happen, panic if it does. Results will be
-		// undefined.
-		panic(errMakeSessionCloseInfoNilCloseInfo)
+		return nil, errMakeSessionCloseInfoNilCloseInfo
 	}
 
 	result := make(map[string][]*pbs.CloseConnectionResponseData)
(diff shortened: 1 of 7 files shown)


internal/servers/worker/session.go
```diff
@@ -331,6 +349,28 @@ func (w *Worker) makeSessionCloseInfo(
...
 
+// makeFakeSessionCloseInfo makes a "fake" makeFakeSessionCloseInfo, intended
+// for use when we can't contact the controller.
+func (w *Worker) makeFakeSessionCloseInfo(
+	closeInfo map[string]string,
+) map[string][]*pbs.CloseConnectionResponseData {
+	if closeInfo == nil {
+		// Should never happen, panic if it does. Results will be
+		// undefined.
+		panic(errMakeSessionCloseInfoNilCloseInfo)
```
jefferai, line 360: Don't put panics into server code, please. Return an error that can be logged.

jefferai: (Please also fix this in `makeSessionCloseInfo`)

internal/servers/worker/session.go
```diff
@@ -331,6 +349,28 @@ func (w *Worker) makeSessionCloseInfo(
...
+	for connectionId, sessionId := range closeInfo {
+		result[sessionId] = append(result[sessionId], &pbs.CloseConnectionResponseData{
+			ConnectionId: connectionId,
+			Status:       pbs.CONNECTIONSTATUS_CONNECTIONSTATUS_CLOSED,
+		})
+	}
+
+	return result
+}
+
```
jefferai, line 373: In setCloseTimeForResponse below, do we really want it to always mark the connection's close time as now? What if it was already closed -- we probably don't want to keep moving the close time forward, but want to preserve it as it was.

internal/servers/worker/session.go
```diff
@@ -267,17 +267,35 @@ func (w *Worker) closeConnections(ctx context.Context, closeInfo map[string]stri
...
+	response, err := w.closeConnection(closeConnCtx, w.makeCloseConnectionRequest(closeInfo))
 	if err != nil {
 		w.logger.Error("error marking connections closed", "error", err)
 		w.logger.Warn(
 			"error contacting controller, connections will be closed only on worker",
 			"session_and_connection_ids", fmt.Sprintf("%#v", closeInfo),
 		)
+
+		// Since we could not reach the controller, we have to make a "fake" response set.
+		sessionCloseInfo = w.makeFakeSessionCloseInfo(closeInfo)
```
jefferai, line 290: Rather than have a second function to generate fake data, can you simply keep the same single function but have behavior triggered on whether the response passed in is `nil` or not?

internal/servers/worker/status.go
```diff
@@ -16,7 +16,7 @@ import (
 // In the future we could make this configurable
 const (
 	statusInterval           = 2 * time.Second
-	statusTimeout            = 10 * time.Second
+	statusTimeout            = 3 * time.Second
```
jefferai, line 19: This makes me a bit nervous. I wonder if a heavily loaded DB might take more than 3 seconds to return a reply, especially if there are large numbers of sessions and only a single worker.

internal/tests/cluster/session_cleanup_test.go
```diff
@@ -363,6 +398,88 @@ func (s *testSession) TestNoConnectionsLeft(t *testing.T) {
...
+	for {
+		if ctx.Err() != nil {
+			break
+		}
+
+		for _, conn := range s.testWorkerConnectionInfo(t, tw) {
+			actualStates[conn.Id] = session.ConnectionStatusFromProtoVal(conn.Status)
+		}
+
+		if reflect.DeepEqual(expectStates, actualStates) {
```
jefferai, line 441: Not important to change here, but we tend to use go-cmp for this as it prints out much easier to understand information, like the exact differences.

internal/servers/worker/session.go
vancluever: Sure, just want to point out though that this is mainly just carrying forward existing functionality. This logic does not exist currently, IIRC. Here's the [existing code from v0.3.0](https://github.com/hashicorp/boundary/blob/407a21991aa7dc550967720466d4b10c2e02ee1c/internal/servers/worker/session.go#L288-L294).

internal/servers/worker/session.go
vancluever: Is the response guaranteed to be nil if there is an error in the API call?

internal/servers/worker/status.go
vancluever: This was discussed internally, I did check in, I can forward the discussion if you like. Let me know!

internal/tests/cluster/session_cleanup_test.go
vancluever: Yeah, this is not for assertion's sake (would have used testify if that was the case). This is just to check for equality.

internal/servers/worker/session.go
vancluever: Now fixed!

internal/servers/worker/session.go
vancluever: Discussed offline that this is something that we want to defer to post-release, given that it would ultimately be new behavior. Additionally, connection close requests at the controller level are something that could probably use a bit more work in general.

internal/servers/worker/session.go
vancluever: Discussed offline that this could be something for a larger refactor to perform short-circuit local state management when a connection cannot be made to the controller.

internal/servers/worker/status.go
vancluever: Discussed internally to change the timeout to 5 seconds (will be done shortly).

jimlambrt (approved):

PR #1375, opened 2021-07-02
Fix Build UI CI job

mdeggies:
The `Build ui layer` CI job is failing on the step that builds packages under `ui/desktop/electron-app`. The version of `electron-rebuild` being used depends on a pretty old version of `lzma-native`, which calls `node-pre-gyp install --fallback-to-build && rimraf build` if the regular install returns any errors. `node-pre-gyp` [requires](https://github.com/nodejs/node-gyp#on-unix) python3 to run, which isn't available in the [base image](https://github.com/hashicorp/boundary/blob/main/packages-oss.yml#L57) where we build the UI. 

Adding `build-essential` and `python3` to [packages-oss.yml](https://github.com/hashicorp/boundary/blob/main/packages-oss.yml#L72) and then running `make packages` fixes the issue, but I think we could also try upgrading to the latest version (^0.8.1) of `lzma-native` in `boundary-ui/ui/desktop/electron-app`. For good measure, we could also upgrade to the latest versions of the `gyp` packages. After pushing up a commit to `boundary-ui`, we can then run `export UI_COMMITISH=$SHA` and run `make update-ui-version` from within boundary to include our test version of the boundary-ui.. and see if that works! If so, we can remove the changes made in this PR which install `build-essential` and `python3`.


jimlambrt:
@mdeggies can we close this PR in favor of #1377 ?

jimlambrt:
I think we can safely close this PR for now.

PR #1377, opened 2021-07-02
Fixing ui build issue by adding build-essential and python3

jimlambrt:
@mdeggies This PR would replace #1375 which was opened on a feature branch which contains a lot of changes not related to the ui build problem.

Michele's original PR description:
The `Build ui layer` CI job is failing on the step that builds packages under `ui/desktop/electron-app`. The version of `electron-rebuild` being used depends on a pretty old version of `lzma-native`, which calls `node-pre-gyp install --fallback-to-build && rimraf build` if the regular install returns any errors. `node-pre-gyp` [requires](https://github.com/nodejs/node-gyp#on-unix) python3 to run, which isn't available in the [base image](https://github.com/hashicorp/boundary/blob/main/packages-oss.yml#L57) where we build the UI. 

Adding `build-essential` and `python3` to [packages-oss.yml](https://github.com/hashicorp/boundary/blob/main/packages-oss.yml#L72) and then running `make packages` fixes the issue, but I think we could also try upgrading to the latest version (^0.8.1) of `lzma-native` in `boundary-ui/ui/desktop/electron-app`. For good measure, we could also upgrade to the latest versions of the `gyp` packages. After pushing up a commit to `boundary-ui`, we can then run `export UI_COMMITISH=$SHA` and run `make update-ui-version` from within boundary to include our test version of the boundary-ui.. and see if that works! If so, we can remove the changes made in this PR which install `build-essential` and `python3`.


mdeggies:
Hi Jim, the PR you are attempting to close is still a WIP. I opened this
pretty late at night and forgot to add that to the title, but I’ll do that
when I’m back in front of my laptop. The goal in opening this PR was just
to demonstrate a quick fix for Sarah to unblock her branch, and provide
enough context in the description to allow the on-call release engineer to
investigate updating the necessary dependencies.

If this is something that urgently needs to be fixed, feel free to merge
your PR. Otherwise, we can have the on-call release engineer pick this back
up on the 6th.

jimlambrt:
Looks like this isn't needed.

PR #1386, opened 2021-07-08
internal/db: add enumeration constraints for the server table

vancluever:
This adds enumeration constraints to the server table, ensuring that no
servers can be added other than the types "controller" and "worker".

This should already be impossible to do in code, but this adds a check
on the database level to ensure it's the case as well.


internal/db/schema/migrations/postgres/11/01_server_type_enum.up.sql
```diff
@@ -0,0 +1,17 @@
+begin;
+
+create table server_type_enm (
+  name text primary key not null,
```
mgaffney, line 4: No need for `not null` since primary keys are always not null.

mgaffney: Also, there should not be a comma at the end of this line since the constraint is a column constraint. Adding the comma makes the constraint a table constraint. Postgres will allow it but other database will not.

internal/db/schema/migrations/postgres/11/01_server_type_enum.up.sql
```diff
@@ -0,0 +1,17 @@
...
+  constraint only_predefined_server_types_allowed
+    check (name = ANY (array['controller'::text, 'worker'::text]))
```
mgaffney: We have a [standard](https://github.com/hashicorp/boundary/blob/c9e4c877b89038c9e9d4f1edecb15a1123073519/internal/db/schema/migrations/postgres/2/03_oidc_enm.up.sql#L8-L10) [pattern](https://github.com/hashicorp/boundary/blob/c9e4c877b89038c9e9d4f1edecb15a1123073519/internal/db/schema/migrations/postgres/10/03_credential.up.sql#L243-L249) we use for checking for valid values. Is there a specific reason for not following that pattern?

jimlambrt, line 6: why not use the current pattern and just do `name in (...)`?  Is there some advantage to using `ANY` and an array?

internal/db/schema/migrations/postgres/11/01_server_type_enum.up.sql
```diff
@@ -0,0 +1,17 @@
...
+);
+
+insert into server_type_enm (name) values
+  ('controller'),
+  ('worker');
+
+alter table server
+    add foreign key (type) references server_type_enm(name)
```
mgaffney, line 19: We name foreign key constraints. Can you add a name following our [foreign key](https://github.com/hashicorp/boundary/blob/c9e4c877b89038c9e9d4f1edecb15a1123073519/internal/db/schema/migrations/postgres/2/04_oidc.up.sql#L16-L17) [naming pattern](https://github.com/hashicorp/boundary/blob/c9e4c877b89038c9e9d4f1edecb15a1123073519/internal/db/schema/migrations/postgres/10/04_vault_credential.up.sql#L167-L168)?

internal/db/schema/migrations/postgres/11/01_server_type_enum.up.sql
vancluever: Hmm... this might be just how psql writes this and not from a pattern I saw. Will fix!

mgaffney (changes requested):

internal/db/schema/migrations/postgres/11/01_server_type_enum.up.sql
vancluever: Updated!

vancluever: Addressed [above](https://github.com/hashicorp/boundary/pull/1386#discussion_r666504684).

internal/db/schema/migrations/postgres/11/01_server_type_enum_test.go
```diff
@@ -0,0 +1,191 @@
...
+	require.NoError(err)
+	require.Len(actualWorkers, 1)
+	require.Equal(expectedWorkers, actualWorkers)
+
+	// Assert the state of the newly created enum table
+	rows, err := d.QueryContext(ctx, "select * from server_type_enm")
+	require.NoError(err)
+
+	// Column types
+	colTypes, err := rows.ColumnTypes()
```
jimlambrt, line 128: I'm not sure we need to test that the DDL is correct, given that you have a FK that ensures it's type and name as well.  The same is likely true of the inserted data.

internal/db/schema/migrations/postgres/11/01_server_type_enum_test.go
```diff
@@ -0,0 +1,191 @@
...
+	require.EqualError(err, expectEnumConstraintErr)
+	require.Nil(result)
+
+	// Try adding a broken server type
+	badServer := &servers.Server{
+		PrivateId: "test-bad",
+		Type:      "bad",
+		Address:   "127.0.0.1",
+	}
+	_, rowsUpdated, err := repo.UpsertServer(ctx, badServer)
```
jimlambrt, line 157: Perhaps consider not using the repo.UpsertServer here and creating the dep between this test and how the repo function is implemented.  This sort of applies to all these repo functions...

internal/servers/testing.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/hashicorp/boundary/internal/types/scope"
+	wrapping "github.com/hashicorp/go-kms-wrapping"
+	"github.com/jinzhu/gorm"
+	"github.com/stretchr/testify/require"
+)
+
+// TestRepo creates a repo that can be used for various purposes.
+// Crucially, it ensures that the global scope contains a valid root
+// key.
+func TestRepo(t *testing.T, conn *gorm.DB, rootWrapper wrapping.Wrapper) *Repository {
```
jimlambrt, line 19: Hum... I'm not sure we continued using this pattern after iam.TestRepo(...)  I think all the other repos are just created by following a pattern of getting a conn, wrapper, kmsCache and then creating the repo.  In this case, I think there's only one current reference...  perhaps we don't need this?

jimlambrt (changes requested):
Overall looks great.  Just a few questions.

internal/db/schema/migrations/postgres/11/01_server_type_enum.up.sql
vancluever: Should both be addressed!

internal/db/schema/migrations/postgres/11/01_server_type_enum.up.sql
vancluever: Fixed!

vancluever:
@jimlambrt @mgaffney thanks for the feedback! I've addressed the issues with the SQL conventions/etc.

@jimlambrt I think the gist of most of the comments regarding the tests is, aside from not worrying about asserting the DDL, is that we should probably just focus on testing on the DB and not through the domains to ensure things are as decoupled as possible? I think that could be pretty reasonable especially given how minimal the `server` table is at the time. Let me know!

internal/db/schema/migrations/postgres/11/01_server_type_enum.up.sql
```diff
@@ -0,0 +1,24 @@
...
+create table server_type_enm (
+  name text primary key
+    constraint only_predefined_server_types_allowed
+      check (
+        name in (
+          'controller',
+          'worker'
+        )
+      )
+);
```
mgaffney, line 12: Can you add a [comment](https://github.com/hashicorp/boundary/blob/c9e4c877b89038c9e9d4f1edecb15a1123073519/internal/db/schema/migrations/postgres/10/04_vault_credential.up.sql#L135-L137) to the table?

(I left a similar comment in my original review but it wasn't saved)

vancluever: Done!

internal/db/schema/migrations/postgres/11/01_server_type_enum.up.sql
```diff
@@ -0,0 +1,27 @@
...
+comment on table server_type_enm is
+  'server_type_enm is an enumeration table for server types.'
```
mgaffney, line 14:
```suggestion
  'server_type_enm is an enumeration table for server types. '
```
Postgres automatically concatenates multi line comment strings but you need to add an extra space at the end of each line except the final line.

internal/db/schema/migrations/postgres/11/01_server_type_enum_test.go
mgaffney: I agree. It feels like an anti-pattern to start testing the DDL.

mgaffney (approved):

jimlambrt (approved):

PR #1389, opened 2021-07-09
Remove errant tls_disable 

mikegreen:
Per https://www.boundaryproject.io/docs/configuration/listener/tcp#tls - having this here is invalid.
All tls parameters are valid only for the api listener. cluster and proxy connections use their own ephemeral TLS stacks


jefferai:
Thanks!

PR #1390, opened 2021-07-09
Events: add node for formatting/filtering cloudevents

jimlambrt:
PR includes:

- Updated eventlogger dependency that includes cloud events
- New event node for formatting and filtering cloudevents

A subsequent PR will integrate this new node into the EventerConfig and NewEventer(...)


internal/observability/event/event_observation.go
```diff
@@ -4,17 +4,21 @@ import (
...
 
 // observationVersion defines the version of observation events
 const observationVersion = "v0.1"
 
 type observation struct {
-	*eventlogger.SimpleGatedPayload
-	Version     string       `json:"version"`
-	Op          Op           `json:"op,omitempty"`
-	RequestInfo *RequestInfo `json:"request_info,omitempty"`
+	Version     string                 `json:"version"`
```
jefferai, line 15: Can you comment this struct? E.g. what Header vs. Detail are, when you'd want to set Flush, etc.

internal/servers/controller/testing.go
```diff
@@ -468,7 +479,7 @@ func NewTestController(t *testing.T, opts *TestControllerOpts) *TestController {
 		opts.Config.Controller = new(config.Controller)
 	}
 	if opts.Config.Controller.Name == "" {
-		opts.Config.Controller.Name, err = base62.Random(5)
+		opts.Config.Controller.Name, err = opts.Config.Controller.InitNameIfEmpty()
```
jefferai, line 482: I think this section can be removed right? It'll initialize above now.

internal/observability/event/event_observation.go
jimlambrt: this struct is not exported and devs will never use this struct directly.

internal/servers/controller/testing.go
jimlambrt: I've updated the code block above...

internal/cmd/base/servers.go
```diff
@@ -150,7 +150,9 @@ func (b *Server) SetupEventing(logger hclog.Logger, serializationLock *sync.Mute
 	if serializationLock == nil {
 		return berrors.New(berrors.InvalidParameter, op, "missing serialization lock")
 	}
-
+	if serverName == "" {
+		return berrors.New(berrors.InvalidParameter, op, "missing server name lock")
```
jefferai, line 154:
```suggestion
		return berrors.New(berrors.InvalidParameter, op, "missing server name")
```

internal/cmd/commands/dev/dev.go
```diff
@@ -262,6 +269,40 @@ func (c *Command) Flags() *base.FlagSets {
...
+		Usage:      `Emit observation events. Supported values are "true" and "false".`,
+	})
+	f.StringVar(&base.StringVar{
+		Name:       "audit",
+		Target:     &c.flagAudit,
+		Completion: complete.PredictSet("true", "false"),
+		Usage:      `Emit audit events. Supported values are "true" and "false".`,
+	})
+	f.StringVar(&base.StringVar{
+		Name:       "sys-events",
```
jefferai, line 291: It feels incongruous to have this flag have `-events` on the end but not the other two. Might I suggest `audit-events` and `observation-events`? That also then makes it clear that the filters will apply to those as well.

Another nit: If it's spelled out "observation-events" (and `observations` right now) it should probably be `system-events`.

internal/cmd/config/config.go
```diff
@@ -128,6 +129,19 @@ type Controller struct {
 	StatusGracePeriodDuration time.Duration `hcl:"-"`
 }
 
+func (c *Controller) InitNameIfEmpty() (string, error) {
+	if c == nil {
+		return "", fmt.Errorf("controller config is empty")
```
jefferai, line 134: It's not the config that's empty, it's the controller itself. More to the point though, can you just remove the specific controller/worker references and combine this into a shared function in the `base` package? Maybe:

`c.Name, err = base.InitNameIfEmpty(c.Name)`

internal/observability/event/context.go
```diff
@@ -213,7 +216,7 @@ func addCtxOptions(ctx context.Context, opt ...Option) ([]Option, error) {
 			// there's no RequestInfo, so there's no id associated with the
 			// observation and we'll generate one and flush the observation
 			// since there will never be another with the same id
-			id, err := newId(string(ObservationType))
+			id, err := NewId("e")
```
jefferai, line 219: It's not like you to not use string constants. Curious about this change.

internal/cmd/base/servers.go
jimlambrt: ty!

internal/cmd/commands/dev/dev.go
jimlambrt: ty!

internal/observability/event/context.go
jimlambrt: It's not just "observations" any longer and I need to update the comment. ty!

I'll add a const as well.

internal/observability/event/node.go
```diff
@@ -0,0 +1,163 @@
...
+	// There isn't a clear way to differentiate between a JSON Pointer which doesn't represent
+	// the structure of the object being Matched and a JSON Pointer which references a field which
+	// is part of a sub structure that is nil in this item. Because of this, any filter which would
+	// result in an error using the underlying library is simply interpreted as not a match.
+	return err == nil && m
+}
+
+// wellKnownTypeFilterHook is passed to bexpr to treat all proto well-known
+// types as the types they wrap for comparison.
+func wellKnownTypeFilterHook(v reflect.Value) reflect.Value {
```
jefferai, line 133: I believe one of these functions exists already in the codebase (@talanknight wrote one IIRC). Might be good to see if they can be consolidated.

internal/observability/event/node.go
```diff
@@ -0,0 +1,163 @@
...
+	opts := getOpts(opt...)
+	n := Node{
+		FormatterFilter: &cloudevents.FormatterFilter{
+			Source: source,
+			Schema: opts.withSchema,
+			Format: format,
+		},
+	}
+
+	// intentionally not checking of allow and/or deny optional filters were
```
jefferai, line 46:
```suggestion
	// intentionally not checking if allow and/or deny optional filters were
```

internal/observability/event/node.go
talanknight: The hook @jefferai referred to is in internal/servers/controller/handlers/filtering.go

internal/cmd/config/config.go
jimlambrt: refactored the bits that generates into a shared func but I kept the paired down receiver functions.  I'm confused about it not being the Config that's empty since it's `c.Config.Controller` or `w.Config.Worker`

internal/observability/event/node.go
jimlambrt: Yeah, I just cut/pasted it. Where would these sort of general functions live in our packages?

jimlambrt: under `libs`?

jimlambrt: I've refactored WellKnownTypeFilterHook(...) into a new filter package.

internal/cmd/config/config.go
jefferai: The first line there -- `if c == nil` -- c is a `*Controller`. So the check is on that pointer, not `c.Config`. Does that make sense?

jefferai (approved):

jefferai:
Looks fantastic!

PR #1393, opened 2021-07-12
Update usage of shared-secure-libs

jefferai:
This removes copied code in our SDK module and removes dependency on Vault's SDK module.


internal/db/schema/migrations/postgres/0/01_domain_types.up.sql
```diff
@@ -5,38 +5,38 @@ check(
   length(trim(value)) > 10
 );
 comment on domain wt_public_id is
-'Random ID generated with github.com/hashicorp/vault/sdk/helper/base62';
+'Random ID generated with github.com/hashicorp/shared-secure-libs/base62';
```
louisruch, line 8: It probably doesn't really matter, but this wont cause the comment to update during a migration

louisruch (approved):
LGTM just one small comment re migrations

internal/db/schema/migrations/postgres/0/01_domain_types.up.sql
jefferai: Yeah, really not an issue!

PR #1395, opened 2021-07-13
Adding support to build multi-arch Boundary images

jacobmammoliti:
- I've updated the release Dockerfile to support multiple architectures.

- I've also added a snippet in the Makefile called `docker-multiarch-build` that will build and push multiple architectures of the image to our Dockerhub. I've done it with Docker's `buildx` feature.

### DIFF
--- a/docker/Release.dockerfile
+++ b/docker/Release.dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.13
+FROM docker.mirror.hashicorp.services/alpine:3.13
 
 ARG VERSION=0.4.0
 
--- a/docker/Release.dockerfile
+++ b/docker/Release.dockerfile
@@ -1,4 +1,4 @@
-FROM docker.mirror.hashicorp.services/alpine:3.10
+FROM alpine:3.13
 
 ARG VERSION=0.4.0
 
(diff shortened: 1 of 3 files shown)


docker/Release.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM docker.mirror.hashicorp.services/alpine:3.10
+FROM alpine:3.13
```
malnick, line 1: @mdeggies I think we need to stick to the hashicorp mirror for this image, is that correct? Do we have one for 3.13?

malnick (commented):
Looks great! I need to check with @mdeggies on whether or not we have a 3.13 alpine image in our HashiCorp mirror. Going to relay this to her for final review.

docker/Release.dockerfile
alvin-huang: Yes, this should be `docker.mirror.hashicorp.services/alpine:3.13`

jacobmammoliti: Cool. Updated now!

mdeggies: Just a note, in the future you can run `docker pull docker.mirror.hashicorp.servies/$whatever` locally to see if that image is supported in the docker mirror. If it's not, you can create a PR to add that here: https://github.com/hashicorp/docker-registry-mirror/blob/main/images.tf.

malnick (approved):

PR #1396, opened 2021-07-14
Ensure job progress is always reported

louisruch:
### What does this PR do

Adds a `WithProgress` option to the `CompleteRun` and `FailRun` calls so the final job status is always persisted.  Currently progress is only updated in the `MonitorJobs` loop, fast running jobs often start and finish before a single monitor job loop is triggered.  Even for long running jobs the final status should not be left to a race between `CompleteRun` and the `MonitorJobs` loop

### DIFF
--- a/internal/scheduler/job/repository_run.go
+++ b/internal/scheduler/job/repository_run.go
@@ -125,6 +125,10 @@ func (r *Repository) CompleteRun(ctx context.Context, runId string, nextRunIn ti
 	run.PrivateId = runId
 	_, err := r.writer.DoTx(ctx, db.StdRetryCnt, db.ExpBackoff{},
 		func(r db.Reader, w db.Writer) error {
+			// TODO (lcr 07/2021) this can potentially overwrite completed and total values
+			// persisted by the scheduler's monitor jobs loop.
+			// Add an on update sql trigger to protect the job_run table, once progress
+			// values are used in the critical path.
 			rows, err := r.Query(ctx, completeRunQuery, []interface{}{completed, total, runId})
 			if err != nil {
 				return errors.Wrap(err, op)
@@ -201,6 +205,10 @@ func (r *Repository) FailRun(ctx context.Context, runId string, completed, total
 	run.PrivateId = runId
 	_, err := r.writer.DoTx(ctx, db.StdRetryCnt, db.ExpBackoff{},
 		func(r db.Reader, w db.Writer) error {
+			// TODO (lcr 07/2021) this can potentially overwrite completed and total values
+			// persisted by the scheduler's monitor jobs loop.
+			// Add an on update sql trigger to protect the job_run table, once progress
+			// values are used in the critical path.
 			rows, err := r.Query(ctx, failRunQuery, []interface{}{completed, total, runId})
 			if err != nil {
 				return errors.Wrap(err, op)
--- a/internal/scheduler/job/repository_run.go
+++ b/internal/scheduler/job/repository_run.go
@@ -105,7 +105,8 @@ func (r *Repository) UpdateProgress(ctx context.Context, runId string, completed
 }
 
 // CompleteRun updates the Run repository entry for the provided runId.
-// It sets the status to 'completed' and updates the run's EndTime to the current database time.
+// It sets the status to 'completed', updates the run's EndTime to the current database
+// time, and sets the completed and total counts.
 // CompleteRun also updates the Job repository entry that is associated with this run,
 // setting the job's NextScheduledRun to the current database time incremented by the nextRunIn
 // parameter.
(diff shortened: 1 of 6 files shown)


louisruch:
@vancluever PTAL This is work that came up re a comment I made in your PR about always sending the status I realized that we will hardly ever report the status of fast running jobs

vancluever:
@louisruch I think the idea sounds good, but just wondering why this is an option versus just built into the scheduler? Is there a situation where we would _not_ want to record the final job status on completion?

louisruch:
> @louisruch I think the idea sounds good, but just wondering why this is an option versus just built into the scheduler? Is there a situation where we would _not_ want to record the final job status on completion?

I was worried about how to handle zero values which could clear a valid status in the db, and felt having it as an option removed this worry, perhaps its not a worry though, but having it as an option did seem clean...Thoughts @mgaffney

vancluever:
> I was worried about how to handle zero values which could clear a valid status in the db, and felt having it as an option removed this worry, perhaps its not a worry though, but having it as an option did seem clean...Thoughts @mgaffney

Hmm... maybe job progress belongs in its own table? That way, progress could ultimately be nil if need be.

louisruch:
> Hmm... maybe job progress belongs in its own table? That way, progress could ultimately be nil if need be.

That was something we discussed previously, but decided to not take action on at the time due to delivery of Vault phase one.  The full use case of progress is still being thought through, one idea was to allow schedulers to determine if a job is `stuck` and not making sufficient progress.  Determining if a job is `stuck` and then interrupting it to be started again is however a delicate balance.

vancluever:
> If we did break it into its own table, I would argue having it as an option is then preferred.

I think this loops back to my first question then... 🙂 

Basically, _when_ would we not want to log progress at the end of a job, and I guess more specifically how is this controlled? Is this just always going to be an internal only option never exposed to job implementations or scheduler engines or will it ever be of use to those use cases to not?

louisruch:
> > If we did break it into its own table, I would argue having it as an option is then preferred.
> 
> I think this loops back to my first question then... 🙂
> 
> Basically, _when_ would we not want to log progress at the end of a job, and I guess more specifically how is this controlled? Is this just always going to be an internal only option never exposed to job implementations or scheduler engines or will it ever be of use to those use cases to not?

The scheduler and job repo are currently only internal.  How this is exposed to external plugins and jobs in the future is still TBD.  Something I was thinking of adding to this PR but didn't was not including the `WithProgress` option if a job has 0 / 0 (completed/total).   If progress was in its own table, would we want to have no entry or have a entry with 0/0, I can see it going either way TBH, but we still need to have a couple discussions around that and the extent of how progress will be used.  

I am trying to think through who should be the decision maker in whether progress is reported, the job most likely knows best, so perhaps changing the interface to return a `*Status` is the best option, clearly indicating to the scheduler what it needs to do.

mgaffney:
I agree that recording the final progress of a job should not be optional. However, I don't think there is much benefit from moving it to a new table since we don't currently do anything with the values. @louisruch you might want to consider adding an "on update" trigger to prevent the current or total values from being updated to values less than the current values.

mgaffney (commented):
I agree that recording the final progress of a job should not be optional. However, I don't think there is much benefit from moving it to a new table since we don't currently do anything with the values. @louisruch you might want to consider adding an "on update" trigger to prevent the current or total values from being updated to values less than the current values.

louisruch:
@mgaffney @vancluever PTAL

mgaffney (dismissed):

internal/scheduler/job/repository_run.go
```diff
@@ -124,7 +125,7 @@ func (r *Repository) CompleteRun(ctx context.Context, runId string, nextRunIn ti
 	run.PrivateId = runId
 	_, err := r.writer.DoTx(ctx, db.StdRetryCnt, db.ExpBackoff{},
 		func(r db.Reader, w db.Writer) error {
-			rows, err := r.Query(ctx, completeRunQuery, []interface{}{runId})
+			rows, err := r.Query(ctx, completeRunQuery, []interface{}{completed, total, runId})
```
vancluever, line 132: Should this be using the reader and not the writer, given that the query is an update? Not too sure if it matters.

internal/scheduler/job/repository_run.go
```diff
@@ -201,6 +205,10 @@ func (r *Repository) FailRun(ctx context.Context, runId string, completed, total
 	run.PrivateId = runId
 	_, err := r.writer.DoTx(ctx, db.StdRetryCnt, db.ExpBackoff{},
 		func(r db.Reader, w db.Writer) error {
+			// TODO (lcr 07/2021) this can potentially overwrite completed and total values
+			// persisted by the scheduler's monitor jobs loop.
+			// Add an on update sql trigger to protect the job_run table, once progress
+			// values are used in the critical path.
 			rows, err := r.Query(ctx, failRunQuery, []interface{}{completed, total, runId})
```
vancluever, line 212: Same comment WRT to reader/writer usage here.

vancluever (approved):

PR #1399, opened 2021-07-15
test(db): Thin line for new sql testing suite

tmessi:
This sets up the configuration for running a new suite of tests that are
written in sql, to test the sql logic implemented in plpgsql functions
and triggers. It is primarily designed to provides tests for the data
warehouse functionality that is implemented purely in sql.

The infrasturture to run these tests leverages docker. It creates one
docker container using the `library/postgres` image and provisions the
database with the boundary sql migrations and test helper functions.
A separate docker container is run that has the pgtap extension and
pg_prove cli to execute the tests.

As noted, pgtap is used for the tests. This postgresql extension
provides a number of helpful assertion functions to make writting sql
tests easier. It also, along with the pg_prove cli, provides readable
test output.

This commit also provides an initial test for the wh_user_dimension to
establish an inital pattern for these kind of tests.

Lastly this configures circleci to run this test suite as a separate job
in parallel with the other tests.

See:
    https://pgtap.org/

### DIFF
--- a/.circleci/config/jobs/test-sql.yml
+++ b/.circleci/config/jobs/test-sql.yml
@@ -1,0 +1,9 @@
+machine:
+  image: 'ubuntu-1604:201903-01'
+working_directory: ~/boundary
+steps:
+- checkout
+- run:
+    name: "Run SQL PgTap Tests"
+    command: |
+      make test-sql
--- a/internal/db/sqltest/README.md
+++ b/internal/db/sqltest/README.md
@@ -1,0 +1,26 @@
+# SQL Tests
+
+This test suite is used to test behavior of the database logic.
+In particular the data warehouse implementation is
+completely in sql via plpgsql, functions, and triggers.
+This test suite is also implemented directly in sql.
+
+## Organization
+
+- `initdb.d`: contains init scripts/sql that is run on the test database
+ when it starts. It ensures the sql migrations are run and creates test helper
+ functions.
+- `tests`: contains the tests. Each file can contain only a single test that
+    runs in a transaction so it can rollback.
+
+The tests leverage the
+[pgTap](https://pgtap.org/documentation.html)
+postgres extension to make assertions and provide
+more readable test output.
+
+## Usage
+
+To run the test run `make` or `make test`. This will:
+
+- Start a docker `postgres` container and initialize it.
+- Start a docker `pgtap` container to execute the tests.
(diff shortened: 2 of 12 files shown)


internal/db/sqltest/README.md
```diff
@@ -0,0 +1,26 @@
+# SQL Tests
```
mgaffney, line 1: Thanks for adding this README. Very helpful!

mgaffney (approved):
Looks great!

.circleci/config/jobs/test-sql.yml
```diff
@@ -0,0 +1,9 @@
+machine:
+  image: 'ubuntu-1604:201903-01'
```
jefferai, line 2: Not that it necessarily matters but any reason to use such an old machine image? LTS is reasonable but why not 20.04?

internal/scheduler/additional_verification_test.go
```diff
@@ -338,3 +339,102 @@ func TestSchedulerMonitorLoop(t *testing.T) {
 	require.NoError(err)
 	assert.Equal(string(job.Interrupted), run.Status)
 }
+
+func TestSchedulerFinalStatusUpdate(t *testing.T) {
```
jefferai, line 343: Did this come in from merging in from `main` or something? Or is this relevant to the current PR in a way I'm not seeing?

It's not a problem if it's the former, but _if so_ you may want to get in the habit of doing a `main` -> `base long-term branch` -> `feature branch` merge instead of merging directly from `main` to `feature branch` to keep PRs clean.

jefferai (commented):
A couple of comments, more questions than anything else. As a non-DB expert this looks good to me!

internal/scheduler/additional_verification_test.go
tmessi: Yes this is from merging `main`. I will cleanup this PR to follow that pattern of `main` -> `long-term` to clean this up.

.circleci/config/jobs/test-sql.yml
tmessi: I picked this to just follow what the [existing jobs was using](https://github.com/hashicorp/boundary/blob/tmessi-wh-test-baseline/.circleci/config/jobs/build.yml) to be consistent.

internal/scheduler/additional_verification_test.go
tmessi: cleaned the PR up

jimlambrt (approved):
Ty!  Really amazing work and a good foundation.

PR #1400, opened 2021-07-16
Apply extra review feedback from #1340

vancluever:
This PR applies review feedback made by @jefferai for #1340.

### DIFF
--- a/internal/session/query.go
+++ b/internal/session/query.go
@@ -291,59 +291,30 @@ with
 	// The query returns the set of servers that have had connections closed
 	// along with their last update time and the number of connections closed on
 	// each.
+
 	closeConnectionsForDeadServersCte = `
-with
-  -- Get dead servers, parameterized off of grace period in seconds
-  dead_servers as (
-    select private_id, update_time
-    from server
-    where update_time < wt_sub_seconds_from_now($1)
-  ),
-  -- Find connections that are not closed so we can reference those IDs
-  unclosed_connections as (
-    select connection_id
-      from session_connection_state
-    where
-      -- It's the current state
-      end_time is null
-        and
-      -- Current state isn't closed state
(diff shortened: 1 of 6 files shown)


mgaffney (approved):

internal/session/query.go
```diff
@@ -334,9 +334,9 @@ with
       returning public_id, server_id
   )
   select
-    dead_servers.private_id,
-    dead_servers.update_time,
-    count(closed_connections.public_id)
+    dead_servers.private_id as server_id,
```
jefferai, line 337: These got named via `as` but those names aren't actually then used in the statement below.

jefferai (changes requested):
Looks good -- just the one thing to change in the SQL query. Seems like either the `as` statements can be reverted or the rest of the select clause should be updated to use those values.

internal/session/query.go
vancluever: @jefferai this was due to fbccee6df93743d1a77c03d1da32b3f6c53e57f3. Unless there's a better way to make these results available in a struct thru gorm?

This was done after https://github.com/hashicorp/boundary/pull/1400/commits/1ee29688da118da6d622b72024701845de6b5a44.

vancluever: In turn related to https://github.com/hashicorp/boundary/pull/1340#discussion_r669944414

jefferai: There's no issue with having names assigned...I'm simply calling out that since they are being given a name, other parts of the query could use the name to simplify things.

vancluever:
@jefferai Now that @mgaffney has merged #1410, I think the other review comment is resolved?

jefferai (approved):

jefferai:
Yep!

PR #1401, opened 2021-07-16
Remove Telemetry from config tests

louisruch:
As @tmessi reported in slack CI builds on `main` are currently failing:

```
Tim Messier Today at 07:50
It looks like ed6c59a0539ecfba54835cd3c04b04f709303c60 is causing CI to fail on main (and any new branch since that commit). It updated the version of github.com/hashicorp/go-secure-stdlib/configutil which has this commit: 1b80d53b4662d4b15ea0c23754dd81e3ae21d58b that removes configutil.Telemetry but the tests in internal/cmd/config are using that struct so the tests fail to compile
```

Commit https://github.com/hashicorp/boundary/commit/9ba1948298b8d0d57919fe480b41c9aeb97fa03b removed telemetry as it was not being used but config_tests still used it, but it is only failing now that it has been removed from go-secure-stdlib

### DIFF
--- a/internal/cmd/base/servers.go
+++ b/internal/cmd/base/servers.go
@@ -18,12 +18,11 @@ import (
 	"time"
 
 	"github.com/armon/go-metrics"
-	berrors "github.com/hashicorp/boundary/internal/errors"
-
 	"github.com/hashicorp/boundary/globals"
 	"github.com/hashicorp/boundary/internal/cmd/base/logging"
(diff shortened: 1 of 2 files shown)


louisruch:
@tmessi @mgaffney PTAL 
If @jefferai sees this while out PTAL

tmessi:
This looks good to me and makes sense. It should fix CI builds, but I'll defer to others to confirm that this is the correct solution here.

jefferai (approved):

PR #1402, opened 2021-07-16
internal/servers/worker: fix TestWorkerWaitForNextSuccessfulStatusUpdate

vancluever:
This fixes this test which was failing intermittently.

* Ups the grace period timeout to 2s.
* Re-positions the testify require.New() so that incorrect code path
  warnings are not given.


louisruch (approved):
LGTM thanks!

PR #1405, opened 2021-07-19
web: migrate to web platform- packages

BRKalow:
Migrate away from @hashicorp/nextjs-scripts to the new suite of @hashicorp/platform-* packages

[_Created by Sourcegraph campaign `bkalow/migrate-nextjs-scripts-packages`._](https://sourcegraph.hashi-mktg.com/users/bkalow/campaigns/migrate-nextjs-scripts-packages)


website/components/subnav/index.jsx
```diff
@@ -27,6 +27,7 @@ export default function ProductSubnav() {
       menuItems={subnavItems}
       constrainWidth
       Link={Link}
+      matchOnBasePath
```
jescalan, line 30: Still curious about this one - is there any place it's not needed?

jescalan (approved):

website/pages/downloads/index.jsx
```diff
@@ -11,7 +11,7 @@ export default function DownloadsPage({ binaryReleases, desktopReleases }) {
   return (
     <div className={styles.root}>
       <HashiHead title={`Downloads | ${productName} by HashiCorp`} />
-      <ProductDownloader
+      <ProductDownloadsPage
```
jescalan, line 14: This needs wider changes - the head section and wrapper are included in the page version of the component

jescalan (changes requested):
Sorry I approved too quick - there are more adjustments needed to change over to the `ProductDownloadsPage` component

website/components/subnav/index.jsx
BRKalow: All the subnavs in www don't need it.

website/pages/downloads/index.jsx
BRKalow: Thanks, good looking out! I left out the usage of the server helper as we are also fetching for boundary desktop here in parallel.

jescalan: Still can have the wrapping div, styles, and extra head component removed, as those are all repeated 😬

BRKalow: Oh wow I think I have a local commit that I didn't push up 🤦 one moment...

website/pages/downloads/index.jsx
```diff
@@ -1,103 +1,99 @@
...
+        {
+          label: 'Amazon Linux',
+          commands: [
+            'sudo yum install -y yum-utils',
+            'sudo yum-config-manager --add-repo https://rpm.releases.hashicorp.com/AmazonLinux/hashicorp.repo',
+            'sudo yum -y install boundary',
+          ],
+          os: 'linux',
+        },
+      ]}
```
jescalan, line 58: Unless these commands are different than the standard set on purpose, they can be omitted, they are built in to the core

jescalan (changes requested):
Sorry one more small thing with the downloads page

jescalan (approved):
👏

PR #1408, opened 2021-07-21
test(db): Add baseline tests for wh_user_dimension

tmessi:
This creates two kinds of tests.

- Unit tests for individual views and functions:
  - whx_user_dimension_source
  - whx_user_dimension_target
  - wh_upsert_user
- Integration tests for that verify that the wh_user_dimension gets
 populated correctly when sessions are inserted or updated. This
 ensures that:
   - the trigger runs on session insert
   - the trigger results in the correct changes to the
     wh_user_dimension.

---

test(db): Setup matrix builds for sql tests in circleci

This refactors the sqltest Makefile to support:

- Setting the postgres docker image to easily test multiple versions.
- Passing through options to pg_prove
- Passing the test list to run individual tests
- Additional targets to separately startup the database image to quickly
re-run tests while developing new tests.

It then re-configures circleci to run the sqltests against multiple
versions of postgres:

- latest
- 13-alpine
- 12-alpine
- 11-alpine


.circleci/config.yml
```diff
@@ -5,6 +5,16 @@
 # Orb 'circleci/slack@3.4.2' resolved to 'circleci/slack@3.4.2'
 version: 2
 jobs:
+  test-sql-13-alpine:
```
mgaffney, line 8: This test matrix is great! Thanks for adding it!

mgaffney (approved):
@tmessi This is amazing work!

PR #1409, opened 2021-07-21
Events: convert from hclog to events

jimlambrt:
This PR converts all the existing logging to events.  The PR shouldn't be merged until #1390 is merged and we're ready for events to be generally available.  

In that vein, this PR removes the BOUNDARY_DEVELOPER_ENABLE_EVENTS from events.WriteSysEvent(...), event.WriteObservation(...) and event.WriteError(...) but it keeps it in place for WriteAudit(...) which means audit events will not be enabled by this PR.

Contributors which relied on trace, debug and warn logging for instrumentation are encouraged to review the conversion of their logging contributions to events.

Note: the conversion from logging to events means that log levels are not longer a thing in Boundary and operators will need to use event filtering to control which events are emitted.


mgaffney (dismissed):
Nice work!

vancluever (dismissed):
Looks good! Also great to know what logging is going to look like going forward :+1:

internal/observability/event/options.go
```diff
@@ -70,6 +71,13 @@ func WithFlush() Option {
 	}
 }
 
+// WithInfo allows an optional map as info
+func WithInfo(d map[string]interface{}) Option {
```
tmessi, line 75: Looking at the usage of this function and how it changes the existing code, this introduces some very long lines, which I find reduces the readability of the code.

For example compare these two lines:
```go
s.logger.Error("error creating job repo", "error", err)
event.WriteError(ctx, op, err, event.WithInfo(map[string]interface{}{"msg": "error creating job repo"}))
```

In particular I think the `map[string]interface{}` is distracting. Perhaps a type to shorten this:

```go
// in event/options.go
type Info map[string]interface{}

// in scheduler/scheduler.go
event.WriteError(ctx, op, err, event.WithInfo(event.Info{"msg": "error creating job repo"}))
```

Or even shorter:

```go
// in event/options.go
type I map[string]interface{}

// in scheduler/scheduler.go
event.WriteError(ctx, op, err, event.WithInfo(event.I{"msg": "error creating job repo"}))
```

I also notice that `options.withDetails` and `options.WithHeader` are also `map[string]interface{}`. Defining types could help clarify if there are any conceptual differences between them:

```go
// separate types to be explicit and make it clear that Details are different
// from Info
type Info map[string]interface{}
type Detail map[string]interface{}
type Header map[string]interface{}

// or a shorter version
type I map[string]interface{}
type D map[string]interface{}
type H map[string]interface{}

// a single type to show that they are all the same kv data, and it is the With
// functions that provide the distinction.
type KeyValue map[string]interface{}
// or perhaps
type KV map[string]interface{}
```

Also, it looks like there are a lot of places where the `Info` is just a single key of `msg`. It might be useful to have a conience function for this:

```go
// in event/options.go
func WithInfoMsg(msg string) Option {
    return WithInfo(map[string]interface{
      "msg": msg,
    })
}

// in scheduler/scheduler.go
event.WriteError(ctx, op, err, event.WithInfoMsg("error creating job repo"}))
```

That said, this might be a conversation for a separate PR, since it would impact code that isn't changing in this PR.

internal/db/db.go
```diff
@@ -47,11 +50,15 @@ func Open(dbType DbType, connectionUrl string) (*gorm.DB, error) {
...
 
 func GetGormLogFormatter(log hclog.Logger) func(values ...interface{}) (messages []interface{}) {
+	const op = "db.GetGormLogFormatter"
+	ctx := context.TODO()
 	return func(values ...interface{}) (messages []interface{}) {
 		if len(values) > 2 && values[0].(string) == "log" {
 			switch values[2].(type) {
 			case *pq.Error:
-				log.Trace("error from database adapter", "location", values[1], "error", values[2])
+				if log.IsTrace() {
```
tmessi, line 59: In the PR message you say:

> Note: the conversion from logging to events means that log levels are not longer a thing in Boundary and operators will need to use event filtering to control which events are emitted.

How does that impact this `log.IsTrace()`?

internal/libs/alpnmux/mux.go
```diff
@@ -110,13 +113,15 @@ func (l *ALPNMux) RegisterProto(proto string, tlsConf *tls.Config) (net.Listener
 	}
 
 	if l.log != nil && l.log.IsDebug() {
-		l.log.Debug("registered", "proto", proto)
+		event.WriteSysEvent(ctx, op, map[string]interface{}{"msg": "registered", "proto": proto})
```
tmessi, line 116: Similar question here, how is the conditional above impacted by there being no log levels?

tmessi (commented):
Looks really good. I do have a few questions and comments.

internal/observability/event/options.go
```diff
@@ -70,6 +71,13 @@ func WithFlush() Option {
 	}
 }
 
+// WithInfo allows an optional map as info
```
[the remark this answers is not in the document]
jimlambrt: I'm happy to add a bit more docs... but they are only used in WriteObservation(...) where there usage is fairly explicit.  If it's not explict there then I need to add more docs there def.

internal/observability/event/options.go
jimlambrt: Hum... I'd like to incorporate any changes to these options now.. since they would ripple across all the code base where ever they're used. 

I like the `WithInfoMsg(...)` suggestion and will add that.  ty!

I'm not sure adding `KV` adds any clarity.

Pondering the whole set of types for `I, D, H`.

louisruch: `WithInfoMsg` does seem to be a convenient helper if we foresee that most Info maps will be a single key of `msg`

louisruch (approved):
This is great!

internal/observability/event/options.go
jimlambrt: I've adopted your suggestions about `I, D, H` and I've include using `I` in `WriteSysEvent(...)` as well.  ty!

internal/observability/event/options.go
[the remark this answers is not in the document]
tmessi: Just based on the names, "Info" and "Detail" feel a bit synonymous. I see that the doc for `WriteObersvation` states:

> At least one and any combination of the supported options may be used:
> WithHeader, WithDetails, WithId, WithFlush and WithRequestInfo. All other
> options are ignored.

Which means for `WriteObservation` `WithDetails` should be used and not `WithInfo`. 

The doc for `WriteError` should probably be updated since it currently says:

> The options WithId and WithRequestInfo are supported and all other options
> are ignored.

But based on the code changes, it seems like `WithInfo` is being used in places where `event.WriteError` is called. 

So it isn't clear to me:
- What the difference between `WithInfo` and `WithDetails` is.
- Why an Observation can take a `Detail`, but not `Info`.
- Why an Error can take `Info` but not `Detail`.

Documenting this on the `WriteError` and `WriteObservation` would be fine. Also if we do go with specific types instead of `map[string]interface` it might be a good spot to document the difference between a `Detail` and an `Info`.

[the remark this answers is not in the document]
tmessi: It looks like you made all of those changes while I was typing that response.

internal/db/db.go
jimlambrt: `GetGormLogFormatter` is specically about specifying a logger for gorm and the only current single available is the level of logging in the logger.

internal/libs/alpnmux/mux.go
jimlambrt: Here is another instance of where the only current signal is coming from an embedded logger in the receiver so at this point there's not much to do but allow it to provide a signal

internal/observability/event/options.go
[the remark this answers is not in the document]
jimlambrt: Are you all set here then?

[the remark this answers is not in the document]
tmessi: Yep

internal/cmd/commands/server/server.go
```diff
@@ -645,7 +647,7 @@ func (c *Command) WaitForInterrupt() int {
 		case <-c.SigUSR2Ch:
 			buf := make([]byte, 32*1024*1024)
 			n := runtime.Stack(buf[:], true)
-			c.Logger.Info("goroutine trace", "stack", string(buf[:n]))
+			event.WriteSysEvent(context.TODO(), op, map[string]interface{}{"msg": "goroutine trace", "stack": string(buf[:n])})
```
tmessi, line 650: I am seeing several of spots like this that are still using `map[string]interface{}` instead of `event.I`, `event.H`, or `event.D`.

jimlambrt: ty!

internal/auth/auth.go
```diff
@@ -254,7 +257,21 @@ func Verify(ctx context.Context, opt ...Option) (ret VerifyResults) {
 		if v.requestInfo.DisableAuthzFailures {
 			ret.Error = nil
 			// TODO: Decide whether to remove this
-			v.logger.Info("failed authz info for request", "resource", pretty.Sprint(v.res), "user_id", ret.UserId, "action", v.act.String())
+			err := event.WriteObservation(ctx, op, event.WithHeader(map[string]interface{}{
```
tmessi, line 260: `map[string]interface{}` -> `event.H`

internal/cmd/base/servers.go
```diff
@@ -711,21 +717,23 @@ func (s *Server) SetStatusGracePeriodDuration(value time.Duration) {
 		v := os.Getenv(statusGracePeriodEnvVar)
 		n, err := strconv.Atoi(v)
 		if err != nil {
-			s.Logger.Error(fmt.Sprintf("could not read setting for %s", statusGracePeriodEnvVar),
-				"err", err,
-				"value", v,
-			)
+			event.WriteError(ctx, op, err, event.WithInfo(
+				map[string]interface{}{
```
tmessi, line 721: `map[string]interface{}` -> `event.I`

internal/servers/controller/handlers/workers/worker_service.go
```diff
@@ -403,16 +412,20 @@ func (ws *workerServiceServer) AuthorizeConnection(ctx context.Context, req *pbs
 		ret.ConnectionsLeft -= int32(authzSummary.CurrentConnectionCount)
 	}
 
-	ws.logger.Info("authorized connection",
-		"session_id", req.GetSessionId(),
-		"connection_id", ret.ConnectionId,
-		"connections_left", ret.ConnectionsLeft)
+	event.WriteSysEvent(ctx, op,
+		map[string]interface{}{
```
tmessi, line 416: `map[string]interface{}` -> `event.I`

internal/servers/controller/session_cleanup_job.go
```diff
@@ -104,30 +103,28 @@ func (j *sessionCleanupJob) Run(ctx context.Context) error {
...
 			j.totalClosed += result.NumberConnectionsClosed
 		}
 	}
 
-	j.logger.Debug(
-		"job finished",
-		"op", op,
-		"total_connections_closed", j.totalClosed,
+	event.WriteSysEvent(ctx, op,
+		map[string]interface{}{
```
tmessi, line 123: `map[string]interface]{}` -> `event.I`

tmessi (commented):
These are the last few that I can find.

internal/auth/auth.go
jimlambrt: ty

internal/cmd/base/servers.go
jimlambrt: ty!

internal/servers/controller/handlers/workers/worker_service.go
jimlambrt: ty!

internal/servers/controller/session_cleanup_job.go
jimlambrt: ty!

internal/observability/event/options.go
jefferai: A lot of libs that take in this kind of data use a `...args interface{}` kind of pattern that expect a key, value, key, value kind of pattern, and check that it matches and that the keys are strings. `hclog` itself is an example of this. So another possibility is something like `event.WithInfo("mykey", 1, "other key", false)`

jimlambrt:
@louisruch @vancluever @jefferai -> would you mind reviewing commit 3c1a3bd 

I took the liberty of deleting some converted to system events and I believe you're the authors of the original debug logging.  ty!

internal/auth/auth.go
```diff
@@ -254,7 +257,22 @@ func Verify(ctx context.Context, opt ...Option) (ret VerifyResults) {
 		if v.requestInfo.DisableAuthzFailures {
 			ret.Error = nil
 			// TODO: Decide whether to remove this
-			v.logger.Info("failed authz info for request", "resource", pretty.Sprint(v.res), "user_id", ret.UserId, "action", v.act.String())
+			err := event.WriteObservation(ctx, op,
+				event.WithHeader(
```
jefferai, line 261: Eventually this is turned into JSON. However, because this is an inline struct that does not specify JSON tags, the eventual keys will be `Msg`, `UserId`, etc. (https://play.golang.org/p/1OO7GTs2EC9). Unless that is the preferred format for CloudEvents then probably this needs to have JSON tags specified as our general JSON format is lower snake case.

internal/cmd/base/dev.go
```diff
@@ -29,7 +31,7 @@ func (b *Server) CreateDevDatabase(ctx context.Context, opt ...Option) error {
 	// We should only get back postgres for now, but laying the foundation for non-postgres
 	switch opts.withDialect {
 	case "":
-		b.Logger.Error("unsupported dialect. wanted: postgres, got: %v", opts.withDialect)
+		event.WriteError(ctx, op, err, event.WithInfoMsg(fmt.Sprintf("unsupported dialect. wanted: postgres, got: %v", opts.withDialect)))
```
jefferai, line 34: Probably this should be something like
`event.WithInfoMsg("unsupported dialect", "wanted", "postgres", "got", opts.withDialect`

internal/cmd/base/servers.go
```diff
@@ -711,21 +717,16 @@ func (s *Server) SetStatusGracePeriodDuration(value time.Duration) {
 		v := os.Getenv(statusGracePeriodEnvVar)
 		n, err := strconv.Atoi(v)
 		if err != nil {
-			s.Logger.Error(fmt.Sprintf("could not read setting for %s", statusGracePeriodEnvVar),
-				"err", err,
-				"value", v,
-			)
+			event.WriteError(ctx, op, err, event.WithInfoMsg(fmt.Sprintf("could not read setting for %s", statusGracePeriodEnvVar), "value", v))
```
jefferai, line 720: Maybe `"WithInfoMsg("could not read status grace period setting", "envvar", statusGracePeriodEnvVar, "value", v)`

internal/observability/event/context.go
```diff
@@ -267,28 +266,61 @@ func WriteSysEvent(ctx context.Context, caller Op, data map[string]interface{})
...
+		args = append(args[:len(args)-1], MissingKey, extra)
+	}
+
+	m := map[string]interface{}{}
+	for i := 0; i < len(args); i = i + 2 {
+		var key string
+		switch st := args[i].(type) {
+		case string:
+			key = st
+		default:
```
jefferai, line 320: Curious if you think this is what we want, as opposed to forcing it to be a string. OTOH I guess you're trying to avoid having to return an error or panic...

internal/observability/event/context.go
```diff
@@ -267,28 +266,61 @@ func WriteSysEvent(ctx context.Context, caller Op, data map[string]interface{})
...
 	}
 }
+
+// MissingKey defines a key to be used as the "missing key" when ConvertArgs has
+// an odd number of args (it's missing a key in its key/value pairs)
+const MissingKey = "EXTRA_VALUE_AT_END"
+
+// ConvertArgs will convert the key/value pair args to a map.  If the args
+// provided are an odd number (they're missing a key in their key/value pairs)
+// then MissingKey is used to the missing key.
```
tmessi, line 304: Interesting, I think I prefer the previous pattern with the `map[string]interface` (and the short hand types of `I`, `D`, H`). This approach of taking `args ...interface{}` for something that needs to be a map seems error prone. With the `map[string]interface` it is enforced at compile time by the type. With this approach you either need to have an error, panic, or this `MissingKey` approach which would likely not be caught until much later after a bug is introduced.

internal/servers/worker/controller_connection.go
```diff
@@ -34,7 +36,6 @@ func (w *Worker) startControllerConnections() error {
 		default:
 			host, port, err := net.SplitHostPort(addr)
 			if err != nil && strings.Contains(err.Error(), "missing port in address") {
-				w.logger.Trace("missing port in controller address, using port 9201", "address", addr)
```
jefferai, line 37: Did you choose to elide this instead of convert it on purpose? Not saying it _needs_ to be something published but also not sure if it was just an oversight.

jefferai (dismissed):
Looks good! I pushed a couple of commits after discussing with you so you may want to give those a once over, and I there's one or two outstanding questions/comments. Really fantastic work!

internal/observability/event/context.go
jimlambrt: Honestly, I like the `args ...interface{}` better and it aligns with the existing hclog pattern most devs coming to boundary will be familiar with for sort of key/value set of parameters

jimlambrt: I do agree that there's more chance for a dev to not pass in key/value but I also don't think it's a huge problem if they do.  The data will still find its way into an event.

internal/observability/event/context.go
jimlambrt: I do think this is a reasonable default and eliminates the need for an error or panic.

internal/servers/worker/controller_connection.go
jimlambrt: It was intentional and part of commit 3c1a3bd

jefferai (dismissed):

internal/observability/event/context.go
louisruch: I agree with the concerns of @tmessi , but I am a fan of the `args ...interface{}` and am happy with how this PR handles MissingKey is correct.

louisruch:
> @louisruch @vancluever @jefferai -> would you mind reviewing commit [3c1a3bd](https://github.com/hashicorp/boundary/commit/3c1a3bdd722fc8a5a2f48f89130826f28b3d2697)
> 
> I took the liberty of deleting some converted to system events and I believe you're the authors of the original debug logging. ty!

I am good with the log lines you have deleted

louisruch (commented):
Seems to be a number of test failures, otherwise LGTM

internal/observability/event/context.go
tmessi: The `MissingKey` logic assumes that:

1. It is a key that is missing, and not a value.
2. The missing key is the last k/v pair

I think it is equally possible that a value is missing, or that a key/value is missing somewhere in the middle of the args, which would mess up all of the following pairs.

While this may not be the most critical part, these kinds of errors are non-issues with the `map[string]interface{}` since it is compile time.

louisruch: True, it is clearly a best effort solution, either way it should be fixed once detected (whether key or value). However, as @jimlambrt pointed out this is the existing pattern in `hclog` and the pattern Boundary is currently using for all logging. We could also consider adding a custom vet in the future to catch this during CI, something similar to what hclog has: https://github.com/hashicorp/go-hclog/tree/master/hclogvet

jimlambrt: @tmessi I understand the tradeoffs and assumptions in the proposed solution that uses `args... interface{}` when it comes to divining the callers intent if the key/value pairs have an off-by-one problem.

vancluever (dismissed):

jimlambrt:
Reviewers: I'm calling out a few specific commits for re-review:

- 4e17657 add waitgroups to manage shutdown of goroutines started by the controller and workers for things like tickers, jobs and status
- d9cb92b when sending events: detect ctx cancellation and change it to a ctx with a timeout. otherwise the event broker doesn't properly track success thresholds and we get false positive errors

jimlambrt:
@tmessi ty!  I've converted them to references.  

@louisruch I've refactored the remaining calls to Start(...) in the tests.  ty!

internal/servers/worker/worker.go
```diff
@@ -26,6 +27,8 @@ type Worker struct {
 	baseCancel  context.CancelFunc
 	started     *ua.Bool
 
+	tickerWg *sync.WaitGroup
```
tmessi, line 30: Similarly here, this can probably be `sync.WaitGroup` and `Start` and `Shutdown` can be updated to remove the `new` and nil check.

jefferai (approved):
Nice!!!

PR #1414, opened 2021-07-23
feat(wh): Add oidc related columns to wh_user_dimension

tmessi:
This adds new columns to the wh_user_dimension to provide additional
data from OIDC that does not apply to Password Auth.

Note: Given the specifics of the data, it is possible these columns may
be used by other auth methods in the future, hence the lack of `oidc` in
the column names.

The specific columns are:

- auth_method_external_id
    For OIDC this is the Issuer
- auth_account_external_id
    For OIDC this is the Subject
- auth_account_full_name
    For OIDC this is the account's Full Name, and can be `None`.
- auth_account_email
    For OIDC this is the account's Email, and can be `None`.

For all of these new columns, they will be `Not Applicable` if the auth
method is Password.

### DIFF
--- a/internal/db/schema/migrations/postgres/13/00_wh_user_dimension_oidc.up.sql
+++ b/internal/db/schema/migrations/postgres/13/00_wh_user_dimension_oidc.up.sql
@@ -1,0 +1,154 @@
+begin;
+
+  alter table wh_user_dimension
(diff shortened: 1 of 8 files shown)


mgaffney (approved):
Nice work!

PR #1415, opened 2021-07-23
Move base interfaces of the domain into a new boundary package

mgaffney:
This PR creates a new internal package named `boundary` and moves three interfaces from the `credential` package into it. The intent is to put the interfaces and types that define the Boundary domain into this package over time.

### DIFF
--- a/internal/boundary/boundary.go
+++ b/internal/boundary/boundary.go
@@ -1,0 +1,27 @@
+// Package boundary contains global interfaces and other definitions that
+// define the Boundary domain.
+package boundary
+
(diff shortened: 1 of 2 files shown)


talanknight (approved):
I like it.  It'll be interesting to see how this grows to include the other specific domain types, but even as is this helps define a common language and provides an interface to work with when we only need a specific abstraction.

louisruch (approved):
Neat!

jimlambrt (approved):
Nice work and an excellent choice of interfaces to begin with.

vancluever (approved):

tmessi (approved):

PR #1416, opened 2021-07-23
Subtype Id registration moved to its own package

talanknight:
Created a subtype registry which repo subpackages can now register to their outer package with. This removes the outer package from having to know about the inner package.

To avoid dependency cycles and to place the business logic closer to where it is actually used I moved the request auth specific stuff under a new auth package under server/controller sibling with handler package.  This is because this specific auth logic is specific to incoming requests on the API and how we authenticate and authorize them.  The top level auth package now solely focuses on the logic around managing the auth related resources.

I had intended for this PR to only be an example of what a refactoring would look like but things compounded and now the only package left to refactor after this PR is approved is target.


internal/types/subtypes/registry.go
```diff
@@ -0,0 +1,81 @@
...
+	r.subtypeMu.RLock()
+	subtype, ok := r.subtypesPrefixes[prefix]
+	r.subtypeMu.RUnlock()
+	if !ok {
+		return UnknownSubtype
+	}
+	return subtype
+}
+
+// Register registers all the prefixes for a provided Subtype. Register panics if the
```
jefferai, line 62: The locking in this file isn't fully consistent, and additionally panics are very much frowned upon at HC in application code. However -- is the intent (as I assume, from the rest of the code) that registration only happens at package init time? If so, I'd recommend documenting that as required behavior and just getting rid of the mutex. That would also make it "safer" to panic since it'd only happen at startup.

jefferai (commented):
Generally speaking this looks good, however see my comments about the new subtypes code.

jefferai:
Thanks for fixing the panics; I'm still not sold on the mutexes.

Do you expect that register will ever be called _not_ during startup via package initialization? If not, document that the package isn't thread safe and get rid of the mutexes.

If you expect it might be called at some other point we're likely still better off with a `sync.Map` as it's optimized for writing values once and then reading them lots of times:

> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

But I still think we're better off documenting a lack of thread safety as a feature not a bug.

internal/types/subtypes/registry.go
```diff
@@ -0,0 +1,87 @@
+package subtypes
```
mgaffney, line 1: Can you add a package level doc?

internal/types/subtypes/registry.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/hashicorp/boundary/internal/errors"
+)
+
+type Subtype string
```
mgaffney, line 11: Doc?

internal/types/subtypes/registry.go
```diff
@@ -0,0 +1,87 @@
...
+
+// Registry stores a collection of subtypes along with their prefixes and
+// allows for translating prefixes back to registered subtypes.
+type Registry struct {
+	subtypeMu        sync.RWMutex
+	subtypesPrefixes map[string]Subtype
+	knownSubtypes map[Subtype]interface{}
+}
+
+func NewRegistry() *Registry {
```
mgaffney, line 29: Doc?

mgaffney (changes requested):
Looks good. Just some comments regarding documentation for exported bits.

mgaffney (approved):

jefferai (approved):

PR #1419, opened 2021-07-26
Add token/userinfo claims to account read output

jefferai:
This is very useful when trying to write managed groups filters as you can directly look at the structure of the claims for the accounts you're trying to filter.

The output of the claims in text format on the CLI currently leaves something to be desired (see example below). I started revamping some of the formatting functions and it quickly got very fragile. Fundamentally at this point we're likely better off -- now that we have a good idea what our output format/standards are like -- rewriting the formatting code. I'll take that as a tasker as I have a lot of ideas...

```
  Attributes:
    email:             admin@localhost
    full_name:         Admin User
    issuer:            http://127.0.0.1:38453
    subject:           admin
    token_claims:
    {
    "at_hash": "EdDSA-hash",
    "aud": [
    "zziZuCmjAft9k4o17qok"
    ],
    "auth_time": 1627322774,
    "azp": "zziZuCmjAft9k4o17qok",
    "exp": 1627322779,
    "iat": 1627322774,
    "iss": "http://127.0.0.1:38453",
    "mode": "dev",
    "nbf": 1627322769,
    "nonce": "GXG5KJX2GjSHGIXB0jIy",
    "sub": "admin"
    }
    userinfo_claims:
    {
    "email": "admin@localhost",
    "name": "Admin User",
    "sub": "admin"
    }
```


internal/db/schema/migrations/postgres/13/01_oidc_add_claims.up.sql
```diff
@@ -0,0 +1,8 @@
+begin;
+
+alter table auth_oidc_account
+  add column token_claims text;
```
mgaffney, line 4: Is an empty string allowed? If not, can a check be added?

jefferai: Originally when I wrote this I thought empty string would be allowed; however it turns out in the code a nil map is not allowed so marshaling empty claims would actually turn into `{}`. So I can put a check in, yes.

talanknight (approved):
Looks good.

mgaffney (approved):

PR #1422, opened 2021-07-28
Fix race with base context cancel and final run update in scheduler

louisruch:
### What does this PR do

Skips final job run update if the base context is no longer valid.  All repo actions are dependent on the base context, attempting to perform an update without a valid context will always fail.

### Testing

The following error was found intermittently in [circle-ci](https://circleci.com/api/v1.1/project/github/hashicorp/boundary/71541/output/102/0?file=true&allocation-id=61004daab3df255367f81b9d-0-build%2F5F75C3D6):
```
testing.go:54: 
        	Error Trace:	testing.go:54
        	            				testing.go:901
        	            				testing.go:1007
        	            				testing.go:1187
        	            	write tcp 127.0.0.1:36706->127.0.0.1:33410: use of closed network connection
        	Test:       	TestSchedulerJobProgress
        	Messages:   	Got error closing gorm db.
--- FAIL: TestSchedulerJobProgress (11.85s)
```

Running the test in a loop locally would hit this error occasionally, in three attempts it hit the error after 34, 59 and 24 attempts.  Debug logging revealed the error was always hit in the final update attempt of a job_run that was racing with the context being cancelled in the unit tests. 

While this error would likely not cause production issues it would add confusing logs to the controller shutdown.

### DIFF
--- a/internal/scheduler/scheduler.go
+++ b/internal/scheduler/scheduler.go
@@ -228,23 +228,25 @@ func (s *Scheduler) runJob(ctx context.Context, r *job.Run) error {
 	jobContext, rj.cancelCtx = context.WithCancel(ctx)
 
 	go func() {
-		defer rj.cancelCtx()
 		runErr := j.Run(jobContext)
-		var updateErr error
+		// Job has ended, cancel job run specific context to clean up
+		rj.cancelCtx()
 
 		// Get final status report to update run progress with
 		status := j.Status()
-
-		switch runErr {
-		case nil:
+		var updateErr error
+		switch {
+		case ctx.Err() != nil:
+			// Base context is no longer valid, skip repo updates as they will fail and exit
+		case runErr == nil:
 			nextRun, inner := j.NextRunIn()
 			if inner != nil {
 				event.WriteError(ctx, op, inner, event.WithInfoMsg("error getting next run time", "name", j.Name()))
 			}
-			_, updateErr = repo.CompleteRun(jobContext, r.PrivateId, nextRun, status.Completed, status.Total)
+			_, updateErr = repo.CompleteRun(ctx, r.PrivateId, nextRun, status.Completed, status.Total)
 		default:
 			event.WriteError(ctx, op, runErr, event.WithInfoMsg("job run failed", "run id", r.PrivateId, "name", j.Name()))
-			_, updateErr = repo.FailRun(jobContext, r.PrivateId, status.Completed, status.Total)
+			_, updateErr = repo.FailRun(ctx, r.PrivateId, status.Completed, status.Total)
 		}
 
 		if updateErr != nil {
--- a/internal/scheduler/scheduler.go
+++ b/internal/scheduler/scheduler.go
@@ -163,7 +163,7 @@ func (s *Scheduler) Start(ctx context.Context, wg *sync.WaitGroup) error {
 	wg.Add(2)
 	go func() {
 		defer wg.Done()
-		s.start(ctx)
+		s.start(ctx, wg)
 	}()
 	go func() {
 		defer wg.Done()
@@ -173,7 +173,7 @@ func (s *Scheduler) Start(ctx context.Context, wg *sync.WaitGroup) error {
 	return nil
 }
 
-func (s *Scheduler) start(ctx context.Context) {
+func (s *Scheduler) start(ctx context.Context, wg *sync.WaitGroup) {
 	const op = "scheduler.(Scheduler).start"
 	timer := time.NewTimer(s.runJobsInterval)
 	for {
@@ -182,7 +182,6 @@ func (s *Scheduler) start(ctx context.Context) {
 			event.WriteSysEvent(ctx, op, "scheduling loop shutting down", "server id", s.serverId)
 			return
 		case <-timer.C:
-
 			repo, err := s.jobRepoFn()
 			if err != nil {
 				event.WriteError(ctx, op, err, event.WithInfoMsg("error creating job repo"))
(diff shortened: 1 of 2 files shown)


internal/scheduler/scheduler.go
```diff
@@ -228,23 +228,25 @@ func (s *Scheduler) runJob(ctx context.Context, r *job.Run) error {
...
+		switch {
+		case ctx.Err() != nil:
+			// Base context is no longer valid, skip repo updates as they will fail and exit
+		case runErr == nil:
 			nextRun, inner := j.NextRunIn()
 			if inner != nil {
 				event.WriteError(ctx, op, inner, event.WithInfoMsg("error getting next run time", "name", j.Name()))
 			}
-			_, updateErr = repo.CompleteRun(jobContext, r.PrivateId, nextRun, status.Completed, status.Total)
+			_, updateErr = repo.CompleteRun(ctx, r.PrivateId, nextRun, status.Completed, status.Total)
```
mgaffney, line 246: If `inner != nil` what will the value of `nextRun` be when this is called?

internal/scheduler/scheduler.go
```diff
@@ -228,23 +228,25 @@ func (s *Scheduler) runJob(ctx context.Context, r *job.Run) error {
 	jobContext, rj.cancelCtx = context.WithCancel(ctx)
 
 	go func() {
-		defer rj.cancelCtx()
 		runErr := j.Run(jobContext)
-		var updateErr error
+		// Job has ended, cancel job run specific context to clean up
+		rj.cancelCtx()
```
mgaffney, line 233: I think the original `defer rj.cancelCtx()` was cleaner and more idiomatic. Is the `defer` causing a problem?

internal/scheduler/scheduler.go
louisruch: Depends on the job implementation https://github.com/hashicorp/boundary/blob/main/internal/scheduler/job.go#L20

internal/scheduler/scheduler.go
louisruch: No it is not the issue.  I changed it back was on the fence about this change after I did it.  I did it because I wanted to be clear that the `jobContext` should not be used outside of the `Run`.

louisruch:
@jimlambrt this resolves the issue you pinged me about yesterday

internal/scheduler/scheduler.go
```diff
@@ -230,21 +230,22 @@ func (s *Scheduler) runJob(ctx context.Context, r *job.Run) error {
 	go func() {
 		defer rj.cancelCtx()
 		runErr := j.Run(jobContext)
-		var updateErr error
```
jimlambrt, line 233: Is there still a possible race condition here since we don't "wait" for the `j.Run(jobContext)`?  Can jobs be writing to the database when the goroutine ends?

jimlambrt (changes requested):

internal/scheduler/scheduler.go
louisruch: Not sure I follow?  Run is a blocking call once it returns the job has exited

jimlambrt: Run is blocking but it's invoked within a goroutine.  I'm just pondering if that coroutine's parent completes without waiting for it, does it have any affect on possible DB writes within the job that's being "Run".

louisruch: Aha I see, yeah it could.  Ill ping you when ready for review.

louisruch:
@jefferai as discussed 
@jimlambrt PTAL this extends the waitgroup you added to the scheduler to also wait on any running jobs.  Please note this does mean if a job ignores the context and does not exit early, we will block the controller shutdown.  I am ok with this as jobs should be using the context correctly.  We could consider adding some timeout at some point if needed

jefferai:
I'm slightly concerned about using the _same_ wait group. Maybe I shouldn't be, but:

> Note that calls with a positive delta that occur when the counter is zero must happen before a Wait. Calls with a negative delta, or calls with a positive delta that start when the counter is greater than zero, may happen at any time. Typically this means the calls to Add should execute before the statement creating the goroutine or other event to be waited for. If a WaitGroup is reused to wait for several independent sets of events, new Add calls must happen after all previous Wait calls have returned. See the WaitGroup example.

This will only be a problem if it's possible that we might launch a job *after* hitting Wait. Given that from what I can tell from the code the scheduler _itself_ also is a part of the group, then so long as the scheduler will not be able to have called Done then launch a job, everything is fine.

Otherwise this approach makes sense to me.

louisruch:
> I'm slightly concerned about using the _same_ wait group. Maybe I shouldn't be, but:
> 
> > Note that calls with a positive delta that occur when the counter is zero must happen before a Wait. Calls with a negative delta, or calls with a positive delta that start when the counter is greater than zero, may happen at any time. Typically this means the calls to Add should execute before the statement creating the goroutine or other event to be waited for. If a WaitGroup is reused to wait for several independent sets of events, new Add calls must happen after all previous Wait calls have returned. See the WaitGroup example.
> 
> This will only be a problem if it's possible that we might launch a job _after_ hitting Wait. Given that from what I can tell from the code the scheduler _itself_ also is a part of the group, then so long as the scheduler will not be able to have called Done then launch a job, everything is fine.
> 
> Otherwise this approach makes sense to me.

The scheduler calls done twice, once in a defer for `start` and once in a defer for `monitorJobs`.  `start` sits in a for-select performing two steps: waiting for ctx.Done() and a timer for running jobs.  The ctx.Done() will return from `start` eventually triggering the wg.Done() mentioned above.  However, all job level `wg.Add(1)` will always execute before returning to the for-select to check the ctx.Done().  Therefore, I am not concerned about the specific point you raise, however it is easy enough to spin up a new waitgroup that I can use within start and remove this concern entirely.  My main worry would be future development that could change my above statement.

jefferai (approved):
I think this looks good. I agree about the idea that a future change might have invalidated the previous waitgroup behavior so I think the change you made was probably the right thing to do.

jimlambrt (approved):

PR #1423, opened 2021-07-28
prevent panics when waitgroup == nil

jimlambrt:


internal/scheduler/scheduler_test.go
```diff
@@ -366,3 +367,47 @@ func TestScheduler_UpdateJobNextRunInAtLeast(t *testing.T) {
...
+			assert, require := assert.New(t), require.New(t)
+			err := sched.Start(tt.ctx, tt.wg)
+			if tt.wantErr {
+				require.Error(err)
+				if tt.wantErrContains != "" {
+					assert.Contains(err.Error(), tt.wantErrContains)
+				}
+				return
+			}
+			require.NoError(err)
```
louisruch, line 410: Perhaps add a `valid` test so this path is exercised

louisruch (dismissed):
Really minor comment in the test but otherwise LGTM

louisruch (approved):

PR #1426, opened 2021-07-28
Add session cleanup changes to CHANGELOG.md

vancluever:
This adds the changelog entries associated with the following PRs:

#1340 
#1354
#1400
#1410

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,21 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
   it significantly easier to write filters to create [managed
   groups](https://www.boundaryproject.io/docs/concepts/filtering/oidc-managed-groups).
   ([PR](https://github.com/hashicorp/boundary/pull/1419))
+* Controllers will now mark connections as closed in the database if the worker
+  has not reported its status; this can be seen as the controller counterpart to
+  the worker-side session cleanup functionality released in 0.4.0. As with the
+  worker, this timeout for this behavior is 15s. Additionally, in a split-brain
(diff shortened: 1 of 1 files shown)


jefferai (dismissed):
Looks good. One comment: in the first bullet point, the last sentence could be seen to mean that workers won't close connections _until_ the status update comes after connectivity is restored. Might be better to instead note that it will be cleaned up at that point, but will have already been closed.

vancluever:
I think I'm just going to strike that part as I think the extra context can be confusing, I can't seem to find a good way to explain it more without writing a book. 😜 I think just saying that the controller is now marking the connections closed like the worker does should be good enough.

CHANGELOG.md
```diff
@@ -30,6 +30,17 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
   it significantly easier to write filters to create [managed
   groups](https://www.boundaryproject.io/docs/concepts/filtering/oidc-managed-groups).
   ([PR](https://github.com/hashicorp/boundary/pull/1419))
+* Controllers will now mark connections as closed in the database if the worker
+  has not reported its status; this can be seen as the controller counterpart to
+  the worker-side session cleanup functionality released in 0.4.0. As with the
+  worker, this timeout for this behavior is 15s.
```
jimlambrt, line 36:
```suggestion
  worker, the timeout for this behavior is 15s.
```

jimlambrt (changes requested):
one small suggestion

jimlambrt (approved):

PR #1432, opened 2021-07-30
server: Add back controller shutdown on worker startup failure

vancluever:
This adds back the controller shutdown call on worker failure.

This was added in a15961e4 to address a panic discovered in testing, and
was mistakenly thought to be unnecessary. This, however, covers
multi-role scenarios where we have already started the controller, which
needs to be shut down gracefully if possible if we can't start the
worker.

A nil check has been added to ensure the panic does not resurface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,7 +99,9 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
   well as the error message)
   ([issue](https://github.com/hashicorp/boundary/issues/1305),
   [PR](https://github.com/hashicorp/boundary/pull/1384))
-
+* server: Fixed a regression in 0.4.0 that caused the controller to not shut
+  down on a multi-role server when there were issues starting the worker role.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -99,7 +99,9 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
   well as the error message)
   ([issue](https://github.com/hashicorp/boundary/issues/1305),
   [PR](https://github.com/hashicorp/boundary/pull/1384))
-
+* server: Fixed a regression in 0.4.0 that caused the controller to not shut
```
jefferai, line 102: It's actually not a regression...this was changed post 0.4.0 so this is just a bug fix for the panic!

jefferai (dismissed):
Update the changelog and 👍

CHANGELOG.md
vancluever: Oh! I think I misread our offline convo then... will fix!

PR #1434, opened 2021-08-01
Event log adpaters

jimlambrt:


internal/cmd/base/dev.go
```diff
@@ -289,7 +286,7 @@ func (b *Server) CreateDevOidcAuthMethod(ctx context.Context) error {
 		clientSecret := string(b.DevOidcSetup.clientSecret)
 
 		b.DevOidcSetup.testProvider = capoidc.StartTestProvider(
-			logger,
+			&oidcLogger{},
```
jefferai, line 289: Unless there's a reason not to, I suggest allowing a context to be set into oidcLogger (either via a New function or just directly) that can be used instead of TODO in its functions. The functions could check if it's `nil` and if not, use it.

jefferai: This does assume that the event system will do the right thing if the context being passed in is expired. If not, maybe leave as is :-)

jefferai (approved):
Just that one comment, seems good!

internal/cmd/base/dev.go
jimlambrt: It does the "right thing", so I'll make this change.  ty!

PR #1438, opened 2021-08-02
Make gen after make tools.

talanknight:
This also had the side effect of running make fmt on a few previously committed files.


jimlambrt (approved):
ty!

PR #1440, opened 2021-08-02
support new event formats of: hclog-text and hclog-json

jimlambrt:


internal/observability/event/node.go
```diff
@@ -11,17 +11,17 @@ import (
 	"github.com/hashicorp/go-bexpr"
 )
 
-// Node represents an eventlogger.Node which filters events based on allow and
+// cloudEventsFormatterFilter represents an eventlogger.cloudEventsFormatterFilter which filters events based on allow and
```
jefferai, line 14: Does it make sense to rename the file based on the renaming inside it? Just for organizational purposes?

jefferai (approved):
10/10, would observe again

internal/observability/event/node.go
jimlambrt: ty!

jefferai (approved):

internal/observability/event/hclog_formatter_node_test.go
```diff
@@ -0,0 +1,368 @@
...
+				return
+			}
+			require.NoError(err)
+			assert.NotNil(got)
+
+			assert.Equal(tt.jsonFormat, got.jsonFormat)
+
+			assert.Len(got.allow, len(tt.wantAllow))
+			for _, f := range got.allow {
+				assert.Contains(tt.wantAllow, f.raw)
```
s-christoff, line 360: Should these be flipped (we are checking `f.raw` for something matching `tt.wantAllow`)?

s-christoff (approved):
one minor comment but 🤙

internal/observability/event/hclog_formatter_node_test.go
jimlambrt: we're checking the list of all possible wants (a slice of strings) to see if a specific got string is present.

PR #1444, opened 2021-08-04
Normalize event configuration

jefferai:
This updates event configuration stanzas to more closely match behavior
for listeners and kmses:

* Singular "sink" instead of "sinks" (probably not the most correct
  option, but a consistent option, and one that makes sense for HCL)
* Allows sinks to have the type defined at the start of the block
* Moves type-specific configuration to a specific named block, which
  will make it easier to support more types in the future, especially if
  go-eventlogger gains plugin support -- the block can be directly
  passed as a `map[string]interface{}` to a plugin.

It also updates `sink_type` to be `type` (since we're in the sink block
already, and we don't use e.g. `sink_name`) and fixes an event URL
discrepancy with test workers.


jefferai:
Note: I'm not 100% sold on `type_config`. `config` is easier but may seem a bit odd to the user if they think that the other parameters, like filters, are already configuration. So this was an attempt to make it be clear that it's configuration _for a specific type_. Better nomenclature ideas welcome.

jimlambrt (changes requested):
I noticed an existing issue with detecting duplicate file sink names and opened a sub-PR  #1445.

tmessi:
> Note: I'm not 100% sold on `type_config`. `config` is easier but may seem a bit odd to the user if they think that the other parameters, like filters, are already configuration. So this was an attempt to make it be clear that it's configuration _for a specific type_. Better nomenclature ideas welcome.

Just brainstorming some alternatives:

- options
- settings

jimlambrt (dismissed):
Nice work!

tmessi (approved):

PR #1446, opened 2021-08-04
parse the duration string specified in an event file config into a time.Duration

jimlambrt:


internal/observability/event/sink_config.go
```diff
@@ -75,11 +75,12 @@ type StderrSinkTypeConfig struct{}
...
-	FileName       string        `hcl:"file_name"        mapstructure:"file_name"`        // FileName defines the file name for the sink
-	RotateBytes    int           `hcl:"rotate_bytes"     mapstructure:"rotate_bytes"`     // RotateByes defines the number of bytes that should trigger rotation of a FileSink
-	RotateDuration time.Duration `hcl:"rotate_duration"  mapstructure:"rotate_duration"`  // RotateDuration defines how often a FileSink should be rotated
-	RotateMaxFiles int           `hcl:"rotate_max_files" mapstructure:"rotate_max_files"` // RotateMaxFiles defines how may historical rotated files should be kept for a FileSink
+	Path              string        `hcl:"path"             mapstructure:"path"`             // Path defines the file path for the sink
+	FileName          string        `hcl:"file_name"        mapstructure:"file_name"`        // FileName defines the file name for the sink
+	RotateBytes       int           `hcl:"rotate_bytes"     mapstructure:"rotate_bytes"`     // RotateByes defines the number of bytes that should trigger rotation of a FileSink
```
mgaffney, line 80:
```suggestion
	RotateBytes       int           `hcl:"rotate_bytes"     mapstructure:"rotate_bytes"`     // RotateBytes defines the number of bytes that should trigger rotation of a FileSink
```

internal/observability/event/sink_config.go
```diff
@@ -75,11 +75,12 @@ type StderrSinkTypeConfig struct{}
...
+	RotateDuration    time.Duration `mapstructure:"rotate_duration"`                         // RotateDuration defines how often a FileSink should be rotated
+	RotateDurationHCL string        `hcl:"rotate_duration" json:"-"`                         // RotateDurationHCL defines hcl string version of RotationDuration
```
mgaffney, line 82:
```suggestion
	RotateDurationHCL string        `hcl:"rotate_duration" json:"-"`                         // RotateDurationHCL defines hcl string version of RotateDuration
```

mgaffney (approved):

mgaffney (approved):

internal/cmd/config/config.go
```diff
@@ -473,6 +473,15 @@ func parseEventing(eventObj *ast.ObjectItem) (*event.EventerConfig, error) {
 			s.StderrConfig = new(event.StderrSinkTypeConfig)
 		}
 
+		// parse the duration string specified in a file config into a time.Duration
+		if s.FileConfig != nil && s.FileConfig.RotateDurationHCL != "" {
+			var err error
+			s.FileConfig.RotateDuration, err = time.ParseDuration(s.FileConfig.RotateDurationHCL)
```
jefferai, line 479: We're not incredibly consistent here -- we should probably update other usages -- but can you use `parseutil.ParseDurationSecond` instead? This allows a bare value without a suffix to be treated as seconds instead of nanoseconds.

jefferai (changes requested):
See comment

PR #1455, opened 2021-08-11
feat(wh): Add oidc related columns to wh_user_dimension

tmessi:
This adds new columns to the wh_user_dimension to provide additional
data from OIDC that does not apply to Password Auth.

Note: Given the specifics of the data, it is possible these columns may
be used by other auth methods in the future, hence the lack of `oidc` in
the column names.

The specific columns are:

- auth_method_external_id
    For OIDC this is the Issuer
- auth_account_external_id
    For OIDC this is the Subject
- auth_account_full_name
    For OIDC this is the account's Full Name, and can be `None`.
- auth_account_email
    For OIDC this is the account's Email, and can be `None`.

For all of these new columns, they will be `Not Applicable` if the auth
method is Password.

### DIFF
--- a/internal/db/schema/migrations/postgres/14/01_wh_user_dimension_oidc.up.sql
+++ b/internal/db/schema/migrations/postgres/14/01_wh_user_dimension_oidc.up.sql
@@ -1,0 +1,171 @@
+begin;
+
+  alter table wh_user_dimension
+    add column auth_method_external_id  wh_dim_text,
+    add column auth_account_external_id wh_dim_text,
+    add column auth_account_full_name   wh_dim_text,
+    add column auth_account_email       wh_dim_text
+  ;
+
+  drop view whx_user_dimension_source;
+  create view whx_user_dimension_source as
+       select -- id is the first column in the target view
+              u.public_id                       as user_id,
+              coalesce(u.name, 'None')          as user_name,
+              coalesce(u.description, 'None')   as user_description,
+              coalesce(aa.public_id, 'None')    as auth_account_id,
+              case
+                   when apa.public_id is not null then 'password auth account'
+                   when aoa.public_id is not null then 'oidc auth account'
+                   else 'None'
+                   end                          as auth_account_type,
+              case
+                   when apa.public_id is not null then coalesce(apa.name, 'None')
+                   when aoa.public_id is not null then coalesce(aoa.name, 'None')
+                   else 'None'
+                   end                          as auth_account_name,
+              case
+                   when apa.public_id is not null then coalesce(apa.description, 'None')
+                   when aoa.public_id is not null then coalesce(aoa.description, 'None')
+                   else 'None'
+                   end                          as auth_account_description,
+              case
+                  when apa.public_id is not null then 'Not Applicable'
+                  when aoa.public_id is null then 'None'
+                  else aoa.subject
+                  end                           as auth_account_external_id,
+              case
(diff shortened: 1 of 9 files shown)


internal/db/schema/migrations/postgres/14/01_wh_user_dimension_oidc.up.sql
```diff
@@ -0,0 +1,154 @@
...
+              coalesce(u.name, 'None')          as user_name,
+              coalesce(u.description, 'None')   as user_description,
+              coalesce(aa.public_id, 'None')    as auth_account_id,
+              case
+                   when apa.public_id is not null then 'password auth account'
+                   when aoa.public_id is not null then 'oidc auth account'
+                   else 'None'
+                   end                          as auth_account_type,
+              coalesce(apa.name, 'None')        as auth_account_name,
+              coalesce(apa.description, 'None') as auth_account_description,
```
mgaffney, line 23: I think the `auth_oidc_account` table also has name and description columns. If so, these need to be `case` statements.

internal/db/schema/migrations/postgres/14/01_wh_user_dimension_oidc.up.sql
```diff
@@ -0,0 +1,154 @@
...
+                  else 'None'
+                  end                           as auth_account_email,
+              coalesce(am.public_id, 'None')    as auth_method_id,
+              case
+                   when apa.public_id is not null then 'password auth method'
+                   when aoa.public_id is not null then 'oidc auth method'
+                   else 'None'
+                   end                          as auth_method_type,
+              coalesce(apm.name, 'None')        as auth_method_name,
+              coalesce(apm.description, 'None') as auth_method_description,
```
mgaffney, line 48: The `auth_oidc_method` table has `name` and `description` columns so these should be `case` statements too.

internal/db/schema/migrations/postgres/14/01_wh_user_dimension_oidc.up.sql
```diff
@@ -0,0 +1,154 @@
...
+              case
+                  when apa.public_id is not null then 'Not Applicable'
+                  when aoa.public_id is null then 'None'
+                  else aoa.issuer
```
mgaffney, line 52: This should be the `issuer` from the `auth_oidc_method` table.

internal/db/schema/migrations/postgres/14/01_wh_user_dimension_oidc.up.sql
```diff
@@ -0,0 +1,154 @@
...
+                  end                           as auth_method_external_id,
+              org.public_id                     as user_organization_id,
+              coalesce(org.name, 'None')        as user_organization_name,
+              coalesce(org.description, 'None') as user_organization_description
+         from iam_user as u
+    left join auth_account as aa on           u.public_id = aa.iam_user_id
+    left join auth_method as am on            aa.auth_method_id = am.public_id
+    left join auth_password_account as apa on aa.public_id = apa.public_id
+    left join auth_password_method as apm on  am.public_id = apm.public_id
+    left join auth_oidc_account as aoa on     aa.public_id = aoa.public_id
```
mgaffney, line 62: Need a join for `auth_oidc_method`

mgaffney (changes requested):

mgaffney (dismissed):

mgaffney (approved):

PR #1458, opened 2021-08-11
refactor(worker): Move tcp_proxy into its own package

louisruch:
This refactors:

- the tcp_proxy.go from the worker package into worker/proxy/tcp
- adds a missing unit test for the HandleTcpProxyV1
- refactors session.go from the worker package into worker/session, needed to avoid import cycle
- adds a baseline for a mock controller session client
- go lint fixes


louisruch:
@jefferai @vancluever this is a refactor of the worker code to break the actual proxy handle into its own package, there will probably be more to come but mainly wanted to get tcp_proxy into its own package for now as this will allow us to add new proxies types moving forward smoothly.  

Please read description for full details of what this PR does - also note I fixed some lint issues, removed commented out code and added a unit test.  In the future I will try and separate these actions into different commits or PRs to make review easier.

internal/servers/worker/package_registration.go
```diff
@@ -0,0 +1,5 @@
+package worker
+
+import (
+	_ "github.com/hashicorp/boundary/internal/servers/worker/proxy/tcp"
```
louisruch, line 4: I need to side effects import here to trigger the tcp proxy init.  I am open to discussing where this belongs, I figured worker package is best location because then it only builds if we build a worker...

internal/servers/worker/handler.go
```diff
@@ -89,73 +91,86 @@ func (w *Worker) handleProxy() http.HandlerFunc {
...
 		defer conn.Close(websocket.StatusNormalClosure, "done")
 
-		connCtx, connCancel := context.WithDeadline(r.Context(), expiration.AsTime())
+		connCtx, connCancel := context.WithDeadline(ctx, expiration.AsTime())
 		defer connCancel()
 
+		sessClient, err := w.ControllerSessionConn()
+		if err != nil {
+			event.WriteError(ctx, op, err)
+			_ = conn.Close(websocket.StatusInternalError, "unable to get controller session client")
```
jefferai, line 100: Any reason not to write the close error as an event if we hit one?

jefferai: Same question in the blocks below

internal/servers/worker/handler.go
```diff
@@ -165,114 +180,18 @@ func (w *Worker) handleProxy() http.HandlerFunc {
...
+func (w *Worker) wrapGenericHandler(h http.Handler, _ HandlerProperties) http.Handler {
 	return http.HandlerFunc(func(wr http.ResponseWriter, r *http.Request) {
 		// Set the Cache-Control header for all responses returned
 		wr.Header().Set("Cache-Control", "no-store")
 		h.ServeHTTP(wr, r)
 	})
 }
-
-/*
-func WrapForwardedForHandler(h http.Handler, authorizedAddrs []*sockaddr.SockAddrMarshaler, rejectNotPresent, rejectNonAuthz bool, hopSkips int) http.Handler {
```
jefferai, line 191: Hah, you finally took this out. We may need to add it back at some point, but we can deal with it then.

internal/servers/worker/handler.go
louisruch: Not specifically this was more a linting exercise to explicitly deal with the unhandled error. Can update to write all these now that we have events

internal/servers/worker/handler.go
louisruch: Yeah I was not sure if this was needed down the road, but I am not a fan of commented out code.  Happy to put back if we will need it in the future

internal/servers/worker/package_registration.go
jefferai: LGTM

internal/servers/worker/worker.go
```diff
@@ -211,3 +218,110 @@ func (w *Worker) ParseAndStoreTags(incoming map[string][]string) {
...
+	if sessClient == nil {
+		return nil, fmt.Errorf("session client is nil")
+	}
+
+	return sessClient, nil
+}
+
+func (w *Worker) getSessionTls(hello *tls.ClientHelloInfo) (*tls.Config, error) {
+	const op = "worker.(Worker).getSessionTls"
+	ctx := context.TODO()
```
jefferai, line 240: Use the worker context instead of TODO if you aren't sure what to use.

jefferai: (Maybe -- see Slack ping)

internal/servers/worker/handler.go
jefferai: Not sure if it's deemed too immaterial/inconsequential or not. Maybe a good question for @jimlambrt

jefferai (commented):
Looks great! Left a couple of comments to follow up on.

internal/servers/worker/proxy/options.go
```diff
@@ -0,0 +1,11 @@
+package proxy
```
louisruch, line 1: Even though there are no packages now I wanted to define the signature as these will be used shortly

louisruch:
@jefferai perhaps went a little left of field here, I decided to create a base proxy `Config` for what I see as the must have parameters in order for the worker to establish a proxy.  I then defined the options to be used in follow up PRs

louisruch:
@tmessi @mgaffney I know you are trying to get the wh work in I have @jefferai and @vancluever looking at this PR as they are most familiar with worker code, but FYI if you wanted to have a look.

internal/servers/worker/handler.go
jimlambrt: I think, we should probably WriteError(...) for the close error.  I noticed this when "porting" it to events and forgot to raise the question in my PR.

louisruch:
ping to anyone who wants to give me a thumbs up
@jefferai @vancluever @jimlambrt :)

PR #1462, opened 2021-08-14
recommend layer 4 network load balancers for HA

rileytg:
current reference architecture for an aws deployment uses a network (layer 4) load balancer. 

https://github.com/hashicorp/boundary-reference-architecture/blob/main/deployment/aws/aws/lb.tf#L3

I have tried and tried to get an application load balancer to work to no avail. im not an expert in these things... 

my understanding is i would setup the tls termination on the controller itself, so from an aws perspective the lb is handling tcp  (L4) not https or http (L7). 

if i am correct, there are further clarifications to this doc i can make, namely the "HTTPS listener" part below this.

if i am incorrect, i would appreciate some advice in getting an alb working. i can submit a pull to  the ref architecture repo if i get it working. (not really sure how to implement health checks using a network lb, so any advice there would also help me update the ref repo.)

### DIFF
--- a/website/content/docs/installing/high-availability.mdx
+++ b/website/content/docs/installing/high-availability.mdx
@@ -25,7 +25,7 @@ The general architecture for the server infrastructure requires 3 controllers an
 ![](/img/production.png)
 
 As shown above, Boundary is broken up into its controller and worker server components across 3 [EC2 instances](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/instance), in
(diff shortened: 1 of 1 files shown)


rileytg:
I was wrong. Here is a working example. Notice some things are not production ready like it only routes to the first controller. 
```tf
module "alb" {
  source  = "terraform-aws-modules/alb/aws"
  version = "~> 6.0"

  name = "my-alb"

  load_balancer_type = "application"

  vpc_id             = local.vpc_id
  subnets            = local.controller_subnet_ids
  security_groups    = [aws_security_group.controller_lb.id, aws_security_group.controller.id]

  target_groups = [
    {
      name_prefix      = "pref-"
      backend_protocol = "HTTP"
      backend_port     = 9200
      target_type      = "instance"
      targets = [
        {
          target_id = aws_instance.controller[0].id
          port = 9200
        }
      ]
    }
  ]

  https_listeners = [
    {
      port               = 443
      protocol           = "HTTPS"
      certificate_arn    = aws_acm_certificate.https.arn
      target_group_index = 0
    }
  ]

  http_tcp_listeners = [
    {
      port               = 80
      protocol           = "HTTP"
      target_group_index = 0
    }
  ]

  tags = {
    Environment = "Test"
  }
}
```

PR #1466, opened 2021-08-17
Update controller.mdx

smilelikeshit:
hi im suggest to add ?sslmode=disable . to aware SSL enable or not . coz I'm having issue for that documentation when starting a boundary

### DIFF
--- a/website/content/docs/configuration/controller.mdx
+++ b/website/content/docs/configuration/controller.mdx
@@ -14,7 +14,7 @@ controller {
   name = "example-controller"
   description = "An example controller"
   database {
-    url = "postgresql://<username>:<password>@10.0.0.1:5432/<database_name>"
+    url = "postgresql://<username>:<password>@10.0.0.1:5432/<database_name>?sslmode=disable"
     max_open_connections = 5
   }
 }


website/content/docs/configuration/controller.mdx
```diff
@@ -14,7 +14,7 @@ controller {
   name = "example-controller"
   description = "An example controller"
   database {
-    url = "postgresql://<username>:<password>@10.0.0.1:5432/<database_name>"
+    url = "postgresql://<username>:<password>@10.0.0.1:5432/<database_name>?sslmode=disable"
```
goshlanguage, line 17: As it's not advised to run postgres like this in production environments, shouldn't you make another example for development environments?

s-christoff:
Howdy @smilelikeshit,
Thank you for bringing this up! I believe this may be better situated [in the comment](https://github.com/hashicorp/boundary/blame/c4d83bb315b1dfdec5d564e99c4a5ed3bed75151/website/content/docs/configuration/controller.mdx#L132), a couple lines above this. A note just pointing out that adding `?sslmode=disable` may need to be added to the string based off the users configuration.

website/content/docs/configuration/controller.mdx
goshlanguage: You should separate your use case from the provided example, indicating that if your database is setup in insecure mode, such as the case in development environments, that the connection string addition is needed. The default should remain the safe, best practices example.

PR #1470, opened 2021-08-18
Add a Connect() function to the public API

remilapeyre:


jefferai:
Hi there, thanks for this!

We've been wanting to do something like this for a while but there are a number of things we haven't sorted out yet, such as:

* Where should this functionality live -- API, SDK, its own module (for dependency reasons)
* What do we want the breakdown of functions to be? The code that is in `connect.go` is mostly linear for that specific function and want a more machine-oriented connect capability to have APIs and structs that work well for more machine-oriented tasks.
* How do we want to refactor the main module to enable this? (`api` cannot depend on the main module, as it does in this PR)
* What primitives do we want to expose? For example, your code exposes a `*websocket.Conn` but it's not clear this is the right approach. Users in most cases would have to know to put it in `net.Conn` mode -- perhaps we'd want to expose that instead.
* How will we handle credential brokering for applications 
* How we can be set up for some of the features coming down the line

All that is to say that we aren't ready to merge a PR like this yet, but as we figure out some of these bits internally, we can reach back out to you and get your input!

remilapeyre:
Thanks for the feedback, I mainly opened this PR as a part of a larger change I'm working on in Terraform, it's also what I'm using in a CLI tool I have. I'll convert this PR to a draft to better reflect its state and will be happy to rework on it when you have a better idea of what the design for this feature should be.

PR #1473, opened 2021-08-19
sdk: add host plugin protobufs and interface

vancluever:
This adds the protobufs and interface for the upcoming plugin-based host
feature for boundary, providing a common interface for building targets
and host catalogs off of dynamic cloud data.


vancluever:
Currently in draft, tests pending.

talanknight:
Lets use a long running feature branch to add our PRs against so we can iterate on the db schema.  I've created https://github.com/hashicorp/boundary/tree/plugin-hostcatalogs which you can base these PRs against.

internal/proto/local/plugin/host.proto
```diff
@@ -0,0 +1,27 @@
...
+  string address = 10;
+
+	// Name of the host.
+	string name = 20;
+
+	// The host catalog of which this host is a part.
+	string host_catalog_id = 30;
+
+	// The list of host sets containing this host.
+	repeated string host_set_ids = 40;
```
talanknight, line 21: The risk here is that the plugin will only be able to populate this from the set of host sets provided in the ListHosts request.  This will not necessarily be a complete list.  MVP we are not planning on populating the host_set list on the API resource, so I don't think this is needed yet.

internal/proto/local/plugin/host.proto
```diff
@@ -0,0 +1,27 @@
...
+// Host contains all fields related to a host returned from a plugin.
+message Host {
+  // The primary address for the host result.
+  string address = 10;
+
+	// Name of the host.
+	string name = 20;
+
+	// The host catalog of which this host is a part.
+	string host_catalog_id = 30;
```
talanknight, line 18: All requests to the plugins currently will only ever include a single host catalog.  Do we want to require the plugin to populate this data or can we let the controller derive it based on the request?

internal/proto/local/plugin/host.proto
```diff
@@ -0,0 +1,27 @@
...
+
+	// The list of host sets containing this host.
+	repeated string host_set_ids = 40;
+
+  // The attributes that are applicable to the specific host. This
+  // may or may not be opaque and can be used for further
```
talanknight, line 24: What does "may or may not be opaque" mean here?

internal/proto/local/plugin/host.proto
```diff
@@ -0,0 +1,27 @@
...
+  // post-processing by Boundary.
+	google.protobuf.Struct attributes = 50;
```
talanknight, line 26: Please label this as optional in the comment above.

This seems useful for allowing the plugin to give options around filtering at list time and maybe when building the filters that go into the host set, right?  Is there any other programmatic purpose you see for this field?

internal/proto/local/plugin/host_catalog.proto
```diff
@@ -0,0 +1,27 @@
+syntax = "proto3";
+
+package plugin.sdk.proto;
+
+option go_package = "github.com/hashicorp/boundary/internal/plugin/sdk/proto;proto";
+
+import "google/protobuf/struct.proto";
+
+// HostCatalog represents a host catalog as passed in to a plugin call. 
+message HostCatalog {
```
talanknight, line 10: Should this include name, description, version, create time, update time, etc?  I'm wondering if this should be as close to the API resource as possible since that is what is what the plugin author will be using as a baseline for understanding what the resource itself is.

internal/proto/local/plugin/host_catalog.proto
```diff
@@ -0,0 +1,27 @@
...
+	// Attributes specific to the host catalog.
+	google.protobuf.Struct attributes = 20;
+	
+  // A set of secrets specific to the host catalog.
+	google.protobuf.Struct secrets = 30;
+}
+
+// HostCatalogPersisted represents state persisted between host
+// catalog calls. This data is encrypted at-rest by Boundary, but is
+// supplied to various plugin calls unencrypted.
```
talanknight, line 23: Can you include in the comment that this data is never returned to the end user and it's purpose is solely for the plugin specific storage per host catalog.

internal/proto/local/plugin/host_set.proto
```diff
@@ -0,0 +1,19 @@
...
+
+import "google/protobuf/struct.proto";
+
+// HostSet represents a host catalog as passed in to a plugin call. 
+message HostSet {
+	// The ID of the host set.
+	string id = 10;
+
+	// The host catalog of which this host set is a part.
+	string host_catalog_id = 20;
```
talanknight, line 15: I don't think this is needed since all calls that include a host set will also include that host set's catalog.

internal/proto/local/plugin/host_catalog.proto
talanknight: Actually thinking about this more...  In protos it is easy to add fields as they become necessary but deprecating them can be a little more painful.  I'm ok with keeping the number of fields at a minimum until we identify the need for more fields.

internal/proto/local/plugin/host.proto
```diff
@@ -0,0 +1,27 @@
...
+
+import "google/protobuf/struct.proto";
+
+// Host contains all fields related to a host returned from a plugin.
+message Host {
+  // The primary address for the host result.
+  string address = 10;
+
+	// Name of the host.
+	string name = 20;
```
talanknight, line 15: Can you identify this as optional?

internal/proto/local/plugin/host.proto
```diff
@@ -0,0 +1,27 @@
...
+
+package plugin.sdk.proto;
+
+option go_package = "github.com/hashicorp/boundary/internal/plugin/sdk/proto;proto";
+
+import "google/protobuf/struct.proto";
+
+// Host contains all fields related to a host returned from a plugin.
+message Host {
+  // The primary address for the host result.
```
talanknight, line 11: Can you please identify this as required?

talanknight (commented):
It feels good to see these created!

internal/plugin/sdk/host/interface.go
```diff
@@ -0,0 +1,78 @@
...
+	//
+	// These hooks take the catalog as it exists in the respective
+	// request (along with any secrets), with both the old and new
+	// state provided in update. These hooks are designed to provide a
+	// way for the provider to both do validation and possibly return a
+	// persisted state. The returned persisted state is encrypted.
+	//
+	// Any error returned will cause the state to not be persisted and
+	// an error returned to the caller.
+	OnCreateCatalog(
```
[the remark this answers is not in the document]
vancluever: Just wrapping, it can be changed though!

internal/proto/local/plugin/host_catalog.proto
vancluever: Yeah, that was the intention for sure. I was just asking myself the question "are the plugins really going to initially use these fields?", and could not really think of a situation where they would be.

internal/proto/local/plugin/host.proto
vancluever: Really only means that we may or may not use it.

I'll take this snippet out so the language is less confusing.

internal/proto/local/plugin/host.proto
vancluever: Yeah, this would be the "for future use" metadata to be used for filtering. Figured it would be useful to scaffold this now, but maybe not?

vancluever:
@talanknight I've changed the base on this PR to the long-running branch, just FYI!

internal/plugin/sdk/host/interface.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"context"
+
+	"github.com/hashicorp/boundary/internal/plugin/sdk/proto"
+)
+
+// HostPlugin describes the interface for host plugins.
+type HostPlugin interface {
+	// The following hook functions run create, update, and delete
```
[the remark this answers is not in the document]
vancluever: Done now!

internal/plugin/sdk/host/interface.go
```diff
@@ -0,0 +1,78 @@
...
+	// functionality for host catalogs.
+	//
+	// These hooks take the catalog as it exists in the respective
+	// request (along with any secrets), with both the old and new
+	// state provided in update. These hooks are designed to provide a
+	// way for the provider to both do validation and possibly return a
```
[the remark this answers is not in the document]
vancluever: Done! Also will add a bit more extra context in the protobuf with some of the other changes.

internal/plugin/sdk/host/interface.go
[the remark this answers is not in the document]
vancluever: Now fixed!

internal/proto/local/plugin/host.proto
```diff
@@ -0,0 +1,27 @@
...
+  string address = 10;
```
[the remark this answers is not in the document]
vancluever: Should be fixed in an upcoming commit, although I'm trying to find something that will auto-format for me... don't think `buf` does it, unfortunately.

internal/proto/local/plugin/host.proto
```diff
@@ -0,0 +1,27 @@
+syntax = "proto3";
+
+package plugin.sdk.proto;
+
+option go_package = "github.com/hashicorp/boundary/internal/plugin/sdk/proto;proto";
+
+import "google/protobuf/struct.proto";
+
+// Host contains all fields related to a host returned from a plugin.
+message Host {
```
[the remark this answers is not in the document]
vancluever: Yeah, wondering how we go about this. This is why I had the `name` field, as I'm not too sure if having multiple identifier fields is particularly a good thing in terms of what would be redundant and what would not be. Thoughts? Maybe we just move `name` to `id`.

internal/proto/local/plugin/host.proto
vancluever: 👍 Fair point, will remove for now. I think that we might need to figure out the best way to determine exactly what host is a member of what set for caching, maybe possibly by way of a mechanism that allows us to track what sets and catalogs a host has been "seen" in. I think that's the best way I can think of to actually make those kinds of associations that would be pretty hard-set in static hosts.

Given this logic and pursuant to the next comment It think we can just remove the ID from the payload too.

internal/proto/local/plugin/host.proto
vancluever: Yeah, given the discussion on the member set field, just going to remove this too.

internal/proto/local/plugin/host.proto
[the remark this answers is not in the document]
talanknight: My problem with using `name` for that purpose is that `name` has a meaning and usage already in boundary which is to provide a boundary resource with a human friendly and unique name (unique at least for the host catalog).  An external id which satisfies the requirements I listed above may not be human readable or useful, and it might be cool if we could eventually allow end users to assign some of these auto generated hosts with a boundary resource name.

If we have a field named "external_id" or something similar and document the requirements (must be stable and unique for the endpoint which this host represents), we could then use this field + the catalog id as a way to generate a stable boundary resource id.

internal/proto/local/plugin/host.proto
vancluever: Will be done in the next commit!

internal/proto/local/plugin/host.proto
vancluever: Going to keep it as available. Fixed the comments though, and made them more descriptive to the purpose. I think we could possibly surface this in API call results, which would be good just for information purposes at least!

internal/proto/local/plugin/host_catalog.proto
vancluever: Resolving, I think we're on the same page here 🙂

internal/proto/local/plugin/host_catalog.proto
vancluever: Done!

internal/proto/local/plugin/host_set.proto
vancluever: Removed!

internal/proto/local/plugin/host.proto
vancluever: I've actually removed this now and replace it with `id` (we can maybe mull on this after too)

internal/proto/local/plugin/host.proto
vancluever: Done!

internal/proto/local/plugin/host.proto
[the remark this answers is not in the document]
vancluever: I've moved this to `id` for now, let's continue this conversation in other comments 🙂

vancluever:
Actually going to mark this one ready for review for now because I don't think we need to have tests in this. We can possibly add a test plugin in a later PR (maybe in the upcoming plugin manager PR) that implements just barebones end-to-end functionality.

internal/proto/local/plugin/host_catalog.proto
```diff
@@ -0,0 +1,30 @@
...
+
+	// Attributes specific to the host catalog.
+	google.protobuf.Struct attributes = 20;
+	
+  // A set of secrets specific to the host catalog.
+	google.protobuf.Struct secrets = 30;
+}
+
+// HostCatalogPersisted represents state persisted between host
+// catalog calls. Its primary purpose is to store authentication data
```
talanknight, line 22:
```suggestion
// catalog calls. Its intended purpose is to store authentication data
```

Although the plugin could use this for anything really.

Consider also adding a TODO to specify a size limit for this data.  This is something we should add prior to exposing the plugin SDK publicly since it's easier to ease size restrictions than it is to add them after the fact.

internal/proto/local/plugin/host_catalog.proto
talanknight: `id` also has a meaning in boundary (the boundary supplied resource id) and this field is also included in host catalogs and host sets in this interface so from a plugin perspective they are already exposed to this meaning.  What do you think about `external_id`?

internal/plugin/sdk/host/interface.go
```diff
@@ -0,0 +1,83 @@
...
+	// at-rest encrypted. This should be used to store authentication
+	// data and other necessary configuration to be used in later hooks
+	// and lookup calls. If an error is returned, data persistence is
+	// aborted.
+	OnCreateCatalog(ctx context.Context, cat proto.HostCatalog) (proto.HostCatalogPersisted, error)
+
+	// OnUpdateCatalog is a hook that runs when a host catalog is
+	// updated. It takes a context and the host catalog to update, both
+	// in its pre-update, and post-update form. The post-update catalog
+	// may contain (optional) secret data. The existing persisted state
```
talanknight, line 29: I think we need to work on figuring out our approach for updating secrets.  Since this is a write only field in the API do we 1. require all secret values to be provided every time the secrets field is updated or 2. do we allow partial updates of the secrets field?

The benefit of approach 1 is that it sort of matches what we are doing with the host catalog in that it provides the complete old and the complete new, assuming the plugin saves the complete old in the persisted data.

I like approach 2 because it matches our existing APIs approach for updating fields (although we've never really had a write only object w/ updatable subfields before).  Following this route we'll need to also pass in an update mask for secrets in the update catalog calls to the plugin.

@jefferai you and I have discussed update semantics w/ the API quite a bit.  Do you have an opinion here?

talanknight (changes requested):

internal/proto/local/plugin/host_catalog.proto
vancluever: Added the TODO!

internal/proto/local/plugin/host_catalog.proto
vancluever: Sounds good to me!

vancluever: Hmm, I'll make this change, but wondering if we need something new for plugins to note that this ID may not be exactly what is reflected in Boundary?

talanknight: Woops! I added my comment to the wrong closed conversation.  This was meant for Host and not HostCatalog.  TO be clear, I think Host Catalog and Host Set should use `id` since this is the value provided by boundary.  I think Host should use `external_id` since that is the value that is populated by the plugin and will be used to calculate the stable id.

internal/plugin/sdk/host/interface.go
vancluever: I also think there's something to be said about the simplicity of the first approach, but yeah, let's see what Jeff has to say.

internal/plugin/sdk/host/interface.go
```diff
@@ -0,0 +1,83 @@
...
+package host
+
+import (
+	"context"
+
+	"github.com/hashicorp/boundary/internal/plugin/sdk/proto"
+)
+
+// HostPlugin describes the interface for host plugins.
+type HostPlugin interface {
```
jefferai, line 13: Can you explain the purpose of this interface? This seems like it's an analogue to the grpc service interface that will be generated, but it's using different parameter styles (e.g. for a service, the signature for `OnCreateCatalog` would be something like `OnCreateCatalog(ctx context.Context, req *pb.OnCreateCatalogRequest) (*pb.OnCreateCatalogResponse, error)`). Shouldn't the service be created, and its methods are the interface? Otherwise you have to translate between that one and this one.

internal/proto/local/plugin/host.proto
```diff
@@ -0,0 +1,21 @@
...
```
jefferai, line 10: Here and for the other proto files: Why redefine `Host`, `HostSet`, and `HostCatalog` rather than use the existing protos that are already defined? That way the request object for something like `OnUpdateCatalog` above could consist of the normal HostCatalog object we already have defined (which includes attributes), along with a protobuf Struct for the secret data.

internal/plugin/sdk/host/interface.go
talanknight: IMO it is good to hide that we are using grpc, and possibly even protos at all, from the plugin author.  Adding a layer between what they code against and how we are communicating with the plugin process gives us flexibility to change in the future.

Regarding what to expose in the interface, I also think the it should be the minimal one needed to easily do what they need to do.  This is especially true when we are just starting out and don't have a use case for a lot of these fiels (update time, allowed actions, boundary resource version, etc...)

jefferai: Protos simply compile to Go structs; grpc services turn into Go interfaces. I think the question is whether these structs and interfaces are sufficient for the task at hand. If we wanted to switch off grpc, would we also switch the interface? If not, why not just use the one it generates as a starting point?

At a broader level, why is it bad for plugin authors to know that it's a grpc-generated service or proto-generated structs?

internal/plugin/sdk/host/interface.go
jefferai: I like approach 1. I think that what's in secrets should be extremely minimal (e.g. secret key only, not key ID which would live in attributes) so the only time you'd actually write it is to explicitly provide a credential, and it's not worth dealing with update masks for that case.

internal/plugin/sdk/host/interface.go
jefferai: FWIW -- I don't think there is an explicit need to hide that it's grpc or proto messages; I think an interface _just to do that_ is unnecessary. I do think there is value in an interface that exists to provide convenience to the user, e.g. taking in two string parameters that under the hood go into two values in a `XRequest` object instead of creating that object yourself, as part of a plugin SDK. But, given that we can change the implementation at any time in the near term because it's only our plugins to start with and we're not doing versioning, I don't think this is _necessary_ at the moment. Fine to have, but if it forces us to define a Go API right now that we otherwise don't need to define, it might not be worth the time.

As for the proto messages -- if the intention is to make the types easier for Go developers to work with, I would argue that an interface like this one could transform those into our `api` Go structs. So we can reuse the same proto messages across the plugin interface, then parse into the API struct if we wish. But I'm not really understanding why we need to define a third version of the same types we already have in both proto and Go form, along with all the tooling we have to convert between them.

talanknight: For the proto message, the state we pass on is captured mid transaction and inadvertently exposes implementation detail.  For example, version, update time, create time, all will have different values depending on where in repository we make a call to the plugin.  It's also difficult to imagine how those fields would be helpful to the plugin, so they would just be added mental load for no benefit.

internal/proto/local/plugin/host_plugin_service.proto
```diff
@@ -0,0 +1,174 @@
...
+
+import "google/protobuf/struct.proto";
+
+import "controller/api/resources/hostcatalogs/v1/host_catalog.proto";
+import "controller/api/resources/hostsets/v1/host_set.proto";
+import "controller/api/resources/hosts/v1/host.proto";
+
+// HostPluginService describes the service for host plugins.
+service HostPluginService {
+  // OnCreateCatalog is a hook that runs when a host catalog is
```
talanknight, line 15: Maybe I'm off base with this observation, but happy to talk about it: Are these really "hooks"?  To me hooks suggest a runtime component relying on callbacks.  I guess you could argue that is what happens w/ plugins.  I don't know. Thoughts?

internal/proto/local/plugin/host_plugin_service.proto
```diff
@@ -0,0 +1,174 @@
...
+  // The host catalog to create. The request may contain optional
+  // secret data to help authenticate the request against a cloud
+  // API.
+  controller.api.resources.hostcatalogs.v1.HostCatalog catalog = 10;
+}
+
+message OnCreateCatalogResponse {
+  // Secret data to persist encrypted within Boundary. This should be
+  // used to store authentication data and other necessary
+  // configuration to be used in later hooks and calls. Returning an
```
talanknight, line 50:
```suggestion
  // configuration to be used in later calls to the host plugin service. Returning an
```

internal/proto/local/plugin/host_plugin_service.proto
```diff
@@ -0,0 +1,174 @@
...
+  // The hosts to return.
+  repeated ListHostsResponseHost hosts = 10;
+}
+
+message ListHostsResponseHost {
+  // Required. A stable identifier for this host. This field is used
+  // to generate a stable ID for the host within Boundary and is
+  // included in audit logs. It should be set to something unique and
+  // useful, ie: a compute instance ID.
+ 	string external_id = 10;
```
talanknight, line 151: There is still some strange spacing happening in this proto file.  Here and elsewhere.

talanknight (approved):
Looks good!  We spoke offline about where to put the generated protos, which I think is different from what is here currently.  Once that is fixed I approve!

internal/proto/local/plugin/host_plugin_service.proto
vancluever: Thinking through the process:

* It's a hook because of the idioms ("on create catalog"). The plugin has the option of sending back data to persist, and returning errors (possibly due to additional validation, etc). It also is not expected to do anything additional to facilitate creation - in fact, we likely do not expect anything to be created here (other than maybe credential rotation).
* It's not a hook because we would likely always be relying on it for initial authentication data persistence.

I think it still is more of a hook than not, mainly based on idioms and lack of pure necessity of action on part of the plugin here during creation - this could no-op and the host catalog would still be created.

internal/proto/local/plugin/host.proto
vancluever: Old commit, just resolving the conversation now.

internal/proto/local/plugin/host_plugin_service.proto
vancluever: Resolving this now as Todd and I discussed this offline.

internal/proto/local/plugin/host_plugin_service.proto
vancluever: Fixed in next commit!

PR #1474, opened 2021-08-20
Properly accept grant format type=<type>;output_fields=<of>

jefferai:
Although we document that this is allowed, and the ACL code handles it,
some of the validation code did not. This would cause a semi
intermittent problem because that validation code is used only in some
code paths; specifically we skip it (for now) to avoid erroring out on
some deprecated grant string formats when reading grants from the DB,
but we use it to error out when validating incoming grants.


louisruch (approved):
LGTM

jimlambrt (approved):

PR #1475, opened 2021-08-20
feat(host): Add host plugin related schemas and tests

talanknight:
This adds the schema files for host plugins, plugin based catalogs and
plugin based sets.  This also adds the protos and tests which utilize
gorm to ensure the schemas enforce the constraints we expect.

Along with this change was a migration which moved catalog names to the
base type table so we can ensure all catalogs in a single scope,
regardless of subtype, have a unique name.


internal/plugin/immutable_fields_test.go
```diff
@@ -0,0 +1,238 @@
...
+	w := db.New(conn)
+
+	ts := timestamp.Timestamp{Timestamp: &timestamppb.Timestamp{Seconds: 0, Nanos: 0}}
+	plg := testPlugin(t, conn, "test")
+	plgVer := testPluginVersion(t, conn, plg.GetPublicId(), "0.0.1")
+	plgExe := testPluginExecutable(t, conn, plgVer.GetPublicId(), "windows", "amd64", []byte("test"))
+
+	newPluginExe := plgExe
+
+	tests := []struct {
```
vancluever, line 156: Is the executable `bytea` column missing from the test set here? Just noticed there wasn't a case calling it out.

internal/plugin/plugin_executable.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"github.com/hashicorp/boundary/internal/plugin/store"
+)
+
+// OperatingSystem defines the operating systems supported for plugin executables
+type OperatingSystem string
+
+const (
+	UnknownOS   OperatingSystem = "unknown"
```
vancluever, line 11: Again, might want to restrict this to just the arches we actually support for now.

internal/plugin/plugin_version_test.go
```diff
@@ -0,0 +1,154 @@
...
+			if tt.wantErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
+		})
+	}
+}
+
+func TestPluginVersion_Delete(t *testing.T) {
```
vancluever, line 108: Noticed that the executable delete tests had cascade tests for both plugin and plugin version, do we want to add a cascade test for this too (ie: deleting plugin deletes all versions)?

vancluever (approved):
Looking great!

internal/db/schema/migrations/postgres/15/01_plugin.up.sql
```diff
@@ -0,0 +1,222 @@
...
+  create trigger immutable_columns before update on plugin_version
+    for each row execute procedure immutable_columns('public_id', 'plugin_id', 'create_time', 'semantic_version');
+
+  -- Values retrieved by using $ go tool dist list | cut -d / -f1 | uniq
+  -- TODO: Remove this and rely on go logic to enforce these values so we
+  --   avoid the toil of keeping this up to date.
+  create table plugin_operating_system_enm (
+    name text not null primary key
+      constraint only_predefined_operating_systems_allowed
+      check(name in ('unknown', 'aix', 'android', 'darwin', 'dragonfly',
```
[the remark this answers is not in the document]
talanknight: I've left the DB more permisive for the values they'll allow and decreased the consts defined in the go code to the ones we support.  In practice we could potentially only allow a binary to add a plugin using a value of it's own arch and os (probably done on startup).   Being more permissive on the db seems like a middle ground between allowing any values here and potentially getting garbage and restricting it aggressively to only the values we currently support and having to update the list frequently.

vancluever (approved):
Looks good!

internal/db/schema/migrations/postgres/16/01_plugin.up.sql
```diff
@@ -0,0 +1,149 @@
...
+    add constraint plugin_scope_id_public_id_uq
+      unique (scope_id, public_id),
+    add constraint plugin_scope_id_name_uq
+      unique (scope_id, name);
+
+
+  -- insert, update, and delete plugin_subtypes are created since we are adding
+  -- subtyped plugins and we need to keep the base table plugin in sync with all
+  -- subtype tables.
+  create or replace function insert_plugin_subtype()
```
mgaffney, line 34: In general, we should not use `create or replace` statements because they can hide errors. If we are replacing something, we should know what we are replacing because there are probably dependencies on it and our code needs to account for that. So we should have a `drop` statement followed by a `create` statement.  If either returns an error, there is a problem that needs to be fixed. 

Likewise, if we are creating something new, we should only have a `create` statement. If something already exists, it means we didn't know it existed and the migration should error out.

internal/db/schema/migrations/postgres/16/03_host.up.sql
```diff
@@ -0,0 +1,63 @@
...
+    host_catalog.public_id = st.public_id and
+    st.name is not null;
+
+  -- Replace the insert_host_catalog_subtype function defined in 0/20_host.up.sql
+  -- to include the name.
+  -- insert_host_catalog_subtype() is a before insert trigger
+  -- function for subtypes of host_catalog
+  create or replace function
+    insert_host_catalog_subtype()
+    returns trigger
```
mgaffney, line 28:
```suggestion
  create or replace function insert_host_catalog_subtype()
    returns trigger
```

internal/db/schema/migrations/postgres/16/01_plugin.up.sql
```diff
@@ -0,0 +1,149 @@
...
+    unique(plugin_id, semantic_version)
+  );
+
+  create trigger default_create_time_column before insert on plugin_version
+    for each row execute procedure default_create_time();
+
+  create trigger immutable_columns before update on plugin_version
+    for each row execute procedure immutable_columns('public_id', 'plugin_id', 'create_time', 'semantic_version');
+
+  create table plugin_executable (
```
mgaffney, line 118: Is the executable supposed to be encrypted?

internal/db/schema/migrations/postgres/16/03_host.up.sql
```diff
@@ -0,0 +1,63 @@
...
+    static_host_catalog st
+  where
+    host_catalog.public_id = st.public_id and
+    st.name is not null;
+
+  -- Replace the insert_host_catalog_subtype function defined in 0/20_host.up.sql
+  -- to include the name.
+  -- insert_host_catalog_subtype() is a before insert trigger
+  -- function for subtypes of host_catalog
+  create or replace function
```
mgaffney, line 26: Should there be a `create trigger` statement for this function?

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
```diff
@@ -0,0 +1,157 @@
...
+    │public_id       │      │host_catalog_id (fk)│
+    │host_catalog_id │┼┼──○┼│name                │
+    └────────────────┘      │description         │
+                            │attributes          │
+                            └────────────────────┘
+
+*/
+
+  create table plugin_host_catalog (
+    public_id wt_public_id primary key,
```
mgaffney, line 41: Does the prefix of the `public_id` have to match the `id_prefix` of the `plugin_id`?

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
```diff
@@ -0,0 +1,157 @@
...
+    description text,
+    create_time wt_timestamp,
+    update_time wt_timestamp,
+    version wt_version,
+    attributes bytea,
+    foreign key (scope_id, public_id)
+      references host_catalog (scope_id, public_id)
+      on delete cascade
+      on update cascade,
+    unique(scope_id, name)
```
mgaffney, line 62: Can you name this constraint?

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
```diff
@@ -0,0 +1,157 @@
...
+
+  create trigger update_host_catalog_subtype before update on plugin_host_catalog
+    for each row execute procedure update_host_catalog_subtype();
+
+  create trigger delete_host_catalog_subtype after delete on plugin_host_catalog
+    for each row execute procedure delete_host_catalog_subtype();
+
+  create table plugin_host_catalog_secret (
+    catalog_id wt_public_id primary key
+      references plugin_host_catalog (public_id)
```
mgaffney, line 88: Can you name this constraint?

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
```diff
@@ -0,0 +1,157 @@
...
+  create trigger default_create_time_column before insert on plugin_host_catalog_secret
+      for each row execute procedure default_create_time();
+
+  create trigger immutable_columns before update on plugin_host_catalog_secret
+      for each row execute procedure immutable_columns('catalog_id', 'create_time');
+
+  create table plugin_host_set (
+    public_id wt_public_id primary key,
+    catalog_id wt_public_id not null
+      constraint host_catalog_fkey
```
mgaffney, line 115:
```suggestion
      constraint plugin_host_catalog_fkey
```

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
```diff
@@ -0,0 +1,157 @@
...
+        references plugin_host_catalog (public_id)
+        on delete cascade
+        on update cascade,
+    name wt_name,
+    description text,
+    create_time wt_timestamp,
+    update_time wt_timestamp,
+    version wt_version,
+    attributes bytea,
+    unique(catalog_id, name),
```
mgaffney, line 125: Can you name this constraint?

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
```diff
@@ -0,0 +1,157 @@
...
+    foreign key (catalog_id, public_id)
```
mgaffney, line 126: Can you name this constraint?

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
```diff
@@ -0,0 +1,157 @@
...
+      constraint host_plugin_fkey
+        references host_plugin (public_id)
+        on delete cascade
+        on update cascade,
+    name wt_name,
+    description text,
+    create_time wt_timestamp,
+    update_time wt_timestamp,
+    version wt_version,
+    attributes bytea,
```
mgaffney, line 57: Can this be null?

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
```diff
@@ -0,0 +1,157 @@
...
+    foreign key (scope_id, public_id)
```
mgaffney, line 58: Can you name this constraint?

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
```diff
@@ -0,0 +1,157 @@
...
+      constraint host_catalog_fkey
+        references plugin_host_catalog (public_id)
+        on delete cascade
+        on update cascade,
+    name wt_name,
+    description text,
+    create_time wt_timestamp,
+    update_time wt_timestamp,
+    version wt_version,
+    attributes bytea,
```
mgaffney, line 124: Why can this be null?

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
```diff
@@ -0,0 +1,157 @@
...
+    unique(catalog_id, name),
+    foreign key (catalog_id, public_id)
+      references host_set (catalog_id, public_id)
+      on delete cascade
+      on update cascade,
+    unique(catalog_id, public_id)
```
mgaffney, line 130: Constraint name?

mgaffney (changes requested):
I only reviewed the sql. I will review the rest of the code in another review.

talanknight (commented):
For some reason github isn't letting me respond to a number of your questions Mike.

> Should there be a create trigger statement for this function?
There is one in 04_plugin_host.up.sql file

> In general, we should not use create or replace statements
If we do a drop then a create, doesn't that mean we'll need to do the same for all triggers using those functions?  is that the intent of your suggestion to not use create or replace?

internal/db/schema/migrations/postgres/16/01_plugin.up.sql
talanknight: removed the executable table.

internal/db/schema/migrations/postgres/16/03_host.up.sql
talanknight: This is in the 04_plugin_host.up.sql file.

internal/db/schema/migrations/postgres/16/01_plugin.up.sql
```diff
@@ -0,0 +1,149 @@
...
+  create table plugin_version (
+    public_id wt_public_id primary key,
+    plugin_id wt_public_id not null
+      constraint plugin_fkey
+        references plugin (public_id)
+        on delete cascade
+        on update cascade,
+    semantic_version text not null
+      constraint plugin_version_requires_semantic_version
+      check(length(semantic_version) > 4), -- minimum length is length("0.0.0")
```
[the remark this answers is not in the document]
talanknight: Moved semantic version to host_plugin table and added a regex provided by https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

We don't have many regexes in our constraints though so what are your thoughts on being so restrictive in db about the values that can go into this column?

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
talanknight: Made not null.

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
talanknight: made not null.

internal/db/schema/migrations/postgres/16/01_plugin.up.sql
```diff
@@ -0,0 +1,78 @@
...
+      default 'global'
+      constraint plugin_must_be_in_global
+      references iam_scope_global(scope_id)
+        on delete cascade
+        on update cascade;
+
+  alter table plugin
+    alter column scope_id drop default;
+
+  -- Add constraints that enforce names are unique across all plugin subtypes
```
jefferai, line 23: Given that for now we have no plans nor design for what non-`global` plugins would look like, can you also add in a constraint that the scope ID is global? We can remove this later if needed.

jefferai: Oop, I missed that in the alter table statement above.

internal/host/plugin/ids.go
```diff
@@ -0,0 +1,41 @@
...
+	HostSetPrefix     = "hsplg"
+
+	Subtype = subtypes.Subtype("plugin")
+)
+
+func newHostCatalogId() (string, error) {
+	// TODO: Take in a prefix portion as an argument.
+	id, err := db.NewPublicId(HostCatalogPrefix)
+	if err != nil {
+		return "", errors.WrapDeprecated(err, "plugin.newHostCatalogId")
```
jefferai, line 29: Let's please not use deprecated methods in new code...

internal/host/plugin/host_catalog_secret.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// SetTableName sets the table name. If the caller attempts to
+// set the name to "" the name will be reset to the default name.
+func (c *HostCatalogSecret) SetTableName(n string) {
+	c.tableName = n
+}
+
+func (c *HostCatalogSecret) encrypt(ctx context.Context, cipher wrapping.Wrapper) error {
+	const op = "vault.(HostCatalogSecret).encrypt"
```
jefferai, line 53: Not in `vault` :-)

internal/host/plugin/host_catalog.go
```diff
@@ -0,0 +1,63 @@
+package plugin
```
mgaffney, line 1: Can you add doc for the package?

internal/host/plugin/host_catalog_secret.go
```diff
@@ -0,0 +1,90 @@
...
+	"context"
+	"encoding/json"
+
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/host/plugin/store"
+	wrapping "github.com/hashicorp/go-kms-wrapping"
+	"github.com/hashicorp/go-kms-wrapping/structwrapping"
+)
+
+type HostCatalogSecret struct {
```
mgaffney, line 13: Doc?

internal/plugin/ids.go
```diff
@@ -0,0 +1,19 @@
+package plugin
```
mgaffney, line 1: Can you add package doc?

internal/plugin/host/plugin.go
```diff
@@ -0,0 +1,43 @@
+package host
```
mgaffney, line 1: Can you add package doc?

jefferai (approved):

internal/db/schema/migrations/postgres/16/02_hostplugin.up.sql
```diff
@@ -43,19 +43,21 @@ begin;
         check(length(trim(plugin_name)) > 0)
       constraint plugin_name_must_be_lowercase
         check(lower(trim(plugin_name)) = plugin_name)
-      constraint plugin_name_must_be_unique
+      constraint plugin_host_name_uq
```
mgaffney, line 46:
```suggestion
      constraint plugin_host_plugin_name_uq
```

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
```diff
@@ -54,13 +54,15 @@ begin;
     create_time wt_timestamp,
     update_time wt_timestamp,
     version wt_version,
-    attributes bytea not null,
+    attributes bytea not null
+      constraint attributes_must_not_empty
```
mgaffney, line 58:
```suggestion
      constraint attributes_must_not_be_empty
```

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
```diff
@@ -124,15 +126,17 @@ begin;
     create_time wt_timestamp,
     update_time wt_timestamp,
     version wt_version,
-    attributes bytea not null,
-    constraint host_plugin_set_name_must_be_unique_in_catalog
+    attributes bytea not null
+      constraint attributes_must_not_empty
```
mgaffney, line 130:
```suggestion
      constraint attributes_must_not_be_empty
```

mgaffney (approved):
Looks good. Just some minor fixes to constraint names.

PR #1477, opened 2021-08-22
Adjust syntax in example

hb9cwp:


jimlambrt:
Thank you for the contribution.  Unfortunately, I'm not sure your suggestion is less confusing for most readers.

PR #1480, opened 2021-08-23
Refactor warehouse columns

tmessi:
refactor(wh): Remove host prefix from wh_user_dimension org fields

These fields are being refactored so they conform with the new 
wh_credential_dimension. This will inform Business Intelligence tools that
these attributes conform across dimensions.

---

refactor(wh): Rename primary and foreign key columns

The tables in the operational part of the Boundary database (meaning 
non-warehouse tables) suffix primary key columns and foreign key columns 
with `_id`. Many of these columns appear in the warehouse dimension and 
fact tables as normal attributes, not primary keys or foreign keys. To 
disambiguate ID columns from operational tables, the primary key column for
dimension tables will be renamed from `id` to `key` and the foreign key
columns in the fact tables will be renamed to change the suffix from
`_id` to `_key`.

---

db: Refactor test setup for pre-populated data

Now when initializing the test database for sql tests, it will load the
colors persona data into the database. This will provide a data set that
tests can use without needing to call
`wtt_load`.

The main motivation is to allow for a separation of setup and test 
execution that requires crossing a transaction boundary. The pgTap tests 
are required to exectue as a single transaction, so that the rollback can
reset the state, and each test can be isolated. However, can cause issues
with some test senarios, particuarly with the use of some times releated
functions like `current_timestamp` and `now` which return the time that the
current transaction started. This creates an issue if a test relies on some
setup that sets a time via `now`, then an update that expects a new time
via `now`.

See:
    https://www.postgresql.org/docs/13/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT


internal/db/schema/migrations/postgres/15/01_wh_rename_key_columns.up.sql
```diff
@@ -0,0 +1,431 @@
...
+  drop function wh_insert_session_connection;
+  create function wh_insert_session_connection()
+    returns trigger
+  as $$
+  declare
+    new_row wh_session_connection_accumulating_fact%rowtype;
+  begin
+    with
+    authorized_timestamp (date_dim_id, time_dim_id, ts) as (
+      select wh_date_id(start_time), wh_time_id(start_time), start_time
```
mgaffney, line 294: Same comments as above about renaming these

internal/db/schema/migrations/postgres/15/01_wh_rename_key_columns.up.sql
```diff
@@ -0,0 +1,431 @@
...
+      -- been inserted into the wh_session_accumulating_fact table.
+      return null;
+    end if;
+
+    date_col = 'session_' || new.state || '_date_key';
+    time_col = 'session_' || new.state || '_time_key';
+    ts_col   = 'session_' || new.state || '_time';
+
+    q = format('update wh_session_accumulating_fact
+                   set (%I, %I, %I) = (select wh_date_id(%L), wh_time_id(%L), %L::timestamptz)
```
mgaffney, line 371: Same comments as above about renaming these

internal/db/schema/migrations/postgres/15/01_wh_rename_key_columns.up.sql
```diff
@@ -0,0 +1,431 @@
...
+      -- wh_session_connection_accumulating_fact table.
+      return null;
+    end if;
+
+    date_col = 'connection_' || new.state || '_date_key';
+    time_col = 'connection_' || new.state || '_time_key';
+    ts_col   = 'connection_' || new.state || '_time';
+
+    q = format('update wh_session_connection_accumulating_fact
+                   set (%I, %I, %I) = (select wh_date_id(%L), wh_time_id(%L), %L::timestamptz)
```
mgaffney, line 415: Same comments as above about renaming these

internal/db/schema/migrations/postgres/15/02_wh_rename_host_dimension_org.up.sql
```diff
@@ -0,0 +1,156 @@
+begin;
+  alter table wh_host_dimension
+    rename column host_organization_id to organization_id;
+  alter table wh_host_dimension
+    rename column host_organization_name to organization_name;
+  alter table wh_host_dimension
+    rename column host_organization_description to organization_description;
+
+  -- replaces function from 0/65_wh_session_dimensions.up.sql
```
mgaffney, line 9:
```suggestion
  -- replaces view from 0/65_wh_session_dimensions.up.sql
```

internal/db/schema/migrations/postgres/15/02_wh_rename_host_dimension_org.up.sql
```diff
@@ -0,0 +1,156 @@
...
+     and s.public_id = m.set_id
+     and t.public_id = ts.target_id
+     and s.public_id = ts.host_set_id
+     and p.public_id = t.scope_id
+     and p.type = 'project'
+     and o.public_id = p.parent_id
+     and o.type = 'org'
+  ;
+
+  -- replaces function from 15/01_wh_rename_key_columns.up.sql
```
mgaffney, line 58:
```suggestion
  -- replaces view from 15/01_wh_rename_key_columns.up.sql
```

mgaffney (changes requested):

mgaffney (dismissed):
Great job!

mgaffney (approved):

PR #1482, opened 2021-08-24
Speed up test execution

tmessi:
This PR changes how test execute to speed up the execution both locally
and in CI.

The main change is to use a single docker image instead of spawning
docker images for individual tests. Instead the tests create a fresh
database in the same container, using a template database that already
has the migrations run.

See the individual commits for more details.


testing/dbtest/docker/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM postgres:13-alpine
```
jefferai, line 1: Our base supported PG version is 11, so probably it'd be good to start with that if possible as that way if we are using something incompatible with any new code we'll catch it in the tests.

testing/dbtest/docker/Makefile
```diff
@@ -0,0 +1,35 @@
+all: docker-build
+
+CWD := $(shell pwd)
+DOCKER_ARGS ?=
+
+REGISTRY_NAME?=docker.io/hashicorpboundary
+IMAGE_NAME=postgres
+IMAGE_TAG=$(REGISTRY_NAME)/$(IMAGE_NAME):13-alpine
```
jefferai, line 8: Same comment here

Makefile
```diff
@@ -152,14 +152,20 @@ website-install:
...
+
 test-ci: install-go
 	~/.go/bin/go test ./... -v $(TESTARGS) -timeout 120m
 
 test-sql:
 	$(MAKE) -C internal/db/sqltest/ test
 
-test: 
-	~/.go/bin/go test ./... -timeout 30m
+test:
```
jefferai, line 167: Can we integrate `test-database-up` and `test-database-down` into this target so that if we want to run all tests locally we get the same behavior?

jefferai:
This PR seems to be fully CI-oriented, and as a result breaks the capability to run a single test. I realize we could use the makefile target to pull up a DB to then run the test against, but the port number is hardcoded everywhere (as the default) so if you want to run a single test while debugging some other issue you can't do it.

It'd be better if it uses a templated DB if available, and falls back to our current method of launching and using a Docker container if not.

testing/dbtest/docker/Dockerfile
tmessi: Yea I can make that change. I was also thinking that we could expand this to build multiple images so we could test against multiple versions.

It is also worth noting that most of the tests on `main` currently run against `postgres:latest`.

Makefile
tmessi: Yea I can do that. I wasn't sure if people tend to run `make test` frequently (I would guess not given how long it tends to take), but with these changes the workflow I was envisioning would be to run `make test-database-up` once, and then be able to run `make test` multiple times without needing to teardown and recreate the test database. The only time you really need to recreate the test database is if new migrations are added.

But this can just as easily be done by using `make test-database-up` and calling `go test` directly.

testing/dbtest/docker/Dockerfile
jefferai: 👍 on expanding to run on multiple versions. When you say that most of the tests run on `postgres:latest`, where do you see that? Our old dockertest config specified version 11...was CI doing something different?

tmessi: Maybe "most" isn't accurate. It does look like the dockertest config would default to 11. But I did notice a number of `postgres:latest` containers getting created when I was monitoring docker and running the full test suite from `main`. I didn't track down where they were specifically coming from.

Makefile
tmessi: I looks like it is still the case that `./...` stays within a single module. Should we be explicitly running the tests for `api/` and `sdk/` in CI?

internal/cmd/base/option.go
```diff
@@ -149,3 +150,10 @@ func WithStatusCode(statusCode int) Option {
 		o.withStatusCode = statusCode
 	}
 }
+
+// WithDatabaseTemplate allows for using an existing database template for initializing the boundary database.
```
jimlambrt, line 154: Very small thing, but this comment is particularly long and I think most of us are wrapping comments around 80chars.

internal/servers/controller/handlers/credentialstores/credentialstore_service_test.go
```diff
@@ -124,7 +124,7 @@ func TestList(t *testing.T) {
 				return
 			}
 			require.NoError(t, gErr)
-			assert.Empty(t, cmp.Diff(got, tc.res, protocmp.Transform()))
+			assert.ElementsMatch(t, got.Items, tc.res.Items)
```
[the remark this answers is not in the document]
jimlambrt: ty for the comments here.

internal/cmd/base/dev.go
```diff
@@ -39,7 +40,11 @@ func (b *Server) CreateDevDatabase(ctx context.Context, opt ...Option) error {
 
 	switch b.DatabaseUrl {
```
jimlambrt, line 41: I realize that there are not existing unit tests for this receiever func... but would it be possible to add one that tests the new option's behavior?

internal/db/schema/postgres/postgres_test.go
```diff
@@ -33,490 +33,352 @@ package postgres
...
-		}
-		defer func() {
-			if err := d.close(t); err != nil {
-				t.Error(err)
-			}
-		}()
-		test(t, d, []byte("SELECT 1"))
+	t.Parallel()
+	ctx := context.Background()
+	c, u, _, err := dbtest.StartUsingTemplate("postgres", dbtest.WithTemplate("template1"))
```
jimlambrt, line 49: Would you mind defining consts for the supported dialects and templates and using them in this PR?

testing/dbtest/template.go
```diff
@@ -0,0 +1,111 @@
+package dbtest
+
```
jimlambrt, line 2: Any thoughts on writing unit tests for these funcs?

testing/dbtest/template.go
```diff
@@ -0,0 +1,111 @@
...
+const letterBytes = "abcdefghijklmnopqrstuvwxyz"
+const (
+	letterIdxBits = 6                    // 6 bits to represent a letter index
+	letterIdxMask = 1<<letterIdxBits - 1 // All 1-bits, as many as letterIdxBits
+	letterIdxMax  = 63 / letterIdxBits   // # of letter indices fitting in 63 bits
+)
+
+func randStr(n int) string {
+	b := make([]byte, n)
```
jimlambrt, line 30: is n == 0 a valid parameter?

testing/dbtest/template.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+		cache >>= letterIdxBits
+		remain--
+	}
+
+	return string(b)
+}
+
+func startUsingTemplate(dialect string, opt ...Option) (cleanup func() error, retURL, dbname string, err error) {
+	db, err := sql.Open("postgres", "postgres://boundary:boundary@127.0.0.1/boundary?sslmode=disable")
```
jimlambrt, line 48: Is an empty dialect a valid parameter?

testing/dbtest/template.go
```diff
@@ -0,0 +1,111 @@
...
+const killconns = `
+select
+  pg_terminate_backend(pg_stat_activity.pid)
+from
+  pg_stat_activity
+where pg_stat_activity.datname = $1
+  and pid <> pg_backend_pid();
+`
+
+func dropDatabase(dbname string) error {
```
jimlambrt, line 95: Is an empty dbname a valid paramter?

testing/dbtest/docker/postgresql.conf
```diff
@@ -0,0 +1,32 @@
+listen_addresses           = '*'
```
jimlambrt, line 1: I have no insight or expertise to comment on these settings overall.

testing/dbtest/docs.go
```diff
@@ -0,0 +1,22 @@
+// Package dbtest provides a way to create a clean database for tests using template databases.
+// For it to function properly a postgres database server must be running with the expected templates.
```
jimlambrt, line 2: Really long lines again.. feel free to ignore this but it's a bit jarring compared with others.

jimlambrt (changes requested):
Amazing work!

As discussed in our zooms, we probably need to update contribution docs about how to run tests, recommended docker settings, troubleshooting common issues when running tests. 

Also, how are new templates built?

tmessi:
> Also, how are new templates built?

The template database is created when the `boundary-sql-test` container is started. It runs [this script](https://github.com/hashicorp/boundary/pull/1482/files#diff-de322499ed681d14bd12ef5dd4c6514fe5241ea55a125583ef4ffab8bdb95bfcR1-R44) which creates the database, applies the migrations, and then configures it to be a template.

internal/cmd/base/option.go
tmessi: fixed

testing/dbtest/docs.go
tmessi: fixed

testing/dbtest/docker/postgresql.conf
tmessi: Do you think this needs some comments to explain the options? Or is it ok to just point people to the official postgres docs if they want to learn about each setting?

testing/dbtest/template.go
tmessi: An empty dbname would cause an error when the `drop database` is run. But given that this is an un-exported function that is only called by `startUsingTempalate` which should always have a non-empty dbname, idk if we need an explicit check here, thoughts?

testing/dbtest/template.go
tmessi: It is not, and really it should only be one of the supported dialects (so it should only be `postgres` currently). I can add a check along with defining the constants you mention in another comment.

testing/dbtest/template.go
tmessi: Hmm...its "valid" in that it works and returns a string of length 0. I don't know why anyone would want to do that though. But given that this is an un-exported function used to just create a random suffix for a test database I don't know if we need a check for that case.

Also, if we already have something to generate a random string that is efficient and doesn't need crypto, I can use that instead of defining this function here, but I wasn't able to find one when I looked.

internal/db/schema/postgres/postgres_test.go
tmessi: added consts and updated the tests to use them

testing/dbtest/template.go
tmessi: Added some unit tests for this package.

testing/dbtest/template.go
tmessi: Added a check for this

tmessi:
> As discussed in our zooms, we probably need to update contribution docs about how to run tests, recommended docker settings, troubleshooting common issues when running tests.

Added some additional configuration options to the makefile and updated the contributing docs with details on how to run the tests, configuration options, and troubleshooting.

internal/cmd/base/dev.go
tmessi: I can look into this. This code path is exercised by test that use the `base.CreateDevDatabase`, but having a dedicated test for it would be a good idea if it isn't too complicated to add (given that there isn't an existing test around this).

testing/dbtest/template.go
jimlambrt: Take a look at ` base62.Random` to see if it fits your requirements.   

When it's cheap (perf wise) to do, then adding a check for invalid params always feels like the right pattern to follow, even for un-exported funcs.  BTW, I agree that passing zero makes no sense... unfortunately, I'm one of those devs that occasionally writes code inadvertently that makes no sense.

testing/dbtest/docker/postgresql.conf
jimlambrt: It might be helpful to give a very succinct reason for each setting so devs understand why and then go explore the docs if they want more info about each setting.

testing/dbtest/template.go
jimlambrt: personally, I'd check for a empty dbname before sending an invalid stmt to the db.

testing/dbtest/template.go
jimlambrt: discussed over slack. ty!

testing/dbtest/docker/postgresql.conf
tmessi: Added some comments to the config

testing/dbtest/template.go
tmessi: added a check for empty dbname

jimlambrt (dismissed):

PR #1483, opened 2021-08-25
ci: Check that generated migration code is up to date

tmessi:
This adds a new job to CircleCI to generate the go code from the sql
migration files and check if there are any changes. This will result in
CI failing if `make migrations` was not run after adding new sql
migration files.


tmessi:
After doing some scoping, I will be refactoring the migrations code some to use [embed](https://pkg.go.dev/embed) instead of this generate command. At which point this ci check will not be needed. I am going to just close this for now but will open another PR for 783c5b7

PR #1490, opened 2021-08-30
Events: add redact/encrypt node for audit events to Eventer

jimlambrt:
Integrate the encrypt.Filter node from the eventlogger into the event.Eventer, so audit events can be redacted/encrypted based on data classifications.

This PR establishes that the default filter is "redaction" of sensitive/secret audit data.  We still need to take some additional steps before we can support other redaction filters like encryption and hmac-sha256 like add a new audit DEK for these operations and initialize it at start up.  But for now the default "redaction" operation is integrated.


internal/observability/event/event.go
```diff
@@ -12,40 +12,39 @@ type (
...
+	UserEmail   string      `json:"email,omitempty" class:"sensitive"`
+	UserName    string      `json:"name,omitempty" class:"sensitive"`
 }
 
 type Request struct {
-	Operation string        `json:"operation"` // std audit field
-	Endpoint  string        `json:"endpoint"`  // std audit field
-	Details   proto.Message `json:"details"`   // boundary field
+	Operation string        `json:"operation,omitempty" class:"public"` // std audit field
+	Endpoint  string        `json:"endpoint,omitempty" class:"public"`  // std audit field
```
jefferai, line 46: What's the difference between endpoint here and path in requestinfo?

internal/observability/event/sink_config.go
```diff
@@ -62,11 +63,22 @@ func (sc *SinkConfig) Validate() error {
...
 	}
+
 	for _, et := range sc.EventTypes {
 		if err := et.Validate(); err != nil {
 			return fmt.Errorf("%s: %w", op, err)
 		}
+		// well, if there's an event type of audit, we need to check the audit
+		// config, if it's optionally provided.  We are intentionally only
+		// checking the FilterOverrides, because there's no way to specify the
+		// wrapper in a config.
```
jefferai, line 74: This brings up a good point. Would we want an embedded `kms` block? Or a single top-level block for type `audit`?

Doing it in an embedded way potentially means different sinks could use different KMSes. OTOH if we allow API-level configuration later then this just becomes the default.

jefferai (approved):
Looks great!

internal/observability/event/event.go
jimlambrt: Great question... and I don't think there is any difference.  Aligning on what fields should be audited is still outstanding and your comment reminds me to restart that effort.  I'm going to dig up Sarah's RFC which referenced this and perhaps try to get alignment via that.  With all that said, I believe this `Request` struct is included in Kit's work that attempts to standarize and align upon a common set of data fields and format for HashiCorp products.

internal/observability/event/sink_config.go
jimlambrt: I'm leaning towards a single top-level audit kms block just for simplicity.  I sort of get the allure of the flexibilty that providing it the embedded way provides... I'm just not sure the added complexity is worth it... unless customers start asking for it in the future.

internal/observability/event/eventer.go
```diff
@@ -259,11 +261,38 @@ func NewEventer(log hclog.Logger, serializationLock *sync.Mutex, serverName stri
...
+				Wrapper: opts.withAuditWrapper,
+			}
+			if len(s.AuditConfig.FilterOverrides) > 0 {
+				overrides := encrypt.DefaultFilterOperations()
+				for k, v := range s.AuditConfig.FilterOverrides {
+					overrides[encrypt.DataClassification(k)] = encrypt.FilterOperation(v)
+				}
+				encryptFilter.FilterOperationOverrides = overrides
+			}
+			id, err := NewId("encrypt-audit")
```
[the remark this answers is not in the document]
jimlambrt: Hum... this is a really excellent question.  It hadn't occured to me that operators may want and non-redacted audit sink.  I just checked the go-eventlogger library and currently there's no way to set a filter to "no operation", even with an explicit override that specifies "no operation".  That's a defect and I'll get a PR open to fix that.   I also think, the node in go-eventlogger should check if all the filters are set to "no operation" and short-circuit the node to immediately return the unfiltered event.   If we go that route, then we don't have to worry about checking/changing things here.  Thanks again for the insightful question!

internal/observability/event/sink_config.go
jefferai: OK -- in that case we can just add per-sink-level overrides later if we want.

internal/observability/event/eventer.go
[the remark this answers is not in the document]
jefferai: I actually wasn't thinking "not redacted" so much as "not encrypted" -- e.g. hmac'd or something. It just seemed based on this that if we were doing something other than encrypting it would still be `encrypt-audit`...?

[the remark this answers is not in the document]
jimlambrt: I think so... it's an `encrypt.Filter` node for auditing.

PR #1491, opened 2021-08-30
refactor(worker): refactors the proxy handler to use endpoint scheme

louisruch:
refactor(worker): refactors the proxy handler to use endpoint scheme


louisruch:
@jefferai PTAL

jefferai (approved):

PR #1492, opened 2021-08-30
Add preferred_address field to API host message

vancluever:
This adds a top-level, read-only "address" field to the host API
message. This field will be used to populate the host address for
plugin-based hosts when their respective host sets are queried.
Attributes will be populated with plugin-specific data that does not
include address data.

Note that this does not affect how static addresses are created or
updated - the API still expects the address in attributes. This keeps
the CRUD workflow consistent with the new plugin-based host catalog and
set types, which take their configuration data in attributes as well.

A small CLI update has been added to output the address at the top-level
as well.


api/hosts/host.gen.go
```diff
@@ -24,6 +24,7 @@ type Host struct {
 	Type              string                 `json:"type,omitempty"`
 	HostSetIds        []string               `json:"host_set_ids,omitempty"`
 	Attributes        map[string]interface{} `json:"attributes,omitempty"`
+	Address           string                 `json:"address,omitempty"`
```
talanknight, line 27: This might be a little unfortunate.  I wonder if there is a way to comment on this generated field to notify SDK users that this is a read only field and that they should use the Option to set the address field or do we just hope the user tests or figures this out...

internal/servers/controller/handlers/hosts/host_service.go
```diff
@@ -456,6 +456,9 @@ func toProto(ctx context.Context, in *static.Host, hostSets []*static.HostSet, o
 			out.HostSetIds = append(out.HostSetIds, hs.GetPublicId())
 		}
 	}
+	if outputFields.Has(globals.HostAddressField) {
```
talanknight, line 459: In addition, please add checks to validateUpdateRequest and validateCreateRequest to ensure that the top level address is not passed in.

talanknight (changes requested):

api/hosts/host.gen.go
vancluever: Now that we've resolved to change this to preferred_address do you think this resolves the issue?

talanknight: Have we decided on renaming it as `prefered_address`?

I wonder if we should have a new read only resource that is called something like "endpoint".   This resource would have the `host_id` that it came from, the `host_set` which is used to determine the address from that `host` and the address selected from that collection of addresses associated with that host.  The reason I think this might make sense is that it feels strange to have a single resource return different values depending on if we have `GET v1/hosts?host_set=foo` and `GET v1/hosts?host_set=bar`.

@jefferai, opinion?

internal/servers/controller/handlers/hosts/host_service.go
vancluever: Done!

vancluever:
Should be up to date based on current discourse, barring conversations about whether or not we want to keep `preferred_address` or take a different approach.

vancluever:
Sounds like we're going a different route with this now so I'm just going to close it. Will keep the branch just in case.

PR #1493, opened 2021-08-31
updating db init without generated resources example

devops-rob:
The existing documentation references this command to initialize the database without generated resources:

```shell
boundary database init -skip-initial-login-role-creation -config /etc/boundary.hcl
```

This is no longer supported and has been superceded with the following:

```shell
boundary database init \
   -skip-auth-method-creation \
   -skip-host-resources-creation \
   -skip-scopes-creation \
   -skip-target-creation \
   -config /etc/boundary.hcl
```

I was getting errors when following the existing example and had to look into the code to figure out its been changed so updating the docs to reflect this change.


malnick (approved):
Thank you!

PR #1500, opened 2021-09-02
fix(connect): Remove erroneous event error message

louisruch:
When a session is terminated by the user, the cancel session path is
triggered, in this path the websocket connection is closed, however, If the
connection is already closed, an EOF is returned. In this situation an
EOF is expected and should not be logged to the user.

Example log line:
```
{
  "id": "ezqk6w9dOQ",
  "source": "https://hashicorp.com/boundary/dev-controller/boundary-dev",
  "specversion": "1.0",
  "type": "error",
  "data": {
    "error": "failed to close WebSocket: failed to read frame header: EOF",
    "error_fields": {},
    "id": "e_n60YPsDkaj",
    "version": "v0.1",
    "op": "worker.(Worker).handleProxy",
    "info": {
      "msg": "error closing client connection"
    }
  },
  "time": "2021-09-02T10:05:46.587513-07:00"
}
```

### DIFF
--- a/internal/servers/worker/handler.go
+++ b/internal/servers/worker/handler.go
@@ -3,6 +3,7 @@ package worker
 import (
 	"context"
 	"errors"
+	"io"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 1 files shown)


vancluever (approved):

jimlambrt (approved):
ty!

jefferai (approved):

PR #1501, opened 2021-09-02
test(scheduler) Remove un-needed test causing intermittent failures

louisruch:


louisruch:
@jimlambrt This is the test that has been causing failures every now and then.  I had a look at this for a little while now, but decided it is probably not worth my time and rather just dropping the test. 
I added this test after we made a change to controller to wait on the scheduler waitgroup, effectively this test is just testing a `wg.Add` and `wg.Done` which doesn't seem really worth much...

louisruch:
Saw this fail on `main` recently, going to merge as is since it just removes an unnecessary test

PR #1502, opened 2021-09-02
Multi-arch Docker Build fixes

jacobmammoliti:
Fixed a few things with the multi-arch build:

- The makefile build would fail because it couldn't find the entrypoint and config file. Fixed that by passing in the docker directory
- Alpine 3.13 was failing the multi-arch builds so I've switched it Alpine 3.11 right now as this successfully builds
- Updated the version from Boundary 0.5.0 to 0.5.1

I've tested the Makefile build locally and had success building arm/amd images to a local registry.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -191,7 +191,7 @@ docker-multiarch-build:
 	docker buildx build \
 	--push \
 	--tag $(IMAGE_TAG) \
-	--tag hashicorp/boundary:latest \
+	--tag hashicorp/boundary:latest \	
 	--build-arg VERSION=$(VERSION) \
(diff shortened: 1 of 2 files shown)


jacobmammoliti:
I've updated the Boundary version to 0.6.1 and updated the alpine image to 3.13.6.

I was running into issues with the arm build with newer versions of Alpine on my Ubuntu machine. I resolved this with the notes here: https://github.com/docker/buildx#building-multi-platform-images. Specifically running the following docker container:

`docker run --privileged --rm tonistiigi/binfmt --install all`

tmessi (approved):

PR #1504, opened 2021-09-03
Add plugin manager scaffolding, add testhostplugin

vancluever:
This adds the scaffolding for the plugin manager for plugin-based hosts.

The manager is a simple method that does the following:

* Checks the repo for the plugin ID to ensure a plugin is present.
* Looks up the name in a built-in factory.
* Loads the plugin client via the factory function.

Additionally, this adds a built-in plugin for testing the plugin-based
host interface. The idea is to use this for E2E testing within the
appropriate plugin-based host repositories.


vancluever:
Draft for now, still depends on a lot of in-flight PRs. But should be helpful in visualizing what it looks like!

vancluever:
Looks like the failing test is due to a missing postgres image:

```
=== CONT  TestDbStuff/postgres:12
    dktest.go:36: Pulling image: postgres:12
    dktest.go:54: Error parsing image pull response: received unexpected HTTP status: 504 Gateway Time-out
    dktest.go:191: Failed to run image: postgres:12 error: Error response from daemon: No such image: postgres:12
--- FAIL: TestDbStuff (0.00s)
    --- FAIL: TestDbStuff/postgres:12 (30.52s)
```

internal/plugin/host/plugin_client.go
```diff
@@ -0,0 +1,72 @@
...
+// These plugins are loaded directly and linked in the Boundary
+// binary, instead of being installed from the embedded filesystem.
+var embeddedPluginClients = map[string]func() plugin.HostPluginServiceClient{
+	// "aws": awsplugin.NewClient,
+	"testing": testhostplugin.NewClient,
+}
+
+// PluginManager is a helper for loading and managing host plugins.
+type PluginManager struct {
+	repo *Repository
```
talanknight, line 22: We probably don't want to store the repo here since I think one of the main use cases for the plugin manager is for the repo in `internal/host/plugin` to be able to use it and nested transactions are problematic and might be hard to track down when doing repo -> repo calls.

Since initially we are going to be only running a very small number of plugins (3 - 5) consider having `NewPluginmanager` take a db object instead and load the db data into memory.  That could also be the time that the plugin processes are started and the clients are created.  Then when the users of the plugin manager requests a plugin (by id, or maybe by name) the manager can return the plugin data and the plugin client w/o having to hit the db or start any processes.

internal/plugin/builtin/testing/testhostplugin/testing.go
```diff
@@ -0,0 +1,439 @@
...
+// plugin system.
+type TestHostPlugin struct{}
+
+// NewClient returns a HostPluginServiceClient for TestHostPlugin.
+func NewClient() plugin.HostPluginServiceClient {
+	return &TestHostPlugin{}
+}
+
+// OnCreateCatalog implements HostPluginServiceClient for TestHostPlugin.
+func (p *TestHostPlugin) OnCreateCatalog(ctx context.Context, in *plugin.OnCreateCatalogRequest, _ ...grpc.CallOption) (*plugin.OnCreateCatalogResponse, error) {
```
talanknight, line 81: The desired behavior of this method seems like it would depend on the test that is using it.  Can we remove the validation logic from these methods and instead let the users of the TestHostPlugin provide the behavior?  If no behavior is provided it could fall back to returning an unimplemented error like [generated grpc services](UnimplementedAccountServiceServer) do.  This means tests could fake a plugin which is hanging or which is returning grpc errors, none of which is possible in this current implementation.

talanknight (changes requested):

vancluever:
Closing this since we're far past this point now.

PR #1505, opened 2021-09-07
wh: Add wh_credential_dimension and bridge tables

tmessi:
This adds new dimension table to record credential information. Since a
session can have multiple credentials, the wh_session_accumulating_fact
is is linked to the credential_dimensions via bridge tables,
wh_credential_group and wh_credential_group_membership. When a session
is inserted, the credental group and credential dimensions will be
upserted. If the group changed in any way, a new group is inserted along
with all of the corresponding credentials.

### DIFF
--- a/internal/db/schema/migrations/postgres/16/03_wh_credential_dimension.up.sql
+++ b/internal/db/schema/migrations/postgres/16/03_wh_credential_dimension.up.sql
@@ -1,0 +1,170 @@
+begin;
+  -- wh_upsert_credential_dimension compares the current vaules in the wh_credential_dimension
+  -- with the current values in the operational tables for the given parameters. IF the values
+  -- between operational tables and the wh_credential_dimension differ, a new row is inserted in
+  -- the wh_credential_dimension to match the current values in the operational tables.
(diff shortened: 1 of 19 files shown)


mgaffney (approved):

internal/db/schema/migrations/postgres/16/03_wh_credential_dimension.up.sql
```diff
@@ -0,0 +1,170 @@
...
+             where a.credential_key = b.credential_key
+               and b.credential_key in (select key from credential_list)
+           )
+       and not exists
+           (
+            select 1
+              from wh_credential_group_membership b
+             where a.credential_key = b.credential_key
+               and b.credential_key not in (select key from credential_list)
+           )
```
talanknight, line 135: Does this where statement using this subquery do anything? We checked on line 121 that a.credential_key is in credential_list.  Since b.credential_key = a.credential_key this where clause can be reduced to confirming that a record doesn't exist where false, no?

talanknight: Maybe this should be 
```
       and not exists
           (
            select 1
              from wh_credential_group_membership b
             where a.credential_group_key = b.credential_group_key
               and b.credential_key not in (select key from credential_list)
           )
```

PR #1507, opened 2021-09-07
Migrate sdk/recovery to api/ and sdk/wrapper to internal

jefferai:
* `recovery` is only used by the API (and the main module, for decryption)
* `wrapper` is only used by the CLI right now

Moving these brings them closer to the code that uses them, and makes
the SDK package a bit more clear in terms of functionality.


jefferai:
Note: once this merges I will update the referenced versions of packages in the various go.mod files.

mgaffney (approved):

PR #1511, opened 2021-09-08
Move wrapper package back to SDK

jefferai:
TF uses this, so SDK is a decent place for it. It doesn't create
dependencies between it and API.


louisruch (approved):
lgtm

PR #1512, opened 2021-09-08
Add secrets to host catalogs API struct

jefferai:
This also changes the address from a plugin from a scalar value to other
fields, starting with an IP address array.


internal/proto/local/plugin/v1/host_plugin_service.proto
```diff
@@ -70,7 +70,7 @@ message OnUpdateCatalogResponse {
 	// It's important that this be returned if it existed previously,
 	// as the returned data overwrites the previously existing copy. If
 	// an error is returned, the update of the persisted data is
-	// aborted.
+	// aborted. If this is nil, no changes are written.
```
talanknight, line 73: This appears to contradict the statement from 2 sentences prior.  Can you update this comment paragraph to be consistent?

talanknight (approved):

internal/proto/local/plugin/v1/host_plugin_service.proto
jefferai: Fixed!

PR #1516, opened 2021-09-08
Create a repository to create, lookup, and list host plugins.

talanknight:
This repository only has the ability to create, list, and lookup host plugins.


vancluever (approved):

vancluever:
Looks good. One thing that I noticed after hitting the ✅ was the presence of the plugin version and executable messages - do these show up anywhere? Maybe we should hold off on these until we actually have versions/executable data in the model?

vancluever (approved):

PR #1525, opened 2021-09-13
feat(credential): Add specific credential types

mgaffney:
This adds interfaces for the following credential types:
- Username and password
- Username and private key
- Certificate and private key

It also adds the following types for redacting secret data:
- Password
- PrivateKey

### DIFF
--- a/internal/credential/credential.go
+++ b/internal/credential/credential.go
@@ -87,3 +87,31 @@ type Revoker interface {
 	// Revoke revokes the dynamic credentials issued for sessionid.
 	Revoke(ctx context.Context, sessionId string) error
 }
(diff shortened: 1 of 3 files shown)


tmessi (approved):

jimlambrt:
This will likely be reverted as soon as we finish with the audit event work which will support: redact, encrypt and hmac1-sha256.  Should we just hold up?

mgaffney:
I spoke with @jimlambrt about his concerns regarding the audit event work and we agreed that it's unlikely anything in this PR will need to be reverted.

louisruch (approved):

PR #1528, opened 2021-09-14
refact: Changes required for gorm v2

jimlambrt:
This PR refactors boundary to work with gorm v2 and includes:

Refactoring due to breaking gorm v2 changes:
- Semantics of the tag: `gorm:"-"`
- How a logger is initialized
- How a db dialect is specified and errors returned from opening a connection
- Postgres driver change from pq to pgx (this includes new error types)
- Use of gorm's inclusion of `sql.Named(...)` for named parameters 
- Different handling of postgres infinity datetimes.

NOTE: This PR is currently pointed at a hashicorp gorm fork, as I work on a required upstream PR.


jimlambrt:
Might as well start reviewing the changes.  We won't merge until we know the disposition of the upstream PR: https://github.com/go-gorm/gorm/pull/4710

internal/db/db.go
```diff
@@ -35,11 +25,30 @@ func (db DbType) String() string {
...
+	case "postgres":
+		return Postgres, nil
+	default:
+		return UnknownDB, fmt.Errorf("%s is an unknown dialect", dialect)
+	}
+}
 
 // Open a database connection which is long-lived.
 // You need to call Close() on the returned gorm.DB
 func Open(dbType DbType, connectionUrl string) (*gorm.DB, error) {
```
tmessi, line 39: I think if there was a struct in the db package that was returned here instead of returning `*gorm.DB` directly, that would eliminate the need to import `gorm` throughout the rest of the code base, especially in a number of tests. Currently, in a number of tests it looks like `gorm` is import along with `db` since the result of `db.Open` is passed to other test helper functions. If instead we had something like:

```go
type DB {
  *gorm.DB
}

func Open(dbType DbType, connectionUrl string) (*DB, error) {
```
Then [this test](https://github.com/hashicorp/boundary/blob/23d2526e9a23c13958799de33cc0a9f1b46d3496/internal/db/schema/migrations/postgres/14/warehouse_user_dim_test.go#L143) for example would not need to import gorm.

internal/cmd/base/servers.go
```diff
@@ -548,20 +549,37 @@ func (b *Server) RunShutdownFuncs() error {
...
-		gorm.LogFormatter = db.GetGormLogFormatter(b.Logger)
-		b.Database.SetLogger(db.GetGormLogger(b.Logger))
+		newLogger := logger.New(
+			db.GetGormLogger(b.Logger),
+			logger.Config{
+				LogLevel: logger.Error, // Log level
+				Colorful: false,        // Disable color
+			},
+		)
+		dbase = dbase.Session(&gorm.Session{Logger: newLogger})
```
tmessi, line 578: Likewise I think this could be refactored into the db package so that this package would not need to import `gorm` at all.

internal/cmd/base/servers.go
```diff
@@ -548,20 +549,37 @@ func (b *Server) RunShutdownFuncs() error {
...
 		return fmt.Errorf("unable to create db object with dialect %s: %s", dialect, "max_open_connections must be unlimited by setting 0 or at least 2")
 	} else {
-		dbase.DB().SetMaxOpenConns(b.DatabaseMaxOpenConnections)
+		underlyingDB, err := dbase.DB()
+		if err != nil {
+			return fmt.Errorf("unable retreive db: %w", err)
+		}
+		underlyingDB.SetMaxOpenConns(b.DatabaseMaxOpenConnections)
 	}
 	b.Database = dbase
```
tmessi, line 569: this line can be removed since `b.Database = dbase` is being set after the `if` and `b.Database` is not used in that `if` block.

internal/cmd/base/servers.go
```diff
@@ -548,20 +549,37 @@ func (b *Server) RunShutdownFuncs() error {
...
 	}
 	if b.DatabaseMaxOpenConnections == 1 {
 		return fmt.Errorf("unable to create db object with dialect %s: %s", dialect, "max_open_connections must be unlimited by setting 0 or at least 2")
 	} else {
-		dbase.DB().SetMaxOpenConns(b.DatabaseMaxOpenConnections)
+		underlyingDB, err := dbase.DB()
+		if err != nil {
+			return fmt.Errorf("unable retreive db: %w", err)
+		}
+		underlyingDB.SetMaxOpenConns(b.DatabaseMaxOpenConnections)
```
tmessi, line 567: If, like my other comment suggests, we create a `db.DB` struct, then this functionality can move moved into the `db` package as a function on that struct, so this package doesn't need to access the underlying `sql.DB` directly, ie:

```go
package db

type DB {
  *gorm.DB
}

func (d *DB) SetMaxOpenConns(n int) error {
  underlyingDB, err := d.DB()
  if err != nil {
     return err
  }

  underlyingDB.SetMaxOpenConns(n)
  return nil
}
```

tmessi (commented):
Overall looks good to me. I do have some suggestions that I think will help reduce the need for other packages to even import `gorm` which feels cleaner to me.

internal/db/db.go
tmessi: this is being addressed in #1535

internal/cmd/base/servers.go
tmessi: this or something like it can be addressed in #1535

internal/cmd/base/servers.go
tmessi: this is can be addressed in #1535

internal/cmd/base/servers.go
jimlambrt: see:  https://github.com/hashicorp/boundary/pull/1535

internal/credential/vault/query.go
```diff
@@ -11,13 +11,13 @@ insert into credential_vault_token (
...
-  $5, -- status
-  $6, -- last_renewal_time
-  wt_add_seconds_to_now($7)  -- expiration_time
+  @1, -- token_hmac
+  @2, -- token
+  @3, -- store_id
+  @4, -- key_id
+  @5, -- status
+  @6, -- last_renewal_time
+  wt_add_seconds_to_now(@7)  -- expiration_time
```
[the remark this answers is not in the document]
jimlambrt: there's some bits that parse the query based on parameters and rebuild bits.... I spent a few min trying to untangle it and decide it was tech debt for future PR.

tmessi (commented):
looks good to me given #1535 but others with better knowledge of the various packages should probably review to check how they there impacted by the refactor.

internal/scheduler/job/repository_job.go
```diff
@@ -171,12 +174,12 @@ func (r *Repository) deleteJob(ctx context.Context, name string, _ ...Option) (i
 		return db.NoRowsAffected, errors.New(ctx, errors.InvalidParameter, op, "missing name")
 	}
 
-	j := allocJob()
+	// j := allocJob()
```
louisruch, line 177: Can probably drop this line instead of commenting it out

louisruch (dismissed):
This looks great, one really minor nit

louisruch (dismissed):

jimlambrt:
Merged by adding a replace `gorm.io/gorm => github.com/hashicorp/gorm v0.0.2` to temporarily point at our fork until the upstream PR is resolved.

PR #1529, opened 2021-09-14
Add preferred endpoints logic to API/handler/DB and some CLI bits

jefferai:


internal/servers/controller/handlers/host_sets/host_set_service.go
```diff
@@ -850,17 +866,23 @@ func validateCreateRequest(req *pbs.CreateHostSetRequest) error {
...
 		switch host.SubtypeFromId(req.GetId()) {
 		case static.Subtype:
 			if req.GetItem().GetType() != "" && req.GetItem().GetType() != static.Subtype.String() {
-				badFields["type"] = "Cannot modify the resource type."
+				badFields[globals.TypeField] = "Cannot modify the resource type."
 			}
 		}
 		return badFields
-	}, static.HostSetPrefix)
+	}, static.HostSetPrefix, plugin.HostSetPrefix)
```
talanknight, line 885: This won't end up working when we support plugin host sets unless we do not have plugin ids with 2 `_`s.  The check for valid id currently checks for a single `_` after the prefix and only alphanumeric values following that.

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
```diff
@@ -129,6 +129,7 @@ begin;
     attributes bytea not null
       constraint attributes_must_not_be_empty
         check(length(attributes) > 0),
+    preferred_endpoints text,
```
mgaffney, line 132: Is this a list?

talanknight, line 132: The benefit of separating this out into a different table is it would allow checks on appropriate prefixes and not nulls.  Having this be a free-form text field which needs to have certain format which is only enforced by go logic has been something we've shied away from in the past.

internal/host/plugin/repository_host_set_test.go
```diff
@@ -60,8 +60,8 @@ func TestRepository_CreateSet(t *testing.T) {
 			name: "invalid-public-id-set",
 			in: &HostSet{
 				HostSet: &store.HostSet{
```
talanknight, line 62: Should we confirm that prefered endpoints column gets written to?

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
jefferai: Yes, it is a delimited list.

internal/servers/controller/handlers/host_sets/host_set_service.go
jefferai: Agreed, however, since we haven't decided on the final format...and if that final format only has one underscore it will work fine. 🤷

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
jefferai: I had it as a different table and had it mostly plumbed through, but when working on the view logic with some help from @jimlambrt we decided it wasn't worth it here. We don't need to check for nulls, validating the data is extremely onerous on the database side (e.g. checking for a valid CIDR or host name, etc.), and it's not a fixed set of data -- it's free text. It's more complicated than what was done in OIDC because it's multiple columns (due to ordering) instead of a single column. It was a huge amount of DB logic, as well as multiple extra calls to the DB to fetch and write and delete values, and it didn't seem worth it just so that the data layer could ensure that there is one of a set of specific prefixes to otherwise free text data.

internal/host/plugin/repository_host_set_test.go
jefferai: Ahh yeah, thanks

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
talanknight: Consider adding a comment on this column to explain why we are breaking from convention a little.

internal/host/plugin/repository_host_set_test.go
jefferai: Added a test!

internal/db/schema/migrations/postgres/16/04_plugin_host.up.sql
jefferai: Resolving for now while I work on the other approach.

jefferai: Resolving for now while I work on the other approach.

jefferai:
@talanknight @mgaffney I've revamped the approach. It's ready for another look -- there are still tests to write/update but that won't fundamentally change the mechanism, so review can probably happen at any point.

internal/host/plugin/repository.go
```diff
@@ -24,7 +25,7 @@ type Repository struct {
 // only be used for one transaction and it is not safe for concurrent go
 // routines to access it. WithLimit option is used as a repo wide default
 // limit applied to all ListX methods.
-func NewRepository(r db.Reader, w db.Writer, kms *kms.Kms, opt ...Option) (*Repository, error) {
+func NewRepository(r db.Reader, w db.Writer, kms *kms.Kms, opt ...host.Option) (*Repository, error) {
```
talanknight, line 28: IMO it's a little strange to put the options 1 level up in the package hierarchy, and only for a portion of the methods on this repo.  This is especially true since this plugin package's option still has withLimit option.  We'll just need to break them out again if Lookup/List behavior diverges.  Is the intent to be able to build a shared interface used by authorize session?

talanknight (commented):
Looks good!  Just a question for discussion purposes and concern about building in memory slices based on priority values retrieved from DB.

internal/host/plugin/repository_host_set.go
```diff
@@ -83,49 +124,150 @@ func (r *Repository) CreateSet(ctx context.Context, scopeId string, s *HostSet,
...
+					ep := strings.Split(val, priorityDelimiter)
+					if len(ep) != 2 {
+						return nil, errors.New(ctx, errors.NotSpecificIntegrity, op, fmt.Sprintf("preferred endpoint %s had unexpected fields", val))
+					}
+					index, err := strconv.Atoi(ep[0])
+					if err != nil {
+						return nil, errors.Wrap(ctx, err, op)
+					}
+					// These are 1-indexed so remove one from the value
+					eps[index-1] = ep[1]
```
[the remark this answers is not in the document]
jefferai: Good catch. I'll make sure to use the given relative priorities bit not assume that the numbers actually match.

internal/host/plugin/repository.go
jefferai: Mainly my intent was to be able to pass options from the service handlers, and we have a lot of places where logic is duplicated (good catch on the limit option, forgot to remove). My thinking was that if we did want to pass an actual type specific option we could simply wrap it within a host option that chains into a call into the typed object.

talanknight: We just want to be careful to not add dependencies for packages `internal/A/B` and `internal/A/C` where code in A depends on code in either `B` or `C`.

jefferai: Yeah -- I figure the more specific code would rely on the less specific code, but not the other way around. We can iterate on this as we go.

jefferai: Removed duplicate option; closing this for now.

internal/host/plugin/repository_host_set.go
```diff
@@ -83,49 +125,172 @@ func (r *Repository) CreateSet(ctx context.Context, scopeId string, s *HostSet,
...
+					}
+					return indexi < indexj
+				})
+				if sortErr != nil {
+					return nil, sortErr
+				}
+				for i, ep := range eps {
+					// At this point they're in the correct order, but we still
+					// have to strip off the priority
+					eps[i] = strings.Split(ep, priorityDelimiter)[0]
```
talanknight, line 268:
```suggestion
					eps[i] = strings.Split(ep, priorityDelimiter)[1]
```

internal/host/plugin/repository_host_set.go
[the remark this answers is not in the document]
jefferai: Fixed!

internal/host/plugin/repository_host_set.go
jefferai: Ugh I think I fell victim to the nest test framework. I definitely ran tests and they definitely worked.

internal/db/schema/migrations/postgres/17/05_preferred_endpoints.up.sql
```diff
@@ -0,0 +1,56 @@
...
+    check(priority > 0),
+  condition text not null
+    check(length(condition) > 0),
+  primary key(host_set_id, priority),
+  unique(host_set_id, condition)
+);
+
+-- host_set_immutable_preferred_endpoint() ensures that endpoint conditions
+-- assigned to host sets are immutable.
+create or replace function
```
mgaffney, line 20:
```suggestion
create function
```
Avoid `create or replace` statements. They can hide errors.

mgaffney (changes requested):

internal/db/schema/migrations/postgres/17/05_preferred_endpoints.up.sql
```diff
@@ -0,0 +1,56 @@
...
+
+create table host_set_preferred_endpoint (
+  create_time wt_timestamp,
+  host_set_id wt_public_id not null
+    constraint host_set_id_fkey
+    references host_set(public_id)
+    on delete cascade
+    on update cascade,
+  priority int not null
+    check(priority > 0),
```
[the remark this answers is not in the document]
jefferai: Fixed

internal/db/schema/migrations/postgres/17/05_preferred_endpoints.up.sql
```diff
@@ -0,0 +1,56 @@
...
+  condition text not null
+    check(length(condition) > 0),
+  primary key(host_set_id, priority),
+  unique(host_set_id, condition)
```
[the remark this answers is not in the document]
jefferai: Sure, I didn't know what it was. (The fact that you said you _think_ it should be that suggests that it maybe isn't as standard as you think it is. :-D  )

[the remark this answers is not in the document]
mgaffney: It is, but I didn't want to sound harsh :)

internal/db/schema/migrations/postgres/17/05_preferred_endpoints.up.sql
```diff
@@ -0,0 +1,56 @@
...
+  create_time wt_timestamp,
+  host_set_id wt_public_id not null
+    constraint host_set_id_fkey
+    references host_set(public_id)
+    on delete cascade
+    on update cascade,
+  priority int not null
+    check(priority > 0),
+  condition text not null
+    check(length(condition) > 0),
```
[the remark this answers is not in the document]
jefferai: Updated!

talanknight (approved):
LGTM!

mgaffney (approved):

PR #1532, opened 2021-09-16
Placeholder Plugin Providers for Repos and added testPlugin struct

talanknight:
This puts into place a way to provide plugins to the repositories.  A TODO is to move from providing the HostPluginServiceServer and change it to the client provided by go plugin once we are using go plugin.

Additionally I created a very light weight testPlugin which allows overwriting specific methods for test cases.  I use this in the CreateSet and CreateCatalog repo tests.

### DIFF
--- a/internal/host/plugin/host_catalog.go
+++ b/internal/host/plugin/host_catalog.go
@@ -19,7 +19,7 @@ type HostCatalog struct {
 	*store.HostCatalog
 	tableName string `gorm:"-"`
 
-	secrets map[string]interface{} `gorm:"-"`
+	secrets []byte `gorm:"-"`
 }
 
 // NewHostCatalog creates a new in memory HostCatalog assigned to a scopeId
@@ -35,15 +35,21 @@ func NewHostCatalog(ctx context.Context, scopeId, pluginId string, opt ...Option
 			Name:        opts.withName,
 			Description: opts.withDescription,
 		},
-		secrets: opts.withSecrets,
 	}
 
 	if opts.withAttributes != nil {
 		attrs, err := json.Marshal(opts.withAttributes)
 		if err != nil {
-			return nil, errors.Wrap(ctx, err, op, errors.WithCode(errors.InvalidParameter))
+			return nil, errors.Wrap(ctx, err, op, errors.WithCode(errors.InvalidParameter), errors.WithMsg("marshaling attributes"))
 		}
 		hc.Attributes = attrs
+	}
+	if opts.withSecrets != nil {
+		secrets, err := json.Marshal(opts.withSecrets)
+		if err != nil {
+			return nil, errors.Wrap(ctx, err, op, errors.WithCode(errors.InvalidParameter), errors.WithMsg("marshaling secrets"))
+		}
+		hc.secrets = secrets
 	}
 	return hc, nil
 }
@@ -54,8 +60,7 @@ func allocHostCatalog() *HostCatalog {
 	}
 }
 
-// clone provides a deep copy of the HostCatalog with the exception of the
-// secret.  The secret shallow copied.
+// clone provides a deep copy of the HostCatalog.
 func (c *HostCatalog) clone() *HostCatalog {
 	cp := proto.Clone(c.HostCatalog)
 
@@ -63,10 +68,8 @@ func (c *HostCatalog) clone() *HostCatalog {
 		HostCatalog: cp.(*store.HostCatalog),
 	}
 	if c.secrets != nil {
-		newSecret := make(map[string]interface{}, len(c.secrets))
-		for k, v := range c.secrets {
-			newSecret[k] = v
-		}
+		newSecret := make([]byte, len(c.secrets))
+		copy(newSecret, c.secrets)
 		hc.secrets = newSecret
 	}
 	return hc
--- a/internal/host/plugin/repository_host_catalog.go
+++ b/internal/host/plugin/repository_host_catalog.go
@@ -2,6 +2,7 @@ package plugin
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 
 	"github.com/hashicorp/boundary/internal/db"
@@ -9,6 +10,9 @@ import (
 	"github.com/hashicorp/boundary/internal/kms"
 	"github.com/hashicorp/boundary/internal/oplog"
 	hostplg "github.com/hashicorp/boundary/internal/plugin/host"
+	pb "github.com/hashicorp/boundary/sdk/pbs/controller/api/resources/hostcatalogs"
+	plgpb "github.com/hashicorp/boundary/sdk/pbs/plugin"
+	"google.golang.org/protobuf/types/known/structpb"
 )
 
 // CreateCatalog inserts c into the repository and returns a new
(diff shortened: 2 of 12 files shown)


internal/host/plugin/host_catalog.go
```diff
@@ -35,16 +35,22 @@ func NewHostCatalog(ctx context.Context, scopeId, pluginId string, opt ...Option
...
 	if opts.withAttributes != nil {
 		attrs, err := json.Marshal(opts.withAttributes)
 		if err != nil {
-			return nil, errors.Wrap(ctx, err, op, errors.WithCode(errors.InvalidParameter))
+			return nil, errors.Wrap(ctx, err, op, errors.WithCode(errors.InvalidParameter), errors.WithMsg("marshaling attributes"))
 		}
 		hc.Attributes = attrs
 	}
+	if opts.withSecrets != nil {
+		secrets, err := json.Marshal(opts.withSecrets)
```
jefferai, line 48: Given that secrets comes in as a structpb.Struct and is passed to plugins as a structpb.Struct, why are we using JSON to marshal it? Seems like we should just proto.Marshal the struct.

talanknight: I have fixed this.  It was an experiment I was playing with in this PR but never reverted.  Thanks for catching it.

internal/host/plugin/repository_host_catalog.go
```diff
@@ -59,23 +62,38 @@ func (r *Repository) CreateCatalog(ctx context.Context, c *HostCatalog, _ ...Opt
...
+	plgHc, err := toPluginCatalog(ctx, c)
 	if err != nil {
-		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("unable to get oplog wrapper"))
+		return nil, errors.Wrap(ctx, err, op)
 	}
-	dbWrapper, err := r.kms.GetWrapper(ctx, c.ScopeId, kms.KeyPurposeDatabase)
+	plgResp, err := plgClient.OnCreateCatalog(ctx, &plgpb.OnCreateCatalogRequest{Catalog: plgHc})
 	if err != nil {
-		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("unable to get db wrapper"))
+		return nil, errors.Wrap(ctx, err, op)
```
jefferai, line 75: The error should be ignored if it's an unimplemented error. There's no particular reason that plugins _must_ include an OnCreateCatalog function; if they don't the Unimplemented one will return an error that we can safely ignore.

As a corollary this means there needs to be a nil check on `plgResp` before using it in code below.

internal/host/plugin/repository_host_catalog.go
```diff
@@ -59,23 +62,38 @@ func (r *Repository) CreateCatalog(ctx context.Context, c *HostCatalog, _ ...Opt
...
 	}
+	persistedData := plgResp.GetPersisted().GetData()
 
 	var hcSecret *HostCatalogSecret
-	if c.secrets != nil {
-		// TODO: Create the secret using the returned value from the call to the plugin.
-		hcSecret, err = newHostCatalogSecret(ctx, id, c.secrets)
+	if persistedData != nil {
+		hcSecret, err = newHostCatalogSecret(ctx, id, persistedData.AsMap())
```
jefferai, line 81: Rather than converting to a map and then json marshaling, we should proto marshal directly from the Struct.

internal/host/plugin/repository_host_catalog.go
```diff
@@ -164,3 +181,91 @@ func (r *Repository) ListCatalogs(ctx context.Context, scopeIds []string, opt ..
 	}
 	return hostCatalogs, nil
 }
+
+// getCatalog retrieves the *HostCatalog with the provided id.  If it is not found or there
+// is an problem getting it from the database an error is returned instead.
+func (r *Repository) getCatalog(ctx context.Context, id string) (*HostCatalog, error) {
```
jefferai, line 187: Maybe a TODO, but it feels like rather than these two functions it should be a view.

internal/host/plugin/repository_host_catalog.go
```diff
@@ -164,3 +181,91 @@ func (r *Repository) ListCatalogs(ctx context.Context, scopeIds []string, opt ..
...
+	if in == nil {
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "nil storage plugin")
+	}
+	hc := &pb.HostCatalog{
+		Id:      in.GetPublicId(),
+		ScopeId: in.GetScopeId(),
+	}
+	if in.GetAttributes() != nil {
+		attrs := map[string]interface{}{}
+		if err := json.Unmarshal(in.GetAttributes(), &attrs); err != nil {
```
jefferai, line 251: Same here -- shouldn't the attributes be getting stored as proto marshalled instead of json marshalled?

internal/host/plugin/repository_host_catalog.go
talanknight: I've added a TODO.

internal/host/plugin/repository_host_catalog.go
talanknight: We no longer error on unimplemented.  Added tests for this as well.

internal/host/plugin/host_catalog.go
```diff
@@ -28,23 +28,22 @@ type HostCatalog struct {
 func NewHostCatalog(ctx context.Context, scopeId, pluginId string, opt ...Option) (*HostCatalog, error) {
 	const op = "plugin.NewHostCatalog"
 	opts := getOpts(opt...)
+
+	attrs, err := proto.Marshal(opts.withAttributes)
```
jefferai, line 32: Not sure what the behavior of marshal is if opts.withAttributes is nil, but do we need to perform the same "check if nil, if so, set to allocated zero-length byte slice" behavior here?

internal/host/plugin/host_catalog.go
```diff
@@ -58,16 +57,16 @@ func allocHostCatalog() *HostCatalog {
...
 	}
-	if c.secrets != nil {
-		newSecret := make(map[string]interface{}, len(c.secrets))
-		for k, v := range c.secrets {
-			newSecret[k] = v
-		}
-		hc.secrets = newSecret
+	// proto.Clone will convert slices with length and capacity of 0 to nil.
+	// Fix this since gorm treats empty slices differently than nil.
+	if c.Attributes != nil && len(c.Attributes) == 0 && hc.Attributes == nil {
```
jefferai, line 68: This check is different between here and host set below (in host sets there is no zero-length check).

internal/servers/controller/handlers/host_sets/host_set_service.go
```diff
@@ -756,16 +756,13 @@ func toProto(ctx context.Context, in host.Set, hosts []host.Host, opt ...handler
...
+			err := proto.Unmarshal(h.Attributes, attrs)
 			if err != nil {
 				return nil, errors.Wrap(ctx, err, op)
 			}
-			if len(attrs) > 0 {
-				out.Attributes, err = structpb.NewStruct(attrs)
-				if err != nil {
-					return nil, errors.Wrap(ctx, err, op)
-				}
+			if len(attrs.AsMap()) > 0 {
```
jefferai, line 764: Same comment.

internal/host/plugin/host_catalog.go
talanknight: proto.Marshal returns nil, nil if the input is nil.

internal/host/plugin/host_catalog.go
talanknight: Woops, thank you for catching that.

jefferai (approved):

PR #1533, opened 2021-09-17
chore(targets): remove deprecated fields on target api/cli/sdk and service

louisruch:
Removes credential libraries with respect to target resources


louisruch:
@jefferai @talanknight PTAL this removes deprecated fields from the target domain
@mgaffney @tmessi FYI/PTAL

louisruch:
Tagging @randallmorey to confirm the UI has moved off libraries and sets

randallmorey:
@louisruch working on this update today.  To clarify, the UI will change to `*-sources`, but we do not support multiple source types just yet.

louisruch:
> @louisruch working on this update today. To clarify, the UI will change to `*-sources`, but we do not support multiple source types just yet.

talanknight (commented):
Do we need to do a database migration for any grants which were previously on sets and libraries and change them to sources?

jefferai:
I'm not sure re: migrations, only because it might take a bit if people have a lot of roles. OTOH, people are unlikely to have a lot of roles right now. :-)

So yeah, I _think_ we do want to migrate, and disallow new grants from being saved using the deprecated actions.

louisruch:
> I'm not sure re: migrations, only because it might take a bit if people have a lot of roles. OTOH, people are unlikely to have a lot of roles right now. :-)
> 
> So yeah, I _think_ we do want to migrate, and disallow new grants from being saved using the deprecated actions.

I removed the `*-credential_library` actions from the [map](https://github.com/hashicorp/boundary/blob/73df4bc2b39bf9b65ae0bdcb1a1b33df353993e7/internal/types/action/action.go#L58) so they would be disallowed.  However, I noticed that `host-sets` actions were shared with `host_set` repo, so I actually removed the `DEPRECATED` comment on them

randallmorey:
Looks like FE will need Monday to complete this work.

randallmorey:
FE has now migrated to *-sources hashicorp/boundary-ui#770

louisruch:
@jefferai @talanknight PTAL

internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -722,7 +605,12 @@ func (s Service) AddTargetCredentialSources(ctx context.Context, req *pbs.AddTar
 	}
 	authResults := s.authResult(ctx, req.GetId(), action.AddCredentialSources)
 	if authResults.Error != nil {
-		return nil, authResults.Error
+		// TODO AddCredentialLibraries was deprecated but grant actions were never migrated
+		// remove this check once actions have been migrated
```
johanbrandhorst, line 609: I know you mentioned this, but do we have a ticket for migrating grant actions?

johanbrandhorst (approved):

PR #1534, opened 2021-09-17
chore(authenticate): remove deprecated authenticate:login and credentials field

louisruch:
Removes `authenticate:login` as well as the deprecated `credentials` field in `authenticate`


louisruch:
@jefferai @talanknight PTAL this removes deprecated fields from the authenticate domain
@randallmorey just confirming there will be no UI issues with this

randallmorey:
@louisruch can you summarise the change in plain English?  I mostly understand the Go but want to confirm.  It looks like the `:authenticate:login` endpoint goes away and the `credentials` field is removed.

louisruch:
> @louisruch can you summarise the change in plain English? I mostly understand the Go but want to confirm. It looks like the `:authenticate:login` endpoint goes away and the `credentials` field is removed.

@randallmorey yes, so the `:authenticate:login` endpoint was deprecated in `0.2.0` https://github.com/hashicorp/boundary/blob/main/CHANGELOG.md#deprecationschanges-4 this undoing the 0.1.6 deprecation of `:authenticate`
We also deprecated the `credentials` field in `:authenticate` to instead use `attributes` I believe this was done to support OIDC

randallmorey:
> > @louisruch can you summarise the change in plain English? I mostly understand the Go but want to confirm. It looks like the `:authenticate:login` endpoint goes away and the `credentials` field is removed.
> 
> @randallmorey yes, so the `:authenticate:login` endpoint was deprecated in `0.2.0` https://github.com/hashicorp/boundary/blob/main/CHANGELOG.md#deprecationschanges-4 this undoing the 0.1.6 deprecation of `:authenticate`
> We also deprecated the `credentials` field in `:authenticate` to instead use `attributes` I believe this was done to support OIDC also see 0.2.0 changelog

Thank you.  Approved!  FE is on the latest.

randallmorey (approved):

jefferai (dismissed):

PR #1535, opened 2021-09-17
refact: Add db.DB wrapper and refact all test fixtures to use it.

jimlambrt:
This PR will allow us to stop leaking the gorm abstraction within
our unit tests.

This PR is a fast-follower to https://github.com/hashicorp/boundary/pull/1528. It cannot be merged until that PR is accepted and then it should be rebased against main.


internal/cmd/base/servers.go
```diff
@@ -575,7 +575,7 @@ func (b *Server) ConnectToDatabase(dialect string) error {
 				Colorful: false,        // Disable color
 			},
 		)
-		dbase = dbase.Session(&gorm.Session{Logger: newLogger})
+		dbase.DB = dbase.Session(&gorm.Session{Logger: newLogger})
```
tmessi, line 578: This is still causing `gorm` and `gorm/logger` to be imported here. I think that can be elimnated by moving this logger configuration into the `db` package. Probably by adding some `Options` to `db`, i.e.:

```go
opts := []db.Option
if os.Getenv("BOUNDARY_DISABLE_GORM_FORMATTER") == "" {
  opts = append(opts, db.WithGormFormatter(b.Logger))
}
dbase, err := db.Open(dbType, b.DatabaseUrl, opts...)
```

then in `db`:
```go
package db

// i dont think i have this exactly correct, but hopefully its close enough to get the idea
func Open(dbType DbType, connectionUrl string, options ...Option) (*DB, error) {
  // ...
  if opts.WithGormFormatter != nil {
    newLogger := logger.New(
      db.GetGormLogger(opts.WithGormFormatter),
      logger.Config{
        LogLevel: logger.Error,
        Colorful: false,
      },
    )
    db.DB = db.Session(&gorm.Session{Logger: newLogger})
  }

  // ...
}

File: internal/cmd/base/servers.go [hunk 2]
```diff
@@ -560,7 +560,7 @@ func (b *Server) ConnectToDatabase(dialect string) error {
 	if b.DatabaseMaxOpenConnections == 1 {
 		return fmt.Errorf("unable to create db object with dialect %s: %s", dialect, "max_open_connections must be unlimited by setting 0 or at least 2")
 	} else {
-		underlyingDB, err := dbase.DB()
+		underlyingDB, err := dbase.SqlDB()
```
tmessi, line 563: likewise, the setting of MaxOpenConnections could be moved into an Option for the call to `db.Open`

internal/cmd/base/servers.go
```diff
@@ -620,7 +620,7 @@ func (b *Server) CreateGlobalKmsKeys(ctx context.Context) error {
 
 func (b *Server) DestroyDevDatabase() error {
 	if b.Database != nil {
-		underlyingDB, err := b.Database.DB()
+		underlyingDB, err := b.Database.SqlDB()
 		if err != nil {
 			return err
 		}
```
tmessi, line 626: I think this `Close` logic should probably be moved into a method on `db.DB`, ie:

```go
func (d *DB) Close() error {
  underlyingDB, err := d.SqlDB()
  if err != nil {
    return err
  }
  return underlying.Close()
}
```
}

internal/db/db.go
```diff
@@ -34,9 +35,23 @@ func StringToDbType(dialect string) (DbType, error) {
...
 }
 
+// DB is a wrapper around the ORM
+type DB struct {
+	*gorm.DB
+}
+
+// SqlDB returns the underlying driver.
+//
+// Note: This func is not named DB(), because our chosen of ORM (gorm) which is
```
tmessi, line 45: nitpick: `s/chosen/choice/`

internal/db/db.go
```diff
@@ -34,9 +35,23 @@ func StringToDbType(dialect string) (DbType, error) {
...
+// embedded for various reasons, already exports a DB() func and it's base type
+// is also an export DB type
+func (d *DB) SqlDB() (*sql.DB, error) {
+	return d.DB.DB()
```
tmessi, line 49: I think this should check if `d.DB` is `nil` and return an error if it is.

internal/db/db.go
jimlambrt: ty!

internal/db/db.go
```diff
@@ -34,9 +35,23 @@ func StringToDbType(dialect string) (DbType, error) {
 	}
 }
 
+// DB is a wrapper around the ORM
+type DB struct {
+	*gorm.DB
+}
+
+// SqlDB returns the underlying driver.
```
tmessi, line 43: nitpick: I think technically this isn't the "driver" its the `sql.DB` which its [docs](https://pkg.go.dev/database/sql#DB) describe as:

> DB is a database handle representing a pool of zero or more underlying connections.

The drivers are [these packages](https://github.com/golang/go/wiki/SQLDrivers)

internal/db/db.go
jimlambrt: ty!

internal/cmd/base/servers.go
jimlambrt: ty!

internal/db/db.go
```diff
@@ -34,9 +37,42 @@ func StringToDbType(dialect string) (DbType, error) {
...
+	if err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+	return underlying.Close()
+}
+
+// Open a database connection which is long-lived. The options of
+// WithGormFormatter and WithMaxOpenConnections are supported.
+//
+// Note: you need to call Close() on the returned DB
```
tmessi, line 74: interestingly, the `sql` docs for `Close` state:

> It is rare to Close a DB, as the DB handle is meant to be long-lived and shared between many goroutines.

And for `Open`:

> It is rarely necessary to close a DB.

And I think the only spot we are calling Close (outside of tests) in is `DestroyDevDatabase` to ensure the connections are closed before destroying the database.

jimlambrt: updated comment/doc

tmessi (approved):

tmessi (approved):

PR #1540, opened 2021-09-17
Update various dependencies

jefferai:
- [x] go-bexpr
- [x] strcase
- [x] pretty
- [x] mapstructure
- [x] go-proxyproto


jefferai:
Note for posterity: I think there are still some go-proxyproto PRs I'd like to see merged before we start using it, but as we aren't actually wrapping listeners with it right now I'm happy to merge an update as it makes dependabot happy.

PR #1541, opened 2021-09-17
Update various grpc deps

jefferai:
Jeff:

- [x] grpc
- [x] grpc-gateway

Todd:

- [x] grpc
- [x] grpc-gateway


jefferai (commented):
I have a slight concern about https://github.com/grpc-ecosystem/grpc-gateway/pull/2314 but I'm not sure honestly if it will cause problems or not. So I think better to get this in earlier and use it in testing and see if things break.

talanknight (approved):
For GRPC gateway, We discussed how https://github.com/grpc-ecosystem/grpc-gateway/pull/2314 shouldn't impact us since we use a custom error handler.  Most other changes seem reasonable.

GRPC looks ok.

PR #1545, opened 2021-09-17
Bump dockertest and atomic

jefferai:


jefferai:
Checked out the changes, looks good.

PR #1546, opened 2021-09-19
Fixes multiarch build context

mister2d:
Replaces `.` with `docker/` to correct the build context relative path.


jefferai:
@tmessi PTAL?

tmessi:
@mister2d thanks for the submission. It looks like issue was being addressed in #1502 which was just merged in.

PR #1547, opened 2021-09-20
fix (credential/vault): Correct typo in error msg

jimlambrt:


louisruch (approved):
LGTM

jefferai:
Looks good, but before you get rid of it, please submit a definition for "repsonse"

jimlambrt:
> Looks good, but before you get rid of it, please submit a definition for "repsonse"

It's a compound word: rep + sonse which means representing lucky or prosperity.  Since the error msg was "no repsonse", I believe the meaning to be: "no representation of prosperity from vault"

PR #1550, opened 2021-09-20
tests(scheduler): fix unit test deadlock

louisruch:
If a new job is started as the test successfully compeltes and begins
the scheduler shutdown process.  The new job will deadlock waiting on
channel reads.


jimlambrt (approved):
ty!

PR #1552, opened 2021-09-20
Run `make tools` and `make gen` since deps are updated

talanknight:


jefferai:
Thanks, I'd done `make gen` but had forgotten to run `make tools` first :-/

PR #1559, opened 2021-09-23
Add authorize session for plugin based host sets.

talanknight:
Added a new Endpoint struct at the top level host package level to capture host, set, and address chosen.
Temporarily add a host_plugin_host table to capture the hosts returned by the plugins.  This ensures the datawarehouse is able to query the table to decide what host was chosen.  The view used to update the datawarehouse dimension table provides rows with host set ids and hosts which are not in that host set.  This is ok since the function which updates the dimension table provides the host set and host id chosen so that relationship doesn't need to be captured in the view itself.

TODO: The datawarehouse does not accurately capture hosts with multiple addresses yet.


internal/host/plugin/repository_host_set.go
```diff
@@ -338,3 +339,117 @@ func toPluginSet(ctx context.Context, in *HostSet) (*pb.HostSet, error) {
 	}
 	return hs, nil
 }
+
+// Endpoints provides all the endpoints available for a given set id.
+// An error is returned if the set, related catalog, or related plugin are
+// unable to be retrieved.  If a host does not contain an addressible endpoint
+// it is not included in the resulting slice of endpoints.
+func (r *Repository) Endpoints(ctx context.Context, setId string) ([]*host.Endpoint, error) {
```
jefferai, line 347: Could this take in a `[]string` of sets so that we could potentially reduce the number of calls to the DB? ListHosts below already takes in multiple sets.

talanknight: Changed to allow a slice of set ids to be passed to it.  I didn't have time to add many tests to exercise that functionality.

PR #1563, opened 2021-09-24
Remove Prefix Id and Plugin Name fields

talanknight:
Creating catalogs now require either the plugin id or the name of the plugin resource (but not both).  Host Catalogs and Host Sets API resources now have a read only plugin field populated with plugin info.  Host Catalogs have a plugin id field as a top level field.


internal/host/plugin/ids.go
```diff
@@ -19,24 +18,22 @@ func init() {
 // PublicId prefixes for the resources in the plugin package.
 const (
 	// TODO: Pull these out of being constants and have them derivable at run time.
-	HostCatalogPrefix = "hcplg"
-	HostSetPrefix     = "hsplg"
+	HostCatalogPrefix = "hc"
```
jefferai, line 21: I think rather than have this here we should just move it to `globals` somewhere (maybe a new file) so that as other resources are modified they can all reference the same spot.

internal/proto/local/controller/api/services/v1/host_catalog_service.proto
```diff
@@ -103,6 +103,9 @@ message ListHostCatalogsResponse {
 
 message CreateHostCatalogRequest {
   api.resources.hostcatalogs.v1.HostCatalog item = 1;
+  // As an alternative to providing the plugin id in the provided HostCatalog,
```
jefferai, line 106: Could we just look at the `name` field in the host catalog iff id is empty?

talanknight: Then how would we set the name of the host catalog to something different from the plugin?

internal/host/plugin/ids.go
talanknight: I started down that path but decided to wait until we did this more broadly.  I'm hoping to avoid discussions around what structure we want to change to since this is already a decently sized PR.

jefferai: OK

internal/proto/local/controller/api/services/v1/host_catalog_service.proto
jefferai: Ah, sorry, right.

internal/perms/output_fields_test.go
```diff
@@ -253,6 +253,8 @@ func Test_ACLSelfOrDefault(t *testing.T) {
 				globals.IdField:                          true,
 				globals.ScopeField:                       true,
 				globals.ScopeIdField:                     true,
+				globals.PluginField:                       true,
+				globals.PluginIdField:                     true,
```
jefferai, line 257: `make fmt`? :-)

jefferai (approved):
Looks great!

PR #1568, opened 2021-09-28
Add host IDs to plugin host sets read response

jefferai:
This is not an ideal implementation but as we address some of the underlying database bits we can make this nicer.

I've also included a change to the loopback plugin to allow it to receive a slice of multiple hosts to return for a given host set.


internal/host/plugin/repository_host_set.go
```diff
@@ -153,28 +154,67 @@ func (r *Repository) CreateSet(ctx context.Context, scopeId string, s *HostSet,
 	return returnedHostSet, plg, nil
 }
 
-// LookupSet will look up a host set in the repository and return the host
-// set. If the host set is not found, it will return nil, nil.
-// All options are ignored.
-func (r *Repository) LookupSet(ctx context.Context, publicId string, opt ...host.Option) (*HostSet, *hostplugin.Plugin, error) {
+// LookupSet will look up a host set in the repository and return the host set,
+// as well as host IDs that match. If the host set is not found, it will return
+// nil, nil, nil, nil. All options are ignored.
```
talanknight, line 159: All options are no longer ignored.

internal/host/plugin/repository_host_set.go
```diff
@@ -153,28 +154,67 @@ func (r *Repository) CreateSet(ctx context.Context, scopeId string, s *HostSet,
...
+		if !ok {
+			return nil, nil, nil, errors.New(ctx, errors.InvalidParameter, op, fmt.Sprintf("no plugin found for plugin id %s", plg.GetPublicId()))
+		}
+		resp, err := plgClient.ListHosts(ctx, &plgpb.ListHostsRequest{
+			Sets: []*hspb.HostSet{plgSet},
+		})
+		switch {
+		case err != nil:
+			// If it's just not implemented, e.g. for tests, don't error out, return what we have
+			if !strings.Contains(err.Error(), codes.Unimplemented.String()) {
```
talanknight, line 201: status.Code will automatically try to convert to a grpcstatus and if it us unable it will call the code Unknown. 
so assuming the plugin server uses the grpc status errors you can just write `status.Code(err) != codes.Unimplemented`

talanknight (approved):
Looks good!

internal/host/plugin/repository_host_set.go
jefferai: Oops yeah.

internal/host/plugin/repository_host_set.go
jefferai: Ah cool, thanks!

PR #1571, opened 2021-09-30
Move host ID lookup to shared function

jefferai:
This allows us to perform a pre-flight host ID lookup when creating a
host set. This has two benefits:

* We don't create the host-set if a call with the given filter errors
  out
* We can return a set of matching host IDs in the response from a create
  call, instead of having to then issue a separate read response

### DIFF
--- a/internal/host/plugin/repository_host_catalog.go
+++ b/internal/host/plugin/repository_host_catalog.go
@@ -264,8 +264,9 @@ func toPluginCatalog(ctx context.Context, in *HostCatalog) (*pb.HostCatalog, err
 		return nil, errors.New(ctx, errors.InvalidParameter, op, "nil storage plugin")
 	}
 	hc := &pb.HostCatalog{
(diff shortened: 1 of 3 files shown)


vancluever (approved):
So my only main concern here is that this could perform a real lookup which could ultimately take a long period of time and hold up creation. This sounded like more of a concern in my head, but typing it out makes it seem like less of a concern (especially if we were doing this already anyway). So ✅ 

PS: WRT dry runs, the AWS plugin does do a dry run on creation (see [here](https://github.com/hashicorp/boundary-plugin-host-aws/blob/d98e41f91c2997dd76ab54c7faa6a6552bd21a83/plugin/plugin.go#L252-L267)), but this doesn't seem to catch all errors (for some reason, invalid filter names do _not_ get caught).

jefferai:
Based on discussion with @vancluever going to close this for now. This may not be worth the tradeoffs.

PR #1574, opened 2021-10-01
Add mapping to Vault credential library create path

louisruch:


louisruch:
@mgaffney @tmessi This is a draft PR with support on the Vault credential library create path.  I wanted to get eyes on before I move too far down this path to ensure consensus on naming and locations of some interfaces

PR #1576, opened 2021-10-03
refactor (controller): Use a grpc server for the grpc-gateway

jimlambrt:
Just sharing some WIP bits.


internal/servers/controller/handler.go
```diff
@@ -78,127 +79,169 @@ func handleGrpcGateway(c *Controller, props HandlerProperties) (http.Handler, er
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to create user handler service: %w", err)
-	}
-	if err := services.RegisterUserServiceHandlerServer(ctx, mux, us); err != nil {
-		return nil, fmt.Errorf("failed to register user service handler: %w", err)
+
+	currentServices := c.gatewayServer.GetServiceInfo()
+	dialOptions := gatewayDialOptions(c.gatewayListener)
+
+	if _, ok := currentServices["controller.api.services.v1.HostCatalogService"]; !ok {
```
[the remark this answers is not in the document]
jimlambrt: ty

internal/proto/local/controller/auth/v1/auth.proto
```diff
@@ -0,0 +1,42 @@
...
+  string user_id_override = 80;
+
+  // disable_authz_failures for the request (helpful for tests)
+  bool disable_authz_failures = 90;
+
+  // disable_auth_entirely for the request (helpful for tests)
+  bool disable_auth_entirely = 100;
+
+  // ticket is a unique id that allows the grpc-gateway to verify that the info
+  // came from it's companion http proxy
```
jefferai, line 40:
```suggestion
  // came from its companion http proxy
```

internal/proto/local/errors/v1/errors.proto
```diff
@@ -0,0 +1,40 @@
...
+  uint32 code = 10;
+
+  // msg for the error
+  string msg = 20;
+
+  // op represents the operation raising/propagating an error and is optional.
+  string op = 30;
+
+  oneof wrapped {
+    // err is a wrapped Err and i
```
jefferai, line 22: and i..............?

internal/servers/controller/auth/auth.go
```diff
@@ -67,6 +69,39 @@ type RequestInfo struct {
 	DisableAuthEntirely  bool
 }
 
+// ToPb will convert a RequestInfo to a protobuf version that can be serialized
+// across the wire to the grpc-gateway server.
+func (r *RequestInfo) ToPb() *authpb.RequestInfo {
```
jefferai, line 74: Rather than do conversions, should we just remove the RequestInfo struct in this file and only use the pb message everywhere?

jimlambrt: done.

PR #1592, opened 2021-10-13
internal/libs/patchstruct: add package for patching structpb

vancluever:
This adds a small helper package for patching *structpb.Struct values
using the same patterns that our APIs use for updating records. The
intent of this package is to provide a fallback mechanism for working
with these values when said record update behavior is unavailable, such
as when working with plugin-based host subtypes where the exact
structure of the subtype's attributes is unavailable.

patchstruct.Patch follows the following logic:

* The source (src) map is merged into the destination map (dst).
* Values are overwritten by the source map if they exist in both.
* Values are deleted from the destination if they are set to null in the
  source.
* Maps are recursively applied, meaning that a nested map at key "foo" in
  the destination would be patched with a map at key "foo" in the
  source.
* A map in the destination is overwritten by a non-map in the source,
  and a non-map in the destination is overwritten by a map in the
  source.

Patch returns the updated map as a copy, dst and src are not altered.

UPDATE: `Patch` has been renamed to `PatchStruct` to make room for `PatchBytes`, which works in a similar fashion, but patches the protobuf-encoded binary data directly.

### DIFF
--- a/internal/libs/patchstruct/patchstruct_test.go
+++ b/internal/libs/patchstruct/patchstruct_test.go
@@ -143,21 +143,10 @@ func TestPatchBytes(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			require := require.New(t)
 			dst, src := mustMarshal(tc.dst), mustMarshal(tc.src)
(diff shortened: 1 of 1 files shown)


vancluever:
Depends on #1592.

PR #1627, opened 2021-10-21
Backport of Adding reference architecture docs into stable-website

randallmorey:
## Backport

This PR is auto-generated from #1625 to be assessed for backporting due to the inclusion of the label website/cherrypick.

The below text is copied from the body of the original PR.

---

_Original PR had no description content._


jefferai:
Weirdly this is not showing as completely conflicted even though stable-website is now equivalent to main. Is that because everything has changed again since this was created?

randallmorey:
I am confused as to why this is autogenerated in my name.  I didn't participate in the original PR.  I'll just close this for now.

PR #1634, opened 2021-10-23
internal/host/plugin: Add UpdateCatalog

vancluever:
This adds support for updating plugin-based host catalogs.

The workflow is:

* The existing catalog is fetched.
* Any attribute changes are patched against it, along with other changes
  such as name and description.
* Both the existing catalog, and the state of the catalog as it will
  exist post-update, along with any persisted secrets, are sent to the
  plugin via an OnUpdateCatalog request.

Upon success, the following is done:

* The record is updated if non-secret updates need to be made.
* Secrets are updated if they still exist.
* If an empty secret set was sent back, secrets are instead deleted.
* If the secrets field was instead null, secret update is skipped
  altogether.

There has also been a lot of helper code added to the tests to help make
the update tests more readable and reduce boilerplate. While the create
tests have not been updated to use these new helpers, it might be a good
idea to do so for consistency.


internal/host/plugin/repository_host_catalog.go
```diff
@@ -282,8 +549,8 @@ func (r *Repository) getCatalog(ctx context.Context, id string) (*HostCatalog, e
 }
 
 // getPersistedDataForCatalog returns the persisted data for a catalog if
-// present.  c must have a valid Public Id and Scope Id set.
-// TODO: consider merging the functions for getting catalog and persisted data into a view.
+// present. c must have a valid Public Id and Scope Id set.  TODO: consider
+// merging the functions for getting catalog and persisted data into a view.
 func (r *Repository) getPersistedDataForCatalog(ctx context.Context, c *HostCatalog) (*plgpb.HostCatalogPersisted, error) {
```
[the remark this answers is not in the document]
talanknight: This TODO is resolved in https://github.com/hashicorp/boundary/pull/1643

internal/host/plugin/repository_host_catalog.go
```diff
@@ -143,6 +146,270 @@ func (r *Repository) CreateCatalog(ctx context.Context, c *HostCatalog, _ ...Opt
...
+	}
+
+	if currentCatalog == nil {
+		return nil, db.NoRowsAffected, db.NoRowsAffected, errors.New(ctx, errors.RecordNotFound, op, fmt.Sprintf("catalog with id %q not found", c.PublicId))
+	}
+
+	// We also need the secrets for the host catalog - while we don't
+	// patch them in the same way as we do attributes, we do need them
+	// for the plugin request.
+	currentCatalogPersisted, err := r.getPersistedDataForCatalog(ctx, currentCatalog)
```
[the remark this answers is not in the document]
talanknight: the catalog is the aggregate here so if persisted updates it should check and update the version on the catalog.

internal/host/plugin/host_catalog_secret.go
```diff
@@ -90,7 +90,7 @@ func (c *HostCatalogSecret) decrypt(ctx context.Context, cipher wrapping.Wrapper
 	return nil
 }
 
-func (c *HostCatalogSecret) insertQuery() (query string, queryValues []interface{}) {
+func (c *HostCatalogSecret) upsertQuery() (query string, queryValues []interface{}) {
 	query = upsertHostCatalogSecretQuery
```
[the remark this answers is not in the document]
talanknight: The OnConflict changes are recent.  This could probably be updated to use that now.

internal/host/plugin/repository_host_catalog.go
```diff
@@ -143,6 +146,270 @@ func (r *Repository) CreateCatalog(ctx context.Context, c *HostCatalog, _ ...Opt
...
+				// updated).
+				returnedCatalog = currentCatalog.clone()
+			}
+
+			if hcSecret != nil {
+				if deleteSecrets {
+					// We didn't set/encrypt the persisted data because there was
+					// none returned. Just delete the entry.
+					deletedSecret := hcSecret.clone()
+					q, v := deletedSecret.deleteQuery()
```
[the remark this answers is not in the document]
talanknight: I forget the reason why I went with the hand written query.  I believe this should use the writer.Delete function.

internal/host/plugin/repository_host_catalog.go
```diff
@@ -143,6 +146,270 @@ func (r *Repository) CreateCatalog(ctx context.Context, c *HostCatalog, _ ...Opt
...
+						return errors.Wrap(ctx, err, op)
+					}
+					if secretsUpdated != 1 {
+						return errors.New(ctx, errors.MultipleRecords, op, fmt.Sprintf("expected 1 catalog secret to be deleted, got %d", secretsUpdated))
+					}
+					msgs = append(msgs, deletedSecret.oplogMessage(db.DeleteOp))
+				} else {
+					// Update the secrets.
+					updatedSecret := hcSecret.clone()
+					q, v := updatedSecret.upsertQuery()
```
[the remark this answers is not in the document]
talanknight: OnConflict is new and should replace all these places where we use upserts.

vancluever:
The current state of this branch is working to try and address https://github.com/hashicorp/boundary/pull/1634#discussion_r735803189.

The current method I'm trying to use is via [gorm hooks](https://gorm.io/docs/hooks.html), but it seems to be able to lack the capability to update a query in-flight to allow the modification of columns as shown [here](https://gorm.io/docs/update.html#Change-Updating-Values). It seems that you can only add new columns with `SetColumn` and not add existing ones. This seems to be necessary because gorm builds updates queries _before_ calling `BeforeSave` or `BeforeUpdate`.

internal/host/plugin/repository_host_catalog.go
```diff
@@ -143,6 +146,270 @@ func (r *Repository) CreateCatalog(ctx context.Context, c *HostCatalog, _ ...Opt
...
+		case strings.EqualFold("name", f) && c.Name != "":
+			dbMask = append(dbMask, "name")
+			newCatalog.Name = c.Name
+		case strings.EqualFold("description", f) && c.Description == "":
+			nullFields = append(nullFields, "description")
+			newCatalog.Description = c.Description
+		case strings.EqualFold("description", f) && c.Description != "":
+			dbMask = append(dbMask, "description")
+			newCatalog.Description = c.Description
+		case strings.EqualFold("attributes", f) && c.Attributes != nil:
```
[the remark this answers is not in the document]
talanknight: I have confirmed it is in fact how I stated above.   "attributes.foo" is included and not "attributes"

internal/host/plugin/repository_host_catalog.go
```diff
@@ -143,6 +146,270 @@ func (r *Repository) CreateCatalog(ctx context.Context, c *HostCatalog, _ ...Opt
...
+		return nil, db.NoRowsAffected, db.NoRowsAffected, errors.New(ctx, errors.InvalidParameter, op, fmt.Sprintf("plugin %q not available", currentCatalog.GetPluginId()))
+	}
+
+	// Convert the catalog values to API protobuf values, which is what
+	// we use for the plugin hook calls.
+	currPlgHc, err := toPluginCatalog(ctx, currentCatalog)
+	if err != nil {
+		return nil, db.NoRowsAffected, db.NoRowsAffected, errors.Wrap(ctx, err, op)
+	}
+	newPlgHc, err := toPluginCatalog(ctx, newCatalog)
```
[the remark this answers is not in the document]
vancluever: After reviewing this and trying out solutions that needed to be reverted due to lack of functionality in gorm and our own standards with regards to where such abstractions should lie, I'd recommend abandoning this for now. Here's my reasons:

Performing this work "around the edges" currently does not really give much value other than saving a single marshal/unmarshal cycle if attributes are to be updated. The values need to be unmarshaled anyway to be sent to the plugin, and as explained in Todd's message, service handlers will be doing more marshaling/unmarshaling anyway. This seems to need broad architectural changes to the data to effectively change this, probably out of the scope of this PR.
I still don't believe update would be in any hot path where performance is particularly important.
Concerns have been raised internally about storing a raw *structpb.Struct in the database. While I think we need to talk more about what can actually be done about it, given we don't know exactly the shape of attribute data looks like for plugin-based host catalogs, it might be a good idea to treat this data as opaquely as possible within the repository for now, and perform marshaling/unmarshaling as we need to do it.

For now, I'm just going to keep it as-is, move the PR to ready for review, and work on resolving any other comments.

internal/host/plugin/repository_host_catalog.go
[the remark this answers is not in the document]
vancluever: Updated code to deal with this.

internal/host/plugin/host_catalog_secret.go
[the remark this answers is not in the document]
vancluever: Will make the changes.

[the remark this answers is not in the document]
vancluever: So the issue here that seems to be happening after the update seems to have to do with the fact that this table lacks a primary key? I get this when testing update:

```
--- FAIL: TestHostCatalogSecret_Update (0.66s)
    --- FAIL: TestHostCatalogSecret_Update/OK (0.00s)
        host_catalog_secret_test.go:262:
            	Error Trace:	host_catalog_secret_test.go:262
            	            	db.Update: db.lookupAfterWrite: db.LookupById: db.primaryKeyWhere: unsupported interface type *plugin.HostCatalogSecret: parameter violation: error #100
            	Test:       	TestHostCatalogSecret_Update/OK
```

Will investigate this more tomorrow.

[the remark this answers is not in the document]
mgaffney: The `host_plugin_catalog_secret` [does have a primary key](https://github.com/hashicorp/boundary/blob/1312628a1ef8b4ab95e642138462112f1b178925/internal/db/schema/migrations/oss/postgres/19/05_plugin_host.up.sql#L89).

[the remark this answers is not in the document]
vancluever: Alright, looks like we can't support this as there's no way to change the primary key field in our DB abstractions at the moment. After internal discussions it was decided to leave the field as is, note a TODO, and move on.

internal/host/plugin/repository_host_catalog.go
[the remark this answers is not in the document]
vancluever: Leaving this for now. See https://github.com/hashicorp/boundary/pull/1634#discussion_r741330899

internal/host/plugin/repository_host_catalog.go
```diff
@@ -111,7 +114,7 @@ func (r *Repository) CreateCatalog(ctx context.Context, c *HostCatalog, _ ...Opt
 
 			if hcSecret != nil {
 				newSecret := hcSecret.clone()
-				q, v := newSecret.insertQuery()
+				q, v := newSecret.upsertQuery()
```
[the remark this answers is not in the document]
vancluever: Resolving this for the time being to move this PR along.

internal/host/plugin/repository_host_catalog.go
```diff
@@ -143,6 +146,270 @@ func (r *Repository) CreateCatalog(ctx context.Context, c *HostCatalog, _ ...Opt
...
+	// ensures that we don't alter anything in the original passed in
+	// parameters, which we don't do by convention. We will be adding
+	// to this "working set" as we move through the method.
+	c = c.clone()
+
+	// Get the old catalog first. We patch the record first before
+	// sending it to the DB for updating so that we can run on
+	// OnUpdateCatalog. Note that the field masks are still used for
+	// updating.
+	currentCatalog, _, err := r.LookupCatalog(ctx, c.PublicId)
```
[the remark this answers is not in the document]
vancluever: Guard added now!

internal/host/plugin/repository_host_catalog.go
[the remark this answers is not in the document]
vancluever: Just decided to keep as-is to keep the commit moving forward. We will come back to this soon, @jimlambrt is working on extending the DB abstractions to accommodate.

vancluever:
Marked as ready for (hopefully final) review.

We have left the manually-generated queries as-is for now to move the PR along. @jimlambrt is working on extending the DB abstractions to accommodate, and that can be added afterwards.

internal/host/plugin/host_catalog_secret.go
```diff
@@ -90,7 +90,13 @@ func (c *HostCatalogSecret) decrypt(ctx context.Context, cipher wrapping.Wrapper
 	return nil
 }
 
-func (c *HostCatalogSecret) insertQuery() (query string, queryValues []interface{}) {
+// TODO: this is necessary as update logic requires a static
```
talanknight, line 93: Reference the jira task inside the TODO as suggested to make it obvious that when the jira task is resolved the TODO should be gone.

```suggestion
// TODO(ICU-2835): this is necessary as update logic requires a static
```

internal/host/plugin/repository_host_catalog.go
```diff
@@ -143,6 +146,279 @@ func (r *Repository) CreateCatalog(ctx context.Context, c *HostCatalog, _ ...Opt
...
+	)
+
+	if err != nil {
+		if errors.IsUniqueError(err) {
+			return nil, db.NoRowsAffected, db.NoRowsAffected, errors.Wrap(ctx, err, op, errors.WithMsg(fmt.Sprintf("in %s: name %s already exists", c.PublicId, c.Name)))
+		}
+		return nil, db.NoRowsAffected, db.NoRowsAffected, errors.Wrap(ctx, err, op, errors.WithMsg(fmt.Sprintf("in %s", c.PublicId)))
+	}
+
+	return returnedCatalog, catalogsUpdated, secretsUpdated, nil
```
talanknight, line 419: Don't separate out the catalogs updated or secrets updated.  Provide a single count for both of them.  A catalog whose secret gets updated should just count as a catalog which is updated.  A catalog whose secret and non-secret gets updated should still just count as a single catalog that got updated.

internal/host/plugin/repository_host_catalog.go
```diff
@@ -143,6 +146,279 @@ func (r *Repository) CreateCatalog(ctx context.Context, c *HostCatalog, _ ...Opt
...
+				}
+				msgs = append(msgs, &cOplogMsg)
+			} else {
+				// Returned catalog needs to be the old copy, as no fields in the
+				// catalog itself are being updated (note: secrets may still be
+				// updated).
+				returnedCatalog = currentCatalog.clone()
+			}
+
+			if hcSecret != nil {
```
talanknight, line 373: The catalog's version should be incremented if it's secret is updated even if it is the only thing that got updated.

Here is an example of how oidc auth methods increase the version when the primary table isn't the thing being updated: https://github.com/hashicorp/boundary/blob/main/internal/auth/oidc/repository_auth_method_update.go#L269

internal/host/plugin/repository_host_catalog.go
```diff
@@ -143,6 +146,279 @@ func (r *Repository) CreateCatalog(ctx context.Context, c *HostCatalog, _ ...Opt
...
+			nullFields = append(nullFields, "description")
+			newCatalog.Description = c.Description
+		case strings.EqualFold("description", f) && c.Description != "":
+			dbMask = append(dbMask, "description")
+			newCatalog.Description = c.Description
+		case strings.EqualFold("attributes", strings.Split(f, ".")[0]):
+			// Flag attributes for updating. While multiple masks may be
+			// sent, we only need to do this once.
+			updateAttributes = true
+		case strings.EqualFold("secrets", f):
```
talanknight, line 243: Don't rely on the secret field being present.  If the passed in catalog has anything in the secrets field, just pass that on to the plugin.

internal/host/plugin/repository_host_catalog.go
```diff
@@ -143,6 +146,279 @@ func (r *Repository) CreateCatalog(ctx context.Context, c *HostCatalog, _ ...Opt
...
+		}
+
+		// Patch the working set with the new attributes.
+		c.Attributes = newCatalog.Attributes
+	}
+
+	// Get the plugin client.
+	plgClient, ok := r.plugins[currentCatalog.GetPluginId()]
+	if !ok || plgClient == nil {
+		return nil, db.NoRowsAffected, db.NoRowsAffected, errors.New(ctx, errors.InvalidParameter, op, fmt.Sprintf("plugin %q not available", currentCatalog.GetPluginId()))
```
talanknight, line 270: I don't think this is an invalid parameter anymore when we already have a catalog with this plugin id.  IMO it makes sense on create since we don't necessarily know if the plugin id exists, but now we know that it does.

This seems like a plugin specific error code might be good to create or make it an internal error.

internal/host/plugin/repository_host_catalog.go
```diff
@@ -143,6 +146,279 @@ func (r *Repository) CreateCatalog(ctx context.Context, c *HostCatalog, _ ...Opt
...
+	if err != nil {
+		return nil, db.NoRowsAffected, db.NoRowsAffected, errors.Wrap(ctx, err, op)
+	}
+	newPlgHc, err := toPluginCatalog(ctx, newCatalog)
+	if err != nil {
+		return nil, db.NoRowsAffected, db.NoRowsAffected, errors.Wrap(ctx, err, op)
+	}
+
+	// Get the secrets for the host catalog.
+	currentCatalogPersisted, err := r.getPersistedDataForCatalog(ctx, currentCatalog)
```
talanknight, line 285: I think you need to update your client.  Last week the persisted data for a catalog was changed to be returned with getCatalog.

internal/host/plugin/repository_host_catalog.go
```diff
@@ -143,6 +146,279 @@ func (r *Repository) CreateCatalog(ctx context.Context, c *HostCatalog, _ ...Opt
...
+	// ensures that we don't alter anything in the original passed in
+	// parameters, which we don't do by convention. We will be adding
+	// to this "working set" as we move through the method.
+	c = c.clone()
+
+	// Get the old catalog first. We patch the record first before
+	// sending it to the DB for updating so that we can run on
+	// OnUpdateCatalog. Note that the field masks are still used for
+	// updating.
+	currentCatalog, _, err := r.LookupCatalog(ctx, c.PublicId)
```
talanknight, line 206: use getCatalog instead of LookupCatalog.

internal/host/plugin/repository_host_catalog.go
```diff
@@ -338,8 +614,10 @@ func toPluginCatalog(ctx context.Context, in *HostCatalog) (*pb.HostCatalog, err
 		return nil, errors.New(ctx, errors.InvalidParameter, op, "nil storage plugin")
 	}
 	hc := &pb.HostCatalog{
-		Id:      in.GetPublicId(),
-		ScopeId: in.GetScopeId(),
+		Id:          in.GetPublicId(),
+		ScopeId:     in.GetScopeId(),
+		Name:        wrapperspb.String(in.GetName()),
+		Description: wrapperspb.String(in.GetDescription()),
```
talanknight, line 620: Name and Description should only be set with a wrapperspb.String if in.GetName and in.GetDescription are not empty strings.

internal/host/plugin/repository_host_catalog_test.go
```diff
@@ -366,12 +367,645 @@ func TestRepository_CreateCatalog(t *testing.T) {
...
+				OnUpdateCatalogFn: func(_ context.Context, req *plgpb.OnUpdateCatalogRequest) (*plgpb.OnUpdateCatalogResponse, error) {
+					gotOnUpdateCatalogRequest = req
+					return &plgpb.OnUpdateCatalogResponse{Persisted: &plgpb.HostCatalogPersisted{Secrets: req.GetNewCatalog().GetSecrets()}}, pluginError
+				},
+			},
+		},
+	}
+
+	// Set up two existing catalogs for duplicate tests, one with org
+	// scope, one with project scope.
```
talanknight, line 399: I think catalogs are only able to be project scoped.  At least that is enforced in the service handler.  It should probably be enforced deeper if that is something we actually want to enforce.

talanknight (changes requested):
I haven't gotten far into the tests but will follow up tomorrow w/ my review of the tests.

internal/host/plugin/host_catalog_secret.go
vancluever: Is this a convention that should be followed in the future? Just noting as it's mentioned in the comments a few lines down.

jimlambrt:
https://github.com/hashicorp/boundary/pull/1658 has been merged and includes the required bits to address some of the comments on this PR.

internal/host/plugin/repository_host_catalog.go
vancluever: Fixed!

internal/host/plugin/repository_host_catalog.go
vancluever: So this goes back to the question: there's a guard that prevents an empty field mask from being sent in. Do we now get rid of that guard if we are to pass in secrets indiscriminately? If we do that, we lose some ability to validate what has been passed in:

* Fields that have well-defined masks will still be iterated over and accepted properly, or errored out on.
*  `secrets`, or any deeper path thereof, will fail due to invalid field mask.
* Secrets could be passed in without an accompanying field mask. 

It seems to me that this breaks the integrity of the validation work that's done here in processing masks. I know that we discussed adding secrets to the passThroughPrefix to allow for these to be passed in a similar way to attributes, has your thinking on this changed?

internal/host/plugin/repository_host_catalog.go
vancluever: Done!

internal/host/plugin/repository_host_catalog.go
vancluever: Just discussed this offline and I think Todd is aware of this and was about to comment on it. Going to proceed with partial validation in the same way that we are doing with attributes.

vancluever: Updated now!

internal/host/plugin/repository_host_catalog.go
vancluever: Yep, merged most recent changes yesterday (which led to #1661).

internal/host/plugin/repository_host_catalog.go
vancluever: Yep, updated on merge yesterday.

internal/host/plugin/repository_host_catalog.go
vancluever: Yeah, maybe internal for now (unless we already have a class/code range for plugin errors). Will update.

vancluever: Done!

internal/host/plugin/repository_host_catalog.go
vancluever: Updated!

internal/host/plugin/repository_host_catalog_test.go
vancluever: Do we have an example of a similar constraint? I can work on adding it and a few tests in a new PR and then fix this up to just create/use two separate projects.

vancluever: I've just updated this to use two separate project scopes for now.

internal/host/plugin/repository_host_catalog_test.go
```diff
@@ -368,12 +369,736 @@ func TestRepository_CreateCatalog(t *testing.T) {
...
+		require.NoError(err)
+		scopeWrapper, err := dbKmsCache.GetWrapper(ctx, cat.GetScopeId(), kms.KeyPurposeDatabase)
+		require.NoError(err)
+		require.NoError(cSecret.encrypt(ctx, scopeWrapper))
+		cSecretQ, cSecretV := cSecret.upsertQuery()
+		secretsUpdated, err := dbRW.Exec(ctx, cSecretQ, cSecretV)
+		require.NoError(err)
+		require.Equal(1, secretsUpdated)
+
+		cleanupFunc := func() {
```
talanknight, line 1053: t.Cleanup() takes a function and calls it once the *testing.T receiver's tests/subtests finish.  That way you don't have to pass the cleanup onto the caller.  Would that be helpful here?

internal/host/plugin/repository_host_catalog_test.go
```diff
@@ -368,12 +369,736 @@ func TestRepository_CreateCatalog(t *testing.T) {
...
+				checkSecrets(map[string]interface{}{
+					"one": "two",
+				}),
+				checkNumUpdated(1),
+				checkVerifyCatalogOplog(oplog.OpType_OP_TYPE_UPDATE),
+			},
+		},
+		{
+			name:        "update name to same",
+			changeFuncs: []changeHostCatalogFunc{changeName("")},
```
talanknight, line 785: This actually fails a db check if it is actually going through.  Names must have length > 0.  see wt_name constraints in internal/db/schema/migrations/postgres/1/02_domains.up.sql.   Is the row actually updated (the update_time being updated?)

talanknight (approved):
Looks good!

internal/host/plugin/repository_host_catalog_test.go
vancluever: It _should_ be nulling out as per https://github.com/hashicorp/boundary/pull/1634/files#diff-9e3211879fb53616a797ca22d9549ac9f558ecc6e5ad6ca5b1a16bfcfc5be6b9R222-R224. Should be okay?

internal/host/plugin/repository_host_catalog.go
```diff
@@ -145,6 +148,299 @@ func (r *Repository) CreateCatalog(ctx context.Context, c *HostCatalog, _ ...Opt
...
+// c.Attributes is a no-op for modifications. Rather, if fields need
+// to be reset, its field in c.Attributes should individually set to
+// null.
+//
+// Updates are sent to OnUpdateCatalog with a full copy of both the
+// current catalog, and the state of the new catalog should it be
+// updated, along with any secrets included in the new request. This
+// request may alter the returned persisted state. Update of the
+// record in the database is aborted if this call fails.
+func (r *Repository) UpdateCatalog(ctx context.Context, c *HostCatalog, version uint32, fieldMask []string, _ ...Option) (*HostCatalog, int, error) {
```
talanknight, line 173: Can you make this UpdateCatalog call also return the *hostplugin.Plugin like CreateCatalog does?  This is so the response can include the plugin info for the specific resource being updated.

internal/host/plugin/repository_host_catalog_test.go
vancluever: Discussed with @talanknight and confirmed this was okay.

internal/host/plugin/host_catalog_secret.go
vancluever: 👍 updated it a couple of days ago, so we should be fine here.

internal/host/plugin/repository_host_catalog.go
```diff
@@ -145,6 +148,299 @@ func (r *Repository) CreateCatalog(ctx context.Context, c *HostCatalog, _ ...Opt
...
+
+	if updateAttributes && c.Attributes != nil {
+		dbMask = append(dbMask, "attributes")
+		newCatalog.Attributes, err = patchstruct.PatchBytes(newCatalog.Attributes, c.Attributes)
+		if err != nil {
+			return nil, db.NoRowsAffected, errors.Wrap(ctx, err, op, errors.WithMsg("error in catalog attribute JSON"))
+		}
+
+		// Patch the working set with the new attributes.
+		c.Attributes = newCatalog.Attributes
```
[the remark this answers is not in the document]
vancluever: The original `c` _is_ used in the update query just to make sure that the fields that need updating (and only those fields) are included in the update set. This was designed to mimic how other update functions work but we don't do this degree of operation of them at the repository level, I don't think - they have much less complexity.

So we _could_ probably not use `c` for this but that was the intention.

There are a lot of in-flight variables and I'm wondering some abstraction in a method would make sure things were a bit clearer?

internal/host/plugin/repository_host_catalog_test.go
```diff
@@ -368,12 +369,736 @@ func TestRepository_CreateCatalog(t *testing.T) {
...
+		}
+	}
+
+	// checkUpdateCatalogRequestCurrentName := func(want string) checkFunc {
+	// 	return func(t *testing.T, ctx context.Context) {
+	// 		t.Helper()
+	// 		assert := assert.New(t)
+	// 		assert.Equal(wrapperspb.String(want), gotOnUpdateCatalogRequest.CurrentCatalog.Name)
+	// 	}
+	// }
```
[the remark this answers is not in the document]
vancluever: Yeah, I kept them around just in case they were needed later (being just unused test checks). Removed them now.

internal/host/plugin/repository_host_catalog_test.go
vancluever: Will noodle on it for a little bit (not too long). It might add some idiomatic value but I think the workflow is clear enough that it's not 100% needed, and there's some other defers going on that I think would be good.

I think it depends on if cleanup is called after the parent finishes or if the subtest finishes, since the intention is to have it cleaned up after the subtest does.

internal/host/plugin/repository_host_catalog.go
vancluever: Done!

internal/host/plugin/repository_host_catalog.go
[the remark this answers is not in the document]
vancluever: Slept on this and just decided to use `newCatalog`. Should be good now.

internal/host/plugin/repository_host_catalog_test.go
vancluever: Think this will be fine. According to t.Cleanup():

https://pkg.go.dev/testing#T.Cleanup

> Cleanup registers a function to be called when the test (or subtest) and all its subtests complete. Cleanup functions will be called in last added, first called order. 

I've updated the code to use this.

talanknight (approved):

PR #1646, opened 2021-10-28
Update view to return host IDs in a set on set lookup

jefferai:
This removes the optional call to the plugin.


internal/host/plugin/host.go
```diff
@@ -31,6 +31,7 @@ func NewHost(ctx context.Context, catalogId, externalId string, opt ...Option) *
 	h := &Host{
 		PluginId: opts.withPluginId,
 		Host: &store.Host{
+			PublicId:    opts.withPublicId,
```
talanknight, line 34: update comment to include withPublicId option as supported.

jefferai:
Added a test on the returned values.

talanknight (approved):

PR #1647, opened 2021-10-28
Add plugin set sync and orphaned host cleanup jobs

talanknight:
This moved some  helper methods from the plugin host repo to the job.  This also added a test helper function to run the set sync job's behavior once.

A follow up PR will add an endpoint which schedules a host set for immediate re-sync.


jefferai (approved):
Looks fantastic!

PR #1650, opened 2021-10-30
refactor: Move functions from kms pkg to new libs/crypto pkg

jimlambrt:
We needed to create a new lib/crypto package and move
 kms.NewDerivedReader into the new pkg. This will allow us to
use NewDerivedReader in the events pkg without a circular
dependency.

Also, consolidate the implementations of hmac-sha256 into the
lib/crypto pkg and refactor existing stuff to use the new refactored
version.

### DIFF
--- a/internal/libs/crypto/options.go
+++ b/internal/libs/crypto/options.go
@@ -1,0 +1,57 @@
+package crypto
+
+// getOpts - iterate the inbound Options and return a struct.
+func getOpts(opt ...Option) options {
+	opts := getDefaultOptions()
+	for _, o := range opt {
+		if o != nil {
+			o(&opts)
+		}
+	}
+	return opts
+}
+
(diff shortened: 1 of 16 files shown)


internal/libs/crypto/options.go
```diff
@@ -0,0 +1,57 @@
+package crypto
+
+// getOpts - iterate the inbound Options and return a struct.
+func getOpts(opt ...Option) options {
```
jefferai, line 4: As this is a new package, you may want `getOpts` to return `(options, error)`

jefferai (dismissed):
Just the one comment, which I don't feel strongly about. so either way. Looks good!

internal/libs/crypto/options.go
jimlambrt: done.

jefferai (approved):

PR #1652, opened 2021-11-01
Make test container names configurable

tmessi:


hugoamvieira (commented):
This looks good! 👍 

I found some hardcoded references to `boundary-sql-tests` as a docker container in config files / docs:
- https://github.com/hashicorp/boundary/blob/main/CONTRIBUTING.md?plain=1#L83-L89
- https://github.com/hashicorp/boundary/blob/main/.circleci/config.yml#L984 (This file seems auto-gen'd, so I'm not sure it's a product of any of the changes in this PR)
- https://github.com/hashicorp/boundary/blob/main/.circleci/config/jobs/build.yml#L9

We're now making this configurable in the Makefiles, so if it changes (presumably you could just set an env var and run `make X`?), those other config files / docs would also have to change. I'm not sure this is necessarily a problem since the default value _is_ still the same - I just thought I'd mention it to get your thoughts.

tmessi:
> This looks good! +1
> 
> I found some hardcoded references to `boundary-sql-tests` in config files / docs:
> 
> * https://github.com/hashicorp/boundary/blob/main/CONTRIBUTING.md?plain=1#L83-L89
> * https://github.com/hashicorp/boundary/blob/main/.circleci/config.yml#L984 (This file seems auto-gen'd, so I'm not sure it's a product of any of the changes in this PR)
> * https://github.com/hashicorp/boundary/blob/main/.circleci/config/jobs/build.yml#L9
> 
> We're now making this configurable in the Makefiles, so if it changes (presumably you could just set an env var and run `make X`?), those other config files / docs would also have to change. I'm not sure this is necessarily a problem since the default value _is_ still the same - I just thought I'd mention it to get your thoughts.

The circleci config files are used by circleci, so they should just use the default value and be fine. I think the docs are fine to use the default, but I can update them to note that this can be customized via env vars in the same way the contrib doc mentions you can customize the port

tmessi:
@hugoamvieira see 3520d21

hugoamvieira (approved):
LGTM 👍

mgaffney (approved):

PR #1653, opened 2021-11-01
test(targets): Change test to not depend on response order

tmessi:
This test started failing when the order of the list response changed.
Since the order does not matter, this test was updated to not depend on
the order.


hugoamvieira (approved):
LGTM!

Have a Q about our usage of `assert` and `require`, but it's just for my own sake.

Is the usage of both packages intentional? I think we should probably coalesce into `require` if there's no special reason for using both, since, IIRC, `require` automatically fails a test if an assertion fails, and `assert` will happily continue along.

mgaffney (approved):

PR #1654, opened 2021-11-01
Mask manager can now include mask paths w/ a specified prefix which would be filtered otherwise

talanknight:
This will be useful when updating plugin based host catalog and sets since attributes and secrets cannot be contained in the mask manager mapping.

### DIFF
--- a/internal/servers/controller/handlers/mask_manager.go
+++ b/internal/servers/controller/handlers/mask_manager.go
@@ -19,8 +19,9 @@ type (
 	MaskSource      []protoreflect.ProtoMessage
 )
 
-// NewMaskManager returns a mask manager that can translate field masks into the first proto from all subsequent
-// protos assuming they are both using the mask_mapping custom option.  Error is returned if no mappings are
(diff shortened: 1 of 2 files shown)


vancluever (commented):
@talanknight I'm trying to understand how this will translate to something like what we need it for, ie: a request update for `attributes` in host catalogs. `Translate` seems to split on period, so maybe we can get some tests/assertions for what `attributes.region` or something similar would look like?

vancluever:
Just noticed that it's a comma, not a period.

@talanknight maybe can you mention what's changed here? Will this now just translate `attributes.region` into just `attributes`? Or was this just not working before at all?

vancluever:
Alright, misread that split (was not a dot, was a comma). Talked to @talanknight and clarified that indeed what was happening was due to a lack of prefix matching, all paths that weren't an exact mapping were getting dropped. So for our situation in plugin-based host catalogs, updating fields in `attributes`, which would have been presented from the gateway as `attributes.region`, would have translated to nothing, and not passed into repository update functionality.

vancluever (approved):

PR #1658, opened 2021-11-02
feature (db): Add support for multi-column PKs

jimlambrt:
Add support for the db layer operations like lookup, delete, update,
etc to support resources with mulit-column primary keys.  It also
adds support for PKs which are not either public_id or private_id.

These new capabilities are needed for things in #1634 ( FYI: @talanknight @vancluever)

### DIFF
--- a/internal/db/read_writer_test.go
+++ b/internal/db/read_writer_test.go
@@ -571,6 +571,17 @@ func TestDb_Update(t *testing.T) {
 		assert.Equal(0, rowsUpdated)
 		assert.Contains(err.Error(), "db.Update: oplog validation failed: db.validateOplogArgs: missing metadata: parameter violation: error #100")
 	})
+	t.Run("multi-column", func(t *testing.T) {
+		assert, require := assert.New(t), require.New(t)
+		w := Db{underlying: db}
+		scooter := testScooter(t, db, "", 0)
+		accessory := testAccessory(t, db, "test accessory")
+		scooterAccessory := testScooterAccessory(t, db, scooter.Id, accessory.AccessoryId)
+		scooterAccessory.Review = "this is great"
+		rowsUpdated, err := w.Update(context.Background(), scooterAccessory, []string{"Review"}, nil)
+		require.NoError(err)
+		assert.Equal(1, rowsUpdated)
+	})
 }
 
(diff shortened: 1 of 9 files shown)


internal/db/schema/migrations/oss/postgres/19/01_db.up.sql
```diff
@@ -0,0 +1,56 @@
+begin;
```
jefferai, line 1: Should this be in a migration for everyone? Or should it be part of the tests utilizing it?

jefferai (commented):
Looks pretty good. There's a failing test and I'm casting some side-eye at that migration but outside of that the approach looks solid.

internal/db/read_writer_test.go
```diff
@@ -2013,16 +2061,34 @@ func TestDb_LookupById(t *testing.T) {
...
+			name:       "simple-non-public-non-private-id",
+			underlying: db,
+			args: args{
+				resource: accessory,
+			},
+			wantErr: false,
+			want:    accessory,
+		},
+		{
+			name:       "compond",
```
mgaffney, line 2079: Can we add a test to verify an error is returned when one or more values of a compound key are the zero value?

mgaffney (changes requested):
Nice work! Just one request for an additional test.

internal/db/schema/migrations/oss/postgres/19/01_db.up.sql
jimlambrt: I agree, that is shouldn't be.  With that said, this PR is aligned with the current pattern used by internal/db for test schema objects.  I've opened an issue to refactor all of them: https://hashicorp.atlassian.net/browse/ICU-2841 in a separate PR.

internal/db/read_writer_test.go
jimlambrt: ty!

jimlambrt:
> Looks pretty good. There's a failing test and I'm casting some side-eye at that migration but outside of that the approach looks solid.

Working on the failed test.  happy to say, the failure shed light on some areas for improvement in that set of unit tests, which led to finding/fixing a new issue.

jefferai (approved):

mgaffney (approved):

PR #1660, opened 2021-11-03
Pass namespace to the Vault client

### LINKED ISSUE #1597: Credential Store creation fails when using Vault namespace
**Describe the bug**
Credential stores cannot be registered against a Vault namespace.

**To Reproduce**
1. Generate a token `vault token create -policies=boundary_credential_store -orphan -period=60 -renewable=true`
2. Add vault URL, token, and namespace to UI. Click save
3. Error response:
```
credentialstores.(Service).createInRepo: unable to create credential store: vault.(Repository).CreateCredentialStore: unable to lookup vault token: vault.(client).lookupToken: vault: http://127.0.0.1:8200: unknown: error #0: Error making API request. URL: GET http://127.0.0.1:8200/v1/auth/token/lookup-self Code: 403. Errors: * permission denied
```
Please note that this works as expected in the default/root namespace of Vault.

**Expected behavior**
Token checks query occur within the configured namespace.

**Additional context**
This is a blocker for configuration of Boundary with HCP Vault which doesn't give access to the root namespace, and defaults into the "admin" namespace.

jefferai:
The namespace was plumbed all the way through the API but isn't actually
set on the client when it is created.

For various reasons this is hard to test right now except manually (although that will change in the future), but it works in my testing and works for @joatmon08 too.

Fixes #1597

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
   a missing mandated cipher suite of the HTTP/2 (`h2`) specification that could
   result in no shared cipher suites between the Boundary API listener and those
   clients. ([PR](https://github.com/hashicorp/boundary/pull/1637))
+* vault: Fix credential store support when using Vault namespaces
+  ([Issue](https://github.com/hashicorp/boundary/issues/1597),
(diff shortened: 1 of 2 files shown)


mgaffney (dismissed):
ty!

PR #1670, opened 2021-11-05
fix(schema): Prevent duplicate versions per edition

tmessi:
This should help prevent issues like the one fixed in
f2ad4de554d5e10365bd39c110f29f9a4c481fa5.

A version for a particular edition should only exist once, previously
which ever file was read last for a version would be included, which
would lead to some migrations not running.


jimlambrt (dismissed):
ty!

PR #1671, opened 2021-11-05
internal/host/plugin: catch duplicate catalog names early during update

vancluever:
Two host catalogs cannot have the same name in the same scope. We need
to catch this early so that it does not get sent to the plugin where
destructive change operations can occur (ie: rotating credentials) that
would then not get committed to the database.


talanknight:
I think this is now superseded by PR 1683.  Closing this PR to avoid confusion.

PR #1673, opened 2021-11-05
Add update catalog support for plugin based catalogs in handler and go sdk

talanknight:
Ran `make fmt` to fix up some pending formatting issues.
Creating plugin based sets and catalogs silently drops attributes where the value is set to nil.
Added gosdk tests alongside the newly added handler tests.


vancluever (commented):
Still have to look over the rest of it but the patchstruct fix looks good! Good catch 🙂

internal/host/plugin/repository_host_catalog.go
```diff
@@ -512,10 +529,6 @@ func (r *Repository) DeleteCatalog(ctx context.Context, id string, _ ...Option)
 		return db.NoRowsAffected, errors.Wrap(ctx, err, op)
 	}
 	sets, _, err := r.getSets(ctx, "", c.GetPublicId())
-	if err != nil {
```
jefferai, line 515: Was this on purpose? Do we not want to check the error? If not should replace with `_`.

jefferai (approved):
Looks good with that one comment that should be addressed.

internal/host/plugin/repository_host_catalog.go
talanknight: Woops.  I started this branch based off of an unfinished branch where the update functionality was being added to the repo.  I hadn't updated this branch with those changes yet.  I've done so now.

PR #1674, opened 2021-11-05
Update Go version to 1.17.3

jefferai:
This includes security fixes and a large number of `http2` fixes. See [the list here](https://github.com/golang/go/issues?q=milestone%3AGo1.17.3+label%3ACherryPickApproved).

This also attempts switching the build container distro to current debian stable (bullseye) instead of buster. We'll see how that goes, and I'll back it out if tests don't pass.


jimlambrt (approved):

jimlambrt:
👍 as long as it passes CI/CD.

tmessi (approved):

PR #1675, opened 2021-11-05
refactor (db): Remove the internal/db test tables from the migrations

jimlambrt:


internal/db/schema/migrations/oss/postgres/19/03_db.up.sql
```diff
@@ -0,0 +1,23 @@
+begin;
```
jefferai, line 1: Does this file need to be in `20/` or is it fine to append to the latest migration major number so long as it's strictly increasing?

jefferai (approved):
Looks good with the one question outstanding.

internal/db/schema/migrations/oss/postgres/19/03_db.up.sql
jimlambrt: I confirmed with the team, that major versions can include additional migrations for new PRs as long as the major version hasn't be included within a Boundary release.

PR #1676, opened 2021-11-05
internal/host/plugin: remove custom queries for secrets

vancluever:
This removes the custom queries for host catalog secrets, now that #1658
has been merged.

It also does a lot of cleanup surrounding the related oplog entries,
using db.WithOplog natively.


vancluever:
@talanknight @jimlambrt one thing particular that I'm wondering is if the oplog changes are correct. This setup seems to match the pattern we use in static host catalogs, but with that said, we don't necessarily need to work with a separate table over there.

jefferai:
You could take a look at the upsert query in `internal/host/plugin/job_set_sync.go`. In a related question -- can Create and Update now just be combined into Upsert?

internal/host/plugin/host_catalog_secret_test.go
```diff
@@ -118,57 +118,85 @@ func TestHostCatalogSecret_New(t *testing.T) {
...
+	})
+	newSecretUpsert := secret.clone()
+	newSecretUpsert.Secret = newStructUpsert
+	require.NoError(t, newSecretUpsert.encrypt(ctx, databaseWrapper))
+	require.NoError(t, w.Create(ctx, newSecretUpsert, db.WithOnConflict(&db.OnConflict{
+		Target: db.Columns{"catalog_id"},
+		Action: db.SetColumnValues(map[string]interface{}{
+			"secret": newSecretUpsert.CtSecret,
+			"key_id": newSecretUpsert.KeyId,
+		}),
```
talanknight, line 159: Instead of using SetColumnValues, which require you passing in the values into the map, consider this potentially more succinct Action.  Here and in `repository_host_catalog.go`
```suggestion
		Action: db.SetColumns([]string{"secret", "key_id"}),
```

internal/host/plugin/repository_host_catalog.go
```diff
@@ -328,21 +314,18 @@ func (r *Repository) UpdateCatalog(ctx context.Context, c *HostCatalog, version
 		db.StdRetryCnt,
 		db.ExpBackoff{},
 		func(_ db.Reader, w db.Writer) error {
-			msgs := make([]*oplog.Message, 0, 3)
-			ticket, err := w.GetTicket(newCatalog)
```
talanknight, line 332: I think we still want to track this using tickets since we are updating 2 different tables (the catalog table and the catalog secret table) with the aggregate being the catalog.

So this would mean adding a db.NewOplogMsg to the delete and the create(onconflict) calls and writing the oplog entries with the messages like existed before.

talanknight (changes requested):
Looks good for the most part.  @jimlambrt will need to step in if I gave bad guidance regarding the oplog, but I think I'm correct :).

vancluever:
FYI going to hold off on this until the long-running branch is merged to `main` and then we can do this work against that.

internal/host/plugin/repository_host_catalog.go
vancluever: Fixed. Resolved this while I was resolving the merge conflict so that it still uses the tickets both in create and update.

internal/host/plugin/host_catalog_secret_test.go
vancluever: Done!

vancluever:
@talanknight merge conflicts resolved and comments addressed. Can you give it another look?

vancluever:
@talanknight @jefferai @jimlambrt looking to get this closed off. Can one of you give this a review? Thanks!

jefferai (approved):
Looks good to me, would like to get a review from @talanknight too though

internal/host/plugin/host_catalog_secret.go
```diff
@@ -90,41 +88,10 @@ func (c *HostCatalogSecret) decrypt(ctx context.Context, cipher wrapping.Wrapper
...
-// public_id or private_id field. Once post-lookup updates can
-// support alternate primary key fields, we can port this to native
-// Update and upserts w/OnConflict.
-func (c *HostCatalogSecret) upsertQuery() (query string, queryValues []interface{}) {
-	query = upsertHostCatalogSecretQuery
-	queryValues = []interface{}{
-		sql.Named("catalog_id", c.CatalogId),
-		sql.Named("secret", c.CtSecret),
-		sql.Named("key_id", c.KeyId),
+func (c *HostCatalogSecret) oplog(op oplog.OpType) oplog.Metadata {
```
talanknight, line 91: I don't think you need to define oplog() at all, do you?

talanknight (approved):
Looks good!

internal/host/plugin/host_catalog_secret.go
vancluever: Is there an idiomatic way to add oplog messages in write queries without it?

Static catalogs have a similar pattern ([reference](https://github.com/hashicorp/boundary/blob/0a83e21aa82004f72b0c52d38166c099594db2ce/internal/host/static/repository_host_catalog.go#L146-L173)), but it's just a separate function versus a method. This more follows the pattern we use in the plugin-based host catalogs (ie: [`oplog`](https://github.com/hashicorp/boundary/blob/0a83e21aa82004f72b0c52d38166c099594db2ce/internal/host/plugin/host_catalog.go#L122-L132) for `HostCatalog` itself).

vancluever: Oh right, you know what, I do remember now - I was looking at adding this into the credential maintenance job, but it needs to be removed here for sure. I'll fix that before merging.

vancluever:
@talanknight @jefferai @jimlambrt still need final review for this after addressing @talanknight's comments.

talanknight (approved):

PR #1677, opened 2021-11-06
fix (events): Only "sign" audit events

jimlambrt:
We should only "sign" audit events.  This fix required the latest version of go-eventlogger.


louisruch (approved):
LGTM

PR #1678, opened 2021-11-06
feature: Add client ip to inbound request information

jimlambrt:


internal/servers/common/client_ip.go
```diff
@@ -0,0 +1,134 @@
...
+		a = strings.TrimSpace(a)
+		isPrivate, err := isPrivateAddr(ctx, privateNetworks, a)
+		if !isPrivate && err == nil {
+			return a, nil
+		}
+	}
+
+	// finally, just fallback to the contexts of the X-Real-Ip header (which
+	// could be empty)
+	return realIp, nil
```
louisruch, line 59: Would we really want to fall back to an empty `"X-Real-Ip"`?  Wouldn't it be better to fall back to `r.RemoteAddr`
Also if we have `X-Real-Ip` wouldn't that be better than `X-Forwarded-For`?

Seems logic here should be check for `X-Real-Ip` if empty check for `X-Forwarded-For` that is private, if none then fall back to r.RemoteAddr

internal/servers/controller/listeners.go
```diff
@@ -16,12 +16,18 @@ import (
...
 	pbs "github.com/hashicorp/boundary/internal/gen/controller/servers/services"
 	"github.com/hashicorp/boundary/internal/libs/alpnmux"
+	"github.com/hashicorp/boundary/internal/servers/common"
 	"github.com/hashicorp/boundary/internal/servers/controller/handlers/workers"
 	"github.com/hashicorp/go-multierror"
 	"google.golang.org/grpc"
 )
 
 func (c *Controller) startListeners(ctx context.Context) error {
+
```
louisruch, line 26: `make fmt`

louisruch (commented):
Need to run `make fmt` also I have one question on logic to get client IP, haven't used the headers too much in the past so going on names to what I think is a potentially better ordering.

internal/servers/controller/listeners.go
jimlambrt: ty!

internal/servers/common/client_ip.go
jimlambrt: ty!

louisruch (approved):
LGTM

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 
 ### Deprecations/Changes
 
+* controllers: Add client IP to inbound request information which is included in
```
louisruch, line 19: This should be moved to `Next`

louisruch (approved):
One comment around the changelog entry

CHANGELOG.md
jimlambrt: ty!

jimlambrt: ty!

CHANGELOG.md
```diff
@@ -2,15 +2,10 @@
 
 Canonical reference for changes, improvements, and bugfixes for Boundary.
 
-## Next
-
 ### New and Improved
 
-* config: Add support for go-sockaddr templates to Worker and Controller
```
louisruch, line 9: Seems you wiped the previous changelog entries and the `Next` title

jimlambrt: Oops.  Ty!

internal/servers/common/client_ip.go
```diff
@@ -0,0 +1,63 @@
...
+	"net/http"
+	"strings"
+	"sync/atomic"
+
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/go-secure-stdlib/listenerutil"
+)
+
+const (
+	RealIpHeader        = "X-Real-Ip"
```
jefferai, line 15: Seems this isn't used anywhere right now?

jefferai (dismissed):

internal/servers/common/client_ip.go
jimlambrt: ty!

louisruch (dismissed):

jefferai (approved):

PR #1680, opened 2021-11-09
Add sync interval to host sets

jefferai:
This adds a parameter to host sets allowing the sync interval to be
user-specified on a per-host-set basis. A value of -1 disables syncing;
a value of null uses the current default.

There are some FIXMEs to take care of once set updates is in place.


internal/host/plugin/job_set_sync.go
```diff
@@ -135,6 +135,13 @@ func (r *SetSyncJob) Description() string {
 }
 
 func nextSync(j scheduler.Job) (time.Duration, error) {
+	// FIXME: This function should query the database and figure out the right
+	// time, however, at the time of writing we don't have an update function
+	// for sets which means we cannot call UpdateJobNextRunInAtLeast if a value
+	// changes. So rather than use the DB query, we simply check every five
+	// seconds. The actual job will ensure that only sets needing updates will
+	// be updated.
+	return 15 * time.Second, nil // Scheduler only wakes up every minute anyways
```
talanknight, line 144: The comment says five seconds, the code says 15 seconds.

Another thing to keep in mind is that while the scheduler wakes up only every minute that is per controller.  If there are multiple controllers running they could be running in a way to have at least 1 of the controller's schedulers waking up more frequently than that.  The running of the job can short circuit itself by running just a single query but that query relies on a column that doesn't have an explicit index so it could potentially hit every plugin host set in the system every time.  Is 15 seconds to frequent of a time for that?  

Thinking about this more also, the value cannot change from what it is set to at create time because there is no update set call yet so there is no risk of miscalculating the next sync job call.

Also, although right now it looks like it's only called at the end of a successful run of a job to decide when it should be run again, if this function is ever called in the flow to determine if the job should run this could starve the job since it will always be 15 seconds away from being ready to run (but hopefully the fixme will be fixed by then)

internal/host/plugin/query.go
```diff
@@ -17,12 +17,35 @@ delete from host_plugin_catalog_secret
...
 	setSyncNextRunInQuery = `
 select
-	need_sync as sync_now,
+  case
+    when sync_interval_seconds = -1
+      then
+        false
+      else
+        need_sync
+  end sync_now,
```
talanknight, line 29: need_sync is meant to indicate a one time requirement to sync the data.  It feels a little weird to overload the sync interval field to disable all syncs, even the ones triggered by change or potentially explicit requests to sync in the future.  What do you think about having `sync_interval_seconds` be either null (for default) or positive and having a different field to disable syncing all together.  That way when we have an endpoint in the future that can force the resync of a specific set we can return false if the disable sync flag is set and there won't be confusion around why a sync interval field is getting in the way of manually syncing a set.

internal/host/plugin/job_set_sync.go
```diff
@@ -288,7 +296,7 @@ func (r *SetSyncJob) syncSets(ctx context.Context, setIds []string) error {
 		}
 
 		if _, err := r.upsertHosts(ctx, ci.storeCat, catSetIds, resp.GetHosts()); err != nil {
-			errors.Wrap(ctx, err, op, errors.WithMsg("upserting hosts"))
+			return errors.Wrap(ctx, err, op, errors.WithMsg("upserting hosts"))
```
talanknight, line 299: thank you!

internal/proto/local/controller/api/resources/hostsets/v1/host_set.proto
```diff
@@ -48,9 +48,21 @@ message HostSet {
 	repeated string host_ids = 100 [json_name="host_ids"];
 
 	// An ordered list of endpoint preferences used to choose from among
-	// multiple possible endpoints for a host.
+	// multiple possible endpoints for a host. Preferences are specified by
+	// "cidr:<valid IPv4/6 CIDR>" or "dns:<globbed name>", specifying which IP
+	// address or DNS name out of a host's available possibilities should be
+	// preferred. If no preferences are specified, a value will be chosen _at
+	// random_ from all available values. May not be valid for all plugin types.
```
talanknight, line 55: A value won't be selected at random though, right?  Don't we still prioritize private ipv4 addresses, etc?

internal/proto/local/controller/api/resources/hostsets/v1/host_set.proto
```diff
@@ -48,9 +48,21 @@ message HostSet {
...
 	repeated string preferred_endpoints = 101 [json_name="preferred_endpoints", (custom_options.v1.generate_sdk_option) = true];
 
+	// An interger number of seconds indicating the amount of time that should
+	// elapse between syncs of the host set. The interval will be applied to the
+	// end of the previous sync operation, not the start. Setting to any
+	// negative value will disable syncing for that host set; setting to zero
+	// (or null) will cause the set to use Boundary's default. The default may
```
talanknight, line 62: When we discussed I think we decided on not allowing zero and requiring the standard used elsewhere of setting it to null to use the default.

internal/proto/local/controller/storage/host/plugin/store/v1/host.proto
```diff
@@ -93,6 +93,10 @@ message HostSet {
   // preferred_endpoints stores string preference values
   // @inject_tag: `gorm:"-"`
   repeated string preferred_endpoints = 11;
+
+  // Sync interval is a value representing a duration in seconds
+  // @inject_tag: `gorm:"default:null"`
+  int32 sync_interval_seconds = 12 [(custom_options.v1.mask_mapping) = {this: "sync_interval_seconds" that: "sync_interval_seconds"}];
```
talanknight, line 99: I think the repo expects casing like `SyncIntervalSeconds` in the mask.

Here and in static.proto

talanknight (changes requested):

internal/host/plugin/job_set_sync.go
jefferai: > The comment says five seconds, the code says 15 seconds.

Forgot to update the comment, apologies.

> The running of the job can short circuit itself by running just a single query but that query relies on a column that doesn't have an explicit index so it could potentially hit every plugin host set in the system every time. Is 15 seconds to frequent of a time for that?

The query doesn't seem like it should be very expensive, so running once a minute seems like it should be fine -- even once a minute per controller. Do you think we should be adding an index?

> Thinking about this more also, the value cannot change from what it is set to at create time because there is no update set call yet so there is no risk of miscalculating the next sync job call.
True, but we'll have that by release time...but that's actually still a problem. Imagine you have three controllers and sync is disabled. You run update on one, which informs the scheduler that it should run sooner -- but the other two controllers don't know this. So now only one controller will ever be able to run that job because the scheduler never checks for a new value until <never> happens. (This may well already affect the Vault jobs -- @louisruch ? ) If that controller goes down, the job may never be run(?) The way out of that is to have the scheduler check periodically anyways.

Generally speaking I'm not sure at this point if it makes sense to calculate next run times rather than have jobs define an interval for checking and just run the queries on that interval to see if anything needs to be run now. I'd rather have more queries than starvation.

internal/host/plugin/query.go
jefferai: I don't think it's overloading at all -- why would you force a sync of a set that has syncing disabled? What kind of change would force a sync of a set where the user has requested that you _not sync the set_? If there is an endpoint to force a resync, it could error out if sync is disabled, or there could be a `force` flag set to do a one-time override of the disabled synchronization.

internal/proto/local/controller/api/resources/hostsets/v1/host_set.proto
jefferai: Correct, I need to update this.

internal/proto/local/controller/api/resources/hostsets/v1/host_set.proto
jefferai: Copied and pasted from elsewhere, need to update it clearly.

internal/proto/local/controller/storage/host/plugin/store/v1/host.proto
jefferai: When I was looking in the protos most of the mask mapping values did not have such casing, e.g. higher up in the same file.

It's not clear to me if it matters, although I can't test update yet for obvious reasons.

internal/host/plugin/job_set_sync.go
louisruch: Everything updated is persisted to the db, if you call `UpdateJobNextRunInAtLeast` the scheduler updates the next run for that job in the database so all controllers would see the change and which ever controller wakes up when/or after the job is scheduled to run would run that job.

internal/host/plugin/query.go
talanknight: The reason I feel like it may be overloaded is that the naming of the `sync_interval_seconds` field suggests to me that it is meant to manage how often automatic syncing of a set happens and not decide if the set syncs at all.

I may have a very stable network that I want to connect boundary with so when I create my host set I can set `sync_interval_seconds` to -1 since I don't need to refresh the set of hots since they never change, but because I set -1 I never get the initial set of hosts which should be in my set.

In a slightly different situation I may have an AWS account with instances which never change their characteristics so I don't need automatic syncing, but while I am deciding which sets to create in boundary, I would like to update the filters of 1 set and have them reflected in boundary otherwise I can have a set with a very restricted filter (say to a single instance) but because the `sync_interval_seconds` is set to -1 the set still contains all my aws instances.

jefferai: OK -- I'm fine changing it so it doesn't ignore need_sync -- but I think generally we need to be careful/cautious as to when we force syncs.

jefferai: Updated. Note that we do not actually run this query right now, though.

internal/proto/local/controller/api/resources/hostsets/v1/host_set.proto
jefferai: Fixed.

internal/host/plugin/job_set_sync.go
jefferai: @louisruch Ahh OK, thanks.

internal/host/plugin/query.go
talanknight: I think this impacts the `setSyncJobQuery` query as well since that query only returns a record when the sync time is either null or > 0.

jefferai: Good point; I'll update that too.

internal/proto/local/controller/storage/host/plugin/store/v1/host.proto
jefferai: Updated

internal/libs/patchstruct/patchstruct_test.go
```diff
@@ -155,12 +155,12 @@ func TestPatchBytesErr(t *testing.T) {
 	t.Run("dst", func(t *testing.T) {
 		require := require.New(t)
 		_, err := PatchBytes([]byte("foo"), nil)
-		require.EqualError(err, "error reading destination data: proto: cannot parse invalid wire-format data")
+		require.Contains(err.Error(), "cannot parse invalid wire-format data")
```
talanknight, line 158: Merge plugin-hostcatalogs, this bug should be fixed in there.

internal/proto/local/controller/storage/host/static/store/v1/static.proto
```diff
@@ -105,6 +105,11 @@ message HostSet {
   // version allows optimistic locking of the resource
   // @inject_tag: `gorm:"default:null"`
   uint32 version = 7;
+
+  // This is not used! This is here to satisfy mask manager requirements due to
+  // shared API proto.
+  // @inject_tag: `gorm:"-"`
+  int32 sync_interval_seconds = 22 [(custom_options.v1.mask_mapping) = {this: "SyncIntervalSeconds" that: "sync_interval_seconds"}];
```
talanknight, line 112: Consider using the approach I used [here](https://github.com/hashicorp/boundary/pull/1684/files#diff-11ec4c832d135259149a5225ad14f972473201078b9f058b9ab31fa6b9ff67abR124) and [here](https://github.com/hashicorp/boundary/pull/1684/files#diff-38ab747f010825b0e56d6afa5b7b6cb55af54f24706c918f170120e36e74b1d7R68) which pulls the fields which are not relevant to the type into their own proto message.

internal/servers/controller/handlers/host_sets/host_set_service.go
```diff
@@ -817,7 +820,10 @@ func toProto(ctx context.Context, in host.Set, hosts []host.Host, opt ...handler
 	switch h := in.(type) {
 	case *plugin.HostSet:
 		if outputFields.Has(globals.PreferredEndpointsField) {
-			out.PreferredEndpoints = h.GetPreferredEndpoints()
+			out.PreferredEndpoints = h.PreferredEndpoints
+		}
+		if outputFields.Has(globals.SyncIntervalSecondsField) {
+			out.SyncIntervalSeconds = &wrapperspb.Int32Value{Value: h.GetSyncIntervalSeconds()}
```
talanknight, line 826: consider only setting this if GetSyncIntervalSeconds is not 0.  Since this is in a Int32Value I believe it'll be present despite it being a zero value.

talanknight (approved):
Looks great!

internal/libs/patchstruct/patchstruct_test.go
jefferai: 👍

internal/servers/controller/handlers/host_sets/host_set_service.go
jefferai: Updated

internal/proto/local/controller/storage/host/static/store/v1/static.proto
jefferai: Great idea, thank you!

PR #1682, opened 2021-11-09
Add redirect to managed groups domain model documentation

calcaide:
Add redirect to domain model documentation for managed groups. This is used when the user click the icon help in the managed group screen.

Screen Shot 2021-11-09 at 8 04 13 AM


gsusmi (approved):
LGTM. Ty!

PR #1683, opened 2021-11-09
Call OnCreate or OnUpdate on the plugin after db constraints are checked

talanknight:
This move the calls which could potentially mutate data into the transaction so our db constraints can be checked prior.  The plugins are only called once if successful even if the transaction is retried later.  This doesn't solve all failure cases but arguably avoids calling the plugin when there is a clear error with the request.


jefferai (approved):
Love it

PR #1684, opened 2021-11-10
Add update set handler and sdk functionality

talanknight:


internal/servers/controller/handlers/host_sets/host_set_service.go
```diff
@@ -539,6 +546,47 @@ func (s Service) updateInRepo(ctx context.Context, scopeId, catalogId string, re
...
+	out, hosts, plg, rowsUpdated, err := repo.UpdateSet(ctx, scopeId, h, item.GetVersion(), dbMask)
+	if err != nil {
+		return nil, nil, nil, errors.Wrap(ctx, err, op, errors.WithMsg("unable to update host set"))
+	}
+	if rowsUpdated == 0 {
+		return nil, nil, nil, handlers.NotFoundErrorf("Host Set %q doesn't exist or incorrect version provided.", req.GetId())
+	}
+	var hl []host.Host
+	for _, h := range hosts {
+		hl = append(hl, h)
```
jefferai, line 574: You could just return `hosts` instead of `hl` here -- usually we do this because we're `toProto`-ing the values.

Alternately, replace with `hl := append([]host.Host, hosts...)` or so.

jefferai (approved):
One nit, PR looks great otherwise!

internal/servers/controller/handlers/host_sets/host_set_service.go
talanknight: hosts is of type *plugin.Host and we are needing to return a slice of host.Host.

PR #1686, opened 2021-11-10
internal/host/plugin: add UpdateSet

vancluever:
This adds the UpdateSet repository function.

UpdateSet runs the OnUpdateSet plugin call within the transaction as per
the changes in #1683.

The plugin call returns the updated set, the hosts in the set, the rows
updated, and any error.

### DIFF
--- a/internal/host/plugin/repository_host_set.go
+++ b/internal/host/plugin/repository_host_set.go
@@ -328,7 +328,7 @@ func (r *Repository) UpdateSet(ctx context.Context, scopeId string, s *HostSet, 
 	// the transaction failed and is being retried.
 	var pluginCalledSuccessfully bool
 
-	var recordUpdated bool
+	var setUpdated, preferredEndpointsUpdated bool
 	var returnedSet *HostSet
 	var hosts []*Host
 	_, err = r.writer.DoTx(
@@ -361,7 +361,7 @@ func (r *Repository) UpdateSet(ctx context.Context, scopeId string, s *HostSet, 
 					return errors.New(ctx, errors.MultipleRecords, op, fmt.Sprintf("expected 1 set to be updated, got %d", numUpdated))
 				}
 
-				recordUpdated = true
+				setUpdated = true
 				msgs = append(msgs, &hsOplogMsg)
 			}
 
@@ -383,32 +383,7 @@ func (r *Repository) UpdateSet(ctx context.Context, scopeId string, s *HostSet, 
 				// Only append the oplog message if an operation was actually
 				// performed.
 				if peDeleteOplogMsg.Message != nil {
-					if !recordUpdated {
-						// we only updated secrets, so we need to increment the
-						// version of the host catalog manually.
-						returnedSet = newSet.clone()
-						returnedSet.Version = uint32(version) + 1
-						var hsOplogMsg oplog.Message
-						numUpdated, err := w.Update(
-							ctx,
-							returnedSet,
-							[]string{"version"},
-							[]string{},
-							db.NewOplogMsg(&hsOplogMsg),
-							db.WithVersion(&version),
-						)
-						if err != nil {
-							return errors.Wrap(ctx, err, op)
-						}
-
-						if numUpdated != 1 {
-							return errors.New(ctx, errors.MultipleRecords, op, fmt.Sprintf("expected 1 set to be updated, got %d", numUpdated))
-						}
-
-						recordUpdated = true
-						msgs = append(msgs, &hsOplogMsg)
-					}
-
+					preferredEndpointsUpdated = true
 					msgs = append(msgs, &peDeleteOplogMsg)
 				}
 
@@ -438,32 +413,7 @@ func (r *Repository) UpdateSet(ctx context.Context, scopeId string, s *HostSet, 
 					return err
 				}
 
-				if !recordUpdated {
-					// we only updated secrets, so we need to increment the
-					// version of the host catalog manually.
-					returnedSet = newSet.clone()
-					returnedSet.Version = uint32(version) + 1
-					var hsOplogMsg oplog.Message
-					numUpdated, err := w.Update(
-						ctx,
-						returnedSet,
-						[]string{"version"},
-						[]string{},
-						db.NewOplogMsg(&hsOplogMsg),
-						db.WithVersion(&version),
-					)
-					if err != nil {
-						return errors.Wrap(ctx, err, op)
-					}
-
-					if numUpdated != 1 {
-						return errors.New(ctx, errors.MultipleRecords, op, fmt.Sprintf("expected 1 set to be updated, got %d", numUpdated))
-					}
-
-					recordUpdated = true
-					msgs = append(msgs, &hsOplogMsg)
-				}
-
+				preferredEndpointsUpdated = true
 				msgs = append(msgs, peCreateOplogMsgs...)
 			}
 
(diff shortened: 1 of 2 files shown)


internal/host/plugin/repository_host_set.go
```diff
@@ -152,6 +156,364 @@ func (r *Repository) CreateSet(ctx context.Context, scopeId string, s *HostSet,
...
+			// Flag attributes for updating. While multiple masks may be
+			// sent, we only need to do this once.
+			updateAttributes = true
+
+		default:
+			return nil, nil, nil, db.NoRowsAffected, errors.New(ctx, errors.InvalidFieldMask, op, fmt.Sprintf("invalid field mask: %s", f))
+		}
+	}
+
+	if updateAttributes && s.Attributes != nil {
```
talanknight, line 265: if s.Attributes is nil but updateAttributes is true, then we should remove all attributes from this host set since that means the body of the post was something like
```
{
   "attributes": null
}
```

internal/host/plugin/repository_host_set.go
```diff
@@ -152,6 +156,364 @@ func (r *Repository) CreateSet(ctx context.Context, scopeId string, s *HostSet,
...
+			}
+
+			switch endpointOp {
+			case endpointOpDelete:
+				// Delete all old endpoint entries.
+				var peDeleteOplogMsg oplog.Message
+				pep := &host.PreferredEndpoint{
+					PreferredEndpoint: &store.PreferredEndpoint{
+						HostSetId: newSet.PublicId,
+						Priority:  1,
```
talanknight, line 371: is Priority needed here?   Including it suggests that it is important for the delete query, but I don't think it should be needed.

internal/host/plugin/repository_host_set.go
```diff
@@ -152,6 +156,364 @@ func (r *Repository) CreateSet(ctx context.Context, scopeId string, s *HostSet,
...
+					},
+				}
+				_, err := w.Delete(ctx, pep, db.WithWhere("host_set_id = ?", newSet.PublicId), db.NewOplogMsg(&peDeleteOplogMsg))
```
talanknight, line 374: Cool, I didn't think this would work using `db.WithWhere` work here.  The documentation for w.Delete seems like it should be updated to explain the functionality of the db.WithWhere option since it's not clear to me what data needs to be present in the item being deleted (`pep` in this case).

internal/host/plugin/repository_host_set.go
```diff
@@ -152,6 +156,364 @@ func (r *Repository) CreateSet(ctx context.Context, scopeId string, s *HostSet,
...
+				if err != nil {
+					return errors.Wrap(ctx, err, op)
+				}
+
+				// Only append the oplog message if an operation was actually
+				// performed.
+				if peDeleteOplogMsg.Message != nil {
+					if !recordUpdated {
```
talanknight, line 382: instead of updating the version buried deep in these if statements, keep track of whether the record was updated (like you are with `recordUpdated` and if the preferred endpoints were updated (with a new bool) and update the version accordingly at the end.

internal/host/plugin/repository_host_set.go
```diff
@@ -152,6 +156,364 @@ func (r *Repository) CreateSet(ctx context.Context, scopeId string, s *HostSet,
...
+					}
+
+					recordUpdated = true
+					msgs = append(msgs, &hsOplogMsg)
+				}
+
+				msgs = append(msgs, peCreateOplogMsgs...)
+			}
+
+			if !pluginCalledSuccessfully {
```
talanknight, line 466: Move this to be the last thing that happens in the transaction.

internal/host/plugin/repository_host_set.go
```diff
@@ -152,6 +156,364 @@ func (r *Repository) CreateSet(ctx context.Context, scopeId string, s *HostSet,
...
+				}
+			}
+
+			if returnedSet == nil {
+				// Nothing updated, so returned set will not have been
+				// cloned.  Clone it now so that we can fetch hosts from it.
+				returnedSet = newSet.clone()
+			}
+
+			hosts, err = listHostBySetIds(ctx, r, []string{returnedSet.PublicId}, opt...)
```
talanknight, line 493: Don't do this in the transaction.  The transaction should have the minimum amount of work done inside of it to minimize the chances for errors to happen.

internal/host/plugin/repository_host_set.go
```diff
@@ -152,6 +156,364 @@ func (r *Repository) CreateSet(ctx context.Context, scopeId string, s *HostSet,
...
+				metadata := s.oplog(oplog.OpType_OP_TYPE_UPDATE)
+				if err := w.WriteOplogEntryWith(ctx, oplogWrapper, ticket, metadata, msgs); err != nil {
+					return errors.Wrap(ctx, err, op, errors.WithMsg("unable to write oplog"))
+				}
+			}
+
+			if returnedSet == nil {
+				// Nothing updated, so returned set will not have been
+				// cloned.  Clone it now so that we can fetch hosts from it.
+				returnedSet = newSet.clone()
```
talanknight, line 490: Instead of doing this repeatedly throughout the transaction, just do it once at the very beginning unconditionally.

talanknight (changes requested):

internal/host/plugin/repository_host_set.go
```diff
@@ -152,6 +156,364 @@ func (r *Repository) CreateSet(ctx context.Context, scopeId string, s *HostSet,
...
+	)
+
+	if err != nil {
+		if errors.IsUniqueError(err) {
+			return nil, nil, nil, db.NoRowsAffected, errors.Wrap(ctx, err, op, errors.WithMsg(fmt.Sprintf("in %s: name %s already exists", newSet.PublicId, newSet.Name)))
+		}
+		return nil, nil, nil, db.NoRowsAffected, errors.Wrap(ctx, err, op, errors.WithMsg(fmt.Sprintf("in %s", newSet.PublicId)))
+	}
+
+	var numUpdated int
```
talanknight, line 509: Similar to how we did it in the create set call,  if attributes have been updated call
```
	_ = r.scheduler.UpdateJobNextRunInAtLeast(ctx, setSyncJobName, 0)
```

internal/host/plugin/repository_host_set.go
vancluever: It is. I got an error when attempting to delete without it.

internal/host/plugin/repository_host_set.go
vancluever: Going to leave this for now as per conversation with @talanknight and @jefferai. Possibly need some input from @jimlambrt on whether nor not this is a good idea though. Will add a test to make sure that we get the correct values here.

internal/host/plugin/repository_host_set.go
vancluever: Going to leave this for now as per conversation with @talanknight and @jefferai. Possibly need some input from @jimlambrt on whether nor not this is a good idea though. Will add a test to make sure that we get the correct values here.

internal/host/plugin/repository_host_set.go
vancluever: Done!

internal/host/plugin/repository_host_set.go
vancluever: Done!

internal/host/plugin/repository_host_set.go
vancluever: Done!

internal/host/plugin/repository_host_set.go
vancluever: Done!

internal/host/plugin/repository_host_set.go
vancluever: Done!

internal/host/plugin/repository_host_set.go
jimlambrt: Currently, you must set the primary key for `Delete(...)`.  ICU-2843 would refactor `Delete(...)` would remove this requirement and allow you do issue a delete that only requires a `WithWhere(...)`

talanknight (approved):
Looks great!

internal/host/plugin/repository_host_set.go
vancluever: Done!

internal/host/plugin/repository_host_set.go
vancluever: Added a test for this, but still would like to hear from @jimlambrt still on whether or not this is correct.

vancluever:
Merging based on Todd's comments here.

internal/host/plugin/repository_host_set.go
vancluever: @jimlambrt I think `WithWhere` overrides it though correct? As per https://github.com/hashicorp/boundary/pull/1686#discussion_r747127484. I've added some bits to the the tests that seem to indicate yes.

PR #1688, opened 2021-11-11
Reference PreferredEndpoints by top level struct name

talanknight:
Also reorganizes the sql files a little.


jefferai:
We actually don't need the value in the proto file -- that was literally the one place it was ever referenced. If we can pull it out, but not have mask manager die, that might be the best thing to do.

PR #1691, opened 2021-11-12
Fix secrets clearing bug in update. Changes to update count behavior for catalogs

talanknight:
When updating static catalogs, even if you are changing a name to what it already is, the call will succeed but the version will not be incremented.

Similarly, when we make a call to plugin, even if no primary data or secrets data is updated we should report 1 record updated even if the version has not been incremented.  If the secrets data _has_ been updated then that will be apparent with the version value incremented.

This PR also implements the OnUpdateCatalog  call for the loopback plugin.

The bug this fixes is caused by multiple secret values being passed to update.  The case statement for secrets would be called twice which would set secrets to for both catalogs `c` and `newCatalog`.

### DIFF
--- a/internal/host/plugin/repository_host_catalog.go
+++ b/internal/host/plugin/repository_host_catalog.go
@@ -250,8 +250,6 @@ func (r *Repository) UpdateCatalog(ctx context.Context, c *HostCatalog, version 
 			// field, as secrets do not have a database entry. Clear the
 			// secrets out of the working set after.
 			newCatalog.Secrets = c.Secrets
-			c.Secrets = nil
-
 		default:
 			return nil, nil, db.NoRowsAffected, errors.New(ctx, errors.InvalidFieldMask, op, fmt.Sprintf("invalid field mask: %s", f))
 		}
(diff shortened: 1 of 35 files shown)


internal/host/plugin/repository_host_catalog.go
```diff
@@ -250,8 +250,6 @@ func (r *Repository) UpdateCatalog(ctx context.Context, c *HostCatalog, version
 			// field, as secrets do not have a database entry. Clear the
 			// secrets out of the working set after.
 			newCatalog.Secrets = c.Secrets
-			c.Secrets = nil
```
jefferai, line 253: Why the change? Just curious. Also the comment should probably be updated.

talanknight: In the case where we have 2 secret fields we are updating this code gets executed twice.  The first time assigns value to newCatalog.Secrets and nil's out c.Secrets, the second time through it nils out newCatalog.Secrets since we nil'ed out c.Secrets in the first pass.

talanknight: Updated comment.

jefferai: Huh, ok -- I thought we were only doing wholesale updates of secrets.

talanknight: We are but the field mask is still populated with all the subfields of secrets.

jefferai: Ah! That's what I was missing. Thanks!

jefferai (approved):

PR #1696, opened 2021-11-12
doc: [ssh] Use `%n` for ProxyCommand

gechr:
Not sure if this is a general issue or a macOS issue, but when using the default SSH settings on macOS Monterey (12.0.1), `%h` causes the hostname to be lowercased before being substituted into `ProxyCommand`, which then fails for any target IDs that contain uppercase letters.

Instead, use `%n` because it preserves the case of the hostname.

Ref: https://man7.org/linux/man-pages/man5/ssh_config.5.html#TOKENS


jefferai (approved):
Thanks!

PR #1698, opened 2021-11-12
Update sync interval seconds field.

talanknight:


jefferai (changes requested):
Updating the sync interval shouldn't cause a sync always; it should change the schedule for the next sync. If this is in the past it should be run immediately but if it's in the future it probably shouldn't.

jefferai (approved):
Thanks for making the update!

PR #1702, opened 2021-11-13
docs (db): Update Create and CreateItems godocs

jimlambrt:


jimlambrt:
Deprecated in favor of #1706

PR #1703, opened 2021-11-13
feat(sqltest): Make container port configurable

mgaffney:


internal/db/sqltest/Makefile
```diff
@@ -32,6 +33,7 @@ test:
 	@echo Using $(PG_TAP_DOCKER_IMAGE)
 	@docker run -d \
 		--name $(SQL_TEST_CONTAINER_NAME) \
+		-p $(TEST_DB_PORT):5432 \
```
tmessi, line 36: Can you make this consistent with the `SQL_TEST_CONTAINER_NAME` env var, so make this `SQL_TEST_DB_PORT`

internal/db/sqltest/README.md
```diff
@@ -63,6 +63,15 @@ for available options, i.e:
 make PROVE_OPTS='-j9 -v'
 ```
 
+By default the container uses the host port of 5432.
+This can changed using an environment variable:
```
tmessi, line 67: A host port isn't needed to run the tests, since it uses another container for executing `pg_prove` that is networked with the database container. The only reason for exposing the database would be for debugging tests. 

As such, it might make more sense to just expose an env var to pass through docker args, so  by default this does not use any host port.

tmessi (changes requested):

internal/db/sqltest/Makefile
mgaffney: done

internal/db/sqltest/README.md
mgaffney: I'm not sure what you mean by this. Could you be more specific?

tmessi: For example, in the makefile for the docker image used for the go tests it has this:

https://github.com/hashicorp/boundary/blob/5b326b59f5778f97bab24e40db389e0ea32a7dcd/testing/dbtest/docker/Makefile#L31-L43

So instead of adding `-p $(SQL_TEST_DB_PORT):5432` to the makefile, just add `$(DOCKER_ARGS)` (Note that in the case of the go tests, the container *does* need to expose the port by default so you can run the go tests directly on your host). 

So if you need to debug something in the database container, it can be run like:

```bash
export DOCKER_ARGS="-p 5433:5432"
make database-up
make run-tests
```

But, if `DOCKER_ARGS` is not specified, it would run without exposing the database port to the host.

mgaffney: Got it. I think connecting to the database during development will be a normal workflow so assigning a host port seems reasonable to me. I realize this could be achieved with the `DOCKER_ARGS` solution you are proposing but that requires developers to lookup/remember the docker parameters to set.

tmessi: Best of both worlds?:
```diff
diff --git a/internal/db/sqltest/Makefile b/internal/db/sqltest/Makefile
index 7c6ccd28..70f0f30f 100644
--- a/internal/db/sqltest/Makefile
+++ b/internal/db/sqltest/Makefile
@@ -6,7 +6,12 @@ CWD := $(shell pwd)
 PG_DOCKER_TAG ?= 13-alpine
 # Version of pg_tap docker image
 PG_TAP_DOCKER_TAG ?= pg13
-SQL_TEST_DB_PORT ?= 5432
+SQL_TEST_DB_PORT ?=
+DOCKER_ARGS ?=
+ifneq ($(strip $(SQL_TEST_DB_PORT),)
+	DOCKER_ARGS += "-p $(SQL_TEST_DB_PORT):5432"
+endif
+

 # Pass through options to pg_prove
 # See: https://pgtap.org/pg_prove.html
@@ -33,7 +38,7 @@ test:
 	@echo Using $(PG_TAP_DOCKER_IMAGE)
 	@docker run -d \
 		--name $(SQL_TEST_CONTAINER_NAME) \
-		-p $(SQL_TEST_DB_PORT):5432 \
+		$(DOCKER_ARGS) \
 		-e POSTGRES_PASSWORD=boundary \
 		-e POSTGRES_USER=boundary \
 		-e POSTGRES_DB=boundary \
@@ -57,7 +62,7 @@ database-up:
 	@echo Using $(POSTGRES_DOCKER_IMAGE)
 	@docker run -d \
 		--name $(SQL_TEST_CONTAINER_NAME) \
-		-p $(SQL_TEST_DB_PORT):5432 \
+		$(DOCKER_ARGS) \
 		-e POSTGRES_PASSWORD=boundary \
 		-e POSTGRES_USER=boundary \
 		-e POSTGRES_DB=boundary \
```

mgaffney: Good idea!

tmessi (approved):

PR #1704, opened 2021-11-13
feat(credential): Use credential type when issuing credentials

mgaffney:
This change uses the credential type attribute when issuing credentials
including mapping overrides.

### DIFF
--- a/internal/credential/vault/private_library.go
+++ b/internal/credential/vault/private_library.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/hashicorp/boundary/internal/credential"
 	"github.com/hashicorp/boundary/internal/db/timestamp"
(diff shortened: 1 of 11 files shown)


internal/credential/vault/private_library.go
```diff
@@ -5,91 +5,146 @@ import (
...
+	return bc, nil
+}
+
+func baseToUsrPass(ctx context.Context, bc *baseCred) (*usrPassCred, error) {
+	uAttr, pAttr := bc.lib.UsernameAttribute, bc.lib.PasswordAttribute
+	if uAttr == "" {
+		uAttr = "username"
+	}
+	if pAttr == "" {
+		uAttr = "password"
```
louisruch, line 61:
```suggestion
		pAttr = "password"
```

louisruch (approved):
Took a quick look, will review properly tomorrow.  But found one bug

internal/credential/vault/private_library.go
```diff
@@ -5,91 +5,146 @@ import (
...
+	}
+
+	var username, password string
+	if u, ok := bc.secretData[uAttr]; ok {
+		username = fmt.Sprintf("%v", u)
+	}
+	if p, ok := bc.secretData[pAttr]; !ok {
```
louisruch, line 68: Haha, so I was trying to figure out why the test was passing given the not setting `pAttr` above, the `!ok` should become `ok`.

What this was doing was setting `username` = passwordAttribue and `password` = `<nil>` :)

mgaffney: I fixed it and added a bunch of tests.

tmessi (approved):

PR #1705, opened 2021-11-13
Add external ID to host output

jefferai:
This also fixes a bug if no hosts were returned where the job would run
over and over in a loop, and some other cleanup.


internal/servers/controller/handlers/options.go
```diff
@@ -110,3 +111,11 @@ func WithMemberIds(ids []string) Option {
 		o.WithMemberIds = ids
 	}
 }
+
+// WithHostSetIds provides an option when creating responses to include the
+// given host set IDs if allowed
+func WithHostSetIds(ids []string) Option {
```
jimlambrt, line 117: Do we want a test for this new option?

jimlambrt (approved):
We might want to add one minor test.  Otherwise, looks great.

internal/servers/controller/handlers/options.go
jefferai: Will add!

PR #1706, opened 2021-11-14
Add more opts when using Create + WithOnConflict

jimlambrt:
-- feature (db): Add more opts when using Create + WithOnConflict 
-- refactor (db): Migrate Update(...) to use whereClausesFromOpts(...)
-- tests (db): Add missing assert for WithDebug unit tests


jefferai (approved):
LGTM! :shipit:

PR #1707, opened 2021-11-15
Increment version of host during updates

jefferai:
Currently host does not use a version, which means it's not protected
against race conditions of either itself or its value objects. This adds
and plumbs through version to ensure that if a host or its value objects
change that they are not going to be affected by a race condition.

### DIFF
--- a/internal/host/plugin/job_set_sync.go
+++ b/internal/host/plugin/job_set_sync.go
@@ -414,22 +414,36 @@ func (r *SetSyncJob) upsertHosts(
 			db.ExpBackoff{},
 			func(r db.Reader, w db.Writer) error {
 				msgs := make([]*oplog.Message, 0)
-				ticket, err := w.GetTicket(hi.h)
+				ticket, err := w.GetTicket(ret)
 				if err != nil {
 					return errors.Wrap(ctx, err, op, errors.WithMsg("unable to get ticket"))
 				}
 
-				if hi.dirtyHost {
-					var hOplogMsg oplog.Message
-					onConflict := &db.OnConflict{
-						Target: db.Constraint("host_plugin_host_pkey"),
-						Action: db.SetColumns([]string{"name", "description"}),
-					}
-					if err := w.Create(ctx, ret, db.NewOplogMsg(&hOplogMsg), db.WithOnConflict(onConflict)); err != nil {
-						return errors.Wrap(ctx, err, op)
-					}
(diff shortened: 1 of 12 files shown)


internal/host/plugin/job_set_sync.go
```diff
@@ -466,7 +480,7 @@ func (r *SetSyncJob) upsertHosts(
 							return err
```
jimlambrt, line 480: Does this code path need test coverage?

jimlambrt (changes requested):
Just one min question about testing...

internal/host/plugin/job_set_sync.go
jefferai: Should be tested via https://github.com/hashicorp/boundary/pull/1707/files/ddadd262d8c89e309a1dc7ca83b8088fac5b450f

Originally the upsert logic was in the repo but Todd pulled them into the job, so the test functions are a bit split...

jimlambrt: ty!

jimlambrt (approved):

internal/host/plugin/job_set_sync.go
```diff
@@ -451,7 +465,7 @@ func (r *SetSyncJob) upsertHosts(
 							Action: db.DoNothing(true),
 						}
 						if err := w.CreateItems(ctx, hi.ipsToAdd.toArray(), db.NewOplogMsgs(&oplogMsgs), db.WithOnConflict(onConflict)); err != nil {
-							return errors.Wrap(ctx, err, op, errors.WithMsg(fmt.Sprintf("adding ips %v for host %q", hi.ipsToAdd.toArray(), hi.h.GetPublicId())))
+							return errors.Wrap(ctx, err, op, errors.WithMsg(fmt.Sprintf("adding ips %v for host %q", hi.ipsToAdd.toArray(), ret.GetPublicId())))
```
talanknight, line 468: Just taking the chance to get your input on this error message.  I added it kind of incidentally on friday to help debug.  Prior there was no WithMsg attached.  I'm worried the msg is too verbose.

talanknight (approved):

internal/host/plugin/job_set_sync.go
jefferai: I think it's probably okay. This won't go to a user; it'd go to observability. And the number of IPs should be pretty small.

PR #1708, opened 2021-11-15
feature(config): Enable more config values to be sourced from env/file

hugoamvieira:
We allow fields like Controller Name and Worker name to be specified `env://<ENV_VAR>` or `file://<FILE_PATH>`values (support from `github.com/hashicorp/go-secure-stdlib`). This library will look at the environment variables or file respectively, extract the values and place them in the variable.

This task is meant to look into other fields that could benefit from this functionality and add them.

### DIFF
--- a/internal/cmd/config/config.go
+++ b/internal/cmd/config/config.go
@@ -298,16 +298,11 @@ func Parse(d string) (*Config, error) {
 
 	// Perform controller configuration overrides for auth token settings
 	if result.Controller != nil {
-		result.Controller.Name, err = parseutil.ParsePath(result.Controller.Name)
-		if err != nil && !errors.Is(err, parseutil.ErrNotAUrl) {
+		result.Controller.Name, err = parsePath(result.Controller.Name)
+		if err != nil {
 			return nil, fmt.Errorf("Error parsing controller name: %w", err)
 		}
-		if result.Controller.Name != strings.ToLower(result.Controller.Name) {
-			return nil, errors.New("Controller name must be all lower-case")
-		}
-		if !strutil.Printable(result.Controller.Name) {
-			return nil, errors.New("Controller name contains non-printable characters")
-		}
+
 		if result.Controller.AuthTokenTimeToLive != "" {
 			t, err := parseutil.ParseDurationSecond(result.Controller.AuthTokenTimeToLive)
 			if err != nil {
@@ -327,16 +322,11 @@ func Parse(d string) (*Config, error) {
 
 	// Parse worker tags
 	if result.Worker != nil {
-		result.Worker.Name, err = parseutil.ParsePath(result.Worker.Name)
-		if err != nil && !errors.Is(err, parseutil.ErrNotAUrl) {
+		result.Worker.Name, err = parsePath(result.Worker.Name)
+		if err != nil {
 			return nil, fmt.Errorf("Error parsing worker name: %w", err)
 		}
-		if result.Worker.Name != strings.ToLower(result.Worker.Name) {
-			return nil, errors.New("Worker name must be all lower-case")
-		}
-		if !strutil.Printable(result.Worker.Name) {
-			return nil, errors.New("Worker name contains non-printable characters")
-		}
+
 		if result.Worker.TagsRaw != nil {
 			switch t := result.Worker.TagsRaw.(type) {
 			// HCL allows multiple labeled blocks with the same name, turning it
@@ -439,6 +429,22 @@ func Parse(d string) (*Config, error) {
 	}
 
 	return result, nil
+}
+
+func parsePath(urlField string) (string, error) {
+	// input is returned in error states to preserve `parseutil.ParsePath` functionality.
+	field, err := parseutil.ParsePath(urlField)
+	if err != nil && !errors.Is(err, parseutil.ErrNotAUrl) {
+		return urlField, err
+	}
+	if field != strings.ToLower(field) {
+		return urlField, errors.New("field must be all lower-case")
+	}
+	if !strutil.Printable(field) {
+		return urlField, errors.New("field contains non-printable characters")
+	}
+
+	return field, nil
 }
 
 func parseEventing(eventObj *ast.ObjectItem) (*event.EventerConfig, error) {
--- a/internal/cmd/config/config_test.go
+++ b/internal/cmd/config/config_test.go
@@ -531,36 +531,42 @@ func TestParsingPath(t *testing.T) {
 		name      string
 		in        string
 		expOut    string
+		expErr    bool
 		expErrStr string
 	}{
 		{
 			name:      "env: not present",
 			in:        "env://TEST_ENV_FIELD_THAT_SURELY_DOESNT_EXIST",
 			expOut:    "",
+			expErr:    false,
 			expErrStr: "",
 		},
 		{
 			name:      "file: not present",
 			in:        "file://test_file_that_surely_doesnt_exist",
 			expOut:    "file://test_file_that_surely_doesnt_exist",
+			expErr:    true,
 			expErrStr: "error reading file at file://test_file_that_surely_doesnt_exist: open test_file_that_surely_doesnt_exist: no such file or directory",
 		},
 		{
 			name:      "not a url",
 			in:        "some-value",
 			expOut:    "some-value",
+			expErr:    false,
 			expErrStr: "",
 		},
 		{
 			name:      "upper case value",
 			in:        "some-VALUE",
 			expOut:    "some-VALUE",
+			expErr:    true,
 			expErrStr: "field must be all lower-case",
 		},
 		{
 			name:      "non-printable value",
 			in:        "some-val\u0000ue",
 			expOut:    "some-val\u0000ue",
+			expErr:    true,
 			expErrStr: "field contains non-printable characters",
 		},
 	}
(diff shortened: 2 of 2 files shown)


internal/cmd/config/config_test.go
```diff
@@ -524,3 +524,58 @@ func TestController_EventingConfig(t *testing.T) {
...
+			in:        "some-val\u0000ue",
+			expOut:    "some-val\u0000ue",
+			expErrStr: "field contains non-printable characters",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			out, err := parsePath(tt.in)
+			if len(tt.expErrStr) > 0 {
```
louisruch, line 571: nit: in most of the test I have encountered we create a `bool` to explicitly set the expectation of an error instead of checking the string.  I feel this helps in ready the test, usually its `wantErr` but `expErr` should be good

internal/cmd/config/config_test.go
```diff
@@ -524,3 +524,58 @@ func TestController_EventingConfig(t *testing.T) {
 		})
 	}
 }
+
+func TestParsingPath(t *testing.T) {
```
louisruch, line 528: Would be good to create a temp file and set a `os.SetEnv` to fully test the happy path

louisruch (commented):
Laptop about to die going to find a place with power but have two comments in the mean time

internal/cmd/config/config_test.go
hugoamvieira: In those tests, do we assert the error itself? 
i.e: Do we have a boolean to assert whether we expect an error or not and _then_ we check against some other field to assert the correct_error was thrown, or do we just use the boolean to say _an_ error _should_ be thrown?

internal/cmd/config/config_test.go
hugoamvieira: I thought about this, but I didn't end up going that path since that functionality should be the responsibility of `parseutil.ParsePath` to get right, so I was divided as to whether I should've included that in these tests or not.

I can certainly add it though, not much more work really!

internal/cmd/config/config_test.go
louisruch: Great question, so we generally have the following
1) Use a bool to set expectation of an error or not. In the test we then ‘if-else’ that book and require error or noError
2) Inside the ‘if’ we often also have a error ‘Is’ check.  In this case you won’t be able to check that
3) Finally we often have an error string check like you did here with ErrorEqual. Sometimes we do not do this as it can make the tests brittle (if the error string changes all tests have to change, and any test in files consuming this package higher up might have to change).  However, I prefer to add the error string check to tests because it helps me visualize the full error message.  In this case since the errors are in the cmd package I would argue we should definitely assert the error string

hugoamvieira: Makes sense, thanks for clarifying! 🙇 

With that in mind, we should add a `wantErr` or `expErr` boolean as that'll probably improve readability later down in the test exec and then when we _do_expect an error, match against `wantErrStr` or `expErrStr` to ensure it's the expected one.

WDYT?

louisruch: SGTM. I have a slight preference to want, as I feel it reads better but exp is used in other tests in this package so as you prefer

hugoamvieira:
I've been walking the code and checking where we could add this functionality, and here are my findings:

In  `go-secure-stdlib`, we have `SharedConfig`, here's an abbreviated version with some of my comments:
```go
type SharedConfig struct {
	EntSharedConfig // N/A, no applicable fields within.

	Listeners []*listenerutil.ListenerConfig // N/A, not string

	Seals []*KMS // N/A, not string
	Entropy *Entropy // N/A, not string

	DisableMlock bool // N/A, not string
	DisableMlockRaw interface{} // N/A, not string

	Telemetry interface{} // N/A, not string

	DefaultMaxRequestDuration time.Duration // N/A, not string
	DefaultMaxRequestDurationRaw interface{} // N/A, not string

	LogFormat string // Could be implemented, but it's already wrapped by Boundary CLI in BOUNDARY_LOG_FORMAT.
	LogLevel string // Could be implemented, but it's already wrapped by Boundary CLI in BOUNDARY_LOG_LEVEL.

	PidFile string // N/A, is already a file

	ClusterName string // I'm not sure what this is used for, but could technically be implemented.
}
```

Then, in our own Boundary code, we have the `Config` struct and its dependents.
Here are the abbreviated structs, with some of my comments.
```go
type Config struct {
	*configutil.SharedConfig // Discussed above

	Worker *Worker // Discussed below
	Controller *Controller // Discussed below

	DevController bool // N/A, not string.
	DevUiPassthroughDir string // Possible contender, but already wrapped by Boundary CLI on BOUNDARY_DEV_UI_PASSTHROUGH_DIR
	DevControllerKey string // Possible contender, but why not make it wrapped by Boundary CLI like the others?
	DevWorkerAuthKey string // Possible contender, but already wrapped by Boundary CLI on BOUNDARY_DEV_WORKER_AUTH_KEY
	DevRecoveryKey string // Possible contender, but already wrapped by Boundary CLI on BOUNDARY_DEV_RECOVERY_KEY

	Eventing *event.EventerConfig // N/A - no applicable fields within.

	Plugins Plugins // Discussed below
}

type Controller struct {
	Name string // Already in (config.go Parse())
	Description string // Possible contender
	Database *Database // Discussed below
	PublicClusterAddr string // Already in (servers.go SetupControllerPublicClusterAddress())

	AuthTokenTimeToLive interface{} // N/A, not string
	AuthTokenTimeToLiveDuration time.Duration // N/A, not string

	AuthTokenTimeToStale interface{} // N/A, not string
	AuthTokenTimeToStaleDuration time.Duration // N/A, not string

	StatusGracePeriodDuration time.Duration // N/A, not string

	SchedulerRunJobInterval time.Duration // N/A, not string
}

type Worker struct {
	Name string // Already in (config.go Parse())
	Description string // Possible contender
	Controllers []string // N/A, not string
	PublicAddr string // Already in (servers.go SetupWorkerPublicAddress() L677)

	TagsRaw interface{}  // N/A, not string
	Tags    map[string][]string // N/A, not string

	StatusGracePeriodDuration time.Duration // N/A, not string
}


type Database struct {
	Url string // Already done in Run() of `boundary server` - commands/server/server.go
	MigrationUrl string // Already done in Run() of `boundary database migrate` - commands/database/init.go
	MaxOpenConnections int // N/A, not string
}

type Plugins struct {
	ExecutionDir string // Possible contender, but already wrapped by Boundary CLI on BOUNDARY_DEV_PLUGIN_EXECUTION_DIR
}
```

In terms of the task at hand, my thoughts come down not to adding support for _more_ fields, as it's already pretty comprehensive, but to standardise this to a centralized function where possible - We seem to do a lot of this parsing in different places in the code, and, while I'm not 100% sure about this, on certain fields we might end up even doing these `ParsePath` calls multiple times per field over the code execution path.

Also, I found this method very limiting because we rely on the field being of string value so we can obtain whatever `env://` or `file://` the user set it to, and replace it with the actual string value - The field must be a string and a string only, whereas with other methods (such as the CLI env variables), we can grab the value from the flag and interpret it into other types, so we have a bit more flexibility there.

internal/cmd/config/config.go
```diff
@@ -441,6 +431,22 @@ func Parse(d string) (*Config, error) {
...
+	field, err := parseutil.ParsePath(urlField)
+	if err != nil && !errors.Is(err, parseutil.ErrNotAUrl) {
+		return urlField, err
+	}
+	if field != strings.ToLower(field) {
+		return urlField, errors.New("field must be all lower-case")
+	}
+	if !strutil.Printable(field) {
+		return urlField, errors.New("field contains non-printable characters")
+	}
```
tmessi, line 445: I am not sure that these checks belong here. While this DRYs things up, in particular the lower case check may not apply in all cases where we might want to use `parseutil.ParsePath`. Similarly, while it probably makes sense to check for non-printable characters in all cases, it a separate concern from what `parseutil.ParsePath` accomplishes.

hugoamvieira: Yep good point. I'm thinking about changing the function signature like this

`func parsePath(urlField string, opts ...parseFieldOpts) (string, error)`

So you could pass some options in to say what you want to be checked. Something like:

```go
parsed, err := parsePath("env://blablabla", withLowerCaseOnlyCheck, withPrintableCheck)
```

I'm a bit torn on this since it seems so overkill for something easy. WDYT?

tmessi: Yea, I think that is overkill, and I think it will still mix concerns. The `parseutil.ParsePath` is concerned with parsing the initial value to extract the true value that might be contained in either an `env://` or `file://` (and in theory additional "paths" in the future). The other two checks are concerned with validation, not parsing. I could potentially see combining those two validation steps into a function, there seems to be a common thing here where a "Name" field should:
- be all lower case
- not contain non-printable characters

But I don't think it should be grouped with the parsing.

hugoamvieira: Perhaps then having another function which enforces these checks that then calls to `parsePath`?

hugoamvieira: Something like this

```go
func parseNamePath(nameField string) (string, error) {
	field, err := parsePath(nameField)
	if err != nil {
		return nameField, err
	}

	if field != strings.ToLower(field) {
		return nameField, errors.New("field must be all lower-case")
	}
	if !strutil.Printable(field) {
		return nameField, errors.New("field contains non-printable characters")
	}

	return field, nil
}

func parsePath(urlField string) (string, error) {
	// input is returned in error states to preserve `parseutil.ParsePath` functionality.
	field, err := parseutil.ParsePath(urlField)
	if err != nil && !errors.Is(err, parseutil.ErrNotAUrl) {
		return urlField, err
	}

	return field, nil
}
```

hugoamvieira:
Discussed offline - Closing as the refactor above isn't really needed anymore given what we discovered digging into this area of the code.

PR #1710, opened 2021-11-15
Add plugin-based host entry to CHANGELOG.md

vancluever:


CHANGELOG.md
```diff
@@ -11,6 +11,12 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 
 ### New and Improved
 
+* Boundary now supports _plugin-based_ host catalogs, host sets, and hosts. This
```
talanknight, line 14: Do we want to advertise unqualified use of plugins for creating these host catalogs?  I'm fine if we do but an alternative would be to say something like "... supports dynamic discovery of host resources using our internal (for now) plugin system..."

Otherwise looks great!

vancluever: Yeah, no problem. I can change that. I agree, it might be premature to use this language, in lieu of this consideration I was just using our domain language. Will fix!

talanknight (approved):

PR #1711, opened 2021-11-16
fix (kms): Add kms.ReconcileKeys(...) to fix missing audit DEK scenario

jimlambrt:
This fix address the situation where an existing boundary database
doesn't have the audit DEKs in the global scope.

### DIFF
--- a/internal/kms/kms.go
+++ b/internal/kms/kms.go
@@ -3,6 +3,8 @@ package kms
 import (
 	"context"
 	"fmt"
+	"io"
+	"reflect"
 	"sync"
 
 	"github.com/hashicorp/boundary/internal/db"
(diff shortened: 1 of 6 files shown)


internal/kms/kms.go
```diff
@@ -269,6 +271,43 @@ func (k *Kms) loadRoot(ctx context.Context, scopeId string, opt ...Option) (*mul
...
+func (k *Kms) ReconcileKeys(ctx context.Context, randomReader io.Reader) error {
+	const op = "kms.ReconcileKeys"
+	if isNil(randomReader) {
+		return errors.New(ctx, errors.InvalidParameter, op, "missing rand reader")
+	}
+
+	// it's possible that the global audit key was created after this instance's
+	// database was initialized... so check if the audit wrapper is available
+	// for the global scope and if not, then add one to the global scope
+	if _, err := k.GetWrapper(ctx, scope.Global.String(), KeyPurposeAudit); err != nil {
```
jefferai, line 284: It'd be better if there was a way of ensuring that the error is because the key doesn't exist rather than something else going wrong, e.g. switching on error type.

internal/kms/kms.go
```diff
@@ -269,6 +271,43 @@ func (k *Kms) loadRoot(ctx context.Context, scopeId string, opt ...Option) (*mul
...
+			return errors.Wrap(ctx, err, op, errors.WithMsg(fmt.Sprintf("error generating random bytes for database key in scope %s", scope.Global.String())))
+		}
+		if _, _, err := k.repo.CreateAuditKey(ctx, globalRootWrapper, key); err != nil {
+			return errors.Wrap(ctx, err, op, errors.WithMsg(fmt.Sprintf("error creating audit key in scope %s", scope.Global.String())))
+		}
+	}
+	return nil
+}
+
+func isNil(i interface{}) bool {
```
jefferai, line 300: Curious about this -- do you foresee randomReader being nil in a way that isn't simply a nil interface?

jefferai (commented):
Looks good -- if the error checking can be improved, great. If not, probably good enough for now.

internal/kms/kms.go
jimlambrt: ty!

internal/kms/kms.go
jimlambrt: I added a test to illustrate what this protects against.

jefferai (approved):
LGTM!

tmessi (approved):

PR #1712, opened 2021-11-16
Add secrets_hmac to host catalog output

jefferai:
This field is updated _when new secrets are passed via the API_. This value does not depend on what the plugin does internally with those values (e.g. if it rotates them) since we can't assume the values won't be changed. The point is to be able to determine if something has updated secrets outside the purview of e.g. Terraform; regardless of any internal rotation, if the value is the same, it hasn't been modified by a user.


internal/host/plugin/host_catalog.go
```diff
@@ -54,6 +57,35 @@ func allocHostCatalog() *HostCatalog {
...
+	if cipher == nil {
+		return errors.New(ctx, errors.InvalidParameter, op, "missing cipher")
+	}
+	// Go's JSON encoding is stable (that is, it alphabetizes keys) so it's a
+	// good option to produce an HMAC-able string.
+	jsonSecrets, err := json.Marshal(secretsMap)
+	if err != nil {
+		return errors.Wrap(ctx, err, op, errors.WithCode(errors.Code(errors.Encryption)))
+	}
+	hm, err := crypto.HmacSha256(ctx, jsonSecrets, cipher, []byte(c.PublicId), nil, crypto.WithBase64Encoding())
```
jimlambrt, line 81: It's great to see the new crypto lib functions being used.

jimlambrt (approved):
LGTM!

internal/db/schema/migrations/oss/postgres/20/08_plugin_host_views.up.sql
```diff
@@ -6,12 +6,11 @@ alter table host_plugin_host
 create trigger update_version_column after update on host_plugin_host
   for each row execute procedure update_version_column();
 
-drop view host_plugin_host_with_value_obj_and_set_memberships;
-
 -- host_plugin_host_with_value_obj_and_set_memberships is useful for reading a
 -- plugin host with its associated value objects (ip addresses, dns names) and
 -- set membership as columns with delimited values. The delimiter depends on
 -- the value objects (e.g. if they need ordering).
+drop view if exists host_plugin_host_with_value_obj_and_set_memberships;
```
talanknight, line 13: shouldn't we know for sure if the view exists or not at this point?

jimlambrt (dismissed):

internal/proto/local/controller/storage/host/plugin/store/v1/host.proto
```diff
@@ -42,9 +42,15 @@ message HostCatalog {
   // @inject_tag: `gorm:"default:null"`
   uint32 version = 8;
 
+  // secrets_hmac is a sha256-hmac of the unencrypted secrets that is returned
+  // from the API for read.  It is recalculated everytime the raw secrets are
+  // updated.
+  // @inject_tag: `gorm:"default:null"`
+  string secrets_hmac = 9;
```
talanknight, line 49: This should probably be bytes.  We can't use use the value for anything except to return it.

internal/db/schema/migrations/oss/postgres/20/10_plugin_host_secrets_hmac.up.sql
```diff
@@ -0,0 +1,33 @@
+begin;
+
+alter table host_plugin_catalog
+  add column secrets_hmac text
```
talanknight, line 4: Instead of text, this should probably be bytes since the purpose of this field is only to return it and we can't do any other string related operations on it that would be meaningful.  I'm channeling my inner @mgaffney here :).

internal/db/schema/migrations/oss/postgres/20/08_plugin_host_views.up.sql
jefferai: Yes, we do know it exists, but this is just extra safety.

internal/proto/local/controller/storage/host/plugin/store/v1/host.proto
jefferai: It goes in as a base64'd string. Is it still useful to put it in as bytes? I can, just not sure of the difference.

talanknight: The token_hmac for credential_vault_token is stored as bytes (see [this](https://github.com/hashicorp/boundary/blob/main/internal/proto/local/controller/storage/credential/vault/store/v1/vault.proto#L79) and [this](https://github.com/hashicorp/boundary/blob/main/internal/db/schema/migrations/oss/postgres/10/04_vault_credential.up.sql#L148) ) as well as most other hmacs (except maybe an oidc hmac that might has slipped through the cracks)

internal/db/schema/migrations/oss/postgres/20/10_plugin_host_secrets_hmac.up.sql
jimlambrt: I'm not sure this is an arg for storing it as bytes over text.  If we're only returning a base64 encoded format and that's all we're ever going to do, why store any other format over that?

internal/proto/local/controller/storage/host/plugin/store/v1/host.proto
jimlambrt: Yes, we can only use the value as a base64 encoded string, so why store it in mem or the db in any other format?

internal/db/schema/migrations/oss/postgres/20/08_plugin_host_views.up.sql
talanknight: Just a nit, so I'll go ahead and resolve the conversation after this follow up.

If we know it should exist but it doesn't then that should be a problem we know about.  Tim or Mike (I forget which) suggested I stop using "create or replace function" for this purpose when I know the function wasn't defined previously.

mgaffney: I agree with @talanknight. If the view does not exist it is an error and this migration should fail.

internal/proto/local/controller/storage/host/plugin/store/v1/host.proto
mgaffney: Why are we base64 encoding it? The hmac is binary data not a string. That fact that we are encoding binary data into a string doesn't change the fact that the data itself is binary data.

internal/db/schema/migrations/oss/postgres/20/10_plugin_host_secrets_hmac.up.sql
mgaffney: A couple of reasons off the top of my head:

* Text fields in the database are UTF-8 but [base64](https://en.wikipedia.org/wiki/Base64) is limited to a subset of ASCII characters.
* Any normal text transformations to the data will cause an error when the text is decoded. For example, returning the first 10-20 characters followed by an ellipse to represent a clipping, would be meaningless and cause errors.
* Base64 is for encoding binary data into text. The output of the encoding is not meant to be human readable.

internal/proto/local/controller/storage/host/plugin/store/v1/host.proto
jimlambrt: while true, the data only has one format for the entire domain, which is a string not bytes.

internal/db/schema/migrations/oss/postgres/20/10_plugin_host_secrets_hmac.up.sql
jimlambrt: * ??? you're first point would seem to suggest that there's some problem storing a subset of ASCII characters as UTF-8.  If there is some holistic problem with storing a char like `A` in utf-8 then we've have lots of other issues.
* Any clipping of hmac data is meaningless and will cause errors. 
* All binary data that's not encoded, will be encoded when it's returned json since by definition that's what it does.

internal/proto/local/controller/storage/host/plugin/store/v1/host.proto
jefferai: Any string is binary data :-D

internal/db/schema/migrations/oss/postgres/20/10_plugin_host_secrets_hmac.up.sql
jefferai: >Any normal text transformations to the data will cause an error when the text is decoded.
The text is never decoded. It is only ever used as a string in string comparisons by e.g. Terraform.

internal/proto/local/controller/storage/host/plugin/store/v1/host.proto
mgaffney: After discussing with @jimlambrt, we recommend this be switched to bytes.

internal/db/schema/migrations/oss/postgres/20/10_plugin_host_secrets_hmac.up.sql
mgaffney: After discussing with @jimlambrt, we recommend the database column type be changed to `bytea`.

mgaffney (approved):

PR #1714, opened 2021-11-16
api: Returns connections as part of sessions 

thorfour:
Adds connections to the session object in the api layer


internal/proto/local/controller/api/resources/sessions/v1/session.proto
```diff
@@ -23,6 +23,35 @@ message SessionState {
   google.protobuf.Timestamp end_time = 30 [json_name = "end_time"];
 }
```
jimlambrt, line 25: Mind adding a doc comment for the `Connection` message?

jimlambrt (dismissed):
Looks great!  One very small request before you merge.

internal/proto/local/controller/api/resources/sessions/v1/session.proto
thorfour: Done.

jimlambrt (approved):

internal/proto/local/controller/api/resources/sessions/v1/session.proto
```diff
@@ -23,6 +23,36 @@ message SessionState {
...
 
+// Connection contains information about a specific connection in a session
+message Connection {
+    // public_id is used to access the connection via an API
+    string public_id = 1;
```
talanknight, line 29: Right now it looks like the only place a connection exists is as part of a session.  Why does it need a public id (These should only be called `id` in api resources) and a session id?  Are connections going to be able to be modified and retrieved through a different API endpoint?

If so they should be defined in their own directory to match the pattern we follow for all the other public API resources.  Then for the resource representation in the context of a session you would have an abbreviated proto message like we have for [group members](https://github.com/hashicorp/boundary/blob/main/internal/proto/local/controller/api/resources/groups/v1/group.proto#L12)  or [accounts on a user](https://github.com/hashicorp/boundary/blob/main/internal/proto/local/controller/api/resources/users/v1/user.proto#L12).

PR #1731, opened 2021-11-18
feature(config): Add go-sockaddr templating support to Controller and Worker Addresses

hugoamvieira:
On the back of https://github.com/hashicorp/go-secure-stdlib/pull/14, add support for [this templating](https://pkg.go.dev/github.com/hashicorp/go-sockaddr/template) as a way to set up Controller and Worker addresses in config.


internal/cmd/base/servers.go
```diff
@@ -692,6 +707,10 @@ func (b *Server) SetupWorkerPublicAddress(conf *config.Config, flagValue string)
 	return nil
 }
 
+func looksLikeTemplate(s string) bool {
+	return strings.HasPrefix(s, "{{") && strings.HasSuffix(s, "}}")
+}
+
```
hugoamvieira, line 713: This is not my favourite thing in the world, but in my head the alternative is actually trying and checking the error, and even that isn't 100% reliable because `ParseSingleIPTemplate` doesn't error in cases like, for example, `"bla"` is input.

Do we have any different ideas here, or do we think this is OK?

internal/cmd/base/servers.go
```diff
@@ -672,13 +680,20 @@ func (b *Server) SetupWorkerPublicAddress(conf *config.Config, flagValue string)
 				}
 			}
 		}
+	} else if looksLikeTemplate(conf.Worker.PublicAddr) {
```
jefferai, line 683: If you make this a separate else case, you forego running the value through ParsePath below. Instead, I think the value you should pass into the template function should be the output from ParsePath. You don't need to check if it looks like a template or not -- if it's not a template, ParseSingleIPTemplate should just return the string unmodified (judging from [how it's used in listenerutil](https://github.com/hashicorp/go-secure-stdlib/blob/main/listenerutil/parse.go#L111)).

tmessi (dismissed):

jefferai (commented):
Looks good...can you update the website docs for those parameters? Also please add a changelog entry.

hugoamvieira:
@jefferai In terms of the Changelog, I'd have to put this change under a version right? Should I just bump it to 0.7.2?

tmessi (approved):

PR #1732, opened 2021-11-18
Don't always assume plugins should be activated

jefferai:
When running test controllers from e.g. Terraform tests we don't have the external plugins built in, so don't always include them. Additionally, add the loopback plugin to the test controller.

### DIFF
--- a/internal/servers/controller/controller.go
+++ b/internal/servers/controller/controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/rand"
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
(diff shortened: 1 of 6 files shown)


internal/servers/controller/testing.go
```diff
@@ -414,6 +416,17 @@ func NewTestController(t *testing.T, opts *TestControllerOpts) *TestController {
 		t.Fatal(err)
 	}
 
+	tc.b.DevLoopbackHostPluginId = DefaultTestPluginId
+	plg := plugin.NewWrappingPluginClient(plugin.NewLoopbackPlugin())
+	plugOpts := []hostplugin.Option{
+		hostplugin.WithDescription("Provides an initial loopback host plugin in Boundary"),
+		hostplugin.WithPublicId(tc.b.DevLoopbackHostPluginId),
+	}
+	if _, err = tc.b.RegisterHostPlugin(ctx, "loopback", plg, plugOpts...); err != nil {
```
talanknight, line 425: Please remove the part of this file which adds the loopback plugin already (see the section close to `CreateDevDatabase` in TestControllerConfig)

internal/servers/controller/controller.go
```diff
@@ -119,32 +123,38 @@ func New(ctx context.Context, conf *Config) (*Controller, error) {
...
-		external_host_plugins.WithLogger(hclog.NewNullLogger()))
-	if err != nil {
-		return nil, fmt.Errorf("error creating aws host plugin")
-	}
-	conf.ShutdownFuncs = append(conf.ShutdownFuncs, awsCleanup)
-	if _, err := conf.RegisterHostPlugin(ctx, "aws", awsSvcClient, hostplugin.WithDescription("Built-in AWS host plugin")); err != nil {
-		return nil, fmt.Errorf("error registering aws host plugin: %w", err)
+	for _, enabledPlugin := range c.enabledPlugins {
+		switch enabledPlugin {
+		case base.EnabledPluginHostAzure:
```
talanknight, line 128: These 2 case statements seem very similar.  Consider containing the plugin information (like the name and... is that it?) with the EnabledPluginHost type so there is a centralized place for that info and this code can avoid being duplicated.

talanknight (changes requested):

internal/servers/controller/testing.go
jefferai: TIL we already have loopback in test controllers!

internal/servers/controller/controller.go
jefferai: Good idea, fixed!

talanknight (approved):

PR #1733, opened 2021-11-18
Fix Windows

jefferai:
The embed package always uses Unix-style forward slashes. filepath.Join
utilization looking for files would cause the slashes to be the wrong
direction on Windows and result in the files not being found.


tmessi (approved):

jefferai:
Note: tests broken from a previous PR, which will be fixed in a forthcoming PR.

PR #1736, opened 2021-11-19
internal/host/plugin: ensure UpdateCatalog triggers updates for sets

vancluever:
This ensures that attribute updates to a host catalog triggers an update
to any of its downstream host sets. This ensures that updates to
elements that affect a host set, such as the region a host catalog is
configured for, take effect immediately instead of requiring a wait of
long as 10m after an update.


vancluever:
Currently in draft - need to fix some stuff surrounding the scheduler with regards to ensuring that a job was queued right away _only_ when we actually did update sets for sync.

internal/host/plugin/repository_host_catalog.go
```diff
@@ -454,6 +458,36 @@ func (r *Repository) UpdateCatalog(ctx context.Context, c *HostCatalog, version
...
+					return errors.Wrap(ctx, err, op, errors.WithMsg("unable to get sets for host catalog"))
+				}
+
+				for _, set := range setsForCatalog {
+					newSet := set.clone()
+					newSet.NeedSync = true
+					currentVersion := newSet.Version
+					newSet.Version = currentVersion + 1
+					var msg oplog.Message
+					n, err := w.Update(ctx, newSet, []string{"NeedSync", "version"}, []string{}, db.NewOplogMsg(&msg), db.WithVersion(&currentVersion))
```
talanknight, line 475: If we are actually updating the set we shouldn't need to include the version as part of the mask.  In fact we shouldn't since if the set is already marked as NeedSync = true we wouldn't need to increment the version.  This is handled by the `update_version_column` db trigger.

internal/host/plugin/repository_host_catalog.go
```diff
@@ -454,6 +458,36 @@ func (r *Repository) UpdateCatalog(ctx context.Context, c *HostCatalog, version
...
+					if err != nil {
+						return errors.Wrap(ctx, err, op, errors.WithMsg("unable to update host set"))
+					}
+
+					if n != 1 {
```
talanknight, line 480: Some of the sets might already have NeedSync = true which would cause it to not be updated I don't think.  It might be better for this conditional to be updated.
```suggestion
					if n > 1 {
```

internal/host/plugin/repository_host_catalog_test.go
```diff
@@ -1441,6 +1443,162 @@ func TestRepository_DeleteCatalogX(t *testing.T) {
...
+	require.Equal(t, set1.Version+1, gotSet1.Version)
+
+	gotSet2, _, err = repo.LookupSet(ctx, set2.PublicId)
+	require.NoError(t, err)
+	require.True(t, gotSet2.NeedSync)
+	require.Equal(t, set2.Version+1, gotSet2.Version)
+
+	// Wait for the job again, for the same amount of time. This time the job should have ran.
+	select {
+	case <-time.After(time.Minute * 2):
```
talanknight, line 1594: Having these potentially super long running tests aren't great.  The scheduler is tested separately so if it gets called with the expected input it should function properly.  Is there a way we can mock the scheduler or fake it in a way that we just have to confirm that it is called iunstead of waiting for up to 2 minute?

Beyond making sure the scheduler is called with scheduling the next job run soon we can try running the job using `TestRunSetSync` and see the sets are picked up.

vancluever: I don't think mocking it is an option, unfortunately, unless we change the repo over to take a full-on interface, which seems excessive for this purpose.

Let me see if I can whip something up so that we can check the next run time and assert that it's been set to 0 or not.

vancluever: Updated!

internal/host/plugin/repository_host_catalog.go
vancluever: Updated!

internal/host/plugin/repository_host_catalog_test.go
```diff
@@ -1441,6 +1444,136 @@ func TestRepository_DeleteCatalogX(t *testing.T) {
...
+
+	numSetsUpdated, err = dbRW.Update(ctx, set2, []string{"LastSyncTime", "NeedSync"}, []string{})
+	require.NoError(t, err)
+	require.Equal(t, 1, numSetsUpdated)
+
+	j := &testSyncJob{}
+	err = sched.RegisterJob(ctx, j, scheduler.WithNextRunIn(setSyncJobRunInterval))
+	require.NoError(t, err)
+	var wg sync.WaitGroup
+	err = sched.Start(ctx, &wg)
```
talanknight, line 1505: With the new way of validating the job's next run time was scheduled, you don't need to start the scheduler.

internal/host/plugin/repository_host_catalog.go
```diff
@@ -454,6 +458,34 @@ func (r *Repository) UpdateCatalog(ctx context.Context, c *HostCatalog, version
...
+					if err != nil {
+						return errors.Wrap(ctx, err, op, errors.WithMsg("unable to update host set"))
+					}
+
+					if n > 1 {
+						return errors.New(ctx, errors.MultipleRecords, op, fmt.Sprintf("expected no more than 1 host set to be updated while flagging host set id %q for synchronization, got %d", newSet.PublicId, n))
+					}
+
+					msgs = append(msgs, &msg)
+					if len(setsForCatalog) != 0 {
```
talanknight, line 483: Either remove the if statement (since inside this loop the `setsForCatalog` will always have a length > 0).  Or move it to after the for loop.  Maybe you meant to only set this if n ==  1 ?

talanknight (changes requested):

internal/host/plugin/repository_host_catalog.go
vancluever: Fixed!

internal/host/plugin/repository_host_catalog_test.go
vancluever: Removed.

talanknight (dismissed):

talanknight (approved):

PR #1741, opened 2021-11-24
refact(servers): Cleanup redundant if else

louisruch:


louisruch:
@vancluever @jimlambrt Found this strange if-else statement. Seems when @jimlambrt added events he removed the logger.Log in the if leaving in the current state.

vancluever (approved):

PR #1743, opened 2021-11-25
Fix setting plugin host catalog attributes to nil

jefferai:
Internally attributes are not allowed to be the null value. However, as
currently implemented, this then breaks our normal model of sending an
explicit `null` in the JSON to reset/clear a field. This is a problem
because the CLI doesn't have a way to say "set this but to empty"
without a lot of hoops, and causes some issues for API users and TF as
well.

This changes PatchBytes and PatchStruct to return an empty value (or
marshaled empty value in the case of PatchBytes) if the source is null
or empty. In the host catalog update function we then always set this
value if it's in the db mask (removing the nil check). Together this
means a null value ends up coming out as an empty value (and thus valid
for the DB) from PatchStruct/PatchBytes, and our API translation
function already returns nothing for attributes if the length of the
fields are zero.

### DIFF
--- a/internal/host/plugin/repository_host_catalog.go
+++ b/internal/host/plugin/repository_host_catalog.go
@@ -286,7 +286,7 @@ func (r *Repository) UpdateCatalog(ctx context.Context, c *HostCatalog, version 
 		}
 	}
 
-	if updateAttributes && c.Attributes != nil {
+	if updateAttributes {
 		dbMask = append(dbMask, "attributes")
 		newCatalog.Attributes, err = patchstruct.PatchBytes(newCatalog.Attributes, c.Attributes)
 		if err != nil {
--- a/internal/libs/patchstruct/patchstruct.go
+++ b/internal/libs/patchstruct/patchstruct.go
@@ -7,15 +7,16 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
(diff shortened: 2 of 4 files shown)


vancluever (commented):
Looks generally sound, but are we sure that we've surfaced up the no-up behavior that has not been removed from `PatchBytes`? Or will it achieve the same effect?

jefferai:
Not quite following, what do you mean the no-up behavior?

vancluever:
Yep. Particularly the test case [here](https://github.com/hashicorp/boundary/pull/1743/files#diff-b74ac45d49e86538756195df483275fb05b42e75c9c160621d018acb90ac1f33R121-R122) seems to indicate that we're changing behavior so that patching no longer skips but rather removes. How are we accounting for that in anything that is affected by this change?

jefferai:
The point of this change is to make that the behavior. I'm not following what you mean by how we are accounting for it. There is a changed test that ensures that that is the behavior of the function.

vancluever:
Hmm, I think I was thinking about how we would skip updating attributes, but I'm guessing that is appropriately handled but ensuring that service handlers are not passing the mask in. As long as that's the case, this seems okay.

vancluever (approved):

jefferai:
Gotcha. Yes, if nothing is passed in the mask is not populated and thus nothing is cleared.

PR #1679, opened 2021-11-08
feature (workers): Add the user's client IP to a session's connection information

jimlambrt:
- feature: Add client ip to inbound request information
- feature (workers): Add user client IP to connection info
- feature (db): Add user_client_ip to session_connection
- feature (workers): add user client ip to ConnectConnection API
- feature (sessions): Add support for user client ip in connection info
- feature (workers): Add support for user client ip in tcp protocol

This PR is dependent on #1678 and cannot be merged until it's approved and merged.

PR #1749, opened 2021-11-30
feature (workers): Add the user's client IP to a session's connection information

jimlambrt:
- feature: Add client ip to inbound request information
- feature (workers): Add user client IP to connection info
- feature (db): Add user_client_ip to session_connection
- feature (workers): add user client ip to ConnectConnection API
- feature (sessions): Add support for user client ip in connection info
- feature (workers): Add support for user client ip in tcp protocol

This PR is a co-dependent of on #1678 and should be merged into the `jimlambrt-audit-client-ip` branch before #1678 is merged to main.

This PR replaces #1679


jimlambrt:
@mgaffney I've incorporated your previous suggestion in this PR to add a `not null` constraint on the `session_connection.user_client_ip` and set a default value of `::` (the [IPv6 Unspecified Address](https://en.wikipedia.org/wiki/IPv6_address#Unspecified_address)).

internal/session/connection_test.go
```diff
@@ -42,13 +43,15 @@ func TestConnection_Create(t *testing.T) {
 				clientTcpPort:      22,
 				endpointTcpAddress: "127.0.0.1",
 				endpointTcpPort:    2222,
+				userClientIp:       "127.0.0.1",
```
louisruch, line 46: nit: Might be good to test a different address to ensure what is coming back comes from this field.

louisruch (approved):
One minor comment about a test otherwise this LGTM

internal/session/connection_test.go
jimlambrt: ty!

internal/db/schema/migrations/oss/postgres/21/01_session.up.sql
```diff
@@ -0,0 +1,6 @@
+begin;
+
+alter table session_connection
+  add column user_client_ip inet not null default '::';
```
jefferai, line 4: Why `::`?

internal/proto/local/controller/servers/services/v1/session_service.proto
```diff
@@ -92,6 +92,10 @@ message ConnectConnectionRequest {
   string endpoint_tcp_address = 40;  // @gotags: `class:"public"`
   uint32 endpoint_tcp_port = 50;     // @gotags: `class:"public"`
   string type = 60;                  // @gotags: `class:"public"`
+
+  // user_client_ip is the user's client ip for the connection as determined
+  // the inbound http request
```
jefferai, line 97:
```suggestion
  // user_client_ip is the user's client ip for the connection as determined by
  // the inbound http request handler
```

internal/servers/worker/proxy/proxy.go
```diff
@@ -15,6 +15,11 @@ import (
 // a provided ClientConn and the RemoteEndpoint, as well as the parameters to update
 // the connection in the connection repository.
 type Config struct {
+	// UserClientIp is the user's client IP
+	UserClientIp string
```
jefferai, line 19: Should this be a `net.IP` at this point, given client address is `net.TCPAddr`?

internal/servers/worker/listeners.go
```diff
@@ -41,9 +41,12 @@ func (w *Worker) startListeners() error {
 				return fmt.Errorf("unknown listener purpose %q", purpose)
 			}
 
-			handler := w.handler(HandlerProperties{
+			handler, err := w.handler(HandlerProperties{
 				ListenerConfig: ln.Config,
 			})
+			if err != nil {
```
jefferai, line 47: ty!

internal/session/session_connect_with.go
```diff
@@ -31,5 +32,8 @@ func (c ConnectWith) validate() error {
 	if c.EndpointTcpPort == 0 {
 		return errors.NewDeprecated(errors.InvalidParameter, op, "missing endpoint ctp port")
 	}
+	if c.UserClientIp == "" {
```
jefferai, line 35: This is one of the places that made me wonder about the consequences of the default of `::`. If there's no IP here, we shouldn't have a session connection right?

My guess is that the default is to handle the upgrade case. But I wonder if a better upgrade case is a non-null empty string as a less surprising possibility?

internal/db/schema/migrations/oss/postgres/21/01_session.up.sql
jimlambrt: It's an ipv6 specified address: https://en.wikipedia.org/wiki/IPv6_address#Unspecified_address

internal/servers/worker/proxy/proxy.go
jimlambrt: I'm not sure []byte is significanlty better than string.  ClientAddress makes sense a net.TCPAddr since there's a port with the IP.

internal/session/session_connect_with.go
jimlambrt: `::` means something contextually (unspecified) and an empty string could be interpreted that the zero value was sent to the database.  I'd prefer to keep the `::` default.

internal/db/schema/migrations/oss/postgres/21/01_session.up.sql
jefferai: I know what it is, I'm asking why you chose it :-)

internal/servers/worker/proxy/proxy.go
jefferai: Using a `net.IP` guarantees that it is a valid address (if created via correct means). We can always convert later if we want to do checks, but by that argument, shouldn't we use a string for ClientAddress since it's just `ip:port`.

To me it just feels weird to have a typed value and a parsed value next to each other when they represent very similar things at the same point in the code path.

internal/session/session_connect_with.go
jefferai: Why not `0.0.0.0` though? After all, it's used to specify an invalid or unknown address.

We don't know if it was IPv4 or IPv6 -- and it's probably the former -- and while we could just use that and say "if we see this, we didn't know the address at all, regardless of protocol", it technically means that we didn't know _the IPv6_ address.

That's why I had suggested empty instead, because to me it better conveys that we don't know _any_ address, IPv4 or IPv6.

internal/db/schema/migrations/oss/postgres/21/01_session.up.sql
jimlambrt: because it's the only "official spec" way to represent an unspecified IP address.

internal/session/session_connect_with.go
jimlambrt: `0.0.0.0` is part of any spec that I'm aware of to represent an unspecified IP address.   Empty string is an anti-pattern for unknown value. With that said, we also aren't big fans of null fields in Boundary for required fields, for very good reasons.   Unfortunately, it this case we're going to need to compromise given that we have existing rows.  @mgaffney suggested `::` as a compromise and I'm in agreement that it's better than allowing nulls and better than using an empty string which can lead to all sorts of odd errors since empty string is a zero value.

internal/servers/worker/proxy/proxy.go
jimlambrt: Add the requested changes in 241d222d1
Looking at the changes I'm not sure it adds any value, but happy to resolve this.

jimlambrt: BTW, I can admit that it looks better.

internal/session/session_connect_with.go
jefferai: 0.0.0.0 is incidentally specified in the DOCSIS standard (https://datatracker.ietf.org/doc/html/rfc4639) to mean unknown address. But incidental.

On the other hand, explicitly making it a v6 spec value means you need to be careful if ever doing comparisons. How do you decide that's an unknown value in the context of IPv4? You're changing up the lack of a standard in v4 with a standard from a different domain.

My main issue is actually not internally to Boundary -- it's surfacing this to users. If most users are only using IPv4 and they are returned an IPv6 address in place of an unknown address, this will be confusing and unexpected. And my assumption is that they will, because it'll surface when doing a read of a session, and it will surface in the warehouse. If this info goes to the warehouse I think it is very much not expected that I as a user need to formulate queries with an IPv6 address when my entire infra is IPv4.

jimlambrt: Ty for the background about your concerns.  I think, I understand them now which is very helpful.  

I guess, we have a couple of choices to move forward:
- Allow nulls with no default value
- Don't allow nulls and set a default to an address that represents unspecified like `::`, `0.0.0.0` or an empty string

Based on @jefferai concerns, it feels like `::` and `0.0.0.0` may not be ideal and could confuse users.  

At this point, I'd vote for (in order of preference): `null` or an empty string.

@mgaffney ???

jefferai: I know `null` is meant as "value unspecified". What's the difference from a DB modeling perspective between that and "value unknown"?

At an API level null in the DB vs empty string doesn't really matter. But I think there might be some effects in terms of the warehouse if it's `null`...?

mgaffney: I vote we use `0.0.0.0` as the default value and not `null`.

jimlambrt:
@jefferai : do you mind reviewing/approving/merging?

internal/session/session_connect_with.go
jefferai: Just circling back for history -- we realized it must be able to be null because we don't know the value at the time the session is created, only after the worker has filled it in.

jefferai (approved):

PR #1753, opened 2021-12-02
ci: Run sdk and api tests in ci

tmessi:
This adds new makefile targets and configuration to run the tests in the 
api and sdk modules. These do not run by default since `go test ./...` does
not cross module boundaries.

### DIFF
--- a/.circleci/config/jobs/test-api.yml
+++ b/.circleci/config/jobs/test-api.yml
@@ -1,0 +1,9 @@
+executor: go-machine-medium
+steps:
+- checkout
+- install-go
+- run:
+    name: "Run API Tests"
+    no_output_timeout: 15m
+    command: |
+      make test-api
--- a/.circleci/config/jobs/test-sdk.yml
+++ b/.circleci/config/jobs/test-sdk.yml
@@ -1,0 +1,9 @@
+executor: go-machine-medium
+steps:
+- checkout
+- install-go
+- run:
+    name: "Run SDK Tests"
+    no_output_timeout: 15m
+    command: |
+      make test-sdk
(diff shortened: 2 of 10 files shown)


.circleci/config/jobs/test-api.yml
```diff
@@ -0,0 +1,13 @@
...
+- checkout
+- install-go
+- run:
+    name: "Run API Tests"
+    no_output_timeout: 15m
+    command: |
+      make test-api
+- slack/status:
+    fail_only: true
+    only_for_branches: master
```
jimlambrt, line 12: main branch?

.circleci/config/jobs/test-sdk.yml
```diff
@@ -0,0 +1,13 @@
...
+- checkout
+- install-go
+- run:
+    name: "Run SDK Tests"
+    no_output_timeout: 15m
+    command: |
+      make test-sdk
+- slack/status:
+    fail_only: true
+    only_for_branches: master
```
jimlambrt, line 12: main branch?

jimlambrt (changes requested):

jimlambrt:
Ty for this PR! I had forgotten about this.

.circleci/config/jobs/test-api.yml
tmessi: good catch, and actually its set to `master` for the [current ci job](https://github.com/hashicorp/boundary/blob/739222d06e3a812c44eb64494ad21cd39d54dcf6/.circleci/config/jobs/build.yml#L21-L24) which probably means it isn't working at all...I might just remove these slack status blocks for now and look into that some other time.

jimlambrt: I'm ready for  the proposed migration to GitHub actions.

jimlambrt (dismissed):
ty!

PR #1756, opened 2021-12-02
refact(controller/targets): Move tcp functionality into subpackage

tmessi:
This follows a similar pattern as the target domain package, providing a 
subpackage for the tcp target functionality, allowing this to be pluggable.


louisruch (approved):
LGTM!

PR #1757, opened 2021-12-03
feat(credentials): add support for vault KV-v2

louisruch:


louisruch:
@mgaffney @tmessi this enables getting credentials from Vault KV 2 which has an embedded `data map[string]interface{}` in the secret Data response. Happy to discuss alternatives.

internal/credential/vault/internal/userpassword/userpassword.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+	}
+
+	return
+}
+
+// kv2UserPass looks for the the usernameAttr and passwordAttr in the embedded
+// 'data' field within the data map.
+//
+// Additionaly it validates the data is in the expected KV-v2 format:
```
tmessi, line 47: I would phrase this so it is clear that the function uses this structure to determine that it is a kv2 secret and if this structure is not found, it returns.

internal/credential/vault/internal/userpassword/userpassword.go
```diff
@@ -0,0 +1,90 @@
+package userpassword
+
+type (
+	data         map[string]interface{}
+	userPassFunc func(sd data, usernameAttr, passwordAttr string) (string, string)
```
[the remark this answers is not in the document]
tmessi: Also, I would add a doc here describing this function and its expected behavior

internal/credential/vault/internal/userpassword/userpassword.go
```diff
@@ -0,0 +1,90 @@
...
+			username = u
+		}
+	}
+	if p, ok := data[passwordAttr]; ok {
+		if p, ok := p.(string); ok {
+			password = p
+		}
+	}
+
+	return
```
tmessi, line 89: personally not a fan of naked returns, i think it is more readable and easier to verify if the returns are explicit

tmessi (approved):

PR #1758, opened 2021-12-03
feat(config): Add env and file support to Worker Tags

hugoamvieira:
We do this by taking a look at the HCL before the actual tag parsing and resolving the string to whatever value exists underneath, and then pass it on to the actual parsing.

This ensures that previous functionality is not affected.

### DIFF
--- a/internal/cmd/config/config.go
+++ b/internal/cmd/config/config.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"crypto/rand"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -157,9 +158,10 @@ type Worker struct {
 
 	// We use a raw interface for parsing so that people can use JSON-like
 	// syntax that maps directly to the filter input or possibly more familiar
-	// key=value syntax. This is trued up in the Parse function below.
-	TagsRaw interface{}         `hcl:"tags"`
+	// key=value syntax, as well as accepting a string denoting an env or file
+	// pointer. This is trued up in the Parse function below.
 	Tags    map[string][]string `hcl:"-"`
+	TagsRaw interface{}         `hcl:"tags"`
 
 	// StatusGracePeriod represents the period of time (as a duration) that the
 	// worker will wait before disconnecting connections if it cannot make a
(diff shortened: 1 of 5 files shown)


internal/cmd/config/config.go
```diff
@@ -337,7 +337,29 @@ func Parse(d string) (*Config, error) {
...
+			// We then lookup whatever value is within and put it back
+			// into TagsRaw to get parsed.
+			case string:
+				rawTags, err := parseutil.ParsePath(t)
+				if err != nil && !errors.Is(err, parseutil.ErrNotAUrl) {
+					return nil, fmt.Errorf("Error parsing worker tags: %w", err)
+				}
+				result.Worker.TagsRaw = nil
+
+				err = hcl.Decode(result, rawTags)
```
tmessi, line 354: I wondering if this should decode into a new `*Config` instead of decoding again into `result`. If something other then tags is in the env/file referenced in the result of `parseutil.ParsePath(t)` won't this replace it? For example what happens if you add a test like:

```go
{
    name: "additional configuration in env",
    in: `
    worker {
        name = "dev-worker"
        tags = "env://BOUNDARY_WORKER_TAGS"
    }`,
    actualTags: `
    worker {
        name = "prod-worker"
        tags {
            type = ["dev", "local"]
        }
    }`,
    expWorkerTags: map[string][]string{
        "type": {"dev", "local"},
    },
    expErr: false,
},
```
And then add an assertion around `c.Worker.Name`?

tmessi (changes requested):

internal/cmd/config/config.go
hugoamvieira: I knew it'd be replaced but I wasn't sure how to fix that problem - Can't believe I didn't see the way to do it right in front of me 😮‍💨 
This is a great suggestion, thank you!

hugoamvieira: Fixed in dcb0944

hugoamvieira: New and improved syntax 5675753

hugoamvieira:
e387b1b adds support for setting the tag keys in the HCL config, but then getting the values from env or file, so we support both block-level abstraction as well as per key abstraction.

internal/cmd/config/config.go
```diff
@@ -337,12 +339,58 @@ func Parse(d string) (*Config, error) {
...
+					for k, v := range m {
+						// We allow the user to pass in only the keys in HCL, and
+						// then set the values to point to an env var/file.
+						valStr, ok := v.(string)
+						if !ok {
+							continue
+						}
+
+						parsed, err := parseutil.ParsePath(valStr)
+						if err != nil {
```
tmessi, line 378: Should this be checking:

```go
if err != nil && !errors.Is(err, parseutil.ErrNotAUrl) 
```

hugoamvieira: I removed this due to concerns of getting bad state below but on second thought I think you're right. There's also another check below to prevent bad payloads from getting into the json bit

hugoamvieira: Resolved in 1c2ae70

hugoamvieira:
^ These commits mostly add some more tests for completeness as well as docs changes to reflect the new functionality.

internal/cmd/config/config.go
tmessi: Can you add a test case for this? So something that would have failed with the code prior to 1c2ae70 where `parseutil.ParsePath` returns `parseutil.ErrNotAUrl`?

hugoamvieira: Yep - The problem with that is that the checks behind that error are quite easy to bypass due to the nature of URLs, but probably worth still adding a test case for that

internal/cmd/config/config.go
```diff
@@ -337,12 +339,58 @@ func Parse(d string) (*Config, error) {
 		if !strutil.Printable(result.Worker.Name) {
 			return nil, errors.New("Worker name contains non-printable characters")
 		}
+
 		if result.Worker.TagsRaw != nil {
 			switch t := result.Worker.TagsRaw.(type) {
+			// We allow `tags` to be a simple string containing a reference to
+			// an environment variable (env://) / file path (file://).
```
tmessi, line 346: parseutil currently supports `env://` and `file://`, but it is possible that will be expanded, at which point this comment would get stale, perhaps rephrase this to something like:

```go
// We allow `tags` to be a simple string containing a URL with schema.
// See: https://github.com/hashicorp/go-secure-stdlib/blob/main/parseutil/parsepath.go
```

hugoamvieira: Good point, thank you!

internal/cmd/config/config.go
tmessi: Yea, its a strange case, but I think having a test to cover it will help to understand the expected behavior. If it helps here is how `parseutil` unit tests that case: https://github.com/hashicorp/go-secure-stdlib/blob/35ba60c96fcdd1ab809281de44df2240011ec71b/parseutil/parsepath_test.go#L53-L58

hugoamvieira: Yep, that's exactly what I used!

internal/cmd/config/config.go
hugoamvieira: Addressed in 9e383a9

internal/cmd/config/config.go
hugoamvieira: Addressed in 9e383a9

tmessi (dismissed):

tmessi (approved):

PR #1760, opened 2021-12-04
Add support for credential mapper overrides to the domain for update and lookup

mgaffney:
An alternative implementation is in #1761. This version treats a credential mapping override as a value object which only allows the mapping override to be updated as a whole, not the individual fields of a credential mapping override.

### DIFF
--- a/internal/credential/vault/repository_credential_library.go
+++ b/internal/credential/vault/repository_credential_library.go
@@ -114,8 +114,8 @@ func (r *Repository) CreateCredentialLibrary(ctx context.Context, scopeId string
 // number of records updated. l is not changed.
 //
 // l must contain a valid PublicId. Only Name, Description, VaultPath,
-// HttpMethod, and HttpRequestBody can be updated. If l.Name is set to a
-// non-empty string, it must be unique within l.StoreId.
+// HttpMethod, HttpRequestBody, and MappingOverride can be updated. If
+// l.Name is set to a non-empty string, it must be unique within l.StoreId.
 //
 // An attribute of l will be set to NULL in the database if the attribute
 // in l is the zero value and it is included in fieldMaskPaths except for
@@ -141,6 +141,7 @@ func (r *Repository) UpdateCredentialLibrary(ctx context.Context, scopeId string
 	}
 	l = l.clone()
 
+	var updateMappingOverride bool
 	for _, f := range fieldMaskPaths {
 		switch {
 		case strings.EqualFold(nameField, f):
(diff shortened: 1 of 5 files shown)


internal/credential/vault/repository_credential_library_test.go
```diff
@@ -897,6 +1162,24 @@ func TestRepository_UpdateCredentialLibrary(t *testing.T) {
 			if tt.wantCount > 0 {
 				assert.NoError(db.TestVerifyOplog(t, rw, got.GetPublicId(), db.WithOperation(oplog.OpType_OP_TYPE_UPDATE), db.WithCreateNotBefore(10*time.Second)))
 			}
+
+			if tt.orig.MappingOverride != nil {
```
louisruch, line 1166: This check invalidates the `no-mapping-override-change-username-and-password-attributes` test because it has a `orig.MappingOverride` of `nil`, but it does expect to have a mapping override after the update that should be checked.

Verifying this locally revealed an issue with this test returning early and never testing any of the override mappings, on line 1154 and line 1159 we return after doing a `dbassert.IsNull`, these should probably be updated to 
```
if tt.want.Name == "" {
	dbassert.IsNull(got, "name")
} else {
	assert.Equal(tt.want.Name, got.Name)
}
```

louisruch (changes requested):
Found one test issue, other than that we should probably sync up on the behavior of the update. When working on the CLI/API for the `create` we decided to go with a `structpb` over a `map[string]string` because if gave us field specific masks for the overrides (`"credential_mapping_overrides.username"`)

This allows the user to do an `update` to the password mapping without having to reapply the username mapping if it didn't change.  With these changes the API service will need to do a `lookup` before calling `update` in order to correctly apply the update, since this adds a lookup as well that means our update path is now `lookup`, `lookup`, `update`, `lookup`

As a side note if the user uses the `withAutomaticVersioning` the SDK does another `lookup` to get the version :/ before hitting the service

internal/credential/vault/repository_credential_library_test.go
mgaffney: Great catch! Thanks for reviewing the tests so thoroughly! I made changes to address the issues you found. Can you take another look?

internal/credential/vault/repository_credential_library.go
```diff
@@ -214,17 +331,68 @@ func (r *Repository) LookupCredentialLibrary(ctx context.Context, publicId strin
...
 		if errors.IsNotFoundError(err) {
 			return nil, nil
 		}
 		return nil, errors.Wrap(ctx, err, op, errors.WithMsg(fmt.Sprintf("failed for: %s", publicId)))
 	}
-	return l, nil
+	return l.toCredentialLibrary(), nil
+}
+
+type publicLibrary struct {
```
tmessi, line 345: can you add docs for this? There is a nice comment on corresponding sql view, but some docs here would be nice as well.

mgaffney: done

tmessi (approved):

louisruch (approved):
lgtm

PR #1761, opened 2021-12-06
(alt 2) Add support for credential mapper overrides to the domain for update and lookup

mgaffney:
An alternative implementation of #1760. This version allows individual fields of a credential mapping override to be updated.

### DIFF
--- a/internal/credential/vault/repository_credential_library.go
+++ b/internal/credential/vault/repository_credential_library.go
@@ -114,14 +114,15 @@ func (r *Repository) CreateCredentialLibrary(ctx context.Context, scopeId string
 // number of records updated. l is not changed.
 //
 // l must contain a valid PublicId. Only Name, Description, VaultPath,
-// HttpMethod, and HttpRequestBody can be updated. If l.Name is set to a
-// non-empty string, it must be unique within l.StoreId.
+// HttpMethod, HttpRequestBody, UsernameAttribute, and PasswordAttribute
+// can be updated. If l.Name is set to a non-empty string, it must be
+// unique within l.StoreId.
 //
 // An attribute of l will be set to NULL in the database if the attribute
 // in l is the zero value and it is included in fieldMaskPaths except for
-// HttpMethod.  If HttpMethod is in the fieldMaskPath but l.HttpMethod
-// is not set it will be set to the value "GET".  If storage has a value
-// for HttpRequestBody when l.HttpMethod is set to GET the update will fail.
+// HttpMethod. If HttpMethod is in the fieldMaskPath but l.HttpMethod is
+// not set it will be set to the value "GET". If storage has a value for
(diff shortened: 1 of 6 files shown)


internal/credential/vault/repository_credential_library.go
```diff
@@ -207,24 +320,221 @@ func (r *Repository) UpdateCredentialLibrary(ctx context.Context, scopeId string
...
 		if errors.IsNotFoundError(err) {
 			return nil, nil
 		}
 		return nil, errors.Wrap(ctx, err, op, errors.WithMsg(fmt.Sprintf("failed for: %s", publicId)))
 	}
-	return l, nil
+	return l.toCredentialLibrary(), nil
+}
+
+type publicLibrary struct {
```
tmessi, line 487: can you add docs for this? There is a nice comment on corresponding sql view, but some docs here would be nice as well.

internal/credential/vault/repository_credential_library.go
```diff
@@ -207,24 +320,221 @@ func (r *Repository) UpdateCredentialLibrary(ctx context.Context, scopeId string
 	return returnedCredentialLibrary, rowsUpdated, nil
 }
 
+type changeFunc func(ctx context.Context, w db.Writer) (*oplog.Message, error)
+
+func calcMappingOverrideChange(op errors.Op, ol, nl *CredentialLibrary, dbMask []string) (changeFunc, error) {
```
tmessi, line 325: This function is very dense and difficult (at least for me) to grok. I think a few things could be done to help:

- add some function docs and some inline comments
- potentially break out the `insertFn`, `deleteFn` `updateFn` functions so they can be documented independently
- use some more descriptive variable names, there are a lot of short but similar variable names in here, `om`, `mo`, `nm`, `nmp`, `dmp`

internal/credential/vault/repository_credential_library.go
```diff
@@ -183,16 +237,75 @@ func (r *Repository) UpdateCredentialLibrary(ctx context.Context, scopeId string
...
+			if updateMappingOverride {
+				chg, err := calcMappingOverrideChange(op, origLib, l, fieldMaskPaths)
+				if err != nil {
+					return errors.Wrap(ctx, err, op, errors.WithMsg("unable to calculate change to mapping override"))
+				}
+				msg, err := chg(ctx, w)
+				if err != nil {
+					return errors.Wrap(ctx, err, op, errors.WithMsg("unable to change mapping override"))
+				}
+				msgs = append(msgs, msg)
```
louisruch, line 295: Should probably add a nil check here, since if the `noChange` function is returned from `calcMappingOverrideChange` this will append `nil` to `msgs`

mgaffney:
We have decide to go with the approach in #1760.

PR #1775, opened 2021-12-08
Create the network address dimension

talanknight:
This also migrates existing addresses into the new dimension.

Left TODO:
- [ ] Populate the wh_network_address_group and wh_network_address_group_membership tables
- [ ] drop the `host_address` column and add a reference to the wh_network_address_group in the wh_host_dimension table
- [ ] Populate the newly created reference to the wh_network_address_group

### DIFF
--- a/internal/db/schema/migrations/oss/postgres/22/02_wh_network_address_dimensions.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/22/02_wh_network_address_dimensions.up.sql
@@ -37,6 +37,14 @@ begin;
     ('Unsupported', 'Unknown', 'Not Applicable', 'Not Applicable', 'None', 'None', 'None'),
     ('Unknown', 'Unknown', 'Not Applicable', 'Not Applicable', 'None', 'None', 'None');
 
+  -- prepare a group which can be referenced by the newly created
+  -- wh_host_dimension column.
+  insert into wh_network_address_group(key)
+  values('Unknown'), ('Unsupported');
+  insert into wh_network_address_group_membership(network_address_group_key, network_address)
+  values('Unknown', 'Unknown'), ('Unsupported', 'Unsupported');
+
+
   -- Migrate all the ip addresses and ignore any addresses which aren't ip.
   with
   ip_addresses(address, inet_address) as (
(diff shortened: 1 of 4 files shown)


internal/db/schema/migrations/oss/postgres/22/02_wh_network_address_dimensions.up.sql
```diff
@@ -0,0 +1,88 @@
+begin;
+
+  -- wh_try_cast_inet returns either the provided text cast into inet or a null.
+  create function wh_try_cast_inet(text)
```
tmessi, line 4: can you add some [sqltests](https://github.com/hashicorp/boundary/blob/main/internal/db/sqltest/README.md) for this function?

internal/db/schema/migrations/oss/postgres/22/02_wh_network_address_dimensions.up.sql
```diff
@@ -0,0 +1,88 @@
...
+  exception when others then
+    return null::inet;
+  end;
+  $$ language plpgsql;
+
+  -- wh_private_address_status returns a warehouse appropriate string
+  -- representing if the address is private, public, or not applicable for the
+  -- provided address.  An address which cannot be cast to an inet results in
+  -- 'Not Applicable' being returned.
+  create function wh_private_address_status(inet) returns text
```
tmessi, line 18: can you add some [sqltests](https://github.com/hashicorp/boundary/blob/main/internal/db/sqltest/README.md) for this function?

internal/db/schema/migrations/oss/postgres/22/02_wh_network_address_dimensions.up.sql
```diff
@@ -0,0 +1,88 @@
...
+  select
+    address,
+    'IP Address',
+    case
+      when family(inet_address) = 4 then 'IPv4'
+      when family(inet_address) = 6 then 'IPv6'
+      else 'Not Applicable'
+    end,
+    wh_private_address_status(inet_address),
+    'None',
```
tmessi, line 60: Should `dns_name` be `None` here? I would think it should be `Not Applicable`

tmessi (changes requested):

talanknight:
FYI, I am trying to break this WH work up into smaller PRs so we can have more targets conversations but I decided after creating this PR that where I stopped was probably not a good stopping point.  I added checklists to the description of this PR and plan on adding those before I send it out for review again.

Sorry for the noise, and I'm also happy to discuss what's already here if that would be valuable.

talanknight:
This PR overly focused on the migration of old data to new tables aspect.  I'm going to close this PR and change my approach to consider the capturing of live data first and then, only after all that is resolved, work out the migration of old data.

PR #1776, opened 2021-12-08
feat(config): Add env and file support for database max connections

irenarindos:


internal/cmd/config/config.go
```diff
@@ -332,6 +334,24 @@ func Parse(d string) (*Config, error) {
 			}
 			result.Controller.AuthTokenTimeToStaleDuration = t
 		}
+
+		if result.Controller.Database != nil {
+			if result.Controller.Database.MaxOpenConnectionsRaw != nil {
+				switch t := result.Controller.Database.MaxOpenConnectionsRaw.(type) {
```
hugoamvieira, line 340: I think this switch should have a `default` case to error on any other type.

Then, you could have a test case testing what happens when you set `max_open_connections` to, say, ["bad"] - It should fall under that default case

internal/cmd/config/config.go
```diff
@@ -349,6 +349,8 @@ func Parse(d string) (*Config, error) {
 					}
 				case int:
 					result.Controller.Database.MaxOpenConnections = t
+				default:
+					return nil, fmt.Errorf("Database max open connections value is not an int: %v", t)
```
hugoamvieira, line 359: IIRC this message goes all the way back to the user in their CLI. In that context, I think this error might be a bit misleading since we also support strings (i.e `"8"` is a valid value).

Maybe more like "Database max open connections: unsupported type %q", with `%q` being the value of   `reflect.TypeOf(t).String()`, which is the actual type. WDYT?

hugoamvieira (commented):
Looks good! I have one last thought:

internal/cmd/config/config.go
irenarindos: Sounds good to me!

hugoamvieira (approved):

PR #1784, opened 2021-12-10
Update AWS host provider

vancluever:
This updates the AWS host provider to include https://github.com/hashicorp/boundary-plugin-host-aws/pull/7.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,8 +25,8 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
   ([PR](https://github.com/hashicorp/boundary/pull/1736))
 * controllers/workers: Add client IP to inbound request information which is included in
   Boundary events ([PR](https://github.com/hashicorp/boundary/pull/1678))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
   ([PR](https://github.com/hashicorp/boundary/pull/1736))
 * controllers/workers: Add client IP to inbound request information which is included in
   Boundary events ([PR](https://github.com/hashicorp/boundary/pull/1678))
+* plugins/aws: AWS host sets will now return DNS hosts in addition to IP
```
talanknight, line 28: I think the terminology here could be confusing for someone not super familiar with boundary (boundary doesn't have the concept of a "dns host", host sets don't return dns values, etc...).  How about something like "AWS plugin based hosts now include dns names in addition to the ip addresses they already provide."

vancluever: Applied suggestion!

talanknight (approved):

PR #1785, opened 2021-12-11
refactor: Update internal/db and oplog to use go-dbw package for database operations.

jimlambrt:
This PR refactors Boundary to use https://github.com/hashicorp/go-dbw for database operations.   

The internal/db pkg remains primarily to convert between boundary and dbw types and errors.


jimlambrt:
I will be adding some additional oplog tests next week, but I think everything else is in reasonable shape and ready to begin review.

internal/db/db.go
```diff
@@ -86,6 +79,7 @@ func (d *DB) Close(ctx context.Context) error {
 // answer is no, but there are occasions when it's necessary.  See the sql.DB
 // docs for more information.
 func Open(dbType DbType, connectionUrl string, opt ...Option) (*DB, error) {
+	const op = "db.Open"
 	var dialect gorm.Dialector
```
tmessi, line 83: this looks to be the reason that gorm is still being imported into this file. This is passed to `dbw.OpenWith`, but it looks like there is a `dbw.Dialector` that is a thin wrapper ofr `gorm.Dialector`, so I think this should be changed to `var dialect dbw.Dialector` and that will remove `gorm` from the imports.

jimlambrt: ty!

jimlambrt:
@tmessi @mgaffney - I'm curious if we can get this reviewed/merged before the long break.

tmessi (commented):
I have gotten through most of this, but I mainly focused around the `db` and `oplog` packages. It seems like most of the changes in other packages are just renames due to the package changes, but I might have missed/not yet seen important changes in other packages. 

The `db` package changes look good to me. Likewise `oplog` seems good, but I would note that there was more happening in that package then I was expecting to facilitate the change to `go-dbw`. It seems like new functionality is coming in along with the refactor, which would be good to note if that is the case.

I can continue to review this tomorrow, but I also don't want to hold it up since I don't have any pending questions or concerns.

jimlambrt:
@tmessi - You are correct about the changes outside of the `db` and `oplog` pkgs.  

The `oplog` refactor did require some significant changes.  We were able to eliminate the oplog's writer completely and rely on `dbw` for those capabilities.  But... we did need to deal with serializing `dbw` Options for oplog messages.      Also, we weren't properly serializing options for the `db` pkg, so it would appear like this is new functionality... but it's actually just addressing an oversight in the current implementation.

jimlambrt:
@tmessi @mgaffney - Just a reminder of this PR as we return from the holidays.  ty!

tmessi (approved):

internal/db/testing.go
```diff
@@ -229,6 +252,27 @@ func WithTemplate(template string) TestOption {
...
+	DefaultTestLogLevel TestLogLevel = iota
+	SilentTestLogLevel
+	ErrorTestLogLevel
+	WarnTestLogLevel
+	InfoTestLogLevel
+)
+
+// WithTestLogLevel provides a way to specify a test log level for the
+// underlying database package (if applicable).  To be clear, this has no affect
+// on the events that may be emitted.
```
mgaffney, line 269: I'm having trouble understanding the last sentence in the context of the `db` package. Can we either remove it or clarify it by adding more details?

internal/oplog/writer.go
```diff
@@ -1,175 +1,70 @@
...
-		if err := w.Tx.Table(tabler.TableName()).Create(i).Error; err != nil {
-			return errors.WrapDeprecated(err, op)
-		}
-	} else {
-		if err := w.Tx.Create(i).Error; err != nil {
-			return errors.WrapDeprecated(err, op)
-		}
-	}
-	return nil
+// Writer uses a gorm DB connection for writing
```
mgaffney, line 12: Why mention gorm here?

mgaffney (changes requested):
Overall, this looks great. Nice work! Can you run `make fmt` on the code? When I ran it locally there were changes. Other than that, I left some minor comments and questions.

internal/errors/error.go
```diff
@@ -211,7 +211,7 @@ func Convert(e error) *Err {
 		}
 		switch pgxError.Code {
 		case "42P01":
-			return E(ctx, WithoutEvent(), WithCode(MissingTable), WithMsg(pgxError.Message)).(*Err)
+			return E(ctx, WithoutEvent(), WithCode(MissingTable), WithMsg(pgxError.Message), WithWrap(e)).(*Err)
```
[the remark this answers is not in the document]
jimlambrt: This was just an existing issue where we were dropping the error from Postgres, so I updated it to wrap the underlying error and stop be lossy.

internal/db/testing.go
```diff
@@ -55,7 +58,12 @@ func TestSetup(t *testing.T, dialect string, opt ...TestOption) (*DB, string) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	db.Logger.LogMode(logger.Error)
+	switch {
+	case opts.withLogLevel != 0:
+		db.wrapped.LogLevel(dbw.LogLevel(opts.withLogLevel))
+	default:
+		db.wrapped.LogLevel(dbw.LogLevel(dbw.Error))
```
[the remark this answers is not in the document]
jimlambrt: ty!

internal/oplog/writer.go
jimlambrt: Not needed.  ty!

internal/db/testing.go
```diff
@@ -55,7 +58,12 @@ func TestSetup(t *testing.T, dialect string, opt ...TestOption) (*DB, string) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	db.Logger.LogMode(logger.Error)
+	switch {
+	case opts.withLogLevel != 0:
+		db.wrapped.LogLevel(dbw.LogLevel(opts.withLogLevel))
```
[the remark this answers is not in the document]
jimlambrt: If this happens, we could get the incorrect log level from the underlying db.  For instance, if for some reason there's a breaking change and the underlying database changes from Silent = 1 to Silent =20 and Error now == 1, then our logging from the underlying db would be incorrect.  With that said, if this happens it's a horribly inconvenient defect but it has no effect on the "correctness" of db operations.

internal/db/testing.go
```diff
@@ -55,7 +58,12 @@ func TestSetup(t *testing.T, dialect string, opt ...TestOption) (*DB, string) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	db.Logger.LogMode(logger.Error)
+	switch {
+	case opts.withLogLevel != 0:
```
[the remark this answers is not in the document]
jimlambrt: ty!

internal/db/testing.go
jimlambrt: Happy to remove it... especially if it's just confusing the reader. ty!

mgaffney (approved):

PR #1786, opened 2021-12-12
[DRAFT] Onboard to CRT

mdeggies:
Onboard Boundary ENT to Common Release Tooling. 

There are a few changes to call out in this PR:

- GitHub Actions is being used as the CI provider to build the go binary artifacts. This logic lives in .`github/workflows/build.yml` and workflow runs are available under https://github.com/hashicorp/boundary-enterprise/actions. 
- Shared [documentation exists here](https://docs.google.com/document/d/14v-KQKhwSsfduJQPaD13GZtHwxIUx0nRDFspKtZAslQ/edit#heading=h.2t5zdr4q7jom) that describes how to stage/promote artifacts to production.
- The `.release/ci.hcl` file contains some metadata for our release service. The only part you'll likely ever need to change is the `release_branches` array -- when you want to release off of a new release branch, update this value to the release branch name and be sure to update the branches array in `.github/workflows/build.yml` to the same value. This should be set to either `main` or the release branch that we'll be releasing off of.

Things to call out in the changelog for the next release using CRT, assuming Packagespec is fully deprecated:

- Artifact builds will now only run on merges to the release branches or to `main`
- The [dockerfile](https://github.com/hashicorp/boundary/blob/main/Dockerfile) that is used to build the boundary docker image available at [hashicorp/boundary](https://hub.docker.com/repository/docker/hashicorp/boundary) now lives in the root of this repo, and the entrypoint is available under [.release/docker/docker-entrypoint.sh](https://github.com/hashicorp/boundary/blob/main/.release/docker/docker-entrypoint.sh)
- The boundary linux packaging service configs and pre/post install scripts are now available under [.release/linux](https://github.com/hashicorp/boundary/blob/main/.release/linux)
- Boundary linux packages are now available for all supported linux architectures including arm, arm64, 386, and amd64


.circleci/Makefile
```diff
@@ -92,4 +92,4 @@ config-up-to-date: $(TMP) # Note this must not depend on $(OUT)!
 	else \
 		echo "Generated $(OUT) is out of date, run make $(CONFIG) to update."; \
 		exit 1; \
-	fi
+	fi
```
mdeggies, line 95: This fixes the previously broken `make ci-config` command

mdeggies:
Hey @kpenfound, was hoping you could review the linux packaging parts here at some point :D I realize it's not quite right, and I wasn't sure whether we'd want separate controller/worker service files or if we'd want to combine them together in `boundary.service`. I used https://stackoverflow.com/a/48232848/622508 as a reference.

Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM docker.mirror.hashicorp.services/alpine:3.13.6 as default
```
tmessi, line 1: Since this replaces the previous release dockerfile, can you also remove `docker/Release.dockerfile` in this PR?

tmessi: Also a question: Is it a requirement of CRT that the dockerfile be at the repo root? Moving the dockerfile out of the `docker` directory makes several things stated in the [docker/README.md](https://github.com/hashicorp/boundary/blob/main/docker/README.md#boundary-docker-image) incorrect.

If the dockerfile must be here that is ok. I can update that README in a followup PR, but I just want to check before doing so.

tmessi (changes requested):
This looks great, I just have one minor question/requested change.

Dockerfile
tmessi: I think we still want the `Dev.dockerfile`, since the `docker-build-dev` make target uses it, and it is helpful for developers. I suppose that file could be moved into the repo root, and the `docker-build-dev` target could be updated. Alternatively I can do those updates in a separate PR since they are not directly related to the release CRT changes, but would just be some cleanup and consolidation of docker for local dev.

> I'm happy to update the README -- do you still want this under the docker directory, or should we move it under .release/docker since there won't be anything else in this dir?

I think if everything else gets removed from the `docker` directory, moving the README to `.release/docker` would be good

.circleci/config/executors/@executors.yml
```diff
@@ -1,10 +1,10 @@
 go-machine:
   machine:
-    image: 'ubuntu-2004:202107-02'
+    image: 'ubuntu-2004'
```
mdeggies, line 3: Note: This is required, as the old image `ubuntu-2004:202107-02` was deprecated by circleci.

Dockerfile
mdeggies: I just pushed up my latest changes with the different docker targets defined in the root Dockerfile, and the docker README moved to `.release/docker` and a few edits made there. Let me know what you think- can split this out into another PR if needed.

mdeggies:
The docker image builds are failing due to an issue with github's hosted runners that we are also seeing in the consul repo. We have an open support case w/github about this -- hopefully this will be fixed soon! FWIW, local builds of all docker targets (except the default target which can only be built in CI) are successful.

Dockerfile
tmessi: Nice, this all looks good. I think it make sense to keep this all in the same PR

tmessi (dismissed):
This looks great to me.

tmessi (dismissed):

tmessi (approved):

PR #1792, opened 2021-12-15
internal/host/plugin: add job for synchronizing secrets

vancluever:
This adds the RefreshHostCatalogPersistedJob, and additional supporting
functionality:

* Addition of the TtlSeconds and RefreshAtTime values to
  HostCatalogPersisted for describing a refresh TTL and refresh
  timestamp. Plugins should return TtlSeconds to specify that secrets
  should be refreshed, and look at RefreshAtTime if needed to see when
  secrets are next scheduled to be refreshed.

* The addition of the RefreshAtTime field to HostCatalogSecret and its
  respective field in the database. This is computed from the returned
  TtlSeconds value and stored as a timestamp.

* The plugin call RefreshHostCatalogPersisted is used by
  RefreshHostCatalogPersistedJob to, if a plugin wishes, refresh
  secrets that are found in the host catalog persisted state. This
  should be used in conjunction with the values specified above, in
  addition to the catalog and secret data normally provided.

A number of refactors have been done to existing repository code to
support the changes.


vancluever:
Just a note: This is a draft PR designed to showcase the changes that are being made for this feature. It's likely that this PR will be closed in favor of smaller pull requests (example: scheduler additions may be moved to their own PR and merged ahead of time).

vancluever:
Updated and squashed one more time. The one main reason being was so that we could update the main commit message and remove details surrounding the scheduler refactoring, which has now been removed.

PR #1803, opened 2021-12-17
Add target resource listing to default roles

### LINKED ISSUE #1791: Permissions Model is Untenable
**Describe the bug**
The existing permissions model is untenable.  It is so complicated that I would imagine folks will avoid least privilege just to ensure functionality.  I don't have any confidence that the permissions I set are 1) correct or 2) least privilege.

**To Reproduce**
1. Start boundary in dev mode
2. Create a role at global scope and set grant scope to generated project scope
3. Add unprivileged `user` as principal
4. Add grant string `id=ttcp_1234567890;actions=read,authorize-session`
5. Login as user
6. See no targets in web UI or boundary app

**Expected behavior**
If I have a role that says `user` should be able to `read` and `authorize-session` on target `ttcp_1234567890` on the project scope, then I expect to see the target listed, else how would it be possible for user to do those things?

**Additional context**
Any permissions that are required in order for other permissions to take effect need to be adequately documented, but preferably avoided altogether.  I love the idea of this product but I have been unable to dial-in permissions since I launched our infrastructure six months ago.  It has been a big problem for us.

jefferai:
This addresses the issue of a user being created in a normal or
dev environment that is given access to authorize sessions against
specific targets but the permission hasn't been configured for the UI
clients to actually list them.

This solves it by enable listing for targets as part of default
grants. Listing visibility rules will still apply.

Resolves #1791


jefferai:
Note: I didn't add tests for TestProject (and there aren't any for TestOrg); after this is merged I'd like to combine testProject/TestProject and testOrg/TestOrg. There's really no need for both anymore, and there are already tests for one pair (which will be the tests used for the combined pairs).

jimlambrt (dismissed):
ty!

jimlambrt (dismissed):

PR #1805, opened 2021-12-17
fix(cli): Remove duplication of target host sources

A-440Hz:
Before this fix, calling `boundary targets read -id ttcp_1234567890` in the boundary dev environment would produce duplicate Host Catalog ID and ID values in the Host Sources section. Now, the same output should look like the sample below:

Target information:
  Created Time:               Fri, 17 Dec 2021 19:05:14 PST
  Description:                Provides an initial target in Boundary
  ID:                         ttcp_1234567890
  Name:                       Generated target
  Session Connection Limit:   -1
  Session Max Seconds:        28800
  Type:                       tcp
  Updated Time:               Fri, 17 Dec 2021 19:05:14 PST
  Version:                    2

  Scope:
    ID:                       p_1234567890
    Name:                     Generated project scope
    Parent Scope ID:          o_1234567890
    Type:                     project

  Authorized Actions:
    no-op
    read
    update
    delete
    add-host-sets
    set-host-sets
    remove-host-sets
    add-credential-libraries
    set-credential-libraries
    remove-credential-libraries
    add-credential-sources
    set-credential-sources
    remove-credential-sources
    authorize-session

  Host Sources:
    Host Catalog ID:          hcst_1234567890
    ID:                       hsst_1234567890

  Attributes:
    Default Port:             22

### DIFF
--- a/internal/cmd/commands/targetscmd/funcs.go
+++ b/internal/cmd/commands/targetscmd/funcs.go
@@ -676,41 +676,31 @@ func printItemTable(result api.GenericResult) string {
 	maxLength := base.MaxAttributesLength(nonAttributeMap, item.Attributes, keySubstMap)
 
 	var hostSourceMaps []map[string]interface{}
-	// used as a key to ensure there are no repeat printed values among HostSources and HostSets
-	type setInfo struct {
-		id            interface{}
-		hostCatalogID interface{}
-	}
-	hostSourceInfo := make(map[setInfo]map[string]interface{})
-	if len(item.HostSources) > 0 {
+	switch {
+	case len(item.HostSources) > 0:
 		for _, set := range item.HostSources {
 			m := map[string]interface{}{
 				"ID":              set.Id,
 				"Host Catalog ID": set.HostCatalogId,
 			}
-			k := setInfo{set.Id, set.HostCatalogId}
-			hostSourceInfo[k] = m
+			hostSourceMaps = append(hostSourceMaps, m)
 		}
 		if l := len("Host Catalog ID"); l > maxLength {
 			maxLength = l
 		}
-	}
-	if len(item.HostSets) > 0 {
+	case len(item.HostSets) > 0:
 		for _, set := range item.HostSets {
 			m := map[string]interface{}{
 				"ID":              set.Id,
 				"Host Catalog ID": set.HostCatalogId,
 			}
-			k := setInfo{set.Id, set.HostCatalogId}
-			hostSourceInfo[k] = m
+			hostSourceMaps = append(hostSourceMaps, m)			
 		}
 		if l := len("Host Catalog ID"); l > maxLength {
 			maxLength = l
 		}
 	}
-	for _, m := range hostSourceInfo {
-		hostSourceMaps = append(hostSourceMaps, m)
-	}
+
 
 	var credentialSourceMaps map[credential.Purpose][]map[string]interface{}
 	if len(item.ApplicationCredentialSources) > 0 {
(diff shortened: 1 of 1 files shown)


internal/cmd/commands/targetscmd/funcs.go
```diff
@@ -676,13 +676,20 @@ func printItemTable(result api.GenericResult) string {
 	maxLength := base.MaxAttributesLength(nonAttributeMap, item.Attributes, keySubstMap)
 
 	var hostSourceMaps []map[string]interface{}
+	// used as a key to ensure there are no repeat printed values among HostSources and HostSets
+	type setInfo struct {
+		id            interface{}
```
louisruch, line 681: Curious to why these are both `interface{}`.  Both `id` and `hostCatalogID` are strings

internal/cmd/commands/targetscmd/funcs.go
```diff
@@ -676,13 +676,20 @@ func printItemTable(result api.GenericResult) string {
...
+		hostCatalogID interface{}
+	}
+	hostSourceInfo := make(map[setInfo]map[string]interface{})
```
louisruch, line 684: So the need for the map in `CredentialLibraries`/`CredentialSources` makes sense because we want to group them by `Purpose`.  I am not sure its the best approach here, a slightly more maintainable and readable approach would be to use a switch statement, so something like:

```
switch {
case len(item.HostSets) > 0:
    // use HostSets
case len(item.HostSources) > 0:
   // use HostSources
}
```

I also have a hunch that the usage of the map in the credential library case was not by design to avoid duplication but rather a side effect of grouping them by purpose.

louisruch (changes requested):
Great to see your first PR!!  I added a comment about a possible alternative here. Happy to chat it through

internal/cmd/commands/targetscmd/funcs.go
A-440Hz: I did it this way because in line 687 `m` is defined as `map[string]interface{}`

internal/cmd/commands/targetscmd/funcs.go
A-440Hz: I agree your suggestion makes more sense. I can test the implementation, but I didn't start in this direction because I had a hunch that this wouldn't work since the json data had host_sets and host_sources both populated.

internal/cmd/commands/targetscmd/funcs.go
louisruch: Aha, so the CLI uses https://github.com/hashicorp/boundary/blob/main/internal/cmd/base/format.go#L119 to allow uniform rendering of maps.  Notice the input to that function is a `map[string]interface{}` here the use of `interface{}` is needed because the `WrapMap` function needs to deal embedded maps.  When possible its best to be specific about the type

internal/cmd/commands/targetscmd/funcs.go
louisruch: That is a great point, so the reason for this is backward compatibility.  This means we will only have one of the following scenarios:
1) Only `HostSets` is populated
2) Both `HostSets` and `HostSources` are populated, but both have the exact same value.

internal/cmd/commands/targetscmd/funcs.go
A-440Hz: I see. Seems like I need to get better at looking through the code to figure out why things are written a certain way rather than just copying a format across changes

louisruch (approved):
LGTM

PR #1828, opened 2022-01-25
feat(host): Add host set information to static host

A-440Hz:
Updates LookupHost and ListHost API to include sets a static host is part of, as like with plugin hosts.

### DIFF
--- a/internal/db/schema/migrations/oss/postgres/25/01_static_host_view.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/25/01_static_host_view.up.sql
@@ -1,0 +1,24 @@
+begin;
+
+-- static_host_with_set_memberships is used for associating a static host instance with all its related host sets
+-- in the set_ids column. Currently there are no size limits.
+create view static_host_with_set_memberships as
+select
+  h.public_id,
+  h.create_time,
+  h.update_time,
+  h.name,
+  h.description,
+  h.catalog_id,
+  h.address,
+  h.version,
+  -- the string_agg(..) column will be null if there are no associated value objects
+  string_agg(distinct hsm.set_id, '|') as set_ids
+from
+  static_host h
+    left outer join static_host_set_member hsm on h.public_id = hsm.host_id
+group by h.public_id;
+comment on view static_host_with_set_memberships is
+  'static host with its associated host sets';
+
+commit;
--- a/internal/host/static/host.go
+++ b/internal/host/static/host.go
@@ -1,6 +1,10 @@
 package static
 
 import (
+	"sort"
+	"strings"
+
+	"github.com/hashicorp/boundary/internal/db/timestamp"
 	"github.com/hashicorp/boundary/internal/errors"
 	"github.com/hashicorp/boundary/internal/host/static/store"
 	"github.com/hashicorp/boundary/internal/oplog"
@@ -15,7 +19,8 @@ const (
 // A Host contains a static address.
 type Host struct {
 	*store.Host
-	tableName string `gorm:"-"`
+	SetIds    []string `gorm:"-"`
+	tableName string   `gorm:"-"`
 }
 
 // NewHost creates a new in memory Host for address assigned to catalogId.
@@ -70,9 +75,18 @@ func allocHost() *Host {
 
 func (h *Host) clone() *Host {
 	cp := proto.Clone(h.Host)
-	return &Host{
+	nh := &Host{
 		Host: cp.(*store.Host),
 	}
+	switch {
+	case h.SetIds == nil:
+	case len(h.SetIds) == 0:
+		nh.SetIds = make([]string, 0)
+	default:
+		nh.SetIds = make([]string, len(h.SetIds))
+		copy(nh.SetIds, h.SetIds)
+	}
+	return nh
 }
 
 func (h *Host) oplog(op oplog.OpType) oplog.Metadata {
@@ -85,4 +99,56 @@ func (h *Host) oplog(op oplog.OpType) oplog.Metadata {
 		metadata["catalog-id"] = []string{h.CatalogId}
 	}
 	return metadata
+}
+
+// GetSetIds returns host set ids
+func (h *Host) GetSetIds() []string {
+	return h.SetIds
+}
+
+type hostAgg struct {
+	PublicId    string `gorm:"primary_key"`
+	CatalogId   string
+	Name        string
+	Description string
+	CreateTime  *timestamp.Timestamp
+	UpdateTime  *timestamp.Timestamp
+	Version     uint32
+	Address     string
+	SetIds      string
+}
+
+func (agg *hostAgg) toHost() *Host {
+	h := allocHost()
+	h.PublicId = agg.PublicId
+	h.CatalogId = agg.CatalogId
+	h.Name = agg.Name
+	h.Description = agg.Description
+	h.CreateTime = agg.CreateTime
+	h.UpdateTime = agg.UpdateTime
+	h.Version = agg.Version
+	h.Address = agg.Address
+	h.SetIds = agg.getSetIds()
+	return h
+}
+
+// TableName returns the table name for gorm
+func (agg *hostAgg) TableName() string {
+	return "static_host_with_set_memberships"
+}
+
+// GetPublicId returns the host public id as a string
+func (agg *hostAgg) GetPublicId() string {
+	return agg.PublicId
+}
+
+// GetSetIds returns a list of all associated host sets to the host
+func (agg *hostAgg) getSetIds() []string {
+	const aggregateDelimiter = "|"
+	var ids []string
+	if agg.SetIds != "" {
+		ids = strings.Split(agg.SetIds, aggregateDelimiter)
+		sort.Strings(ids)
+	}
+	return ids
 }
--- a/internal/host/static/repository_host.go
+++ b/internal/host/static/repository_host.go
@@ -171,6 +171,9 @@ func (r *Repository) UpdateHost(ctx context.Context, scopeId string, h *Host, ve
 			if err != nil {
 				return errors.Wrap(ctx, err, op)
 			}
+			if rowsUpdated > 1 {
+				return errors.New(ctx, errors.MultipleRecords, op, "more than 1 resource would have been updated")
+			}
 			ha := &hostAgg{
 				PublicId: h.PublicId,
 			}
@@ -178,9 +181,6 @@ func (r *Repository) UpdateHost(ctx context.Context, scopeId string, h *Host, ve
 				return errors.Wrap(ctx, err, op, errors.WithMsg("failed to lookup host after update"))
 			}
 			returnedHost.SetIds = ha.getSetIds()
-			if rowsUpdated > 1 {
-				return errors.New(ctx, errors.MultipleRecords, op, "more than 1 resource would have been updated")
-			}
 			return nil
 		},
 	)
--- a/internal/host/static/repository_host.go
+++ b/internal/host/static/repository_host.go
@@ -171,6 +171,13 @@ func (r *Repository) UpdateHost(ctx context.Context, scopeId string, h *Host, ve
 			if err != nil {
 				return errors.Wrap(ctx, err, op)
 			}
+			ha := &hostAgg{
+				PublicId: h.PublicId,
+			}
+			if err := r.reader.LookupByPublicId(ctx, ha); err != nil {
+				return errors.Wrap(ctx, err, op, errors.WithMsg("failed to lookup host after update"))
+			}
+			returnedHost.SetIds = ha.getSetIds()
 			if rowsUpdated > 1 {
 				return errors.New(ctx, errors.MultipleRecords, op, "more than 1 resource would have been updated")
 			}
@@ -203,15 +210,16 @@ func (r *Repository) LookupHost(ctx context.Context, publicId string, opt ...Opt
 	if publicId == "" {
 		return nil, errors.New(ctx, errors.InvalidParameter, op, "no public id")
 	}
-	h := allocHost()
-	h.PublicId = publicId
-	if err := r.reader.LookupByPublicId(ctx, h); err != nil {
+	ha := &hostAgg{
+		PublicId: publicId,
+	}
+	if err := r.reader.LookupByPublicId(ctx, ha); err != nil {
 		if errors.IsNotFoundError(err) {
 			return nil, nil
 		}
 		return nil, errors.Wrap(ctx, err, op, errors.WithMsg(fmt.Sprintf("failed for %s", publicId)))
 	}
-	return h, nil
+	return ha.toHost(), nil
 }
 
 // ListHosts returns a slice of Hosts for the catalogId.
@@ -227,11 +235,16 @@ func (r *Repository) ListHosts(ctx context.Context, catalogId string, opt ...Opt
 		// non-zero signals an override of the default limit for the repo.
 		limit = opts.withLimit
 	}
-	var hosts []*Host
-	err := r.reader.SearchWhere(ctx, &hosts, "catalog_id = ?", []interface{}{catalogId}, db.WithLimit(limit))
+	var aggs []*hostAgg
+	err := r.reader.SearchWhere(ctx, &aggs, "catalog_id = ?", []interface{}{catalogId}, db.WithLimit(limit))
 	if err != nil {
 		return nil, errors.Wrap(ctx, err, op)
 	}
+	hosts := make([]*Host, 0, len(aggs))
+	for _, ha := range aggs {
+		hosts = append(hosts, ha.toHost())
+	}
+
 	return hosts, nil
 }
 
--- a/internal/host/static/repository_host_test.go
+++ b/internal/host/static/repository_host_test.go
@@ -606,6 +606,12 @@ func TestRepository_UpdateHost(t *testing.T) {
 			assert.NoError(err)
 			require.NotNil(orig)
 
+			if tt.wantIsErr == 0 {
+				set := TestSets(t, conn, catalog.GetPublicId(), 1)[0]
+				TestSetMembers(t, conn, set.PublicId, []*Host{orig})
+				tt.want.SetIds = []string{set.PublicId}
+			}
+
 			if tt.chgFn != nil {
 				orig = tt.chgFn(orig)
 			}
@@ -629,6 +635,9 @@ func TestRepository_UpdateHost(t *testing.T) {
 			if tt.want.Name == "" {
 				dbassert.IsNull(got, "name")
 				return
+			}
+			if tt.wantIsErr == 0 {
+				assert.Equal(tt.want.SetIds, got.SetIds)
 			}
 			assert.Equal(tt.want.Name, got.Name)
 			if tt.want.Description == "" {
--- a/internal/host/static/repository_host_test.go
+++ b/internal/host/static/repository_host_test.go
@@ -2,6 +2,7 @@ package static
 
 import (
 	"context"
+	"sort"
 	"testing"
 	"time"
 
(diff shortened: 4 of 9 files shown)


internal/db/schema/migrations/oss/postgres/25/01_static_host_view.up.sql
```diff
@@ -0,0 +1,22 @@
+begin;
+
+create view static_host_with_set_memberships as
```
talanknight, line 3: Please add a comment to this view explaining its purpose and details regarding the unique return value of the `set_ids` column (the delimitation and that there is no size limits currently).

internal/servers/controller/handlers/hosts/host_service_test.go
```diff
@@ -320,12 +326,20 @@ func TestList_Static(t *testing.T) {
...
 			assert.Empty(cmp.Diff(got, tc.res, protocmp.Transform()), "ListHosts(%q) got response %q, wanted %q", tc.req, got, tc.res)
 
 			// Test anonymous listing
 			got, gErr = s.ListHosts(auth.DisabledAuthTestContext(iamRepoFn, proj.GetPublicId(), auth.WithUserId(auth.AnonymousUserId)), tc.req)
 			require.NoError(gErr)
 			assert.Len(got.Items, len(tc.res.Items))
+			gItems = got.GetItems()
+			sort.Slice(gItems, func(i int, j int) bool {
+				return gItems[i].GetId() < gItems[j].GetId()
+			})
```
talanknight, line 342: This sorting doesn't seem useful.  Please remove.

internal/servers/controller/handlers/hosts/host_service_test.go
```diff
@@ -253,8 +253,11 @@ func TestList_Static(t *testing.T) {
 	hcs := static.TestCatalogs(t, conn, proj.GetPublicId(), 2)
 	hc, hcNoHosts := hcs[0], hcs[1]
 
+	hset := static.TestSets(t, conn, hc.GetPublicId(), 1)[0]
 	var wantHs []*pb.Host
-	for _, h := range static.TestHosts(t, conn, hc.GetPublicId(), 10) {
+	testHosts := static.TestHosts(t, conn, hc.GetPublicId(), 10)
+	static.TestSetMembers(t, conn, hset.GetPublicId(), testHosts)
+	for _, h := range testHosts {
 		wantHs = append(wantHs, &pb.Host{
```
talanknight, line 261: Where is the SetIds expected on these wanted hosts?  I would expect either the test to fail if we got host set ids being returned by the API.  If they aren't failing then it seems like we are still not returning the set ids.

talanknight (changes requested):
Please also return the host's sets as a response to the UpdateHost call.

internal/servers/controller/handlers/hosts/host_service_test.go
A-440Hz: yep seems like the `group by` in the view already covers this sorting

talanknight: I meant it wasn't useful because there isn't anything following this sort which relies on these values being sorted.

A-440Hz: I put in a sort.Slice on 274 which sorted the `wantHS` and passed the "List Many Hosts" test that was previously failing

internal/db/schema/migrations/oss/postgres/25/01_static_host_view.up.sql
```diff
@@ -0,0 +1,24 @@
...
+  h.name,
+  h.description,
+  h.catalog_id,
+  h.address,
+  h.version,
+  -- the string_agg(..) column will be null if there are no associated value objects
+  string_agg(distinct hsm.set_id, '|') as set_ids
+from
+  static_host h
+    left outer join static_host_set_member hsm   on h.public_id = hsm.host_id
```
louisruch, line 19: nit: white space
```suggestion
    left outer join static_host_set_member hsm on h.public_id = hsm.host_id
```

internal/host/static/repository_host.go
```diff
@@ -171,6 +171,16 @@ func (r *Repository) UpdateHost(ctx context.Context, scopeId string, h *Host, ve
...
+			ha := &hostAgg{
+				PublicId: h.PublicId,
+			}
+			if err := r.reader.LookupByPublicId(ctx, ha); err != nil {
+				if errors.IsNotFoundError(err) {
+					return errors.Wrap(ctx, err, op)
+				}
+				return errors.Wrap(ctx, err, op)
+			}
+			returnedHost.SetIds = ha.GetSetIds()
```
louisruch, line 183: We can probably drop this `if errors.IsNotFoundError(err)` and just do a single return on error.  Since we are in a transaction if the update succeeded we should be be able to get a not found, also it doesn't do anything differently from the return outside the if statement.
This is also a good case for adding a custom message to the wrapped error.
```suggestion
			if err := r.reader.LookupByPublicId(ctx, ha); err != nil {
				return errors.Wrap(ctx, err, op, errors.WithMsg("failed to lookup host after update")
			}
			returnedHost.SetIds = ha.GetSetIds()
```

internal/servers/controller/handlers/hosts/host_service.go
```diff
@@ -134,6 +134,8 @@ func (s Service) ListHosts(ctx context.Context, req *pbs.ListHostsRequest) (*pbs
 		switch h := item.(type) {
 		case *plugin.Host:
 			outputOpts = append(outputOpts, handlers.WithHostSetIds(h.SetIds))
+		case *static.Host:
```
louisruch, line 137: Since all host types now support setIds it seems we should extend the Host interface defined [here](https://github.com/hashicorp/boundary/blob/main/internal/host/host.go#L24) to include a `GetSetIds() []string` method. That way we can skip this type switch and just do:
```
outputOpts = append(outputOpts, handlers.WithHostSetIds(item.GetSetIds()))
```

internal/servers/controller/handlers/hosts/host_service.go
```diff
@@ -184,6 +186,8 @@ func (s Service) GetHost(ctx context.Context, req *pbs.GetHostRequest) (*pbs.Get
 	switch h := h.(type) {
 	case *plugin.Host:
 		outputOpts = append(outputOpts, handlers.WithHostSetIds(h.SetIds))
+	case *static.Host:
```
louisruch, line 189: Same comment re updating the interface

louisruch (changes requested):
This looks really good! I left a couple of nits and comments, lets chat tomorrow about updating the interface!

internal/host/static/host.go
```diff
@@ -86,3 +100,47 @@ func (h *Host) oplog(op oplog.OpType) oplog.Metadata {
...
+
+func (agg *hostAgg) TableName() string {
+	return "static_host_with_set_memberships"
+}
+
+func (agg *hostAgg) GetPublicId() string {
+	return agg.PublicId
+}
+
+func (agg *hostAgg) GetSetIds() []string {
```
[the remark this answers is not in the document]
louisruch: Given to point around updating the interface this should remain exported and a comment be added

internal/host/static/host.go
```diff
@@ -86,3 +100,47 @@ func (h *Host) oplog(op oplog.OpType) oplog.Metadata {
...
+	h.Description = agg.Description
+	h.CreateTime = agg.CreateTime
+	h.UpdateTime = agg.UpdateTime
+	h.Version = agg.Version
+	h.Address = agg.Address
+	h.SetIds = agg.GetSetIds()
+	return h
+}
+
+func (agg *hostAgg) TableName() string {
```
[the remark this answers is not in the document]
A-440Hz: should I do the same for `GetPublicId` and `GetSetIds`?

internal/servers/controller/handlers/hosts/host_service_test.go
```diff
@@ -266,8 +272,12 @@ func TestList_Static(t *testing.T) {
 				"address": structpb.NewStringValue(h.GetAddress()),
 			}},
 			AuthorizedActions: testAuthorizedActions[static.Subtype],
+			HostSetIds:        []string{hset.GetPublicId()},
```
talanknight, line 275: Something similar needs to be added for the tests around UpdateHost.

internal/host/static/repository_host_test.go
```diff
@@ -798,6 +800,72 @@ func TestRepository_LookupHost(t *testing.T) {
 	}
 }
 
+func TestRepository_LookupHost_HostSets(t *testing.T) {
```
talanknight, line 803: Can you add one of these tests for UpdateHost?

internal/tests/api/hostsets/host_set_test.go
```diff
@@ -246,27 +253,24 @@ func TestCrud(t *testing.T) {
...
 		hostsets.WithPreferredEndpoints([]string{"dns:update"}))
-	require.NoError(err)
 	assert.Equal("bar", h.Item.Name)
-	switch retried {
-	case true:
-		assert.Equal(uint32(3), h.Item.Version)
-	default:
-		assert.Equal(uint32(2), h.Item.Version)
-	}
+	// If the plugin syncs after update, its version will increment, so it can either be 2 or 3
```
talanknight, line 272:
```suggestion
	// If the plugin set has synced since creation it can either be 3, otherwise it'll be 2.
```

internal/tests/api/hostsets/host_set_test.go
```diff
@@ -246,27 +253,24 @@ func TestCrud(t *testing.T) {
...
+		hostsets.WithSyncIntervalSeconds(-1))
 	require.NoError(err)
 	assert.Equal("foo", h.Item.Name)
 	assert.Equal(uint32(1), h.Item.Version)
 
 	h, err = hClient.Read(tc.Context(), h.Item.Id)
 	require.NoError(err)
 	assert.Equal("foo", h.Item.Name)
-	assert.Equal(uint32(1), h.Item.Version)
+	// If the plugin syncs after creation, its version will increment, so it can either be 1 or 2
```
talanknight, line 265:
```suggestion
	// If the plugin set has synced after creation this can be 2, otherwise it'll be 1.
```

talanknight (changes requested):
Great progress!

internal/host/static/repository_host.go
```diff
@@ -171,6 +171,13 @@ func (r *Repository) UpdateHost(ctx context.Context, scopeId string, h *Host, ve
...
 				return errors.Wrap(ctx, err, op)
 			}
+			ha := &hostAgg{
+				PublicId: h.PublicId,
+			}
+			if err := r.reader.LookupByPublicId(ctx, ha); err != nil {
+				return errors.Wrap(ctx, err, op, errors.WithMsg("failed to lookup host after update"))
+			}
+			returnedHost.SetIds = ha.getSetIds()
 			if rowsUpdated > 1 {
```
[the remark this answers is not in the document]
A-440Hz: moved the rowsUpdated check to before I instantiate the hostAgg

internal/tests/api/hostsets/host_set_test.go
```diff
@@ -246,27 +253,24 @@ func TestCrud(t *testing.T) {
...
+	assert.Contains([]uint32{1, 2}, h.Item.Version)
 
-	h, retried := retryableUpdate(hClient, h.Item.Id, h.Item.Version, hostsets.WithName("bar"),
+	h, _ = retryableUpdate(hClient, h.Item.Id, h.Item.Version, hostsets.WithName("bar"),
```
louisruch, line 268: Since we are not using the bool returned from retryable update and doing a contains check instead, can you update the function definition to be something like:
```
retryableUpdate := func(c *hostsets.Client, hcId string, version uint32, opts ...hostsets.Option) *hostsets.HostSetUpdateResult {
```

internal/host/static/repository_host_test.go
```diff
@@ -857,6 +934,82 @@ func TestRepository_ListHosts(t *testing.T) {
...
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			assert, require := assert.New(t), require.New(t)
+			repo, err := NewRepository(rw, rw, kms)
+			assert.NoError(err)
+			require.NotNil(repo)
+			got, err := repo.ListHosts(context.Background(), tt.in)
+			require.NoError(err)
+			// Make sure outputs match. Ignore timestamps.
```
talanknight, line 997: Nit: This comment doesn't provide any additional details and simply states what the next line does.  Please remove.

internal/host/static/repository_host_test.go
```diff
@@ -857,6 +934,82 @@ func TestRepository_ListHosts(t *testing.T) {
...
+	tests := []struct {
+		name      string
+		in        string
+		want      []*Host
+		wantIsErr errors.Code
+	}{
+		{
+			name: "with-hostsets",
+			in:   catalogA.PublicId,
+			want: hostsA,
```
talanknight, line 975: Nit, and FYI (This is my opinion and other's opinions may differ.  I'm more than happy to discuss further regarding if this makes following the expected in and out of each test easier or harder.):

When the test setup is simply creating variables for a number of inputs to pair with a number of precomputed outputs then much of the benefit of table driven tests is removed since you could have easily just written out 3 different test cases with potentially fewer variable names and inlined the test logic which tends to be a little easier to follow.


Much of the benefit of the table driven tests is to be able to easily see: "For this input, I expect that output." which makes it clearer when a test case might not be covered.

For example, if the tests struct had `getWantedHosts   func(*HostCatalog) []*Host` then you could have a test cases that looked like:

```
{
	name: "with-hostsets",
	getWantedHosts:   func(c *HostCatalog) []*Host {
           hosts := TestHosts(t, conn, c.PublicId, 3)
	   s := TestSets(t, conn, c.PublicId, 1)[0]
	   TestSetMembers(t, conn, s.PublicId, hosts)
	   for _, h := range hosts {
		  h.SetIds = []string{s.PublicId}
	   }
           return hosts
         },
}
```

and in t.Run you could have:
```
...
cat := TestCatalogs(t, conn, prj.PublicId, 1)[0]
want := tt.setupWantedHosts(cat)
...
```

While the getWantedHosts is a bit clunky, it at least it pairs input with the output.  I realize other tests could do better in this regard, especially ones that I've written, so again I'd be super happy to discuss further.

internal/host/static/repository_host_test.go
```diff
@@ -857,6 +934,82 @@ func TestRepository_ListHosts(t *testing.T) {
...
+		h.SetIds = []string{setC[i].PublicId}
+		TestSetMembers(t, conn, setC[i].PublicId, []*Host{hostsC[i]})
+	}
+	hostsC = append(hostsC, hostsC0...)
+
+	tests := []struct {
+		name      string
+		in        string
+		want      []*Host
+		wantIsErr errors.Code
```
talanknight, line 970: This is not used.  Remove it.

internal/servers/controller/handlers/hosts/host_service.go
```diff
@@ -265,6 +257,9 @@ func (s Service) UpdateHost(ctx context.Context, req *pbs.UpdateHostRequest) (*p
 		idActions := idActionsTypeMap[host.SubtypeFromId(req.GetId())]
 		outputOpts = append(outputOpts, handlers.WithAuthorizedActions(authResults.FetchActionSetForId(ctx, h.GetPublicId(), idActions).Strings()))
 	}
+	if outputFields.Has(globals.HostSetIdField) {
+		outputOpts = append(outputOpts, handlers.WithHostSetIds(h.GetSetIds()))
+	}
```
talanknight, line 262: Why is UpdateHost treated differently from ReadHost or ListHosts regarding the outputFields?

talanknight (changes requested):

internal/host/static/repository_host_test.go
```diff
@@ -604,6 +606,12 @@ func TestRepository_UpdateHost(t *testing.T) {
 			assert.NoError(err)
 			require.NotNil(orig)
 
+			if tt.wantIsErr == 0 {
```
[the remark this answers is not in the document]
A-440Hz: I get an invalid memory address runtime error when I attempt to set `tt.want.SetIds` on line 612 when `tt.want` does not exist. I could wrap an if statement around that line only to check if `tt.want` exists. It is still the same number of if statements, but I think it might be slightly better for cognitive load. Do you have any alternate suggestions?

internal/host/static/repository_host_test.go
A-440Hz: Yeah. I copied this comment over from a different test which probably looked messier

internal/host/static/repository_host_test.go
A-440Hz: Yeah I prefer this method a lot better because it gets rid of a lot of repetitive code. The one nit I have is that the function calls `TestSetMembers` to create the `host_set_member` object as part of its call, which could be confusing in the sense of clarity (function name suggests nothing about actually setting the hostset-host association). I prefer this method over what I currently have because it reduces a lot of repetition.

internal/servers/controller/handlers/hosts/host_service.go
A-440Hz: Good question. Before I wrote this if case, the function had nothing to append the host set ids option to proto. What determines whether a field needs an if case or is just appended by default?

internal/host/static/repository_host_test.go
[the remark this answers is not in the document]
louisruch: The error will be due to the fact that `tt.want` is nil. An alternative would be to define the want in the test case instead of adding it during the test so something like:

```
want: &Host{
	Host: &store.Host{
		Name:        "test-update-name-repo",
		Description: "test-description-repo",
		Address:     "127.0.0.1",
	},
        SetIds: []string{....},
}
```
We would also want to make sure we handle all the test cases here, two off the top of my head:
- Orig does not have any sets and we add new sets in the update
- Orig has sets and we change them in the update path

This will likely require reworking this unit test a bit to create orig with and without sets outside of the loop here.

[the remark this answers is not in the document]
A-440Hz: @louisruch I'm not quite sure what you mean about the test cases. I thought `UpdateHost` only worked on the name, description, and address. 
I am currently running tests with a singular hostset to make sure the set is persisted, upon calling `UpdateHost`. I understand the meaning of the first bullet point, but if I were to test that, by the same logic, wouldn't I have to test every field that isn't name, description, or address?
For the second bullet point, could you expand what you mean by "change them in the update path"? Thanks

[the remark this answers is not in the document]
louisruch: Sorry you are right, we have a separate call for SetHostSets!

internal/host/static/repository_host_test.go
```diff
@@ -604,6 +606,10 @@ func TestRepository_UpdateHost(t *testing.T) {
 			assert.NoError(err)
 			require.NotNil(orig)
 
+			set := TestSets(t, conn, catalog.GetPublicId(), 1)[0]
+			TestSetMembers(t, conn, set.PublicId, []*Host{orig})
+			want_SetIds := []string{set.PublicId}
```
talanknight, line 611: Variables should be camel case.

```suggestion
			wantSetIds := []string{set.PublicId}
```

internal/servers/controller/handlers/hosts/host_service.go
```diff
@@ -265,6 +257,7 @@ func (s Service) UpdateHost(ctx context.Context, req *pbs.UpdateHostRequest) (*p
 		idActions := idActionsTypeMap[host.SubtypeFromId(req.GetId())]
 		outputOpts = append(outputOpts, handlers.WithAuthorizedActions(authResults.FetchActionSetForId(ctx, h.GetPublicId(), idActions).Strings()))
 	}
+	outputOpts = append(outputOpts, handlers.WithHostSetIds(h.GetSetIds()))
```
talanknight, line 260: I didn't realize the UpdateHost call didn't already return set ids for plugin based hosts.  Since this was something that didn't exist before can you add the test for TestUpdate_Plugin as well (and update the update host flow in the plugin repo if needed)?

talanknight (changes requested):
Looks close!

internal/servers/controller/handlers/hosts/host_service.go
talanknight: Woops! Plugin doesn't have "update" so my comment doesn't make sense.

PR #1829, opened 2022-01-26
website: display warning in IE 11

dstaley:
This PR displays a warning to users of IE 11 that their browser is unsupported.

[_Created by Sourcegraph batch change `dylan.staley/ie11-warning-banner`._](https://hashicorp-mktg.sourcegraph.com/users/dylan.staley/batch-changes/ie11-warning-banner)

image


website/public/ie-warning.js
```diff
@@ -0,0 +1,40 @@
...
+
+  const link = document.createElement('a')
+  link.textContent = 'Learn more.'
+  link.href = 'https://www.microsoft.com/en-us/edge?form=MA13DL&OCID=MA13DL'
+
+  message.appendChild(link)
+  el.appendChild(message)
+
+  document.body.insertBefore(el, document.body.childNodes[0])
+})()
```
randallmorey, line 40: We should assume web standards fluency and express the markup directly as HTML and CSS, rather than obscuring it with JavaScript.  I believe this reads better and makes the intent very clear, whilst being more maintainable (in the off chance it needs to change):

```suggestion
!(function () {
  'use strict'

  const el = document.createElement('div');
  el.innerHtml = `
    <div class="ie-warning">
      <p class="ie-warning-description">
        Internet Explorer is no longer supported.
        <a href="https://www.microsoft.com/en-us/edge?form=MA13DL&OCID=MA13DL">
          Learn more.
        </a>
      </p>
    </div>

    <style>
      .ie-warning {
        background-color: #FCF0F2;
        border-bottom: 1px solid #FFD4D6;
        color: #BA2226;
        text-align: center;
        font-family: "Segoe UI", sans-serif;
        font-weight: bold;
      }
      .ie-warning-description {
        padding: 16px 0;
        margin: 0;
        color: #BA2226;
      }
    </style>
  `;

  document.body.insertBefore(el, document.body.childNodes[0]);
})();
```

randallmorey (changes requested):
See code suggestion.  Also, should we direct users to Chrome or Firefox instead?

website/public/ie-warning.js
calcaide: I agree with Randy on embracing web standards and directly expressing the markup as HTML and CSS.

As per what browser to suggest to the user, I do not have a strong opinion. I do like to re-direct it to a website where the user can choose, i.e. https://browsehappy.com/

dstaley: @randallmorey @calcaide Thank you both for the feedback!

One unfortunate aspect is that IE11 doesn't support ES6 template literals, and this file deliberately bypasses our transforms, so we'd need to do something like the following:

```js
!(function () {
  'use strict'
  const el = document.createElement('div')
  el.innerHtml =
    '<div class="ie-warning">' +
    '<p class="ie-warning-description">' +
    'Internet Explorer is no longer supported.' +
    '<a href="https://www.microsoft.com/en-us/edge?form=MA13DL&OCID=MA13DL">' +
    'Learn more.' +
    '</a>' +
    '</p>' +
    '</div>' +
    '<style>' +
    '.ie-warning {' +
    'background-color: #FCF0F2;' +
    'border-bottom: 1px solid #FFD4D6;' +
    'color: #BA2226;' +
    'text-align: center;' +
    'font-family: "Segoe UI", sans-serif;' +
    'font-weight: bold;' +
    '}' +
    '.ie-warning-description {' +
    'padding: 16px 0;' +
    'margin: 0;' +
    'color: #BA2226;' +
    '}' +
    '</style>'
  document.body.insertBefore(el, document.body.childNodes[0])
})()
```

(We could also do the ol' join-an-array-of-strings trick to get rid of the concatenate operator.)

I'm fine with using this method if you still think it has improved readability. Otherwise totally down to brainstorm some other options!

As for where we point the "Learn more" link to, we discussed using a page that provides multiple options, but ultimately went with the Microsoft link since it included language about the deprecation of IE11. That language only shows up if you're using IE11, so here's a screenshot of what that page looks like:

image

We chose that page since it provides a bit more explanation ("Some websites no longer support Internet Explorer"), and provides the Microsoft approved solution (as opposed to asking the user to make a choice). It also avoids the potentially awkward situation of appearing to recommend Microsoft's competitors.

randallmorey: I updated my original suggestion to single quoted strings.  It's similar to your suggestion, but with indentation intact.

As for the browser recommendation, I think in many cases IE is a default, not something they necessarily chose for themselves.  The could be an opportunity to educate by presenting additional options.  A third option is to not link.  Personally I'm in favour of neutrality, i.e. present a list of evergreen browsers and let the user choose.

dstaley: Ah good call, that does read a bit better!

I'm going to explore some options for where we can send the user. I'd love something that both has an explanation for why IE11 is no longer supported, but also provides that choice of browser to the user.

randallmorey (approved):

PR #1835, opened 2022-01-31
Update authenticate command/subcommands help text

### LINKED ISSUE #1833: boundary authenticate (type) can conflict with BOUNDARY_AUTH_METHOD_ID
**Describe the bug**

The command `boundary authenticate` requires the user to provide a type `oidc` or `password`. 
It also accepts setting the auth-method via `BOUNDARY_AUTH_METHOD_ID` ... which makes the user selection unnecessary, and potentially wrong. If a user environment is set up correctly for OIDC auth but they type the wrong option here they get incorrect instructions.

```
$ env |grep BO
BOUNDARY_AUTH_METHOD_ID=amoidc_wLLVtNa5kA

$ boundary authenticate
Usage: boundary authenticate [sub command] [options] [args]

  This command authenticates the Boundary commandline client using a specified auth method. Examples:

    Authenticate with password auth method:

      $ boundary authenticate password -auth-method-id ampw_1234567890 -login-name foo -password "bar"

  Please see the auth method subcommand help for detailed usage information.

$ boundary authenticate password --login-name innocent-user --password exposing-my-login-password-in-shell-history
```

**To Reproduce**
1. Set `BOUNDARY_AUTH_METHOD_ID` to an oidc method
2. Run `boundary authenticate` by itself
3. Receive instructions to login with password from help output
4. Run `boundary authenticate password` (because user doesn't recognize oidc)
5. Failure. User is frustrated, contacts helpdesk

**Expected behavior**

* User should not be required to select oidc or password when `BOUNDARY_AUTH_METHOD_ID` is set
* User should not be given an example with `ampw_1234567890` without any mention that this value cannot be used as-is
* User should not be instructed to put their password on the command line by a security application!!  The help example should demonstrate good behavior, like prompting for it

jefferai:
* Make it clear in the subcommands that the given strings are examples
* Add an OIDC example to show that there are different types
* Remove password flag from examples

Closes #1833


irenarindos (approved):
Looks good to me

hugoamvieira (approved):
LGTM 👍

PR #1841, opened 2022-02-03
bug(host): Fix sync job failure blocking other set syncs

louisruch:
During the plugin_host_set_sync job, a single set failing to sync
can cause all following syncs to fail.

### DIFF
--- a/internal/host/plugin/job_set_sync.go
+++ b/internal/host/plugin/job_set_sync.go
@@ -9,6 +9,7 @@ import (
 	"github.com/hashicorp/boundary/internal/errors"
 	"github.com/hashicorp/boundary/internal/kms"
 	"github.com/hashicorp/boundary/internal/libs/endpoint"
+	"github.com/hashicorp/boundary/internal/observability/event"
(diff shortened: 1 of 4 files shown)


louisruch:
@talanknight PTAL
This mainly adds a event.WriteError and continue instead of returning error after a failure during `ListHost` or `upsertAndCleanHosts`.

internal/host/plugin/job_set_sync.go
```diff
@@ -574,5 +547,14 @@ func (r *SetSyncJob) upsertAndCleanHosts(
 		}
 	}
 
+	// update last sync time on the sets
+	i, err := r.writer.Exec(ctx, updateSyncDataQuery, []interface{}{setIds})
```
louisruch, line 551: @talanknight lets chat about this, I was originally going to add this to the transaction for the set, but we don't actually do a transaction if there are no changes...

louisruch:
@talanknight @A-440Hz PTAL

talanknight (approved):
Looks like you cleaned up a lot of dead/duplicate code.  Thanks!  Looks great!

PR #1842, opened 2022-02-03
ci(proto): Detect unchecked API and proto changes

irenarindos:
Addition of workflow job, `make-gen-deltas`, to run `make gen` and detect diffs, in case proto changes are made but generated files are not checked in.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,6 +49,47 @@ jobs:
         command: |
           make test-sql POSTGRES_DOCKER_IMAGE_BASE=docker.mirror.hashicorp.services/postgres PG_DOCKER_TAG=11-alpine
         name: Run SQL PgTap Tests
+  make-gen-deltas:
+    machine:
+      image: ubuntu-2004:202111-01
+    resource_class: medium
+    working_directory: ~/boundary
+    steps:
+    - checkout
+    - run:
+        command: |
+          make install-go
+          source ~/.bashrc
+          echo 'export GOROOT=$GOROOT' >> "$BASH_ENV"
+          echo 'export GOPATH=$GOPATH' >> "$BASH_ENV"
+          echo 'export PATH=$PATH' >> "$BASH_ENV"
+          echo "$ go version"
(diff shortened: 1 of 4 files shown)


.circleci/config/jobs/proto-and-api-deltas.yml
```diff
@@ -0,0 +1,18 @@
...
+steps:
+- checkout
+- install-go
+- run:
+    name: "Check proto deltas"
+    command: |
+      make proto
+      rm -f diffFile
+      git diff > diffFile
+      [ -s diffFile ] && exit -1
```
tmessi, line 11: You can eliminate the file here by leveraging an option to git diff:

```
git diff --exit-code
```

From `git diff --help`:
```
--exit-code
    Make the program exit with codes similar to diff(1). That is, it exits with 1 if there were
    differences and 0 means no differences.
```

This also has a nice benefit of producing the full diff output, so if this does fail the CI output will provide something to compare against locally.

However, this will still return 0 in the case where untracked files are added. So for example, if a new protobuf file was added that would result in a new generated file, but the generated file was not added. To detect this, in the past I have used `git status --porcelain` which will have no output if there are no changes or untracked files. So a command like:

```
test -z $(git status --porcelain)
```
That will return a non-0 if the git status command has any output.

So to summarize all of that I would suggest:

```suggestion
      git diff --exit-code
      git status --porcelain
      test -z $(git status --porcelain)
```

tmessi (changes requested):

irenarindos:
Passed with main branch, no changes: https://app.circleci.com/pipelines/github/hashicorp/boundary/8479/workflows/2b8dfc1d-cb5a-4626-ab86-619cc202eab3/jobs/119523
Pushed code with a proto change and CI job failed as expected: https://app.circleci.com/pipelines/github/hashicorp/boundary/8481/workflows/a2a07803-3dc4-41cd-b603-7a8b881969f7/jobs/119537

tmessi (approved):

PR #1843, opened 2022-02-03
RDX 223 crt dockerversion parity

modrake:
This modifies the logic by which version strings are created in docker builds, for parity with the enterprise build change here: https://github.com/hashicorp/boundary-enterprise/pull/18


mdeggies (approved):
Thank you!

tmessi (approved):

PR #1846, opened 2022-02-09
Include addresses from the static host table in whx_network_address_dimension_source

talanknight:


internal/db/schema/migrations/oss/postgres/26/02_wh_network_address_dimensions.up.sql
```diff
@@ -131,18 +131,52 @@ begin;
...
+      sh.public_id as host_id,
+      sh.address as address,
+      'DNS Name' as address_type,
+      'Not Applicable' as ip_address_family,
+      'Not Applicable' private_ip_address_indicator,
+      sh.address as dns_name,
+      'Not Applicable' as ip4_address,
+      'Not Applicable' as ip6_address
+    from static_host as sh
+    where wh_try_cast_inet(sh.address) is null
```
mgaffney, line 157: Can we change `wh_try_cast_inet` to not use exceptions? The `wh_try_cast_inet` uses exceptions to determine if an address is an ip address or a dns address. Exception blocks are [expensive](https://www.postgresql.org/docs/current/plpgsql-control-structures.html#id-1.8.8.8.10.2) in plpgsql and this view uses the `wh_try_cast_inet` function 12 times. The query planner might be able to minimize calls but if it can't then `wh_try_cast_inet` could be called at least once for every row in the `static_host` table every time this view is used.

mgaffney (changes requested):

internal/db/schema/migrations/oss/postgres/26/02_wh_network_address_dimensions.up.sql
talanknight: That's a good question.  Maybe we can create a regex to try to confirm the text matches something that would be accepted by an inet cast before actually doing so, that might avoid entering the `begin/exception` block when the exception section is likely to be entered.

Thoughts?  Googling for ipv4 and ipv6 regex patterns brings up a wide variety of possible regexs.

mgaffney: +1 on using regex patterns

mgaffney: [This](http://bonesmoses.org/2016/07/01/pg-phriday-exceptional-performance/) article measured the cost of exceptions. In their test, adding an `EXCEPTION` block to their test function increased the execution time by 25%. And the exception block they added did nothing and would never actually be run. This may be a premature optimization but given how often this function will be called, I think it is justified.

talanknight: For posterity, Mike and Tim profiled variations on these functions with and without regex and exception blocks
The following is the result of different variations which resulted in us deciding on sticking with the original approach of a single cast and exception block:

mg_to_inet_1: Todd's original function that a cast and catches exceptions
mg_to_inet_2: Todd's second function that a uses regex and catches exceptions
mg_to_inet_3: Puts Todd's regexs into a case statement no exception block
mg_to_inet_4: Same as mg_to_inet_3 plus an additional when statement
```
# select mg_to_inet_1('x' || s.id) from generate_series(1, 100000) s(id);
Time: 573.238 ms
# select mg_to_inet_2('x' || s.id) from generate_series(1, 100000) s(id);
Time: 1316.799 ms (00:01.317)
# select mg_to_inet_3('x' || s.id) from generate_series(1, 100000) s(id);
Time: 1671.002 ms (00:01.671)
# select mg_to_inet_4('x' || s.id) from generate_series(1, 100000) s(id);
Time: 514.216 ms
```

tmessi (approved):

PR #1849, opened 2022-02-09
bug(plugin): Fix incorrect use of WithIpAddrs in Repository.Endpoints

### LINKED ISSUE #1848: [Bug] incorrect use of WithIpAddrs in plugin.Repository.Endpoints
**Describe the bug**

Dynamic AWS Host Catalog creates an error when trying to connect to targets which have DNS Names.

```
$ boundary connect ssh -target-id=ttcp_jSD3e4avYm
Error from controller when performing authorize-session action against given target

Error information:
  Kind:                Internal
  Message:             plugin.(Repository).Endpoints: endpoint.(preferencer).Choose: unknown, unknown: error #0: input 'ec2-52-42-134-187.us-west-2.compute.amazonaws.com' is
  not parseable as an ip address
  Status:              500
  context:             Error from controller when performing authorize-session action against given target
```

**To Reproduce**

1. Create a Dynamic Host Catalog and Host Set follow the guidance in [AWS Host Plugin](https://github.com/hashicorp/boundary-plugin-host-aws#getting-started). No special configuration required here as long as it can and register a host of type `plugin`
2. Create a tcp target `boundary targets create tcp --default-port=22 -scope-id=<SCOPE> -name=<NAME>`
3. Associate the host set created in step 1 to the target created in step 2: `boundary targets add-host-sets -host-set=<string>   -id=<string>`
4. Connect to the target created in step 3: `boundary connect ssh -target-id=ttcp_jSD3e4avYm`
5. Encounter the error

**Expected behavior**

A session should be created for the target host.

**Additional context**

The bug appears to be from here.
https://github.com/hashicorp/boundary/blob/ac78ee63151188083ae17c4eebb21d8d2117b57a/internal/host/plugin/repository_host_set.go#L819

It seems like it was a copy/paste error where the copied code was not altered to use the WithDnsNames option

justenwalker:
When creating the options for the preferencer, this used WithIpAddrs for both the host addresses and also DNS names. This causes errors when a host contains DNS names that are not IP addresses (ie: whenever dns names are present).

This change will:
- Replace WithIpAddrs with WithDNSNames option for DNS Names when constructing options for the preferencer.
- Add a test which includes a DNS name to prevent regressions

Fixes #1848

### DIFF
--- a/internal/host/plugin/repository_host_set_test.go
+++ b/internal/host/plugin/repository_host_set_test.go
@@ -1314,12 +1314,14 @@ func TestRepository_Endpoints(t *testing.T) {
 	hostSet10 := TestSet(t, conn, kms, sched, catalog, plgm, WithName("hostSet10"), WithPreferredEndpoints([]string{"cidr:10.0.0.1/24"}))
 	hostSet192 := TestSet(t, conn, kms, sched, catalog, plgm, WithName("hostSet192"), WithPreferredEndpoints([]string{"cidr:192.168.0.1/24"}))
 	hostSet100 := TestSet(t, conn, kms, sched, catalog, plgm, WithName("hostSet100"), WithPreferredEndpoints([]string{"cidr:100.100.100.100/24"}))
+	hostSetDNS := TestSet(t, conn, kms, sched, catalog, plgm, WithName("hostSetDNS"), WithPreferredEndpoints([]string{"dns:*"}))
 	hostlessSet := TestSet(t, conn, kms, sched, hostlessCatalog, plgm)
 
-	h1 := TestHost(t, conn, catalog.GetPublicId(), "test", withIpAddresses([]string{"10.0.0.5", "192.168.0.5"}))
+	h1 := TestHost(t, conn, catalog.GetPublicId(), "test", withIpAddresses([]string{"10.0.0.5", "192.168.0.5"}), withDnsNames([]string{"example.com"}))
 	TestSetMembers(t, conn, hostSet10.GetPublicId(), []*Host{h1})
 	TestSetMembers(t, conn, hostSet192.GetPublicId(), []*Host{h1})
 	TestSetMembers(t, conn, hostSet100.GetPublicId(), []*Host{h1})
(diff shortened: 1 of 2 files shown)


internal/host/plugin/repository_host_set_test.go
```diff
@@ -1321,6 +1321,10 @@ func TestRepository_Endpoints(t *testing.T) {
 	TestSetMembers(t, conn, hostSet192.GetPublicId(), []*Host{h1})
 	TestSetMembers(t, conn, hostSet100.GetPublicId(), []*Host{h1})
 
+	hostSetDNS := TestSet(t, conn, kms, sched, catalog, plgm, WithName("hostSetDNS"), WithPreferredEndpoints([]string{"cidr:5.4.0.0/16"}))
+	h2 := TestHost(t, conn, catalog.GetPublicId(), "dns", withIpAddresses([]string{"5.4.3.2"}), withDnsNames([]string{"example.com"}))
+	TestSetMembers(t, conn, hostSetDNS.GetPublicId(), []*Host{h2})
```
talanknight, line 1326: Thank you so much for identifying this gap in our testing.  I'd like to ensure we also verify the selection of the DNS address that might be on a host.
To make this new test case be closer to the other tests in this same file, instead of creating a new host here, could you add the `withDnsNames([]string{"example.com"})` to the host created on line 1319 and then move your hostset alongside the ones already created in lines 1314-1316 and have it uses `WithPreferredEndpoints([]string{"dns:*"})`?  That'll not only add coverage to the line that you fixed in this PR, but it'll also ensure that the DNS name on the host is properly selected and returned.

talanknight (changes requested):
This is great!  Thank you so much for identifying and fixing this bug as well as adding regression testing.  I have a suggestion for a very minor tweak regarding the test case.

justenwalker:
> This is great! Thank you so much for identifying and fixing this bug as well as adding regression testing. I have a suggestion for a very minor tweak regarding the test case.

I've applied your suggestions to this PR. Please have another look and let me know if there is anything else you think I should do.

talanknight (approved):
This looks great and ready to merge!  Thank you again for the issue and pull request.

Issue #1850, opened 2022-02-10
boundary connect's -host-id flag expects the `hst_` prefix instead of the actual `h_` one


psypuff:
**Describe the bug**
When using `boundary connect` with the `-host-id` flag in order to connect to a specific host using its `h_` prefixed ID, it fails with the following error:
```
Error information:
  Kind:                InvalidArgument
  Message:             Errors in provided fields.
  Status:              400
  context:             Error from controller when performing authorize-session action against given target

  Field-specific Errors:
    Name:              -host-id
      Error:           Incorrectly formatted identifier.
```
 Trying the same with the (former?) `hst_` prefixed ID, it passes validation but fails with another error:
```
Error information:
  Kind:                FailedPrecondition
  Message:             No workers are available to handle this session, or all have been filtered.
  Status:              400
  context:             Error from controller when performing authorize-session action against given target
```
Omitting the `-host-id` flag makes `boundary connect` work as expected.

**To Reproduce**
1. Create whatever resources it takes to have a SSHable host in Boundary
2. Fetch the host ID using `boundary host-sets read -id hs_<host_set_id>`
3. Run `boundary connect ssh -target-id ttcp_<target_id> -host-id h_<host_id>` 

**Expected behavior**
`boundary connect ssh -target-id ttcp_<target_id> -host-id h_<host_id>` should connect to the specified host

**Additional context**
Boundary version - 0.7.4
Connecting through Boundary Desktop (1.4.1) fails as well with a less verbose error:
```
Errors in provided fields.

Would you like to retry?
```

psypuff:
Thanks for the quick response here.
Please ignore the other error, we had actual issues with the worker.

PR #1853, opened 2022-02-10
fix(controller/targets): authorize-session should allow h_ ids

justenwalker:
The targets controller previously only allowed hosts from a static catalog
since it checked for the ID subtype from the static package.

This change validates host ids with the plugin.Subtype as well

Fixes: #1850


justenwalker:
> Thank you for the fix! Currently our target_service test doesn't cover specifying the host id when authorizing a session so I'll submit a seperate fast follow up PR to this one which adds regression tests.

Hi @talanknight -- I have an update coming which should add a regression test for this case - I can push it to this PR if you're interested

talanknight:
Adding that regression test would be wonderful.  Thank you!

justenwalker:
> Adding that regression test would be wonderful. Thank you!

Sure thing. Pushed for your review.

Issue #1857, opened 2022-02-15
[Bug] panic when executing kube commands with '--' in args list


justenwalker:
**Describe the bug**

When executing a `kube` command that requires `--` in the arguments like `kubectl exec`; the connect command panics

```
panic: runtime error: slice bounds out of range [7:1]

goroutine 1 [running]:
github.com/hashicorp/boundary/internal/cmd/commands/connect.(*Command).Run(0xc0001b1400, {0xc00003a0e0, 0x8, 0x8})
        /go/internal/cmd/commands/connect/connect.go:265 +0x2a65
github.com/mitchellh/cli.(*CLI).Run(0xc0000f6dc0)
        /root/go/pkg/mod/github.com/mitchellh/cli@v1.1.2/cli.go:262 +0x5f8
github.com/hashicorp/boundary/internal/cmd.RunCustom({0xc00003a0c0, 0x60, 0x0}, 0x0)
        /go/internal/cmd/main.go:186 +0x9d6
github.com/hashicorp/boundary/internal/cmd.Run(...)
        /go/internal/cmd/main.go:92
main.main()
        /go/cmd/boundary/main.go:13 +0xc9
```

**To Reproduce**

```sh
boundary connect kube -target-id=foo -- exec -n default bar -- /bin/sh
```

**Note**: The panic is so early that the target doesn't really need to exist

**Expected behavior**

No panic

**Additional Details**

```
 $ boundary version

Version information:
  Git Revision:        221acff4cc4d1f9be7619a657274c043999e62cc
  Version Number:      0.7.4
```

PR #1858, opened 2022-02-15
fix(connect): do not panic on multiple `--` args

justenwalker:
Passthrough args looks for the an instance of -- and re-slices the arg
list to pass through those args to the underlying connect exec command.
However such commands may also need to parse their own '--', such as
kubectl exec.

This change updates the logic so that boundary arg parser only considers
the first instance of '--' when deciding which args to pass to the
-exec= commands.

Fixes: #1857

### DIFF
--- a/internal/cmd/commands/connect/connect.go
+++ b/internal/cmd/commands/connect/connect.go
@@ -264,6 +264,7 @@ func (c *Command) Run(args []string) (retCode int) {
 		if v == "--" {
 			passthroughArgs = args[i+1:]
 			args = args[:i]
+			break // only consider the first instance of '--' in the args list
 		}
 	}
 


justenwalker:
**Note** Currently, this is the most minimal change to make this work.  
I'm not sure how often this sort of operation might be re-used, but it could also be extracted into a function with tests.

Something like this: https://go.dev/play/p/igILb-wPYe6

justenwalker:
@talanknight does this fix things in the right way? Anything I can update to make it better? ( could extract the function and add tests like in https://go.dev/play/p/igILb-wPYe6)

**Note**: I'm pretty sure there are no compatibility concerns since the way it is currently implemented guarantees a panic if multiple `--` are given; so no one could have a dependency on multiple `--` being parsed in a different way. (https://go.dev/play/p/TNK9qyxsor6).

talanknight (approved):
This is great! Thank you again. :)

PR #1870, opened 2022-02-17
[RDX-309] UI build timeout fix

claire-labry:
There’s a pesky UI build issue where `yarn install` times out with a network issue causing folks to have to re-run the build manually in the Github UI. This is kind of build failure is easy to miss and this PR is a fix for that. 

I'd like to merge in this fix + have the Boundary team notify our team (release engineering) if they're still running into this issue. This error is sporadic and hard to reproduce unfortunately. 😕

The error looks like this: 

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,8 +5,7 @@ on:
     # Sequence of patterns matched against refs/heads
     branches:    
       # Push events on main branch
-      # - 'main'
-      - 'yarn-timeout-fix'
+      - 'main'
 
 env:
   PKG_NAME: "boundary"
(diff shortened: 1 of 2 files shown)


tmessi (approved):
Nice, I *just* ran into this issue so hopefully this will fix it. I was also noticing that we run this step for every os/arch combination. I think it should be possible to do this step once and re-use the built assets for each os/arch. At least then if it does still fail, it happens early so a re-run doesn't re-do as much work. But that might require some more testing.

claire-labry:
@tmessi Thank you for your quick PR review! I originally added the flag to each step for every os/arch and you bring up a good point re: doing this step once. I'd like to bring @kpenfound for his thoughts just adding the flag once + reusing the built assets.

mdeggies:
We do have a backlog item to look into build job caching, but it's lower on the priority list. This looks like a good change to make until we can get to that.

mdeggies (approved):

PR #1874, opened 2022-02-18
refactor(controller/worker): Listener code improvements

hugoamvieira:
Improving controller/worker listener code:

- (Controller) Only start a single gRPC Server for all API Listener
purposes.
- (Both) Improve logic around handling multiple listeners with the same
purpose.
- (Both) Improve naming around objects and functions.
- (Controller) Decouple gRPC Server and gRPC Gateway Service 
Registration.
- (Both) Improve shutdown process.
- (Both) Improve test cases and coverage.
- (Both) First pass at centralizing configuration and validation
on Controller/Worker objects.

### DIFF
--- a/internal/servers/controller/controller.go
+++ b/internal/servers/controller/controller.go
@@ -337,15 +337,15 @@ func (c *Controller) registerSessionCleanupJob() error {
 	return nil
 }
 
-func (c *Controller) Shutdown(serversOnly bool) error {
+func (c *Controller) Shutdown() error {
 	const op = "controller.(Controller).Shutdown"
 	if !c.started.Load() {
 		event.WriteSysEvent(context.TODO(), op, "already shut down, skipping")
 	}
 	defer c.started.Store(false)
 	c.baseCancel()
-	if err := c.stopListeners(serversOnly); err != nil {
-		return fmt.Errorf("error stopping controller listeners: %w", err)
+	if err := c.stopServersAndListeners(); err != nil {
+		return fmt.Errorf("error stopping controller servers and listeners: %w", err)
 	}
 	c.schedulerWg.Wait()
 	c.tickerWg.Wait()
--- a/internal/servers/controller/controller.go
+++ b/internal/servers/controller/controller.go
@@ -52,6 +52,9 @@ type Controller struct {
 
 	workerAuthCache *sync.Map
 
+	apiListeners    []*base.ServerListener
+	clusterListener *base.ServerListener
+
 	// Used for testing and tracking worker health
 	workerStatusUpdateTimes *sync.Map
 
@@ -123,6 +126,32 @@ func New(ctx context.Context, conf *Config) (*Controller, error) {
 				err)
 		}
 	}
+
+	clusterListeners := make([]*base.ServerListener, 0)
+	for i := range conf.Listeners {
+		l := conf.Listeners[i]
+		if l == nil || l.Config == nil || l.Config.Purpose == nil {
+			continue
+		}
+		if len(l.Config.Purpose) != 1 {
+			return nil, fmt.Errorf("found listener with multiple purposes %q", strings.Join(l.Config.Purpose, ","))
+		}
+		switch l.Config.Purpose[0] {
+		case "api":
+			c.apiListeners = append(c.apiListeners, l)
+		case "cluster":
+			clusterListeners = append(clusterListeners, l)
+		}
+	}
+	if len(c.apiListeners) == 0 {
+		return nil, fmt.Errorf("no api listeners found")
+	}
+	if len(clusterListeners) != 1 {
+		// in the future, we might pick the cluster that is exposed to the outside
+		// instead of limiting it to one.
+		return nil, fmt.Errorf("exactly one cluster listener is required")
+	}
+	c.clusterListener = clusterListeners[0]
 
 	for _, enabledPlugin := range c.enabledPlugins {
 		switch enabledPlugin {
--- a/internal/servers/controller/controller.go
+++ b/internal/servers/controller/controller.go
@@ -7,8 +7,6 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
-
 	"github.com/hashicorp/boundary/internal/auth/oidc"
 	"github.com/hashicorp/boundary/internal/auth/password"
 	"github.com/hashicorp/boundary/internal/authtoken"
@@ -58,11 +56,9 @@ type Controller struct {
 	// Used for testing and tracking worker health
 	workerStatusUpdateTimes *sync.Map
 
-	// grpc gateway server
-	gatewayServer   *grpc.Server
-	gatewayTicket   string
-	gatewayListener gatewayListener
-	gatewayMux      *runtime.ServeMux
+	apiGrpcServer         *grpc.Server
+	apiGrpcServerListener grpcServerListener
+	apiGrpcGatewayTicket  string
 
 	// Repo factory methods
 	AuthTokenRepoFn       common.AuthTokenRepoFactory
@@ -279,7 +275,7 @@ func (c *Controller) Start() error {
 	if err := c.registerJobs(); err != nil {
 		return fmt.Errorf("error registering jobs: %w", err)
 	}
-	if err := c.startListeners(c.baseContext); err != nil {
+	if err := c.startListeners(); err != nil {
 		return fmt.Errorf("error starting controller listeners: %w", err)
 	}
 	if err := c.scheduler.Start(c.baseContext, c.schedulerWg); err != nil {
--- a/internal/servers/controller/controller_test.go
+++ b/internal/servers/controller/controller_test.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"testing"
 
+	"github.com/hashicorp/boundary/internal/cmd/base"
 	"github.com/hashicorp/boundary/internal/db"
 	"github.com/hashicorp/boundary/internal/kms"
+	"github.com/hashicorp/go-secure-stdlib/listenerutil"
 	"github.com/stretchr/testify/require"
 )
 
@@ -31,4 +33,154 @@ func TestController_New(t *testing.T) {
 		_, err = New(testCtx, conf)
 		require.NoError(err)
 	})
+}
+
+func TestControllerNewListenerConfig(t *testing.T) {
+	tests := []struct {
+		name       string
+		listeners  []*base.ServerListener
+		assertions func(t *testing.T, c *Controller)
+		expErr     bool
+		expErrMsg  string
+	}{
+		{
+			name: "valid listener configuration",
+			listeners: []*base.ServerListener{
+				{
+					Config: &listenerutil.ListenerConfig{
+						Purpose: []string{"api"},
+					},
+				},
+				{
+					Config: &listenerutil.ListenerConfig{
+						Purpose: []string{"api"},
+					},
+				},
+				{
+					Config: &listenerutil.ListenerConfig{
+						Purpose: []string{"cluster"},
+					},
+				},
+			},
+			assertions: func(t *testing.T, c *Controller) {
+				require.Len(t, c.apiListeners, 2)
+				require.NotNil(t, c.clusterListener)
+			},
+		},
+		{
+			name:      "listeners are required",
+			listeners: []*base.ServerListener{},
+			expErr:    true,
+			expErrMsg: "no api listeners found",
+		},
+		{
+			name:      "listeners are required - not nil",
+			listeners: []*base.ServerListener{nil, nil},
+			expErr:    true,
+			expErrMsg: "no api listeners found",
+		},
+		{
+			name:      "listeners are required - with config",
+			listeners: []*base.ServerListener{{}, {}},
+			expErr:    true,
+			expErrMsg: "no api listeners found",
+		},
+		{
+			name: "listeners are required - with purposes",
+			listeners: []*base.ServerListener{
+				{
+					Config: &listenerutil.ListenerConfig{Purpose: nil},
+				},
+				{
+					Config: &listenerutil.ListenerConfig{Purpose: nil},
+				},
+			},
+			expErr:    true,
+			expErrMsg: "no api listeners found",
+		},
+		{
+			name: "both api and cluster listeners are required",
+			listeners: []*base.ServerListener{
+				{
+					Config: &listenerutil.ListenerConfig{
+						Purpose: []string{"api"},
+					},
+				},
+			},
+			expErr:    true,
+			expErrMsg: "exactly one cluster listener is required",
+		},
+		{
+			name: "both api and cluster listeners are required 2",
+			listeners: []*base.ServerListener{
+				{
+					Config: &listenerutil.ListenerConfig{
+						Purpose: []string{"cluster"},
+					},
+				},
+			},
+			expErr:    true,
+			expErrMsg: "no api listeners found",
+		},
+		{
+			name: "only one cluster listener is allowed",
+			listeners: []*base.ServerListener{
+				{
+					Config: &listenerutil.ListenerConfig{
+						Purpose: []string{"api"},
+					},
+				},
+				{
+					Config: &listenerutil.ListenerConfig{
+						Purpose: []string{"cluster"},
+					},
+				},
+				{
+					Config: &listenerutil.ListenerConfig{
+						Purpose: []string{"cluster"},
+					},
+				},
+			},
+			expErr:    true,
+			expErrMsg: "exactly one cluster listener is required",
+		},
+		{
+			name: "only one purpose is allowed per listener",
+			listeners: []*base.ServerListener{
+				{
+					Config: &listenerutil.ListenerConfig{
+						Purpose: []string{"api", "cluster"},
+					},
+				},
+			},
+			expErr:    true,
+			expErrMsg: `found listener with multiple purposes "api,cluster"`,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx, cancel := context.WithCancel(context.Background())
+
+			tc := &TestController{
+				t:      t,
+				ctx:    ctx,
+				cancel: cancel,
+				opts:   nil,
+			}
+			conf := TestControllerConfig(t, ctx, tc, nil)
+			conf.Listeners = tt.listeners
+
+			c, err := New(ctx, conf)
+			if tt.expErr {
+				require.EqualError(t, err, tt.expErrMsg)
+				require.Nil(t, c)
+				return
+			}
+
+			require.NoError(t, err)
+			require.NotNil(t, c)
+			tt.assertions(t, c)
+		})
+	}
 }
--- a/internal/servers/controller/cors_test.go
+++ b/internal/servers/controller/cors_test.go
@@ -20,50 +20,54 @@ const corsTestConfig = `
 disable_mlock = true
 
 telemetry {
-        prometheus_retention_time = "24h"
-        disable_hostname = true
+	prometheus_retention_time = "24h"
+	disable_hostname = true
 }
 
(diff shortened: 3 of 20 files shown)


internal/servers/controller/handler.go
```diff
@@ -68,99 +68,70 @@ func (c *Controller) handler(props HandlerProperties) (http.Handler, error) {
...
+	return eventsHandler, err
 }
 
-func handleGrpcGateway(c *Controller, props HandlerProperties) (http.Handler, error) {
-	// Register*ServiceHandlerServer methods ignore the passed in ctx. Using it
-	// now however in case this changes in the future.
-	ctx := props.CancelCtx
-	currentServices := c.gatewayServer.GetServiceInfo()
-	dialOptions := gatewayDialOptions(c.gatewayListener)
+func (c *Controller) registerGrpcServices(ctx context.Context, s *grpc.Server) (*grpc.Server, error) {
```
hugoamvieira, line 75: TBD: We're receiving the gRPC Server `s` as a reference and then using it directly in the Register calls below. At the end, we're returning the object back to the caller. This isn't a great flow as it makes the caller think that what they call with (`s`) isn't going to be mutated, when in fact it is.

Solutions:
1. Don't return gRPC Server back. This makes it a bit clearer to the caller that we're mutating what they passed in, and because these are private functions, I think there should be the assumption that the caller should be aware of the inner workings of this function (to an extent).
2. Deep copy of `s`, and then use that into the Register calls and return that object. This is an interesting approach, but can be complex considering we're working with an object that already has state and has pointer fields within it.

Comments welcome here, thanks! 🙇

internal/servers/controller/handler.go
```diff
@@ -176,73 +147,104 @@ func handleGrpcGateway(c *Controller, props HandlerProperties) (http.Handler, er
...
-		if err := services.RegisterCredentialLibraryServiceHandlerFromEndpoint(ctx, c.gatewayMux, gatewayTarget, dialOptions); err != nil {
-			return nil, fmt.Errorf("failed to register credential library service handler: %w", err)
-		}
+		services.RegisterCredentialLibraryServiceServer(s, cl)
+	}
+
+	return s, nil
+}
+
+func registerGrpcGatewayEndpoints(ctx context.Context, gwMux *runtime.ServeMux, dialOptions ...grpc.DialOption) (*runtime.ServeMux, error) {
```
hugoamvieira, line 198: https://github.com/hashicorp/boundary/pull/1874/files#r811054909 same situation

internal/servers/controller/controller.go
```diff
@@ -55,11 +53,10 @@ type Controller struct {
 	// Used for testing and tracking worker health
 	workerStatusUpdateTimes *sync.Map
 
-	// grpc gateway server
+	// TODO: Rename.
 	gatewayServer   *grpc.Server
+	gatewayListener grpcServerListener
 	gatewayTicket   string
```
hugoamvieira, line 59: TODO: Deal with these (Either rename to match what the objects actually are, or remove depending on final implementation)

hugoamvieira, line 59: TODO: Deal with these (Either rename to match what the objects actually are, or remove depending on final implementation)

internal/servers/controller/listeners.go
```diff
@@ -168,6 +83,104 @@ func (c *Controller) startListeners(ctx context.Context) error {
...
+		apiServers = append(apiServers, func() { go server.Serve(l) })
+
+	default:
+		protos := []string{"", "http/1.1", "h2"}
+		for _, v := range protos {
+			l := ln.Mux.GetListener(v)
+			if l == nil {
+				return nil, fmt.Errorf("could not get tls proto %q listener", v)
+			}
+			apiServers = append(apiServers, func() { go server.Serve(l) })
```
hugoamvieira, line 139: This is the problematic bit: We have multiple listeners - Therefore I wasn't able to use the same pattern here as I did on [`configureForCluster`](https://github.com/hashicorp/boundary/pull/1874/files#diff-954e2ccd0cd7c7d7991745a59b951b9149a750bdf8d18ee04e88be3e2524310fR70) (not returning anything and mutating the listener object and then using the new objects within it to start the server).

Possible solutions:
1. We change `base.ServerListener`'s `ALPNListener` field to be a slice of `net.Listener` instead of just one.
2. We add a new field to `base.ServerListener` which would be a slice of funcs to be used to startup that particular listener.

Thoughts very much welcome!

internal/servers/controller/listeners.go
```diff
@@ -168,6 +83,104 @@ func (c *Controller) startListeners(ctx context.Context) error {
...
+		),
+	)
+
+	workerService := workers.NewWorkerServiceServer(c.ServersRepoFn, c.SessionRepoFn, c.workerStatusUpdateTimes, c.kms)
+	pbs.RegisterServerCoordinationServiceServer(workerServer, workerService)
+	pbs.RegisterSessionServiceServer(workerServer, workerService)
+
+	interceptor := newInterceptingListener(c, l)
+	ln.ALPNListener = interceptor
+	ln.GrpcServer = workerServer
```
hugoamvieira, line 179: I've found these listeners to be a great place to store this state. They provide some key benefits:
- Means we don't have to return something like `[]func()` in this func;
- We can easily correlate the listener config and the server/listener that was created and ultimately will be started to serve it. This is important to support our future efforts of starting these listeners more intelligently.
- Conceptually speaking, they seem like the correct place to be storing this state in.
- If everything is stored centrally as it would be this way, shutting the servers down becomes easy.

But there's some problems with this that need to be addressed if we're going this way. More details [here](https://github.com/hashicorp/boundary/pull/1874/files#r811193116)

internal/servers/controller/controller.go
hugoamvieira: Addressed in [e31aa95](https://github.com/hashicorp/boundary/pull/1874/commits/e31aa9574ad44efa380d4cad5922d694ff4e8d0d)

internal/servers/controller/handler.go
hugoamvieira: Went with 1. Addressed in [c02c198](https://github.com/hashicorp/boundary/pull/1874/commits/c02c1980037a178a6bf6acd05b63ae11dc568d09)

internal/servers/controller/handler.go
hugoamvieira: Also went with 1. Addressed in [64a5d98](https://github.com/hashicorp/boundary/pull/1874/commits/64a5d98342d9db493fdcc167593d6a1397f2ca98)

internal/servers/controller/listeners.go
```diff
@@ -19,20 +19,41 @@ import (
...
+		c.gatewayServer, err = c.registerGrpcServices(ctx, grpcServer)
+		if err != nil {
+			return err
+		}
+
+		c.gatewayListener, _ = newGrpcServerListener()
+		servers = append(servers, func() {
+			go c.gatewayServer.Serve(c.gatewayListener)
+		})
+	}
```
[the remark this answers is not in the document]
hugoamvieira: Resolving as me and @tmessi discussed a better way of doing this

internal/servers/controller/controller.go
```diff
@@ -124,6 +123,25 @@ func New(ctx context.Context, conf *Config) (*Controller, error) {
...
+		}
+	}
+	if len(c.apiListeners) == 0 {
+		return nil, fmt.Errorf("no api listeners found")
+	}
+	if len(c.clusterListeners) == 0 {
+		return nil, fmt.Errorf("no cluster listeners found")
+	}
+	// TBD: Are we supposed to only allow one cluster listener?
+
```
hugoamvieira, line 144: Are we supposed to only allow one cluster listener? Boundary doesn't seem to support multiple cluster listeners.

internal/servers/controller/controller.go
```diff
@@ -124,6 +123,24 @@ func New(ctx context.Context, conf *Config) (*Controller, error) {
...
+		case "cluster":
+			c.clusterListeners = append(c.clusterListeners, l)
+		}
+	}
+	if len(c.apiListeners) == 0 {
+		return nil, fmt.Errorf("no api listeners found")
+	}
+	if len(c.clusterListeners) == 0 {
+		return nil, fmt.Errorf("no cluster listeners found")
+	}
```
tmessi, line 142: should this check for one and only one?

internal/servers/controller/controller_test.go
```diff
@@ -32,3 +34,94 @@ func TestController_New(t *testing.T) {
...
+			listeners: []*base.ServerListener{
+				{
+					Config: &listenerutil.ListenerConfig{
+						Purpose: []string{"api"},
+					},
+				},
+			},
+			expErr:    true,
+			expErrMsg: "no cluster listeners found",
+		},
```
tmessi, line 87: Can you add another test condition similar to this one, where there is only a `cluster` purpose provided and no `api`.

internal/servers/controller/controller_test.go
```diff
@@ -32,3 +34,94 @@ func TestController_New(t *testing.T) {
...
+		{
+			name: "only one purpose is allowed per listener",
+			listeners: []*base.ServerListener{
+				{
+					Config: &listenerutil.ListenerConfig{
+						Purpose: []string{"one", "two"},
```
tmessi, line 93: Can you make these both valid purposes? i.e. `[]string{"api", "cluster"}`?

internal/servers/controller/cors_test.go
```diff
@@ -38,6 +38,10 @@ kms "aead" {
 		key_id = "global_worker-auth"
 }
 
+listener "tcp" {
+		purpose = "cluster"
```
tmessi, line 42: This indentation seems off compared to the other blocks here.

internal/servers/controller/interceptor.go
```diff
@@ -230,9 +230,7 @@ func isNil(i interface{}) bool {
 	return false
 }
 
-func auditRequestInterceptor(
-	_ context.Context,
-) grpc.UnaryServerInterceptor {
+func auditRequestInterceptor() grpc.UnaryServerInterceptor {
```
tmessi, line 233: I would have these still take `_ context.Context` so they are consistent with the other interceptor functions

tmessi (changes requested):

internal/servers/controller/gateway.go
```diff
@@ -51,15 +45,21 @@ func newGatewayMux() *runtime.ServeMux {
 	)
 }
 
-func newGatewayServer(
+// newGrpcServerListener will create an in-memory listener for the gRPC server.
+func newGrpcServerListener() (grpcServerListener, string) {
+	buffer := globals.DefaultMaxRequestSize // seems like a reasonable size for the ring buffer, but then happily change the size if more info becomes available
+	return bufconn.Listen(int(buffer)), ""
```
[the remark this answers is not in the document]
hugoamvieira: Good question - it was like that and I didn't notice 😂 - Addressed in [028e9d5](https://github.com/hashicorp/boundary/pull/1874/commits/028e9d5c1cea7caf826a0e33f3e4caca02e30b95)

internal/servers/controller/controller.go
```diff
@@ -52,14 +50,15 @@ type Controller struct {
...
 	workerStatusUpdateTimes *sync.Map
 
-	// grpc gateway server
-	gatewayServer   *grpc.Server
-	gatewayTicket   string
-	gatewayListener gatewayListener
-	gatewayMux      *runtime.ServeMux
+	grpcServer         *grpc.Server
+	grpcServerListener grpcServerListener
+	grpcGatewayTicket  string
```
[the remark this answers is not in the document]
hugoamvieira: Yep, good idea - Added in [028e9d5](https://github.com/hashicorp/boundary/pull/1874/commits/028e9d5c1cea7caf826a0e33f3e4caca02e30b95)

internal/servers/controller/controller.go
hugoamvieira: Yeah, this is still outstanding and linked to https://github.com/hashicorp/boundary/pull/1874#pullrequestreview-899415126. Currently waiting on someone to shed light on whether we're supposed to have multiple clusters or not - perhaps I should assume in this PR that it'd just ever be one cluster and go from there?

In which case, `clusterListeners` really becomes `clusterListener`

internal/servers/controller/cors_test.go
hugoamvieira: Could you share what you're seeing on your end just to make sure? The only difference my editor reports is that I actually used tabs there and not spaces.

internal/servers/controller/controller_test.go
hugoamvieira: Yep, added in [37603e0](https://github.com/hashicorp/boundary/pull/1874/commits/37603e0728145592d5163ea9cd3c2b9f926e6e47)

internal/servers/controller/controller_test.go
hugoamvieira: Good idea, changed in [37603e0](https://github.com/hashicorp/boundary/pull/1874/commits/37603e0728145592d5163ea9cd3c2b9f926e6e47)

internal/servers/controller/cors_test.go
tmessi: Ok that is probably why. It looks like there is a mix of spaces and tabs within the raw string. We should probably just use one or the other here, but not an issue that needs to be solved now.

internal/servers/controller/interceptor.go
hugoamvieira: Addressed in [1063c66](https://github.com/hashicorp/boundary/pull/1874/commits/1063c6668fd1da5c1484d248c8d0cefead6a9078)

internal/servers/controller/controller.go
hugoamvieira: Assuming for now that only one cluster listener is allowed, considering we've seen examples in the codebase that point towards this direction

internal/servers/controller/controller.go
hugoamvieira: https://github.com/hashicorp/boundary/pull/1874#discussion_r819602594

internal/servers/controller/listeners.go
hugoamvieira: Considering we only need access to the `Serve` functions ephemerally, I decided to go with returning functions.

internal/servers/controller/listeners.go
hugoamvieira: https://github.com/hashicorp/boundary/pull/1874#discussion_r819607622

internal/servers/worker/listeners.go
```diff
@@ -100,38 +46,120 @@ func (w *Worker) startListeners() error {
...
+func (w *Worker) stopServersAndListeners() error {
+	var eg errgroup.Group
+
+	eg.Go(w.stopHttpServersAndListeners)
+	eg.Go(w.stopAnyListeners)
+
+	return eg.Wait()
+}
+
+// TODO(hugo): add tests
```
hugoamvieira, line 105: TODO

internal/servers/worker/listeners.go
```diff
@@ -100,38 +46,120 @@ func (w *Worker) startListeners() error {
...
+		}
+	}
+
+	return closeErrors.ErrorOrNil()
+}
+
+// stopAnyListeners does a final once over the known
+// listeners to make sure we didn't miss any;
+// expected to run at the end of stopServersAndListeners.
+// TODO(hugo): add tests
```
hugoamvieira, line 131: TODO

internal/servers/worker/worker.go
```diff
@@ -45,19 +46,18 @@ type Worker struct {
...
 	// request. It can be set via startup in New below, or (eventually) via
 	// SIGHUP.
 	updateTags ua.Bool
-
-	// Test-related values
-	testReuseAuthNonces bool
-	testReusedAuthNonce string
 }
 
+// TODO(hugo) test addition of `listeners`
```
hugoamvieira, line 60: TODO

tmessi (changes requested):
couple comments, mostly style questions, otherwise looking good

internal/servers/controller/controller.go
```diff
@@ -124,7 +124,8 @@ func New(ctx context.Context, conf *Config) (*Controller, error) {
 	}
 
 	clusterListeners := make([]*base.ServerListener, 0)
-	for _, l := range conf.Listeners {
+	for i := range conf.Listeners {
+		l := conf.Listeners[i]
```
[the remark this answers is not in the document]
hugoamvieira: We're iterating like this quite a bit in this flow, but honestly it's mostly because I keep forgetting when I _need_ to do this to iterate over a slice of pointers 😂 
I don't think it's needed in here though.

internal/servers/controller/listeners.go
```diff
@@ -159,56 +160,103 @@ func (c *Controller) configureForCluster(ln *base.ServerListener) (func(), error
...
-			}()
+func (c *Controller) stopServersAndListeners() error {
+	var eg errgroup.Group
+
+	eg.Go(c.stopClusterGrpcServerAndListener)
+	eg.Go(c.stopHttpServersAndListeners)
+	eg.Go(c.stopApiGrpcServerAndListener)
+	eg.Go(c.stopAnyListeners)
+
+	return eg.Wait()
```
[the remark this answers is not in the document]
hugoamvieira: @tmessi hm, that's a good point.

At this point, we could fail to stop any servers or listeners. As the process is on it's way out, _I think_ everything would terminate even if there had been a failure here when the process actually exits (although this could be OS dependent - not very familiar with how process lifecycle management works across OSes)

Regardless, now that you mention this, I think we should return _any and all_ errors for completeness, if anything. I'll change the code a bit.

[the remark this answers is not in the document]
hugoamvieira: Addressed in [4580b3d](https://github.com/hashicorp/boundary/pull/1874/commits/4580b3df019b925ae947c0df7d40f5c077208a95)

internal/servers/worker/worker.go
hugoamvieira: [4580b3d](https://github.com/hashicorp/boundary/pull/1874/commits/4580b3df019b925ae947c0df7d40f5c077208a95)

internal/servers/worker/listeners.go
hugoamvieira: [4580b3d](https://github.com/hashicorp/boundary/pull/1874/commits/4580b3df019b925ae947c0df7d40f5c077208a95)

internal/servers/worker/listeners.go
hugoamvieira: [4580b3d](https://github.com/hashicorp/boundary/pull/1874/commits/4580b3df019b925ae947c0df7d40f5c077208a95)

tmessi (dismissed):

CHANGELOG.md
```diff
@@ -27,6 +27,12 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 * hosts: Static hosts now include the host sets they are in.
   ([PR](https://github.com/hashicorp/boundary/pull/1828))
 
+### Bug Fixes
+
+* controller: Fix creating and starting one Controller gRPC Server
```
jefferai, line 32: This isn't user facing and doesn't really affect anything non-internal so shouldn't go in the changelog.

internal/tests/cluster/worker_replay_test.go
```diff
@@ -1,64 +0,0 @@
-package cluster
```
jefferai, line 1: This is an important test that we need to keep.

jefferai (changes requested):
This is a very, very large PR and it's not clear why many of the changes are being made. Many changes seem to be simply moving logic around or renaming things, which in turn makes it much harder to figure out the functional differences.

This needs to be split into more digestible chunks so that it's clear what changes are happening and why.

internal/tests/cluster/worker_replay_test.go
hugoamvieira: I have brought this test back but made it work with our changes

PR #1875, opened 2022-02-23
Fix yarn caching and set network-timeout to 5 mins

sarahethompson:
Fixed yarn caching in the build workflow and upped network timeout to 5 mins when doing a yarn install (the default seems to be 30 secs) in case of yarn network errors when the cache is invalidated.


claire-labry (approved):
Thank you so much for doing this @sarahethompson!

mdeggies (approved):

tmessi (approved):

PR #1877, opened 2022-02-24
Remove setup git CI step

mdeggies:
This isn't needed for non-enterprise projects. Thanks to @claire-labry for pointing this out!!


claire-labry (approved):
:shipit:

PR #1879, opened 2022-02-25
refact(session connections): Create session connection repository

irenarindos:
- Create a new repository, connection_repository, for Session Connections in the Session package.
- Have repository_connection functions use the new Session Connection repo
- Rename sessionCleanupJob to sessionConnectionsCleanupJob for clarity.
- Move AuthorizeConnection from repository_session.go into repository_connection.go.
- Creation of domain service, AuthorizeConnection()
- authorizeConnectionCte refactored so that constraints are more explicit.

### DIFF
--- a/internal/session/repository_session_test.go
+++ b/internal/session/repository_session_test.go
@@ -541,108 +541,16 @@ func TestRepository_updateState(t *testing.T) {
 	}
 }
 
-func TestRepository_AuthorizeConnect(t *testing.T) {
-	t.Parallel()
-	conn, _ := db.TestSetup(t, "postgres")
-	rw := db.New(conn)
-	wrapper := db.TestWrapper(t)
-	iamRepo := iam.TestRepo(t, conn, wrapper)
-	kms := kms.TestKms(t, conn, wrapper)
-	repo, err := NewRepository(rw, rw, kms)
-	require.NoError(t, err)
-
-	var testServer string
-	setupFn := func(exp *timestamp.Timestamp) *Session {
-		composedOf := TestSessionParams(t, conn, wrapper, iamRepo)
-		if exp != nil {
-			composedOf.ExpirationTime = exp
-		}
-		s := TestSession(t, conn, wrapper, composedOf)
-		srv := TestWorker(t, conn, wrapper)
-		testServer = srv.PrivateId
-		tofu := TestTofu(t)
-		_, _, err := repo.ActivateSession(context.Background(), s.PublicId, s.Version, srv.PrivateId, srv.Type, tofu)
-		require.NoError(t, err)
-		return s
-	}
-	testSession := setupFn(nil)
-
-	tests := []struct {
-		name          string
-		session       *Session
-		wantErr       bool
-		wantIsError   error
-		wantAuthzInfo ConnectionAuthzSummary
-	}{
-		{
-			name:    "valid",
-			session: testSession,
-			wantAuthzInfo: ConnectionAuthzSummary{
-				ConnectionLimit:        1,
-				CurrentConnectionCount: 1,
-				ExpirationTime:         testSession.ExpirationTime,
-			},
-		},
-		{
-			name: "empty-sessionId",
-			session: func() *Session {
-				s := AllocSession()
-				return &s
-			}(),
-			wantErr: true,
-		},
-		{
-			name: "exceeded-connection-limit",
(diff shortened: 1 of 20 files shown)


internal/session/connection_repository.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"context"
+
+	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/kms"
+)
+
+// Repository is the session database repository
```
tmessi, line 11: This comment should start with `ConnectionRepository`, and the rest of the comment seems wrong. I think it should be something like:

```go
// ConnectionRepository is the session connection database repository.
```

internal/session/connection_repository.go
```diff
@@ -0,0 +1,69 @@
...
+type ConnectionRepository struct {
+	reader db.Reader
+	writer db.Writer
+	kms    *kms.Kms
+
+	// defaultLimit provides a default for limiting the number of results returned from the repo
+	defaultLimit int
+}
+
+// NewRepository creates a new session Connection Repository. Supports the options: WithLimit
```
tmessi, line 21: this should start with `NewConnectionRepository`

internal/session/connection_repository.go
```diff
@@ -0,0 +1,69 @@
...
+// which sets a default limit on results returned by repo operations.
+func NewConnectionRepository(r db.Reader, w db.Writer, kms *kms.Kms, opt ...Option) (*ConnectionRepository, error) {
+	const op = "sessionConnection.NewRepository"
+	if r == nil {
+		return nil, errors.NewDeprecated(errors.InvalidParameter, op, "nil reader")
```
tmessi, line 26: it would be nice to not create new errors using `NewDeprecated`, do to so I think requires that a `context.Context` is passed in here, which would rippled back up into the repo factory functions in other packages, but I think that should be do-able as part of this.

tmessi (changes requested):

internal/session/repository_session_test.go
```diff
@@ -543,12 +543,14 @@ func TestRepository_updateState(t *testing.T) {
 
 func TestRepository_AuthorizeConnect(t *testing.T) {
```
tmessi, line 544: These tests should probably be moved to `repository_connection_test.go` or perhaps `service_authorize_connetion_test.go`. The test name should probably be updated as will from `TestRepository_AuthorizeConnect` to `TestService_AuthorizeConnection`

internal/session/service_authorize_connection.go
```diff
@@ -0,0 +1,27 @@
+package session
+
+import (
+	"context"
+
+	"github.com/hashicorp/boundary/internal/errors"
+)
+
+// AuthorizeConnection will check to see if a connection is allowed.
```
tmessi, line 9: I would start this comment off with:

```
// AuthorizeConnection is a domain service function that ....
```

I think this is a good place to describe the checks that are performed in determining if a Connection is authorized.

internal/session/connection_repository.go
```diff
@@ -0,0 +1,69 @@
+package session
```
tmessi, line 1: I would place the contents of this file in `repository_connection.go`

tmessi (changes requested):

internal/session/service_authorize_connection.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"context"
+
+	"github.com/hashicorp/boundary/internal/errors"
+)
+
+// AuthorizeConnection is a domain service function that will:
+// * create a connection, if the session has not reached its connection limit and the session is not expired and is still active
+// * Return the connection, its states, and an AuthzSummary.
+// If authorization fails, it returns an error with Code InvalidSessionState.
```
tmessi, line 12: maybe something like this to highlight the conditions it uses?
```suggestion
// AuthorizeConnection is a domain service function that will create a Connection
// for a session if the following criteria are met:
// * The session is active.
// * The session is not expired.
// * The session has not reached its connection limit or has a connection limit of -1.
// If any of these criteria is not met, it returns an error with Code InvalidSessionState.

tmessi (commented):
looking good, a couple more minor things

tmessi (approved):

PR #1882, opened 2022-03-02
feat(controller): Add health endpoint

hugoamvieira:
New health endpoint for the controller API (on `/health`) to enable health reporting to other automated systems. (eg: Load Balancers)


internal/servers/controller/handlers/health/health_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"testing"
+
+	"github.com/hashicorp/boundary/internal/gen/controller/ops/services"
+	"github.com/hashicorp/boundary/sdk/pbs/controller/ops/resources/health"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+func TestGetHealth(t *testing.T) {
```
talanknight, line 15: This appears to only test the grpc response.  Consider writing a test against the http server similar to the TestAuthenticationHandler in handler_test.go.  That way you can confirm http status codes along with the response body.

hugoamvieira: Thanks for the comment! Yes, this only tests the gRPC side of things. There will be some middleware that I will be writing to set the correct HTTP status code based on the gPRC message, but that's not in yet.

internal/cmd/config/config.go
```diff
@@ -128,6 +133,12 @@ type Controller struct {
 	AuthTokenTimeToStale         interface{} `hcl:"auth_token_time_to_stale"`
 	AuthTokenTimeToStaleDuration time.Duration
 
+	// OpsListenersGracePeriod is the amount of time that we'll wait before actually
+	// starting the Controller shutdown. This is used to allow the health endpoint
+	// to return a bad status code for a little while before the instance goes away.
+	OpsListenersGracePeriod         interface{} `hcl:"ops_listeners_grace_period"`
+	OpsListenersGracePeriodDuration time.Duration
```
hugoamvieira, line 140: TBD: Default value? We don't seem to do those anywhere

internal/servers/controller/handlers/health/health_test.go
hugoamvieira: Added in [6fad34e](https://github.com/hashicorp/boundary/pull/1882/commits/6fad34e659e7280de1dfad9be839f6e562447c1a)

internal/cmd/config/config.go
hugoamvieira: We decided the zero value makes sense here

internal/servers/controller/controller.go
```diff
@@ -90,6 +94,7 @@ func New(ctx context.Context, conf *Config) (*Controller, error) {
 		workerAuthCache:         new(sync.Map),
 		workerStatusUpdateTimes: new(sync.Map),
 		enabledPlugins:          conf.Server.EnabledPlugins,
+		apiListeners:            make([]*base.ServerListener, 0),
 	}
```
johanbrandhorst, line 98: Should we default `StartServiceUnavailableReplies` to something safe here, so that calling it without registering the health service doesn't panic?

internal/proto/local/controller/ops/services/v1/health_service.proto
```diff
@@ -0,0 +1,20 @@
...
+
+import "google/api/annotations.proto";
+import "controller/ops/resources/health/v1/health.proto";
+
+service HealthService {
+  rpc GetHealth(GetHealthRequest) returns (GetHealthResponse) {
+    option (google.api.http) = {
+      get : "/health"
+    };
+  }
```
johanbrandhorst, line 15: If we wanted to, we could add some extra metadata about expected responses to our generated swagger file by adding annotations here. See https://github.com/grpc-ecosystem/grpc-gateway/blob/c20d40b67e395dbd6f8810d462c6909bcfca4fa4/examples/internal/proto/examplepb/a_bit_of_everything.proto#L593-L604 for an example. I assume we either return 200 or 503? I don't know if that sort of thing is something we'd be interested in but I thought it worth mentioning 😁.

website/content/docs/configuration/controller.mdx
```diff
@@ -61,15 +61,23 @@ description will be read.
 - `auth_token_time_to_live` - Maximum time to live (TTL) for all auth tokens globally (pertains
   to all tokens from all auth methods). Valid time units are anything specified by Golang's
   [ParseDuration()](https://golang.org/pkg/time/#ParseDuration) method. Default is 7 days.
+
 - `auth_token_time_to_stale` - Maximum time of inactivity for all auth tokens globally (pertains
   to all tokens from all auth methods). Valid time units are anything specified by Golang's
   [ParseDuration()](https://golang.org/pkg/time/#ParseDuration) method. Default is 1 day.
 
+- `health_grace_period` - Amount of time Boundary will wait before initiating the shutdown procedure,
```
johanbrandhorst, line 69: This may already have been widely discussed and decided, if so, ignore me, but my initial impression upon reading this configuration variable name was that it had something to do with how often the health value would update or something. What do you think about sticking a `shutdown` somewhere into this variable? `shutdown_health_grace_period`?

internal/servers/controller/controller.go
hugoamvieira: As part of the startup process (in `startListeners`), this handler is always set, but yeah let's set that to something just in case.

internal/proto/local/controller/ops/services/v1/health_service.proto
hugoamvieira: I played with this but couldn't get swagger to work properly - It kept returning a broken definition for the second `response` object - I'll try this example 👍

website/content/docs/configuration/controller.mdx
hugoamvieira: This var is actually controlling how long Boundary will wait before shutdown - having it as `shutdown_health_grace_period` or `health_shutdown_grace_period` seems a bit misleading. At least for me, it leads to thinking about the particular health endpoint and how this var might control something _within_ that endpoint.

Maybe we could have this just be `shutdown_grace_period`. Considering that this var is exposed within a `controller` block and it'll affect the controller. This still seems misleading though - this var will actually prevent _any_ Boundary component from shutdown on the entire instance. Perhaps the best thing to do here is call it `shutdown_grace_period` but have it defined in top level context? I'm not sure.

WDYT?

internal/proto/local/controller/ops/services/v1/health_service.proto
hugoamvieira: Added in [db3bd22](https://github.com/hashicorp/boundary/pull/1882/commits/db3bd222f93a84977c791a4b2b86c1c39e9e3087) but it looks like our structure would add the swagger information [here](https://www.boundaryproject.io/api-docs/) which isn't ideal because health isn't part of the API.

website/content/docs/configuration/controller.mdx
johanbrandhorst: I see, yeah it risks being confusing no matter how you slice it I think - this is essentially a `wait_this_long_until_actually_shutting_down` right? What is the important part to convey to the user when they configure this? Why do they want to configure it?

internal/proto/local/controller/ops/services/v1/health_service.proto
johanbrandhorst: If we don't want it to be part of the `controller.swagger.json` we can generate it to a separate file, or maybe we don't need to generate this swagger documentation at all?

website/content/docs/configuration/controller.mdx
hugoamvieira: > this is essentially a wait_this_long_until_actually_shutting_down right?


Exactly. The reason a user might want to configure this is explained in the listeners documentation:

>This is designed to allow an operator to configure load-balancers to preemptively stop new traffic to a Boundary instance that is going away.

internal/servers/controller/handler.go
```diff
@@ -74,6 +76,29 @@ func (c *Controller) apiHandler(props HandlerProperties) (http.Handler, error) {
 	return eventsHandler, err
 }
 
+func (c *Controller) GetHealthHandler(lcfg *listenerutil.ListenerConfig) (http.Handler, error) {
```
tmessi, line 79: can you add a godoc?

internal/servers/controller/handler.go
```diff
@@ -74,6 +76,29 @@ func (c *Controller) apiHandler(props HandlerProperties) (http.Handler, error) {
...
+
+	wrapped := wrapHandlerWithCommonFuncs(healthGrpcGwMux, c, HandlerProperties{lcfg, c.baseContext})
+	return common.WrapWithEventsHandler(wrapped, c.conf.Eventer, c.kms, lcfg)
+}
+
+func registerHealthGrpcGatewayEndpoint(ctx context.Context, gwMux *runtime.ServeMux, dialOptions ...grpc.DialOption) error {
+	return opsservices.RegisterHealthServiceHandlerFromEndpoint(ctx, gwMux, gatewayTarget, dialOptions)
+}
+
+func (c *Controller) registerGrpcHealthService(s *grpc.Server) {
```
tmessi, line 94: Is there a reason to not do this as part of `registerGrpcServices`?

website/content/docs/configuration/listener/index.mdx
```diff
@@ -9,7 +9,33 @@ description: |-
...
+Currently, Boundary recognizes the following listener purposes:
+
+- `api`: Starts up and exposes Boundary's controller API.
+  At least one `api` listener must be present if a `cluster`
+  purpose listener is present. By default, it runs on `:9200`.
+
+- `cluster`: Starts up and exposes Boundary's worker and
+  controller communication layer. Only one `cluster` listener
+  is allowed per Boundary instance. Must be present if an `api`
+  purpose listener is present. By default, it runs on `:9201`.
```
tmessi, line 30: I think it should be made clear that these are needed for a boundary Controller instance.

tmessi (changes requested):

internal/servers/controller/controller.go
```diff
@@ -90,6 +95,8 @@ func New(ctx context.Context, conf *Config) (*Controller, error) {
 		workerAuthCache:         new(sync.Map),
 		workerStatusUpdateTimes: new(sync.Map),
 		enabledPlugins:          conf.Server.EnabledPlugins,
+		apiListeners:            make([]*base.ServerListener, 0),
+		HealthService:           &health.Service{},
```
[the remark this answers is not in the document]
hugoamvieira: You mean `HealthService`? The diff makes it a little ambiguous.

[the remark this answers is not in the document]
tmessi: Yea `HealthService` i think should default to nil, we do a nil check on it during shutdown: https://github.com/hashicorp/boundary/pull/1882/files#diff-39d8b05de61be8fba5e909420b32a4182700e95200188a0be03efccc858000f0R653-R655

website/content/docs/configuration/listener/index.mdx
hugoamvieira: Addressed in [95f3c08](https://github.com/hashicorp/boundary/pull/1882/commits/95f3c08a0fc1d54b3cb55e1d7dbbfe838d0318b7)

internal/servers/controller/handler.go
hugoamvieira: Addressed in [8140f0f](https://github.com/hashicorp/boundary/pull/1882/commits/8140f0ffd35b36bdce2ffdc2eafca9070f98615e)

internal/cmd/config/config.go
```diff
@@ -128,6 +133,12 @@ type Controller struct {
 	AuthTokenTimeToStale         interface{} `hcl:"auth_token_time_to_stale"`
 	AuthTokenTimeToStaleDuration time.Duration
 
+	// OpsListenersGracePeriod is the amount of time that we'll wait before actually
+	// starting the Controller shutdown. This is used to allow the health endpoint
+	// to return a bad status code for a little while before the instance goes away.
```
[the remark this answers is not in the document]
hugoamvieira: That's clearer - Addressed in [8140f0f](https://github.com/hashicorp/boundary/pull/1882/commits/8140f0ffd35b36bdce2ffdc2eafca9070f98615e)

internal/cmd/commands/server/server.go
```diff
@@ -41,6 +43,7 @@ var (
 
 type Command struct {
 	*base.Server
+	opsServers []*base.OpsServer
```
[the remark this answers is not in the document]
hugoamvieira: Addressed in [3a34b26](https://github.com/hashicorp/boundary/pull/1882/commits/3a34b26470abe8e671f8627cb0ac0c94e4988c84)

website/content/docs/configuration/controller.mdx
tmessi: I agree with a lot of this discussion, I am not sold on any of the options mentioned so far. Some additional options:

- `graceful_shutdown_period`
- `graceful_shutdown_wait_duration`

hugoamvieira: I think one of the considerations we should make is that this variable is related to the health endpoint. We could mention it just in the docs blurb but it feels strange to me because I see it as a crucial bit of information - If the health endpoint didn't start, we won't use it, even if it's configured... so I'd think the name ought to allude to health _somehow_

internal/cmd/ops/server.go
```diff
@@ -0,0 +1,202 @@
...
+	controller *controller.Controller
+}
+
+type opsBundle struct {
+	ln      *base.ServerListener
+	h       http.Handler
+	startFn []func()
+}
+
+func NewServer(l hclog.Logger, c *controller.Controller, listeners ...*base.ServerListener) (*Server, error) {
```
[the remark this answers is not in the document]
hugoamvieira: [2d07917](https://github.com/hashicorp/boundary/pull/1882/commits/2d079172b95e947028416b45e8a411d676c6c344)

internal/cmd/ops/server.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/hashicorp/boundary/internal/servers/controller"
+	"github.com/hashicorp/go-cleanhttp"
+	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-multierror"
+	"github.com/hashicorp/go-secure-stdlib/listenerutil"
+	"github.com/mitchellh/cli"
+	"github.com/pkg/errors"
+)
+
+type Server struct {
```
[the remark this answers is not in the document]
hugoamvieira: [2d07917](https://github.com/hashicorp/boundary/pull/1882/commits/2d079172b95e947028416b45e8a411d676c6c344)

internal/cmd/ops/server.go
```diff
@@ -0,0 +1,202 @@
+package ops
```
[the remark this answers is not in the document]
hugoamvieira: [2d07917](https://github.com/hashicorp/boundary/pull/1882/commits/2d079172b95e947028416b45e8a411d676c6c344)

website/content/docs/configuration/controller.mdx
hugoamvieira: Changed it to `graceful_shutdown_wait_duration` in [70560f6](https://github.com/hashicorp/boundary/pull/1882/commits/70560f63799e99433406728b0a2c7fa91f6768b8)

tmessi (approved):

PR #1883, opened 2022-03-03
docs(targets): remove default port option

irenarindos:
Remove Targets doc reference to a default port attribute - there is no way for the user to specify a port at connect time and we don't currently plan on allowing it either.


xka5h (approved):
pretty straightforward removal, looks good to me.

talanknight (approved):
Looks good! Thanks.

PR #1884, opened 2022-03-03
updated CLI connection options, including addr

xka5h:
I noticed that the Boundary CLI connection options are currently only displayed when `boundary dev -h` is ran. Not sure if this should also be included in the output when `boundary -h` is ran, but is currently not. As part of this commit, I did mention running `boundary dev -h` will display connection options. 

I'm not sure if we should include that in this page given the intention is to have it displayed only when `boundary -h` is ran. We can leave it out for now or keep it as is and update it in the future if we need to.


PPacent (approved):
LGTM

PR #1887, opened 2022-03-04
Docs: Update connect-to-target page with target-name example

stellarsquall:
This docs PR:

- adds an example of connecting to the default ssh target in dev mode using `target-name` and `target-scope-name`
- removes some passive and ableist language from the copy

On the boundary docs day Task List this task is labeled as:

> target-name and scope-name as arguments for boundary connect

### DIFF
--- a/website/content/docs/getting-started/connect-to-target.mdx
+++ b/website/content/docs/getting-started/connect-to-target.mdx
@@ -83,6 +83,12 @@ across scopes. As a result, when not using the target ID, you must use the
 target's name in conjunction with the scope name or scope ID so that Boundary
 can correctly identify the desired target.
 
+Here is an SSH example in dev mode:
+
+```
+$ boundary connect ssh -target-name "Generated target" -target-scope-name "Generated project scope"
+```
+
 ## Built-In vs. Exec
 
 Boundary comes with built-in wrappers for popular layer 7 connection protocols,
--- a/website/content/docs/getting-started/connect-to-target.mdx
+++ b/website/content/docs/getting-started/connect-to-target.mdx
@@ -35,9 +35,7 @@ within the authorized session. When you are finished making connections, simply
 
 ### Using Connect Helpers
 
-It can be annoying to keep accepting host SSH key prompts as the port changes,
-much less having to keep copying and pasting the current port. To make this
-easier, Boundary includes connect helpers that take care of this work for you.
+Boundary includes connect helpers that automatically accept host SSH key prompts for you.
 These are written as `boundary connect <subcommand>`. In the following example,
 the helper will automatically execute `ssh` for you, filling in the local
 address/port, and setting an expected host ID so that future connections on
@@ -64,7 +62,7 @@ connection information to PuTTY.
 One advantage to styles is that Boundary will provide information to the client
 in the format that makes sense. For example, the `-username` flag is used to
 specify a username other than your currently logged-in user. This ensures that
-regardless of whether you use the default `ssh` style or the `putty` style, the
+regardless of your `-style` choice, the
 username is properly passed to the executed client, and you don't need to figure
 out the syntax yourself.
 
--- a/website/content/docs/getting-started/connect-to-target.mdx
+++ b/website/content/docs/getting-started/connect-to-target.mdx
@@ -9,25 +9,25 @@ description: |-
 
 The default target is a `tcp` target with a default port of `22` (which
 corresponds to the default SSH port using TCP). The host sets for this target
-contain the default host, which has the address `127.0.0.1`. When we run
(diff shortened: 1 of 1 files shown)


website/content/docs/getting-started/connect-to-target.mdx
```diff
@@ -38,51 +38,57 @@ within the authorized session. When you are finished making connections, simply
 It can be annoying to keep accepting host SSH key prompts as the port changes,
 much less having to keep copying and pasting the current port. To make this
 easier, Boundary includes connect helpers that take care of this work for you.
```
hugoamvieira, line 40: I was reading this bit and couldn't help but think about the workshops yesterday. It just seems like extra fluff and I wonder if we should keep it.

Perhaps we could have something lighter like "Boundary includes connect helpers that automatically accept host SSH key prompts for you" ?

hugoamvieira: There's also a bit of repetition around as we talk about these helpers in the `ssh` context above - Perhaps we should introduce the connect helpers concept first, and then go into the SSH example?

website/content/docs/getting-started/connect-to-target.mdx
```diff
@@ -38,51 +38,57 @@ within the authorized session. When you are finished making connections, simply
...
-These take the form of `boundary connect <subcommand>`. In the following
-example, the helper will automatically execute `ssh` for you, filling in the
-local address/port, and setting an expected host ID so that future connections
-on different automatically-allocated ports don't complain about the host ID
-changing (you'll still need to accept a host key the first time):
+These are written as `boundary connect <subcommand>`. In the following example,
+the helper will automatically execute `ssh` for you, filling in the local
+address/port, and setting an expected host ID so that future connections on
+different automatically-allocated ports don't complain about the host ID
+changing (you'll still need to accept a host key the first time).
```
hugoamvieira, line 45: This paragraph (specifically the 2nd sentence) also reads really long and dense. Maybe we could split it or remove some info?

Also, the use of the `complain` verb here is a personification and more of a colloquial use in the English lang. It probably doesn't translate well into different languages (I can vouch for at least one where it doesn't).

We could change to something like this:

"These are written as `boundary connect <subcommand>`. In the following example,
the helper will automatically execute `ssh` for you, filling in the local address and port. 
It will also set the expected host ID to avoid warnings on future connections with different
ports. You'll still need to accept a host key the first time."

^ Not fully happy with this either, but I think it illustrates what I mean.
(I think it doesn't convey the same information the original paragraph does - I suppose we'd have to understand if a simplification is worth here considering the "Getting Started" context)

website/content/docs/getting-started/connect-to-target.mdx
```diff
@@ -38,51 +38,57 @@ within the authorized session. When you are finished making connections, simply
...
 
 One advantage to styles is that Boundary will provide information to the client
-in the format that makes sense. For example, if you want to specify a username
-other than your currently logged-in user, you can do so via the `-username`
-flag. This will ensure that regardless of whether you use the default `ssh`
-style or the `putty` style, the username is properly passed to the executed
-client -- you don't need to figure out the syntax yourself.
+in the format that makes sense. For example, the `-username` flag is used to
+specify a username other than your currently logged-in user. This ensures that
+regardless of whether you use the default `ssh` style or the `putty` style, the
```
hugoamvieira, line 67: We've already described the available styles above, we could make this bit a little lighter.

```suggestion
regardless of your `-style` choice, the
```

hugoamvieira (commented):
Looking good! Few comments below ⬇️

website/content/docs/getting-started/connect-to-target.mdx
stellarsquall: To be honest, I like your rewrite here. It's much more direct and easier to understand

website/content/docs/getting-started/connect-to-target.mdx
stellarsquall: How do you feel about the change I just pushed?

> Boundary includes connect helpers that automatically accept host SSH key prompts
for you. These are written as `boundary connect <subcommand>`. 

> In the following example, the helper will automatically execute `ssh` for you,
filling in the local address and port. An expected host ID is set to avoid
warnings on future connections when a different port is allocated automatically.
A host key must still be accepted upon first connect.

hugoamvieira: That's just waaaaaay better! _\*chef's kiss\*_

Thanks for taking a look at it 😄

PPacent (approved):
Much clearer. Thanks @stellarsquall

PR #1891, opened 2022-03-04
refact(session): Move domain logic of Worker Status into domain service

tmessi:
This creates a domain service function `session.WorkerStatusReport`. The
service performs two functions:

1. Checks the status for all of the reported sessions, returning a
   report of the sessions that are in a "non-active" state, i.e.
   "cacneling" or "terminated".
2. Looks for any "orphaned" session connections, i.e. connections that
   are active but were not reported by the worker. These connections are
   marked as closed.

This also removes the "ShouldCloseConnectionsOnWorker" logic, as this
was determined to not be a possible state, and was unnecessary.

Also includes some refactoring to speed up some tests.

### DIFF
--- a/internal/servers/controller/handlers/workers/worker_service_status_test.go
+++ b/internal/servers/controller/handlers/workers/worker_service_status_test.go
@@ -1,0 +1,509 @@
+package workers_test
+
+import (
+	"context"
+	"sync"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/hashicorp/boundary/internal/authtoken"
+	"github.com/hashicorp/boundary/internal/db"
+	pbs "github.com/hashicorp/boundary/internal/gen/controller/servers/services"
+	"github.com/hashicorp/boundary/internal/host/static"
+	"github.com/hashicorp/boundary/internal/iam"
+	"github.com/hashicorp/boundary/internal/kms"
+	"github.com/hashicorp/boundary/internal/servers"
+	"github.com/hashicorp/boundary/internal/servers/controller/handlers/workers"
+	"github.com/hashicorp/boundary/internal/session"
+	"github.com/hashicorp/boundary/internal/target"
+	"github.com/hashicorp/boundary/internal/target/tcp"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestStatus(t *testing.T) {
+	ctx := context.Background()
+	conn, _ := db.TestSetup(t, "postgres")
+	rw := db.New(conn)
+	wrapper := db.TestWrapper(t)
+	kms := kms.TestKms(t, conn, wrapper)
+	org, prj := iam.TestScopes(t, iam.TestRepo(t, conn, wrapper))
+
+	serverRepo, _ := servers.NewRepository(rw, rw, kms)
+	serverRepo.UpsertServer(ctx, &servers.Server{
+		PrivateId: "test_controller1",
+		Type:      "controller",
+		Address:   "127.0.0.1",
+	})
+	serverRepo.UpsertServer(ctx, &servers.Server{
+		PrivateId: "test_worker1",
+		Type:      "worker",
+		Address:   "127.0.0.1",
+	})
+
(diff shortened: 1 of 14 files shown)


tmessi:
@mgaffney @jimlambrt I want to get both of your thoughts on this approach to make this particular query a bit more testable.

internal/session/repository_connection.go
```diff
@@ -321,67 +289,51 @@ func (r *ConnectionRepository) CloseConnectionsForDeadWorkers(ctx context.Contex
...
-// be a set of sessions we've already submitted close requests for, so adding
-// them again would be redundant.
-//
-// The returned map[string][]string is indexed by session ID.
-func (r *ConnectionRepository) ShouldCloseConnectionsOnWorker(ctx context.Context, foundConns, filterSessions []string) (map[string][]string, error) {
-	const op = "session.(ConnectionRepository).ShouldCloseConnectionsOnWorker"
-	if len(foundConns) < 1 {
-		return nil, nil // nothing to do
-	}
+func (r *ConnectionRepository) orphanedConnections(ctx context.Context, workerId string, reportedConnections []string) ([]string, error) {
```
irenarindos, line 292: Add function documentation

internal/servers/controller/handlers/workers/worker_service_status_test.go
```diff
@@ -0,0 +1,512 @@
...
+		name       string
+		wantErr    bool
+		wantErrMsg string
+		req        *pbs.StatusRequest
+		want       *pbs.StatusResponse
+	}{
+		{
+			name:       "No Sessions",
+			wantErr:    false,
+			wantErrMsg: "",
```
irenarindos, line 106: If wantErr is false, wantErrMsg isn't needed

internal/servers/controller/handlers/workers/worker_service_status_test.go
```diff
@@ -0,0 +1,512 @@
...
+						Type:      "controller",
+						Address:   "127.0.0.1",
+					},
+				},
+			},
+		},
+		{
+			name:       "Still Active",
+			wantErr:    false,
+			wantErrMsg: "",
```
irenarindos, line 123: Same here, wantErrMsg can be removed

internal/servers/controller/handlers/workers/worker_service_status_test.go
```diff
@@ -0,0 +1,512 @@
...
+		wantErr    bool
+		wantErrMsg string
+		setupFn    func(t *testing.T)
+		req        *pbs.StatusRequest
+		want       *pbs.StatusResponse
+	}{
+		{
+			name:       "Connection Canceled",
+			wantErr:    false,
+			wantErrMsg: "",
```
irenarindos, line 283: Same here, wantErrMsg can be removed

internal/session/query.go
```diff
@@ -10,15 +10,15 @@ const (
 insert into session_state
 with not_active as (
 	select session_id, 'active' as state
-	from
```
irenarindos, line 13: Do we want all these space diffs- there's a lot in this file.

internal/session/query.go
```diff
@@ -329,7 +285,45 @@ values
   (?, ?, ?)`
 
 	sessionCredentialDynamicBatchInsertReturning = `
-  returning session_id, library_id, credential_id, credential_purpose
+  returning session_id, library_id, credential_id, credential_purpose;
+`
+
+	orphanedConnectionsCte = `
+with
+  -- Find connections that are not closed so we can reference those IDs
```
irenarindos, line 293: Can this comment be moved before the with?

irenarindos (changes requested):

internal/session/query.go
tmessi: I can rip it out, I put this as a separate commit since I was conflicted on making the change. Generally I don't like seeing trailing whitespace, and `gofmt` removes them from go code, but since this sql code is in a string, they don't get removed automatically.

irenarindos (approved):
Looks good to me!

jefferai:
I think this is looking really good!

jefferai (commented):
Going to defer to others that have been tracking more closely on a final call but from reading through this it looks good to me!

irenarindos (approved):

Issue #1812, opened 2022-01-03
DEADLOCK on warehouse tables


giulio1979:
Hi, we have noticed that after a few sessions have been established (5-6) system is getting slow and at some point boundary controller states deadlocks are happening in the database:

`2022-01-03T12:02:06.481Z [TRACE] error from database adapter: location=/go/internal/db/read_writer.go:618 error="ERROR: deadlock detected (SQLSTATE 40P01)"
[...]

PR #1895, opened 2022-03-07
refactor(session connections): Refactor connection closure

irenarindos:
Under heavy load, we saw deadlocks as competing processes attempted to close the same connection. A connection was closed on proxy exit at the same time as the Worker failed to make its status call and attempted to close all its connections. Connections and sessions were also closed in the same transactions.

This PR resolves this issue by breaking apart transactions for sessions and connections, and removing the Worker connection closure requests on status failure to prevent the resource contention causing the deadlocks.

- Creation of domain service, CloseConnections, which closes connections in the Connection repo and then attempts to terminate sessions if possible using the Sessions repo
- The Session repository function CloseConnections() has been moved to the Session Connection repository.
- The trigger function, terminate_session_if_possible, has been removed. Sessions are terminated if possible in the new domain service
- Refactored warehouse triggers to only do rollup of connection stats on session termination

Refs: #1812

### DIFF
--- a/internal/db/schema/migrations/oss/postgres/27/01_disable_terminate_session.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/27/01_disable_terminate_session.up.sql
@@ -1,0 +1,36 @@
+begin;
+-- Replaces function from 0/51_connection.up.sql
+-- Remove call to terminate_session_if_possible
+drop trigger update_connection_state_on_closed_reason on session_connection;
+drop function update_connection_state_on_closed_reason();
+create function
+    update_connection_state_on_closed_reason()
+    returns trigger
+as $$
+    begin
+        if new.closed_reason is not null then
+            -- check to see if there's a closed state already, before inserting a new one.
+            perform from
+                session_connection_state cs
+            where
+                    cs.connection_id = new.public_id and
+                    cs.state = 'closed';
+            if not found then
+                insert into session_connection_state (connection_id, state)
+                values
+                    (new.public_id, 'closed');
+            end if;
+        end if;
+    return new;
+    end;
+$$ language plpgsql;
+
+create trigger
+    update_connection_state_on_closed_reason
+    after update of closed_reason on session_connection
+    for each row execute procedure update_connection_state_on_closed_reason();
+
+-- Remove function, defined in 0/51_connection.up.sql
+drop function terminate_session_if_possible;
+
+commit;
--- a/internal/db/schema/migrations/oss/postgres/27/02_wh_session_facts.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/27/02_wh_session_facts.up.sql
@@ -1,0 +1,114 @@
+begin;
+-- Updating definition from 16/05_wh_credential_dimension.up.sql
+-- Remove call to wh_rollup_connections(new.session_id) from function
+drop trigger wh_insert_session_connection on session_connection;
+drop function wh_insert_session_connection();
+create function wh_insert_session_connection()
+    returns trigger
+as $$
+declare
+new_row wh_session_connection_accumulating_fact%rowtype;
+begin
+with
+    authorized_timestamp (date_dim_key, time_dim_key, ts) as (
+        select wh_date_key(start_time), wh_time_key(start_time), start_time
+        from session_connection_state
+        where connection_id = new.public_id
+          and state = 'authorized'
+    ),
+    session_dimension (host_dim_key, user_dim_key, credential_group_dim_key) as (
+        select host_key, user_key, credential_group_key
+        from wh_session_accumulating_fact
+        where session_id = new.session_id
+    )
+insert into wh_session_connection_accumulating_fact (
+        connection_id,
+        session_id,
+        host_key,
+        user_key,
+        credential_group_key,
+        connection_authorized_date_key,
+        connection_authorized_time_key,
+        connection_authorized_time,
+        client_tcp_address,
+        client_tcp_port_number,
+        endpoint_tcp_address,
+        endpoint_tcp_port_number,
+        bytes_up,
+        bytes_down
+    )
+select new.public_id,
+       new.session_id,
+       session_dimension.host_dim_key,
+       session_dimension.user_dim_key,
+       session_dimension.credential_group_dim_key,
+       authorized_timestamp.date_dim_key,
+       authorized_timestamp.time_dim_key,
+       authorized_timestamp.ts,
+       new.client_tcp_address,
+       new.client_tcp_port,
+       new.endpoint_tcp_address,
+       new.endpoint_tcp_port,
+       new.bytes_up,
+       new.bytes_down
+from authorized_timestamp,
+     session_dimension
+         returning * into strict new_row;
+return null;
+end;
+$$ language plpgsql;
+
+create trigger wh_insert_session_connection
+    after insert on session_connection
+    for each row
+    execute function wh_insert_session_connection();
+
+-- Updating definition from 0/69_wh_session_facts.up.sql
+-- Remove call to wh_rollup_connections(new.session_id) from function
+drop trigger wh_update_session_connection on session_connection;
+drop function wh_update_session_connection;
+create function wh_update_session_connection()
+    returns trigger
+as $$
+declare
+updated_row wh_session_connection_accumulating_fact%rowtype;
+begin
+update wh_session_connection_accumulating_fact
+set client_tcp_address       = new.client_tcp_address,
+    client_tcp_port_number   = new.client_tcp_port,
+    endpoint_tcp_address     = new.endpoint_tcp_address,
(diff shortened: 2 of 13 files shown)


internal/servers/controller/handlers/workers/worker_service.go
```diff
@@ -469,19 +469,28 @@ func (ws *workerServiceServer) CloseConnection(ctx context.Context, req *pbs.Clo
...
 
+	sessRepo, err := ws.sessionRepoFn()
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "error getting session repo: %v", err)
+	}
+
+	for _, v := range req.GetCloseRequestData() {
+		sessRepo.TerminateSessionIfPossible(ctx, v.GetConnectionId())
+	}
+
```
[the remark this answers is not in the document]
tmessi: The `Connection` in each `CloseConnectionResp` in `closeInfos` has the `SessionId`. So we should be able to get the set of unique sessions that had a connection closed. That would allow `TerminateSessionIfPossible` to be given a session id, or the set of session ids, which I think willl simplify the query a little.

tmessi (changes requested):

internal/db/schema/migrations/oss/postgres/27/01_disable_terminate_session.up.sql
```diff
@@ -0,0 +1,36 @@
...
+            session_connection_state cs
+        where
+                cs.connection_id = new.public_id and
+                cs.state = 'closed';
+        if not found then
+            insert into session_connection_state (connection_id, state)
+            values
+                (new.public_id, 'closed');
+end if;
+end if;
```
tmessi, line 23: the indentation of these seems off. they should align with the corresponding `if`

internal/db/schema/migrations/oss/postgres/27/01_disable_terminate_session.up.sql
```diff
@@ -0,0 +1,36 @@
...
+return new;
```
tmessi, line 24: this should be indented one since it is within the `being` block

internal/db/schema/migrations/oss/postgres/27/02_wh_session_facts.up.sql
```diff
@@ -0,0 +1,113 @@
+begin;
+
+-- Remove call to wh_rollup_connections(new.session_id) from function
```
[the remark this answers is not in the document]
tmessi: can you also add a comment in `16/05_wh_credential_dimension.up.sql`?

internal/db/schema/migrations/oss/postgres/27/02_wh_session_facts.up.sql
```diff
@@ -0,0 +1,113 @@
...
+return null;
+end;
+$$ language plpgsql;
+
+create trigger wh_insert_session_connection
+    after insert on session_connection
+    for each row
+    execute function wh_insert_session_connection();
+
+-- Remove call to wh_rollup_connections(new.session_id) from function
```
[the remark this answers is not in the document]
tmessi: and likewise add a comment in `0/69_wh_session_facts.up.sql`

internal/db/schema/migrations/oss/postgres/27/02_wh_session_facts.up.sql
```diff
@@ -0,0 +1,114 @@
...
+return null;
+end;
+$$ language plpgsql;
+
+create trigger wh_update_session_connection
+    after update on session_connection
+    for each row
+    execute function wh_update_session_connection();
+
+create or replace function
```
tmessi, line 94: since this is a new function, it should just use `create function`, we want to avoid `create or replace` so it doesn't mask an error.

internal/db/schema/migrations/oss/postgres/27/02_wh_session_facts.up.sql
```diff
@@ -0,0 +1,114 @@
...
+    perform_session_rollup()
```
tmessi, line 95: since this is a warehouse function, it should be prefixed with `wh_`, probably something like `wh_session_rollup`

internal/session/repository_session.go
```diff
@@ -375,6 +375,34 @@ func (r *Repository) TerminateCompletedSessions(ctx context.Context) (int, error
 	return rowsAffected, nil
 }
 
+// TerminateSessionIfPossible is called on connection close and will attempt to close the connection's
+// session if the following conditions are met:
+//  * sessions that have exhausted their connection limit and all their connections are closed.
+//	* sessions that are expired and all their connections are closed.
+//	* sessions that are canceling and all their connections are closed
+func (r *Repository) TerminateSessionIfPossible(ctx context.Context, connectionId string, sessionVersion uint32) (int, error) {
```
tmessi, line 383: I think this can be unexported now since it should only be called by the domain service function

internal/session/repository_connection.go
```diff
@@ -184,6 +184,71 @@ func (r *ConnectionRepository) ListConnectionsBySessionId(ctx context.Context, s
...
+// It wraps the connection and its states for each connection closed.
+type CloseConnectionResp struct {
+	Connection       *Connection
+	ConnectionStates []*ConnectionState
+}
+
+// CloseConnections set's a connection's state to "closed" in the repo.  It's
+// called by a worker after it's closed a connection between the client and the
+// endpoint
+func (r *ConnectionRepository) CloseConnections(ctx context.Context, closeWith []CloseWith, _ ...Option) ([]CloseConnectionResp, error) {
```
tmessi, line 197: I think this can be unexported now since it should only be called by the domain service.

internal/session/repository_connection.go
```diff
@@ -184,6 +184,71 @@ func (r *ConnectionRepository) ListConnectionsBySessionId(ctx context.Context, s
 	return connections, nil
 }
 
+// CloseConnectionResp is just a wrapper for the response from CloseConnections.
+// It wraps the connection and its states for each connection closed.
+type CloseConnectionResp struct {
+	Connection       *Connection
+	ConnectionStates []*ConnectionState
+}
```
tmessi, line 192: likewise this can be unexported, since it is a result that only the domain service uses

tmessi (changes requested):

internal/session/service_close_connections.go
```diff
@@ -0,0 +1,33 @@
...
+func CloseConnections(ctx context.Context, sessionRepoFn *Repository, connectionRepoFn *ConnectionRepository,
+	closeWiths []CloseWith, opt ...Option) ([]closeConnectionResp, error) {
+	const op = "session.AuthorizeConnection"
+
+	closeInfos, err := connectionRepoFn.closeConnections(ctx, closeWiths)
+	if err != nil {
+		return nil, errors.Wrap(ctx, err, op)
+	}
+
+	for _, c := range closeInfos {
```
tmessi, line 21: The same session could be in each element of `closeInfos`. I think we should get the set of unique session ids first, and then attempt to terminate them.

internal/session/service_close_connections.go
```diff
@@ -0,0 +1,33 @@
...
+		ses, _, err := sessionRepoFn.LookupSession(ctx, c.Connection.SessionId)
+		if err != nil {
+			return nil, err
+		}
+		if ses != nil {
+			version := ses.Version
+			sessionRepoFn.terminateSessionIfPossible(ctx, c.Connection.SessionId, version)
+		}
```
tmessi, line 29: I still don't follow why this version is needed, but I think if we need it, it should be done in `terminateSessionIfPossible` so it is part of the same transaction instead of calling LookupSession here.

tmessi (changes requested):

internal/session/service_close_conections_test.go
```diff
@@ -0,0 +1,97 @@
+package session
```
tmessi, line 1: Looks like there is a typo in this filename, I think it should be `service_close_connections_test.go`.

tmessi (commented):
one minor thing, otherwise its looking good

internal/session/service_close_connections.go
```diff
@@ -0,0 +1,31 @@
...
+	"context"
+
+	"github.com/hashicorp/boundary/internal/errors"
+)
+
+// CloseConnection is a domain service function that:
+// * closes requested connections
+// * uses the sessionId of the connection to see if the session meets conditions for termination
+func CloseConnections(ctx context.Context, sessionRepoFn *Repository, connectionRepoFn *ConnectionRepository,
+	closeWiths []CloseWith, opt ...Option) ([]closeConnectionResp, error) {
```
tmessi, line 13: nitpick: it looks like `opt` is not used, this should either be removed or named `_` to note that it is not used.

tmessi (commented):
ok, now just a couple more nitpicky things, but then I think its good

tmessi (approved):

PR #1901, opened 2022-03-10
Update to go-kms-wrapping version 2, and plugin-based KMS

jefferai:
This updates to the new version of go-kms-wrapping. Much of the original code here worked its way into host plugins, `pluginutil`/`configutil` inside `go-secure-stdlib`, and so on. So this looks like a much more scary PR than it is. But it's 200 changed files, so....no joke.

Unfortunately this can't be easily broken up. In theory v1 and v2 can encrypt and decrypt each other (and I have a commented-out test that ensures this, but is commented out to not actually pull in the large number of go-kms-wrapping v1 deps) but when passing around the values to various parts of Boundary they need to be the same structures. (I thought about making an interface that both v1 and v2 structures could implement but I think at the moment that will just make things confusing for users of the library, even if it would ease this one-time transition.)

However, the _vast majority_ of the changes are purely swapping in v2 bits for v1 bits. These are bits caught by the compiler. So what I've done is I've added a comment anywhere that this is _not_ the case. So a really easy way to review this is to look for where I've put a comment indicating non-trivial changed code, and review that section of code. Then move on to my next comment.

To get an idea for just how much of a difference v2 and using plugins makes, see `api/go.mod` and `api/go.sum` as good examples.


internal/auth/oidc/service.go
```diff
@@ -62,7 +61,11 @@ func encryptMessage(ctx context.Context, wrapper wrapping.Wrapper, am *AuthMetho
 	if wrapper == nil {
 		return "", errors.New(ctx, errors.InvalidParameter, op, "missing wrapper")
 	}
-	if wrapper.KeyID() == "" {
```
jefferai, line 65: This file contains not-totally-trivial changes and should be reviewed.

internal/cmd/commands/accountscmd/accounts.gen.go
```diff
@@ -155,8 +155,15 @@ func (c *Command) Run(args []string) int {
 	}
 
 	client, err := c.Client()
+	if c.WrapperCleanupFunc != nil {
```
jefferai, line 158: Reviewers: this is representative of how the various generated files changed, so you can take a look at this and ignore the rest.

What's happening here is simply ensuring that _if_ a cleanup func is specified by a wrapper, we ensure that it is run at the end of the command.

internal/cmd/base/base.go
```diff
@@ -220,30 +226,39 @@ func (c *Command) Client(opt ...Option) (*api.Client, error) {
 
 	switch {
 	case c.FlagRecoveryConfig != "":
-		wrapper, err := wrapper.GetWrapperFromPath(c.FlagRecoveryConfig, "recovery")
+		wrapper, cleanupFunc, err := wrapper.GetWrapperFromPath(
```
jefferai, line 229: Reviewers: take a look at this file. This resolves a longstanding issue around finalization of wrappers instantiated as part of a command call (in order to handle any recovery flows in the API) where we weren't ever finalizing. If you look at the comment you can see that the change here was first theorized long ago (storing the function in the command) so this is just a matter of putting that idea in practice. The generated files then ensure to call this function if it exists.

internal/cmd/commands/config/encryptdecrypt.go
```diff
@@ -142,7 +145,14 @@ func (c *EncryptDecryptCommand) Run(args []string) (ret int) {
 		kmsDefFile = strings.TrimSpace(c.flagConfigKms)
 	}
 
-	wrapper, err := wrapper.GetWrapperFromPath(kmsDefFile, "config")
+	wrapper, cleanupFunc, err := wrapper.GetWrapperFromPath(
```
jefferai, line 148: Reviewers: here's a file to look at. Note that cleanupFunc should never be nil; the reason I do a nil check is simply to avoid deferring another func that does error reporting if I don't need to. It could safely be taken out though. But also, nil checking is never bad. :-)

internal/cmd/commands/database/init.go
```diff
@@ -8,10 +8,13 @@ import (
 	"github.com/hashicorp/boundary/internal/cmd/config"
 	"github.com/hashicorp/boundary/internal/errors"
 	"github.com/hashicorp/boundary/internal/types/scope"
+	kms_plugin_assets "github.com/hashicorp/boundary/plugins/kms"
```
jefferai, line 11: Reviewers: Same changes as in base.go. Leaving a comment here mostly for me so that if there were concerns in base.go I remember to address them here too.

internal/cmd/commands/database/migrate.go
```diff
@@ -7,12 +7,14 @@ import (
 	"github.com/hashicorp/boundary/internal/cmd/config"
 	"github.com/hashicorp/boundary/internal/errors"
 	host_plugin_assets "github.com/hashicorp/boundary/plugins/host"
+	kms_plugin_assets "github.com/hashicorp/boundary/plugins/kms"
```
jefferai, line 10: Reviewers: Same changes as in base.go. Leaving a comment here mostly for me so that if there were concerns in base.go I remember to address them here too.

internal/kms/repository_audit_key.go
```diff
@@ -44,7 +44,10 @@ func createAuditKeyTx(ctx context.Context, r db.Reader, w db.Writer, rkvWrapper
 	if len(key) == 0 {
 		return nil, nil, errors.New(ctx, errors.InvalidParameter, op, "missing key")
 	}
-	rootKeyVersionId := rkvWrapper.KeyID()
+	rootKeyVersionId, err := rkvWrapper.KeyId(ctx)
```
jefferai, line 47: Reviewers: this file could be looked at if desired, very minor changes but repeated throughout the various key types.

internal/kms/repository_audit_key_version.go
```diff
@@ -7,7 +7,7 @@ import (
 
 	"github.com/hashicorp/boundary/internal/db"
 	"github.com/hashicorp/boundary/internal/errors"
-	wrapping "github.com/hashicorp/go-kms-wrapping"
+	wrapping "github.com/hashicorp/go-kms-wrapping/v2"
```
jefferai, line 10: Reviewers: this file could be looked at if desired, very minor changes but repeated throughout the various key types.

scripts/build.sh
```diff
@@ -43,21 +43,23 @@ if [ "${GOOS}x" == "windowsx" ]; then
...
 
 # Build needed plugins first
-ORIG_PATH=$(pwd);
-echo "==> Building Host Plugins..."
-for PLUGIN_TYPE in host; do
-    rm -f $ORIG_PATH/plugins/$PLUGIN_TYPE/assets/boundary-plugin-${PLUGIN_TYPE}*
-    for CURR_PLUGIN in $(ls $ORIG_PATH/plugins/$PLUGIN_TYPE/mains); do
-        cd $ORIG_PATH/plugins/$PLUGIN_TYPE/mains/$CURR_PLUGIN;
-        go build -v -o $ORIG_PATH/plugins/$PLUGIN_TYPE/assets/boundary-plugin-${PLUGIN_TYPE}-${CURR_PLUGIN}${BINARY_SUFFIX} .;
+if [ "${SKIP_PLUGIN_BUILD}x" == "x" ]; then
```
jefferai, line 46: Reviewers: This ensure we build both KMS and host plugins, and also allows for skipping plugin build if you know they're up-to-date for faster builds.

internal/cmd/base/servers.go
```diff
@@ -471,7 +473,7 @@ func (b *Server) SetupListeners(ui cli.Ui, config *configutil.SharedConfig, allo
 	return nil
 }
 
-func (b *Server) SetupKMSes(ui cli.Ui, config *config.Config) error {
+func (b *Server) SetupKMSes(ctx context.Context, ui cli.Ui, config *config.Config) error {
```
jefferai, line 476: Reviewers: this file has changed a non-trivial bit

internal/kms/kms.go
```diff
@@ -10,9 +10,9 @@ import (
 	"github.com/hashicorp/boundary/internal/db"
 	"github.com/hashicorp/boundary/internal/errors"
 	"github.com/hashicorp/boundary/internal/types/scope"
-	wrapping "github.com/hashicorp/go-kms-wrapping"
-	"github.com/hashicorp/go-kms-wrapping/wrappers/aead"
-	"github.com/hashicorp/go-kms-wrapping/wrappers/multiwrapper"
+	wrapping "github.com/hashicorp/go-kms-wrapping/v2"
```
jefferai, line 13: Reviewers: the changes in here are straightforward but there's a bunch so calling it out for a look.

sdk/plugins/host/load.go
```diff
@@ -1,18 +1,14 @@
...
 	"fmt"
 	"io/fs"
-	"io/ioutil"
-	"os"
-	"path/filepath"
-	"runtime"
 	"strings"
 
 	pb "github.com/hashicorp/boundary/sdk/pbs/plugin"
+	"github.com/hashicorp/go-plugin"
```
jefferai, line 10: Reviewers: The actual plugin instantiating code moved over to pluginutil where it's now shared between KMS and hosts (and can satisfy any other interface), hence all the red.

sdk/plugins/host/options.go
```diff
@@ -1,13 +1,10 @@
...
 
 import (
-	"errors"
 	"fmt"
-	"io/fs"
-	"os"
 
-	pb "github.com/hashicorp/boundary/sdk/pbs/plugin"
 	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-secure-stdlib/pluginutil/v2"
```
jefferai, line 7: Reviewers: Same deal; this now supports pass-through options to pluginutil instead of hosting them directly.

sdk/wrapper/wrapper.go
```diff
@@ -1,49 +1,56 @@
 package wrapper
 
 import (
+	"context"
```
jefferai, line 4: Reviewers: Pretty straightforward but worth a look.

internal/cmd/base/base.go
```diff
@@ -220,30 +227,46 @@ func (c *Command) Client(opt ...Option) (*api.Client, error) {
...
+			c.Context,
+			c.FlagRecoveryConfig,
+			"recovery",
+			configutil.WithPluginOptions(
+				pluginutil.WithPluginsMap(kms_plugin_assets.BuiltinKmsPlugins()),
+				pluginutil.WithPluginsFilesystem("boundary-plugin-kms-", kms_plugin_assets.FileSystem()),
+			),
+			// TODO: How would we want to expose this kind of log to users when
```
jefferai, line 238: Reviewers: Note that this here and a few other places is for recovery wrappers. In general since those are sourced as part of a separate command we don't have logging set up. So it's not clear what to do here. For the moment swallowing it is likely fine, as we would have swallowed it anyways if it was built-in -- this means swallowing go-plugin status info but there are other ways to verify if that is an issue (such as with `boundary dev`).

internal/cmd/commands/server/server.go
```diff
@@ -133,21 +134,12 @@ func (c *Command) AutocompleteFlags() complete.Flags {
...
-	ctx := context.TODO()
 	c.CombineLogs = c.flagCombineLogs
 
 	if result := c.ParseFlagsAndConfig(args); result > 0 {
 		return result
 	}
 
-	if c.configWrapper != nil {
-		defer func() {
-			if err := c.configWrapper.Finalize(c.Context); err != nil {
```
jefferai, line 145: Reviewers: Note the change in context where this is moved to -- it uses context.Background(). This is because if the context is canceled due to Ctrl-C, we still want to attempt to finalize the wrapper.

Future work could include adding a timeout somewhere in the finalize call so it can't hang forever, but this is still safer than using an already-canceled context.

internal/cmd/commands/server/server.go
```diff
@@ -133,21 +134,12 @@ func (c *Command) AutocompleteFlags() complete.Flags {
 }
 
 func (c *Command) Run(args []string) int {
-	ctx := context.TODO()
```
jefferai, line 136: Reviewers: In a few of these commands context.TODO had been inserted. This was from a previous change that added updated error into various files that required context. I believe the person that did that work simply did not realize that each command already has a context baked in. I can see no reason to use TODO or Background here.

internal/kms/kms.go
```diff
@@ -244,31 +256,36 @@ func (k *Kms) loadRoot(ctx context.Context, scopeId string, opt ...Option) (*mul
...
 		}
-		if err := wrapper.SetAESGCMKeyBytes(key.GetKey()); err != nil {
+		if err = wrapper.SetAesGcmKeyBytes(key.GetKey()); err != nil {
 			return nil, "", errors.Wrap(ctx, err, op, errors.WithMsg(fmt.Sprintf("error setting key bytes on aead root wrapper in scope %s", scopeId)))
 		}
 		if i == 0 {
-			multi = multiwrapper.NewMultiWrapper(wrapper)
+			pooled, err = multi.NewPooledWrapper(ctx, wrapper)
+			if err != nil {
+				return nil, "", errors.Wrap(ctx, err, op, errors.WithMsg(fmt.Sprintf("error getting root multiwrapper for key version 0 in scope %s", scopeId)))
```
jimlambrt, line 278: do we want to call it a multi or pooled wrapper in the error msg?

internal/kms/kms.go
```diff
@@ -390,29 +407,34 @@ func (k *Kms) loadDek(ctx context.Context, scopeId string, purpose KeyPurpose, r
...
 		}
-		if err := wrapper.SetAESGCMKeyBytes(keyVersion.GetKey()); err != nil {
+		if err = wrapper.SetAesGcmKeyBytes(keyVersion.GetKey()); err != nil {
 			return nil, errors.Wrap(ctx, err, op, errors.WithMsg(fmt.Sprintf("error setting key bytes on aead %s wrapper in scope %s", purpose.String(), scopeId)))
 		}
 		if i == 0 {
-			multi = multiwrapper.NewMultiWrapper(wrapper)
+			pooled, err = multi.NewPooledWrapper(ctx, wrapper)
+			if err != nil {
+				return nil, errors.Wrap(ctx, err, op, errors.WithMsg(fmt.Sprintf("error getting %s multiwrapper for key version 0 in scope %s", purpose.String(), scopeId)))
```
jimlambrt, line 429: same question about multi vs pooled in the error msg.

internal/kms/repository_database_key.go
```diff
@@ -44,7 +44,10 @@ func createDatabaseKeyTx(ctx context.Context, r db.Reader, w db.Writer, rkvWrapp
 	if len(key) == 0 {
 		return nil, nil, errors.New(ctx, errors.InvalidParameter, op, "missing key")
 	}
-	rootKeyVersionId := rkvWrapper.KeyID()
+	rootKeyVersionId, err := rkvWrapper.KeyId(ctx)
+	if err != nil {
+		return nil, nil, errors.Wrap(ctx, err, op, errors.WithMsg("unable to get key id"))
```
jimlambrt, line 49: a bit inconsistent with other error msgs for this.. I think, you were using something like "unable to lookup root key"

internal/kms/repository_oidc_key.go
```diff
@@ -44,7 +44,10 @@ func createOidcKeyTx(ctx context.Context, r db.Reader, w db.Writer, rkvWrapper w
 	if len(key) == 0 {
 		return nil, nil, errors.New(ctx, errors.InvalidParameter, op, "missing key")
 	}
-	rootKeyVersionId := rkvWrapper.KeyID()
+	rootKeyVersionId, err := rkvWrapper.KeyId(ctx)
+	if err != nil {
+		return nil, nil, errors.Wrap(ctx, err, op, errors.WithMsg("unable to fetch key ID"))
```
jimlambrt, line 49: yet another error msg for the same error bits.

sdk/wrapper/upgrade_test.go
```diff
@@ -0,0 +1,56 @@
+package wrapper
+
+// NOTE: The test below is commented out for a very specific reason: to not
```
jimlambrt, line 3: ty!

internal/servers/controller/controller.go
```diff
@@ -140,9 +141,11 @@ func New(ctx context.Context, conf *Config) (*Controller, error) {
 			client, cleanup, err := external_host_plugins.CreateHostPlugin(
 				ctx,
 				pluginType,
-				external_host_plugins.WithHostPluginsFilesystem("boundary-plugin-host-", host_plugin_assets.FileSystem()),
-				external_host_plugins.WithHostPluginExecutionDir(conf.RawConfig.Plugins.ExecutionDir),
-				external_host_plugins.WithLogger(hclog.NewNullLogger()))
+				external_host_plugins.WithPluginOptions(
+					pluginutil.WithPluginExecutionDirectory(conf.RawConfig.Plugins.ExecutionDir),
+					pluginutil.WithPluginsFilesystem("boundary-plugin-host-", host_plugin_assets.FileSystem()),
```
jimlambrt, line 146: perhaps a const since it's used in a few places?

internal/servers/worker/controller_connection.go
```diff
@@ -199,7 +199,7 @@ func (w *Worker) workerAuthTLSConfig() (*tls.Config, *base.WorkerAuthInfo, error
 	if err != nil {
 		return nil, nil, err
 	}
-	encInfo, err := w.conf.WorkerAuthKms.Encrypt(context.Background(), marshaledInfo, nil)
+	encInfo, err := w.conf.WorkerAuthKms.Encrypt(context.Background(), marshaledInfo)
```
jimlambrt, line 202: perhaps w.baseContext?

plugins/kms/assets/README.md
```diff
@@ -0,0 +1,3 @@
+This directory contains assets for kms plugins. This file in particular exists
```
jimlambrt, line 1: ty!

jimlambrt (changes requested):
Amazing work and the removed deps makes me very happy.  I did have a couple of minor bits...

internal/servers/worker/controller_connection.go
jefferai: 👍

internal/servers/controller/controller.go
jefferai: Done

internal/kms/kms.go
jefferai: Updated

internal/cmd/base/servers.go
```diff
@@ -471,8 +478,9 @@ func (b *Server) SetupListeners(ui cli.Ui, config *configutil.SharedConfig, allo
...
 	sharedConfig := config.SharedConfig
+	var err error
 	for _, kms := range sharedConfig.Seals {
 		for _, purpose := range kms.Purpose {
 			purpose = strings.ToLower(purpose)
```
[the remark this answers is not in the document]
jefferai: Done

internal/cmd/commands/database/migrate.go
```diff
@@ -32,7 +36,7 @@ type MigrateCommand struct {
 
 	Config *config.Config
 
-	configWrapper wrapping.Wrapper
+	configWrapperCleanupFunc func() error
```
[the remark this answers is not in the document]
jefferai: Done (here and in init.go)

jimlambrt (approved):

internal/kms/repository_audit_key.go
```diff
@@ -44,7 +44,10 @@ func createAuditKeyTx(ctx context.Context, r db.Reader, w db.Writer, rkvWrapper
...
+	if err != nil {
+		return nil, nil, errors.Wrap(ctx, err, op, errors.WithMsg("unable to lookup root key id"))
```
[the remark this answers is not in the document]
jimlambrt: @louisruch I might offer that it's not an encrypt operation... so maybe the `errors.Encrypt` isn't quite right?

internal/oplog/testing_test.go
```diff
@@ -1,8 +1,11 @@
 package oplog
 
 import (
+	"context"
 	"testing"
 
+	wrapping "github.com/hashicorp/go-kms-wrapping/v2"
+
```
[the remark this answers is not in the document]
jefferai: Fixed

jimlambrt (approved):

PR #1904, opened 2022-03-10
chore(dbtest): Build test database docker image for arm64

tmessi:
This adds a new makefile target to build and publish the test database
docker image for multiple architectures. It will create images for:

- linux/amd64
- linux/arm64

The target uses docker buildx to build the images. To build the images:

    docker buildx create --driver docker-container --use
    make docker-buildx
    docker buildx rm

See: https://docs.docker.com/buildx/working-with-buildx/


jefferai:
Do we still need the `docker builds` target with this change? I would assume we'd always want to create and publish both.

tmessi:
Yea, good point. We shouldn't need the old targets

PR #1906, opened 2022-03-11
refactor(controller): Bring in Controller-only listeners

hugoamvieira:
With this change, the Controller is now aware of its own listeners in a
more codified and structured way.

The reasoning behind these changes are as follows:

1. Currently, we're iterating over these listeners in multiple spots
along this initialization flow to check each listeners' purpose. By
doing this once when we create a Controller and essentially storing the
result, we can clean up the downstream code to use those results - This
will be the focus of a later PR.

2. Initial effort to centralize Controller validation. At the moment, we
are doing a lot of validation all over the `cmd` and `config` packages.
From a conceptual perspective, Controller-related validation should
be performed in the function that is responsible for constructing the
Controller object itself. This enables any developer to look in a
single place for all the constraints that we place on creating a
Controller.

### DIFF
--- a/internal/servers/controller/controller.go
+++ b/internal/servers/controller/controller.go
@@ -52,6 +52,9 @@ type Controller struct {
 
 	workerAuthCache *sync.Map
 
+	apiListeners    []*base.ServerListener
+	clusterListener *base.ServerListener
+
 	// Used for testing and tracking worker health
 	workerStatusUpdateTimes *sync.Map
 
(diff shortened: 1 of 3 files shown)


internal/servers/controller/controller.go
```diff
@@ -124,6 +127,32 @@ func New(ctx context.Context, conf *Config) (*Controller, error) {
 		}
 	}
 
+	clusterListeners := make([]*base.ServerListener, 0)
```
jefferai, line 130: I'm assuming that you're pulling the loop iteration out of elsewhere in the code. Can you do it in the same PR? If it happens across two PRs it's much harder to assess whether the code that is migrating here will have the same effect as the code removed elsewhere.

hugoamvieira: Yep, just merged these changes onto another [PR](https://github.com/hashicorp/boundary/pull/1907) where they're being used.

hugoamvieira:
See https://github.com/hashicorp/boundary/pull/1907 for merged changes

PR #1908, opened 2022-03-11
refactor(controller): Decouple listeners/servers shutdown from Worker

hugoamvieira:
In general, these changes were introduced in the same vein of the changes to controller `startListeners`.

Specifically, these changes also decouple the Controller and the Worker shutdown processes. Currently, a call to `stopListeners` on an instance that is running as both a Controller and a Worker would stop the listeners for both.

Lastly, we also add unit/integration tests to the shutdown code.

### DIFF
--- a/internal/servers/controller/listeners.go
+++ b/internal/servers/controller/listeners.go
@@ -9,11 +9,9 @@ import (
 	"net"
 	"net/http"
 	"os"
-	"sync"
 	"time"
 
 	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
-	"github.com/hashicorp/boundary/globals"
 	"github.com/hashicorp/boundary/internal/cmd/base"
 	pbs "github.com/hashicorp/boundary/internal/gen/controller/servers/services"
 	"github.com/hashicorp/boundary/internal/libs/alpnmux"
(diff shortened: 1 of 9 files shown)


internal/servers/controller/listeners.go
```diff
@@ -161,61 +159,108 @@ func (c *Controller) configureForCluster(ln *base.ServerListener) (func(), error
...
+
+	c.clusterListener.GrpcServer.GracefulStop()
+	err := c.clusterListener.Mux.Close()
+	return listenerCloseErrorCheck(c.clusterListener.Config.Type, err)
+}
+
+func (c *Controller) stopHttpServersAndListeners() error {
+	var closeErrors *multierror.Error
+	for i := range c.apiListeners {
+		ln := c.apiListeners[i]
```
johanbrandhorst, line 197: Is there a good reason to use indexing here instead of using a loop variable?

internal/servers/controller/listeners.go
```diff
@@ -161,61 +159,108 @@ func (c *Controller) configureForCluster(ln *base.ServerListener) (func(), error
...
-				<-shutdownKill.Done()
-				c.apiGrpcServer.Stop()
-			}()
-			c.apiGrpcServer.GracefulStop()
-		}()
+		ctx, cancel := context.WithTimeout(c.baseContext, ln.Config.MaxRequestDuration)
+		ln.HTTPServer.Shutdown(ctx)
+		cancel()
+
+		err := ln.Mux.Close() // The HTTP Shutdown call should close this, but just in case.
```
johanbrandhorst, line 206: This error is not used, if we just want to ignore it, what do you think about making that explicit?
```go
_ = ln.Mux.Close
```

internal/servers/controller/listeners.go
```diff
@@ -161,61 +159,108 @@ func (c *Controller) configureForCluster(ln *base.ServerListener) (func(), error
...
+// expected to run at the end of stopServersAndListeners.
+func (c *Controller) stopAnyListeners() error {
+	var closeErrors *multierror.Error
+	for i := range c.apiListeners {
+		ln := c.apiListeners[i]
+		if ln == nil || ln.Mux == nil {
+			continue
+		}
+
+		err := ln.Mux.Close()
```
johanbrandhorst, line 237: Same here

internal/servers/controller/listeners.go
johanbrandhorst: Scratch that, I see we're using the error below! My bad. This is a pattern I hadn't seen before.

internal/servers/controller/listeners.go
hugoamvieira: I think using a loop variable here would work just fine.

I think I got confused because of [this](https://github.com/hashicorp/boundary/blob/main/internal/servers/controller/listeners.go#L174-L175) and not being entirely sure as to why it was done like that. 

The only reason I can think of is because we're using `localLn` in goroutines that spawn as part of a loop, there might be some pointer problems there where using `ln` would change under all goroutines and they'd end up only working on the last element of the slice? I'm not sure this would even be the case.

I know there's some [pitfalls](https://medium.com/codezillas/uh-ohs-in-go-slice-of-pointers-c0a30669feee) when working with slices of pointer types in go though.

internal/tests/cluster/ipv6_listener_test.go
```diff
@@ -77,16 +77,18 @@ func TestIPv6Listener(t *testing.T) {
 	time.Sleep(10 * time.Second)
 	expectWorkers(c1, w1)
 
-	require.NoError(w1.Worker().Shutdown(true))
-	time.Sleep(10 * time.Second)
-	expectWorkers(c1)
+	c2 := c1.AddClusterControllerMember(t, &controller.TestControllerOpts{
```
hugoamvieira, line 80: Tests in this `cluster` folder relied on calling `Shutdown(true)` on a particular controller, doing some assertions and then starting the same controller back up. This is no longer possible due to the changes, so instead we spin up a new controller and perform similar assertions against it.

tmessi (approved):
nice to see these new tests

PR #1909, opened 2022-03-11
refactor(worker): Validate listener configuration in New

hugoamvieira:
With this change, the Worker is now aware of its own listeners in a
more codified and structured way.

The reasoning behind these changes are as follows:

1. Currently, we're iterating over these listeners in multiple spots
along this initialization flow to check each listeners' purpose. By
doing this once when we create a Worker and essentially storing the
result, we can clean up the downstream code to use those results.

2. Represents a continuous effort to centralize validation. At the
moment, we are doing a lot of validation all over the `cmd` and `config`
packages.
Just as mentioned in the analogous commit for the Controller, from a
conceptual perspective, Worker-related validation should be performed in
the function that is responsible for constructing the Worker object
itself. This enables any developer to look in a single place for all the
constraints that we place on creating a Worker.

### DIFF
--- a/internal/cmd/commands/dev/dev.go
+++ b/internal/cmd/commands/dev/dev.go
@@ -611,7 +611,7 @@ func (c *Command) Run(args []string) int {
 
 		if err := c.controller.Start(); err != nil {
 			retErr := fmt.Errorf("Error starting controller: %w", err)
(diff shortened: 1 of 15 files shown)


hugoamvieira:
Merged changes into https://github.com/hashicorp/boundary/pull/1910

PR #1913, opened 2022-03-11
test(worker): Remove test-only hooks from actual code

hugoamvieira:
This change introduces a `nonceFn` field that can be changed
by a test but at runtime is transparent to the code that calls it.

The current nonce-generating logic defaults to a test-only field which
opens us up to this value inadvertently being set.
With this change, the default behavior becomes using `base62` to
generate a random nonce, and any tests have to explicitly change that.


hugoamvieira:
Cherry-picked into https://github.com/hashicorp/boundary/pull/1910

PR #1916, opened 2022-03-12
refactor(plugin): convert grpc error codes into Boundary error codes

A-440Hz:


internal/errors/code.go
```diff
@@ -107,10 +107,15 @@ const (
...
 	VaultInvalidMappingOverride   Code = 3016 // VaultInvalidMappingOverride represents an error returned when a credential mapping is unknown or does not match a credential type
 	VaultInvalidCredentialMapping Code = 3017 // VaultInvalidCredentialMapping represents an error returned when a Vault secret failed to be mapped to a specific credential type
 
 	// OIDC authentication provided errors
 	OidcProviderCallbackError Code = 4000 // OidcProviderCallbackError represents an error that is passed by the OIDC provider to the callback endpoint
+
+	// Plugin specific srrors are codes 5000-5999
+	PluginInvalidParameter   Code = 5000 // PluginInvalidParameter represents a plugin error with an invalid parameter for an operation
+	PluginFailedPrecondition Code = 5001 // PluginFailedPrecondition represents a plugin error where the system is not in the state required for an operation
+	PluginAborted            Code = 5002 // PluginAborted represents a plugin error where a non-native library call was not able to run to completion
```
talanknight, line 120: Could you clarify that these are not errors with the plugin system itself but errors that the plugins report back?  To me just saying "a plugin error" leaves that bit ambiguous.

internal/host/plugin/repository_host_catalog.go
```diff
@@ -122,7 +122,14 @@ func (r *Repository) CreateCatalog(ctx context.Context, c *HostCatalog, _ ...Opt
 			if !pluginCalledSuccessfully {
 				plgResp, err = plgClient.OnCreateCatalog(ctx, &plgpb.OnCreateCatalogRequest{Catalog: plgHc})
 				if err != nil {
-					if status.Code(err) != codes.Unimplemented {
+					switch {
```
talanknight, line 125: Can you add tests to verify these behave as expected?  For the changes in this file and the repository_host_set file.

talanknight (changes requested):
This looks like it is moving our errors for plugins in a good direction.

Could you also provide in the PR a before/after example of what aws/azure invalid arguments error looks like?  This can show us what state we have arrived at in terms of readability of the errors and help us evaluate if we should continue or pause on our path towards more readable errors for plugins.

irenarindos:
Closing as old

PR #1917, opened 2022-03-14
Defines the controller api metrics and registers them at controller startup

talanknight:
This PR creates an internal metric package for the controller.

Currently this package exports 2 things:
1. A method to register the api related collectors to a provided registerer and initialize them with all expected labels to zero
2. A handler for all requests to the controller's api handler

To know which labels are expected, this PR uses the protoregistery and protoreflection to extract out the service method options.  This allows new services and service methods to be added to the protobufs and have them reflected automatically in our metrics.


buf.yaml
```diff
@@ -19,7 +19,7 @@ lint:
       - controller/api/view/v1/view.proto
     FIELD_LOWER_SNAKE_CASE:
       - controller/api/v1/error.proto
-      - controller/custom_options/v1/testing.proto
+      - testing/options/v1/mask_mapping.proto
```
tmessi, line 22: Is this change related to metrics? I think this is causing a log of generated files to change and it is not clear if it belongs in this PR.

talanknight: This change is related in that we are adding testing related proto files for options.  I'm adding new testing related proto files for options and this change makes us more in line with the standard that places it under thest /testing/options/.. path.

tmessi: can this change, and the change to the generated files be done in a separate commit (and/or PR)? so that it is easier to see the changes that are directly relevant for the metrics?

talanknight: This was due to an old version of gofumpt.  I've reverted all the reformatted files.   PTAL.

internal/servers/controller/internal/metrics/api.go
```diff
@@ -0,0 +1,285 @@
...
+	return nil
+}
+
+// apiPathsAndMethods provides the pathing information for all services
+// registered to the controller API.  This does not include any paths that are
+// defined outside of the protobufs such as anything using the OPTION method
+// or any other paths registered like the dev UI passthrough path.
+func apiPathsAndMethods() map[string][]string {
+	// TODO: Auto generate the list of file descriptors so new services created
+	//  will automatically be covered by the metrics.
```
tmessi, line 83: I think this should be easy to do in this PR, you should be able to use [protoregistry.GlobalFiles](https://pkg.go.dev/google.golang.org/protobuf/reflect/protoregistry#Files), something like:

```go
paths := make(map[string][]string)
protoregistry.GlobalFiles.RangeFiles(func(f protoreflect.FileDescriptor) bool {
  if err := gatherServicePathsAndMethods(f, paths); err != nil {
    panic(err)
  }
  return true
})

**Review by tmessi — CHANGES_REQUESTED** [review]
Looking good to me. Just one comment about addressing a TODO.

File: internal/servers/controller/internal/metric/api.go [hunk 3]
```diff
@@ -0,0 +1,284 @@
...
+	return nil
+}
+
+// apiPathsAndMethods provides the pathing information for all services
+// registered to the controller API.  This does not include any paths that are
+// defined outside of the protobufs such as anything using the OPTION method
+// or any other paths registered like the dev UI passthrough path.
+func apiPathsAndMethods() map[string][]string {
+	// TODO: Auto generate the list of file descriptors so new services created
+	//  will automatically be covered by the metrics.
```
tmessi, line 81: This was marked as outdated due to this file moving, but to add the comment again:

I think this should be easy to do in this PR, you should be able to use [protoregistry.GlobalFiles](https://pkg.go.dev/google.golang.org/protobuf/reflect/protoregistry#Files), something like:

```go
paths := make(map[string][]string)
protoregistry.GlobalFiles.RangeFiles(func(f protoreflect.FileDescriptor) bool {
  if err := gatherServicePathsAndMethods(f, paths); err != nil {
    panic(err)
  }
  return true
})
```

talanknight: I wasn't aware of the protoregistery.  Thanks for pointing me to it!

tmessi (approved):

internal/servers/controller/internal/metric/api.go
```diff
@@ -0,0 +1,271 @@
...
+			Namespace: globals.MetricNamespace,
+			Subsystem: apiSubSystem,
+			Name:      "http_request_size_bytes",
+			Help:      "Histogram of request sizes for HTTP requests.",
+			Buckets:   msgSizeBuckets,
+		},
+		[]string{labelHttpCode, labelHttpPath, labelHttpMethod},
+	)
+
+	// httpRequestSize collections measurements of how large each rresponse
```
mgaffney, line 151:
```suggestion
	// httpRequestSize collections measurements of how large each response
```

mgaffney (changes requested):

internal/cmd/base/servers.go
```diff
@@ -74,18 +74,16 @@ type Server struct {
 	CombineLogs bool
 	LogLevel    hclog.Level
 
-	StderrLock *sync.Mutex
-	Eventer    *event.Eventer
+	StderrLock           *sync.Mutex
+	Eventer              *event.Eventer
+	PrometheusRegisterer prometheus.Registerer
```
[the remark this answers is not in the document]
A-440Hz: looks like gofumpt added more tabs to them to get them to line up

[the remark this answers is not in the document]
A-440Hz: Could it make more sense to put the registerer with Listeners?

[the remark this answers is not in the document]
talanknight: Thanks for your comment.  I moved it to be a replacement of the removed go-metrics library that was defined here previously.

internal/servers/controller/internal/metric/api.go
talanknight: Applied this change.

go.mod
```diff
@@ -86,6 +86,8 @@ require (
 
 require github.com/hashicorp/go-dbw v0.0.0-20211215222256-2ff0d37184ff // this is a branch and should be updated before merging
 
+require github.com/prometheus/client_golang v1.4.0
```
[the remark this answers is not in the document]
talanknight: Done.

PR #1921, opened 2022-03-15
fix(session): Only check session status for reported sessions

tmessi:
When processing a worker status report, if the worker did not report any
sessions, the query would find all canceling or terminated sessions,
since the `in` clause would not be populated.
Now this will run the query if there are sessions reported and therefore
there is an `in` clause.


tmessi:
This was merged manually

PR #1928, opened 2022-03-17
internal/session: fix dropped test errors

alrs:
This fixes six dropped `err` variables in `internal/session`.

No changelog updates are warranted.


irenarindos (approved):
Thanks for catching this!

PR #1931, opened 2022-03-17
feat(proto): create custom option for attribute subtype

irenarindos:
- Create Protobuf custom option for subtype attributes
- Update internal/api/genapi/input.go to contain fieldOverrides for each updated inProto. This ensures attributes appears as Attributes and has the correct FieldType
- Update internal/api/genapi/parse.go to detect the new custom option for subtype attributes.


internal/proto/local/controller/custom_options/v1/options.proto
```diff
@@ -15,6 +15,9 @@ extend google.protobuf.FieldOptions {
 	// generate_sdk_option is a directive used when generating the SDK to
 	// indicate that an option should be created for the field
 	bool generate_sdk_option = 85412;
+
+	// attrs_for_type is used for oneOf attribute subtypes
+	string attrs_for_type = 85555;
```
tmessi, line 20: I am thinking we might just want to make this `subtype` since when it is used it will look like:

```proto
[(custom_options.v1.subtype) = "<subtype string>"]
```

irenarindos: Todo: Add more detail to this option comment as work fleshes out

tmessi (approved):

PR #1942, opened 2022-03-23
Update buf version

johanbrandhorst:
Upgrades our buf tool and config versions to v1.

PR #1943, opened 2022-03-23
Update buf version

johanbrandhorst:
Upgrades our buf tool and config versions to v1.

This is an alternative to https://github.com/hashicorp/boundary/pull/1942, which removes `buf` from our dependency closure, and avoiding many dependency updates and potential issues.


tmessi (dismissed):
I like this approach better then #1942, but would also like input from @jefferai and @talanknight  if either of them have an opinion on this.

PR #1944, opened 2022-03-23
Switch to buf for protobuf generation

johanbrandhorst:
This greatly simplifies and speeds up protobuf generation by leveraging `buf`. It also makes it easier for us to consider features like [remote plugins](https://buf.build/blog/remote-plugin-execution) and [managed mode](https://docs.buf.build/generate/managed-mode) which could help further simplify generation.

Removes the need to vendor `googleapis` and `grpc-gateway` protos, instead leveraging dependencies from https://buf.build. Note that this means that the first time a generation is done from a new environment it requires network access. Subsequent generations will not require network access since files are cached locally. If we like, we could cache these in CI, but I don't think it should be necessary initially.

Moves protobuf files up one level since `local` doesn't make sense when there's no `third_party`.

This removes the need for `protoc` in our tools.

This sits on top of #1943.


internal/auth/oidc/request/request.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.27.1
-// 	protoc        v3.17.3
+// 	protoc        (unknown)
```
johanbrandhorst, line 4: This is the only change to all of the generated files, and it's introduced because `buf` does not send any version information to generators.

Makefile
```diff
@@ -172,15 +155,14 @@ protobuild:
...
-
+	@buf generate --template buf.testing.gen.yaml --path internal/proto/testing/event/v1/
 
 	@rm -R ${TMP_DIR}
 
 .PHONY: protolint
 protolint:
 	@buf lint
-	@buf breaking --against 'https://github.com/hashicorp/boundary.git#branch=stable-website'
+	#@buf breaking --against 'https://github.com/hashicorp/boundary.git#branch=stable-website'
```
johanbrandhorst, line 165: This is temporary to test CI generation

johanbrandhorst: TODO: remove before merge

Makefile
```diff
@@ -106,28 +106,11 @@ proto: protolint protobuild
...
-		"--proto_path=internal/proto/local" \
-		"--proto_include_path=internal/proto/third_party" \
-		"--plugin_name=go-grpc" \
-		"--plugin_out=${TMP_DIR}"
-	@bash scripts/protoc_gen_plugin.bash \
-		"--proto_path=internal/proto/local/" \
-		"--proto_include_path=internal/proto/third_party" \
-		"--plugin_name=grpc-gateway" \
-		"--plugin_out=logtostderr=true:${TMP_DIR}"
+	@buf generate -o "${TMP_DIR}"
```
tmessi, line 109: Is this tmp dir needed? I think buf can output to the correct location eliminating the need for the tmp dir and the copy step below

.circleci/config.yml
```diff
@@ -69,19 +69,7 @@ jobs:
     - run:
         command: |-
           make tools
-
```
tmessi, line 72: The changes here don't seem correct. This file is generated by running `make ci-config` in the `.circleci` directory. It is build from the various files in `.circleci/config/`

tmessi (changes requested):

.circleci/config.yml
johanbrandhorst: Ah, thanks for the heads up, I manually changed this, I will run the generation instead, thanks!

Makefile
tmessi: Yea, the `source_relative` option is what I was thinking of, which actually could be done without buf, but I havnt tried it on our files to see how it changes the structure.

.circleci/config.yml
johanbrandhorst: Addressed in https://github.com/hashicorp/boundary/pull/1944/commits/2de86ba9f3ac57e824bf42d3575a62996a5a68e7

internal/proto/buf.yaml
```diff
@@ -1,4 +1,8 @@
 version: v1
+name: buf.build/hashicorp/boundary
```
johanbrandhorst, line 2: This module does not exist on buf.build, but I have reserved the `hashicorp` organization just in case: https://buf.build/hashicorp. There's no need to create this to use the BSR anonymously, but it could be useful if we wanted to share protobuf definitions at some point.

internal/proto/buf.yaml
```diff
@@ -1,4 +1,8 @@
...
+deps:
+  - buf.build/googleapis/googleapis
+  - buf.build/grpc-ecosystem/grpc-gateway
```
johanbrandhorst, line 5: These are modules we depend on, one is maintained by Buf (googleapis) and one is the official grpc-gateway module (grpc-gateway).

talanknight (approved):
This looks great! Thank you for doing this.

tmessi (approved):

talanknight (dismissed):

tmessi (approved):

PR #1945, opened 2022-03-24
Prototype: ignore fields

johanbrandhorst:
This is a prototype of how to hide and add new fields to the openapiv2 spec.

Ticket: https://hashicorp.atlassian.net/browse/ICU-3760


internal/proto/controller/api/resources/accounts/v1/account.proto
```diff
@@ -41,7 +42,10 @@ message Account {
 	string auth_method_id = 90 [json_name="auth_method_id"];
 
 	// The attributes that are applicable for the specific Account type.
-	google.protobuf.Struct attributes = 100 [(custom_options.v1.generate_sdk_option) = true];
+	google.protobuf.Struct attributes = 100 [
+		(custom_options.v1.generate_sdk_option) = true,
+		(google.api.field_visibility).restriction = "INTERNAL"
```
johanbrandhorst, line 47: This is how we can hide fields/services/rpcs/enum values from being rendered.

johanbrandhorst:
The rendered API doc looks fine :)

internal/gen/controller.swagger.json
```diff
@@ -1,226 +1,171 @@
 {
-  "swagger": "2.0",
```
johanbrandhorst, line 2: Unfortunately there does not seem to be a way to avoid the contents of this file being completely rearranged. It isn't important, but it is a bit annoying.

internal/cmd/addattributes/main.go
```diff
@@ -0,0 +1,52 @@
...
+	openapiSpec := doc.Spec()
+
+	for name, definition := range openapiSpec.Definitions {
+		switch name {
+		case "controller.api.resources.accounts.v1.Account":
+			attributesSchema := spec.Schema{}
+			attributesSchema.Type = spec.StringOrArray{"object"}
+			attributesSchema.Description = "The attributes that are applicable for the specific Account type."
+			definition.Properties["attributes"] = attributesSchema
+		}
```
johanbrandhorst, line 41: We'd have one of these cases for each message we need updating

johanbrandhorst:
Investigation concluded, this is possible ✔️

PR #1946, opened 2022-03-24
Prototype: add and remove attributes

johanbrandhorst:
This one adds the attributes back without any third party dependencies


johanbrandhorst:
Not pursuing this

PR #1947, opened 2022-03-25
refact(auth): Move subtype registration into global registry

tmessi:


johanbrandhorst (approved):

hugoamvieira (approved):
Reviewing these specific lines that were changed in all the `*_service.go` files, most of them are really big and seem to do quite a lot. I find them quite dense to read. They also have a lot of repeat calls on them (`GetPublicId` for example).

If you think this is something we should improve, perhaps we could do it as part of this PR since we're touching the code? Either way, this wasn't introduced in this PR so maybe we worry about that some other time.

As for your changes, looks good!

irenarindos (approved):

tmessi:
> Reviewing these specific lines that were changed in all the `*_service.go` files, most of them are really big and seem to do quite a lot. I find them quite dense to read. They also have a lot of repeat calls on them (`GetPublicId` for example).
> 
> If you think this is something we should improve, perhaps we could do it as part of this PR since we're touching the code? Either way, this wasn't introduced in this PR so maybe we worry about that some other time.

I think it could be looked at outside of this PR, I want to keep this as a target refactor to reduce the likelihood of introducing bugs.

tmessi:
including this in #1967

PR #1950, opened 2022-03-25
Post-process add attributes for attr oneof messages

johanbrandhorst:
For https://hashicorp.atlassian.net/browse/ICU-3674


johanbrandhorst:
Not going this direction anymore

PR #1951, opened 2022-03-28
Refactor file formatting target (round 2)

johanbrandhorst:
Use the standard Go heuristic for determining whether a file should be ignored by formatting for being a generated file.

This is effectively a no-op change today, but if we add new generated files in the future, we should automatically ignore them in formatting.

This is a slightly updated version of #1941 that should work across all versions of grep.


talanknight (approved):
Tested it on my MBP and works great.

PR #1954, opened 2022-03-29
Upgrade grpc-gateway version to v2.10.0

johanbrandhorst:
This new version will let us hide fields from the OpenAPI output

### DIFF
--- a/internal/gen/controller.swagger.json
+++ b/internal/gen/controller.swagger.json
@@ -6,49 +6,49 @@
   },
   "tags": [
     {
-      "name": "ScopeService"
+      "name": "controller.api.services.v1.ScopeService"
     },
     {
-      "name": "AccountService"
+      "name": "controller.api.services.v1.AccountService"
     },
     {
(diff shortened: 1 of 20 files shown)


internal/gen/controller.swagger.json
```diff
@@ -6,49 +6,49 @@
   },
   "tags": [
     {
-      "name": "ScopeService"
+      "name": "controller.api.services.v1.ScopeService"
```
johanbrandhorst, line 9: This change does not appear to affect the rendered API docs: https://boundary-cgn9uklnw-hashicorp.vercel.app/api-docs/account-service

internal/gen/controller.swagger.json
```diff
@@ -105,7 +105,7 @@
         },
         "parameters": [
           {
-            "name": "body",
+            "name": "item",
```
johanbrandhorst, line 108: This change also appears not to affect the rendered UI

PR #1952, opened 2022-03-28
refact(account): Use oneOf for Account attributes

irenarindos:
Convert Account service protobuf to use OneOf for attributes

PR #1956, opened 2022-03-29
Move default_attributes back to attributes

johanbrandhorst:
There's no need for this to be a new field name with the new transformation strategy.


internal/cmd/base/profiling_on.go
```diff
@@ -4,11 +4,14 @@
...
 
 import (
+	"context"
 	"os"
 	"path/filepath"
 	"runtime"
 	"runtime/pprof"
 	"time"
+
+	"github.com/hashicorp/boundary/internal/observability/event"
```
johanbrandhorst, line 14: I ran `goimports` on the code base and it turned up this diff, I can move it to a separate PR if we want but I figured it was small enough.

internal/proto/local/controller/api/resources/accounts/v1/account.proto
```diff
@@ -40,13 +40,11 @@ message Account {
...
-	reserved 100;
-	reserved "attributes";
 	// The attributes that are applicable for the specific Account type.
 	oneof attrs {
-		google.protobuf.Struct default_attributes = 101 [(custom_options.v1.generate_sdk_option) = true,(custom_options.v1.subtype) = "default"];
-		PasswordAccountAttributes password_account_attributes = 102 [(custom_options.v1.generate_sdk_option) = true,(custom_options.v1.subtype) = "password"];
-		OidcAccountAttributes oidc_account_attributes = 103 [(custom_options.v1.generate_sdk_option) = true,(custom_options.v1.subtype) = "oidc"];
+		google.protobuf.Struct attributes = 100 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.subtype) = "default"];
+		PasswordAccountAttributes password_account_attributes = 101 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.subtype) = "password"];
+		OidcAccountAttributes oidc_account_attributes = 102 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.subtype) = "oidc"];
```
johanbrandhorst, line 47: A follow up PR will hide these two fields

internal/servers/controller/handlers/accounts/validate_test.go
```diff
@@ -5,15 +5,14 @@ import (
...
-
 	"github.com/hashicorp/boundary/internal/auth/oidc"
 	"github.com/hashicorp/boundary/internal/auth/password"
 	pbs "github.com/hashicorp/boundary/internal/gen/controller/api/services"
+	"github.com/hashicorp/boundary/internal/intglobals"
 	pb "github.com/hashicorp/boundary/sdk/pbs/controller/api/resources/accounts"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
 )
```
johanbrandhorst, line 16: This import ordering was introduced accidentally (I think) in #1952

PR #1957, opened 2022-03-29
Hide internal attribute fields

johanbrandhorst:
These attributes are not supposed to be set directly by the user, so we remove them from the generated OpenAPI spec.


internal/proto/local/controller/api/resources/accounts/v1/account.proto
```diff
@@ -43,8 +44,16 @@ message Account {
 	// The attributes that are applicable for the specific Account type.
 	oneof attrs {
 		google.protobuf.Struct attributes = 100 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.subtype) = "default"];
-		PasswordAccountAttributes password_account_attributes = 101 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.subtype) = "password"];
-		OidcAccountAttributes oidc_account_attributes = 102 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.subtype) = "oidc"];
+		PasswordAccountAttributes password_account_attributes = 101 [
+			(google.api.field_visibility).restriction = "INTERNAL",
+			(custom_options.v1.generate_sdk_option) = true,
+			(custom_options.v1.subtype) = "password"
```
johanbrandhorst, line 50: I added newlines between these field options, I'm happy to revert it back to a single line if we prefer.

internal/proto/third_party/google/api/visibility.proto
```diff
@@ -0,0 +1,111 @@
+// Copyright 2021 Google LLC
```
johanbrandhorst, line 1: This file is necessary unless we move to `buf` for generation (#1944), in which case it'll be removed together with the other vendored files.

johanbrandhorst:
Closed on favor of https://github.com/hashicorp/boundary/pull/1958

PR #1958, opened 2022-03-29
Hide non-public attributes, add go-tags

johanbrandhorst:
For https://hashicorp.atlassian.net/browse/ICU-3670

These attributes are not supposed to be set directly by the user, so we remove them from the generated OpenAPI spec.

Generation is failing because the protobuf changes are technically breaking. We do not worry about the break in this case, as it does not break the JSON API in any way.


internal/proto/local/controller/api/resources/accounts/v1/account.proto
```diff
@@ -14,83 +14,83 @@ import "controller/custom_options/v1/options.proto";
...
-	string id = 10;
+	string id = 10;  // @gotags: `class:"public"`
 
 	// Output only. Scope information for the Account.
-	resources.scopes.v1.ScopeInfo scope = 20;
+	resources.scopes.v1.ScopeInfo scope = 20;  // @gotags: `class:"public"`
 
 	// Optional name for identification purposes.
-	google.protobuf.StringValue name = 30 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = {this:"name" that: "Name"}];
+	google.protobuf.StringValue name = 30 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = {this:"name" that: "Name"}];  // @gotags: `class:"sensitive"`
```
johanbrandhorst, line 23: Sensitive because this might contain PII

internal/proto/local/controller/api/resources/accounts/v1/account.proto
```diff
@@ -14,83 +14,83 @@ import "controller/custom_options/v1/options.proto";
...
 
 	// Optional user-set description for identification purposes.
-	google.protobuf.StringValue description = 40 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = {this:"description" that: "Description"}];
+	google.protobuf.StringValue description = 40 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = {this:"description" that: "Description"}];  // @gotags: `class:"sensitive"`
```
johanbrandhorst, line 26: Sensitive because this might contain PII

internal/proto/local/controller/api/resources/accounts/v1/account.proto
```diff
@@ -14,83 +14,83 @@ import "controller/custom_options/v1/options.proto";
...
 	// Output only. The available actions on this resource for this user.
-	repeated string authorized_actions = 300 [json_name="authorized_actions"];
+	repeated string authorized_actions = 300 [json_name="authorized_actions"];  // @gotags: `class:"public"`
 }
 
 // Attributes associated only with Accounts with type "password".
 message PasswordAccountAttributes {
 	// The login name of this Account. This is unique per Auth Method.
-	string login_name = 10 [json_name="login_name", (custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = {this:"attributes.login_name" that: "LoginName"}];
+	string login_name = 10 [json_name="login_name", (custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = {this:"attributes.login_name" that: "LoginName"}];  // @gotags: `class:"sensitive"`
```
johanbrandhorst, line 69: Not sure if this should be sensitive, is this user controlled? Could also be secret perhaps?

internal/proto/local/controller/api/resources/accounts/v1/account.proto
```diff
@@ -14,83 +14,83 @@ import "controller/custom_options/v1/options.proto";
...
 
 	// The password for this Account.
-	google.protobuf.StringValue password = 20 [(custom_options.v1.generate_sdk_option) = true];
+	google.protobuf.StringValue password = 20 [(custom_options.v1.generate_sdk_option) = true];  // @gotags: `class:"secret"`
```
johanbrandhorst, line 72: Secret because duh

internal/proto/local/controller/api/resources/accounts/v1/account.proto
```diff
@@ -14,83 +14,83 @@ import "controller/custom_options/v1/options.proto";
...
-	string email = 110;
+	string email = 110;  // @gotags: `class:"sensitive"`
 
 	// Output only. token_claims are the marshaled claims from the token.
-	google.protobuf.Struct token_claims = 120;
+	google.protobuf.Struct token_claims = 120;  // @gotags: `class:"sensitive"`
 
 	// Output only. userinfo_claims are the marshaled claims from userinfo.
-	google.protobuf.Struct userinfo_claims = 130;
+	google.protobuf.Struct userinfo_claims = 130;  // @gotags: `class:"sensitive"`
```
johanbrandhorst, line 95: I marked all of these as sensitive because they could contain PII. Do we consider `subject` a secret?

internal/cmd/base/profiling_on.go
```diff
@@ -4,11 +4,14 @@
...
 
 import (
+	"context"
 	"os"
 	"path/filepath"
 	"runtime"
 	"runtime/pprof"
 	"time"
+
+	"github.com/hashicorp/boundary/internal/observability/event"
```
johanbrandhorst, line 14: I ran goimports on the code base and it turned up this diff, I can move it to a separate PR if we want but I figured it was small enough.

internal/proto/local/controller/api/resources/accounts/v1/account.proto
```diff
@@ -7,83 +7,90 @@ option go_package = "github.com/hashicorp/boundary/sdk/pbs/controller/api/resour
...
+		PasswordAccountAttributes password_account_attributes = 101 [
+			(google.api.field_visibility).restriction = "INTERNAL",
+			(custom_options.v1.generate_sdk_option) = true,
+			(custom_options.v1.subtype) = "password"
+		];  // @gotags: `class:"public"`
+		OidcAccountAttributes oidc_account_attributes = 102 [
+			(google.api.field_visibility).restriction = "INTERNAL",
+			(custom_options.v1.generate_sdk_option) = true,
+			(custom_options.v1.subtype) = "oidc"
+		];  // @gotags: `class:"public"`
```
johanbrandhorst, line 56: I added newlines between these field options, I'm happy to revert it back to a single line if we prefer.

internal/proto/third_party/google/api/visibility.proto
```diff
@@ -0,0 +1,111 @@
+// Copyright 2021 Google LLC
```
johanbrandhorst, line 1: This file is necessary unless we move to buf for generation (https://github.com/hashicorp/boundary/pull/1944), in which case it'll be removed together with the other vendored files.

johanbrandhorst:
I've re-targeted this against `llb-audit-events`, so this is the only PR to review for the account proto changes.

internal/proto/local/controller/api/services/v1/account_service.proto
```diff
@@ -114,68 +114,68 @@ service AccountService {
...
 message GetAccountResponse {
-  resources.accounts.v1.Account item = 1;
+  resources.accounts.v1.Account item = 1;  // @gotags: `class:"public"`
 }
 
 message ListAccountsRequest {
-  string auth_method_id = 1 [json_name="auth_method_id"];
-  string filter = 30 [json_name="filter"];
+  string auth_method_id = 1 [json_name="auth_method_id"];  // @gotags: `class:"public"`
+  string filter = 30 [json_name="filter"];  // @gotags: `class:"public"`
```
johanbrandhorst, line 126: Does this risk containing any PII?

internal/proto/local/controller/api/resources/accounts/v1/account.proto
tmessi: I think this should be sensitive

internal/servers/controller/handlers/accounts/validate_test.go
```diff
@@ -5,15 +5,14 @@ import (
...
-	"github.com/hashicorp/boundary/internal/intglobals"
-	"google.golang.org/protobuf/types/known/fieldmaskpb"
-
 	"github.com/hashicorp/boundary/internal/auth/oidc"
 	"github.com/hashicorp/boundary/internal/auth/password"
 	pbs "github.com/hashicorp/boundary/internal/gen/controller/api/services"
+	"github.com/hashicorp/boundary/internal/intglobals"
 	pb "github.com/hashicorp/boundary/sdk/pbs/controller/api/resources/accounts"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
```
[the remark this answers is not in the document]
johanbrandhorst: Sounds good

internal/cmd/base/profiling_on.go
johanbrandhorst: Created a new PR for this: https://github.com/hashicorp/boundary/pull/1960

internal/servers/controller/handlers/accounts/validate_test.go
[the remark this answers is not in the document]
johanbrandhorst: This is now part of e21e1868da4e9adc91c16f28606a13785589eae5

internal/proto/local/controller/api/services/v1/account_service.proto
johanbrandhorst: Moved this to sensitive in a89463e4a720bbf62c982772216124300d0ecdf7

tmessi (approved):

mgaffney (approved):

PR #1961, opened 2022-03-30
Initialize and provide handler for proxy worker http metrics

A-440Hz:


internal/cmd/ops/server.go
```diff
@@ -140,7 +140,7 @@ func createOpsHandler(lncfg *listenerutil.ListenerConfig, c *controller.Controll
 		mux.Handle("/health", h)
 	}
 
-	mux.Handle("/metrics", promhttp.Handler())
+	mux.Handle("/metric", promhttp.Handler())
```
hugoamvieira, line 143: This should still be `/metrics` as it's the default for a Prometheus endpoint

internal/servers/worker/internal/metric/proxy_http.go
```diff
@@ -0,0 +1,105 @@
...
+
+// httpTimeUntilHeader collects measurements of how long it takes
+// the boundary system to hijack an HTTP request into a websocket connection
+// for the proxy worker.
+var httpTimeUntilHeader prometheus.ObserverVec = prometheus.NewHistogramVec(
+	prometheus.HistogramOpts{
+		Namespace: globals.MetricNamespace,
+		Subsystem: proxySubSystem,
+		Name:      "http_write_header_duration_seconds",
+		Help:      "Histogram of latencies for HTTP to websocket conversions.",
```
talanknight, line 35: This help text isn't accurate any more.  It should state that this is a histogram of the latency from the time the request is received (post TLS handshake) to when the first http header is written back from the server.

internal/servers/controller/internal/metric/api.go
```diff
@@ -227,7 +227,7 @@ func InstrumentApiHandler(wrapped http.Handler) http.Handler {
 		promhttp.InstrumentHandlerDuration(
 			httpRequestLatency.MustCurryWith(l),
 			promhttp.InstrumentHandlerRequestSize(
-				httpResponseSize.MustCurryWith(l),
```
talanknight, line 230: Looks like you need to update the PR since this is already reflected in the main branch.

internal/servers/controller/testing.go
```diff
@@ -424,7 +424,6 @@ func NewTestController(t *testing.T, opts *TestControllerOpts) *TestController {
 		cancel: cancel,
 		opts:   opts,
 	}
-
```
talanknight, line 427: Unless this is added through `make fmt` you should avoid making these types of changes as it pollutes the files history and git blame.

talanknight (changes requested):
Great work!  I just had a minor change related to help text that I think is important so it is clear the metric being tracked.

internal/servers/controller/testing.go
A-440Hz: I was using the merge feature of my IDE and have no idea what is going on here

talanknight (approved):

PR #1963, opened 2022-03-31
feat(worker): Implement proxy handler websocket instrumentation

hugoamvieira:
Add metrics on read/written bytes and active worker websocket connections.


internal/servers/worker/internal/metric/proxy_ws.go
```diff
@@ -0,0 +1,118 @@
...
+	activeConnCount = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: globals.MetricNamespace,
+			Subsystem: proxyWebsocketSubsystem,
+			Name:      "active_connection_count",
+			Help:      "Count of open websocket connections to Boundary workers.",
+		},
+	)
+
+	bytesReceived = prometheus.NewHistogram(
```
talanknight, line 28: FYI, the bytes sent/received should be a counter, not a histogram.  You can see the relevant part of the RFC, but in summary, histograms make sense if there is a natural observation time to record the observation for.  Since this is a continual stream, that doesn't exist here.  Instead a counter will be able to track the total number of bytes being sent over this worker.

hugoamvieira: > ... histograms make sense if there is a natural observation time to record the observation for. Since this is a continual stream, that doesn't exist here.

Didn't come across my mind but yeah we're working with websockets in this case - great point!

internal/servers/worker/internal/metric/proxy_ws.go
```diff
@@ -0,0 +1,118 @@
...
+type wsConn struct {
+	net.Conn
+}
+
+func (wc *wsConn) Read(b []byte) (int, error) {
+	n, err := wc.Conn.Read(b)
+	if err != nil {
+		return n, err
+	}
+	bytesReceived.Observe(float64(len(b))) // The slice is always at least 4096 bytes long (`bufio.ReadWriter` default)
```
hugoamvieira, line 75: This is a noteworthy point that should be discussed in this RFC.
The byte slice len will be at least `4096`. It'll consist of the actual data and the rest (if not filled up by the data itself) padded by zeroes.

It's probably worth to discuss if we should remove that padding for our counter as we'll be over-reporting received bytes if we don't.

internal/servers/worker/internal/metric/proxy_ws.go
hugoamvieira: Addressed in [7574f02](https://github.com/hashicorp/boundary/pull/1963/commits/7574f02e4e09da0681804e529875258b6db071cd)

internal/servers/worker/internal/metric/proxy_ws.go
```diff
@@ -0,0 +1,123 @@
...
+	*sync.Once
+	net.Conn
+}
+
+func (wc *wsConn) Read(b []byte) (int, error) {
+	n, err := wc.Conn.Read(b)
+	if err != nil {
+		return n, err
+	}
+	bytesReceived.Add(float64(len(b))) // The slice is always at least 4096 bytes long (`bufio.ReadWriter` default)
```
hugoamvieira, line 73: This is a noteworthy point that should be discussed in this RFC.
The byte slice len will be at least 4096. It'll consist of the actual data and the rest (if not filled up by the data itself) padded by zeroes.

It's probably worth to discuss if we should remove that padding for our counter as we'll be over-reporting received bytes if we don't.

talanknight (changes requested):
I haven't gotten to the tests yet.

I have a few concerns about buffering writes which I am happy to discuss more or discover if there is anything to be concerned with.

internal/servers/worker/internal/metric/proxy_ws.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	c = &wsConn{Conn: c, Once: &mw.dec}
+
+	// Our underlying `Hijack` call returns a `bufio.ReadWriter` object
+	// with the `Reader` and `Writer` set to the `net.Conn` object *before* we wrap it.
+	// The websocket library uses this buffer to Read and Write. This means
+	// our Conn implementation would never be called.
+	// You might have noticed we're not replacing the Reader. This is
+	// because the Websocket library already does this in `Accept`.
+	rw = bufio.NewReadWriter(rw.Reader, bufio.NewWriter(c))
```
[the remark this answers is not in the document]
hugoamvieira: I implemented it like this because [this is how it's done in the HTTP library](https://github.com/golang/go/blob/1a955bcdb701d788b048a39d7273621729e257bc/src/net/http/server.go#L310-L329=), so to the extent that it hasn't been a problem until now, I don't foresee this being a problem in the future.

I didn't want to break the pattern just in case I failed to understand some of this. I did try using `c` directly and that also works.

It's probably worth to mention that from my testing `c` is always of type `tls.Conn` since we're always talking via TLS with the Worker. `tls.Conn` has it's own `Write` which [**can** buffer writes at this level](https://github.com/golang/go/blob/master/src/crypto/tls/conn.go#L913-L922=), but it looks like it only does so during handshakes (for both [client](https://github.com/golang/go/blob/master/src/crypto/tls/handshake_client.go#L402=) and [server](https://github.com/golang/go/blob/master/src/crypto/tls/handshake_server.go#L73=) connections). 

The underlying `net.Conn` inside `tls.Conn` seems to just be one of the implementations that come from the `net` package (eg. TCPConn), which seem to eventually end up in `rawConn` methods, which [only write onto the raw connection with no buffering](https://github.com/golang/go/blob/master/src/net/rawconn.go#L51-L61=).

internal/servers/worker/internal/metric/proxy_ws.go
hugoamvieira: 🤦 I was so focused on this problem that I literally forgot the data was right there. Thanks for the great spot.

internal/servers/worker/internal/metric/proxy_ws.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil, nil, err
+	}
+	c = &wsConn{Conn: c, Once: &mw.dec}
+
+	// Our underlying `Hijack` call returns a `bufio.ReadWriter` object
+	// with the `Reader` and `Writer` set to the `net.Conn` object *before* we wrap it.
+	// The websocket library uses this buffer to Read and Write. This means
+	// our Conn implementation would never be called.
+	// You might have noticed we're not replacing the Reader. This is
+	// because the Websocket library already does this in `Accept`.
```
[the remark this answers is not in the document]
hugoamvieira: It always does this in `Accept` as a [bugfix, apparently](https://github.com/nhooyr/websocket/blob/master/accept.go#L141-L143=). 

We could wrap the incoming object in a `bytes.NewReader` but [it's already a `bufio.Reader`](https://github.com/golang/go/blob/65153e478e302eaf3556372ff257ebfc893943c1/src/net/http/server.go#L321=) from our underlying `Hijack` call and I'm not sure if we'd have any side effects from wrapping it, just to be wrapped once again in `Accept`.

internal/servers/worker/internal/metric/proxy_ws.go
hugoamvieira: Addressed in [6156a64](https://github.com/hashicorp/boundary/pull/1963/commits/6156a64961d22cce2c58ec9be30f5e06e0683237)

internal/servers/worker/internal/metric/proxy_ws.go
```diff
@@ -0,0 +1,123 @@
...
+			Help:      "Count of open websocket connections to Boundary workers.",
+		},
+	)
+
+	bytesReceived = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Namespace: globals.MetricNamespace,
+			Subsystem: proxyWebsocketSubsystem,
+			Name:      "received_bytes_total",
+			Help:      "Count of received bytes for Worker connections.",
```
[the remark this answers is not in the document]
hugoamvieira: Addressed in [6156a64](https://github.com/hashicorp/boundary/pull/1963/commits/6156a64961d22cce2c58ec9be30f5e06e0683237)

internal/servers/worker/internal/metric/proxy_ws.go
```diff
@@ -0,0 +1,123 @@
...
+	proxyWebsocketSubsystem = "worker_proxy_websocket"
+)
+
+var (
+	activeConnCount = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: globals.MetricNamespace,
+			Subsystem: proxyWebsocketSubsystem,
+			Name:      "active_connections",
+			Help:      "Count of open websocket connections to Boundary workers.",
```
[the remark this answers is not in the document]
hugoamvieira: Addressed in [6156a64](https://github.com/hashicorp/boundary/pull/1963/commits/6156a64961d22cce2c58ec9be30f5e06e0683237)

internal/servers/worker/internal/metric/proxy_ws.go
```diff
@@ -0,0 +1,123 @@
...
+		prometheus.CounterOpts{
+			Namespace: globals.MetricNamespace,
+			Subsystem: proxyWebsocketSubsystem,
+			Name:      "sent_bytes_total",
+			Help:      "Count of sent bytes for Worker connections.",
+		},
+	)
+)
+
+func MustInitializeProxyMetrics(r prometheus.Registerer) {
```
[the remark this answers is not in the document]
hugoamvieira: Addressed in [6156a64](https://github.com/hashicorp/boundary/pull/1963/commits/6156a64961d22cce2c58ec9be30f5e06e0683237)

internal/servers/worker/internal/metric/proxy_ws.go
```diff
@@ -0,0 +1,123 @@
...
+	if r == nil {
+		return
+	}
+	r.MustRegister(activeConnCount, bytesReceived, bytesSent)
+}
+
+func ProxyWsMetricsHandler(next http.Handler) http.Handler {
```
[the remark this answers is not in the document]
hugoamvieira: Addressed in [6156a64](https://github.com/hashicorp/boundary/pull/1963/commits/6156a64961d22cce2c58ec9be30f5e06e0683237)

internal/servers/worker/handler.go
```diff
@@ -41,9 +41,11 @@ func (w *Worker) handler(props HandlerProperties) (http.Handler, error) {
 	}
 	mux.Handle("/v1/proxy", h)
 
-	genericWrappedHandler := w.wrapGenericHandler(mux, props)
-	metricHandler := metric.InstrumentProxyHttpHandler(genericWrappedHandler)
-	return metricHandler, nil
+	handler := w.wrapGenericHandler(mux, props)
+	handler = metric.ProxyWsMetricsHandler(handler)
+	handler = metric.InstrumentProxyHttpHandler(handler)
```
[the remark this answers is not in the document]
hugoamvieira: Addressed in [6156a64](https://github.com/hashicorp/boundary/pull/1963/commits/6156a64961d22cce2c58ec9be30f5e06e0683237)

talanknight (changes requested):
Took a look at the tests and they look good generally.

Just a comment on table driven tests: They really shine when you have a variety of inputs and need to assert a specific output.  There are a few situations I noticed in this PR which use the table, not for input/output assertions, but to navigate what parts of the test code gets run (through conditionals or for loops) which adds to the cognitive load and makes it harder to use the tests for documentation purposes.

internal/servers/worker/internal/metric/proxy_ws.go
```diff
@@ -0,0 +1,123 @@
...
+	bytesSent.Add(float64(len(b)))
+	return n, nil
+}
+
+func (wc *wsConn) Close() error {
+	wc.Once.Do(func() { activeConnCount.Dec() })
+	return wc.Conn.Close()
+}
+
+type metricsWrapper struct {
```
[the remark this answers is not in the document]
hugoamvieira: Addressed in [48ded61](https://github.com/hashicorp/boundary/pull/1963/commits/48ded61625d3c1a8f9499b267d54f71af78e0dc0)

internal/servers/worker/internal/metric/proxy_ws_test.go
```diff
@@ -0,0 +1,419 @@
...
+type testHttpHijacker struct {
+	// Mocks the underlying http.Hijacker in metricsWrapper.
+	http.ResponseWriter
+	t             *testing.T
+	retConn       net.Conn
+	retErr        error
+	hijackCalledN int
+}
+
+func (h *testHttpHijacker) Hijack() (net.Conn, *bufio.ReadWriter, error) {
```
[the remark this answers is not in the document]
hugoamvieira: > Does the testHttpHijacker you wrote here behave close to the one used in reality?

This `testHttpHijacker` returns a connection along with a newly created `bufio.ReadWriter`, just like the real deal. To that extent, I'd say that it does behave like the real one.

> I think it's possible for the bufio.ReadWriter to have a non empty reader at the time Hijack returns.

This is correct - There can be data in the buffer.

> Would that impact our code at all and should that be covered in our tests?

I don't think so because we're using the incoming reader directly.

We could write a test where our test hijacker returns some data in the reader and we could then verify that the data is present in the ReadWriter we create and return?

internal/servers/worker/internal/metric/proxy_ws_test.go
```diff
@@ -0,0 +1,419 @@
...
+				err = brw.Flush()
+				require.NoError(t, err)
+				require.True(t, tt.underlyingConn.writeCalled)
+			}
+		})
+	}
+}
+
+func TestWsConnRead(t *testing.T) {
+	tests := []struct {
```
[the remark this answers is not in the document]
hugoamvieira: Addressed in [3a96b61](https://github.com/hashicorp/boundary/pull/1963/commits/3a96b61445db999d22924e43569aef16480a4707)

internal/servers/worker/internal/metric/proxy_ws_test.go
```diff
@@ -0,0 +1,419 @@
...
+		expErr                bool
+		expErrStr             string
+	}{
+		{
+			name:                  "underlying hijacker error",
+			underlyingHijackerErr: fmt.Errorf("oops!"),
+			underlyingConn:        nil,
+			hijackCallCount:       1,
+			expErr:                true,
+			expErrStr:             "oops!",
```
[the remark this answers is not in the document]
hugoamvieira: Addressed in [3b9338e](https://github.com/hashicorp/boundary/pull/1963/commits/3b9338ec9e2ac0601c1cf10bc1e0e85e2d5af0b4)

internal/servers/worker/internal/metric/proxy_ws_test.go
```diff
@@ -0,0 +1,419 @@
...
+			name:                  "no underlying error, multiple calls",
+			underlyingHijackerErr: nil,
+			underlyingConn:        &testConn{t: t, readEOF: true, retWriteLen: true},
+			hijackCallCount:       10,
+			expErr:                false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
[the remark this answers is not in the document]
hugoamvieira: I've rewritten the test cases to eliminate table tests - Let me know what you think!

internal/servers/worker/internal/metric/proxy_ws.go
```diff
@@ -0,0 +1,127 @@
...
+	// The `Once` objects constrain incrementing/decrementing active connections.
+	// Only a precautionary measure as:
+	// 1. http.(Hijacker).Hijack() cannot be called more than once (2nd call is an error).
+	// 2. websocket.(*Conn).Close() cannot be called more than once (2nd call is no-op).
+	inc sync.Once
+	dec sync.Once
+
+	// `ServeHTTP` takes an `http.ResponseWriter`, so `metricsWrapper`
+	// needs to implement it. Other than that, `ResponseWriter` is unused.
+	http.ResponseWriter
```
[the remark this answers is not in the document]
hugoamvieira: I'm not familiar with this pattern, but we only wrap `http.ResponseWriter` if it is `http.Hijacker`

internal/servers/worker/internal/metric/proxy_ws_test.go
```diff
@@ -0,0 +1,371 @@
...
+
+		mw := rw.(*hijackerWrapper)
+		require.NotNil(t, mw.Hijacker)
+		require.NotNil(t, mw.ResponseWriter)
+	})
+
+	rw := &testHttpHijacker{ResponseWriter: httptest.NewRecorder(), t: t}
+	req := httptest.NewRequest("GET", "/v1/proxy", nil)
+	wrapped := InstrumentWebsocketWrapper(h)
+	wrapped.ServeHTTP(rw, req)
```
talanknight, line 48: Please also assert that the request reaches `h`.  Here and the other `TestHandlerWrapper` test.

internal/servers/worker/internal/metric/proxy_ws_test.go
```diff
@@ -0,0 +1,371 @@
...
+}
+
+func TestHandlerWrapperWithNoHijacker(t *testing.T) {
```
talanknight, line 51: nit: There is a semi-standard for naming tests to have variations on a test use the naming `TestX_VariationName`.   See [the](https://github.com/hashicorp/boundary/blob/main/internal/auth/password/argon2_test.go#L15) [argon2](https://github.com/hashicorp/boundary/blob/main/internal/auth/password/argon2_test.go#L94) [tests](https://github.com/hashicorp/boundary/blob/main/internal/auth/password/argon2_test.go#L161) or [these](https://github.com/hashicorp/boundary/blob/main/internal/auth/password/repository_account_test.go#L264) [repository](https://github.com/hashicorp/boundary/blob/main/internal/auth/password/repository_account_test.go#L42) [tests](https://github.com/hashicorp/boundary/blob/main/internal/auth/password/repository_authmethod_test.go#L165) [for](https://github.com/hashicorp/boundary/blob/main/internal/auth/password/repository_authmethod_test.go#L237) [example](https://github.com/hashicorp/boundary/blob/main/internal/auth/password/repository_authmethod_test.go#L23).

Here and for any other tests in this file which is a variation on another (like `TestHijacker_Error`).

internal/servers/worker/internal/metric/proxy_ws_test.go
```diff
@@ -0,0 +1,371 @@
...
+	require.NotPanics(t, func() { InitializeWebsocketCollectors(nil) })
+
+	// Try with an actual Registerer.
+	r := prometheus.NewRegistry()
+	require.NotPanics(t, func() { InitializeWebsocketCollectors(r) })
+
+	// Expect that some metrics exist.
+	f, err := r.Gather()
+	require.NoError(t, err)
+	require.Greater(t, len(f), 0)
```
talanknight, line 29: I notice you are using `require` exclusively in this file.  My understanding is that require should be used where progressing would cause a panic thus making it impossible to run subsequent tests.

`assert` should be used everywhere else since it is helpful to see all the assertions that are passing and failing with a single run if possible so the same test doesn't need to be run multiple times while fixes to assertions are made since `require` will stop the tests at the first require failure and thus make it harder to see all the other places assertions may be failing.

internal/servers/worker/internal/metric/proxy_ws.go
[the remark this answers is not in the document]
talanknight: The thing is that we wrap the http.Hijacker interface only so if any downstream code relies on http.Pusher or http.Flusher, the ResponseWriter we pass onto it will never satisfy those interfaces.  Happy to discuss this further.

talanknight (changes requested):

internal/servers/worker/internal/metric/proxy_ws.go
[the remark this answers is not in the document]
hugoamvieira: Ahhh I see - got it now, thanks for explaining 👍

internal/servers/worker/internal/metric/proxy_ws_test.go
hugoamvieira: Not sure I understood this - We [check if the request is not nil inside `h`](https://github.com/hashicorp/boundary/pull/1963/files/d75c3db7450006ce4c69bd97933efa1e1655688a#diff-efaa445d723714b9a9380282573e7f82a3cde48812571f9ba9aada7b189c86fdR34) - Did you have something different in mind here?

internal/servers/worker/internal/metric/proxy_ws_test.go
hugoamvieira: Addressed [37e6be3](https://github.com/hashicorp/boundary/pull/1963/commits/37e6be3f4008efe6f06582586c94c5b8a250fbf8)

internal/servers/worker/internal/metric/proxy_ws_test.go
hugoamvieira: I'm not totally sure what the pattern is in Boundary but in past PRs I've consistently only used `require` in tests. I suspect that so far this might be down to personal preference as I've seen `assert`-only test files, `require`-only test files, and anything in between.

Historically, I typically use `require` exclusively because when we write table tests the `t *testing.T` object inside the `t.Run` function is scoped exclusively for that particular test run. This means the entire overarching test does not abort if one test case fails - all test cases will still run (unless there's a panic somewhere of course).

For me, in general, using `assert` introduces more complexity to the tests (both to read and to write) as now I have to:
1. As the reader, understand why both `assert` and `require` were used and why they are used in the places they are;
2. As the writer, for each `assert` statement, understand how the overall test would continue in case the assertion was successful and unsuccessful. In cases where the assertion is complex (eg: asserting struct equality/struct value equality) this is not trivial. Many fields can be the cause for failure and I don't know which so I have to account for all in my following test code. Even worse, as assertions mount I have to keep in mind all the possible states from all the combinations of previous assertions.

With `require`, the test fails exactly where the assertion failed. Everything above is OK, and everything below is unknown. That's the only line I have to look at to start debugging process.

For me, while it is true that it does require that I run a particular test multiple times and this can be annoying (even for me!), I value the simplicity it affords more than any temporary annoyances that I might experience

I'm happy to discuss this further though if you think we could improve here! 😄

internal/servers/worker/internal/metric/proxy_ws_test.go
hugoamvieira: Ah, I see what you mean now - We don't assert that `h` was called in the first place. Good point!

hugoamvieira: Addressed [7b9e229](https://github.com/hashicorp/boundary/pull/1963/commits/7b9e229dbf34e3c573d4bc39f4e3d67872402cc4)

internal/servers/worker/internal/metric/proxy_ws.go
```diff
@@ -0,0 +1,155 @@
...
+	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		nextRw := rw
+		if rwh, ok := rw.(http.Hijacker); ok {
+			nextRw = wrapProxyHandler(&hijackerWrapper{ResponseWriter: rw, Hijacker: rwh}, rw)
+		}
+		next.ServeHTTP(nextRw, req)
+	})
+}
+
+func wrapProxyHandler(with *hijackerWrapper, wrap http.ResponseWriter) http.ResponseWriter {
```
talanknight, line 67: Since it wasn't immediately intuitive why using this function is helpful, consider adding an explanation for why this function exist (so downstream code can cast to these interfaces).

internal/servers/worker/internal/metric/proxy_ws.go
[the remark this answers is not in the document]
talanknight: I think we should take a defensive approach here.  If the referenced bug gets resolved in some way which causes that referenced bugfix line to be removed the websocket library may rely only on the bufio.Reader we return from this function.  That would bypass our connection (and thus the measurements we add to it) and our code would no longer be correct.

I think the solution is to put into place the same thing the websocket code put into place:

```
b, _ := brw.Reader.Peek(brw.Reader.Buffered())
rw.Reader.Reset(io.MultiReader(bytes.NewReader(b), netConn))
```
which ensures the reader is based off of our `wsConn` and not the underlying reader which is based off of the initial, unwrapped connection. 


Please let me know if I have misunderstood how the readers and connections interact.

internal/servers/worker/internal/metric/proxy_ws.go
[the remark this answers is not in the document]
talanknight: Thank you, that helps clarify things a great deal.

internal/servers/worker/internal/metric/proxy_ws_test.go
```diff
@@ -0,0 +1,450 @@
...
+
+	wsConn := c.(*wsConn)
+	require.Equal(t, underlyingConn, wsConn.Conn) // Underlying conn must be our testConn.
+	require.Equal(t, &mw.dec, wsConn.Once)        // wsConn.Once must be set to a ptr to `dec`.
+
+	// Assert that the buffer Reader and Writer are backed up
+	// by the underlying Conn.
+	_, err = io.ReadAll(brw.Reader)
+	require.NoError(t, err)
+	require.True(t, underlyingConn.readCalled)
```
talanknight, line 170: In addition to read being called on the underlying connection, we want to make sure read is called on wsConn or else we won't be measuring what we should be.   Same thing for write.

talanknight (changes requested):
Looks great!  Just a few more comments.

internal/servers/worker/internal/metric/proxy_ws_test.go
hugoamvieira: I'm unsure how we'd test that `wsConn.Read` has been called without having test fields in `wsConn`, since it's the object under test and thus cannot be mocked.

internal/servers/worker/internal/metric/proxy_ws.go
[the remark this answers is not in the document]
hugoamvieira: I think this is reasonable - As I said, I'm not sure there are any side effects with doing this, but I'll perform some tests

internal/servers/worker/internal/metric/proxy_ws_test.go
hugoamvieira: Addressed [5954753](https://github.com/hashicorp/boundary/pull/1963/commits/59547537e4a5be499d98f5d521b102ab2e057702)

internal/servers/worker/internal/metric/proxy_ws.go
hugoamvieira: [5954753](https://github.com/hashicorp/boundary/pull/1963/commits/59547537e4a5be499d98f5d521b102ab2e057702)

internal/servers/worker/internal/metric/proxy_ws.go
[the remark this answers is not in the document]
hugoamvieira: Addressed [5954753](https://github.com/hashicorp/boundary/pull/1963/commits/59547537e4a5be499d98f5d521b102ab2e057702)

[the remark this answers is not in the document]
hugoamvieira: (I also did a bunch of testing and saw no changes from what we had to what we have now, behavior wise, so it seems like this fine)

internal/servers/worker/internal/metric/proxy_ws_test.go
talanknight: FYI: I just learned about the problem with using require (ie. FailNow) in goroutines.  That isn't relevant here since there are no go routines, but it was something I wasn't aware of and thought I would share :).  See [this](https://github.com/hashicorp/boundary/pull/1978)

talanknight (dismissed):
Looks great!

PR #1964, opened 2022-03-31
Add collector updating interceptor to the controller cluster grpc server

talanknight:
The newly written interceptor measures latency, request size and response size, assuming they are sent and received over proto wireformat (currently they are).

If we are unable to cast the request or response to a proto.Message we record an error on the eventor, but note that this interceptor runs before the interceptor which adds the eventor to the context.  The intent is to capture as much of the actual duration of the request as possible, including any logic that might be added by the eventor related interceptor.


internal/servers/controller/internal/metric/cluster.go
```diff
@@ -0,0 +1,177 @@
...
+		[]string{labelGRpcCode, labelGRpcService, labelGRpcMethod},
+	)
+)
+
+func statusFromError(err error) *status.Status {
+	if s, ok := status.FromError(err); ok {
+		return s
+	}
+	return status.New(codes.Unknown, "Unknown Code")
+}
```
johanbrandhorst, line 78: This is almost identical to https://pkg.go.dev/google.golang.org/grpc/status#Convert (the difference is in the message for the unknown status), do we need our own function for this?

internal/servers/controller/internal/metric/cluster.go
```diff
@@ -0,0 +1,177 @@
...
+
+	if respProto, ok := resp.(proto.Message); ok {
+		respSize := proto.Size(respProto)
+		gRpcResponseSize.With(r.labels).Observe(float64(respSize))
+	} else {
+		event.WriteError(ctx, op, errors.New(ctx, errors.Internal, op, "unable to cast to proto.Message"))
+	}
+}
+
+// InstrumentClusterInterceptor wraps a UnaryServerInterceptor and measures
```
johanbrandhorst, line 137: I think this sentence might be accidentally cut short but I also thing it might not be, sorry if this is intentional 😅.

talanknight: Updated the comment to hopefully be more complete.  Thank you!

johanbrandhorst (approved):

hugoamvieira (dismissed):

A-440Hz (approved):
LGTM

PR #1966, opened 2022-04-01
Put session ID in ALPN

jefferai:
This allows us to use SNI for actual host (e.g. routing) information. We
don't have the client cert yet so we can't look at that, so this
provides a way for us to convey the information needed to look up that
session's TLS stack.

Using SNI for hosts means we also run into the fact that we don't have
automatic agreement in terms of SANs. So when generating the certs we
now also pass worker address information to the function to be encoded
in the cert.

Finally, there is a change in how the websocket dialing happens, because
http.RoundTripper tries to be too clever for its own good and overwrites
NextProtos on a whim.


internal/session/repository_session.go
```diff
@@ -64,12 +64,14 @@ func (r *Repository) CreateSession(ctx context.Context, sessionWrapper wrapping.
...
 
+	opts := getOpts(opt...)
+
 	id, err := newId()
 	if err != nil {
 		return nil, nil, errors.Wrap(ctx, err, op)
 	}
 
-	privKey, certBytes, err := newCert(ctx, sessionWrapper, newSession.UserId, id, newSession.ExpirationTime.Timestamp.AsTime())
+	privKey, certBytes, err := newCert(ctx, sessionWrapper, newSession.UserId, id, opts.withWorkerAddresses, newSession.ExpirationTime.Timestamp.AsTime())
```
jimlambrt, line 74: Is this really optional?  Feels like a required parameter.

internal/session/session.go
```diff
@@ -351,7 +352,8 @@ func contains(ss []string, t string) bool {
 	return false
 }
 
-func newCert(ctx context.Context, wrapper wrapping.Wrapper, userId, jobId string, exp time.Time) (ed25519.PrivateKey, []byte, error) {
+// newCert creates a new session certificate. If addresses are supplied, they will be parsed and added to IP or DNS SANs as appropriate.
```
jimlambrt, line 355: should we allow addresses to be empty?

internal/session/repository_session.go
```diff
@@ -19,7 +19,7 @@ import (
 // its State of "Pending".  The following fields must be empty when creating a
 // session: ServerId, ServerType, and PublicId.  No options are
 // currently supported.
-func (r *Repository) CreateSession(ctx context.Context, sessionWrapper wrapping.Wrapper, newSession *Session, _ ...Option) (*Session, ed25519.PrivateKey, error) {
+func (r *Repository) CreateSession(ctx context.Context, sessionWrapper wrapping.Wrapper, newSession *Session, opt ...Option) (*Session, ed25519.PrivateKey, error) {
```
jimlambrt, line 22: should addrs be optional?  if so, then mind changing the godoc?

jimlambrt: also, might be nice to have unit tests that require addresses.

internal/servers/worker/handler.go
```diff
@@ -52,11 +53,26 @@ func (w *Worker) handleProxy(listenerCfg *listenerutil.ListenerConfig) (http.Han
 	return func(wr http.ResponseWriter, r *http.Request) {
```
jimlambrt, line 53: I know writing handler tests is a pain, but should we add at least one that verifies the error handling?

internal/servers/worker/worker.go
```diff
@@ -265,11 +266,20 @@ func (w *Worker) getSessionTls(hello *tls.ClientHelloInfo) (*tls.Config, error)
 	ctx := w.baseContext
```
jimlambrt, line 266: might be nice to have a test for the unexported func, given how buried it is during runtime...

jimlambrt (changes requested):
Mostly questions/suggestions about tests.  With that said, I do wonder if addrs should be an option for creating a session?

jimlambrt:
Mind updating go.mod with: go get -u golang.org/x/sys

see: https://stackoverflow.com/questions/71507321/go-1-18-build-error-on-mac-unix-syscall-darwin-1-13-go253-golinkname-mus

internal/session/repository_session.go
jefferai: Yes, it's optional. I just didn't want to introduce an options pattern into `newCert` -- but I can easily do so if you prefer!

internal/session/session.go
jefferai: I don't think it hurts anything. It's not always necessary. In practice we'll probably always supply something except in test code.

If you prefer I could ensure it's not empty and update tests, but I don't have a preference.

jimlambrt (approved):
ty!

PR #1967, opened 2022-04-01
feat(subtypes): Add grpc interceptor for transforming attributes

tmessi:
This interceptor is responsible for maintaining compatability with the JSON
API with the change to the protobuf messages. The messages are changing
from a single attribue field to a oneof to explicitly associate subtype
attribute messages with the resource message. The interceptor transforms
the request prior to the auditReqeust interceptor, and transforms the
response after the auditResponse interceptor. This allows the audit
interceptors to use all the necessary classification information when
creating the audit events.


internal/types/subtypes/interceptor.go
```diff
@@ -0,0 +1,193 @@
...
+
+func transformRequestAttributes(_ context.Context, req proto.Message) proto.Message {
+	domain := messageDomain(req)
+
+	r := req.ProtoReflect()
+	fields := r.Descriptor().Fields()
+
+	itemField := fields.ByName("item")
+	idField := fields.ByName("id")
+	attributesField := fields.ByName("attributes")
```
johanbrandhorst, line 37: Could we add a comment here to explain the assumptions that underlay these? E.g.

```
* "item", if it exists, is assumed to be a field in the top level request containing the primary model of the service, e.g. an Account used in a Create request.
* "id", if it exists, is assumed to be a field in the model containing the models id, from which we can infer the subtype of the request.
* "attributes" is assumed to be a field in the model containing a `google.protobuf.Struct` with fields mapping to one of the specific subtypes, or a generic attribute structure if the type is not specific.
```

internal/types/subtypes/interceptor.go
```diff
@@ -0,0 +1,193 @@
...
+	fieldValue := func(m protoreflect.Message, fd protoreflect.FieldDescriptor) string {
+		if fd == nil {
+			return ""
+		}
+		return m.Get(fd).String()
+	}
+
+	var st Subtype
+	switch {
+	case itemField != nil:
```
johanbrandhorst, line 48: Could we add a comment explaining the scenario that this implies and the next steps? E.g.
```
If "item" is set, it implies that we're sending a resource in the request, e.g. via a Create call.
Use the following priority for deducing the subtype:
* Look it up from the "id" field, if set.
* Look it up from any specified source ID field.
* Look it up by from the "type" field,
```

I think this code is pretty self documenting, but I'm worried it's because we've spent weeks in reflect land already, and whenever reflection is concerned, it's never a bad idea to over-comment, IMO.

internal/types/subtypes/interceptor.go
```diff
@@ -0,0 +1,193 @@
...
+		case sourceIdField != nil && sourceId != "":
+			st = SubtypeFromId(domain, sourceId)
+		case typeField != nil && t != "":
+			st = Subtype(t)
+		default: // need either type or id
+			return req
+		}
+		if err := convertAttributesToSubtype(item, st); err != nil {
+			return req
+		}
```
johanbrandhorst, line 76: Should we surface the error somehow? Goes for all instances of this call.

internal/types/subtypes/interceptor.go
```diff
@@ -0,0 +1,193 @@
...
+	case idField != nil && attributesField != nil:
```
johanbrandhorst, line 77: Same here re: commenting

internal/types/subtypes/interceptor.go
```diff
@@ -0,0 +1,193 @@
...
+	st := Subtype(item.ProtoReflect().Get(typeField).String())
+	return convertAttributesToDefault(item, st)
+}
+
+func transformResponseAttributes(_ context.Context, res proto.Message) proto.Message {
+	r := res.ProtoReflect()
+	fields := r.Descriptor().Fields()
+
+	itemField := fields.ByName("item")
+	itemsField := fields.ByName("items")
```
johanbrandhorst, line 106: Could we add a comment for the scenarios we expect here similarly to the request?

internal/types/subtypes/interceptor.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+		items := r.Get(itemsField).List()
+
+		for i := 0; i < items.Len(); i++ {
+			item := items.Get(i).Message().Interface()
+			if err := transformResponseItemAttributes(item); err != nil {
+				return res
+			}
+		}
+
```
johanbrandhorst, line 129: Nit: remove this newline?

internal/types/subtypes/interceptor.go
```diff
@@ -0,0 +1,193 @@
...
+//     }
+//
+// Will result in a proto request like:
+//
+//     type:"password" attributes:{fields:{key:"login_name" value:{string_value:"tim"}}}
+//
+// This request will be transformed into:
+//
+//     type:"password" password_attributes:{login_name:"tim"}
+func AttributeTransformerInterceptor(_ context.Context) grpc.UnaryServerInterceptor {
```
johanbrandhorst, line 178: Do we need the context parameter?

internal/types/subtypes/interceptor_ext_test.go
```diff
@@ -0,0 +1,291 @@
...
+	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/testing/protocmp"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+func TestAttributeTransformerInterceptor(t *testing.T) {
+	cases := []struct {
+		name             string
+		req              interface{}
+		handlerRes       interface{}
```
johanbrandhorst, line 19: Can we name this `handlerResp` to disambiguate from `handlerResult`? I initially read "result" not "response".

internal/types/subtypes/interceptor_ext_test.go
```diff
@@ -0,0 +1,291 @@
...
+		expectHandlerReq interface{}
+		expectRes        interface{}
```
johanbrandhorst, line 21: Obviously this becomes `expectResp` too.

internal/types/subtypes/source_registry.go
```diff
@@ -0,0 +1,73 @@
...
+	if _, present := s.m[fn]; present {
+		return fmt.Errorf("proto message %s already registered", fn)
+	}
+
+	fields := d.Fields()
+	for i := 0; i < fields.Len(); i++ {
+		f := fields.Get(i)
+
+		opts := f.Options().(*descriptorpb.FieldOptions)
+		isSourceId := proto.GetExtension(opts, protooptions.E_SubtypeSourceId).(bool)
```
johanbrandhorst, line 53: Is this safe to assert? What if this option is not set on the field?

internal/types/subtypes/attribute_transform.go
```diff
@@ -112,3 +112,57 @@ func protoToStruct(p proto.Message) (*structpb.Struct, error) {
...
+
+// Filterable converts a proto.Message so any subtype attributes fields are
+// structed like the API so filter strings will be correctly applied. If the
+// given proto.Message does not have any subtype attributes, the original
+// proto.Message is returned.
+func Filterable(item proto.Message) (proto.Message, error) {
+	clone := proto.Clone(item)
+	r := clone.ProtoReflect()
+
+	attrsField := r.Descriptor().Oneofs().ByName("attrs")
```
johanbrandhorst, line 125: Add comments to underscore all the assumptions about field names and structure, please.

internal/types/subtypes/attribute_transform.go
```diff
@@ -112,3 +112,57 @@ func protoToStruct(p proto.Message) (*structpb.Struct, error) {
...
+			break
+		}
+	}
+
+	// no attrs set, so the original item can just be filtered as is
+	if pbAttrs == nil {
+		return item, nil
+	}
+
+	r.Set(defaultAttrField, protoreflect.ValueOfMessage(pbAttrs.ProtoReflect()))
```
johanbrandhorst, line 159: Worth commenting that this also unset the old field, by virtue of `defaultAttrField` being part of a `oneof`.

internal/types/subtypes/attribute_transform.go
```diff
@@ -118,6 +118,7 @@ func protoToStruct(p proto.Message) (*structpb.Struct, error) {
 // given proto.Message does not have any subtype attributes, the original
 // proto.Message is returned.
 func Filterable(item proto.Message) (proto.Message, error) {
+	fmt.Printf("POOP: %#v\n", item)
```
johanbrandhorst, line 121: 😬

internal/types/subtypes/attribute_transform.go
tmessi: That is documents on [Set](https://github.com/protocolbuffers/protobuf-go/blob/32051b4f86e54c2142c7c05362c6e96ae3454a1c/reflect/protoreflect/value.go#L95-L96) do you think it is needed here as well? We call set on a one of in a couple spots in this file. Would you want a comment in both places?

internal/types/subtypes/interceptor.go
tmessi: The other [interceptors](https://github.com/hashicorp/boundary/blob/tmessi-subtype-grpc-interceptor/internal/servers/controller/gateway.go) all take a context. I think it is likely that this context will be needed or used at some point, since we need a context to make error events. So I think it should be included even if it is not used yet.

internal/types/subtypes/interceptor.go
```diff
@@ -0,0 +1,193 @@
...
+	if fd == nil {
+		return ""
+	}
+
+	opts, ok := fd.Options().(*descriptorpb.FileOptions)
+	if !ok {
+		return ""
+	}
+
+	domain := proto.GetExtension(opts, protooptions.E_Domain).(string)
```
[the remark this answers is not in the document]
tmessi: This does seem to be safe, and will result in an empty string being returned if domain is not set on the file. I don't think an error is necessary, since the only thing I can see to do if a domain is set, is potentially return from `transformRequestAttributes` if `domain == ""`

internal/types/subtypes/attribute_transform.go
```diff
@@ -0,0 +1,114 @@
...
+	// TODO: replicate this logic but with a proto extension set on the Field
+	// descriptor There are some attributes where we want to discard unknown
+	// fields, while others that should error if there are unknown fields
+	//
+	// opts := GetOpts(opt...)
+	// if opts.withDiscardUnknownFields {
+	// 	err = (protojson.UnmarshalOptions{DiscardUnknown: true}.Unmarshal(js, p))
+	// } else {
+	// 	err = protojson.Unmarshal(js, p)
+	// }
```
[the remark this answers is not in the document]
tmessi: Ah, yep. I forgot I left this. I should address this in this PR.

internal/types/subtypes/attribute_transform.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		// TODO: error event?
+		return err
+	}
+	r.Set(defaultAttrField, protoreflect.ValueOfMessage(defaultAttrs.ProtoReflect()))
+	return nil
+}
+
+func structToProto(fields *structpb.Struct, p proto.Message) error {
+	if fields == nil {
```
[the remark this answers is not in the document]
tmessi: I am not entirely sure, these `structToProto` and `protoToStruct` functions are lifted from the [handlers pkg](https://github.com/hashicorp/boundary/blob/tmessi-subtype-grpc-interceptor/internal/servers/controller/handlers/subtypes.go). I think we should be able to remove the functions from `handlers` once we have updated all of the packages. Then this functionality should remain internal to the `subtypes` package.

internal/types/subtypes/attributes.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	return tt, nil
+}
+
+// TODO: fix doc
+// protoAttributeField is used by the attrMarshaler to translate between JSON
+// formats for the API and for the protobuf messages. It expects a
+// proto.Message with a OneOf field for the subtype attributes and the subtype
+// string. It returns the string for the JSON key that that should be used for
+// the subtype's attributes fields.
```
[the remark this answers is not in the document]
tmessi: Yea this gets fixed in 026f8b730, I think this will get squashed together before merging, since this is tightly coupled with the grpc interceptor. I think I just missed adding the correct doc in this earlier commit.

internal/types/subtypes/registry_global.go
```diff
@@ -0,0 +1,121 @@
...
+
+	knownSubtypes, present := r.knownSubtypes[domain]
+	if !present {
+		knownSubtypes = make(map[Subtype]interface{})
+		r.knownSubtypes[domain] = knownSubtypes
+		r.subtypesPrefixes[domain] = make(map[string]Subtype)
+	}
+
+	if _, present := knownSubtypes[subtype]; present {
+		return errors.NewDeprecated(errors.SubtypeAlreadyRegistered, op, fmt.Sprintf("subtype %q already registered in domain %s", subtype, domain))
```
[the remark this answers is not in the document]
tmessi: Good catch, I just need to thread a context.Context into this to avoid using `errors.NewDeprecated`.

internal/types/subtypes/attribute_transform.go
```diff
@@ -0,0 +1,114 @@
...
+	// 	err = (protojson.UnmarshalOptions{DiscardUnknown: true}.Unmarshal(js, p))
+	// } else {
+	// 	err = protojson.Unmarshal(js, p)
+	// }
+
+	err = protojson.Unmarshal(js, p)
+	if err != nil {
+		return err
+	}
+	return nil
```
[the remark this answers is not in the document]
tmessi: this will look different once I address the TODO mentioned [here](https://github.com/hashicorp/boundary/pull/1967#discussion_r842048065)

internal/types/subtypes/attribute_transform.go
johanbrandhorst: I think it would be nice, yeah.

internal/types/subtypes/interceptor.go
tmessi: They are not scoped to the request context. If you look at the other interceptors, most take a context but do not use it, the ones that do use it are performing [initialization checks](https://github.com/hashicorp/boundary/blob/dd988c1d867c813759a5aa0f2fb1151c69791cf6/internal/servers/controller/interceptor.go#L55-L73). I don't know what initialization checks could be added to this interceptor, but I think consistency with the other interceptors is a good idea here.

johanbrandhorst: That sounds reasonable 👍🏻, thanks for explaining.

internal/types/subtypes/interceptor.go
tmessi: I can look to add some additional comments in this file. I went back and forth on this a bit. I agree that it is a bit "self documenting", and I think the corresponding tests also provide some good context. The only down-side I have see to over-commenting is when the code gets updated and the comments dont.

johanbrandhorst: Yeah, agree, hopefully this code won't change too much

internal/types/subtypes/interceptor.go
tmessi: I added docs in [7658e9e](https://github.com/hashicorp/boundary/pull/1967/commits/7658e9e489454006241899d5864127cb99e47422) that should hopefully explain this now

internal/types/subtypes/interceptor.go
tmessi: See [7658e9e](https://github.com/hashicorp/boundary/pull/1967/commits/7658e9e489454006241899d5864127cb99e47422)

internal/types/subtypes/interceptor.go
tmessi: See [7658e9e](https://github.com/hashicorp/boundary/pull/1967/commits/7658e9e489454006241899d5864127cb99e47422)

internal/types/subtypes/attribute_transform.go
tmessi: See [24fe7b8](https://github.com/hashicorp/boundary/pull/1967/commits/24fe7b8a7acdd5176c291e72452144c3005a260e)

internal/types/subtypes/attribute_transform.go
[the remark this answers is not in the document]
tmessi: Actually. I think we can revisit this when we get to `authmethods` that is the only package (currently) that requires this functionality. But the general approach will be to create another proto custom option that can be set on the message that will indicate if `DiscardUnknown` should be set to true.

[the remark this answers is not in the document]
johanbrandhorst: OK

internal/types/subtypes/attribute_transform.go
```diff
@@ -0,0 +1,186 @@
...
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+func convertAttributesToSubtype(msg proto.Message, st Subtype) error {
+	r := msg.ProtoReflect()
+	d := r.Descriptor()
+
+	defaultAttrField, err := attributeField(d, defaultSubtype)
+	if err != nil {
+		// no need to convert
```
johanbrandhorst, line 18: This comment is a little confusing, why, when there is an error, is there no need to convert?

internal/types/subtypes/attribute_transform.go
```diff
@@ -0,0 +1,186 @@
...
+		return item, nil
+	}
+
+	var attr proto.Message
+	var pbAttrs proto.Message
+	var err error
+
+	oneOfFields := attrsField.Fields()
+
+	for i := 0; i < oneOfFields.Len(); i++ {
```
johanbrandhorst, line 162: Would it be worth adding a comment here about what we're doing? The only clue is the `if` statement, so maybe worth spelling it out.
```suggestion
	// Find the populated oneof field and turn it into a `*structpb.Struct`.
	for i := 0; i < oneOfFields.Len(); i++ {
```

internal/types/subtypes/attributes.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+var globalAttributeRegistry attributeRegistry
+
+// register examines the given protobuf MessageDescriptor for fields that have
+// the Subtype protobuf extension. It uses these to build a fieldMap of subtype
+// strings to the protobuf FieldDescriptors. If the message does not have any
+// subtypes it will not be registered, but no error is returned, allowing this
+// to be called on any protobuf message. However, if a message has subtypes but
+// does not provide a field with a subtype of "default" an error is returned.
```
johanbrandhorst, line 47:
```suggestion
// does not provide a field with a subtype of "default", or if the default
// field is not a *structpb.Struct type, an error is returned.
```

internal/types/subtypes/interceptor_ext_test.go
```diff
@@ -0,0 +1,333 @@
...
+			}
+
+			got, err := subtypes.AttributeTransformerInterceptor(ctx)(ctx, tc.req, nil, handler)
+			require.NoError(t, err)
+			require.Empty(t, cmp.Diff(got, tc.excpetResp, protocmp.Transform()))
+		})
+	}
+}
+
+func TestAttributeTransformerInterceptorRequestErrors(t *testing.T) {
```
johanbrandhorst, line 294: Could we add a test here as well to see what happens when the response contains an unknown `type`?

internal/types/subtypes/interceptor_test.go
```diff
@@ -0,0 +1,575 @@
...
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			got, err := transformResponseAttributes(tc.resp)
+			require.NoError(t, err)
+			assert.Empty(t, cmp.Diff(got, tc.expected, protocmp.Transform()))
+		})
+	}
+}
```
johanbrandhorst, line 575: Should we have an error test as well for these functions? Personally I'd probably argue that we don't necessarily need both the internal and external tests so I'll leave it up to you.

internal/types/subtypes/registry_global.go
```diff
@@ -0,0 +1,132 @@
...
+		r.subtypesPrefixes[domain] = make(map[string]Subtype)
+	}
+
+	if _, present := knownSubtypes[subtype]; present {
+		return errors.New(
+			ctx,
+			errors.SubtypeAlreadyRegistered,
+			op,
+			fmt.Sprintf("subtype %q already registered in domain %s", subtype, domain),
+		)
```
johanbrandhorst, line 95: Nice 👌🏻

internal/types/subtypes/interceptor.go
```diff
@@ -0,0 +1,297 @@
...
+//    message FooActionRequest {
+//      string id = 1;
+//      oneof attrs {
+//        google.protobuf.Struct attributes = 10 [(custom_options.v1.subtype) = "default"];
+//        // other subtype attributes types
+//      }
+//    }
+//
+// Also note that for any of the id based lookups to function, the file that contains
+// the proto.Message definition must set the "domain" custom option.
```
johanbrandhorst, line 95: This comment is ⭐

johanbrandhorst (commented):
Minor doc and test nits remaining

internal/types/subtypes/attribute_transform.go
tmessi: I can expand the comment to clarify. But `attribueField` will return an error if the given message is not registered, which will be the case for a message like [this test case](https://github.com/hashicorp/boundary/blob/fc2447052a6ddc85da8be422437f13675f5878aa/internal/types/subtypes/interceptor_test.go#L247-L271). So if we are not able to get a `defaultAttrField` there should be no need to convert.

johanbrandhorst: Yeah I think it would be valuable to expand the comment 👍🏻.

internal/types/subtypes/interceptor_test.go
tmessi: I think in this case the external test is more important, since the particular error message and type is important as it will be returned. And that error message is determined in the exported interceptor function. The particular errors from the un-exported functions imo should be free to change.

johanbrandhorst: Agreed the external tests are more important in general.

internal/types/subtypes/interceptor_ext_test.go
tmessi: See d6d2e1f38

internal/types/subtypes/attribute_transform.go
tmessi: See 52cf23ce7

johanbrandhorst (approved):

PR #1969, opened 2022-04-04
Fix missing context

thorfour:
I ran into two errors when trying to use the commit (`58a448fc6`) of main.

image
image

So I changed to use the passed in context instead of the command context. I'm unfamiliar of the difference but this got everything working.


tmessi:
I think @jefferai has a different fix for this in #1959

thorfour:
> I think @jefferai has a different fix for this in #1959

Even better!

PR #1970, opened 2022-04-05
feat(controller): Reorder interceptor stack

tmessi:
The new `subtype.AttributeTransformerInterceptor` can return an error if
it fails to transform the attributes. This is mainly if invalid
attribute fields were included for a particular subtype. This error case
used to be checked in the grpc service handler functions. However, now
that it is happening in the interceptor, the `errorInterceptor` needs to
be moved up in the stack so that it will process the response of the new
transformation interceptor.

The `errorInterceptor` used to be after the `auditRequestInterceptor`
and prior to the `auditResponseInterceptor`, meaning that the
`auditResponseInterceptor` would have access to the result of the
`errorInterceptor`. But it seems the `auditRequestInterceptor` does not
use or need any of the headers set by the `errorInterceptor` so it
should be safe to move the `errorInterceptor`. This is important since
the `subtype.AttributeTransformerInterceptor` needs to be before the
audit interceptors in the stack to properly transform requests/responses
so they can be audited with the complete classification information.

### DIFF
--- a/internal/servers/controller/gateway.go
+++ b/internal/servers/controller/gateway.go
@@ -75,9 +75,9 @@ func newGrpcServer(
 		grpc.UnaryInterceptor(
 			grpc_middleware.ChainUnaryServer(
 				requestCtxInterceptor,                         // populated requestInfo from headers into the request ctx
+				errorInterceptor(ctx),                         // convert domain and api errors into headers for the http proxy
 				subtypes.AttributeTransformerInterceptor(ctx), // convert to/from generic attributes from/to subtype specific attributes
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):

tmessi:
@jimlambrt if you can take a quick look, and make sure my understanding of the errorInterceptor is correct.

jefferai:
Linking to https://github.com/hashicorp/boundary/pull/1955 as it's semi-related. @jimlambrt was going to look into that because I get some failures with that change and it's related to the event decoration happening in a few ways IIRC.

jimlambrt (approved):
ty!

PR #1972, opened 2022-04-06
feat(authmethods): add well typed attribute options

johanbrandhorst:
Add well typed attribute options to auth_method proto
definitions, while hiding them from being rendered in
the openapiv2 spec.

For https://hashicorp.atlassian.net/browse/ICU-3810 and https://hashicorp.atlassian.net/browse/ICU-3812


sdk/pbs/controller/api/resources/authmethods/auth_method_taggable_test.go
```diff
@@ -1,182 +0,0 @@
...
-	"github.com/hashicorp/boundary/sdk/wrapper"
-	"github.com/hashicorp/eventlogger"
-	"github.com/hashicorp/eventlogger/filters/encrypt"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-	"google.golang.org/protobuf/types/known/structpb"
-	"google.golang.org/protobuf/types/known/wrapperspb"
-)
-
-func TestAuthMethod_Tags(t *testing.T) {
```
tmessi, line 19: I think we should be able to keep this test in place. I would potentially rename the file, but these tests should be asserting that the classification is correct, we should probably add similar tests for the other packages.

internal/servers/controller/handlers/authmethods/authmethod_service.go
```diff
@@ -839,13 +831,11 @@ func validateCreateRequest(ctx context.Context, req *pbs.CreateAuthMethodRequest
...
-			if err := handlers.StructToProto(req.GetItem().GetAttributes(), attrs); err != nil {
-				badFields[attributesField] = "Attribute fields do not match the expected format."
-			}
+			// Password attributes are not required when creating a password auth method.
 		case oidc.Subtype:
-			attrs := &pb.OidcAuthMethodAttributes{}
-			if err := handlers.StructToProto(req.GetItem().GetAttributes(), attrs); err != nil {
+			attrs := req.GetItem().GetOidcAuthMethodsAttributes()
+			if attrs == nil {
+				// OIDC attributes are required when creating an OIDC auth method.
```
tmessi, line 838: I am wondering if the error message returned here should be updated. The specific error that was being checked here will now happen in the interceptor, but it is possible that no attributes are provided at all, which would reach this code. I suppose it is technically correct to say that `nil` is that the expected format, but it seems a little confusing to me.

internal/gen/controller/api/services/auth_method_service_taggable_test.go
```diff
@@ -1,148 +0,0 @@
...
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-	structpb "google.golang.org/protobuf/types/known/structpb"
-)
-
-// TestAuthenticate_Tags will test that the response filtering aligns with the
-// AuthenticateResponse and AuthenticateResponse tags.  See:
-// internal/tests/api/authmethods/authenticate_test.go TestAuthenticate where
-// the audit events produced using these tags is unit tested.
-func TestAuthenticate_Tags(t *testing.T) {
```
tmessi, line 23: I think we should be able to keep this test in place. I would potentially rename the file, but these tests should be asserting that the classification is correct, we should probably add similar tests for the other packages.

tmessi (changes requested):

sdk/pbs/controller/api/resources/authmethods/auth_method_taggable_test.go
johanbrandhorst: I'm not sure what value these tests add to be honest. Classification is defined in the Protobuf files, anything downstream of that is automatically generated. These tests would catch us forgetting to run `gotags` on a file, or maybe adding a tag with a typo, but is that worth writing tests for? It feels like it would mostly test that `gotags` does the right thing.

johanbrandhorst: Discussed on zoom, we will keep these tests, but now that I'm playing around with them, I realize that they're going to be wrong until we add the classification anyway, so I'm going to punt on this until the next PR which adds the classification to the fields.

internal/servers/controller/handlers/authmethods/authmethod_service.go
johanbrandhorst: Done in https://github.com/hashicorp/boundary/pull/1972/commits/7899545d25d153bef7a0a7e80e3c03d54a45a815

internal/servers/controller/handlers/authmethods/authmethod_service.go
```diff
@@ -177,7 +177,13 @@ func (s Service) ListAuthMethods(ctx context.Context, req *pbs.ListAuthMethodsRe
...
 		}
 
-		if filter.Match(item) {
+		// This comes last so that we can use item fields in the filter after
+		// the allowed fields are populated above
+		filterable, err := subtypes.Filterable(item)
+		if err != nil {
+			return nil, err
+		}
+		if filter.Match(filterable) {
```
johanbrandhorst, line 186: Forgot to add the filterable conversion here, and it turns out we didn't have any tests for this.

tmessi: hmm yea, good catch, yea some tests for this sounds like a good idea

johanbrandhorst: Addressed in fce9968e68ae923c71f0a51500590f732e52ef5e

johanbrandhorst:
Note that this PR now contains https://github.com/hashicorp/boundary/pull/1982. We can drop it again when we rebase after that is merged.

tmessi (approved):

Issue #1856, opened 2022-02-15
missing root_key_id linked to global scope after a few upgrades


dnini96:
Hello guys

I installed boundary 0.1.2 when it came out and have since upgraded the existing install. 0.1.2 -> 0.1.3 -> 0.1.5 -> 0.2.0 -> 0.3.0 and then 0.4.0. I recently wanted to add an OIDC authentication method and stumbled upon this issue/bug: 
```
Feb 14 11:10:37 DWSE-nonprd-boundary-ansible boundary[21515]: 2022-02-14T11:10:37.821+0100 [ERROR] controller.authmethod_service.oidc: error starting the oidc authentication flow: op=authmethod_service.(Service).authenticateOidcStart error="oidc.StartAuth: oidc.(Repository).oidcWrapper: unable to get oidc wrapper: kms.GetWrapper: error loading oidc for scope global: kms.loadDek: error finding oidc key for scope global: integrity violation: error #105"
Feb 14 11:10:37 DWSE-nonprd-boundary-ansible boundary[21515]: 2022-02-14T11:10:37.821+0100 [ERROR] controller: internal error returned: error="authmethod_service.(Service).authenticateOidcStart: Error generating parameters for starting the OIDC flow. See the controller's log for more information.: unknown: error #500"
```
I looked in the database and in the kms_oidc_key table there were only root keys used which were not linked to a root key with the global scope_id. In a fresh install there is one which did. 

current: 
```
boundary=> select * from kms_root_key;
   private_id   |   scope_id   |          create_time          
----------------+--------------+-------------------------------
 krk_epNm3ryB2E | global       | 2021-01-08 12:08:53.312033+00
 krk_12rAHjQFmm | o_c4HqJ42HzW | 2021-01-18 13:15:51.800903+00
 krk_IuIlUSGosT | p_j3z93Fl705 | 2021-01-18 14:26:40.586433+00
 krk_YeJ4ZglLaF | p_5IASnGvwAY | 2021-01-18 14:26:40.650917+00
 krk_k89dEhTKMH | p_9zUSmEcuAv | 2021-01-22 11:58:09.509812+00
 krk_RekrmdxVAi | p_m1sqqo8lYf | 2021-01-22 11:58:09.739897+00
 krk_ovuhOif3hJ | p_N0uDAzdTdZ | 2021-01-22 11:58:09.907963+00
 krk_CBXkqehTHj | p_fYWSd8RNnu | 2021-01-22 12:03:25.328386+00
[...]

PR #1976, opened 2022-04-06
fix (kms): properly reconcile DEKs for existing scopes

jimlambrt:
When new DEKs are added to boundary we need to ensure that the keys
for existing scopes are reconciled.

This fix should address:
https://github.com/hashicorp/boundary/issues/1856

### DIFF
--- a/internal/kms/kms.go
+++ b/internal/kms/kms.go
@@ -287,8 +287,10 @@ func (k *Kms) loadRoot(ctx context.Context, scopeId string, opt ...Option) (*mul
 	return pooled, rootKeyId, nil
 }
 
-// ReconcileKeys will reconcile the keys in the kms against known possible issues.
-func (k *Kms) ReconcileKeys(ctx context.Context, randomReader io.Reader) error {
+// ReconcileKeys will reconcile the keys in the kms against known possible
+// issues.  This function reconciles the global scope unless the
+// WithScopeIds(...) option is provided
+func (k *Kms) ReconcileKeys(ctx context.Context, randomReader io.Reader, opt ...Option) error {
 	const op = "kms.ReconcileKeys"
 	if isNil(randomReader) {
 		return errors.New(ctx, errors.InvalidParameter, op, "missing rand reader")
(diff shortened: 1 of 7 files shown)


internal/kms/kms_ext_test.go
```diff
@@ -138,16 +138,21 @@ func TestKms(t *testing.T) {
 }
 
 func TestKms_ReconcileKeys(t *testing.T) {
-	t.Parallel()
+	// t.Parallel()
```
johanbrandhorst, line 141: Do we want to just remove this?

jimlambrt: Umm... yes?  ty!

johanbrandhorst (dismissed):
I reserve business logic judgment as I don't feel I have enough context.

johanbrandhorst (approved):

tmessi (approved):

internal/kms/kms.go
```diff
@@ -315,6 +317,46 @@ func (k *Kms) ReconcileKeys(ctx context.Context, randomReader io.Reader) error {
...
+	for _, id := range opts.withScopeIds {
+		var scopeRootWrapper *multi.PooledWrapper
+
+		for _, purpose := range []KeyPurpose{
+			// just add additional purposes as needed going forward to reconcile
+			// new keys as they are added.
+			//
+			// NOTE: don't add an audit key here since it can only be created in
+			// the global scope.
+			KeyPurposeOidc,
```
jefferai, line 331: Are we already creating OIDC DEKs for non-global/org scopes? If not we may want a check here since this will be called with _all_ scopes from the controller New function.

internal/servers/controller/controller.go
```diff
@@ -220,7 +220,21 @@ func New(ctx context.Context, conf *Config) (*Controller, error) {
...
+	}
+	allScopes, err := iamRepo.ListScopesRecursively(ctx, scope.Global.String())
+	if err != nil {
+		return nil, fmt.Errorf("error listing all scopes for reconciling keys: %w", err)
+	}
+	reconcileScopeIds := make([]string, 0, len(allScopes))
+	for _, s := range allScopes {
+		reconcileScopeIds = append(reconcileScopeIds, s.PublicId)
+	}
+	if err := c.kms.ReconcileKeys(ctx, c.conf.SecureRandomReader, kms.WithScopeIds(reconcileScopeIds...)); err != nil {
```
jefferai, line 237: This will turn into potentially a lot of DB calls (depending on the number of scopes) because we'll not have a warmed KMS cache at this point and it's going to fetch roots for all scopes. Doesn't need to be addressed now but should be in a follow up.

jefferai (approved):
Approved pending a check on if oidc keys are supposed to be getting created in project scopes.

internal/kms/kms.go
jimlambrt: We are calling `kms.CreateKeysTx(...)` every time we create a new scope.

internal/servers/controller/controller.go
jimlambrt: This will be addressed in a soonish PR that ports boundary over to use the new go-kms-wrapping/extras/kms package.

PR #1979, opened 2022-04-06
feat(db): Suppress not found events during db lookups

### LINKED ISSUE #1925: Not found errors on startup
On startup we currently get some events:

```
{
  "id": "Q9ZN6uQZoX",
  "source": "https://hashicorp.com/boundary/dev-controller/boundary-dev",
  "specversion": "1.0",
  "type": "error",
  "data": {
    "error": "db.LookupWhere: record not found, search issue: error #1100: dbw.LookupWhere: record not found",
    "error_fields": {
      "Code": 1100,
      "Msg": "",
      "Op": "db.LookupWhere",
      "Wrapped": {}
    },
    "id": "e_c6EmYVKTHg",
    "version": "v0.1",
    "op": "db.LookupWhere"
  },
  "time": "2022-03-15T14:02:21.607812-07:00"
}
{
  "id": "3fS3pYOdAz",
  "source": "https://hashicorp.com/boundary/dev-controller/boundary-dev",
  "specversion": "1.0",
  "type": "error",
  "data": {
    "error": "db.LookupWhere: record not found, search issue: error #1100: dbw.LookupWhere: record not found",
    "error_fields": {
      "Code": 1100,
      "Msg": "",
      "Op": "db.LookupWhere",
      "Wrapped": {}
    },
    "id": "e_T9Pmiq92Ze",
    "version": "v0.1",
    "op": "db.LookupWhere"
  },
  "time": "2022-03-15T14:02:22.472712-07:00"
} 
```

Putting a conversation here with light editing:

Irena:
It's happening when the Controller registers the AWS and Azure plugins. They aren't found, and end up being created (which is the expected behaviour). That code path is here:
https://github.com/hashicorp/boundary/blob/main/internal/cmd/base/initial_resources.go#L633
->
 https://github.com/hashicorp/boundary/blob/main/internal/plugin/host/repository_plugin.go#L115-L117
->
https://github.com/hashicorp/boundary/blob/main/internal/cmd/base/initial_resources.go#L638-L643

I looked into why we're emitting the LookupWhere error. When the lookup fails, we wrap the error, which ends up emitting the error:
https://github.com/hashicorp/boundary/blob/main/internal/db/read_writer.go#L468
->
https://github.com/hashicorp/boundary/blob/main/internal/db/read_writer.go#L516-L517
->
https://github.com/hashicorp/boundary/blob/main/internal/errors/error.go#L140-L155
->
https://github.com/hashicorp/boundary/blob/main/internal/errors/error.go#L111 (edited) 

The read_writer wrapError function ignores any incoming opts; had we passed the opts.withoutEvent opt this would not have been emitted

Jeff:
Ah so it's emitted even before we hit the error case in the calling function. Want to pull @jimlambrt into this. We could pass withoutEvent but I wonder if that's where we should be emitting it in the first place. Like, is it normal to want to do a lookup and have no records found be a perfectly acceptable success criteria

louisruch:
Fixes #1925 

@jimlambrt I have been fixing a couple of related scheduler error spam and also thought that perhaps NotFound should be suppressed as well. This is going off your comment left in #1925 happy to discuss if you like

### DIFF
--- a/internal/db/read_writer.go
+++ b/internal/db/read_writer.go
@@ -3,7 +3,6 @@ package db
 import (
 	"context"
 	"database/sql"
-	stderrors "errors"
 	"fmt"
 	"reflect"
 	"time"
@@ -437,7 +436,7 @@ func (w *Db) DoTx(ctx context.Context, retries uint, backOff Backoff, handler Tx
 }
 
 // LookupByPublicId will lookup resource by its public_id or private_id, which
-// must be unique. Options are ignored.
+// must be unique. WithDebug is the only valid option, all other options are ignored.
 func (rw *Db) LookupById(ctx context.Context, resourceWithIder interface{}, opt ...Option) error {
 	const op = "db.LookupById"
 	if rw.underlying == nil {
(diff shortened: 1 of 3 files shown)


jimlambrt (dismissed):
Looks great!  Would be better with a test or two that checks to see if the event is suppressed but certainly not a blocker.  ty!

jimlambrt (dismissed):
Ty!

PR #1984, opened 2022-04-08
feat(metrics): add build info metric on server start

A-440Hz:


internal/cmd/base/internal/metric/build_info.go
```diff
@@ -0,0 +1,50 @@
...
+	return map[string]string{
+		labelGoVersion:       runtime.Version(),
+		labelGitRevision:     verInfo.Revision,
+		labelBoundaryVersion: verInfo.Version,
+	}
+}
+
+// InitializeBuildInfoVec registers the boundary_build_info metric with its
+// correct labels and sets its value to 1.
+func InitializeBuildInfoVec(r prometheus.Registerer) {
```
talanknight, line 42: nit: Having `Vec` in the name feels a little like an implementation detail.
```suggestion
func InitializeBuildInfo(r prometheus.Registerer) {
```

talanknight (approved):
Looks good!

talanknight (changes requested):
Did you notice that CI is failing?

internal/cmd/base/servers.go
```diff
@@ -134,6 +135,8 @@ type Server struct {
 }
 
 func NewServer(cmd *Command) *Server {
+	reg := prometheus.NewRegistry()
```
talanknight, line 138: Have you tested this solution, using a NewRegistery instead of the DefaultRegisterer?  I think the prometheus http handler relies on the default registerer being used.

A-440Hz: Ah you're right none of the http metrics appear

internal/cmd/base/internal/metric/build_info.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"runtime"
+
+	"github.com/hashicorp/boundary/globals"
+	"github.com/hashicorp/boundary/version"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	labelGoVersion       = "goversion"
```
jefferai, line 14: Is this the standard way to create label names or should it be something like `go_version`? I would have thought separate words would have a separator.

A-440Hz: That's what I thought initially, but the [prometheus collector](https://github.com/prometheus/common/blob/8558a5b7db3c84fa38b4766966059a7bd5bfa2ee/version/info.go#L38) (which does the same thing but doesn't work here) labeled theirs `goversion`, so that's what I went with here

jefferai: Huh, casing is all over the place in that file.

Sounds good, let's use their const.

internal/cmd/base/option.go
```diff
@@ -188,3 +190,10 @@ func WithEventGating(with bool) Option {
 		o.withEventGating = with
 	}
 }
+
+// WithPrometheusRegisterer uses the provided prometheus registerer
+func WithPrometheusRegisterer(with prometheus.Registerer) Option {
```
talanknight, line 195: I like this approach.  Can you also please add a test to /internal/cmd/base/option_test.go ?

talanknight (commented):
Looks good!

talanknight (approved):

PR #1987, opened 2022-04-08
feat(groups): Classify resources for audit events

tmessi:


johanbrandhorst (approved):

internal/proto/controller/api/services/v1/group_service.proto
```diff
@@ -131,80 +131,80 @@ service GroupService {
 }
 
 message GetGroupRequest {
-  string id = 1;
+  string id = 1; // @gotags: `class:"public"`
```
jefferai, line 134: Same comments in here about tags on messages

jefferai (changes requested):
Same comment here about tagging the messages

jimlambrt (approved):

jefferai (approved):

PR #1988, opened 2022-04-08
Add subtype transformation function registration

johanbrandhorst:
## refact(subtypes): Remove return from transformations

This return value is just a reference to the input, which
is being mutated. We may as well make that clear via
the signature, which will also simplify the new custom
transformation function registration logic.

## feat(subtypes): Add new transformation registration option

This allows the user to register a custom transformation callback
for a specific protobuf message, in cases where the transformation
is sufficiently complex for it to be cumbersome to formalise
the logic through protobuf options.


internal/types/subtypes/registry_global.go
```diff
@@ -111,6 +117,62 @@ func (r *registry) register(ctx context.Context, domain string, subtype Subtype,
...
+	return transformRequestAttributes(msg)
+}
+
+func transformResponse(msg proto.Message) error {
+	fqn := msg.ProtoReflect().Descriptor().FullName()
+	if fn, ok := globalRegistry.responseTransformationFuncs[fqn]; ok {
+		return fn(msg)
+	}
+	return transformResponseAttributes(msg)
+}
```
tmessi, line 170: i think these functions should be in `interceptor.go`, as it is now, trying to trace the logic from the interceptor brings you to this file, which then goes back to `interceptory.go` for `transformResponseAttributes`

internal/types/subtypes/registry_global.go
```diff
@@ -7,23 +7,29 @@ import (
...
 
 // registry stores a collection of boundary resource subtypes along with their
 // prefixes and allows for translating prefixes back to registered subtypes.
 type registry struct {
-	subtypesPrefixes map[string]map[string]Subtype
-	knownSubtypes    map[string]map[Subtype]interface{}
+	subtypesPrefixes            map[string]map[string]Subtype
+	knownSubtypes               map[string]map[Subtype]interface{}
+	requestTransformationFuncs  map[protoreflect.FullName]TransformationFunc
+	responseTransformationFuncs map[protoreflect.FullName]TransformationFunc
```
tmessi, line 32: I am not sure this belongs in this registry, given that the keys, and the input for the corresponding register function is a protomessage, while this registry is currently responsible for mapping Subtypes to ids.

I think it should either be a separate registry, or potentially added into the [attributeRegistery](https://github.com/hashicorp/boundary/blob/a6c026fb769004f1b9beca371ff46d9ed2d53cc2/internal/types/subtypes/attributes.go#L35-L40) or [sourceRegistry](https://github.com/hashicorp/boundary/blob/a6c026fb769004f1b9beca371ff46d9ed2d53cc2/internal/types/subtypes/source_registry.go#L28-L36), but they would probably need to be renamed to be more clear what they are.

tmessi (commented):
d6716cc09ff3dce50149a2b0185868fc5d36c955 looks good, and a6c026fb769004f1b9beca371ff46d9ed2d53cc2 looks good except I have some questions/suggestings about organization of some things.

internal/types/subtypes/registry_global.go
johanbrandhorst: Good points, let me think a bit about it and update it.

internal/types/subtypes/registry_global.go
johanbrandhorst: Makes sense 👍🏻

johanbrandhorst: Fixed

internal/types/subtypes/registry_global.go
johanbrandhorst: Created a new `transformationRegistry` type, let me know if you have another idea.

internal/types/subtypes/source_registry.go
```diff
@@ -35,7 +35,7 @@ type sourceRegistry struct {
 	m map[protoreflect.FullName]protoreflect.FieldDescriptor
 }
 
-func (s sourceRegistry) register(d protoreflect.MessageDescriptor) error {
+func (s *sourceRegistry) register(d protoreflect.MessageDescriptor) error {
```
tmessi, line 38: :thinking: is this needed?

tmessi (approved):

internal/types/subtypes/source_registry.go
johanbrandhorst: Yep, the mutex is otherwise copied and provides no mutual exclusion :P

PR #1998, opened 2022-04-12
feat(worker): Instrumentation for Controller Cluster Communication

hugoamvieira:
Adds request latency observations to the worker side (client) of
worker <-> controller cluster communications by integrating a new
grpc client interceptor.

### DIFF
--- a/internal/servers/controller/internal/metric/cluster.go
+++ b/internal/servers/controller/internal/metric/cluster.go
@@ -20,20 +20,18 @@ const (
 	clusterSubSystem = "controller_cluster"
 )
 
-var (
-	// gRpcRequestLatency collects measurements of how long it takes
-	// the boundary system to reply to a request to the controller cluster
-	// from the time that boundary received the request.
-	gRpcRequestLatency prometheus.ObserverVec = prometheus.NewHistogramVec(
-		prometheus.HistogramOpts{
-			Namespace: globals.MetricNamespace,
-			Subsystem: clusterSubSystem,
-			Name:      "grpc_request_duration_seconds",
-			Help:      "Histogram of latencies for gRPC requests.",
-			Buckets:   prometheus.DefBuckets,
-		},
-		[]string{labelGRpcCode, labelGRpcService, labelGRpcMethod},
-	)
+// gRpcRequestLatency collects measurements of how long it takes
+// the boundary system to reply to a request to the controller cluster
+// from the time that boundary received the request.
+var gRpcRequestLatency prometheus.ObserverVec = prometheus.NewHistogramVec(
+	prometheus.HistogramOpts{
+		Namespace: globals.MetricNamespace,
+		Subsystem: clusterSubSystem,
+		Name:      "grpc_request_duration_seconds",
+		Help:      "Histogram of latencies for gRPC requests.",
+		Buckets:   prometheus.DefBuckets,
+	},
+	[]string{labelGRpcCode, labelGRpcService, labelGRpcMethod},
 )
 
 // statusFromError retrieves the *status.Status from the provided error.  It'll
@@ -70,9 +68,11 @@ type statsHandler struct{}
 func (sh statsHandler) TagRPC(ctx context.Context, i *stats.RPCTagInfo) context.Context {
 	return context.WithValue(ctx, metricMethodNameContextKey{}, i.FullMethodName)
 }
+
 func (sh statsHandler) TagConn(ctx context.Context, _ *stats.ConnTagInfo) context.Context {
 	return ctx
 }
+
 func (sh statsHandler) HandleConn(context.Context, stats.ConnStats) {
 }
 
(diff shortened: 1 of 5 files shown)


internal/servers/worker/internal/metric/controller_dialer.go
```diff
@@ -0,0 +1,156 @@
...
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/reflect/protoregistry"
+)
+
+const (
+	labelGrpcCode               = "grpc_code"
+	labelGrpcService            = "grpc_service"
+	labelGrpcMethod             = "grpc_method"
+	workerToControllerSubsystem = "worker_to_controller_cluster"
```
talanknight, line 22: IMO the subsystem metric being used is the cluster client (or server service client).  It doesn't really matter who is using the client for the purposes of these metrics.  For example, if we invented another boundary server "proxy" which could handle the requests to this cluster client, it would be nice to not have to rewrite any portion of this code to automatically get the metrics for it.

The name controller is also a bit misleading since there are 2 apis (maybe 3, if you seperate server management service and the session service) exposed by the boundary controller.  These metrics are only for the server management and session services (I think what we've often called the cluster).

Please update here and elsewhere in this file anywhere that refers to the worker and/or calls the the cluster (or server service) client the controller.

internal/servers/worker/internal/metric/controller_dialer.go
```diff
@@ -0,0 +1,156 @@
...
+
+	controllerServersPackage = "controller.servers.services.v1"
```
talanknight, line 24: Could you instead use a generated files instead of hardcoding this?  That way, if the generated file gets changed, this would automatically get updated or would result in a compile time error instead of just silently being wrong.

internal/servers/worker/internal/metric/controller_dialer.go
```diff
@@ -0,0 +1,156 @@
...
+			labelGrpcMethod:  method,
+			labelGrpcService: service,
+		},
+		start: time.Now(),
+	}
+
+	return r
+}
+
+func (r requestRecorder) record(resp interface{}, err error) {
```
talanknight, line 87: resp is not used in this function.
```suggestion
func (r requestRecorder) record(err error) {
```

internal/servers/worker/internal/metric/controller_dialer.go
```diff
@@ -0,0 +1,156 @@
...
+	r.labels[labelGrpcCode] = statusFromError(err).Code().String()
+	grpcRequestLatency.With(r.labels).Observe(time.Since(r.start).Seconds())
+}
+
+// InstrumentControllerDialer wraps a UnaryClientInterceptor and records
+// observations for the collectors associated with the worker<->controller
+// gRPC connection.
+func InstrumentControllerDialer() grpc.UnaryClientInterceptor {
```
talanknight, line 95: This isn't a dialer.  This is passed in as a dialer option, but is actually an interceptor.  Consider renaming this function to both specify that this is for the cluster (or server serivce) client and it's an interceptor, not a dialer.

Here and elsewhere you refer to this as dialer instrumentation.

talanknight (changes requested):

internal/servers/worker/internal/metric/controller_dialer.go
hugoamvieira: Addressed [072e97e](https://github.com/hashicorp/boundary/pull/1998/commits/072e97e4eeddfe56f09d75d23e49760605c3b21a)

internal/servers/worker/internal/metric/controller_dialer.go
hugoamvieira: Addressed [072e97e](https://github.com/hashicorp/boundary/pull/1998/commits/072e97e4eeddfe56f09d75d23e49760605c3b21a)

internal/servers/worker/internal/metric/controller_dialer.go
talanknight: You can use the `.Package()` method found [here](https://github.com/hashicorp/boundary/blob/main/internal/gen/controller/servers/services/session_service.pb.go#L930).

This is similar to what I did in the internal/metrics/api.go file [here](https://github.com/hashicorp/boundary/blob/main/internal/servers/controller/internal/metric/api.go#L83).

hugoamvieira: Missed that, thanks!

hugoamvieira: Addressed [e0ff5ce](https://github.com/hashicorp/boundary/pull/1998/commits/e0ff5cee189e8405fb4a99c64adaa5ea7687416f)

talanknight (dismissed):

talanknight (approved):

PR #2000, opened 2022-04-12
Update the cluster server metrics to use the gRPC status.Handler.

talanknight:
gRPC status.Handler gives us visibility into gRPC returned errors. In contrast, there are some situations where no metric would be counted at all using the interceptor approach (for example, in out of memory cases where gRPC would return a resource exhausted error).

This PR also removes the request size and response size metrics.  This is because those are difficult to observe along with the status code label.  In the future, we may revisit recording observations for the request and response sizes to the cluster server.

### DIFF
--- a/internal/servers/controller/internal/metric/cluster.go
+++ b/internal/servers/controller/internal/metric/cluster.go
@@ -3,16 +3,14 @@ package metric
 import (
 	"context"
 	"strings"
-	"time"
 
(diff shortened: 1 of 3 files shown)


A-440Hz (approved):
LGTM

hugoamvieira (approved):
LGTM. PR #2000 🥳

PR #2001, opened 2022-04-12
Classify all hostcatalog fields

johanbrandhorst:
For https://hashicorp.atlassian.net/browse/ICU-3921


internal/proto/controller/api/services/v1/host_catalog_service.proto
```diff
@@ -8,6 +8,9 @@ import "protoc-gen-openapiv2/options/annotations.proto";
 import "google/api/annotations.proto";
 import "google/protobuf/field_mask.proto";
 import "controller/api/resources/hostcatalogs/v1/host_catalog.proto";
+import "controller/custom_options/v1/options.proto";
+
+option (custom_options.v1.domain) = "hostcatalog";
```
johanbrandhorst, line 13: Is this the right domain? I am still not sure where to find that

johanbrandhorst: This appears to be a domain shared with `host`, so I'm going to change this to `host`.

johanbrandhorst:
Closing this in favor of doing one big PR with hosts, host-catalogs and host-sets

PR #2004, opened 2022-04-13
Add classification to host, hostsets and hostcatalogs

johanbrandhorst:
For https://hashicorp.atlassian.net/browse/ICU-3921


internal/proto/controller/api/resources/hostcatalogs/v1/host_catalog.proto
```diff
@@ -14,50 +14,52 @@ import "controller/custom_options/v1/options.proto";
...
+		// Attributes specific to the catalog type.
+		google.protobuf.Struct attributes = 100 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.subtype) = "default"];
+	}
 
 	// Secrets specific to the catalog type. These are never output.
 	google.protobuf.Struct secrets = 110 [(custom_options.v1.generate_sdk_option) = true];
 
 	// Output only. The HMAC of the last secrets supplied via the API, if any.
-	string secrets_hmac = 120;
+	string secrets_hmac = 120;  // @gotags: `class:"secret"`
```
jimlambrt, line 59: I think, this is public... it's already hmac'd

jimlambrt (changes requested):
I reviewed the protos only.

tmessi (commented):
other then the one jim mentioned this looks good, so i think if you fix that one you can merge this in

internal/proto/controller/api/resources/hostcatalogs/v1/host_catalog.proto
```diff
@@ -14,50 +14,52 @@ import "controller/custom_options/v1/options.proto";
...
 	// Output only. Scope information for this resource.
 	resources.scopes.v1.ScopeInfo scope = 30;
 
 	// The ID of the plugin of which this catalog is created.
-	string plugin_id = 34 [json_name="plugin_id", (custom_options.v1.generate_sdk_option) = true];
+	string plugin_id = 34 [json_name="plugin_id", (custom_options.v1.generate_sdk_option) = true];  // @gotags: `class:"public"`
 
 	// Output only. Plugin information for this resource.
-	resources.plugins.v1.PluginInfo plugin = 35;
+	resources.plugins.v1.PluginInfo plugin = 35;  // @gotags: `class:"public"`
```
jefferai, line 29: This shouldn't have a tag.

internal/proto/controller/api/resources/hostcatalogs/v1/host_catalog.proto
```diff
@@ -14,50 +14,52 @@ import "controller/custom_options/v1/options.proto";
...
 
 	// Optional name for identification purposes.
-	google.protobuf.StringValue name = 40 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = {this:"name" that: "name"}];
+	google.protobuf.StringValue name = 40 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = {this:"name" that: "name"}];  // @gotags: `class:"public"`
```
jefferai, line 32: Might be worth asking @jimlambrt how this will work. Generally only byte slices and strings will be classified and messages won't (but the contents may be) and this is a strange hybrid where if it's like a message this tag is unnecessary and if it's more like a string we should ensure that the classification library handles these proto types.

internal/proto/controller/api/services/v1/host_set_service.proto
```diff
@@ -135,33 +138,33 @@ service HostSetService {
...
 message ListHostSetsRequest {
-  string host_catalog_id = 1 [json_name="host_catalog_id"];
-  string filter = 30 [json_name="filter"];
+  string host_catalog_id = 1 [json_name="host_catalog_id"];  // @gotags: `class:"public"`
+  string filter = 30 [json_name="filter"];  // @gotags: `class:"public"`
 }
 
 message ListHostSetsResponse {
-  repeated api.resources.hostsets.v1.HostSet items = 1;
+  repeated api.resources.hostsets.v1.HostSet items = 1;  // @gotags: `class:"public"`
```
jefferai, line 154: This shouldn't have a classification

internal/proto/controller/api/resources/hostcatalogs/v1/host_catalog.proto
```diff
@@ -14,50 +14,52 @@ import "controller/custom_options/v1/options.proto";
...
 
-	// Attributes specific to the catalog type.
-	google.protobuf.Struct attributes = 100 [(custom_options.v1.generate_sdk_option) = true];
+	oneof attrs {
+		// Attributes specific to the catalog type.
+		google.protobuf.Struct attributes = 100 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.subtype) = "default"];
+	}
 
 	// Secrets specific to the catalog type. These are never output.
 	google.protobuf.Struct secrets = 110 [(custom_options.v1.generate_sdk_option) = true];
```
jefferai, line 56: This is related to my earlier comment/question about google proto types. `secrets` needs to be redacted and isn't marked here. If it will be done automatically with default handling of those types then no worries. It might be better to mark them explicitly though. I also don't know the current state of redacting maps...

johanbrandhorst: It currently only redacts the values, not the keys: https://github.com/hashicorp/boundary/pull/2004/files#diff-158f011052a662828d487cf6d4fd1f971593ebe611ae7dd6658126ecc100e0a3R124.

PR #2013, opened 2022-04-14
Unify plugin building

jefferai:
This causes the build.sh script to chain to plugins.sh which makes
plugin building behave the same for both local and release builds.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -38,27 +38,13 @@ case $(uname) in
 esac
 
 BINARY_SUFFIX=""
-if [ "${GOOS}x" == "windowsx" ]; then
+if [ "${GOOS}x" = "windowsx" ]; then
     BINARY_SUFFIX=".exe"
 fi
 
 # Build needed plugins first
 if [ "${SKIP_PLUGIN_BUILD}x" == "x" ]; then
-    ORIG_PATH=$(pwd);
-    for PLUGIN_TYPE in {"kms","host"}; do
-        echo "==> Building ${PLUGIN_TYPE} plugins..."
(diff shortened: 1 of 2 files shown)


scripts/build.sh
```diff
@@ -38,27 +38,13 @@ case $(uname) in
 esac
 
 BINARY_SUFFIX=""
-if [ "${GOOS}x" == "windowsx" ]; then
+if [ "${GOOS}x" = "windowsx" ]; then
```
johanbrandhorst, line 41: If two equal signs are too many here, should we be keeping them elsewhere? I prefer `[[` since it has more natural comparisons IIRC.

jefferai: I have zero strong opinions on this. I changed it because a change went into the plugins script related to this and I figured I should port that over. I'm happy either way. I think @tmessi made that change if he wants to comment.

tmessi: I changed it in a2dc9dfd40f0df88df23fb375904d51cf458136a I agree that `[[` is more natural, but is a bash extension. I believe using `[` is POSIX compliant, and technically an equals comparison should be `=` as `==` is not supported, however most versions of bash will allow `==` even with `[`. I don't recall the specific CI failure that led to that change, I think it had to do with how the GH action calls the script with `sh ./scripts/plugins.sh`, but I would say it would be good to be consistent, so either `[[` with `==` or `[` with `=`.

tmessi (approved):
this looks good otherwise, and I am fine if you merge it as is and we can address the `=` vs `==` thing at some other time.

mdeggies (approved):
Looks good to me. Thanks for working on the fix!

PR #2014, opened 2022-04-14
Do not register the build_info collectors multiple times.

talanknight:
Since we call base.NewServer multiple times when building the Commands map this was returning
an error "duplicate metrics collector registration attempted".  This PR only creates a single
server which is used for both the "server" command as well as "dev".


A-440Hz (approved):
thanks, LGTM

PR #2015, opened 2022-04-15
feat(target): Classify resources for audit events

tmessi:


tmessi:
@jefferai @jimlambrt this should be the last one, so if you can take a look at the classification.

Also flagged @mgaffney @louisruch since the refactor here is different compared to the other packages, and they probably have the best context for these changes, so if you two can look at [73103f2](https://github.com/hashicorp/boundary/pull/2015/commits/73103f292af4476292bd3ab7d416ff3d67cbe17b) in particular.

internal/proto/controller/api/resources/targets/v1/target.proto
```diff
@@ -158,8 +159,19 @@ message Target {
   // Output only. The egress credential sources associated with this Target.
   repeated CredentialSource egress_credential_sources = 510 [json_name = "egress_credential_sources"];
 
-  // The attributes that are applicable for the specific Target.
-  google.protobuf.Struct attributes = 200 [(custom_options.v1.generate_sdk_option) = true];
+  oneof attrs {
```
louisruch, line 162: Not that our comments have been great in the protos, even the previous comment didn't give much: `The attributes that are applicable for the specific Target.`.
But it might be useful to have something more than the auto generated `oneof` since I am not sure how clear `attrs` is:
```
// Types that are assignable to Attrs:
	//	*Target_Attributes
	//	*Target_TcpTargetAttributes
```

This is a minor nit and feel free to drop this for later

internal/proto/controller/api/resources/targets/v1/target.proto
```diff
@@ -158,8 +159,19 @@ message Target {
...
+    // The attributes that are applicable for the specific Target.
+    google.protobuf.Struct attributes = 200 [
+			(google.api.field_visibility).restriction = "INTERNAL",
+      (custom_options.v1.generate_sdk_option) = true,
+			(custom_options.v1.subtype) = "default"
+    ];
+		TcpTargetAttributes tcp_target_attributes = 201 [
```
louisruch: Similarly might be good to have a comment on the `TcpTargetAttributes`...

louisruch (approved):
Some minor nits around comments, but I see other PRs have it as you do so feel free to ignore for now

internal/proto/controller/api/resources/targets/v1/target.proto
tmessi: I agree that we should look at these comments. But there is a particular consideration with these in how they relate to the openapi/swagger output. These comments are used in the generated swagger doc, except when a field has `(google.api.field_visibility).restriction = "INTERNAL"`, so in the case of the `TcpTargetAttributes` mentioned in your other comment, any comments would be dropped from the swagger doc, which is intentional. 

As for the Attrs field itself in the generated go code, I found the auto generated doc useful enough to direct a developer to the structs that could be in the oneof, which is think is sufficient *if* those structs are well documented.

jimlambrt (approved):

internal/servers/controller/handlers/targets/registry.go
```diff
@@ -56,45 +60,39 @@ func (r registry) maskManager(s subtypes.Subtype) (handlers.MaskManager, error)
...
 
-	return attr, nil
+	return re.setAttrFunc(in, out)
 }
 
 var subtypeRegistry = registry{}
 
 // Register registers a subtype for used by the service handler.
-func Register(s subtypes.Subtype, maskManager handlers.MaskManager, af attributeFunc) {
+func Register(s subtypes.Subtype, maskManager handlers.MaskManager, af attributeFunc, sf setAttributeFunc) {
```
johanbrandhorst, line 91: Do we plan on keeping this around after the merge to enterprise? I'm not sure I understand the full scope of this but should we not be able to rip all this out and just switch on subtype like we do in other parts of the code?

johanbrandhorst (changes requested):

internal/proto/controller/api/resources/targets/v1/target.proto
```diff
@@ -92,90 +93,100 @@ message SessionCredential {
...
   google.protobuf.UInt32Value default_port = 10
-      [json_name = "default_port", (custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = { this: "attributes.default_port" that: "DefaultPort" }];
+      [json_name = "default_port", (custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = { this: "attributes.default_port" that: "DefaultPort" }]; // @gotags: `class:"public"`
 }
 
 // WorkerInfo contains information about workers, returned in to the client in SessionAuthorization
 message WorkerInfo {
   // Output only. The address of the worker.
-  string address = 10;  // @gotags: `class:"public"`
+  string address = 10;  // @gotags: `class:"public"` // @gotags: `class:"public"`
```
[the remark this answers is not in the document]
tmessi: yep

jefferai (approved):

PR #2016, opened 2022-04-15
feat(event): Remove BOUNDARY_DEVELOPER_ENABLE_EVENTS env var

tmessi:


globals/boundary_developer_env.go
```diff
@@ -1,8 +1 @@
 package globals
```
tmessi, line 1: I guess I could remove this file completely, but I didnt know if we wanted to leave it around in case we create any new developer consts.

jimlambrt: Happy to leave it for now.

jimlambrt (approved):
Ty!

PR #2017, opened 2022-04-15
Switch request.WithContext to request.Clone

jefferai:
The Go docs now say to use request.Clone (introduced in Go
1.13) instead of request.WithContext (introduced in Go 1.7) as it deep
copies instead of shallow copies.


talanknight (approved):
Looks good. Thank you!

PR #2019, opened 2022-04-15
refactor (tests): GetEventDetails(...) will stop returning nil

jimlambrt:


tmessi (approved):
this seems fine to me assuming ci passes, although some additional context would be welcome.

PR #2020, opened 2022-04-15
Register the build info metrics using prometheus.DefaultRegistry in an init function

talanknight:
This fixes errors related to calling NewServer or NewCommand multiple times when using autocomplete.


A-440Hz (approved):
LGTM

Issue #2003, opened 2022-04-13
Flaky tests list


johanbrandhorst:
This issue tracks tests that are known to be flaky.

- `TestServer_ReloadWorkerTags` in `internal/cmd/commands/server`
   This will sometimes fail like this:
   ```
   === CONT  TestServer_ReloadWorkerTags
      worker_tags_reload_test.go:166: 
            Error Trace:	worker_tags_reload_test.go:146
                                    worker_tags_reload_test.go:166
            Error:      	Not equal: 
                           expected: "foo"
                           actual  : "type"
                           
                           Diff:
                           --- Expected
                           +++ Actual
                           @@ -1 +1 @@
                           -foo
                           +type
            Test:       	TestServer_ReloadWorkerTags
   FAIL
   FAIL	github.com/hashicorp/boundary/internal/cmd/commands/server	84.648s
   ```
   [Failing log](https://github.com/hashicorp/boundary/files/8482294/log.txt). The exact cause is unknown at this time, but may be related to the amount of time spent sleeping in the test (sometimes it may not be long enough).
- `Test_statusCodeInterceptor/nil-err` in `internal/servers/controller`
   ```
   testing.go:802: 
            Error Trace:	testing.go:802
                                    testing.go:912
                                    testing.go:1049
                                    testing.go:1253
            Error:      	Received unexpected error:
                           closed
            Test:       	Test_statusCodeInterceptor/nil-err
   === RUN   Test_statusCodeInterceptor/hello
   --- FAIL: Test_statusCodeInterceptor (0.00s)
      --- PASS: Test_statusCodeInterceptor/nil-nil (0.00s)
      --- FAIL: Test_statusCodeInterceptor/nil-err (0.00s)
   ```
   [Failing log](https://github.com/hashicorp/boundary/files/8522634/log.txt).
-  `TestServer_ShutdownWorker` in `internal/cmd/commands/server`
   ```
   worker_shutdown_reload_test.go:148: 
            Error Trace:	testing_helper.go:138
                                    testing_helper.go:312
                                    worker_shutdown_reload_test.go:148
            Error:      	Received unexpected error:
                           failed to read protobuf message: failed to get reader: received close frame: status = StatusProtocolError and reason = "received header with unexpected rsv bits set: false:false:true"
            Test:       	TestServer_ShutdownWorker
   --- FAIL: TestServer_ShutdownWorker (1.78s)
   ```
[...]

PR #2031, opened 2022-04-19
Classify audit events

tmessi:


johanbrandhorst:
```
=== CONT  TestServer_ReloadWorkerTags
    worker_tags_reload_test.go:166: 
        	Error Trace:	worker_tags_reload_test.go:146
        	            				worker_tags_reload_test.go:166
        	Error:      	Not equal: 
        	            	expected: "foo"
        	            	actual  : "type"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-foo
        	            	+type
        	Test:       	TestServer_ReloadWorkerTags
--- FAIL: TestServer_ReloadWorkerTags (23.00s)
FAIL
FAIL	github.com/hashicorp/boundary/internal/cmd/commands/server	85.289s
```

Looks like we messed up the rebase slightly.

tmessi:
> 

I don't think so, see https://github.com/hashicorp/boundary/issues/2003

johanbrandhorst:
> > 
> 
> I don't think so, see #2003

How did I not remember this 😂

internal/api/genapi/parse.go
```diff
@@ -40,24 +40,41 @@ func parsePBs() {
...
+			isAttribute := false
+			if proto.GetExtension(opts, protooptions.E_Subtype).(string) != "" {
+				isAttribute = true
+			}
+			if isAttribute {
+				// Only parse attributes oneOf once and not for each option
+				if strutil.StrListContains(attributesFilter, "Attributes") {
+					continue
+				}
+			}
```
johanbrandhorst, line 56: I wonder if we could simplify this now, by simply checking if the subtype is different from `"default"`, and if so, excluding the field. What do you think?

internal/gen/controller.swagger.json
```diff
@@ -3699,8 +3699,7 @@
           "description": "The ID of the Auth Method that is associated with this Account."
         },
         "attributes": {
-          "type": "object",
-          "description": "The attributes that are applicable for the specific Account type."
```
johanbrandhorst, line 3703: This comment disappearing implies that we misplaced a comment in a proto file somewhere, will investigate.

internal/proto/controller/api/resources/accounts/v1/account.proto
```diff
@@ -7,73 +7,86 @@ option go_package = "github.com/hashicorp/boundary/sdk/pbs/controller/api/resour
...
-	string subject = 90 [json_name="subject", (custom_options.v1.generate_sdk_option) = true];
+	string subject = 90 [json_name="subject", (custom_options.v1.generate_sdk_option) = true];  // @gotags: `class:"sensitive"`
 
 	// Output only. full_name is a string that maps to the OIDC name claim.
-	string full_name = 100 [json_name="full_name"];
+	string full_name = 100 [json_name="full_name"];  // @gotags: `class:"sensitive"`
 
 	// Output only. email is a string that maps to the OIDC email claim.
-	string email = 110;
+	string email = 110;  // @gotags: `class:"sensitive"`
```
johanbrandhorst, line 89: This seems too conservative now

internal/gen/controller.swagger.json
johanbrandhorst: Fixed

internal/proto/controller/api/resources/accounts/v1/account.proto
```diff
@@ -7,73 +7,86 @@ option go_package = "github.com/hashicorp/boundary/sdk/pbs/controller/api/resour
...
 	// Output only. The ID of the Account.
-	string id = 10;
+	string id = 10;  // @gotags: `class:"public"`
 
 	// Output only. Scope information for the Account.
 	resources.scopes.v1.ScopeInfo scope = 20;
 
 	// Optional name for identification purposes.
-	google.protobuf.StringValue name = 30 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = {this:"name" that: "Name"}];
+	google.protobuf.StringValue name = 30 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = {this:"name" that: "Name"}];  // @gotags: `class:"sensitive"`
```
jimlambrt, line 23: perhaps `public`?

internal/proto/controller/api/resources/accounts/v1/account.proto
```diff
@@ -7,73 +7,86 @@ option go_package = "github.com/hashicorp/boundary/sdk/pbs/controller/api/resour
...
-	google.protobuf.StringValue password = 20 [(custom_options.v1.generate_sdk_option) = true];
+	google.protobuf.StringValue password = 20 [(custom_options.v1.generate_sdk_option) = true];  // @gotags: `class:"secret"`
 }
 
 // Attributes associated only with Accounts with type "oidc".
 message OidcAccountAttributes {
 	// issuer is a case sensitive URL that maps to the OIDC iss claim.
 	// This value is immutable after creation time.
-	string issuer = 80 [json_name="issuer", (custom_options.v1.generate_sdk_option) = true];
+	string issuer = 80 [json_name="issuer", (custom_options.v1.generate_sdk_option) = true];  // @gotags: `class:"sensitive"`
```
jimlambrt, line 79: perhaps `public`?

internal/proto/controller/api/resources/accounts/v1/account.proto
```diff
@@ -7,73 +7,86 @@ option go_package = "github.com/hashicorp/boundary/sdk/pbs/controller/api/resour
...
 
 	// subject is a case sensitive string that maps to the OIDC sub claim.
 	// This value is immutable after creation time.
-	string subject = 90 [json_name="subject", (custom_options.v1.generate_sdk_option) = true];
+	string subject = 90 [json_name="subject", (custom_options.v1.generate_sdk_option) = true];  // @gotags: `class:"sensitive"`
```
jimlambrt, line 83: perhaps `public`?

internal/proto/controller/api/resources/accounts/v1/account.proto
jimlambrt, line 89: perhaps `public`?

internal/proto/controller/api/resources/accounts/v1/account.proto
johanbrandhorst: Fixed, thanks!

internal/proto/controller/api/resources/accounts/v1/account.proto
```diff
@@ -7,73 +7,86 @@ option go_package = "github.com/hashicorp/boundary/sdk/pbs/controller/api/resour
...
 	// Output only. Scope information for the Account.
 	resources.scopes.v1.ScopeInfo scope = 20;
 
 	// Optional name for identification purposes.
-	google.protobuf.StringValue name = 30 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = {this:"name" that: "Name"}];
+	google.protobuf.StringValue name = 30 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = {this:"name" that: "Name"}];  // @gotags: `class:"sensitive"`
 
 	// Optional user-set description for identification purposes.
-	google.protobuf.StringValue description = 40 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = {this:"description" that: "Description"}];
+	google.protobuf.StringValue description = 40 [(custom_options.v1.generate_sdk_option) = true, (custom_options.v1.mask_mapping) = {this:"description" that: "Description"}];  // @gotags: `class:"sensitive"`
```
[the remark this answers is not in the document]
johanbrandhorst: Fixed, thanks!

internal/proto/controller/api/resources/accounts/v1/account.proto
johanbrandhorst: Fixed, thanks!

internal/proto/controller/api/resources/accounts/v1/account.proto
johanbrandhorst: Fixed, thanks!

internal/servers/controller/handlers/credentialstores/credentialstore_service.go
```diff
@@ -34,6 +35,7 @@ const (
 	caCertsField        = "attributes.ca_cert"
 	clientCertField     = "attributes.client_certificate"
 	clientCertKeyField  = "attributes.certificate_key"
+	domain              = "credential"
```
johanbrandhorst, line 38: We don't use a constant for the domain anywhere else, is there a good reason we should keep it here, or conversely, should we add it everywhere else?

internal/servers/controller/handlers/targets/target_service.go
```diff
@@ -45,6 +46,10 @@ import (
 	"google.golang.org/protobuf/types/known/wrapperspb"
 )
 
+const (
+	credentialDomain = "credential"
```
johanbrandhorst, line 50: We're using both `credential`, `host` and `target` domains in the target service, is this intentional?

internal/servers/controller/handlers/credentialstores/credentialstore_service.go
tmessi: I thought we had a constant for domain in the other handler subpackages, but maybe we were not consistent with this. Looking now at the host_sets for example, it looks like we are using the same `"host"` string over and over. I think making a constant for that string would be better imo.

johanbrandhorst: Yeah that seems reasonable. I'll do it now.

johanbrandhorst (dismissed):

johanbrandhorst (approved):

PR #2033, opened 2022-04-20
Format protobuf files with buf and enforce formatting in CI

johanbrandhorst:
## [feat(proto): Format protobuf files with buf](https://github.com/hashicorp/boundary/commit/1bd7668774594f87558743f6e2d862264d04012d)

Formats all protobuf files with buf format. This ensures
a consistent style across all our protobuf files, similar
to gofumpt. It is notable that the existing protobuf
files were sometimes using tabs and sometimes using spaces,
causing unsightly and hard-to-catch diffs when copying
contents between files.

<hr>

I propose that immediately after this, we add a `.git-blame-ignore-revs` file that will allow the changes from the formatting commit to be ignored when using `git blame` and on Github. See https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view for more information.


.circleci/config/jobs/fmt.yml
```diff
@@ -0,0 +1,13 @@
+executor: go-machine
+steps:
+- checkout
+- install-go
+- run:
+    name: "Install tools"
+    command: make tools
+- run:
+    name: "Format files"
+    command: make fmt
```
tmessi, line 10: This should be covered by [make-gen-deltas](https://github.com/hashicorp/boundary/blob/1342bd018317a44fd9f97dd93c31523941306c29/.circleci/config/jobs/make-gen-deltas.yml) since `make fmt` is call via `make gen`

mgaffney (approved):

tmessi (commented):
I think this is great, just one comment about the new ci job, it isn't needed.

.circleci/config/jobs/fmt.yml
johanbrandhorst: Good catch!

johanbrandhorst:
Going to wait with merging this until we've got 0.8 out

.circleci/config/jobs/fmt.yml
johanbrandhorst: Removed

tmessi (approved):

PR #2034, opened 2022-04-20
Make the metrics regexp less strict when matching against the {*id} vars for the "path" label.

talanknight:
Previously we were matching what boundary considers an id, but that is determined after requests are routed by grpc-gateway.
This changes the regexp to match any path segment to try to behave the same way that grpc-gateway would when assigning values
to the {id} variable declared in the protobuf.

### DIFF
--- a/internal/servers/controller/internal/metric/api.go
+++ b/internal/servers/controller/internal/metric/api.go
@@ -91,9 +91,10 @@ func apiPathsAndMethods() map[string][]string {
 }
 
 func buildRegexFromPath(p string) *regexp.Regexp {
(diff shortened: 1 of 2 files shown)


A-440Hz (approved):
lgtm

hugoamvieira (approved):

hugoamvieira:
(Build failed though, looks like due to circle ci issues - Might want to retry it)

johanbrandhorst (approved):
I can't say I understand fully the grpc-gateway path matching, but this looks fine to me

PR #2027, opened 2022-04-19
refactor (kms): adopt the go-kms-wrapping/extras/kms package

jimlambrt:
This PR refactors boundary to use [go-kms-wrapping/extras/kms](https://github.com/hashicorp/go-kms-wrapping/tree/main/extras/kms) package.  

Among other things, this refactor entails migrating the existing kms DEK tables into a new table required by the new pkg.  At this point, we've decide to keep the deprecated DEK tables and drop them in future release, when we're confident they won't be needed for any reason.  This future work is tracked in: https://github.com/hashicorp/boundary/issues/2026

Most of the refactor is fairly straightforward.  I will add comments to reviewers to point out areas which should be closely reviewed.

PR #2037, opened 2022-04-25
Add session listing benchmarks

johanbrandhorst:
## [feat(dbtest): Add docker-load Makefile target](https://github.com/hashicorp/boundary/pull/2037/commits/74b2ef9e0c4d22bc7ce7312c1677e0dac0fd3462)

The new docker-load target uses docker buildx to build
a local image of each of the Dockerfiles, and load them
into the local docker socket image store.

## [refact(all): Use testing.TB for test helpers](https://github.com/hashicorp/boundary/pull/2037/commits/4070486a4bf7acdc74e7ca6fadf504cfd3396564)

Our test helper packages were not possible to use
with benchmarks. This refactor switches to testing.TB
to allow both tests and benchmarks to use the helpers.

## [test(session): Ignore unique error on session create](https://github.com/hashicorp/boundary/pull/2037/commits/d685bf59f2ff586546069f472b034d27bcbe992c)

Sometimes when creating a lot of sessions concurrently we
can run into a unique constraint error in the warehouse triggers
in the database. Simply retrying the call will allow the
session creation to succeed.

## [feat(dbtest): Mount DB dumps and create templates](https://github.com/hashicorp/boundary/pull/2037/commits/9c8e29833d97e5f4f60832cf58b9d8ad9867bf81)

Mount the benchmark dump directory when running
the test database and create a new script for creating
templates from the dumps. This makes it easy to create
pre-populated databases for benchmark tests.

Fixes https://hashicorp.atlassian.net/browse/ICU-4040

## [feat(sessions): Add list benchmarks](https://github.com/hashicorp/boundary/pull/2037/commits/fa6cac87b7dedaedf44e110ccbc1db2880525553)

The list benchmark tests uses pre-created database
dumps to run its benchmarks. This benchmark
will be used to guide optimizations around listing
of sessions.

Fixes https://hashicorp.atlassian.net/browse/ICU-4042

## [feat(dbtest): Add benchmark dump generator](https://github.com/hashicorp/boundary/pull/2037/commits/90e87d58c9e50d37d82261cfc5f40c26dfada71c)

Generate dumps ahead of time so that they can
be used for benchmarks without prohibitive
startup costs.

Fixes https://hashicorp.atlassian.net/browse/ICU-4039

### DIFF
--- a/internal/kms/testing.go
+++ b/internal/kms/testing.go
@@ -10,7 +10,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestRootKey(t *testing.T, conn *db.DB, scopeId string) *RootKey {
+func TestRootKey(t testing.TB, conn *db.DB, scopeId string) *RootKey {
 	t.Helper()
 	require := require.New(t)
 	db.TestDeleteWhere(t, conn, func() interface{} { i := AllocRootKey(); return &i }(), "scope_id = ?", scopeId)
@@ -25,7 +25,7 @@ func TestRootKey(t *testing.T, conn *db.DB, scopeId string) *RootKey {
 	return k
 }
 
-func TestRootKeyVersion(t *testing.T, conn *db.DB, wrapper wrapping.Wrapper, rootId string) (kv *RootKeyVersion, kvWrapper wrapping.Wrapper) {
+func TestRootKeyVersion(t testing.TB, conn *db.DB, wrapper wrapping.Wrapper, rootId string) (kv *RootKeyVersion, kvWrapper wrapping.Wrapper) {
 	t.Helper()
 	require := require.New(t)
 	rw := db.New(conn)
@@ -46,7 +46,7 @@ func TestRootKeyVersion(t *testing.T, conn *db.DB, wrapper wrapping.Wrapper, roo
 	return k, rootKeyVersionWrapper
 }
 
-func TestKms(t *testing.T, conn *db.DB, rootWrapper wrapping.Wrapper) *Kms {
+func TestKms(t testing.TB, conn *db.DB, rootWrapper wrapping.Wrapper) *Kms {
 	t.Helper()
 	require := require.New(t)
 	rw := db.New(conn)
@@ -59,7 +59,7 @@ func TestKms(t *testing.T, conn *db.DB, rootWrapper wrapping.Wrapper) *Kms {
 	return kms
 }
 
-func TestDatabaseKey(t *testing.T, conn *db.DB, rootKeyId string) *DatabaseKey {
+func TestDatabaseKey(t testing.TB, conn *db.DB, rootKeyId string) *DatabaseKey {
 	t.Helper()
 	require := require.New(t)
 	rw := db.New(conn)
@@ -75,7 +75,7 @@ func TestDatabaseKey(t *testing.T, conn *db.DB, rootKeyId string) *DatabaseKey {
 	return k
 }
 
-func TestDatabaseKeyVersion(t *testing.T, conn *db.DB, rootKeyVersionWrapper wrapping.Wrapper, databaseKeyId string, key []byte) *DatabaseKeyVersion {
+func TestDatabaseKeyVersion(t testing.TB, conn *db.DB, rootKeyVersionWrapper wrapping.Wrapper, databaseKeyId string, key []byte) *DatabaseKeyVersion {
 	t.Helper()
 	require := require.New(t)
 	rw := db.New(conn)
@@ -95,7 +95,7 @@ func TestDatabaseKeyVersion(t *testing.T, conn *db.DB, rootKeyVersionWrapper wra
 	return k
 }
 
-func TestOplogKey(t *testing.T, conn *db.DB, rootKeyId string) *OplogKey {
+func TestOplogKey(t testing.TB, conn *db.DB, rootKeyId string) *OplogKey {
 	t.Helper()
 	require := require.New(t)
 	rw := db.New(conn)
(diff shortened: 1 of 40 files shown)


testing/dbtest/session_benchmark_test.go
```diff
@@ -0,0 +1,230 @@
...
+// the testing framework for easy access to a boundary server for
+// populating a database with test data. This database is then dumped
+// to predictable files for use in benchmarking.
+func TestGenerateSessionBenchmarkTemplateDumps(t *testing.T) {
+	if os.Getenv("BOUNDARY_DB_TEST_GENERATE_SESSION_BENCHMARK_TEMPLATE_DUMPS") == "" {
+		t.Skip("BOUNDARY_DB_TEST_GENERATE_SESSION_BENCHMARK_TEMPLATE_DUMPS is not set")
+		return
+	}
+
+	_, err := exec.LookPath("pg_dump")
```
tmessi, line 34: might want to make an environment variable for this string. That would make it easier for people to use the correct version. It isn't uncommon to have multiple `pg_dump` versions installed. For example I have `pg_dump11`, `pg_dump13` `pg_dump14`.

testing/dbtest/session_benchmark_test.go
```diff
@@ -0,0 +1,230 @@
...
+		},
+	}
+
+	for _, scenario := range scenarios {
+		scenario := scenario // Parallel test closures act as goroutines, copy iteration variable
+		t.Run(fmt.Sprintf("Generate-%d-sessions-%d-conns-per-session-%d%%-authorized-ratio-dump", scenario.sessions, scenario.connsPerSession, scenario.authorizedRatio), func(t *testing.T) {
+			t.Parallel() // Lets speed things up a bit
+			ctx := context.Background()
+			require := require.New(t)
+			// Name file after number of sessions
```
tmessi, line 103: this comment is a little out of place now, either update it or just remove it

internal/kms/testing.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestRootKey(t *testing.T, conn *db.DB, scopeId string) *RootKey {
+func TestRootKey(t testing.TB, conn *db.DB, scopeId string) *RootKey {
```
tmessi, line 13: This might be a notable thing for @jimlambrt and #2027

testing/dbtest/session_benchmark_test.go
```diff
@@ -0,0 +1,230 @@
...
+	"github.com/hashicorp/boundary/internal/target/tcp"
+	"github.com/hashicorp/boundary/testing/dbtest"
+	"github.com/stretchr/testify/require"
+)
+
+// TestGenerateSessionBenchmarkTemplateDumps is not really a test, it uses
+// the testing framework for easy access to a boundary server for
+// populating a database with test data. This database is then dumped
+// to predictable files for use in benchmarking.
+func TestGenerateSessionBenchmarkTemplateDumps(t *testing.T) {
```
tmessi, line 28: can you add a make target for running this?

internal/servers/controller/handlers/sessions/session_list_benchmarks_test.go
```diff
@@ -0,0 +1,107 @@
...
+			{clients: 2000},
+		*/
+	}
+
+	for _, bc := range cases {
+		for _, template := range []template{{
+			name:                        dbtest.Boundary1000Sessions50ConnsPerSession50Template,
+			sessions:                    1000,
+			connsPerSession:             50,
+			authorizedSessionPercentage: dbtest.BobRatio,
```
tmessi, line 45: This seems odd, the coupling of the string const in the `dbtest` pacakge with the other variables here, it seems like this should all be defined together as part of a single struct.

tmessi (commented):
looking good. I will try this out, but also have some comments

testing/dbtest/docker/Makefile
```diff
@@ -49,4 +59,12 @@ clean:
 	docker stop $(TEST_CONTAINER_NAME) || true
 	docker rm -v $(TEST_CONTAINER_NAME) || true
 
-.PHONY: all docker-build database-up ${docker-buildxs} clean
+generate-database-dumps:
```
tmessi, line 62: can you add a target to the top level Makefile that calls this target?

We do this for other targets in this makefile like [test-database-up](https://github.com/hashicorp/boundary/blob/a1cf3234ed088a738b53997eeee0eeb8883b9ef7/Makefile#L209-L211)

testing/dbtest/template.go
```diff
@@ -44,6 +58,12 @@ var supportedDialects = map[string]struct{}{
 var supportedTemplates = map[string]struct{}{
 	BoundaryTemplate: {},
 	Template1:        {},
+	Boundary1000Sessions10ConnsPerSession10Template:  {},
+	Boundary1000Sessions10ConnsPerSession25Template:  {},
+	Boundary1000Sessions10ConnsPerSession50Template:  {},
+	Boundary1000Sessions10ConnsPerSession75Template:  {},
+	Boundary1000Sessions10ConnsPerSession100Template: {},
+	Boundary1000Sessions10ConnsPerSession500Template: {},
```
tmessi, line 66: this is getting cumbersome as we add additional datasets. I think we could re-work this to be a regex? Could be done in the future if this list grows though.

tmessi (commented):
looks good, just a couple minor comments and thoughts. The only one I would really consider a "requested change" is the makefile one.

testing/dbtest/template.go
johanbrandhorst: I think we can wait and see how this grows, for now I'd prefer not to add a regex here.

testing/dbtest/docker/Makefile
johanbrandhorst: Fixed

PR #2039, opened 2022-04-26
Add concurrent session creation test

johanbrandhorst:
The test currently fails because of warehouse constraint
violations.


johanbrandhorst:
Not going to do anything about this after discussing with Mike. If someone sees this error in the wild it is possible to simply retry.

PR #2042, opened 2022-04-28
Reorder authz check for sessions

jefferai:
This implements a mechanism for service handlers to pass a repo satisfying an interface into the scope IDs listing function. The purpose of this interface is to allow the same function to perform authorization checking against resources rather than requiring service handlers to fetch all resources in the determined set of scopes and then perform this checking. While this could be done in each service handler, this approach centralizes the logic, which has the nice benefit of removing some boilerplate from service handler list functions that have adapted to it.

Tests haven't changed, which is expected -- the function should return exactly what was being returned before, simply faster.

### DIFF
--- a/internal/session/repository_session.go
+++ b/internal/session/repository_session.go
@@ -206,7 +206,7 @@ func (r *Repository) LookupSession(ctx context.Context, sessionId string, _ ...O
 	return &session, authzSummary, nil
 }
 
-// FetchAuthzProtectedEntityInfo implements boundary.AuthzProtectedEntityProvider
+// FetchAuthzProtectedEntitiesByScope implements boundary.AuthzProtectedEntityProvider
 func (r *Repository) FetchAuthzProtectedEntitiesByScope(ctx context.Context, scopeIds []string) (map[string][]boundary.AuthzProtectedEntity, error) {
 	const op = "session.(Repository).FetchAuthzProtectedEntityInfo"
 


internal/servers/controller/common/scopeids/scope_ids.go
```diff
@@ -15,70 +15,121 @@ import (
...
+	// info for each returned value
+	output.ScopeResourceMap = map[string]*ScopeInfoWithResourceIds{}
+
 	// Base case: if not recursive, return the scope we were given and the
 	// already-looked-up info
-	if !recursive {
-		return []string{authResults.Scope.Id}, map[string]*scopes.ScopeInfo{authResults.Scope.Id: authResults.Scope}, nil
+	if !input.Recursive {
+		output.ScopeResourceMap[input.AuthResults.Scope.Id] = &ScopeInfoWithResourceIds{ScopeInfo: input.AuthResults.Scope}
+		// If we don't have information do to the lookup ourselves, return what
```
johanbrandhorst, line 106:
```suggestion
		// If we don't have information to do the lookup ourselves, return what
```

johanbrandhorst (approved):
Don't have a ton of context here but the code looks good!

internal/types/resource/resource.go
```diff
@@ -90,3 +93,17 @@ var Map = map[string]Type{
...
+	PublicId string
+	ScopeId  string
+	UserId   string
+}
+
+type MinimalInfoRepo interface {
+	// Fetches resource IDs for the given scopes. Note that it is assumed that
+	// the set of scopes is either length 1 or is recursive; that is, the reason
+	// we would have multiple scopes is because there is a parent or set of
+	// parents and any children.
```
tmessi, line 107: This note and its implications are a little tough to grok. When we chatted about it it seemed that main thing that an implementer needs to address is the different between fetching for all scopes vs fetching for a specific scope. This could be expressed a few ways such as having separate methods (i.e. `FetchIdsForScopes`, `FetchIds`), or using an option pattern ie `FetchIds(context.Context, Option...)` or clarifying the behavior if `scopesIds []string` is empty.

internal/types/resource/resource.go
```diff
@@ -90,3 +93,17 @@ var Map = map[string]Type{
...
 	CredentialLibrary.String(): CredentialLibrary,
 }
+
+type MinimalInfo struct {
+	PublicId string
+	ScopeId  string
+	UserId   string
+}
+
+type MinimalInfoRepo interface {
```
tmessi, line 103: The names of this struct and this interface could be revisited. At first read it isn't clear what they are and what their purpose is. And perhaps the package they are in adds to the difficulty of understanding what they are.

internal/types/resource/resource.go
```diff
@@ -90,3 +93,17 @@ var Map = map[string]Type{
...
+	// Fetches resource IDs for the given scopes. Note that it is assumed that
+	// the set of scopes is either length 1 or is recursive; that is, the reason
+	// we would have multiple scopes is because there is a parent or set of
+	// parents and any children.
+	FetchIdsForScopes(context.Context, []string) (map[string][]MinimalInfo, error)
```
tmessi, line 108: Does this need to return a map? It looks like just a `[]MinimalInfo` would be sufficient given how it is used [here](https://github.com/hashicorp/boundary/pull/2042/files#diff-dd4cd0303a44969a76a67e6bcd1024a00b29afcf3bb92ca626b844fac278735eR271-R295)

tmessi: I also think this method could be renamed. It is returning more then just ids, perhaps once a different name for `MinimalInfo` is picked, it would help with this name as well.

tmessi (commented):
left some comments to note some of the things we chatted about

internal/types/resource/resource.go
```diff
@@ -94,16 +94,19 @@ var Map = map[string]Type{
...
-	PublicId string
-	ScopeId  string
-	UserId   string
+// BasicInfo is used by some functions (primarily
+// BasicInfoRepo-conforming implementations) to deliver some common information
+// necessary for calculating authz.
+type BasicInfo interface {
+	GetPublicId() string
+	GetScopeId() string
+	GetUserId() string
```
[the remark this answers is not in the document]
jefferai: Ah I thought Mike had done something! I can move it there.

internal/types/resource/resource.go
jefferai: The map is used in the service handlers to fetch authorized actions based on the resource ID.

internal/types/resource/resource.go
```diff
@@ -94,16 +94,19 @@ var Map = map[string]Type{
...
 }
 
-type MinimalInfoRepo interface {
+type BasicInfoRepo interface {
```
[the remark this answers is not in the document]
jefferai: Good point. Updated.

internal/types/resource/resource.go
```diff
@@ -1,6 +1,8 @@
 package resource
 
-import "encoding/json"
+import (
+	"encoding/json"
+)
```
tmessi, line 5: this change could be dropped now that nothing is being added to this file

jefferai: Done

internal/servers/controller/common/scopeids/scope_ids.go
```diff
@@ -15,70 +16,127 @@ import (
...
-// tests in the other service handlers test this function extensively as it
-// forms the basis for all recursive listing tests; see those tests for list
-// functionality in the various service handlers.
-func GetListingScopeIds(
+type authzProtectedEntityProvider interface {
+	// Fetches basic resource info for the given scopes. Note that this is a
+	// "where clause" style of argument: if the set of scopes is populated these
+	// are the scopes to limit to (e.g. to put in a where clause). An empty set
+	// of scopes means to look in *all* scopes, not none!
+	FetchAuthzProtectedEntityInfo(context.Context, []string) (map[string][]boundary.AuthzProtectedEntity, error)
```
tmessi, line 24: having a named type for this map might help clarify what the keys need to be, maybe `AuthProtectedEntitiesByScope`? Figuring out a name for that map might also help with the function name, ie `FetchAuthzProtectedEntitiesByScope`

Also using named parameters for the interface might be a good idea, ie:
```
FetchAuthzProtectedEntityInfo(ctx context.Context, scopeIds []string)
```

internal/session/repository_session.go
```diff
@@ -205,6 +206,51 @@ func (r *Repository) LookupSession(ctx context.Context, sessionId string, _ ...O
 	return &session, authzSummary, nil
 }
 
+// FetchAuthzProtectedEntityInfo implements boundary.AuthzProtectedEntityProvider
```
tmessi, line 209: this comment needs updating to match the function name

tmessi (approved):

PR #2045, opened 2022-04-29
Update targets to new listing method

jefferai:
This uses the prefetch method to fetch resource IDs and validate them
for authz by the user before then fetching the full list of targets.

### DIFF
--- a/internal/session/repository_session.go
+++ b/internal/session/repository_session.go
@@ -220,7 +220,7 @@ func (r *Repository) FetchAuthzProtectedEntitiesByScope(ctx context.Context, sco
 		return nil, errors.New(ctx, errors.InvalidParameter, op, "no scopes given")
 	case 1:
 		inClauseCnt += 1
-		where, args = fmt.Sprintf("where scope_id = @%d", inClauseCnt), append(args, sql.Named("1", scopeIds[0]))
+		where, args = fmt.Sprintf("where scope_id = @%d", inClauseCnt), append(args, sql.Named(fmt.Sprintf("%d", inClauseCnt), scopeIds[0]))
 	default:
 		idsInClause := make([]string, 0, len(scopeIds))
 		for _, id := range scopeIds {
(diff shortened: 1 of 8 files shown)


internal/target/repository.go
```diff
@@ -138,21 +141,98 @@ func (r *Repository) LookupTarget(ctx context.Context, publicIdOrName string, op
...
+	var numRows int
+	for rows.Next() {
+		numRows++
+		var tv targetView
+		if err := r.reader.ScanRows(ctx, rows, &tv); err != nil {
+			return nil, errors.Wrap(ctx, err, op, errors.WithMsg("scan row failed"))
+		}
+		targetsMap[tv.GetScopeId()] = append(targetsMap[tv.GetScopeId()], tv)
+	}
+
```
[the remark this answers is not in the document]
johanbrandhorst: Apparently `ScanRows` checks the iteration error already

internal/session/repository_session.go
```diff
@@ -263,38 +264,38 @@ func (r *Repository) ListSessions(ctx context.Context, opt ...Option) ([]*Sessio
 	case 0:
 	case 1:
 		inClauseCnt += 1
-		where, args = append(where, fmt.Sprintf("scope_id = @%d", inClauseCnt)), append(args, sql.Named(fmt.Sprintf("%d", inClauseCnt), opts.withScopeIds[0]))
+		where, args = append(where, fmt.Sprintf("scope_id = @%d", inClauseCnt)), append(args, sql.Named(strconv.Itoa(inClauseCnt), opts.withScopeIds[0]))
```
johanbrandhorst, line 267: Hm, I made these changes based on my comment, but I see this pattern is pretty common in this file, so maybe I should revert it again. Thoughts @tmessi?

tmessi: Yea I have noticed this in a lot of the repositories. I think this should be changed, but probably as a separate commit that addresses it in all places, so for now I would revert this.

johanbrandhorst: https://github.com/hashicorp/boundary/pull/2047

tmessi (approved):

PR #2047, opened 2022-04-29
refact(all): Replace uses of fmt.Sprintf("%d", x)

johanbrandhorst:
This is just a less efficient and clear way of converting
an integer to a string than strconv.Itoa.


johanbrandhorst:
Nevermind this, it'll introduce type conversions in some places.

PR #2049, opened 2022-04-29
Listing Session and Targets Performance Improvements

tmessi:
Improve response time for `GET` requests to:

- `/v1/targets`
- `/v1/sessions`

These endpoints now perform authorization checks prior to retrieving the
full resources from the database.

This also includes a bug fix for retrieving sessions that could result
in incomplete results when there is a large number (10k+) of sessions.


johanbrandhorst (approved):

johanbrandhorst (approved):
:shipit:

Issue #2064, opened 2022-05-09
Foreign key constraint violation when targets.(Service).deleteFromRepo


jakubbujny:
**Describe the bug**
Boundary v0.8.0
`boundary targets delete -id XXX`
gives

```
Error from controller when performing delete on target

Error information:
  Kind:                Internal
  Message:             targets.(Service).deleteFromRepo: unable to delete target: target.(Repository).DeleteTarget: db.DoTx: target.(Repository).DeleteTarget: db.Delete: insert or update on table "session_state" violates foreign key constraint
  "session_valid_state_enm_fkey": integrity violation: error #1003
  Status:              500
  context:             Error from controller when performing delete on target
```

**To Reproduce**
I'm not sure how to reproduce it exactly - we did the upgrade from 0.7.0 to 0.8.0
It seems only some part of targets is affected (these created in 0.7.0) - when I create new "test" target I'm able to delete it.

jakubbujny:
I found now that we have the same problem for the host-sets

```
boundary host-sets delete -id hsst_q5cqOuwUja
Error from controller when performing delete on host set

Error information:
  Kind:                Internal
  Message:             host_sets.(Service).deleteFromRepo: unable to delete host: static.(Repository).DeleteSet: delete failed for hsst_q5cqOuwUja: db.DoTx: static.(Repository).DeleteSet: db.Delete: insert or update on table "session_state"
  violates foreign key constraint "session_valid_state_enm_fkey": integrity violation: error #1003
  Status:              500
  context:             Error from controller when performing delete on host set
  ```

irenarindos:
Thanks for reporting this! I've been able to recreate it & am working on a fix.

irenarindos:
Thanks again- this has been fixed in #2065 and will be in our next release.

PR #2065, opened 2022-05-09
bug(session): null fkey trigger also checks for session terminated state

irenarindos:
Function cancel_session() cancels sessions on deletion of foreign key constraints. It checked whether a session was in a canceling state but not terminated (sessions can transition to terminated without going through canceling).
This update adds a check for a terminated state to prevent attempts to cancel already terminated sessions.

Fixes https://github.com/hashicorp/boundary/issues/2064

### DIFF
--- a/internal/db/schema/migrations/oss/postgres/29/01_cancel_session_null_fkey.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/29/01_cancel_session_null_fkey.up.sql
@@ -1,0 +1,37 @@
+begin;
+
+drop function cancel_session(in sessionId text);
+-- Updates cancel_session() from 0/50_session to check if a session is either terminated or canceling
+-- Sessions can progress directly to terminated without going through the canceling state
+-- cancel_session will insert a cancel state for the session, if there's isn't
+-- a canceled or terminated state already.  It's used by cancel_session_with_null_fk.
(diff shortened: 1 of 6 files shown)


internal/db/schema/migrations/oss/postgres/29/01_cancel_session_null_fkey.up.sql
```diff
@@ -0,0 +1,37 @@
+begin;
+
+drop function cancel_session(in sessionId text);
+-- Updates cancel_session() from 0/50_session to check if a session is either terminated or canceling
+-- Sessions can progress directly to terminated without going through the canceling state
+-- cancel_session will insert a cancel state for the session, if there's isn't
+-- a canceled or terminated state already.  It's used by cancel_session_with_null_fk.
+create function
+  cancel_session(in sessionId text) returns void
```
mgaffney, line 9: Can we add unit tests for this?

mgaffney: Sorry, I should have been more clear. I was thinking the unit tests would be added to our sql test suite since this is a sql function. Would it be possible to add unit tests there?

mgaffney (dismissed):

PR #2068, opened 2022-05-10
feat(workers): Add storage path parameter to worker config

A-440Hz:


irenarindos (commented):
Looks great so far! 
I think we might want to check that the storage location set is valid & that we can write there. The Parse() method in config.go might be a good location for this. 

Then we can add testing around the storage location- in worker_test.go we have an existing test around listener config; we could extend that with the storage_path and test for valid/ invalid paths

A-440Hz:
thanks! 
I will be adding a test write for the storage location around [here](https://github.com/hashicorp/nodeenrollment/blob/9e82d50663c5f0807ea8d0dd20efceeda14af6ba/nodestorage/file/file_storage.go#L53), with probably os.MkdirTemp()

irenarindos (approved):

PR #2069, opened 2022-05-10
chore: Change how update-ui-version gets the latest commit sha from main

cameronperera:
[Jira Ticket](https://hashicorp.atlassian.net/browse/ICU-3937)

This does remove the message associated with the commit. Please let me know if that is an issue at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 
 ## Next
 
+### Deprecations/Changes
+
+* Update uiupdate.sh to fetch latest commit sha on main from the boundary-ui repo.
+* No longer use uiclone.sh in the update-ui-version step in the Makefile since we do not need to clone the repo to get the latest commit sha.
+
 ## 0.8.0 (2022/05/03)
 
 ### New and Improved
--- a/scripts/uiupdate.sh
+++ b/scripts/uiupdate.sh
@@ -5,18 +5,15 @@ set -e
 if [ -z "$UI_VERSION_FILE" ]; then
 	echo "Must set UI_VERSION_FILE"; exit 1
 fi
-if [ -z "$UI_CLONE_DIR" ]; then
-	echo "Must set UI_CLONE_DIR"; exit 1
-fi
 
 shafileabs="$(pwd)/${UI_VERSION_FILE}"
-cd "${UI_CLONE_DIR}"
-V="$(git log -n1 --pretty=oneline)"
-echo "==> Setting UI version to: $V"
+V="$(git ls-remote https://github.com/hashicorp/boundary-ui main)"
+SHA=${V:0:40}
+echo "==> Setting UI version to: $SHA"
 
 # Write the version file.
 {
-	echo "$V"
+	echo "$SHA"
 	echo "# This file determines the version of the UI to embed in the boundary binary."
 	echo "# Update this file by running 'make update-ui-version' from the root of this repo."
 	echo "# Set UI_COMMITISH when running the above target to update to a specific version."
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,7 @@ update-ui-version:
 	else \
 		echo "==> Setting to latest commit matching '$(UI_COMMITISH)'"; \
 	fi; \
-	./scripts/uiclone.sh && ./scripts/uiupdate.sh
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 
 ## Next
 
+### Deprecations/Changes
```
calcaide, line 7: Since this is mostly used by devs, I think this should not be mentioned in changelog, but happy to discuss.

cameronperera: Agreed but I wanted to include it just in case someone still wanted it include.

scripts/uiupdate.sh
```diff
@@ -5,18 +5,15 @@ set -e
...
-if [ -z "$UI_CLONE_DIR" ]; then
-	echo "Must set UI_CLONE_DIR"; exit 1
-fi
 
 shafileabs="$(pwd)/${UI_VERSION_FILE}"
-cd "${UI_CLONE_DIR}"
-V="$(git log -n1 --pretty=oneline)"
-echo "==> Setting UI version to: $V"
+V="$(git ls-remote https://github.com/hashicorp/boundary-ui main)"
+SHA=${V:0:40}
```
calcaide, line 11: To simplify this I would delete `SHA` transformation and leave `V` as `git ls-remote https://github.com/hashicorp/boundary-ui main | cut -f 1`. That will print only tge SHA without the reference

calcaide (changes requested):
Good work! left a comment

scripts/uiupdate.sh
tmessi: Personally I like the bash substring solution, instead of needing `cut` in a subshell. However if we use that we should update the shabang to explicitly use bash, ie:

```bash
#!/usr/bin/env bash
```

cameronperera: What @tmessi mentioned. I went with this less pretty solution because it doesn't need `cut`. I am fine with either direction though. Just let me know what you would prefer. I will remove the changelog edit as well.

calcaide: I do not have strong opinion at all, bash substring it is!!

calcaide (dismissed):

CHANGELOG.md
cameronperera: Removing changelog edits.

calcaide:
@cameronperera just to double check, do you mind to share the testing procedure for this? I guess you already tried to run the `make install` and `make build-ui` as testing?

Nice work! Appreciate it!

tmessi (changes requested):

scripts/uiupdate.sh
tmessi: Unresolving this since the shebang line still needs updating

scripts/uiupdate.sh
```diff
@@ -5,18 +5,15 @@ set -e
...
 fi
-if [ -z "$UI_CLONE_DIR" ]; then
-	echo "Must set UI_CLONE_DIR"; exit 1
-fi
 
 shafileabs="$(pwd)/${UI_VERSION_FILE}"
-cd "${UI_CLONE_DIR}"
-V="$(git log -n1 --pretty=oneline)"
-echo "==> Setting UI version to: $V"
+V="$(git ls-remote https://github.com/hashicorp/boundary-ui main)"
```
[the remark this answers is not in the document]
johanbrandhorst: If we remove UI_COMMITISH we also need to update the release process: https://hashicorp.atlassian.net/wiki/spaces/ICU/pages/1459880389/Binary+and+Docker+Release#UI-Version

scripts/uiupdate.sh
cameronperera: Fixed now.

scripts/uiupdate.sh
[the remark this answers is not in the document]
cameronperera: Sorry, I was only looking at this from the scope of `uiupdate.sh`, not who `update-ui-version` would really be using it. It now uses `$UI_COMMITSH` instead of just main. I added a check for when a commit is provided via `$UI_COMMITISH`. It will check the response from `git ls-remote` and set `$V` to `$UI_COMMITISH` if the test is true.

johanbrandhorst (approved):

tmessi (approved):

PR #2070, opened 2022-05-11
Remove the server_id from session table, change it to worker_id on session_connection

talanknight:
Subsequent PRs will continue to modify the server_worker table to add new fields. The worker is
becoming a public resource so it must have a public id.  We drop the server_id column on the
session_connection table and add the worker_id column so we can use the wt_public_key type.
Since db migrations are applied while the controller is offline, it should be safe to assume all
session connections are not in a running state, thus allowing us to null out all the worker ids in
the session_connection table.

For now we don't have very good enforcement of the wt_public_id constraints, since the id is currently
defined in the worker config, but this will soon change and the worker's id will be generated by the
controller.

### DIFF
--- a/internal/proto/controller/api/resources/sessions/v1/session.proto
+++ b/internal/proto/controller/api/resources/sessions/v1/session.proto
@@ -44,9 +44,11 @@ message Connection {
 
 // Session contains all fields related to a Session resource
 message Session {
-  // Worker info previously contained only the worker's address.  A session
-  // could have multiple connections through different workers, so if we need
-  // to report worker information, it should probably be done on the connection.
+  // Worker info previously contained only the worker's address but was never
+  // populated either in the database or on the API.  While worker id information
+  // may become helpful on a session in the future, the worker id seems the most
+  // relevant information since the address can change between the time the
+  // session is created and a session connection is made through that worker.
   reserved "worker_info";
 
   // Output only. The ID of the Session.


internal/daemon/worker/worker.go
```diff
@@ -194,6 +194,9 @@ func (w *Worker) Start() error {
 	if err := w.startControllerConnections(); err != nil {
 		return fmt.Errorf("error making controller connections: %w", err)
 	}
+	if len(w.conf.RawConfig.Worker.Name) < 10 {
```
irenarindos, line 197: What does this do?

internal/db/schema/migrations/oss/postgres/99/01_worker_controller_tables.up.sql
```diff
@@ -69,37 +69,44 @@ create table server_worker_tag (
...
-  after update of version, termination_reason, key_id, tofu_token, worker_id on session
+  after update of version, termination_reason, key_id, tofu_token on session
   for each row execute procedure update_version_column();
 
+
 -- Update session_connection table to use worker_id instead of server_id
-alter table session_connection
-  drop constraint server_fkey;
+-- Table last updated in 21/02_session.up.sql
+-- TODO: Add the denormalized worker address column since this value doesn't change
```
irenarindos, line 90: Will this be addressed in a future PR or should it go into this one?

mgaffney (changes requested):

internal/db/schema/migrations/oss/postgres/99/01_worker_controller_tables.up.sql
```diff
@@ -71,37 +71,32 @@ drop table server_tag;
...
-
-alter table session
-  rename column server_id to worker_id;
 alter table session
-  add constraint server_worker_fkey
-    foreign key (worker_id)
-      references server_worker(private_id)
-      on delete set null
-      on update cascade;
+  drop column server_id;
```
mgaffney, line 83: Any reason not to make the three changes to the `session` table one statement?

```sql
drop trigger update_version_column
  on session;

alter table session
  drop constraint session_server_id_fkey,
  drop column server_type,
  drop column server_id;
```

internal/db/schema/migrations/oss/postgres/99/01_worker_controller_tables.up.sql
```diff
@@ -71,37 +71,32 @@ drop table server_tag;
...
 alter table session_connection
-  rename column server_id to worker_id;
+  add column worker_id wt_public_id;
 alter table session_connection
   add constraint server_worker_fkey
     foreign key (worker_id)
-      references server_worker (private_id)
+      references server_worker (public_id)
       on delete set null
       on update cascade;
```
mgaffney, line 101: Same question for the `session_connection` table and the `job_run` table:

```sql
alter table session_connection
  drop column server_id,
  add column worker_id wt_public_id
    constraint server_worker_fkey
      references server_worker (public_id)
      on delete set null
      on update cascade;

-- Update job run table so that server id references controller id
alter table job_run
  drop constraint server_fkey,
  add constraint server_controller_fkey
    foreign key (server_id)
      references server_controller (private_id)
      on delete set null
      on update cascade;
```

mgaffney (approved):
I only reviewed the sql bits

internal/proto/controller/api/resources/sessions/v1/session.proto
```diff
@@ -49,6 +44,11 @@ message Connection {
 
 // Session contains all fields related to a Session resource
 message Session {
+  // Worker info previously contained only the worker's address.  A session
```
jefferai, line 47: This is not quite correct. Note that worker_info is repeated. This is a list of workers that were chosen as allowed proxies at session authorization time. Any connections would need to be made through one of these workers.

IOW, this is the set of workers that matched any worker filter on the target and any load balancing or round robin we might do in the future.

You're right in that _for a particular connection_ you'd have to ask the connection which worker was used. But I'm not convinced that removing this is the right thing to do as it's still potentially valuable information.

Another option might be to change it to something like a list of worker IDs rather than expose the address here directly.

internal/session/repository.go
```diff
@@ -110,7 +110,6 @@ func (r *Repository) convertToSessions(ctx context.Context, sessionList []*sessi
 				PublicId:          sv.PublicId,
 				UserId:            sv.UserId,
 				HostId:            sv.HostId,
-				WorkerId:          sv.WorkerId,
```
jefferai, line 113: See comment on the proto, this is probably not the right thing to do.

internal/session/repository_session.go
```diff
@@ -289,11 +286,6 @@ func (r *Repository) ListSessions(ctx context.Context, opt ...Option) ([]*Sessio
 		where = append(where, fmt.Sprintf("s.public_id in (%s)", strings.Join(idsInClause, ",")))
 	}
 
-	if opts.withWorkerId != "" {
```
jefferai, line 292: Similar to my comments above, this is all sort of predicated on the idea that we will never want to query to find out which workers are _allowed_ for a given session. I'm not sure that's the case. It seems like information we might want to push into the warehouse with the other session information: the session contains the allowed set of workers, then each connection indicates which worker it actually used for that connection.

jefferai (changes requested):
See the comments I left -- I think some of this needs to be reworked and some of the removed code restored...the issue isn't really that session has a worker ID, it's that the worker ID attached to a session is singular.

jefferai (approved):

Issue #2072, opened 2022-05-11
[bug] Listener on the controller's "cluster" endpoint crashes if it is sent an unexpected request


justenwalker:
**Describe the bug**

Listener on the controller's "cluster" endpoint crashes if it is sent an unexpected request

**To Reproduce**

1. run `boundary dev` to set up a local boundary instance
2. Validate that boundary is listening on all the ports
   ```
    $ netstat -an -p tcp | grep LISTEN | grep 920
    tcp4       0      0  127.0.0.1.9202         *.*                    LISTEN
    tcp4       0      0  127.0.0.1.9203         *.*                    LISTEN
    tcp4       0      0  127.0.0.1.9201         *.*                    LISTEN
    tcp4       0      0  127.0.0.1.9200         *.*                    LISTEN
    ```
3. run `curl http://localhost:9201` to send it an unexpected request
4. Validate that boundary is no longer listening on 9201 (the "cluster" endpoint)
   ```
    $ netstat -an -p tcp | grep LISTEN | grep 920
    tcp4       0      0  127.0.0.1.9202         *.*                    LISTEN
    tcp4       0      0  127.0.0.1.9203         *.*                    LISTEN
    tcp4       0      0  127.0.0.1.9200         *.*                    LISTEN
    ```

**Expected behavior**

boundary should gracefully reject malformed packets on 9201 and continue to accept connections

**Additional context**

This seems to happen after the v0.8.0 upgrade

```
$ boundary version

Version information:
  Git Revision:        bc86b4e9e9adfdf07183cbf140a2802fb453b35f
  Version Number:      0.8.0
```

It is not reproduceable on `v0.7.6`

Also, nothing seems to get logged when this happens. I did not see any new output from the logstream after sending the curl.

AdamBouhmad:
Thanks for reporting @justenwalker! We're looking into this currently.

hsimon-hashicorp:
Thank you for the report, @justenwalker! This is fixed in #2073 and will be included in the next release. We really appreciate your contribution. Please feel free to open a new issue for any other potential regressions, bugs, oddities, or enhancement requests!

PR #2073, opened 2022-05-11
Leave cluster listener up when bad data is sent

jefferai:
Removing the ALPN muxer exposed the underlying listener directly to the
gRPC server, which means that a quirk of the ALPN muxer (not surfacing
errors on accept) turned into the gRPC server receiving errors on
accept, and it turns out if the error does not satisfy a special
temporary interface the listener will be closed.

The obviously better thing to do would be to let the listener close
itself if there was truly a fatal error rather than relying on
undocumented interfaces, but here we are.

### DIFF
--- a/internal/daemon/controller/intercepting_listener.go
+++ b/internal/daemon/controller/intercepting_listener.go
@@ -17,8 +17,8 @@ type tempError struct {
 	error
 }
 
-// NewTempError is a "temporary" error
-func NewTempError(inner error) tempError {
+// newTempError is a "temporary" error
+func newTempError(inner error) tempError {
 	return tempError{error: inner}
 }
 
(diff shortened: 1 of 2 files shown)


hsimon-hashicorp:
Fixes #2072

internal/daemon/controller/intercepting_listener.go
```diff
@@ -9,6 +9,23 @@ import (
...
+// tempError is an error that satisfies the temporary error interface that is
+// internally used by gRPC to determine whether an error should cause a listener
+// to die. Any error that isn't an accept error is wrapped in this since one
+// connection failing TLS wise doesn't mean we don't want to accept any more...
+type tempError struct {
+	error
+}
+
+// NewTempError is a "temporary" error
+func NewTempError(inner error) tempError {
```
johanbrandhorst, line 21: Could we unexport this?

jefferai: Ah, sure.

johanbrandhorst (approved):

tmessi (approved):

PR #2074, opened 2022-05-11
Update UI artifacts consumption

calcaide:
Update the UI artifacts consumption within the Boundary CI using a make script to get Boundary Edition.

Updated May 13.

To be aware:
- This PR updates the UI version. **Why?** In the UI we start generating both artifacts (`admin-ui-oss` and `admin-ui-enterprise`) hours before this PR. So there are no artifacts with that naming available for previous SHA's.

Screenshot of the test branch using the make script and the `get-product-edition` job and retrieving the proper UI artifact succesfully:
Screenshot from 2022-05-13 10-46-31

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,6 @@ on:
 
 env:
   PKG_NAME: "boundary"
-  EDITION: oss
 
 jobs:
   get-product-version:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -9,6 +9,7 @@ on:
 
 env:
   PKG_NAME: "boundary"
+  EDITION: oss
```
tmessi, line 12: I think to avoid a conflict on this file, we should derive this variable from something already in the repo. Lets chat to figure out the best way to do that.

randallmorey: This configuration must not live in code.  We may want to use GitHub environment secrets for this value.

tmessi: It depends on what you consider "in code". Using GitHub environment secrets might be a good solution for now. But if we eventually want to support using these pre-built ui assets locally, we will need to solve this issue again, and it would be nice if locally it would just "do the right thing" based on something in the repository.

randallmorey: We can have a default edition, but it must also be configurable outside of the code in order to support arbitrary editions.

calcaide: I will start a thread in Slack.

calcaide: As decide on slack thread:
- We will retrieve the edition using Go.
- We will add a make script `make edition` to run the Go code mentioned above.
- We will update `build.yaml` to use the make script ratther than `ENV` variable.

johanbrandhorst (approved):
LGTM

tmessi (approved):

PR #2076, opened 2022-05-12
feat(worker): implements WorkerStorage wrapper for worker credential encryption

A-440Hz:
also adds "worker-storage-dir" flag to boundary dev to be able to modify the `storage_path` var in worker config in lieu of using an .hcl file


A-440Hz:
adds a skipStorageCleanup bool to the worker struct so that users can toggle whether a worker's credential storage directory is preserved or deleted; with a flag option in `boundary dev`, or within the hcl config file

internal/daemon/worker/worker.go
```diff
@@ -255,6 +256,7 @@ func (w *Worker) Shutdown() error {
 	w.baseCancel()
 	w.Resolver().UpdateState(resolver.State{Addresses: []resolver.Address{}})
 
+	w.NodeeFileStorage.Cleanup()
```
irenarindos, line 259: Do we want to clear out worker storage on shutdown? If a user is taking a worker down for maintenance, I don't think we'd want a worker's creds cleared out

A-440Hz: I took this away for now. Perhaps we want to look for a specific worker tag before calling this function

internal/cmd/config/config.go
```diff
@@ -244,42 +252,47 @@ func DevKeyGeneration() (string, string, string) {
...
 	if n != numBytes {
-		panic(fmt.Errorf("expected to read 64 bytes, read %d", n))
+		panic(fmt.Errorf("expected to read 32 bytes, read %d", n))
 	}
-	controllerKey := base64.StdEncoding.EncodeToString(randBuf.Bytes()[0:32])
-	workerAuthKey := base64.StdEncoding.EncodeToString(randBuf.Bytes()[32:64])
-	recoveryKey := base64.StdEncoding.EncodeToString(randBuf.Bytes()[64:numBytes])
-
-	return controllerKey, workerAuthKey, recoveryKey
+	devKey := base64.StdEncoding.EncodeToString(randBuf.Bytes()[0:numBytes])
```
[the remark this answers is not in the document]
A-440Hz: I am unsure how this will return a consistent buffer size of 32 bytes when I make this change

go.sum
```diff
@@ -741,8 +741,6 @@ github.com/hashicorp/golang-lru v0.5.4 h1:YDjusn29QI/Das2iO9M0BHnIbxPeyuCHsjMW+l
 github.com/hashicorp/golang-lru v0.5.4/go.mod h1:iADmTwqILo4mZ8BN3D2Q6+9jd8WM5uGBxy+E8yxSoD4=
 github.com/hashicorp/hcl v1.0.0 h1:0Anlzjpi4vEasTeNFn2mLJgTSwt0+6sfsiTG8qcWGx4=
 github.com/hashicorp/hcl v1.0.0/go.mod h1:E5yfLk+7swimpb2L/Alb/PJmXilQ/rhwaUYs4T20WEQ=
-github.com/hashicorp/nodeenrollment v0.0.0-20220520141833-ad7323153acc h1:7RE6UxjVxa3uUS0A4RZwVRFnBJqNQnm5nzNwqPJy4VM=
```
irenarindos, line 744: Can you un-check in changes to this file?

internal/kms/kms.go
```diff
@@ -103,6 +103,11 @@ func (k *Kms) AddExternalWrappers(ctx context.Context, opt ...Option) error {
 			return errors.Wrap(ctx, err, op, errors.WithMsg("unable to add worker auth wrapper"))
 		}
 	}
+	if opts.withWorkerStorageWrapper != nil {
```
irenarindos, line 106: WorkerStorage also needs to be added to the `GetExternalWrappers` method in this file

Can you also add the WorkerStorage wrapper to these files:
`globals/kms.go` 
`internal/kms/external_wrappers.go`
`internal/kms/external_wrappers_test.go`

internal/daemon/controller/controller.go
```diff
@@ -218,6 +218,7 @@ func New(ctx context.Context, conf *Config) (*Controller, error) {
 		ctx,
 		kms.WithRootWrapper(c.conf.RootKms),
 		kms.WithWorkerAuthWrapper(c.conf.WorkerAuthKms),
+		kms.WithWorkerStorageWrapper(c.conf.WorkerStorage),
```
irenarindos, line 221: WorkerStorage is only set on the worker config/ used by the worker, so the controller doesn't need it
```suggestion
	
```

internal/cmd/base/servers.go
```diff
@@ -84,6 +84,7 @@ type Server struct {
```
irenarindos, line 84: In `internal/cmd/base/servers.go` can you add the new kms to `SetupKMSes`
And also add to the test `TestServer_SetupKMSes` in  `internal/cmd/base/server_test.go`

irenarindos (changes requested):
Looking good! Just a few changes... I'm also wondering if you can add a test that uses the wrapper and NodeFileStorage to ensure that the wrapper is used and the values written are encrypted.
Thanks!

go.sum
A-440Hz: could I `git checkout --patch llb-byow go.sum` or something, or would I have to redo all my changes in a new branch?

internal/daemon/worker/testing.go
```diff
@@ -173,6 +173,9 @@ type TestWorkerOpts struct {
 	// The worker auth KMS to use, or one will be created
 	WorkerAuthKms wrapping.Wrapper
 
+	// The worker credential storage KMS to use, or one will be created
+	WorkerStorage wrapping.Wrapper
```
jimlambrt, line 177: I don't think you've incorporated this into `NewTestWorker(...)` so it's here but I'm not sure it has any affect.  Based on this, perhaps we should add a unit test that makes sure it does have an affect on a TestWorker?

internal/daemon/worker/worker.go
```diff
@@ -186,7 +186,7 @@ func (w *Worker) Start() error {
 		return err
 	}
 
-	nodeCreds, err := types.NewNodeCredentials(w.baseContext, w.NodeeFileStorage)
+	nodeCreds, err := types.NewNodeCredentials(w.baseContext, w.NodeeFileStorage, nodee.WithWrapper(w.conf.WorkerStorageKms))
```
jimlambrt, line 189: should this be moved up to `New(conf *Config)`?  I'm curious is there any reason to wait until startup to setup these creds?  Also, if we move it up to `New(...)` we could easily write a unit test for this using a "real" worker vs a TestWorker.

internal/cmd/base/server_test.go
```diff
@@ -900,3 +907,24 @@ func TestSetupWorkerPublicAddress(t *testing.T) {
...
 	}
 }
+
+func TestSetupWorkerCredentialStorage(t *testing.T) {
+	ctx := context.Background()
+	assert, require := assert.New(t), require.New(t)
+	ts := db.TestWrapper(t)
+	keyId, err := ts.KeyId(ctx)
+
+	tw := worker.NewTestWorker(t, &worker.TestWorkerOpts{
```
jimlambrt, line 917: I'm not sure this is a valid test... you're testing a `TestWorker` which is completely different than testing a `Worker`.  Perhaps we need to test both?

internal/cmd/base/server_test.go
```diff
@@ -900,3 +907,24 @@ func TestSetupWorkerPublicAddress(t *testing.T) {
...
+		WorkerStorage:    ts,
+		DisableAutoStart: true,
+	})
+
+	err = tw.Worker().Start()
+	require.NoError(err)
+	wKeyId, err := tw.Config().WorkerStorageKms.KeyId(ctx)
```
jimlambrt, line 924: this current panics.

internal/cmd/base/server_test.go
```diff
@@ -900,3 +907,24 @@ func TestSetupWorkerPublicAddress(t *testing.T) {
 		})
 	}
 }
+
+func TestSetupWorkerCredentialStorage(t *testing.T) {
+	ctx := context.Background()
```
jimlambrt, line 912: TestSetupWorkerCredentialStorage(...) into internal/daemon/worker/worker_test.go you’ll eliminate the circular dep

internal/daemon/worker/controller_connection.go
```diff
@@ -78,6 +78,7 @@ func (w *Worker) controllerDialerFunc() func(context.Context, string) (net.Conn,
 				nodeauth.MakeCurrentParametersFactory(
 					ctx,
 					nodee.NopTransactionStorage(w.NodeeFileStorage),
+					nodee.WithWrapper(w.conf.WorkerStorageKms),
```
jimlambrt, line 81: what happens if the storage wrapper is nil (is it an error we can detect before returning)?  Also, is it possible to write a unit test for this?

internal/cmd/config/config.go
```diff
@@ -244,42 +252,47 @@ func DevKeyGeneration() (string, string, string) {
...
 	parsed.DevWorkerAuthKey = workerAuthKey
+	parsed.DevWorkerStorageKey = workerStorageKey
 	parsed.DevRecoveryKey = recoveryKey
 	return parsed, nil
 }
 
 func DevCombined() (*Config, error) {
-	controllerKey, workerAuthKey, recoveryKey := DevKeyGeneration()
-	hclStr := fmt.Sprintf(devConfig+devControllerExtraConfig+devWorkerExtraConfig, controllerKey, workerAuthKey, recoveryKey)
+	controllerKey := DevKeyGeneration()
```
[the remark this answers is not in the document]
A-440Hz: There's no input for this function so is it useful if I only test one case? (That the function returns the correct parsed hcl input with the keys passed in)

internal/daemon/worker/worker.go
A-440Hz: This is a good question. I'm guessing it's because the nodee functions need a context and there already is one in `Start()` but none in `New()`. Let's ask @jefferai what he thinks about moving them, since I believe he put them there in the llb-byow base branch

internal/cmd/base/server_test.go
A-440Hz: Doesn't a `TestWorker` create a new Worker with some default settings using the `New()` function from `worker.go`?

internal/cmd/config/config.go
```diff
@@ -244,42 +252,47 @@ func DevKeyGeneration() (string, string, string) {
...
-	return controllerKey, workerAuthKey, recoveryKey
+	devKey := base64.StdEncoding.EncodeToString(randBuf.Bytes())
+	return devKey
 }
 
 // DevController is a Config that is used for dev mode of Boundary
 // controllers
 func DevController() (*Config, error) {
-	controllerKey, workerAuthKey, recoveryKey := DevKeyGeneration()
+	controllerKey := DevKeyGeneration()
```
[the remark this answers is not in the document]
A-440Hz: What specifically would I be testing for? That the config created is the same except for the generated keys?

internal/cmd/commands/dev/dev.go
```diff
@@ -336,6 +338,18 @@ func (c *Command) Flags() *base.FlagSets {
 		Usage:  "If set, the original \"ephemeral\" method of worker auth will be used to connect the initial dev worker to the controller.",
 	})
 
+	f.StringVar(&base.StringVar{
```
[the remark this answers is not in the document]
A-440Hz: I don't think I've found any unit tests for other dev flags. Is there a specific place where I would write this test?

internal/daemon/worker/controller_connection.go
A-440Hz: Do we want to ensure this is not nil? In the [nodee library](https://github.com/hashicorp/nodeenrollment/blob/2f27f4add8d11675c784a586e26347dd6f2a71af/types/node_credentials.go#L62) it just stores without encryption if this wrapper is nil

A-440Hz: I guess do we want to throw an error any time a worker storage KMS wrapper is not passed in?

internal/daemon/worker/worker.go
jimlambrt: not sure there's anything magical about the `context.WithCancel()` that's created 20 lines above in this same func.  I'm pretty sure there's no way to call `cancel` on that context before this func returns.  As I stated before, moving it to `New(...)` has the extra added bonus of being able to easily create a unit test for it... which in my mind is quite compeling if there's no concrete reason to keep it here.

internal/cmd/config/config.go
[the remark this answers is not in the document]
jimlambrt: You've added an expected outcome which I believe other parts of the implementation rely on.  It might be nice to ensure it's correct.  you might even consider checking that the keys meet some set of constraints, since there are no unit tests for `DevKeyGeneration(...)`.  It's like getting two birds with one stone.

internal/cmd/config/config.go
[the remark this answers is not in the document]
jimlambrt: Sort of the same response I had to the `DevController()`.  Whether or not a func has parameters wouldn't seem to affect whether or not you test your code's correctness.  There's an expected outcome and/or behavior which should be tested if possible unless there's some big lift/change required to make it testable.

internal/cmd/base/server_test.go
jimlambrt: Ah, ty!  I didn't read the func far enough down to see that call.

internal/daemon/worker/testing.go
jimlambrt: I don't see a test for `TestWorker(...)`.  Did I miss something?

internal/daemon/worker/controller_connection.go
jimlambrt: while I understand why the nodee library wants to support not encrypting creds, does boundary want to support that workflow?  sort of a broader question you might consider asking the team.

jimlambrt (changes requested):

internal/daemon/worker/controller_connection.go
jefferai: I think we need to, yes. Not everyone will have access to a suitable KMS, or may have other possibilities (an encrypted USB drive that's inserted and unlocked as long as it has power; a ramdisk; ephemeral KMS workflow; dev/test environments; etc.

internal/cmd/config/config.go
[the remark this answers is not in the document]
A-440Hz: Sorry, I should have asked in the first place whether this unit test is sufficiently covered by the `TestDevController` func in `config_test.go`

A-440Hz:
I have yet to write a test for DevCombined

internal/cmd/commands/dev/dev.go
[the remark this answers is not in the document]
jimlambrt: You can add new types of tests which have been defined before.  Perhaps the original authors were pressed for time or weren't sure how to write a test for these flags.

Something as simple as:
```
unc TestCommand_Flags(t *testing.T) {
	t.Parallel()
	assert := assert.New(t)

	serverCmdUi := &base.BoundaryUI{
		Ui: &cli.ColoredUi{
			ErrorColor: cli.UiColorRed,
			WarnColor:  cli.UiColorYellow,
			Ui: &cli.BasicUi{
				Reader: bufio.NewReader(os.Stdin),
				Writer: os.Stdout,
			},
		},
		Format: "table",
	}
	base.NewServer(base.NewCommand(serverCmdUi))

	cmd := &Command{Server: base.NewServer(base.NewCommand(serverCmdUi))}
	got := cmd.Flags()
	completions := got.Completions()
	
	assert.Contains(completions, "-worker-storage-dir")

       // keep adding assertions for other flags which should be set as a result of cmd.Flags()
}
```

internal/cmd/config/config.go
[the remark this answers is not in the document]
jimlambrt: yes it is.

internal/cmd/config/config.go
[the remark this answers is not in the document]
jimlambrt: this still has no test coverage.

jimlambrt (changes requested):
Just a few more minor bits.  Looks great overall!

internal/cmd/commands/dev/dev.go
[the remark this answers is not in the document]
jefferai: > were pressed for time

Boy howdy

internal/cmd/commands/dev/dev.go
```diff
@@ -335,6 +337,18 @@ func (c *Command) Flags() *base.FlagSets {
...
+		Name:   "worker-storage-dir",
```
jefferai, line 341: There's an inconsistency between the flag name -- which doesn't have "auth" in it -- and the variable names/usage strings, which indicate it's for auth storage. I'm of two minds about the right way to proceed. On the one hand we could define a single top level storage dir using these flags and have auth and perhaps other future purposes be under this one directory. On the other hand for future purposes we may want to allow defining storage in different locations.

My inclination is to say that we should change the flag name to `worker-auth-storage-dir`/`worker-auth-storage-skip-cleanup`.

internal/cmd/config/config.go
```diff
@@ -57,6 +57,13 @@ kms "aead" {
 	key_id = "global_worker-auth"
 }
 
+kms "aead" {
+    purpose = "worker-storage"
```
jefferai, line 61: Similar question here. Would we define different keys for worker auth storage and other future storage needs? Possibly it should be a purpose of `worker-auth-storage`.

internal/kms/kms.go
```diff
@@ -103,6 +103,11 @@ func (k *Kms) AddExternalWrappers(ctx context.Context, opt ...Option) error {
...
+		if err := k.underlying.AddExternalWrapper(ctx, wrappingKms.KeyPurpose(KeyPurposeWorkerStorage.String()), opts.withWorkerStorageWrapper); err != nil {
```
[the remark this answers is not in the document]
jimlambrt: Good catch. I wasn't considering the usage clearly. If workers are going to have additional external wrappers then perhaps we need a WorkerKms in the future, but right now we can nerf it out of the controller kms.

internal/cmd/config/config.go
A-440Hz: So to confirm, we would like to change this WorkerStorageKms wrapper purpose and all its related value names to WorkerAuthStorageKms-flavored things, like I've done with the storage path names in dev.go?

PR #2077, opened 2022-05-13
Cherry picking for 0.8.1 Release

irenarindos:


ci/goinstall.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
-VERSION="1.17.9"
+VERSION="1.17.10"
```
tmessi, line 4: you also need to update the build workflow files to use the correct version. For example [here](https://github.com/hashicorp/boundary/blob/3991a9fdb1d6c298383c0e81fb228b6c7c81693c/.github/workflows/build.yml#L23) and there are others in that file, just search for `go-version` or do a search and replace for `1.17.9` to `1.17.10`.

irenarindos: Sorry about that, it's in my fixup commit- missed this file the first time

.github/workflows/build.yml
```diff
@@ -5,7 +5,7 @@ on:
     # Sequence of patterns matched against refs/heads
     branches:    
       # Push events on main branch
-      - 'main'
+      - 'release/0.8.x'
```
tmessi, line 8: I would just add the release branch in addition to main, instead of replacing main. I suppose it doesn't matter...I am not really sure.

tmessi (approved):
I retested the scenario that [8c71343](https://github.com/hashicorp/boundary/pull/2077/commits/8c71343afbb97db14fa06b210f59e1789da08748) resolves, and it still looks good.

The go version update and build workflow update also look good. I did not test the session fix since I was not familiar with it, but otherwise this looks good to me.

jefferai (approved):

PR #2079, opened 2022-05-16
feat(db): add static credential store schema

mgaffney:


internal/db/schema/migrations/oss/postgres/50/01_static_credential.up.sql
```diff
@@ -0,0 +1,113 @@
...
+    username text not null
+      constraint username_must_not_be_empty
+        check(length(trim(username)) > 0),
+    password_encrypted bytea not null
+      constraint password_encrypted_must_not_be_empty
+        check(length(password_encrypted) > 0),
+    password_hmac bytea not null
+      constraint password_hmac_must_not_be_empty
+        check(length(password_hmac) > 0),
+    dek_id text not null
```
louisruch, line 67: Any reason going to `dek_id` over `key_id` as we do for other tables? I think we still need to have the struct in go be `KeyID` for the decrypt. Maybe not.  Either way this change might be good since it clearly states what specific key this this, just curious on thoughts?

louisruch (approved):
LGTM, one comment which is more a curiosity than anything else

internal/db/schema/migrations/oss/postgres/50/01_static_credential.up.sql
```diff
@@ -0,0 +1,113 @@
...
+    -- identifiers in PostgreSQL is 63 characters.
+    -- credential_static_username_password_credential_store_id_public_id_uq
+    -- is 68 characters.
+    --
+    -- https://www.postgresql.org/docs/current/limits.html
+    constraint credential_static_username_password_store_id_public_id_uq
+      unique(store_id, public_id)
+  );
+  comment on table credential_static_username_password_credential is
+    'credential_static_username_password_credential is a table where each row is a resource that represents a static username password credential. '
```
jimlambrt, line 92: extra space at the end?

internal/db/schema/migrations/oss/postgres/50/01_static_credential.up.sql
jimlambrt: I'm sort of with Louis on this... can we be consistent and use key_id?

jimlambrt (approved):
minor nits... but I would carefully consider changing dek_id to key_id.

internal/db/schema/migrations/oss/postgres/50/01_static_credential.up.sql
louisruch: The lines get concat so the extra space is needed expect on the last line

louisruch:
Mike is PTO for the rest of the week, I would prefer to work off the long lived branch. I am going to merge this and make changes on the `llb` to use `key_id`. I will sync up with Mike when he gets back

PR #2081, opened 2022-05-17
fix(vault_store): Correctly return expected errors

louisruch:


tmessi (commented):
Can you add tests to cover this?

louisruch:
@talanknight I fixed up the vault_token line as well now, I did not update the tests as this has a dependency on the KeyId returning an error which would be tricky to setup.
@tmessi Weirdly enough I see now it was being tested, but since it was failing on the next line due to the missing cert-key it was not testing that specific return path. I updated the test to validated the error code

tmessi (approved):

PR #2083, opened 2022-05-17
Add fields to server_worker to be a boundary resource

talanknight:
This PR adds scope_id, version, address, name and makes the uniqueness constraint be per scope id.

This also sets all running jobs to "interrupted" and sets the server_id to null in the jobs_run table.  This gives us more flexibility in setting these tables up correctly since we don't need to worry about maintaining existing data or migrating it into a new format.  The tables will get populated automatically as new controllers and workers are spun up.

Finally I also updated the WorkerTag storage proto to correctly reflect the columns in the worker tag table.

### DIFF
--- a/internal/db/schema/migrations/oss/postgres/99/01_domain_types.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/99/01_domain_types.up.sql
@@ -8,7 +8,7 @@ create domain wt_network_address as text
     check (length(trim(value)) > 0)
   constraint wt_network_address_too_long
     check (length(trim(value)) < 256);
-comment on domain wt_dns_name is
+comment on domain wt_network_address is
   'standard column for a network address.';
 
 


internal/db/schema/migrations/oss/postgres/99/02_worker_controller_tables.up.sql
```diff
@@ -68,6 +50,28 @@ create table server_worker_tag (
 -- Aaand drop server_tag
 drop table server_tag;
 
+create table server_controller (
```
tmessi, line 53: is there a reason `server_controller` was moved below `server_worker`? it makes this diff harder to review

internal/db/schema/migrations/oss/postgres/99/02_worker_controller_tables.up.sql
```diff
@@ -97,14 +101,30 @@ alter table session_connection
 
 -- Update job run table so that server id references controller id
 alter table job_run
-  drop constraint server_fkey,
+  add column controller_id wt_private_id,
```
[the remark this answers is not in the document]
tmessi: I think that all makes sense, it might be a good idea to include a comment in this migration to explain this rational.

internal/servers/public_ids.go
```diff
@@ -0,0 +1,19 @@
...
+
+// PublicId prefixes for the resources in the servers package.
+const (
+	WorkerPrefix = "w"
+)
+
+func newWorkerId() (string, error) {
+	id, err := db.NewPublicId(WorkerPrefix)
+	if err != nil {
+		return "", errors.WrapDeprecated(err, "vault.newCredentialStoreId")
```
irenarindos, line 16:
```suggestion
		return "", errors.WrapDeprecated(err, "vault.newWorkerId")
```

PR #2093, opened 2022-05-20
feat(workerAuth): Implement nodeenrollment lib workerauth storage

irenarindos:
Implements repository storage around nodeenrollment library storage: https://github.com/hashicorp/nodeenrollment/blob/dev/storage.go


internal/servers/workerauth/repository.go
```diff
@@ -0,0 +1,664 @@
...
+	if err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+	nodeAuth.ControllerEncryptionPrivKey, err = encrypt(ctx, node.ServerEncryptionPrivateKeyBytes, databaseWrapper)
+	if err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+
+	// workerId is passed in the context of the AuthorizeNode call
+	workerId := ctx.Value("workerId")
```
irenarindos, line 135: Is there a better way to pass workerId to storage?

internal/servers/repository_workerauth.go
```diff
@@ -0,0 +1,637 @@
...
+		reader: r,
+		writer: w,
+		kms:    kms,
+		worker: wrk,
+	}, nil
+}
+
+// Flush is called when storage is done being performed. The boolean parameter
+// indicates whether the operation was successful (true) or failed (false).
+func (r *WorkerAuthRepositoryStorage) Flush(b bool) error {
```
jefferai, line 61: This function needs to gate the transaction for the storage operation(s). Is this going to be built out in a future PR?

internal/db/read_writer.go
```diff
@@ -388,6 +397,47 @@ func (rw *Db) DeleteItems(ctx context.Context, deleteItems []interface{}, opt ..
 	return rowsDeleted, nil
 }
 
+// BeginTx will start a transaction and return a RW for use in db operations
+func (w *Db) BeginTx(ctx context.Context) (*Db, *dbw.RW, error) {
```
jefferai, line 401: I'm guessing these functions were pulled out from the logic in DoTx...would we want the backoff/retry/ticket already redeemed logic? Not saying we do, not sure we don't :-D

internal/servers/repository_workerauth.go
```diff
@@ -0,0 +1,729 @@
...
+	}
+
+	nodeAuth.WorkerId = r.worker.PublicId
+
+	err = nodeAuth.ValidateNewWorkerAuth(ctx)
+	if err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+
+	onConflict := &db.OnConflict{
```
jefferai, line 193: Thinking through concurrent operations...

NodeInformation can be written either through the API or, based on a discussion today, possibly at registration request time (if we move back away from caching). If it's written via the API, e.g. through AuthorizeNode, then if a request comes before that to fetch credentials it's fine, and if it comes after that it's fine. If a node reaches out to multiple controllers though and we store a registration request we want to ensure we only generate one ID, so we will need to ensure that an upsert takes into account whether we have a worker with an ID for that key ID already. (I'm not entirely sure what the conflict is doing right below this.)

On the root certificate side if we had multiple controllers seeing no certificates and trying to create an initial set, we'd want to make sure only one is successful. That's probably pretty simple in a transaction, just checking if current and next already exist.

I think NodeInformation is a bit tricker though, we may need a Version or as a constraint in the transaction. So that if we read a NodeInformation object in a transaction, and we are trying to update it, we ensure that we are only updating it if the Version hasn't changed.

internal/servers/repository_workerauth.go
```diff
@@ -0,0 +1,729 @@
...
+type WorkerAuthTransaction struct {
+	transactionWriter *db.Db
+	transaction       *dbw.RW
+}
+
+// NewRepositoryStoragecreates a new Worker Auth WorkerAuthRepositoryStorage that implements the Storage interface
+// If used to authenticate and store a workerAuth record, a worker must be passed
+// If transactional is set to true, all operations conducted in the lifetime of this storage instance are executed
+// within the context of a single transaction; Flush must be called to either commit or rollback this transaction
+func NewRepositoryStorage(ctx context.Context, r db.Reader, w db.Writer, kms *kms.Kms, wrk *store.Worker, transactional bool) (*WorkerAuthRepositoryStorage, error) {
```
jefferai, line 47: I think this will probably work -- I might need to modify the transaction interface so that the key id is passed into the function, I'm not sure how we'd get the worker otherwise since the value is opaque to the library. But that's pretty easy to do. Then when the function is called to create this you'd instead have the key ID passed in and from that you could look up the worker that it's associated with (if any).

internal/servers/repository_workerauth.go
```diff
@@ -0,0 +1,729 @@
...
+	return nil
+}
+
+// Node information is stored in two parts:
+// * the workerAuth record is stored with a reference to a worker
+// * certificate bundles are stored with a reference to the workerAuth record and issuing root certificate
+func (r *WorkerAuthRepositoryStorage) storeNodeInformation(ctx context.Context, node *types.NodeInformation) error {
+	const op = "workerauth.(WorkerAuthRepositoryStorage).storeNodeInformation"
+
+	if r.worker == nil {
```
jefferai, line 162: The first time node information is being stored I don't know for sure you'd have a worker yet. This is part of the discussions that happened today around when a worker and its ID are created but it may be that as part of storing node information worker creation will have to happen first, so either that'd have to happen and this would then not be nil, or this being nil would trigger that.

internal/servers/repository_workerauth.go
irenarindos: I had been assuming the calling function in Boundary (AuthorizeWorker) would do the lookup based on key ID to find/ create a worker before creating storage and passing in the worker.

internal/servers/repository_workerauth.go
irenarindos: I'd been assuming the calling function would handle worker creation; this nil check was in case the caller didn't pass in a worker when attempting to perform nodeInformation storage. 

I think worker lookup/ creation should happen first and should be external, esp when we get to worker rekeying.

jimlambrt:
Suggested edit:
```diff
diff --git a/internal/servers/repository_workerauth.go b/internal/servers/repository_workerauth.go
index 9e5d22c5a..f4cd7feab 100644
--- a/internal/servers/repository_workerauth.go
+++ b/internal/servers/repository_workerauth.go
@@ -194,32 +194,28 @@ func (r *WorkerAuthRepositoryStorage) storeNodeInformation(ctx context.Context,
 		Target: db.Constraint("worker_auth_authorized_pkey"),
 		Action: db.SetColumns([]string{"controller_encryption_priv_key"}),
 	}
-	if r.workerAuthTransaction != nil {
+
+	fn := func(w db.Writer) error {
 		// Store WorkerAuth
-		if err := r.workerAuthTransaction.transactionWriter.Create(ctx, &nodeAuth, db.WithOnConflict(onConflict)); err != nil {
+		if err := w.Create(ctx, &nodeAuth, db.WithOnConflict(onConflict)); err != nil {
 			return errors.Wrap(ctx, err, op)
 		}
 		// Then store cert bundles associated with this WorkerAuth
 		for _, c := range node.CertificateBundles {
-			err := r.storeWorkerCertBundle(ctx, c, nodeAuth.WorkerKeyIdentifier, r.workerAuthTransaction.transactionWriter)
+			err := r.storeWorkerCertBundle(ctx, c, nodeAuth.WorkerKeyIdentifier, w)
 			if err != nil {
 				return errors.Wrap(ctx, err, op)
 			}
 		}
+		return nil
+	}
+	if r.workerAuthTransaction != nil {
+		if err := fn(r.workerAuthTransaction.transactionWriter); err != nil {
+			return errors.Wrap(ctx, err, op)
+		}
 	} else {
 		_, err = r.writer.DoTx(ctx, db.StdRetryCnt, db.ExpBackoff{}, func(read db.Reader, w db.Writer) error {
-			// Store WorkerAuth
-			if err := w.Create(ctx, &nodeAuth, db.WithOnConflict(onConflict)); err != nil {
-				return errors.Wrap(ctx, err, op)
-			}
-			// Then store cert bundles associated with this WorkerAuth
-			for _, c := range node.CertificateBundles {
-				err := r.storeWorkerCertBundle(ctx, c, nodeAuth.WorkerKeyIdentifier, w)
-				if err != nil {
-					return errors.Wrap(ctx, err, op)
-				}
-			}
-			return nil
+			return fn(w)
 		},
 		)
 		if err != nil {

```

internal/servers/repository_workerauth.go
```diff
@@ -0,0 +1,729 @@
...
+		}
+		certBundle = append(certBundle, thisBundle)
+	}
+
+	return certBundle, nil
+}
+
+func (r *WorkerAuthRepositoryStorage) findWorkerAuth(ctx context.Context, node *types.NodeInformation) (*WorkerAuth, error) {
+	const op = "workerauth.(WorkerAuthRepositoryStorage).findWorkerAuth"
+
```
jimlambrt, line 451: node != nil

internal/servers/repository_workerauth.go
```diff
@@ -0,0 +1,729 @@
...
+			return nil, nil
+		}
+		return nil, errors.Wrap(ctx, err, op)
+	}
+
+	return worker, nil
+}
+
+func (r *WorkerAuthRepositoryStorage) loadRootCertificate(ctx context.Context, cert *types.RootCertificate, result proto.Message) error {
+	const op = "workerauth.(WorkerAuthRepositoryStorage).loadRootCertificate"
```
jimlambrt, line 471: check params for nil?

internal/servers/repository_workerauth.go
```diff
@@ -0,0 +1,729 @@
...
+		if err != nil {
+			return errors.Wrap(ctx, err, op)
+		}
+	case *types.RootCertificate:
+		err := r.removeRootCertificate(ctx, msg.GetId())
+		if err != nil {
+			return errors.Wrap(ctx, err, op)
+		}
+	default:
+		return errors.Wrap(ctx, fmt.Errorf("(%s) Message type not supported for Remove", op), op)
```
jimlambrt, line 537: same comment about adding type to error msg

jimlambrt (changes requested):
just some suggestions.

internal/servers/repository_workerauth.go
```diff
@@ -0,0 +1,729 @@
...
+// * the workerAuth record is stored with a reference to a worker
+// * certificate bundles are stored with a reference to the workerAuth record and issuing root certificate
+func (r *WorkerAuthRepositoryStorage) storeNodeInformation(ctx context.Context, node *types.NodeInformation) error {
+	const op = "workerauth.(WorkerAuthRepositoryStorage).storeNodeInformation"
+
+	if r.worker == nil {
+		return errors.Wrap(ctx, fmt.Errorf("Cannot store workerAuthentication without a worker reference"), op)
+	}
+
+	nodeAuth := AllocWorkerAuth()
```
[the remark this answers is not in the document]
irenarindos: Thanks for catching this!

internal/servers/repository_workerauth.go
```diff
@@ -0,0 +1,729 @@
...
+	node.Authorized = true
+
+	return unmarshalNodeToResult(ctx, node, result)
+}
+
+func (r *WorkerAuthRepositoryStorage) findCertBundles(ctx context.Context, workerKeyId string) ([]*types.CertificateBundle, error) {
+	const op = "workerauth.(WorkerAuthRepositoryStorage).findCertBundles"
+
+	where := fmt.Sprintf("worker_key_identifier= '%s'", workerKeyId)
+	var bundles []*WorkerCertBundle
```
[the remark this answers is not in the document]
jimlambrt: I'm not sure the proto has gorm tags for PK

internal/proto/controller/storage/servers/store/v1/worker_auth.proto
```diff
@@ -40,14 +40,14 @@ message WorkerAuth {
 // WorkerCertBundle contains all fields related to a WorkerCertBundle resource
 message WorkerCertBundle {
   // The public key of the issuing root certificate
```
jefferai, line 42: This comment is a bit confusing -- if it's actually part of the root certificate, I would suggest naming it something like `ca_certificate_public_key` or `root_certificate_public_key`...as is it seems like it's talking about the client certificate since that's the worker cert.

internal/proto/controller/storage/servers/store/v1/worker_auth.proto
```diff
@@ -40,14 +40,14 @@ message WorkerAuth {
...
   bytes certificate_public_key = 10;
 
   // The WorkerAuth worker_key_identifier this cert bundle record is for
-  // @inject_tag: `gorm:"not_null"`
+  // @inject_tag: `gorm:"primary_key"`
   string worker_key_identifier = 20;
 
   // CertBundle is the marshaled protobuf certificate bundle for a WorkerAuth
-  // @inject_tag: `gorm:"not_null"`
+  // @inject_tag: `gorm:"primary_key"`
```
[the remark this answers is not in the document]
jefferai: Agree, I don't think it's part of the PK. Although I do think that it should be marked immutable.

internal/servers/worker.go
```diff
@@ -4,12 +4,27 @@ import (
...
 )
 
+type workerAuthState struct {
+	WorkerId string
+}
+
+// AttachWorkerIdToState accepts a workerId and creates a struct
+// This is intended for use in worker authorization; AuthorizeNode accepts the option WithState
+// so that the workerId is passed through to storage
+func AttachWorkerIdToState(workerId string) (*structpb.Struct, error) {
```
jefferai, line 22: Would you need to possibly attach version as well?

internal/servers/repository_workerauth.go
```diff
@@ -0,0 +1,843 @@
...
+	}
+
+	// Determine type of message to store
+	marshaledBytes, err := proto.Marshal(msg.(proto.Message))
+	if err != nil {
+		return errors.New(ctx, errors.Encode, op, "error marshaling nodee.MessageWithId", errors.WithWrap(err))
+	}
+	switch msg.(type) {
+	case *types.NodeInformation:
+		node, err := unmarshalNodeInformation(ctx, marshaledBytes)
```
jefferai, line 81: How come you're marshaling and unmarshaling instead of just using the value directly?

internal/servers/repository_workerauth.go
```diff
@@ -0,0 +1,843 @@
...
+		return errors.New(ctx, errors.Encode, op, "error marshaling nodetypes.CertificateBundle", errors.WithWrap(err))
+	}
+
+	// Extract serial number from CA cert
+	caCert := bundle.CaCertificateDer
+	parsedCert, err := x509.ParseCertificate(caCert)
+	if err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+	workerCertBundle.CertificatePublicKey = parsedCert.AuthorityKeyId
```
jefferai, line 195: Yeah -- this is a bit confusing. I suggest renaming this field.

internal/servers/repository_workerauth.go
```diff
@@ -0,0 +1,843 @@
...
+	if err != nil {
+		return nil, errors.Wrap(ctx, err, op)
+	}
+	rootCert.SerialNumber = parsedCert.SerialNumber.Uint64()
+	rootCert.Certificate = cert.CertificateDer
+	rootCert.NotValidAfter = timestamp.New(cert.NotAfter.AsTime())
+	rootCert.NotValidBefore = timestamp.New(cert.NotBefore.AsTime())
+	rootCert.PublicKey = cert.PublicKeyPkix
+	rootCert.State = cert.Id
+	rootCert.IssuingCa = ca_id
```
jefferai, line 230: I see `ca_id` several times in the PR but I never actually see where it's set...where is this coming from?

internal/servers/worker.go
jefferai: Oh, I see that it's in certAuthState. Although maybe it's just a naming thing but I find it a bit confusing -- certs are generally immutable and workers aren't so it seems odd that cert has version and workers don't?

internal/servers/repository_workerauth.go
```diff
@@ -0,0 +1,843 @@
...
+func (r *WorkerAuthRepositoryStorage) Load(ctx context.Context, msg nodee.MessageWithId) error {
+	const op = "workerauth.(WorkerAuthRepositoryStorage).Load"
+	if err := types.ValidateMessage(msg); err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+	if msg.GetId() == "" {
+		return errors.New(ctx, errors.InvalidParameter, op, "given message cannot be loaded as it has no ID")
+	}
+
+	marshaledBytes, err := proto.Marshal(msg)
```
jefferai, line 336: Same question

internal/servers/repository_workerauth.go
```diff
@@ -0,0 +1,843 @@
...
+	if err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+	if authorizedWorker == nil {
+		return nodee.ErrNotFound
+	}
+
+	node.EncryptionPublicKeyBytes = authorizedWorker.WorkerEncryptionPubKey
+	node.CertificatePublicKeyPkix = authorizedWorker.WorkerSigningPubKey
+	node.RegistrationNonce = authorizedWorker.Nonce
```
jefferai, line 386: Nonce should be encrypted/decrypted too.

internal/servers/repository_workerauth.go
```diff
@@ -0,0 +1,843 @@
...
+}
+
+func (r *WorkerAuthRepositoryStorage) findCertBundles(ctx context.Context, workerKeyId string) ([]*types.CertificateBundle, error) {
+	const op = "workerauth.(WorkerAuthRepositoryStorage).findCertBundles"
+	if workerKeyId == "" {
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "empty workerKeyId")
+	}
+
+	var bundles []*WorkerCertBundle
+	err := r.reader.SearchWhere(ctx, &bundles, "worker_key_identifier=?", []interface{}{workerKeyId}, db.WithLimit(-1))
```
jefferai, line 420: This is a very very minor nit but your spacing isn't consistent in the where clause here and below in findWorkerAuth.

internal/servers/repository_workerauth.go
```diff
@@ -0,0 +1,843 @@
...
+	}
+
+	// Get cert bundles from the other table
+	certBundles, err := r.findCertBundles(ctx, node.Id)
+	if err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+	node.CertificateBundles = certBundles
+
+	return unmarshalNodeToResult(ctx, node, result)
```
jefferai, line 410: Any reason not to just return the node message directly instead of setting it on a proto.Message?

internal/proto/controller/storage/servers/store/v1/worker_auth.proto
irenarindos: Good catch, I'll rename it

internal/servers/repository_workerauth.go
irenarindos: It's in root_certificate.go- it's always "roots"

internal/servers/repository_workerauth.go
```diff
@@ -0,0 +1,843 @@
...
+
+	// Get workerId from state passed in
+	var result workerAuthState
+	err = mapstructure.Decode(node.State.AsMap(), &result)
+	if err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+	nodeAuth.WorkerId = result.WorkerId
+
+	err = nodeAuth.ValidateNewWorkerAuth(ctx)
```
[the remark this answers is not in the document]
irenarindos: The validation is on the constructed nodeAuth object so I can't move it but I'll update the error!

internal/servers/repository_workerauth.go
```diff
@@ -0,0 +1,754 @@
...
+
+	_, err = r.writer.DoTx(ctx, db.StdRetryCnt, db.ExpBackoff{}, func(read db.Reader, w db.Writer) error {
+		// Store WorkerAuth
+		storeNodeAuth := nodeAuth.clone()
+		if err := w.Create(ctx, &storeNodeAuth); err != nil {
+			return errors.Wrap(ctx, err, op)
+		}
+		// Then store cert bundles associated with this WorkerAuth
+		for _, c := range node.CertificateBundles {
+			if err := r.storeWorkerCertBundle(ctx, c, nodeAuth.WorkerKeyIdentifier, w); err != nil {
```
jefferai, line 145: This is an empty branch -- it will do nothing and return nil rather than abort the transaction

internal/servers/repository_workerauth.go
jefferai: Doesn't need to be changed now but we don't generally use underscores in names...something like `rootsId` or `caId` would be preferable

jefferai (approved):
Looks good after addressing my one new comment

PR #2094, opened 2022-05-20
Respect max open connection limits in migrate and init

johanbrandhorst:
## [fix(cmd): Respect DB max open connections](https://github.com/hashicorp/boundary/commit/0b81e1505264081c7c5f65f1999eb32d0772052e)

Both the init and migrate commands were ignoring
the configured max_open_connections database
setting. This fix ensures that both commands
respect the limits.

### DIFF
--- a/internal/cmd/commands/database/funcs.go
+++ b/internal/cmd/commands/database/funcs.go
@@ -16,7 +16,7 @@ import (
 // We expect the database already to be initialized iff initialized is set to true.
 // Returns a cleanup function which must be called even if an error is returned and
 // an error code where a non-zero value indicates an error happened.
-func migrateDatabase(ctx context.Context, ui cli.Ui, dialect, u string, initialized bool) (func(), int) {
+func migrateDatabase(ctx context.Context, ui cli.Ui, dialect, u string, initialized bool, maxOpenConns int) (func(), int) {
 	noop := func() {}
 	// This database is used to keep an exclusive lock on the database for the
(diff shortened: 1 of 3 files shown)


internal/cmd/commands/database/funcs.go
```diff
@@ -25,6 +25,8 @@ func migrateDatabase(ctx context.Context, ui cli.Ui, dialect, u string, initiali
 		ui.Error(fmt.Errorf("Error establishing db connection: %w", err).Error())
```
johanbrandhorst, line 25: I kept this `SqlOpen` in this function because the whole connection is scoped to this function, and it means we can safely defer the close. Moving it out of this scope makes the closing a little messier (and we have to do it everywhere this function is called).

internal/cmd/commands/database/funcs.go
```diff
@@ -25,6 +25,8 @@ func migrateDatabase(ctx context.Context, ui cli.Ui, dialect, u string, initiali
...
 		return noop, 2
 	}
+	dBase.SetMaxOpenConns(maxOpenConns)
+	defer dBase.Close()
```
johanbrandhorst, line 29: Hm, I'm being told this may be intentionally kept open, so this may be incorrect.

johanbrandhorst: Remove it

lilic (approved):
Nice discovery, thank you so much for this! 🎉

PR #2106, opened 2022-05-25
Session performance changes

tmessi:
This makes two changes to the list `/v1/sessions` endpoint to address
some scale issues:

---

perf(session): Remove connections from session list endpoint

Including the nested connection information in the session list response 
had a negative impact on performance as it created a significant multiplier
on the number of rows that needed to be fetched from the database.

It was also determined that this information was used by neither the admin
ui nor the desktop client.

Note that nested connection information can still be retrieved via a read
request for a specific session.

---

perf(session): Change list behavior around terminated sessions

It was determined that it is a more common use case for users to request 
non-terminated sessions (i.e. sessions in `pending`, `active`, or
`canceling` when requesting sessions. Notably, end users would need 
non-terminated sessions to:

- Establish a new connection to a session.
- Terminate a session

So now by default a list request to `/v1/sessions` will *not* include 
terminated sessions.

To still enable use cases where terminated sessions are needed, this adds a
new query parameter to include the terminated session:
`/v1/sessions?include_terminated=true`.

Note that this query parameter is considered experimental and subject to 
change.


internal/daemon/controller/handlers/sessions/session_service_test.go
```diff
@@ -423,14 +409,8 @@ func TestList(t *testing.T) {
 				require.Equal(len(tc.res.GetItems()), len(got.GetItems()), "Didn't get expected number of sessions: %v", got.GetItems())
 				for i, wantSess := range tc.res.GetItems() {
 					assert.True(got.GetItems()[i].GetExpirationTime().AsTime().Sub(wantSess.GetExpirationTime().AsTime()) < 10*time.Millisecond)
-					assert.Equal(1, len(wantSess.GetConnections()))
+					assert.Equal(0, len(wantSess.GetConnections())) // no connections on list
```
johanbrandhorst, line 467: We decided not to change the proto definitions for this change then?

api/sessions/option.gen.go
```diff
@@ -86,3 +87,15 @@ func WithRecursive(recurse bool) Option {
 		o.withRecursive = true
 	}
 }
+
+func WithIncludeTerminated(inIncludeTerminated bool) Option {
+	return func(o *options) {
+		o.queryMap["include_terminated"] = fmt.Sprintf("%v", inIncludeTerminated)
```
johanbrandhorst, line 93: Nit: I would prefer to use `strconv.FormatBool` here to keep the strong typing

internal/proto/controller/api/services/v1/session_service.proto
```diff
@@ -63,6 +63,9 @@ message ListSessionsRequest {
   string scope_id = 1; // @gotags: `class:"public"`
   bool recursive = 20 [json_name = "recursive"]; // @gotags: `class:"public"`
   string filter = 30 [json_name = "filter"]; // @gotags: `class:"public"`
+  // Experimental. By default only non-terminated (i.e. pending, active, canceling) are returned.
```
hugoamvieira, line 66: The "Experimental" keyword strikes me as odd here, considering it's the default behaviour right now. 

As I understand it, we're eventually deprecating this behaviour, so wouldn't this be a legacy/compatibility option or something like that?

hugoamvieira (commented):
Looks good! A Q about documentation - should we add an API change notice/deprecation notice on the changelog?

Looking at the swagger changes, it seems like we'll have some details on the "HTTP API" page, but that page wouldn't be the documentation "entry point" to see what changed when someone updates, so I can see a scenario where someone updates and this catches them off-guard.

api/sessions/option.gen.go
tmessi: This is generated code

johanbrandhorst: 🙄 so it is, my bad!

internal/daemon/controller/handlers/sessions/session_service_test.go
tmessi: Probably less "decided" and more "forgot about that conversation". I can look at doing that as a refactor on top of this though. Since it might cause a bit of code churn.

johanbrandhorst: Sounds good! Personally I think I'd be in favor of a different message for the List endpoint, but we should see if there's any precedent for other list endpoints first.

internal/proto/controller/api/services/v1/session_service.proto
jefferai: Experimental is on the flag though, right? The current behavior is changing, so rather than just change it, there is a flag that may or may not continue existing, hence experimental.

I also don't think it's legacy/compatibility, especially given that we may never deprecate it and may simple officially support it once we are cleaning out terminated sessions with regularity.

PR #2110, opened 2022-05-26
Merge the worker_status table with the worker table

talanknight:
Instead of migrating ListTagsForWorkers repo function to lookup by name (which is what the worker service relies on) only to remove it in a future PR, I have removed ListTagsForWorkers now and moved LookupWorkerByWorkerReportedName methods which returns the worker with all of its tags.


internal/db/schema/migrations/oss/postgres/99/02_worker_controller_tables.up.sql
```diff
@@ -40,6 +40,20 @@ create table server_worker (
   create_time wt_timestamp,
   update_time wt_timestamp,
   version wt_version,
+  last_status_time timestamp with time zone,
+  constraint last_status_time_after_create_time
```
irenarindos, line 44: Should this maybe be `last_status_time_not_before_create_time` since the check is >=?

internal/db/schema/migrations/oss/postgres/99/02_worker_controller_tables.up.sql
```diff
@@ -55,35 +69,41 @@ create trigger default_create_time_column before insert on server_worker
...
+  'function used to update the last_update_time column in server_worker to now';
 
-create trigger default_create_time_column before insert on server_worker_status
-  for each row execute procedure default_create_time();
+create trigger update_server_worker_last_update_time_column before update of worker_reported_address, worker_reported_name on server_worker
+  for each row execute procedure update_server_worker_last_update_time_column();
 
-create trigger worker_insert_time_column before insert on server_worker_status
-  for each row execute procedure update_time_column();
+create function insert_server_worker_last_update_time_column()
```
irenarindos, line 92: Here and elsewhere
```suggestion
create function insert_server_worker_update_last_status_time_column()
```

internal/db/schema/migrations/oss/postgres/99/02_worker_controller_tables.up.sql
```diff
@@ -55,35 +69,41 @@ create trigger default_create_time_column before insert on server_worker
 create trigger worker_insert_time_column before insert on server_worker
   for each row execute procedure update_time_column();
 
-create trigger worker_update_time_column before update on server_worker
+create trigger worker_update_time_column before update of version, description, name, address on server_worker
```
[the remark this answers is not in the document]
mgaffney: I think of the `updated_time` column as the timestamp of when the row was updated. So, to answer your question, I think it should be updated.

internal/db/schema/migrations/oss/postgres/99/02_worker_controller_tables.up.sql
```diff
@@ -55,35 +69,41 @@ create trigger default_create_time_column before insert on server_worker
...
+create function update_server_worker_last_update_time_column()
+  returns trigger
+as $$
+begin
+  new.last_status_time = now();
+  return new;
+end;
+$$ language plpgsql;
+comment on function update_server_worker_last_update_time_column is
+  'function used to update the last_update_time column in server_worker to now';
```
mgaffney, line 87: This description doesn't match what the function does.

internal/db/schema/migrations/oss/postgres/99/02_worker_controller_tables.up.sql
```diff
@@ -55,35 +69,41 @@ create trigger default_create_time_column before insert on server_worker
...
+as $$
+begin
+  if new.worker_reported_address is not null or new.worker_reported_name is not null then
+    new.last_status_time = now();
+  end if;
+  return new;
+end;
+$$ language plpgsql;
+comment on function insert_server_worker_last_update_time_column is
+  'function used to update the last_update_time column in server_worker to now';
```
mgaffney, line 103:
```suggestion
  'function used to update the last_status_time column in server_worker to now';
```

PR #2114, opened 2022-05-26
Create the API Worker resource and service definition

talanknight:
These are not complete by any means.  My hope with getting this in now is that if it is merged before work begins on the service handlers it might make it easier for multiple people to work on it at once.

The worker resource still needs the worker auth fields.
The service definition still needs the add/set/remove tags rpcs as well as any rpcs needed for the auth flow.


irenarindos (approved):
LGTM

PR #2115, opened 2022-05-27
feat(target): Add support for static credential in AddCredentialSources

louisruch:


louisruch:
@mgaffney @ddebko PTAL
I was not able to just do the domain work as it was being used by the service handlers so I did the API as well. I have manually tested this in the CLI but will add more detail to that in a different PR with bats tests.

louisruch:
@mgaffney PTAL

louisruch:
@tmessi PTAL

tmessi (approved):

PR #2117, opened 2022-05-27
Add worker update to server repo

talanknight:


internal/servers/repository_worker.go
```diff
@@ -271,6 +273,83 @@ func setWorkerTags(ctx context.Context, w db.Writer, id string, ts TagSource, ta
...
+// fields, if no updatable fields are included in the fieldMaskPaths, then an
+// error is returned.  If any paths besides those listed above are included in
+// the path then an error is returned.
+func (r *Repository) UpdateWorker(ctx context.Context, worker *Worker, version uint32, fieldMaskPaths []string, opt ...Option) (*Worker, int, error) {
+	const op = "servers.(Repository).UpdateWorker"
+	switch {
+	case worker == nil:
+		return nil, db.NoRowsAffected, errors.New(ctx, errors.InvalidParameter, op, "worker is nil")
+	case worker.PublicId == "":
+		return nil, db.NoRowsAffected, errors.New(ctx, errors.InvalidParameter, op, "missing public id")
```
irenarindos, line 289: Does version need to be checked?

```suggestion
		return nil, db.NoRowsAffected, errors.New(ctx, errors.InvalidParameter, op, "missing public id")
	case version == 0 {
		return nil, db.NoRowsAffected, errors.New(ctx, errors.InvalidParameter, op, "missing version")
	}
```

irenarindos (changes requested):

irenarindos (approved):

internal/servers/repository_worker.go
```diff
@@ -271,6 +273,85 @@ func setWorkerTags(ctx context.Context, w db.Writer, id string, ts TagSource, ta
...
+	case version == 0:
+		return nil, db.NoRowsAffected, errors.New(ctx, errors.InvalidParameter, op, "version is zero")
+	}
+
+	for _, f := range fieldMaskPaths {
+		switch {
+		case strings.EqualFold("name", f):
```
jimlambrt, line 296: perhaps use consts (defined at the top of the func) for these? (they are used a few times in this func)

internal/servers/repository_worker.go
```diff
@@ -271,6 +273,85 @@ func setWorkerTags(ctx context.Context, w db.Writer, id string, ts TagSource, ta
...
+			"description": worker.Description,
+			"address":     worker.Address,
+		},
+		fieldMaskPaths,
+		nil,
+	)
+	if len(dbMask) == 0 && len(nullFields) == 0 {
+		return nil, db.NoRowsAffected, errors.New(ctx, errors.EmptyFieldMask, op, "no fields to update")
+	}
+	worker = worker.clone()
```
jimlambrt, line 317: perhaps move this into the DoTx(...) so a "fresh" clone will be sent with every txn attempt.

internal/servers/repository_worker_test.go
```diff
@@ -551,3 +551,189 @@ func TestRepository_CreateWorker(t *testing.T) {
 		})
 	}
 }
+
+func TestRepository_UpdateWorker(t *testing.T) {
```
jimlambrt, line 555: in general: it might be nice to test unhappy paths since I think it's fairly trivial to test all of them, even without resorting to sqlmock (see CreateWorker tests).

internal/servers/repository_worker_test.go
```diff
@@ -551,3 +551,189 @@ func TestRepository_CreateWorker(t *testing.T) {
...
+	ctx := context.Background()
+	conn, _ := db.TestSetup(t, "postgres")
+	rw := db.New(conn)
+	wrapper := db.TestWrapper(t)
+	kmsCache := kms.TestKms(t, conn, wrapper)
+
+	repo, err := servers.NewRepository(rw, rw, kmsCache)
+	require.NoError(t, err)
+
+	setCases := []struct {
```
jimlambrt, line 566: ty for grouping these test cases into different slices.

jimlambrt (changes requested):
Looks great!  I had just a few possible suggestions.

PR #2123, opened 2022-06-02
Add the Read/List service functions

talanknight:
Additionally I have set up the handler registration for the worker service so requests can be sent the the `/v1/workers` path.

### DIFF
--- a/internal/daemon/controller/handlers/workers/worker_service_test.go
+++ b/internal/daemon/controller/handlers/workers/worker_service_test.go
@@ -173,7 +173,7 @@ func TestList(t *testing.T) {
 			},
 		},
 		{
-			name: "Filter to a 2 workers",
+			name: "Filter to 2 workers",
 			req:  &pbs.ListWorkersRequest{ScopeId: "global", Recursive: true, Filter: `"/item/name" matches "worker[23]"`},
 			res: &pbs.ListWorkersResponse{
 				Items: wantWorkers[2:4],
--- a/internal/servers/testing.go
+++ b/internal/servers/testing.go
@@ -92,13 +92,13 @@ func TestWorker(t *testing.T, conn *db.DB, wrapper wrapping.Wrapper, opt ...Opti
 	require.NoError(t, err)
 	ctx := context.Background()
 
-	namePart, err := newWorkerId(context.Background())
+	namePart, err := newWorkerId(ctx)
 	require.NoError(t, err)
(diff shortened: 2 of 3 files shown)


internal/proto/controller/api/resources/workers/v1/worker.proto
```diff
@@ -18,9 +18,6 @@ message WorkerConfig {
   // Output only.  The name the worker daemon reports itself having.
   string name = 20;
 
-  // Output only.  The boundary version the worker has reported itself as running.
-  string version = 30;
```
irenarindos, line 22: Don't we still need version?

irenarindos (changes requested):

internal/servers/testing.go
```diff
@@ -81,22 +81,48 @@ func TestWorkerAuth(ctx context.Context, t *testing.T, conn *db.DB, worker *Work
...
+func TestWorker(t *testing.T, conn *db.DB, wrapper wrapping.Wrapper, opt ...Option) *Worker {
 	t.Helper()
 	rw := db.New(conn)
 	kms := kms.TestKms(t, conn, wrapper)
 	serversRepo, err := NewRepository(rw, rw, kms)
 	require.NoError(t, err)
+	ctx := context.Background()
 
-	id, err := newWorkerId(context.Background())
+	namePart, err := newWorkerId(context.Background())
```
A-440Hz, line 95: why is this context different from the one used for `UpdateWorker` at the end?

internal/daemon/controller/handlers/workers/worker_service_test.go
```diff
@@ -0,0 +1,223 @@
...
+		},
+		{
+			name: "Filter to a single workers",
+			req:  &pbs.ListWorkersRequest{ScopeId: "global", Recursive: true, Filter: `"/item/name"=="worker2"`},
+			res: &pbs.ListWorkersResponse{
+				Items: wantWorkers[2:3],
+			},
+		},
+		{
+			name: "Filter to a 2 workers",
```
irenarindos, line 176:
```suggestion
			name: "Filter to 2 workers",
```

internal/proto/controller/api/resources/workers/v1/worker.proto
irenarindos: Does the version field on the workerAggregate need to be removed as well?

irenarindos (approved):

PR #2124, opened 2022-06-02
Add delete service function implementation.

talanknight:


internal/daemon/controller/handlers/workers/worker_service_test.go
```diff
@@ -221,3 +221,66 @@ func TestList(t *testing.T) {
...
+				Id: servers.WorkerPrefix + "_doesntexis",
+			},
+			err: handlers.ApiErrorWithCode(codes.NotFound),
+		},
+		{
+			name:    "Bad Worker Id formatting",
+			scopeId: w.GetScopeId(),
+			req: &pbs.DeleteWorkerRequest{
+				Id: "bad_format",
+			},
```
irenarindos, line 271: Should there be a test for delete a worker that doesn't exist?

irenarindos (changes requested):

irenarindos (approved):

PR #2125, opened 2022-06-02
refactor (servers/workerauth): add StoreNodeInformationTx(...)

jimlambrt:
refactor to export StoreNodeInformationTx(...) which can be used
by other repositories within their transactions to store
node information.

### DIFF
--- a/internal/servers/repository_workerauth.go
+++ b/internal/servers/repository_workerauth.go
@@ -74,8 +74,14 @@ func (r *WorkerAuthRepositoryStorage) Store(ctx context.Context, msg nodee.Messa
 	// Determine type of message to store
 	switch t := msg.(type) {
 	case *types.NodeInformation:
-		err := r.storeNodeInformation(ctx, t)
+		// Encrypt the private key
+		databaseWrapper, err := r.kms.GetWrapper(ctx, scope.Global.String(), kms.KeyPurposeDatabase)
 		if err != nil {
+			return errors.Wrap(ctx, err, op)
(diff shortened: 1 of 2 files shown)


internal/servers/repository_workerauth.go
```diff
@@ -90,11 +96,26 @@ func (r *WorkerAuthRepositoryStorage) Store(ctx context.Context, msg nodee.Messa
...
+// allows this capability to be shared with other repositories or just called
+// within a transaction.  To be clear, this repository function doesn't include
+// its own transaction and is intended to be used within a transaction provided
+// by the caller.
+//
 // Node information is stored in two parts:
 // * the workerAuth record is stored with a reference to a worker
 // * certificate bundles are stored with a reference to the workerAuth record and issuing root certificate
-func (r *WorkerAuthRepositoryStorage) storeNodeInformation(ctx context.Context, node *types.NodeInformation) error {
+func StoreNodeInformationTx(ctx context.Context, writer db.Writer, databaseWrapper wrapping.Wrapper, node *types.NodeInformation, _ ...Option) error {
```
irenarindos, line 111: If options aren't used should they be omitted?

irenarindos (approved):
Just one small suggested change

internal/servers/repository_workerauth.go
jimlambrt: it's fine to keep them in my opinion, as long as it's clear (via _ ) that they currently aren't in use.

PR #2130, opened 2022-06-03
feature(WorkerAuth): Create RotateRoots domain and job

irenarindos:
- Create RotateRoots domain function 
- Create RotateRoots job. The current default certificate lifetime is set to 1 year; rotation will occur every 6 months. Rotation should occur on first start up and from then on at the job frequency
- Remove rotate roots call on startup; this is now achieved by the rotate roots job

### DIFF
--- a/internal/servers/job/jobs.go
+++ b/internal/servers/job/jobs.go
@@ -1,0 +1,50 @@
+package servers
+
+import (
+	"context"
+	"reflect"
+
+	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/kms"
+	"github.com/hashicorp/boundary/internal/scheduler"
+)
+
+// RegisterJobs registers the rotate roots job with the provided scheduler.
+func RegisterJobs(ctx context.Context, scheduler *scheduler.Scheduler, r db.Reader, w db.Writer, kms *kms.Kms) error {
+	const op = "servers.(Jobs).RegisterJobs"
+
+	if isNil(scheduler) {
+		return errors.New(ctx, errors.InvalidParameter, op, "missing scheduler")
+	}
+	if isNil(r) {
+		return errors.New(ctx, errors.InvalidParameter, op, "missing reader")
+	}
+	if isNil(w) {
+		return errors.New(ctx, errors.InvalidParameter, op, "missing writer")
+	}
+	if kms == nil {
+		return errors.New(ctx, errors.InvalidParameter, op, "missing kms")
+	}
+
+	rotateRootsJob, err := newRotateRootsJob(ctx, r, w, kms)
+	if err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+	if err = scheduler.RegisterJob(ctx, rotateRootsJob); err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+
+	return nil
+}
+
+func isNil(i interface{}) bool {
+	if i == nil {
+		return true
+	}
+	switch reflect.TypeOf(i).Kind() {
+	case reflect.Ptr, reflect.Map, reflect.Array, reflect.Chan, reflect.Slice:
+		return reflect.ValueOf(i).IsNil()
+	}
+	return false
+}
--- a/internal/servers/job/rotate_roots_job.go
+++ b/internal/servers/job/rotate_roots_job.go
@@ -1,0 +1,82 @@
+package servers
+
+import (
+	"context"
+	"time"
+
+	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/kms"
+	"github.com/hashicorp/boundary/internal/scheduler"
+	"github.com/hashicorp/boundary/internal/servers"
+)
+
+const rotateFrequency = time.Hour
+
+// rotateRootsJob defines a periodic job that initiates root certificate rotation
+// It runs every hour; the root rotation function in the library is designed to not
+// do anything if it's not time to rotate (roots are within their valid ranges)
+type rotateRootsJob struct {
+	workerAuthRepo *servers.WorkerAuthRepositoryStorage
+
+	totalRotates int
+}
+
+// newRotateRootsJob instantiates the rotate roots job.
+func newRotateRootsJob(ctx context.Context, r db.Reader, w db.Writer, kms *kms.Kms) (*rotateRootsJob, error) {
+	const op = "servers.newRotateRootsJob"
+	switch {
+	case isNil(r):
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing reader")
+	case isNil(w):
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing writer")
+	case kms == nil:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing kms")
+	}
+
+	workerAuthRepo, err := servers.NewRepositoryStorage(ctx, r, w, kms)
+	if err != nil {
+		return nil, errors.Wrap(ctx, err, op)
+	}
+
+	return &rotateRootsJob{
+		workerAuthRepo: workerAuthRepo,
+		totalRotates:   0,
+	}, nil
+}
+
+// Name returns a short, unique name for the job.
+func (r *rotateRootsJob) Name() string { return "rotate_roots" }
+
+// Description returns the description for the job.
+func (r *rotateRootsJob) Description() string {
+	return "Rotate root certificates"
+}
+
+// NextRunIn returns the next run time after a job is completed.
+// This is represented by RotateFrequency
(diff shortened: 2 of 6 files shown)


internal/servers/worker_auth.go
```diff
@@ -2,12 +2,19 @@ package servers
...
 
 	"github.com/hashicorp/boundary/internal/errors"
 	"github.com/hashicorp/boundary/internal/servers/store"
 	"google.golang.org/protobuf/proto"
 )
 
+const (
+	// DefaultCertificateLifetime represents how long root certificates are valid for.
+	// The default is set to a year
+	DefaultCertificateLifetime = time.Hour * 24 * 365 * 24
```
louisruch, line 15:
```suggestion
	DefaultCertificateLifetime = time.Hour * 24 * 365
```

internal/servers/job/jobs.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/kms"
+	"github.com/hashicorp/boundary/internal/scheduler"
+)
+
+// RegisterJobs registers rotate roots jobs with the provided scheduler.
```
louisruch, line 12:
```suggestion
// RegisterJobs registers the rotate roots job with the provided scheduler.
```

louisruch: Alternatively if there will be other jobs in the future, the comment could just be generic about registering all required server jobs.

internal/servers/job/rotate_roots_job_test.go
```diff
@@ -0,0 +1,106 @@
...
+				require.Error(err)
+				assert.Nil(got)
+				assert.Truef(errors.Match(errors.T(tt.wantErrCode), err), "Unexpected error %s", err)
+				return
+			}
+			require.NoError(err)
+			require.NotNil(got)
+			require.Equal(0, got.totalRotates)
+
+			// Run job and ensure rotation was performed
```
louisruch, line 98: minor nit: As I am not sure exactly what validation for rotate means. It might be nice to see a `workerAuthRepo.List(` before `got.Run` to see to that the List moved from x to `2`

louisruch (changes requested):
Job registration looks good to me, I noticed one issue with the lifetime const, otherwise left a couple of minor comments!

internal/servers/worker_auth.go
```diff
@@ -2,12 +2,19 @@ package servers
...
+	"time"
 
 	"github.com/hashicorp/boundary/internal/errors"
 	"github.com/hashicorp/boundary/internal/servers/store"
 	"google.golang.org/protobuf/proto"
 )
 
+const (
+	// DefaultCertificateLifetime represents how long root certificates are valid for.
+	// The default is set to a year
```
jefferai, line 14: Can you add a FIXME to make it clear that the year lifetime is temporary?

louisruch: I am not sure what the non-temporary plan is. But if root certificates will each get their own lifetime when created, please also add a `FIXME` to the job as well. The job would need to behave like the Vault Token renewal job which runs more frequently and uses the expiry time in the database to determine if the certificates needs to be renewed or not rather than a blanket renewal of all tokens.

internal/servers/service_rotate_roots.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/hashicorp/nodeenrollment/rotation"
+)
+
+// RotateRoots is a domain service function that initiates a rotation of root certificates
+// via a call to the nodenenrollment RotateRootCertificates function
+// Accepts the nodeenrollment option, WithCertificateLifetime(time.Duration) to specify the lifetime
+// of the generated cert(s)
+func RotateRoots(ctx context.Context, workerAuthRepo *WorkerAuthRepositoryStorage, opt ...nodeenrollment.Option) error {
+	const op = "servers.RotateRoots"
+
```
jimlambrt, line 17: perhaps check workAuthRepo != nil ?  I realized RotateRootCertificates also checks this, but why make the call with an invalid parameter?

internal/servers/service_rotate_roots_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+func TestRotateRoots(t *testing.T) {
+	require, assert := require.New(t), assert.New(t)
+	ctx := context.Background()
+	wrapper := db.TestWrapper(t)
+	conn, _ := db.TestSetup(t, "postgres")
+	kmsCache := kms.TestKms(t, conn, wrapper)
+	err := kmsCache.CreateKeys(context.Background(), scope.Global.String(), kms.WithRandomReader(rand.Reader))
+	require.NoError(err)
+	wrapper, err = kmsCache.GetWrapper(context.Background(), scope.Global.String(), kms.KeyPurposeDatabase)
```
jimlambrt, line 26: same comment as before: is this needed and if it needed perhaps name it something like `databaseWrapper` so it's not confused with the wrapper (of the root variety) used with the test kms.

internal/servers/service_rotate_roots.go
```diff
@@ -0,0 +1,24 @@
...
+	_, err := rotation.RotateRootCertificates(ctx, workerAuthRepo, opt...)
+	if err != nil {
+		return errors.Wrap(ctx, err, op)
```
jimlambrt, line 20: perhaps a unit test for this failure mode?

jimlambrt (changes requested):
Very easy to read and great work!  A few minor things...

internal/servers/worker_auth.go
jefferai: Don't define this. The root rotation function in the library is specifically designed to not do anything if it's not time to rotate. Have it simply run once an hour or so. If in the future we allow people to modify the default certificate lifetime (which, if we needed to reference it, should really be `nodeenrollment.DefaultCertificateLifetime` anyways) we can simply use fractions of that defined value.

internal/servers/job/rotate_roots_job.go
```diff
@@ -0,0 +1,91 @@
...
+	case kms == nil:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing kms")
+	}
+
+	workerAuthRepo, err := servers.NewRepositoryStorage(ctx, r, w, kms)
+	if err != nil {
+		return nil, errors.Wrap(ctx, err, op)
+	}
+
+	frequency := certificateLifetime / 2
```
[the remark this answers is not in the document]
jefferai: Don't game it out to certificate lifetime; run it once an hour or so. If we allow people to set it to a very short period we could want/need to be able to set this to less than an hour but I think we can wait until it's necessary to do that.

internal/servers/worker_auth.go
irenarindos: I've updated the job to run once an hour, and I've added a FIXME indicating the year cert lifetime is temporary

internal/servers/service_rotate_roots.go
irenarindos: What's the best way to force a failure in the underlying function?

jimlambrt: could you just create WorkerAuthRepositoryStorage with a writer == `&db.Db{}`?

internal/servers/job/rotate_roots_job.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+}
+
+// Run executes the job by calling the rotateRoots domain function
+func (r *rotateRootsJob) Run(ctx context.Context) error {
+	const op = "servers.(rotateRootsJob).Run"
+
+	err := servers.RotateRoots(ctx, r.workerAuthRepo, nodeenrollment.WithCertificateLifetime(certificateLifetime))
+	if err != nil {
+		return errors.Wrap(ctx, err, op)
```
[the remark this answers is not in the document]
jimlambrt: Just referencing the other suggestion where you give the workAuthRepo a bad writer to test this failure mode.

internal/servers/job/rotate_roots_job.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/kms"
+	"github.com/hashicorp/boundary/internal/scheduler"
+	"github.com/hashicorp/boundary/internal/servers"
+)
+
+const rotateFrequency = time.Hour
+
+// rotateRootsJob defines a periodic job that initiates root certificate rotation
+// rotateFrequency represents how often rotate is called, and
```
jefferai, line 17: Comment should probably be updated

jimlambrt (approved):
amazing work!  I did suggest one way you could test a few failure modes, whehter you implement that suggestion or not there's no need to re-review

jefferai (approved):

jefferai:
Looks good to me! Just a comment to update.

internal/servers/service_rotate_roots.go
irenarindos: Ooo thanks!

PR #2132, opened 2022-06-03
Migrate storage to database repo

jefferai:


internal/daemon/controller/controller.go
```diff
@@ -414,13 +416,38 @@ func (c *Controller) AuthorizeNodeeWorker(request string) error {
 	const op = "controller.(Controller).AuthorizeWorker"
```
irenarindos, line 416: Could this function have documentation?

irenarindos, line 416: Unsure if the op needs to change or if the function name should
```suggestion
	const op = "controller.(Controller).AuthorizeNodeeWorker"
```

internal/servers/worker.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 type workerAuthWorkerId struct {
-	WorkerId string
+	WorkerId string `mapstructure:"worker_id"`
```
irenarindos, line 16: Can you update the AttachWorkerIdToState function below to reflect this change? `TestStoreWorkerAuth` is failing because it is no longer setting the workerId correctly

irenarindos (changes requested):

internal/daemon/controller/controller.go
jefferai: This function is a temporary function until the create PR lands. I'll actually probably merge Jim's PR first and then remove it.

internal/servers/worker.go
jefferai: I have this fixed in another commit now

irenarindos (approved):

PR #2137, opened 2022-06-04
feature (events): add TestWithoutEventing(t)

jimlambrt:
TestWithoutEventing allows the caller to "disable" all eventing for
a test. You must not run the test in parallel (no calls to t.Parallel)
since the function relies on modifying the system wide default eventer.

### DIFF
--- a/internal/observability/event/testing.go
+++ b/internal/observability/event/testing.go
@@ -27,7 +27,6 @@ func TestWithoutEventing(t testing.TB) *Eventer {
 		ObservationsEnabled: false,
 		SysEventsEnabled:    false,
 	}
-	// testConfig := TestEventerConfig(t, "InitSysEventer")
 	testLock := &sync.Mutex{}
 	testLogger := hclog.New(&hclog.LoggerOptions{
 		Mutex:  testLock,
--- a/internal/observability/event/testing.go
+++ b/internal/observability/event/testing.go
@@ -4,15 +4,41 @@ import (
 	"context"
 	"io/ioutil"
 	"os"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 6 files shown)


internal/observability/event/testing.go
```diff
@@ -4,16 +4,42 @@ import (
...
 	"github.com/hashicorp/boundary/sdk/pbs/controller/api/resources/groups"
 	"github.com/hashicorp/eventlogger"
+	"github.com/hashicorp/go-hclog"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 )
 
+// TestWithoutEventing allows the caller to "disable" all eventing for a test.
+// You must not run the test in parallel (no calls to t.Parallel) since the
+// function relies on modifying the system wide default eventer.
```
jefferai, line 21: I think most tests don't really do eventing already; it happens via those that use TestWorker/TestController right? In which case I wonder if having a flag to these that disables eventing locally (for example perhaps by removing eventing info from the base context?) would allow for some of these to be done in parallel.

jimlambrt: If there's no event info, the system still events because that's a possible error condition and we don't want to drop events because some portion of the event context and/or infra is missing/down.  First the event system falls back to the "system eventer" and if that's not available it will resort to stderr logging.   The only way to completely shutoff eventing is to circumvent all those fails safes by initializing a system eventer that goes to dev/null... unfortunately, that requires updating "global state" (in the form of a package event pkg var) which means you can't run the tests in parallel.  

With all that said, I've added a new option to NewTestController(...) to support no eventing from that path as well.  This path also has the same "no parallel" constraint.

jefferai (dismissed):

PR #2138, opened 2022-06-05
Add worker creation ACLs

jefferai:
This is a bit more than simply adding the action; we had some
checks to ensure that list/create could only happen/not happen
in some cases and those had to be modified to account for
subactions.


jimlambrt (approved):
Looks great!

PR #2139, opened 2022-06-05
Use persistent storage for worker auth credentials

jefferai:
Tests are TODO but throwing this up for review while Jim's around :-D

### DIFF
--- a/internal/cmd/commands/server/server.go
+++ b/internal/cmd/commands/server/server.go
@@ -463,22 +463,10 @@ func (c *Command) Run(args []string) int {
 			}
 			return base.CommandCliError
 		}
-		if c.WorkerAuthKms == nil {
-			req := c.worker.WorkerAuthRegistrationRequest
-			if req == "" {
-				retErr := fmt.Errorf("No worker registration request found at worker startup time")
-				if err := c.worker.Shutdown(); err != nil {
-					c.UI.Error(retErr.Error())
-					retErr = fmt.Errorf("Error shutting down worker: %w", err)
-				}
-				c.UI.Error(retErr.Error())
-				if err := c.controller.Shutdown(); err != nil {
-					c.UI.Error(fmt.Errorf("Error with controller shutdown: %w", err).Error())
-				}
-				return base.CommandCliError
-			}
+
+		if c.WorkerAuthKms == nil && c.worker.WorkerAuthRegistrationRequest != "" {
 			c.InfoKeys = append(c.InfoKeys, "worker auth registration request")
-			c.Info["worker auth registration request"] = req
+			c.Info["worker auth registration request"] = c.worker.WorkerAuthRegistrationRequest
 			c.InfoKeys = append(c.InfoKeys, "worker auth current key id")
 			c.Info["worker auth current key id"] = c.worker.WorkerAuthCurrentKeyId
 		}
@@ -646,6 +634,14 @@ func (c *Command) StartWorker() error {
 	c.worker, err = worker.New(conf)
 	if err != nil {
 		return fmt.Errorf("Error initializing worker: %w", err)
+	}
+
+	if c.WorkerAuthKms == nil {
+		if c.worker.WorkerAuthStorage == nil {
+			return fmt.Errorf("No worker auth KMS specified and no worker auth storage found")
+		}
+		c.InfoKeys = append(c.InfoKeys, "worker auth storage path")
+		c.Info["worker auth storage path"] = c.worker.WorkerAuthStorage.BaseDir()
 	}
 
 	if err := c.worker.Start(); err != nil {
--- a/internal/daemon/worker/testing.go
+++ b/internal/daemon/worker/testing.go
@@ -32,7 +32,7 @@ type TestWorker struct {
 	name   string
 }
 
-// Worker returns the underlying controller
+// Worker returns the underlying worker
 func (tw *TestWorker) Worker() *Worker {
 	return tw.w
 }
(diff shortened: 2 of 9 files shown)


internal/servers/repository_worker_test.go
```diff
@@ -663,7 +663,7 @@ func TestRepository_CreateWorker(t *testing.T) {
 			reader: rw,
 			fetchReq: func() *types.FetchNodeCredentialsRequest {
 				// This happens on the worker
-				fileStorage, err := file.NewFileStorage(testCtx)
+				fileStorage, err := file.New(testCtx)
```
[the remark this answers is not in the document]
jimlambrt: btw, the overal devexperience seems quite good in the nodee framework, this is just being picky.

internal/daemon/worker/worker.go
```diff
@@ -75,7 +75,7 @@ type Worker struct {
 	updateTags *ua.Bool
 
 	// The storage for node enrollment
-	WorkerAuthStorage             nodeenrollment.Storage
+	WorkerAuthStorage             *nodeefile.FileStorage
 	WorkerAuthCurrentKeyId        string
 	WorkerAuthRegistrationRequest string
 	workerAuthSplitListener       *splitlistener.SplitListener
```
jimlambrt, line 81: I know, it's out of scope of this PR, but can we fix this studder?  perhaps change the package name to `net`?

internal/daemon/worker/worker.go
```diff
@@ -188,26 +188,81 @@ func (w *Worker) Start() error {
...
+
+	default:
+		// Some other type of error happened, bail out
+		return fmt.Errorf("error loading worker auth creds: %w", err)
 	}
-	w.WorkerAuthRegistrationRequest = base58.FastBase58Encoding(reqBytes)
-	if err != nil {
-		return fmt.Errorf("error encoding registration request: %w", err)
+
+	if createNodeAuthCreds {
```
jimlambrt, line 238: mind adding a comment that this MUST happen before `if createFetchRequest`... it's great you have a test for that conditional as well btw.

internal/cmd/commands/server/server.go
```diff
@@ -463,22 +463,10 @@ func (c *Command) Run(args []string) int {
...
-					retErr = fmt.Errorf("Error shutting down worker: %w", err)
-				}
-				c.UI.Error(retErr.Error())
-				if err := c.controller.Shutdown(); err != nil {
-					c.UI.Error(fmt.Errorf("Error with controller shutdown: %w", err).Error())
-				}
-				return base.CommandCliError
-			}
+
+		if c.WorkerAuthKms == nil && c.worker.WorkerAuthRegistrationRequest != "" {
```
jimlambrt, line 467: excellent!

internal/daemon/worker/worker.go
```diff
@@ -188,26 +188,81 @@ func (w *Worker) Start() error {
...
-		return fmt.Errorf("error generating new node creds: %w", err)
-	}
+	// Check for worker auth creds and if not create new ones
+	var createNodeAuthCreds bool
+	var createFetchRequest bool
+	nodeCreds, err := types.LoadNodeCredentials(w.baseContext, w.WorkerAuthStorage, nodeenrollment.CurrentId, nodeenrollment.WithWrapper(w.conf.WorkerAuthStorageKms))
+	switch {
+	case err == nil:
+		if nodeCreds == nil {
+			event.WriteSysEvent(w.baseContext, op, "no error loading worker auth creds but nil creds, creating new creds for registration")
```
jimlambrt, line 198: are you seeing this event?  just curious if you manually tested it -> no test coverage :-)  BTW, the test coverage is great overall for these new bits.  so hard to test failure modes.

internal/daemon/worker/testing.go
```diff
@@ -177,6 +177,9 @@ type TestWorkerOpts struct {
 	// The worker credential storage KMS to use, or one will be created
 	WorkerAuthStorageKms wrapping.Wrapper
 
+	// The location of the worker's auth storage
+	WorkerAuthStoragePath string
```
jimlambrt, line 181: I can't figure out how this is used (if it is) anywhere?

jimlambrt (approved):
looks great overall.  I had a few minor questions (some of them for the upstream nodee framework).  There is one bit about the `WorkerAuthStoragePath` that seemed out of place but perhaps I was just missing something.

internal/cmd/commands/server/server.go
jefferai: 😸

internal/daemon/worker/testing.go
jefferai: I think I meant to use it and then didn't end up doing so; I'll take it out for now

jefferai: Wait no it's used in the new test!

internal/daemon/worker/worker.go
jefferai: Yeah -- I'll do that and merge in here before I merge this

jefferai: Done

internal/servers/repository_worker_test.go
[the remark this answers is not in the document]
jefferai: Done...and thanks!

internal/daemon/worker/worker.go
jefferai: I'm not, no. I don't even know how it'd happen given that it should never return nil and a non-nil error so this is just to prevent panic in some absolutely bonkers case.

internal/daemon/worker/worker.go
jefferai: Done!

PR #2143, opened 2022-06-06
feature (worker): add CreateWorker(...) service (aka API)

jimlambrt:

PR #2145, opened 2022-06-07
Add active session count to the worker

talanknight:
This counts all sessions which have a connection w/o a closed reason for the worker.


internal/db/schema/migrations/oss/postgres/99/04_views.up.sql
```diff
@@ -0,0 +1,73 @@
...
+  sc.bytes_up, sc.bytes_down, sc.closed_reason
+from session s
+  join session_state ss on
+    s.public_id = ss.session_id
+  left join session_connection sc on
+    s.public_id = sc.session_id;
+
+-- Drop the server and server_type_enm tables
+drop table server;
+drop table server_type_enm;
```
irenarindos, line 71: Should these two statements be in the views file? Maybe move them and drop the old views in 02

internal/proto/controller/api/resources/workers/v1/worker.proto
```diff
@@ -90,6 +90,9 @@ message Worker {
   // itself.
   WorkerConfig worker_config = 140 [json_name = "worker_config"];
 
+  // Output only. The number of sessions that this worker is currently handling.
+  uint32 active_session_count = 150 [json_name = "active_session_count"]; // @gotags: `class:"public"`
```
irenarindos, line 94: You might want to coordinate with Jim on the number here- he's also set 150: https://github.com/hashicorp/boundary/pull/2143/files#diff-021096cfbf64554a86d2acfaeff08ec6399d84eda7954d84224c1e535380c4a8

internal/db/schema/migrations/oss/postgres/99/04_views.up.sql
```diff
@@ -0,0 +1,73 @@
...
+with worker_config_tags(worker_id, source, tags) as (
+  select
+    ct.worker_id,
+    ct.source,
+    -- keys and tags can be any lowercase printable character so use uppercase characters as delimitors.
+    string_agg(distinct concat_ws('Y', ct.key, ct.value), 'Z') as tags
+  from server_worker_tag ct
+  group by ct.worker_id, ct.source
+),
+ session_count (worker_id, count) as (
```
irenarindos, line 15: A session can only be terminated if all its connections are closed. If we added back the `worker_id` column to the session table, this could be a more straightforward query based on whether or not `termination_reason` has been set for the session.

irenarindos: Looks good to me!

PR #2146, opened 2022-06-07
feature(worker): add workerKeyId to status

irenarindos:
- Adds worker key id to cluster status requests.
- Removes id generation based off name. On upsert, if a KMS worker isn't found, an id will be generated for it. 
- Addition of worker lookup based on keyId


internal/servers/repository_worker.go
```diff
@@ -67,17 +67,72 @@ func (r *Repository) LookupWorkerByWorkerReportedName(ctx context.Context, name
...
+		return "", errors.Wrap(ctx, err, op)
+	}
+
+	return worker.WorkerId, nil
+}
+
+func (r *Repository) lookupWorkerId(ctx context.Context, name string, keyId string) (string, error) {
+	const op = "servers.(Repository).lookupWorkerId"
+
+	if name == "" && keyId == "" {
```
jefferai, line 109: Nit: I think this function could be a bit clearer as a switch.

jefferai (approved):

internal/daemon/cluster/handlers/worker_service.go
```diff
@@ -85,10 +85,13 @@ func (ws *workerServiceServer) Status(ctx context.Context, req *pbs.StatusReques
 		})
 	}
 
+	wKeyId := wStat.GetKeyId()
```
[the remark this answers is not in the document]
irenarindos: Thanks (and thanks for all the other review comments!)

jefferai (approved):

internal/daemon/worker/status.go
```diff
@@ -144,12 +144,15 @@ func (w *Worker) sendWorkerStatus(cancelCtx context.Context) {
 	statusCtx, statusCancel := context.WithTimeout(cancelCtx, common.StatusTimeout)
 	defer statusCancel()
 
+	keyId := w.WorkerAuthCurrentKeyId.Load()
```
jimlambrt, line 147: should we check if keyId is empty?

internal/daemon/cluster/handlers/worker_service.go
```diff
@@ -61,8 +61,8 @@ func (ws *workerServiceServer) Status(ctx context.Context, req *pbs.StatusReques
 
 	wStat := req.GetWorkerStatus()
 	switch {
-	case wStat.GetName() == "":
```
jimlambrt, line 64: find fixing this possible panic: https://github.com/hashicorp/boundary/blob/88ede3f9e5d4e4d81d242871715878e6b5011f80/internal/daemon/cluster/handlers/worker_service.go#L227

internal/servers/repository_worker.go
```diff
@@ -202,18 +227,35 @@ func (r *Repository) UpsertWorkerStatus(ctx context.Context, worker *Worker, opt
...
+		WithKeyId(worker.WorkerReportedKeyId))
 
 	opts := getOpts(opt...)
 
 	var err error
 	switch {
 	case opts.withPublicId != "":
 		worker.PublicId = opts.withPublicId
+	case worker.WorkerReportedKeyId != "":
+		workerId, err := r.LookupWorkerIdByWorkerReportedKeyId(ctx, worker.WorkerReportedKeyId)
```
jimlambrt, line 240: is this path tested?

internal/daemon/worker/status.go
irenarindos: It could be empty if it's a KMS worker and doesn't have a keyId, so I think this is fine- @jefferai ?

internal/servers/repository_worker.go
irenarindos: TestStatusWorkerWithKeyId covers this; it doesn't show up when I run tests with coverage for some reason, but I've run through debug and it does hit this case.

internal/daemon/cluster/handlers/worker_service.go
irenarindos: Thanks! I've updated the LookupSessionRequest to pass keyId/ added a lookup on that

internal/daemon/worker/status.go
jefferai: This is correct in that it's empty if it's a KMS worker, but there's actually a bug -- not your bug, not my bug, just from two things happening in parallel -- it's also empty if we haven't (a) created a fetch request due to already having credentials and (b) haven't rotated. You probably need to update the logic in worker.go (starting around line 197)  so that if `!createFetchRequest` we derive and populate the key ID anyways.

I also think startListeners should go below that block instead of above -- it is totally fine where it is but it's sort of of a piece with startControllerConnections so I think a bit cleaner for them to be together.

jimlambrt: ty for updaing worker.go -> I'm curious if we should still detect if it's empty, since it's def an error condition now I believe.

internal/daemon/cluster/handlers/worker_service.go
```diff
@@ -61,8 +61,8 @@ func (ws *workerServiceServer) Status(ctx context.Context, req *pbs.StatusReques
 
 	wStat := req.GetWorkerStatus()
```
jimlambrt, line 62: what happens if wStat == nil?

jimlambrt (changes requested):
minor questions

jefferai (approved):
Took a look -- looks good pending addressing Jim's comments

internal/proto/controller/storage/servers/store/v1/worker.proto
```diff
@@ -62,6 +62,10 @@ message Worker {
   // The last_status_time is the last time a worker daemon sent a status update.
   // @inject_tag: `gorm:"default:null"`
   timestamp.v1.Timestamp last_status_time = 110;
+
+  // keyId reported by a PKI worker (optional)
```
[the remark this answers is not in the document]
irenarindos: Updated to Self-Managed worker

internal/daemon/cluster/handlers/worker_service.go
```diff
@@ -221,10 +227,24 @@ func (ws *workerServiceServer) LookupSession(ctx context.Context, req *pbs.Looku
 			event.WriteError(ctx, op, err, event.WithInfoMsg("error getting servers repo"))
 			return &pbs.LookupSessionResponse{}, status.Errorf(codes.Internal, "Error acquiring server repo when looking up session: %v", err)
 		}
-		w, err := serversRepo.LookupWorkerByWorkerReportedName(ctx, req.GetServerId())
-		if err != nil {
-			event.WriteError(ctx, op, err, event.WithInfoMsg("error looking up worker by name", "name", req.ServerId))
-			return &pbs.LookupSessionResponse{}, status.Errorf(codes.Internal, "Error looking up tags for server: %v", err)
+		var w *servers.Worker
+		if req.WorkerKeyId != "" {
```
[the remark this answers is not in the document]
irenarindos: This is tested in TestStatusWorkerWithKeyId

internal/db/schema/migrations/oss/postgres/99/02_worker_controller_tables.up.sql
```diff
@@ -54,19 +54,27 @@ create table server_worker (
...
           or
           (last_status_time is not null and worker_reported_name is not null)
+          or
+          (last_status_time is not null and worker_reported_key_id is not null)
         )
     constraint worker_reported_name_must_be_lowercase
       check (lower(trim(worker_reported_name)) = worker_reported_name)
     constraint worker_reported_name_only_has_printable_characters
       check (worker_reported_name !~ '[^[:print:]]'),
+  worker_reported_key_id text
```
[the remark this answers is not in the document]
irenarindos: I've removed this, I had it so there would be errors on create/ update if name was null and worker key id was also null. I've moved those tests to the repo upsert tests.

PR #2148, opened 2022-06-07
Require default ports to be defined for newly created tcp targets.

talanknight:


jefferai:
I think the prime functional change looks fine, but sync up with Tim on the interface bit.

louisruch (approved):
LGTM

tmessi (approved):

PR #2152, opened 2022-06-08
feat(worker): save workerAuth request to a file

irenarindos:
A WorkerAuthRequest is stored to a file on creation and removed after succesful auth


internal/cmd/base/servers.go
```diff
@@ -342,6 +345,29 @@ func (b *Server) RemovePidFile(pidPath string) error {
 	return os.Remove(pidPath)
 }
 
+func (b *Server) StoreWorkerAuthReq(authReq, workerAuthReqPath string) error {
+	// Quit fast if no workerAuthReqFile
+	if workerAuthReqPath == "" {
+		return nil
+	}
+	workerAuthReqFilePath := fmt.Sprintf("%s/%s", workerAuthReqPath, WorkerAuthReqFile)
```
jefferai, line 353: Use `[filepath.Join](https://pkg.go.dev/path/filepath#Join)` as this is safe across OSes.

internal/cmd/base/servers.go
```diff
@@ -342,6 +345,29 @@ func (b *Server) RemovePidFile(pidPath string) error {
...
+
+	// Open the workerAuthReq file
+	workerAuthReqFile, err := os.OpenFile(workerAuthReqFilePath, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0o644)
```
jefferai, line 356: I think `0o640` is probably fine. The request isn't secret, but some might consider it sensitive.

internal/cmd/base/servers.go
```diff
@@ -342,6 +345,29 @@ func (b *Server) RemovePidFile(pidPath string) error {
...
+	if err != nil {
+		return fmt.Errorf("could not open workerAuthReq file: %w", err)
+	}
+	defer workerAuthReqFile.Close()
+
+	// Write out the workerAuthRequest
+	_, err = workerAuthReqFile.WriteString(fmt.Sprintf("%s", authReq))
```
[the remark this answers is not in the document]
irenarindos: woops! I initially also had it printing the keyId on a newline.

internal/cmd/commands/dev/dev.go
```diff
@@ -718,6 +718,10 @@ func (c *Command) Run(args []string) int {
 			c.Info["worker auth current key id"] = c.worker.WorkerAuthCurrentKeyId
 			c.InfoKeys = append(c.InfoKeys, "worker auth storage path")
 			c.Info["worker auth storage path"] = c.worker.WorkerAuthStorage.BaseDir()
+			if err := c.StoreWorkerAuthReq(c.worker.WorkerAuthRegistrationRequest, c.worker.WorkerAuthStorage.BaseDir()); err != nil {
+				c.UI.Error(fmt.Errorf("Error storing workerAuth request: %w", err).Error())
+				return base.CommandUserError
```
jefferai, line 723: This gets a little ugly. See above how if Start fails we shut down the worker? If we're going to exit here we should really do a proper shutdown of the worker and controller.

I was going to suggest doing something like `errorEncountered.Store(true)` as in the goroutine but in fact that logic is already faulty. If there _is_ an error we skip out of the for loop and return command success 😬 If you're up for refactoring it, I think what needs to happen is that you could output the error on the UI (as in the goroutine below) and set encountered error here to true and continue; that will cause the `for` loop below not to run; and everything in the for loop in the `if shutdownTriggered` bit needs to go after the `for` loop, and then whether success or error is returned as the return code depends on if errorEncountered was true or not.

internal/cmd/commands/server/server.go
```diff
@@ -469,6 +469,12 @@ func (c *Command) Run(args []string) int {
 			c.Info["worker auth registration request"] = c.worker.WorkerAuthRegistrationRequest
 			c.InfoKeys = append(c.InfoKeys, "worker auth current key id")
 			c.Info["worker auth current key id"] = c.worker.WorkerAuthCurrentKeyId
+
+			// Write the WorkerAuth request to a file
+			if err := c.StoreWorkerAuthReq(c.worker.WorkerAuthRegistrationRequest, c.worker.WorkerAuthStorage.BaseDir()); err != nil {
+				c.UI.Error(fmt.Errorf("Error storing workerAuth request: %w", err).Error())
```
jefferai, line 475: Same comment

internal/daemon/worker/controller_connection.go
```diff
@@ -77,6 +78,14 @@ func (w *Worker) controllerDialerFunc() func(context.Context, string) (net.Conn,
 			conn, err = w.v1KmsAuthDialFn(ctx, addr)
 		default:
 			conn, err = protocol.Dial(ctx, w.WorkerAuthStorage, addr, nodeenrollment.WithWrapper(w.conf.WorkerAuthStorageKms))
+			// No error and a valid connection means the WorkerAuthRegistrationRequest was populated
+			// We can remove the stored workerAuthRequest file
+			if err == nil && conn != nil {
+				if w.WorkerAuthStorage.BaseDir() != "" {
+					workerAuthReqFilePath := fmt.Sprintf("%s/%s", w.WorkerAuthStorage.BaseDir(), base.WorkerAuthReqFile)
```
jefferai, line 85: Same comment about `filepath.Join`. Also for the os.Remove command I'd so `_ = os.Remove...` to make it clear that this isn't forgotten error handling but we explicitly want to ignore a failure since it isn't really a failure of the system overall (and add a comment about it)

internal/cmd/base/servers.go
jefferai: This got resolved but as there was no comment but also no change it feels like maybe it was resolved by accident. If you want to leave it that's fine, can you just confirm that you want to leave it as-is?

internal/cmd/commands/server/server.go
```diff
@@ -469,6 +469,12 @@ func (c *Command) Run(args []string) int {
 			c.Info["worker auth registration request"] = c.worker.WorkerAuthRegistrationRequest
 			c.InfoKeys = append(c.InfoKeys, "worker auth current key id")
 			c.Info["worker auth current key id"] = c.worker.WorkerAuthCurrentKeyId
+
+			// Write the WorkerAuth request to a file
+			if err := c.StoreWorkerAuthReq(c.worker.WorkerAuthRegistrationRequest, c.worker.WorkerAuthStorage.BaseDir()); err != nil {
+				c.UI.Error(fmt.Errorf("Error storing worker auth request: %w", err).Error())
```
jefferai, line 475: This needs the same treatment as in dev.go -- if it fails, shut down the worker and then the controller before exiting.

jefferai (changes requested):
Two comments, one a fixup, but very close!

internal/cmd/base/servers.go
irenarindos: D'oh my bad, I thought the existing code had the same number you commented. Fixed!

PR #2155, opened 2022-06-09
Add support static credentials in DeleteCredentialSource and SetCredentialSource 

louisruch:
Results from running the new bats tests:

```
bats -p boundary/target_credential_sources.bats
 ✓ boundary/login: can login as admin user
 ✓ boundary/credential-stores: can create test-for-add-credential-sources static store in default project
 ✓ boundary/credentials: can create first-credential credential in test-for-add-credential-sources store
 ✓ boundary/target: can add first-credential credential source
 ✓ boundary/target: validate only first-credential credential source present
 ✓ boundary/target: cannot add duplicate credential source
 ✓ boundary/target: can delete first-credential credential source
 ✓ boundary/target: validate first-credential credential source removed
 ✓ boundary/credentials: can create additional credentials in test-for-add-credential-sources store
 ✓ boundary/target: can add multiple credential sources
 ✓ boundary/target: validate added credential sources present
 ✓ boundary/target: can delete multiple credential sources
 ✓ boundary/target: validate third-credential credential source present
 ✓ boundary/target: validate deleted credential sources not present
 ✓ boundary/target: can set multiple credential sources
 ✓ boundary/target: validate set credential sources present
 ✓ boundary/target: can set just first-credential credential source
 ✓ boundary/target: validate first-credential credential source present
 ✓ boundary/target: validate not set credential sources not present

19 tests, 0 failures
```

### DIFF
--- a/internal/target/query.go
+++ b/internal/target/query.go
@@ -3,46 +3,51 @@ package target
 const (
 	setChangesQuery = `
 with
-set_libraries (library_id) as (
+set_sources (source_id, type) as (
   -- returns the SET list
-  select public_id
-    from credential_library
+  select public_id, type
+    from credential_source_all_types
    where public_id in (%s)
 ),
-current_libraries (library_id) as (
+current_sources (source_id, type) as (
   -- returns the current list
-  select credential_library_id
+  select credential_library_id, 'library'
     from target_credential_library
    where target_id          = @target_id
      and credential_purpose = @purpose
+  union
+  select credential_static_id, 'static'
+    from target_static_credential
+   where target_id          = @target_id
+     and credential_purpose = @purpose
 ),
-keep_libraries (library_id) as (
+keep_sources (source_id) as (
   -- returns the KEEP list
-  select library_id
-    from current_libraries
-   where library_id in (select * from set_libraries)
+  select source_id
+    from current_sources
+   where source_id in (select source_id from set_sources)
 ),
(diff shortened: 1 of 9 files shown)


louisruch:
@tmessi PTAL
@mgaffney If you have the time PTAL as well
@ddebko FYI/PTAL

internal/target/query.go
```diff
@@ -3,46 +3,51 @@ package target
...
-current_libraries (library_id) as (
+current_sources (source_id, type) as (
   -- returns the current list
-  select credential_library_id
+  select credential_library_id, 'library'
     from target_credential_library
    where target_id          = @target_id
      and credential_purpose = @purpose
+  union
+  select credential_static_id, 'static'
```
jefferai, line 19: We have `credential_library_id`, shouldn't we have a similar `credential_id`? For instance it's not `select credential_library_id, 'vault'`, it's `library`. So it feels like this should be `select credential_id, 'credential'`.

tmessi: I think it depends on where this is being selected from. `credential_library_id` is from `target_credential_library`, while `credential_static_id` is from `target_static_credential`, either way they both end up populating `source_id` for the `current_sources` cte.

tmessi (approved):

ddebko (approved):
LGTM. My comments are non-blocking.

PR #2162, opened 2022-06-09
Clarify audit logging file sink doc

johanbrandhorst:
Including the "path" in the file example makes it clear that you need to use both of these.

### DIFF
--- a/website/content/docs/configuration/events/file.mdx
+++ b/website/content/docs/configuration/events/file.mdx
@@ -16,7 +16,8 @@ sink {
     event_types = ["observation"]
     format = "cloudevents-json"
     file {
-      file_name = "file-name"
+      path = "/var/log/boundary"
+      file_name = "events.ndjson"
     }
   }
 ```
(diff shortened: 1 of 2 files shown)


tmessi (commented):
There is another example [here](https://github.com/hashicorp/boundary/blob/3976a50ebba0b50d8eadf9f554f6e79ffc88e04d/website/content/docs/configuration/events/overview.mdx) as well.

Also this is not specific to audit events, but rather to any file sink. And lastly I think technically the `path` is *not* required. Its just that `file_name` can only be the base file name, and if you need to specify a path you need to use the `path` variable.

johanbrandhorst:
I see. Let me update the wording a bit.

tmessi:
I would probably phrase it like this:

> chore: Clarify event file sink doc
>
> Updated examples to show the common use case of
> specifying a directory path and file name for
> a file sink. This should hopefully prevent users
> from making the mistake of putting a full path
> for the `file_name` field and seeing an error like:
>
>     [ERROR] unable to send event: operation="event.(Eventer).retrySend"
>       warning=
>       | 1 error occurred:
>       | 	* mkdir : no such file or directory
>       |

johanbrandhorst:
I updated the commit message

tmessi (approved):

PR #2163, opened 2022-06-09
feat(worker): Map cluster ID in config to upstream address

irenarindos:
- `hcp_boundary_cluster_id` can be set and used as an intial upstream.
- Adds comparison of old vs new upstreams and emits event.
- Testing will be added to this PR.

### DIFF
--- a/internal/cmd/config/config.go
+++ b/internal/cmd/config/config.go
@@ -198,6 +198,9 @@ type Worker struct {
 
 	// AuthStoragePath represents the location a worker stores its node credentials, if set
 	AuthStoragePath string `hcl:"auth_storage_path"`
+
+	// Internal field for use with HCP deployments. Will be used if controllers/ initial_upstreams is not set
+	HCPBClusterId string `hcl:"hcp_boundary_cluster_id"`
 }
 
 func (w *Worker) InitNameIfEmpty() (string, error) {
--- a/internal/cmd/config/config.go
+++ b/internal/cmd/config/config.go
@@ -124,6 +124,9 @@ type Config struct {
 
 	// Plugin-related options
 	Plugins Plugins `hcl:"plugins"`
+
+	// Internal field for use with HCP deployments. Used if controllers/ initial_upstreams is not set
+	HCPBClusterId string `hcl:"hcp_boundary_cluster_id"`
 }
 
 type Controller struct {
@@ -198,9 +201,6 @@ type Worker struct {
 
 	// AuthStoragePath represents the location a worker stores its node credentials, if set
 	AuthStoragePath string `hcl:"auth_storage_path"`
-
-	// Internal field for use with HCP deployments. Will be used if controllers/ initial_upstreams is not set
-	HCPBClusterId string `hcl:"hcp_boundary_cluster_id"`
 }
 
 func (w *Worker) InitNameIfEmpty() (string, error) {
--- a/internal/daemon/worker/controller_connection.go
+++ b/internal/daemon/worker/controller_connection.go
@@ -35,6 +35,8 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+const hcpb_url_suffix = ".proxy.boundary.hashicorp.cloud"
+
 // StartControllerConnections starts up the resolver and initiates controller
 // connection client creation
 func (w *Worker) StartControllerConnections() error {
@@ -57,7 +59,13 @@ func (w *Worker) StartControllerConnections() error {
 	}
 
 	if len(initialAddrs) == 0 {
-		return errors.New("no initial controller addresses found")
+		if w.conf.RawConfig.Worker.HCPBClusterId != "" {
+			clusterAddress := fmt.Sprintf("%s%s", w.conf.RawConfig.Worker.HCPBClusterId, hcpb_url_suffix)
+			initialAddrs = append(initialAddrs, resolver.Address{Addr: clusterAddress})
+			event.WriteSysEvent(context.TODO(), op, fmt.Sprintf("Setting HCPB Cluster address %s as upstream address", clusterAddress))
+		} else {
+			return errors.New("no initial controller addresses found")
+		}
 	}
 
 	w.Resolver().InitialState(resolver.State{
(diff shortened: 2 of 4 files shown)


internal/daemon/worker/controller_connection.go
```diff
@@ -35,6 +35,8 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+const hcpb_url_suffix = ".proxy.boundary.hashicorp.cloud"
```
jefferai, line 38: I realize this is a common const style but I think our normal const style is also camelCase

internal/cmd/config/config.go
```diff
@@ -198,6 +198,9 @@ type Worker struct {
 
 	// AuthStoragePath represents the location a worker stores its node credentials, if set
 	AuthStoragePath string `hcl:"auth_storage_path"`
+
+	// Internal field for use with HCP deployments. Will be used if controllers/ initial_upstreams is not set
+	HCPBClusterId string `hcl:"hcp_boundary_cluster_id"`
```
jefferai, line 203: I think this should be at the top level -- I can imagine that at some point it might be useful for controllers, internally or not. It feels more like an "installation level parameter" vs. specifically a worker parameter.

internal/daemon/worker/status.go
```diff
@@ -3,21 +3,24 @@ package worker
...
 	"github.com/hashicorp/boundary/internal/observability/event"
+	"github.com/hashicorp/boundary/internal/servers"
 	"google.golang.org/grpc/resolver"
 )
 
 type LastStatusInformation struct {
 	*pbs.StatusResponse
-	StatusTime time.Time
+	StatusTime              time.Time
+	LastCalculatedUpstreams []resolver.Address
```
jefferai, line 23: Given that strings go over the RPC, it feels like the logic below could be simplified if this was `[]string` instead.

internal/daemon/worker/status.go
```diff
@@ -191,14 +194,29 @@ func (w *Worker) sendWorkerStatus(cancelCtx context.Context) {
...
 		if len(result.CalculatedUpstreams) > 0 {
-			addrs := make([]resolver.Address, 0, len(result.CalculatedUpstreams))
+			addrs = make([]resolver.Address, 0, len(result.CalculatedUpstreams))
 			for _, v := range result.CalculatedUpstreams {
 				addrs = append(addrs, resolver.Address{Addr: v.Address})
 			}
-			w.Resolver().UpdateState(resolver.State{Addresses: addrs})
+			lastStatus := w.lastStatusSuccess.Load().(*LastStatusInformation)
+			// Compare upstreams; update resolver if there is a difference, and emit an event with old and new addresses
+			if lastStatus != nil && upstreamsHasChanged(lastStatus.LastCalculatedUpstreams, addrs) {
```
jefferai, line 205: If using []string as what's being stored you can get rid of a lot of the logic here and all of `upstreamsHasChanged` by using https://github.com/hashicorp/go-secure-stdlib/blob/main/strutil/strutil.go#L295

jimlambrt (approved):
Looks great!

internal/daemon/worker/status.go
```diff
@@ -191,14 +193,26 @@ func (w *Worker) sendWorkerStatus(cancelCtx context.Context) {
...
+			addrs = make([]resolver.Address, 0, len(result.CalculatedUpstreams))
 			for _, v := range result.CalculatedUpstreams {
 				addrs = append(addrs, resolver.Address{Addr: v.Address})
+				newUpstreams = append(newUpstreams, v.Address)
+			}
+			lastStatus := w.lastStatusSuccess.Load().(*LastStatusInformation)
+			// Compare upstreams; update resolver if there is a difference, and emit an event with old and new addresses
+			if lastStatus != nil && !strutil.EquivalentSlices(lastStatus.LastCalculatedUpstreams, newUpstreams) {
+				upstreamsMessage := fmt.Sprintf("Upstreams has changed; old upstreams were: %s, new upstreams are: %s", lastStatus.LastCalculatedUpstreams, newUpstreams)
+				event.WriteSysEvent(context.TODO(), op, upstreamsMessage)
```
jefferai, line 208: Why use `context.TODO` instead of `cancelCtx` from the input parameters?

internal/daemon/worker/status.go
```diff
@@ -191,14 +193,26 @@ func (w *Worker) sendWorkerStatus(cancelCtx context.Context) {
...
+				w.Resolver().UpdateState(resolver.State{Addresses: addrs})
+			} else if lastStatus == nil {
+				w.Resolver().UpdateState(resolver.State{Addresses: addrs})
+				event.WriteSysEvent(context.TODO(), op, fmt.Sprintf("Upstreams after first status set to: %s", newUpstreams))
```
jefferai, line 212: Same question

internal/daemon/worker/controller_connection.go
```diff
@@ -57,7 +59,13 @@ func (w *Worker) StartControllerConnections() error {
 	}
 
 	if len(initialAddrs) == 0 {
-		return errors.New("no initial controller addresses found")
+		if w.conf.RawConfig.HCPBClusterId != "" {
+			clusterAddress := fmt.Sprintf("%s%s", w.conf.RawConfig.HCPBClusterId, hcpbUrlSuffix)
+			initialAddrs = append(initialAddrs, resolver.Address{Addr: clusterAddress})
+			event.WriteSysEvent(context.TODO(), op, fmt.Sprintf("Setting HCPB Cluster address %s as upstream address", clusterAddress))
```
jefferai, line 65: Could probably use `w.baseCtx`

internal/daemon/worker/status.go
jefferai: Or w.baseCtx, although likely `cancelCtx` is right here since if this function is canceled sending the event up won't matter

internal/daemon/worker/controller_connection.go
irenarindos: Fixed these in https://github.com/hashicorp/boundary/pull/2168

PR #2175, opened 2022-06-13
refact(workers): remove workerKeyId from aggregate

irenarindos:
Addresses comments in https://github.com/hashicorp/boundary/pull/2146; removes worker key id from the aggregate.


internal/db/schema/migrations/oss/postgres/99/02_worker_controller_tables.up.sql
```diff
@@ -60,21 +60,21 @@ create table server_worker (
...
-          (last_status_time is not null and worker_reported_name is not null)
-          or
-          (last_status_time is not null and worker_reported_key_id is not null)
-        ),
+    check (
+      (last_status_time is null and key_id is null)
+      or
+      (last_status_time is not null and worker_reported_name is not null)
+      or
+      (last_status_time is not null and key_id is not null)
```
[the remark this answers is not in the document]
jimlambrt: can you update the constraint name?

internal/db/schema/migrations/oss/postgres/99/02_worker_controller_tables.up.sql
```diff
@@ -83,6 +83,27 @@ create table server_worker (
...
+      raise 'key_id %s does not match a worker auth record', new.key_id;
+    end if;
+  end if;
+  return new;
+end;
+$$ language plpgsql;
+
+create trigger validate_key_id_on_insert before insert on server_worker
+  for each row execute procedure validate_key_id_on_insert();
+
```
jimlambrt, line 106: I think, key_id is immutable and perhaps should be added to the immutable_columns func call for server_worker.  If this is true, then we should add an "update" test for this as well.

internal/servers/worker.go
```diff
@@ -68,7 +69,7 @@ func NewWorkerForStatus(scopeId string, opt ...Option) *Worker {
 			ScopeId:               scopeId,
 			WorkerReportedName:    opts.withName,
```
jimlambrt, line 70: are we allowed to have an opts.withName and opts.withKeyId together?

jimlambrt (changes requested):
still reviewing the tests a bit, but some early suggestions.

internal/db/schema/migrations/oss/postgres/99/02_worker_controller_tables.up.sql
irenarindos: Key_id can change when the worker rotates its creds

internal/db/schema/migrations/oss/postgres/99/02_worker_controller_tables.up.sql
```diff
@@ -83,6 +83,27 @@ create table server_worker (
...
+        worker_key_identifier= new.key_id;
+    if not found then
+      raise 'key_id %s does not match a worker auth record', new.key_id;
+    end if;
+  end if;
+  return new;
+end;
+$$ language plpgsql;
+
+create trigger validate_key_id_on_insert before insert on server_worker
```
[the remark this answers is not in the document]
irenarindos: I've removed the trigger; the key id will be valid for this worker on upsert- the worker could not send a status update if it was invalid, and the worker id lookup is based on keyId

irenarindos:
Closing; we're going in a new direction with KMS and PKI workers.

PR #2176, opened 2022-06-13
feat(config): Add database config options for idle connection management

tmessi:
This adds two new config options for the `database` stanza:

- `max_idle_connections`
- `max_idle_time`

These correspond to settings for the underlying `sql.DB` instance. They
can be used to have finer control over the controller's database
connection pool. For example this can be used to mitigate the high
connection overhead when connecting to postgres by setting a high
`max_idle_connections` and either setting `max_idle_time` to zero or to
a high duration.

Alternatively, if using a connection pool like pgpool-II or pgbouncer,
`max_idle_connections` can be set to 0.

See:
    https://pkg.go.dev/database/sql#DB.SetMaxIdleConns
    https://pkg.go.dev/database/sql#DB.SetConnMaxIdleTime
    https://www.pgpool.net/mediawiki/index.php/Main_Page
    https://www.pgbouncer.org/

### DIFF
--- a/internal/cmd/config/config.go
+++ b/internal/cmd/config/config.go
@@ -215,10 +215,14 @@ type Worker struct {
 }
 
 type Database struct {
-	Url                   string      `hcl:"url"`
-	MigrationUrl          string      `hcl:"migration_url"`
-	MaxOpenConnections    int         `hcl:"-"`
-	MaxOpenConnectionsRaw interface{} `hcl:"max_open_connections"`
+	Url                     string         `hcl:"url"`
+	MigrationUrl            string         `hcl:"migration_url"`
+	MaxOpenConnections      int            `hcl:"-"`
+	MaxOpenConnectionsRaw   interface{}    `hcl:"max_open_connections"`
+	MaxIdleConnections      *int           `hcl:"-"`
+	MaxIdleConnectionsRaw   interface{}    `hcl:"max_idle_connections"`
+	ConnMaxIdleTime         interface{}    `hcl:"max_idle_time"`
+	ConnMaxIdleTimeDuration *time.Duration `hcl:"-"`
 }
 
 type Plugins struct {
@@ -375,7 +379,7 @@ func Parse(d string) (*Config, error) {
 				switch t := result.Controller.Database.MaxOpenConnectionsRaw.(type) {
 				case string:
 					maxOpenConnectionsString, err := parseutil.ParsePath(t)
-					if err != nil {
+					if err != nil && !errors.Is(err, parseutil.ErrNotAUrl) {
 						return nil, fmt.Errorf("Error parsing database max open connections: %w", err)
 					}
 					result.Controller.Database.MaxOpenConnections, err = strconv.Atoi(maxOpenConnectionsString)
@@ -389,6 +393,43 @@ func Parse(d string) (*Config, error) {
 						reflect.TypeOf(t).String())
 				}
 			}
+			if result.Controller.Database.MaxIdleConnectionsRaw != nil {
+				switch t := result.Controller.Database.MaxIdleConnectionsRaw.(type) {
+				case string:
+					maxIdleConnectionsString, err := parseutil.ParsePath(t)
+					if err != nil && !errors.Is(err, parseutil.ErrNotAUrl) {
+						return nil, fmt.Errorf("Error parsing database max idle connections: %w", err)
+					}
+					idleConns, err := strconv.Atoi(maxIdleConnectionsString)
+					if err != nil {
+						return nil, fmt.Errorf("Database max idle connections value is not a uint: %w", err)
+					}
+					result.Controller.Database.MaxIdleConnections = &idleConns
+				case int:
+					result.Controller.Database.MaxIdleConnections = &t
+				default:
+					return nil, fmt.Errorf("Database max idle connections: unsupported type %q",
+						reflect.TypeOf(t).String())
+				}
+			}
+			if result.Controller.Database.ConnMaxIdleTime != nil {
+				switch t := result.Controller.Database.ConnMaxIdleTime.(type) {
+				case string:
+					durationString, err := parseutil.ParsePath(t)
+					if err != nil && !errors.Is(err, parseutil.ErrNotAUrl) {
+						return nil, fmt.Errorf("Error parsing connection max idle time: %w", err)
+					}
+					connMaxIdleTime, err := parseutil.ParseDurationSecond(durationString)
+					if err != nil {
+						return nil, fmt.Errorf("Connection max idle time is not a duration: %w", err)
+					}
+					result.Controller.Database.ConnMaxIdleTimeDuration = &connMaxIdleTime
+				default:
+					return nil, fmt.Errorf("Database connection max idle time: unsupported type %q",
+						reflect.TypeOf(t).String())
+				}
+			}
+
 		}
 	}
 
--- a/internal/db/option.go
+++ b/internal/db/option.go
@@ -3,6 +3,7 @@ package db
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/hashicorp/boundary/internal/errors"
 	"github.com/hashicorp/boundary/internal/oplog"
@@ -166,8 +167,10 @@ type Options struct {
 	// withPrngValues is used to switch the ID generation to a pseudo-random mode
 	withPrngValues []string
 
-	withGormFormatter      hclog.Logger
-	withMaxOpenConnections int
+	withGormFormatter           hclog.Logger
+	withMaxOpenConnections      int
+	withMaxIdleConnections      *int
+	withConnMaxIdleTimeDuration *time.Duration
 
 	// withDebug indicates that the given operation should invoke Gorm's debug
 	// mode
@@ -189,11 +192,13 @@ func getDefaultOptions() Options {
 			wrapper:  nil,
 			metadata: oplog.Metadata{},
 		},
-		withLookup:         false,
-		WithFieldMaskPaths: []string{},
-		WithNullPaths:      []string{},
-		WithLimit:          0,
-		WithVersion:        nil,
+		withLookup:                  false,
+		WithFieldMaskPaths:          []string{},
+		WithNullPaths:               []string{},
+		WithLimit:                   0,
+		WithVersion:                 nil,
+		withMaxIdleConnections:      nil,
+		withConnMaxIdleTimeDuration: nil,
 	}
 }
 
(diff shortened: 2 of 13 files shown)


internal/cmd/config/config.go
```diff
@@ -379,6 +383,36 @@ func Parse(d string) (*Config, error) {
...
+					}
+					result.Controller.Database.MaxIdleConnections, err = strconv.Atoi(maxIdleConnectionsString)
+					if err != nil {
+						return nil, fmt.Errorf("Database max idle connections value is not an int: %w", err)
+					}
+				case int:
+					result.Controller.Database.MaxIdleConnections = t
+				default:
+					return nil, fmt.Errorf("Database max idle connections: unsupported type %q",
+						reflect.TypeOf(t).String())
```
johanbrandhorst, line 401: Can we just use `%T`? https://pkg.go.dev/fmt#hdr-Printing

internal/cmd/config/config.go
```diff
@@ -379,6 +383,36 @@ func Parse(d string) (*Config, error) {
...
+				}
+			} else {
+				result.Controller.Database.MaxIdleConnections = -1
```
johanbrandhorst, line 404: I don't know if this is a well established pattern (using -1 to indicate "unset"), but if it's not, what do you think about using a pointer instead? Then we could make the type unsigned too.

internal/db/option.go
```diff
@@ -306,14 +311,32 @@ func WithGormFormatter(l hclog.Logger) Option {
 	}
 }
 
-// WithMaxOpenConnections specifices and optional max open connections for the
+// WithMaxOpenConnections specifies and optional max open connections for the
```
johanbrandhorst, line 314: 😅 
```suggestion
// WithMaxOpenConnections specifies an optional max open connections for the
```

internal/db/option.go
```diff
@@ -306,14 +311,32 @@ func WithGormFormatter(l hclog.Logger) Option {
...
 // database
 func WithMaxOpenConnections(max int) Option {
 	return func(o *Options) {
 		o.withMaxOpenConnections = max
 	}
 }
 
+// WithMaxIdleConnections specifies and optional max idle connections for the
```
johanbrandhorst, line 322:
```suggestion
// WithMaxIdleConnections specifies an optional max idle connections for the
```

internal/db/option.go
```diff
@@ -306,14 +311,32 @@ func WithGormFormatter(l hclog.Logger) Option {
...
+// database.
+// This corresponds with: https://pkg.go.dev/database/sql#DB.SetMaxIdleConns
+func WithMaxIdleConnections(max int) Option {
+	return func(o *Options) {
+		o.withMaxIdleConnections = max
+	}
+}
+
+// WithConnMaxIdleTimeDuration specifies and optional connection max idle time
```
johanbrandhorst, line 331:
```suggestion
// WithConnMaxIdleTimeDuration specifies an optional connection max idle time
```

website/content/docs/configuration/controller.mdx
```diff
@@ -35,18 +35,35 @@ description will be read.
...
+  - `max_idle_connections` - Can be used to control the maximum number of
+    idle connections in the idle connection pool.
+    If `max_open_connections` is greater than 0 but less than
+    `max_idle_connections`, then `max_idle_connections` will be reduced to match the `max_open_connections` limit. 
+    Setting this value to 0 will mean that no idle connections are retained.
+    If not set or set to less than 0, the default
+    [sql.DB](https://pkg.go.dev/database/sql#DB.SetMaxIdleConns) setting will be used.
+    This value can be a string representing the max number of connections, can refer to a file
+    on disk (file://) from which the number of connections will be read; or an env var (env://) from which the
+    number of connections will be read.
```
johanbrandhorst, line 60: It doesn't look to me like `parseutil.ParsePath` will support a verbatim `"5"`, will it? Do we want
```suggestion
    This value can be an integer representing the max number of connections, can refer to a file
    on disk (file://) from which the number of connections will be read; or an env var (env://) from which the
    number of connections will be read.
```

tmessi: This phrasing was taken from what already existed for `max_open_connections`. I am not sure if `an integer` is more accurate phrasing, but it should be consistent between this and hte `max_open_connections` since they both leverage the same logic.

internal/cmd/config/config.go
tmessi: Yea I am not sure. I don't think we have done `-1` in the past. I thought about using a pointer but it felt cumbersome, but maybe it is cleaner. The main thing I want to avoid is accidentally setting this to `0` since that is different from the default value in `sql.DB`.

johanbrandhorst: I think if it's all the same to you I'd prefer a pointer, but it seems something that maybe @jefferai or @jimlambrt might have opinions about too 😁.

website/content/docs/configuration/controller.mdx
```diff
@@ -35,18 +35,35 @@ description will be read.
...
-  Either URL can refer to a file on disk (file://) from which a URL will be read; an env
-  var (env://) from which the URL will be read; or a direct database URL (postgres://).
+  - `max_idle_connections` - Can be used to control the maximum number of
+    idle connections in the idle connection pool.
+    If `max_open_connections` is greater than 0 but less than
+    `max_idle_connections`, then `max_idle_connections` will be reduced to match the `max_open_connections` limit. 
+    Setting this value to 0 will mean that no idle connections are retained.
+    If not set or set to less than 0, the default
+    [sql.DB](https://pkg.go.dev/database/sql#DB.SetMaxIdleConns) setting will be used.
+    This value an integer representing the max number of connections, can refer to a file
```
johanbrandhorst, line 58:
```suggestion
    This value can be an integer representing the max number of connections, can refer to a file
```

johanbrandhorst (changes requested):

johanbrandhorst (changes requested):
I think this PR is missing a step in `init.go` where we assign from the Config to the Server? See https://github.com/hashicorp/boundary/blob/b6ca61a7f62d01183f47f7fbff761679b7522b54/internal/cmd/commands/database/init.go#L242

internal/db/option.go
```diff
@@ -189,11 +192,13 @@ func getDefaultOptions() Options {
...
-		WithNullPaths:      []string{},
-		WithLimit:          0,
-		WithVersion:        nil,
+		withLookup:                  false,
+		WithFieldMaskPaths:          []string{},
+		WithNullPaths:               []string{},
+		WithLimit:                   0,
+		WithVersion:                 nil,
+		withMaxIdleConnections:      -1,
+		withConnMaxIdleTimeDuration: -1,
```
johanbrandhorst, line 201: Do we want to use the same pattern here 😬?

internal/cmd/config/config.go
tmessi: I put this back after some tests failed expecting a quoted string around the type here. I think using `%q` looks nicer compared to `\"%T\"`

hugoamvieira (approved):
LGTM!

hugoamvieira:
Q: Some of the behaviour mentioned in the docs is not present in this logic - I assumed this is because we're just passing the value along and some other downstream SQL code is the bit that has this behaviour. Is this a correct assumption?

johanbrandhorst (dismissed):
Nice!

website/content/docs/configuration/controller.mdx
```diff
@@ -35,18 +35,35 @@ description will be read.
...
   - `migration_url` - Can be used to specify a different URL for migrations, as that
     usually requires higher privileges.
+    This value can refer to a file on disk (file://) from which a URL will be read; an env
+    var (env://) from which the URL will be read; or a direct database URL (postgres://).
   - `max_open_connections` - Can be used to control the maximum number of
     connections that can be opened by the controller.
     The minimum number of connections required is 5.
     Setting this value to 0 will allow the controller to open as many
-    connections as needed. This value can be a string representing the max number of connections, can refer to a file
+    connections as needed. This value can be an integer representing the max number of connections, can refer to a file
```
jefferai, line 48: This went from something not quite correct to something else not quite correct :-)  This can be a string or integer in the config file which is typed; it's a string for env var or file since those are always strings. Hence string was put here. But probably it should be "string or integer"

website/content/docs/configuration/controller.mdx
```diff
@@ -35,18 +35,35 @@ description will be read.
...
-  Either URL can refer to a file on disk (file://) from which a URL will be read; an env
-  var (env://) from which the URL will be read; or a direct database URL (postgres://).
+  - `max_idle_connections` - Can be used to control the maximum number of
+    idle connections in the idle connection pool.
+    If `max_open_connections` is greater than 0 but less than
+    `max_idle_connections`, then `max_idle_connections` will be reduced to match the `max_open_connections` limit. 
+    Setting this value to 0 will mean that no idle connections are retained.
+    If not set or set to less than 0, the default
+    [sql.DB](https://pkg.go.dev/database/sql#DB.SetMaxIdleConns) setting will be used.
+    This value can be an integer representing the max number of connections, can refer to a file
```
jefferai, line 58: Same comment here and below

website/content/docs/configuration/controller.mdx
tmessi: I reworded this and this for `max_open_connections` and `max_idle_connections`. Also I added some similar wording for `max_idle_time`. Please take another look,

internal/cmd/config/config.go
```diff
@@ -412,12 +412,19 @@ func Parse(d string) (*Config, error) {
...
 				}
 			}
-			if result.Controller.Database.ConnMaxIdleTime != "" {
-				t, err := parseutil.ParseDurationSecond(result.Controller.Database.ConnMaxIdleTime)
-				if err != nil {
-					return nil, err
+			if result.Controller.Database.ConnMaxIdleTime != nil {
+				switch t := result.Controller.Database.ConnMaxIdleTime.(type) {
+				case string:
+					durationString, err := parseutil.ParsePath(t)
```
jefferai, line 418: You're swallowing the error here.

jefferai: Here, and in both stanzas above (the one you didn't modify contains a bug, the same one that was copypasta'd here), if the value passed in is not a URL, ParsePath returns the original value and a known error. Look at line 337 for a correct usage of ParsePath.

website/content/docs/configuration/controller.mdx
```diff
@@ -44,26 +44,32 @@ description will be read.
   - `max_open_connections` - Can be used to control the maximum number of
     connections that can be opened by the controller.
     The minimum number of connections required is 5.
-    Setting this value to 0 will allow the controller to open as many
-    connections as needed. This value can be an integer representing the max number of connections, can refer to a file
-    on disk (file://) from which the number of connections will be read; or an env var (env://) from which the
-    number of connections will be read.
+    Setting this value to 0 will allow the controller to open as many connections as needed.
+    This value can be an integer representing the max number of connections,
+    or a string that can refer to a file
```
jefferai, line 49: It can also be a bare string, once you fix the ParsePath invocation :-)

internal/cmd/config/config.go
tmessi: Updated, I think I caught all the correct spots. I added some more tests to explicitly test these raw string cases, although interestingly the tests passed without needing to make a change. I think because when `parseutil.ParsePath` calls `url.Parse`, it does not get an error back from and ends up at the final return of this block:

https://github.com/hashicorp/go-secure-stdlib/blob/bf6d78ef5b727b83b2f8f41b473eb9764b446df4/parseutil/parsepath.go#L30-L455

See: https://go.dev/play/p/JDkw7nm1pVF

Still probably a good idea to be defensive and consistent in the usage of `parseutil.ParsePath`, but difficult to cover in a test.

jefferai (dismissed):

PR #2180, opened 2022-06-14
docs(workers): Add self-managed workers to website

irenarindos:
- Added Self-Managed worker configuration to the documentation
- Created separate pages for KMS and Self-Managed workers
- Updated Worker overview page to explain the difference between the two worker types

### DIFF
--- a/website/content/docs/oss/configuration/worker/kms-worker.mdx
+++ b/website/content/docs/oss/configuration/worker/kms-worker.mdx
@@ -1,0 +1,77 @@
+---
+layout: docs
+page_title: KMS Worker Configuration
+description: |-
+  The worker stanza configures worker-specific parameters.
+---
+
+
+## KMS Worker Configuration
+
+KMS Workers require a KMS block designated for `worker-auth`. This is the KMS configuration for
+authentication between the workers and controllers and must be present. Example (not safe for production!):
+
+```hcl
+  kms "aead" {
+	purpose = "worker-auth"
+	aead_type = "aes-gcm"
+	key = "8fZBjCUfN0TzjEGLQldGY4+iE9AkOvCfjh7+p0GtRBQ="
+	key_id = "global_worker-auth"
+}
+```
+
+This configuration must be the same for the worker-auth configuration for the controller if you're
+running the controller and worker as separate servers.
+
+And optionally, a KMS stanza for configuration encryption purpose:
+
+```hcl
+# Configuration encryption block: decrypts sensitive values in the
+# configuration file. See `boundary config [encrypt|decrypt] -h`.
+kms "aead" {
+  purpose   = "config"`
+  aead_type = "aes-gcm"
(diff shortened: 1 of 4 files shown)


website/content/docs/oss/configuration/worker/kms-worker.mdx
```diff
@@ -8,7 +8,19 @@ description: |-
 
 ## KMS Worker Configuration
 
-KMS Workers require a KMS block designated for `worker-auth`. This is the KMS configuration for
+KMS Workers require a `name` field. This specifies a unique name of this worker within the Boundary cluster. This value
+can be a direct name string, can refer to a file on disk (file://) from which an name will be read; or an env
+var (env://) from which the name will be read.
```
stellarsquall, line 13: this might be a little cleaner as a list, something like:

```suggestion
KMS Workers require a `name` field. This specifies a unique name of this worker within the Boundary cluster. 

The `name` value can be:
- a direct name string
- a reference to a file on disk (`file://`) from which the name is read
- an env var (`env://`) from which the name is read.
```

stellarsquall (changes requested):
this looks great! A quick comment on the `name` attribute is all I noticed

website/content/docs/oss/configuration/worker/kms-worker.mdx
irenarindos: Thanks- I've updated this!

stellarsquall (approved):
looks great to me 👍

jimlambrt (approved):
just the description bit that Todd mentioned.

website/content/docs/oss/configuration/worker/self-managed-worker.mdx
```diff
@@ -0,0 +1,56 @@
...
+Example (not safe for production!):
+
+```hcl
+worker {
+  auth_storage_path="/tmp/boundary/demo-worker-1"
+  initial_upstreams = ["10.0.0.1"]
+}
+```
+
+__Note__: `name` and `description` fields are not valid for Self-Managed workers. These fields are only valid for
```
jefferai, line 23: Might be good to make it clear that they can still be set, just through the API -- they're not valid in the config for these workers.

irenarindos: Updated to:
__Note__: `name` and `description` fields are not valid config fields for Self-Managed workers. These fields are only valid for [KMS Workers][]. `name` and `description` can only be set for Self-Managed workers through the API.

website/content/docs/oss/configuration/worker/overview.mdx
```diff
@@ -0,0 +1,60 @@
...
+- [Self-Managed Workers][] use certificates issued by Boundary to authenticate with controllers
+
+Different worker types have different configuration requirements, but share the common worker parameters listed below.
+
+## Common Worker Parameters
+The following fields are supported for all worker types.
+
+```hcl
+worker {
+  description = "An example worker"
```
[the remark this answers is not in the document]
irenarindos: done, thanks!

PR #2184, opened 2022-06-14
Reduce log noise for tests

tmessi:
This introduces two changes to reduce the output of tests in CI and for
local developers:

- Silences gorm db logs
- Silences log indicating the lack of an eventer


mgaffney (approved):

internal/observability/event/context.go
```diff
@@ -129,7 +130,9 @@ func WriteError(ctx context.Context, caller Op, e error, opt ...Option) {
 		eventer = SysEventer()
 		if eventer == nil {
 			logger := hclog.New(nil)
-			logger.Error(fmt.Sprintf("%s: no eventer available to write error: %v", op, e))
+			if os.Getenv("CI") == "" {
```
jefferai, line 133: As this can be in the critical path, a very minor nit you can ignore: set a variable to capture whether this env var is set in an `init()` function on the package. Then each check here becomes a bool check instead of an environment lookup.

tmessi: Hopefully for production deployments it is rare that it makes it past the `eventer == nil` check. But it is simple enough to do a check in `init` see [3680a69](https://github.com/hashicorp/boundary/pull/2184/commits/3680a69bd4a7d972870e689e56e8b4110d9c9382). I welcome opinions on which is more readable, I am fine with either.

jimlambrt (changes requested):

jimlambrt (approved):

PR #2187, opened 2022-06-14
Upsert creates a new KMS and Updates PKI workers

talanknight:


internal/daemon/controller/handlers/workers/worker_service.go
```diff
@@ -457,7 +457,7 @@ func toProto(ctx context.Context, in *servers.Worker, opt ...handlers.Option) (*
 		out.AuthorizedActions = opts.WithAuthorizedActions
 		if in.Type == KmsWorkerType {
 			// KMS workers cannot be updated through the API
-			out.AuthorizedActions = strutil.StrListDelete(action.Update.String())
+			out.AuthorizedActions = strutil.StrListDelete(out.AuthorizedActions, action.Update.String())
```
jefferai, line 460: Lol whoops

internal/servers/public_ids.go
```diff
@@ -19,3 +19,15 @@ func newWorkerId(ctx context.Context) (string, error) {
 	}
 	return id, nil
 }
+
+// newWorkerIdFromScopeAndName generates a predictable public id based on the
+// scope and the worker name.  This should only be used on kms workers at in
+// upsert.
```
jefferai, line 25:
```suggestion
// scope and the worker name.  This should only be used on kms workers at
// upsert time.
```

jefferai (approved):

PR #2191, opened 2022-06-15
feat(cli): Support credential brokering with sshpass

louisruch:
### What does this PR do

Adds support for brokered credentials from the static credential store for postgres
Adds new `sshpass` style for brokering username_password credentials for ssh:

https://user-images.githubusercontent.com/6704214/173732755-d7f5fec0-5853-4731-b261-e1b459f2b221.mp4

### DIFF
--- a/internal/cmd/commands/connect/credentials.go
+++ b/internal/cmd/commands/connect/credentials.go
@@ -1,0 +1,61 @@
+package connect
+
+import (
+	"errors"
+
+	"github.com/hashicorp/boundary/api/targets"
+	"github.com/hashicorp/boundary/internal/credential"
+	"github.com/mitchellh/mapstructure"
+)
+
+type usernamePasswordCredentials struct {
+	Username string `mapstructure:"username"`
+	Password string `mapstructure:"password"`
+}
+
+// TODO: this should support multiple types of credentials and return an interface instead
+// of only usernamepassword. Each subcommand can use the returned credentials as it needs.
+func parseCredentials(creds []*targets.SessionCredential) ([]usernamePasswordCredentials, error) {
+	if creds == nil {
+		return nil, nil
+	}
+	var out []usernamePasswordCredentials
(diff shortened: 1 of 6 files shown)


internal/cmd/commands/connect/ssh.go
```diff
@@ -41,18 +42,49 @@ func (s *sshFlags) defaultExec() string {
...
+		}
+		// sshpass requires that the password is passed as env-var "SSHPASS"
+		// when the using env-vars.
+		envs = append(envs, fmt.Sprintf("SSHPASS=%s", cred.Password))
+		args = append(args, "-e", "ssh")
+		args = append(args, "-p", port, ip)
+
+		// sshpass cannot handle host key checking, disable localhost key verification
+		// to avoid error: 'SSHPASS detected host authentication prompt. Exiting.'
+		args = append(args, "-o", "NoHostAuthenticationForLocalhost=yes")
```
jefferai, line 76: I wonder if we should always turn this on, not just for sshpass. Right now we set the host name of the local connection to the host ID in order to not keep seeing host authentication mismatches, but adding this would help too.

internal/cmd/commands/connect/credentials.go
```diff
@@ -0,0 +1,62 @@
...
+		return nil, nil
+	}
+	var out []usernamePasswordCredentials
+	for _, cred := range creds {
+		if cred.CredentialSource == nil {
+			return nil, errors.New("missing credential source")
+		}
+
+		if cred.CredentialSource.CredentialType == string(credential.UsernamePasswordType) {
+			username, ok := cred.Credential["username"].(string)
```
jefferai, line 29: Given you have `mapstructure` tags, why not just `mapstructure.Decode` into the struct? Then you don't need to do the validity checking on the map.

jefferai (approved):
Looks good, just a couple of comments!

PR #2194, opened 2022-06-15
fix (byow): minor updates to UpsertWorkerStatus(...)  on conflict

jimlambrt:


internal/servers/repository_worker_test.go
```diff
@@ -400,11 +444,13 @@ func TestUpsertWorkerStatus(t *testing.T) {
 			_, err = tc.repo.UpsertWorkerStatus(ctx, tc.status)
 			assert.Error(t, err)
 			tc.errAssert(t, err)
+			t.Log("error: ", err)
```
jefferai, line 447: Fine if you keep these two t.Log statements in, just want to check that it's on purpose and not on accident.

jefferai (approved):

internal/servers/repository_worker_test.go
jimlambrt: yes, intentional.

internal/servers/repository_worker.go
```diff
@@ -277,8 +277,17 @@ func (r *Repository) UpsertWorkerStatus(ctx context.Context, worker *Worker, opt
...
+				err := w.Create(ctx, worker, db.WithOnConflict(workerCreateConflict), db.WithReturnRowsAffected(&withRowsAffected),
+					// The intent of this WithWhere option is to operate with the OnConflict such that the action
+					// taken by the OnConflict only applies if the conflict is on a row that is returned by this where
+					// statement, otherwise it should error out.
+					db.WithWhere("server_worker.type = 'kms'"))
+				switch {
+				case err != nil:
 					return errors.Wrap(ctx, err, op, errors.WithMsg("error creating a worker"))
+				case withRowsAffected == 0:
+					return errors.New(ctx, errors.NotUnique, op, "error updating worker")
```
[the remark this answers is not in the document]
jimlambrt: doesn't this represent an "unique constraint" issue?  It's not a db constraint error, so I don't think it should raise one that looks like that.

jimlambrt:
updated PR (removed duplicate test) @talanknight 

and merging.

PR #2210, opened 2022-06-16
docs(credentials): Add static and username-password to domain model

louisruch:


stellarsquall (changes requested):
Looks great, a couple of minor suggestions

stellarsquall (approved):

PR #2211, opened 2022-06-16
Quote tags before printing them out.

talanknight:


jefferai:
If it works to do this inside `format.go` as another case, I have a preference for doing it there with `json.Marshal`, so that this applies to all functions printing `[]string`

jefferai (changes requested):
I think https://github.com/hashicorp/boundary/blob/main/internal/cmd/base/format.go#L143 is the better place to do this which will automatically handle both of these cases as well as any others.

Either that, or convert to use https://github.com/hashicorp/boundary/blob/main/internal/cmd/base/format.go#L107 which has the benefit of not wrapping horizontally if there are lots of values (at the expense of verticality)

jimlambrt (approved):

PR #2212, opened 2022-06-16
Split proto breaking change detection config

johanbrandhorst:
We generally care about preserving JSON compatibility for protobuf
files, but for some files, where we only use pure gRPC, we only
care about preserving wire compatibility.

Split our buf breaking invocation into two invocations, one that checks
for WIRE_JSON compatibility, and one that checks only for WIRE
compatibility. The former check ignores protos in controller/servers,
since they do not need to adhere to this stricter definition.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -211,7 +211,13 @@ protobuild:
 .PHONY: protolint
 protolint:
 	@buf lint
-	@buf breaking --against 'https://github.com/hashicorp/boundary.git#branch=stable-website'
+	# First check all protos except controller/servers for WIRE_JSON compatibility
+	cd internal/proto && buf breaking --against 'https://github.com/hashicorp/boundary.git#branch=stable-website,subdir=internal/proto' \
+		--exclude-path=controller/servers \
+		--config buf.breaking.json.yaml
+	# Next check all protos for WIRE compatibility. WIRE is a subset of WIRE_JSON so we don't need to exclude any files.
+	cd internal/proto && buf breaking --against 'https://github.com/hashicorp/boundary.git#branch=stable-website,subdir=internal/proto' \
+		--config buf.breaking.wire.yaml
 
 .PHONY: website
 # must have nodejs and npm installed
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -211,7 +211,13 @@ protobuild:
 .PHONY: protolint
 protolint:
 	@buf lint
-	@buf breaking --against 'https://github.com/hashicorp/boundary.git#branch=stable-website'
+	# First check all protos except controller/servers for WIRE_JSON compatibility
+	cd internal/proto && buf breaking --against 'https://github.com/hashicorp/boundary.git#branch=stable-website,subdir=internal/proto' \
+		--exclude-path=controller/servers \
+		--config '{"breaking":{"use":["WIRE_JSON"],"ignore_only":{"FIELD_NO_DELETE_UNLESS_NUMBER_RESERVED":["controller/api","controller/storage"]}},"version":"v1","name":"buf.build/hashicorp/boundary","deps":["buf.build/googleapis/googleapis","buf.build/grpc-ecosystem/grpc-gateway"]}'
```
tmessi, line 217: Can you update this to place the configs into separate files? I tested it out and this still works, so something like:

`internal/proto/buf.breaking.api.yaml`:
```yaml
breaking:
  use:
    - WIRE_JSON
  ignore_only:
    FIELD_NO_DELETE_UNLESS_NUMBER_RESERVED:
      - controller/api
      - controller/storage
version: v1
name: buf.build/hashicorp/boundary
deps:
  - buf.build/googleapis/googleapis
  - buf.build/grpc-ecosystem/grpc-gateway
```
`internal/proto/buf.breaking.grpc.yaml`
```yaml
breaking:
  use:
    - WIRE
  ignore_only:
    FIELD_NO_DELETE_UNLESS_NUMBER_RESERVED:
      - controller/api
      - controller/storage
version: v1
name: buf.build/hashicorp/boundary
deps:
  - buf.build/googleapis/googleapis
  - buf.build/grpc-ecosystem/grpc-gateway
```

Then you would need to update this to `--config=./buf.breaking.api.yaml` and the other to `--config=./buf.breaking.grpc.yaml`

tmessi (changes requested):

johanbrandhorst:
> I think it is unfortunate that the best way to do this is inlining the call in the makefile. That will make it harder to reason about or be able to eye what is happening. Maybe it wouldn't be so bad if we could put the value for the `--config` flag on multiple lines and nicely formatted. Would that be possible?

Does the use of separate config files work better for you?

tmessi (approved):

PR #2216, opened 2022-06-17
Have fallback eventing use json format

jefferai:


jimlambrt:
@jefferai I'm not convinced this approach will work given: https://github.com/hashicorp/boundary/blob/f5bf3b2c0ca14c49b1f96fac9990bd320240c105/internal/observability/event/context.go#L281

we can discuss next week the best path forward.

jefferai:
That's fair. I've also been thinking about it and wondering if our fallback shouldn't be to just write events directly to the output. E.g. if we were going to send to hclog on stderr, maybe just keep a global mutex on the fallback writer and write events out directly to stderr.

jimlambrt:
@jefferai I pushed some commits to try and move this PR forward.

jefferai (commented):
Looks good to me

jimlambrt (approved):

PR #2219, opened 2022-06-19
fix (API): check attributes missing appropriately.

jimlambrt:
Check for nil attributes to prevent panics

### DIFF
--- a/internal/daemon/controller/handlers/authmethods/password.go
+++ b/internal/daemon/controller/handlers/authmethods/password.go
@@ -137,27 +137,32 @@ func validateAuthenticatePasswordRequest(req *pbs.AuthenticateRequest) error {
 	badFields := make(map[string]string)
 
 	attrs := req.GetPasswordLoginAttributes()
-	if attrs.LoginName == "" {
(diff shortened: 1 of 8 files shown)


jimlambrt:
I need to add some unit tests.

jefferai (approved):
TY!

internal/daemon/controller/handlers/authmethods/password.go
```diff
@@ -137,27 +137,32 @@ func validateAuthenticatePasswordRequest(req *pbs.AuthenticateRequest) error {
...
-		tokenType = req.GetTokenType()
-	}
-	tType := strings.ToLower(strings.TrimSpace(tokenType))
-	if tType != "" && tType != "token" && tType != "cookie" {
-		badFields[tokenTypeField] = `The only accepted types are "token" and "cookie".`
+	switch {
+	case attrs == nil:
+		badFields["attributes"] = "This is a required field."
+	default:
+		if attrs.LoginName == "" {
```
louisruch, line 144: As a note the `attrs.GetLoginName()` call is nil protected, most of the new handlers make use of these functions.

PR #2221, opened 2022-06-20
[CI-only] Notarize darwin arm

mdeggies:
Currently, darwin arm binaries are not being notarized. Thanks to Louis for the call-out. This PR adds that in.


louisruch (approved):
LGTM thanks!

PR #2226, opened 2022-06-22
bug(scheduler): Fix controller id length check

louisruch:
The job_run table has a foreign key to the server_controller private_id.
While this column is called a prviate_id it does not implement the wt_private_id
postgres domain type, and therefore does not have a minimum length of 10 characters.

### DIFF
--- a/internal/db/schema/migrations/oss/postgres/35/01_job_migrations.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/35/01_job_migrations.up.sql
@@ -1,0 +1,18 @@
+begin;
+
+-- The job_run table has a foreign key to the server_controller private_id.
+-- While this column is called a prviate_id it does not implement the wt_private_id
+-- domain type, and therefore does not have a minimum length of 10 characters.
+
+-- Drop constraint that requires controller_id to be at least 10 chars
+alter table job_run
+  drop constraint controller_id_must_be_at_least_10_characters;
+
+-- Add a not empty contraint to the controller_id
+alter table job_run
+  add constraint controller_id_must_not_be_empty
+    check(
+      length(trim(controller_id)) > 0
+    );
+
+commit;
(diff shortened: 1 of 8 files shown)


ddebko (approved):
LGTM

jefferai (dismissed):
Looks good; would you mind starting a new Changelog section for 0.9.1?

internal/db/schema/migrations/oss/postgres/35/01_job_migrations.up.sql
```diff
@@ -0,0 +1,18 @@
...
+-- Drop constraint that requires controller_id to be at least 10 chars
+alter table job_run
+  drop constraint controller_id_must_be_at_least_10_characters;
+
+-- Add a not empty contraint to the controller_id
+alter table job_run
+  add constraint controller_id_must_not_be_empty
+    check(
+      length(trim(controller_id)) > 0
+    );
```
mgaffney, line 16: This constraint should also be on the `server_controller` table since `controller_id` is a foreign key. Can we create a domain type for this constraint and then change the column types in the `server_controller` and `job_run` tables to use the new domain type?

mgaffney (changes requested):

internal/db/schema/migrations/oss/postgres/35/01_job_migrations.up.sql
```diff
@@ -0,0 +1,18 @@
+begin;
+
+-- The job_run table has a foreign key to the server_controller private_id.
+-- While this column is called a prviate_id it does not implement the wt_private_id
+-- domain type, and therefore does not have a minimum length of 10 characters.
+
+-- Drop constraint that requires controller_id to be at least 10 chars
```
tmessi, line 7: can you add a comment here that references where this constraint was originally added? And can you update that migration to add a comment noting that it was removed in this migration?

mgaffney (changes requested):

mgaffney (approved):

PR #2229, opened 2022-06-22
Add an extra cleanup task for sessions stuck in canceling state

jefferai:
See the comment on the migrations file for details.

Note: in discussion with @mgaffney there didn't seem to be much use in creating a test for this; it's by definition corrupt data. However, in my home Boundary cluster, this query effectively identifies and removes the stuck sessions, both in the case that they are not terminated in the warehouse (in which case they are marked terminated and removed after an hour) and in the case that they are (in which case they are immediately removed).


internal/db/schema/migrations/oss/postgres/35/01_cleanup.up.sql
```diff
@@ -0,0 +1,22 @@
+begin;
+
+-- This adds additional cleanup to what went into 0.8.x. A forgotten case could
+-- mean that sessions can persist in canceled state even if there are no
+-- connections for that session (possibly due to cleanup) and will never be
+-- transitioned because connection state cannot be determined.
+--
+-- There is no reason to transition to terminated state as we remove those
+-- periodically anyways, and there is no useful information to be gleaned from
+-- having it in that state.
```
mgaffney, line 10: Not transitioning to the terminated state may impact the fact tables in the warehouse. Can we either write sql tests to confirm that deleting will not impact the warehouse? The other option is to not delete them and instead just transition the sessions to the terminated state so the normal workflows continue.

mgaffney (changes requested):

mgaffney (approved):
FYI, the migration file needs to be renamed since a `35/01_...` has been merged into main already.

PR #2231, opened 2022-06-23
fix(error) validate credential store id when creating static cred

ddebko:
Passing in an invalid resource type into the -credential-store-id arg when creating a static credential returns the incorrect error type. The actual error is a 404 Resource Not Found, but the expected error type is a 400 Bad Request. The reason why this bug is occurring is because we are not checking the prefix on the -credential-store-id arg.

Example Command:

```
boundary credentials create username-password -credential-store-id cred_PnlQZvgQ74 -username admin -password password
```

Actual Output:
```
Error from controller when performing create on username_password-type credential

Error information:
  Kind:                NotFound
  Message:             Resource not found.
  Status:              404
  context:             Error from controller when performing create on username_password-type credential
```

Expected Output:

```
Error from controller when performing create on username_password-type credential

Error information:
  Kind:                InvalidArgument
  Message:             Error in provided request.
  Status:              400
  context:             Error from controller when performing create on username_password-type credential

  Field-specific Errors:
    Name:              -credential-store-id
      Error:           This field must be a valid credential store id.
```


louisruch (approved):
This is great thanks for catching and fixing!

PR #2234, opened 2022-06-27
feat: set default connection limit to -1 instead of 1

malnick:
Updates the default connection limit on a target from 1 to 5 in order to reduce the first time user gotcha that most things require multiple connections to work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,13 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 * The plugin execution_dir configuration parameter is now respected by kms plugins too
   [PR](https://github.com/hashicorp/boundary/pull/2239).
 
+### Deprecations/Changes
+
+* sessions: The default connect limit for new sessions changed from 1 to unlimited (-1).
+  Specific connection limits is an advanced feature of Boundary and this setting is
+  more friendly for new users.
+  ([PR](https://github.com/hashicorp/boundary/pull/2234))
+
 ## 0.9.0 (2022/06/20)
 
 ### Known Issues
--- a/internal/db/schema/migrations/oss/postgres/0/41_targets.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/0/41_targets.up.sql
@@ -105,6 +105,7 @@ create table target_tcp (
     constraint session_max_seconds_must_be_greater_than_0
     check(session_max_seconds > 0),
   -- limit on number of session connections allowed. -1 equals no limit
+  -- The default was updated in 37/01_set_unlimited_connections_limit.up.sql.
   session_connection_limit int not null default 1
     constraint session_connection_limit_must_be_greater_than_0_or_negative_1
     check(session_connection_limit > 0 or session_connection_limit = -1),
--- a/internal/db/schema/migrations/oss/postgres/37/01_set_unlimited_connections_limit.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/37/01_set_unlimited_connections_limit.up.sql
@@ -1,0 +1,8 @@
+begin;
+
+-- Changes the default that was set in 0/41_targets.up.sql
+alter table target_tcp
+    alter column session_connection_limit
+        set default -1;
+
+commit;
--- a/internal/session/repository_session_test.go
+++ b/internal/session/repository_session_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/hashicorp/boundary/internal/auth/password"
 	"github.com/hashicorp/boundary/internal/authtoken"
 	authtokenStore "github.com/hashicorp/boundary/internal/authtoken/store"
 	cred "github.com/hashicorp/boundary/internal/credential"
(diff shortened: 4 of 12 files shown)


jefferai:
I think it should be -1. 1 makes sense from a security perspective of tying a session to one connection as a default, but 5 seems arbitrary.

jefferai:
Also I don't think this is a `fix`. I don't know what it is (which is partly why I hate these labels people have been increasingly adopting) but it's not a bug so it's not a fix.

malnick:
@jefferai Thanks, updated to -1 and amended my commit message.

jefferai (dismissed):

mgaffney (changes requested):
* Should we update the [default in the database](https://github.com/hashicorp/boundary/blob/3e9c99c217fc7c250b3d55b2c63e4bdb47895f50/internal/db/schema/migrations/oss/postgres/0/41_targets.up.sql#L108) to be consistent?
* Can we update the [website page](https://github.com/hashicorp/boundary/blob/3e9c99c217fc7c250b3d55b2c63e4bdb47895f50/website/content/docs/concepts/domain-model/targets.mdx#L51) to reflect the new default?

johanbrandhorst:
I will commandeer this PR and make the suggested changes

malnick:
Thank you @johanbrandhorst !

internal/session/testing.go
```diff
@@ -135,7 +135,7 @@ func TestSessionParams(t testing.TB, conn *db.DB, wrapper wrapping.Wrapper, iamR
 	sets := static.TestSets(t, conn, cats[0].PublicId, 1)
 	_ = static.TestSetMembers(t, conn, sets[0].PublicId, hosts)
 
-	tcpTarget := tcp.TestTarget(ctx, t, conn, proj.PublicId, "test target")
+	tcpTarget := tcp.TestTarget(ctx, t, conn, proj.PublicId, "test target", target.WithSessionConnectionLimit(1))
```
johanbrandhorst, line 138: Without adding this option we get errors where sessions aren't progressing from `pending`. I will do some more digging, but "for now" this fixes the test errors.

jefferai: Probably good to get this figured out before release as we've had weird issues with session state management before...

johanbrandhorst: Yeah, I will keep looking

johanbrandhorst: Turns out that Boundary was doing the right thing here and not terminating sessions without a connection limit. The tests that expect a single connection limit have been updated to have this set explicitly on the target.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 
 * `azure` host plugin: Support multiple MSI identities
   ([PR](https://github.com/hashicorp/go-kms-wrapping/pull/97)
+* sessions: The default connect limit for new sessions is now unlimited. This avoids
+  the confusing behavior where a session is terminated as soon as a single connection
```
jefferai, line 12: I don't think this is the right way to put it. It's only confusing behavior to people that either don't look at the created target or don't understand what it means...and phrasing it this way might discourage people from doing anything other than `-1` even though in many cases this is indeed what you want. In fact, we've had users confused the other way -- hey, I set my connection count high and I killed my client and now the session shows as still open, does this mean anyone can now access my system, doesn't Boundary close sessions?

internal/session/repository_session_test.go
```diff
@@ -917,7 +919,47 @@ func TestRepository_CancelSession(t *testing.T) {
...
+				exp := &timestamp.Timestamp{Timestamp: future}
+				org, proj := iam.TestScopes(t, iamRepo)
+
+				cats := static.TestCatalogs(t, conn, proj.PublicId, 1)
+				hosts := static.TestHosts(t, conn, cats[0].PublicId, 1)
+				sets := static.TestSets(t, conn, cats[0].PublicId, 1)
+				_ = static.TestSetMembers(t, conn, sets[0].PublicId, hosts)
+
+				// We need to set the session connection limit to 1 so that the session
+				// is terminated when the one connection is closed.
```
jefferai, line 932: I wonder if the test code isn't sending the going-away RPC that the client does when it's killed. Possibly we'd want to allow either/or but if it was this change wouldn't be necessary.

jefferai (changes requested):
Looks good other than changelog suggestions.

internal/db/schema/migrations/oss/postgres/37/01_set_unlimited_connections_limit.up.sql
```diff
@@ -0,0 +1,7 @@
+begin;
+
```
tmessi, line 2: Can you add a comment to note that this changes the default that was set in `0/41_targets.up.sql` and add a comment to `0/41_targets.up.sql` to note that the default changed in this migration.

tmessi (changes requested):

CHANGELOG.md
johanbrandhorst: How should we phrase this?

internal/session/repository_session_test.go
johanbrandhorst: My understanding was that the implicit terminating of a session after a connection is closed is tied to the connection limit of the session, is that wrong? Would we expect the session to terminate when its only connection closes always?

internal/db/schema/migrations/oss/postgres/37/01_set_unlimited_connections_limit.up.sql
johanbrandhorst: Done

internal/db/schema/migrations/oss/postgres/0/41_targets.up.sql
```diff
@@ -105,6 +105,7 @@ create table target_tcp (
     constraint session_max_seconds_must_be_greater_than_0
     check(session_max_seconds > 0),
   -- limit on number of session connections allowed. -1 equals no limit
+  -- This was updated in 37/01_set_unlimited_connections_limit.up.sql.
```
tmessi, line 108:
```suggestion
  -- The default was updated in 37/01_set_unlimited_connections_limit.up.sql.
```

tmessi (approved):

PR #2235, opened 2022-06-28
Create session manager for workers to use to interact with sessions

talanknight:
This moves all the concurrency logic into a session object which manages communication with the controller and concurrency protections.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 
 ## Next
 
+## 0.9.1 (2022/07/06)
+
+### New and Improved
+
+* `azure` host plugin: Support multiple MSI identities
+  ([PR](https://github.com/hashicorp/go-kms-wrapping/pull/97)
+
 ### Bug Fixes
 
 * scheduler: Fix regression causing controller names of less than 10 characters
@@ -14,6 +21,13 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
   ([PR](https://github.com/hashicorp/boundary/pull/2229))
 * The plugin execution_dir configuration parameter is now respected by kms plugins too
   [PR](https://github.com/hashicorp/boundary/pull/2239).
+
+### Deprecations/Changes
+
+* sessions: The default connect limit for new sessions changed from 1 to unlimited (-1).
+  Specific connection limits is an advanced feature of Boundary and this setting is
+  more friendly for new users.
+  ([PR](https://github.com/hashicorp/boundary/pull/2234))
 
 ## 0.9.0 (2022/06/20)
 
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.10.0
 	github.com/hashicorp/boundary/api v0.0.25
-	github.com/hashicorp/boundary/sdk v0.0.16
+	github.com/hashicorp/boundary/sdk v0.0.17
 	github.com/hashicorp/cap v0.1.1
 	github.com/hashicorp/dawdle v0.4.0
 	github.com/hashicorp/dbassert v0.0.0-20210708202608-ecf920cf1ed8
--- a/internal/cmd/base/servers.go
+++ b/internal/cmd/base/servers.go
@@ -135,9 +135,12 @@ type Server struct {
 
 	DevOidcSetup oidcSetup
 
-	DatabaseUrl                string
-	DatabaseMaxOpenConnections int
-	DevDatabaseCleanupFunc     func() error
+	DatabaseUrl                     string
+	DatabaseMaxOpenConnections      int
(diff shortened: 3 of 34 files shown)


internal/daemon/worker/session/manager.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	s, err := newSess(m.controllerSessionConn, resp)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: Periodically clean this up. We can't rely on things in here but
```
jefferai, line 77: This comment may actually be better in ForEachLocalSession. You can imagine that as we range through we could check time-boxed validity and remove invalid sessions at that time, moving on to the next one.

internal/daemon/worker/session/session.go
```diff
@@ -22,26 +22,240 @@ var errMakeSessionCloseInfoNilCloseInfo = errors.New("nil closeInfo supplied to
...
+	// the request.
+	connInfo, ok := s.connInfoMap[connId]
+	if !ok {
+		return fmt.Errorf("could not find connection ID %q for session ID %q in local state",
+			connId,
+			s.GetId())
+	}
+
+	connInfo.Status = status
+	if connInfo.Status == pbs.CONNECTIONSTATUS_CONNECTIONSTATUS_CLOSED {
```
jefferai, line 84: Should this also call the cancel function? I realize it's called elsewhere, but any reason not to do it here too?

internal/daemon/worker/session/session.go
```diff
@@ -22,26 +22,240 @@ var errMakeSessionCloseInfoNilCloseInfo = errors.New("nil closeInfo supplied to
...
+}
+
+// ApplyLocalStatus updates the given session with the status provided by
+// the SessionJobInfo.  It returns an error if any of the connections
+// in the SessionJobInfo are not present, however, it still applies the
+// status change to the session and the connections which are present.
+func (s *Session) ApplyLocalStatus(st pbs.SESSIONSTATUS) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	s.status = st
```
jefferai, line 97: Same question here.

internal/daemon/worker/session/session.go
```diff
@@ -22,26 +22,240 @@ var errMakeSessionCloseInfoNilCloseInfo = errors.New("nil closeInfo supplied to
...
+func (s *Session) RequestCancel(ctx context.Context) error {
+	st, err := cancel(ctx, s.client, s.GetId())
+	if err != nil {
+		return err
+	}
+	s.ApplyLocalStatus(st)
+	return nil
+}
+
+// RequestActivate Sends session activation request to the controller.  The Session's
```
jefferai, line 166:
```suggestion
// RequestActivate sends session activation request to the controller.  The Session's
```

internal/daemon/worker/session/session.go
```diff
@@ -22,26 +22,240 @@ var errMakeSessionCloseInfoNilCloseInfo = errors.New("nil closeInfo supplied to
...
+	s.ApplyLocalConnectionStatus(info.GetConnectionId(), st)
+	return nil
+}
+
+// CancelOpenLocalConnections closes the local connections in this session
+//based on the connection's state by calling the connections context cancel
+// function.
+//
+// The returned slice are connection ids that were closed.
+func (s *Session) CancelOpenLocalConnections() []string {
```
jefferai, line 219: I'm not sure this and the function below need to be different. Seems like CancelOpenLocalConnections should suffice for any need?

internal/daemon/worker/session/session.go
```diff
@@ -112,17 +320,9 @@ func closeConnection(ctx context.Context, sessClient pbs.SessionServiceClient, r
...
 }
 
-// CloseConnections is a helper worker function that sends connection close
-// requests to the controller, and sets close times within the worker. It is
-// called during the worker status loop and on connection exit on the proxy.
-//
-// The boolean indicates whether the function was successful, e.g. had any
-// errors. Individual events will be sent for the errors if there are any.
-//
-// closeInfo is a map of connections mapped to their individual session.
```
jefferai, line 122: Any reason to remove the comment? Was it just wrong?

internal/daemon/worker/session/manager.go
```diff
@@ -0,0 +1,110 @@
...
+	"sync"
+
+	"github.com/hashicorp/boundary/internal/errors"
+	pbs "github.com/hashicorp/boundary/internal/gen/controller/servers/services"
+)
+
+// Manager stores session information locally and exposes ways to operate on the
+// set of sessions locally in batch.
+// This is thread-safe.
+type Manager struct {
```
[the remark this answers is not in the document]
jefferai: Sorry, I muddled the message. I didn't mean thread-safe, I just meant safe. If they instantiate the manager directly instead of with the constructor, the map and the client are never initialized, so unless you wrap all accesses to them with nil checks, you can't guarantee that you won't panic.

Which reminds me -- there should be an IsNil check on the passed-in client and the constructor should return an error if it's nil :-)

jefferai (approved):

Issue #2122, opened 2022-06-02
Generate nicer SDK getters


johanbrandhorst:
**Is your feature request related to a problem? Please describe.**
When using the SDK, the result getters all return `interface{}` when they have a concrete type available. This is a poor experience for the user. See https://github.com/hashicorp/boundary/blob/2a5aaea8b234425007255b3013d7d18b02128e70/api/sessions/session.gen.go#L45-L47 and https://github.com/hashicorp/boundary/blob/2a5aaea8b234425007255b3013d7d18b02128e70/api/sessions/session.gen.go#L74-L76 for examples.

**Describe the solution you'd like**
I would like the generated code to use typed returns.

**Additional context**
This was raised as a concern by internal users of our SDK. The template for this part of the generated SDK is here: https://github.com/hashicorp/boundary/blob/2a5aaea8b234425007255b3013d7d18b02128e70/internal/api/genapi/templates.go#L564-L631

PPacent:
Thank you for raising this item @johanbrandhorst! We'll keep this item open to gauge interest from the community to inform its priority on our roadmap.

jefferai:
Closing as #2238 has been merged

PR #2238, opened 2022-06-28
Ddebko ICU-2122 Enforce Typed Definitions In API Module

ddebko:
[Issue-2122](https://github.com/hashicorp/boundary/issues/2122)

## Summary:

The focus for issue 2122 is to add typed definitions to our return methods in the API module. Currently we are returning generic types (interface{}), which requires the user to type cast the value into the expected typed definition. Otherwise, the values stored in the generic type are hidden. This PR adds the typed definitions to the API module by updating the template in the genapi package. The changes to the API module created a cascading error effect in the cli commands, which is why I listed many items in the changes section for the cli module.

## Changes:
- Updated apigen module template
  - Changed return type for ReadResult.GetItem() from interface{} to typed definition
  - Changed return type for ListResult.GetItems() from interface{} to typed definition
  - Delete interface definitions: GenericResult, GenericDeleteResult, GenericListResult
- Updated cligen module template
  - removed references to typed interface definitions: GenericResult, GenericDeleteResult, GenericListResult
  - added typed references
    - *api.Response
    - *{{ .PackageName }}.{{ Resource Type }}                       
      - Ex.) *authtokens.AuthToken
    - []*{{ .PackageName }}.{{ Resource Type }}                     
      - Ex.) []*authtokens.AuthToken
    - *{{ .PackageName }}.{{ Resource Type }}ReadResult
      - Ex.) *authtokens.AuthTokenReadResult    
    - *{{ .PackageName }}.{{ Resource Type }}DeleteResult 
      - Ex.) *authtokens.AuthTokenDeleteResult
    - *{{ .PackageName }}.{{ Resource Type }}ListResult
      - Ex.) *authtokens.AuthTokenListResult
  - updated cli helper func executeExtraActions
    - removed parameter "origResult api.GenericResult"
    - added parameters/return types:
      - origResp *api.Response
      - origItem  *{{ .PackageName }}.{{ Resource Type }}
      - origItems []*{{ .PackageName }}.{{ Resource Type }}
        - note this is only generated when a command has a Result type with a method GetItems()
  - updated cli helper funcs PrintJsonItem & PrintJsonItems to pass in *api.Response in format.go
  - updated all helper funcs in files labels funcs.go in gencli package
    - updated executeExtraActionsImpl to match func definition in generated files
    - fixed printItemTable by a case to case situation depending on the command

## Concerns:
- In the API module we have a DeleteResult type that always returns a nil value for GetItem and we have a comment for the functions to explain this, therefore I decided not to change the return type.
- This is a breaking change for users that are currently using the API Module


internal/cmd/gencli/templates.go
```diff
@@ -446,6 +461,8 @@ func (c *{{ camelCase .SubActionPrefix }}Command) Run(args []string) int {
 		return base.CommandCliError
 	}
 
+
+
```
johanbrandhorst, line 465: Nit: do we need these lines?

internal/tests/cluster/recursive_anon_listing_test.go
```diff
@@ -33,25 +33,25 @@ func TestListAnonymousRecursing(t *testing.T) {
 	l, err := amClient.List(tc.Context(), scope.Global.String(), amapi.WithRecursive(true))
 	require.NoError(err)
 	require.NotNil(l)
-	require.Len(l.GetItems().([]*amapi.AuthMethod), 3)
```
johanbrandhorst, line 36: :chefs_kiss:

johanbrandhorst (changes requested):
This looks really good. Could you also prepare some notes in the CHANGELOG that detail the exact breaking changes for an external user of the `api` module?

internal/tests/cluster/recursive_anon_listing_test.go
ddebko: thank you

internal/cmd/gencli/templates.go
ddebko: Good catch, I will delete these lines.

tmessi (approved):

johanbrandhorst (changes requested):
Please rebase this on the v0.9.1 release changelog changes

johanbrandhorst (dismissed):
LGTM but wait for Louis and Mike

louisruch (approved):
:shipit:

johanbrandhorst (approved):

PR #2247, opened 2022-07-08
Workers can act as a server to upstreams that it dials to

talanknight:
In order for the worker to host a grpc server for the upstream it connects to this PR creates:

- A dialing listener:  This dials all upstreams that it is provided with using a unique ALPN Proto and when connected sends the resulting net.Conn to a net.Listener interface.

- A new boundary specific split listener: This looks for the unique ALPN proto and if present sends it to 1 of it's child listeners, otherwise it sends it to the other.

- A grpc.ClientConn producer: This listens to a net.Listener and when it gets a net.Conn it creates a grpc.ClienConn using a Dialer that opens a yamux.Client connection from that net.Conn.


internal/daemon/common/reversegrpc/dialinglistener.go
```diff
@@ -0,0 +1,290 @@
...
+// if e close the yamux connection we should close the underlying one as well
+func (dc *dialingListenerConn) Close() error {
+	err := &multierror.Error{}
+	err = multierror.Append(err, dc.Stream.Close())
+	err = multierror.Append(err, dc.underlyingSession.Close())
+	return multierror.Append(err, dc.underlyingConn.Close()).ErrorOrNil()
+}
+
+func newDialingListenerConn(ctx context.Context, l *dialingListener, c net.Conn, addr string) (*dialingListenerConn, error) {
+	yamuxSession, err := yamux.Server(c, yamux.DefaultConfig())
```
irenarindos, line 51: Check for nil params?

internal/daemon/common/reversegrpc/router.go
```diff
@@ -0,0 +1,55 @@
...
+	"context"
+	"sync"
+
+	"github.com/hashicorp/boundary/internal/errors"
+	pbs "github.com/hashicorp/boundary/internal/gen/controller/api/services"
+	"github.com/hashicorp/boundary/internal/observability/event"
+	"google.golang.org/grpc"
+)
+
+// TODO: The router is currently only a proof of concept for how we might
```
irenarindos, line 13: Is this still only a POC?

internal/daemon/common/reversegrpc/router.go
```diff
@@ -0,0 +1,55 @@
...
+	defer r.routesLock.Unlock()
+	if old, ok := r.routes[workerId]; ok {
+		old.Close()
+	}
+	r.routes[workerId] = cc
+}
+
+// AddClientConnection adds a client connection for a specific route.
+func (r *router) AddClientConnection(ctx context.Context, cc *grpc.ClientConn) error {
+	const op = "controllerled.(*router).AddClientConnection"
```
irenarindos, line 39: Check for nil cc?

irenarindos, line 39:
```suggestion
	const op = "reversegrpc.(*router).AddClientConnection"
```

internal/daemon/common/reversegrpc/dialinglistener.go
```diff
@@ -0,0 +1,290 @@
...
+	return &net.TCPAddr{
+		IP:   net.IP("0"),
+		Port: 0,
+	}
+}
+
+// dial blocks until it successfully dials to the provided address or the
+// provided context is done.
+func (d *dialingListener) dial(ctx context.Context, addr string) error {
+	bf := backoff.NewExponentialBackOff()
```
irenarindos, line 234: Check for empty addr?

internal/daemon/worker/worker.go
```diff
@@ -355,17 +376,20 @@ func (w *Worker) Shutdown() error {
 	waitStatusStart := time.Now()
 	nextStatusCtx, nextStatusCancel := context.WithTimeout(w.baseContext, w.conf.StatusGracePeriodDuration)
 	defer nextStatusCancel()
+	secondTimer := time.NewTimer(0)
```
irenarindos, line 379: What is the secondTimer for? Can you document what this does?

PR #2251, opened 2022-07-08
Prototype code for bytes up/down

johanbrandhorst:


api/sessions/connection.gen.go
```diff
@@ -6,7 +6,7 @@ type Connection struct {
 	ClientTcpPort      uint32 `json:"client_tcp_port,omitempty"`
 	EndpointTcpAddress string `json:"endpoint_tcp_address,omitempty"`
 	EndpointTcpPort    uint32 `json:"endpoint_tcp_port,omitempty"`
-	BytesUp            uint64 `json:"bytes_up,omitempty"`
-	BytesDown          uint64 `json:"bytes_down,omitempty"`
+	BytesUp            uint64 `json:"bytes_up,string,omitempty"`
+	BytesDown          uint64 `json:"bytes_down,string,omitempty"`
```
johanbrandhorst, line 10: This is a little annoying. The gRPC-Gateway marshals the uint64s into a string in the response. This is because most JSON implementations implement numbers as float64, which cannot accurately represent integers larger than 2^53 (9,007,199,254,740,992). We have two options:

* Modify the Boundary client (and Desktop client) to expect these uint64s to be wrapped in a string (this is what this change does). This is the safe option, as it means most JSON implementations will support unmarshalling the data for all numbers that may be contained in these fields. However, it means that old clients will not work with new controllers, since the existing code tries to marshal this out of a number literal.
* Modify the output of the gateway to return these as number literals. This would mean that old and new clients would keep working, up until we get connections with more than ~9 PB (PetaBytes) of data uploaded or downloaded. Then things might silently (or loudly) start to fail. This option will probably be easier to implement.

johanbrandhorst:
This has been merged!

PR #2256, opened 2022-07-13
Migrate plugin host/set/catalog prefixes to typed

jefferai:


internal/host/plugin/ids.go
```diff
@@ -10,16 +10,19 @@ import (
...
 const (
-	HostCatalogPrefix = "hc"
-	HostSetPrefix     = "hs"
-	HostPrefix        = "h"
+	HostCatalogPrefix         = "hcplg"
+	PreviousHostCatalogPrefix = "hc"
+	HostSetPrefix             = "hs"
+	PreviousHostSetPrefix     = "hsplg"
+	HostPrefix                = "h"
+	PreviousHostPrefix        = "hplg"
```
johanbrandhorst, line 25: The choice of naming here seems inconsistent, do we really want the _new_ `HostCatalog` prefix to include the `plg` suffix when the other new ones don't?

jefferai: Oh weird

How did that happen

johanbrandhorst (approved):

internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -1665,7 +1665,7 @@ func validateAddSetsRequest(req *pbs.AddTargetHostSetsRequest) error {
 	}
 	for _, id := range req.GetHostSetIds() {
 		if !handlers.ValidId(handlers.Id(id), static.HostSetPrefix) &&
-			!strings.HasPrefix(id, fmt.Sprintf("%s_", plugin.HostSetPrefix)) {
+			!(strings.HasPrefix(id, fmt.Sprintf("%s_", plugin.HostSetPrefix)) || strings.HasPrefix(id, fmt.Sprintf("%s_", plugin.PreviousHostSetPrefix))) {
```
johanbrandhorst, line 1668: Minor improvement maybe?
```go
prefix, _, _ := strings.Cut(id, "_")
if !handlers.ValidId(handlers.Id(id), static.HostSetPrefix) && prefix != plugin.HostSetPrefix && prefix != plugin.PreviousHostSetPrefix

**Review by johanbrandhorst — DISMISSED** [review]

File: internal/credential/static/public_ids.go [hunk 3]
```diff
@@ -10,18 +10,20 @@ import (
...
 	}
 }
 
 // PublicId prefixes for the resources in the static package.
 const (
-	CredentialStorePrefix = "cs"
-	CredentialPrefix      = "cred"
+	CredentialStorePrefix         = "csst"
+	PreviousCredentialStorePrefix = "cs"
+	CredentialPrefix              = "credst"
```
louisruch, line 25: This credential is actually referring to the `username_password` credential, each type should get it's own prefix as they have different attributes. So something like `cup` or `credup`

jefferai:
@louisruch @johanbrandhorst After discussion with @louisruch we realized some more changes needed to be made. I've also simplified the logic -- I had added to logic that was there in target_service.go that already was doing the wrong thing -- thanks for the pointer on that @johanbrandhorst .

louisruch (approved):
LGTM

johanbrandhorst (approved):

PR #2266, opened 2022-07-19
feat(cli): add/set/remove worker tags

A-440Hz:
example commands:

adding key with multiple values, adding multiple keys:

 `bin/boundary workers add-worker-tags -id w_1234567890 -tag key=val1,val2 -tag key2=val`
 
>   Worker information:
>   Active Connection Count:   0
>   Address:                   127.0.0.1:9202
>   Created Time:              Mon, 25 Jul 2022 10:03:39 PDT
>   ID:                        w_1234567890
>   Last Status Time:          2022-07-25 19:41:09.485499 +0000 UTC
>   Type:                      pki
>   Updated Time:              Mon, 25 Jul 2022 12:41:09 PDT
>   Version:                   2
> 
>   Scope:
>     ID:                      global
>     Name:                    global
>     Type:                    global
> 
>   Tags:
>     Configuration:
>       type: ["dev" "local"]
>     Api:
>       key: ["val1" "val2"]
>       key2: ["val"]
>     Canonical:
>       key: ["val1" "val2"]
>       key2: ["val"]
>       type: ["dev" "local"]

removing a single tag:

 `bin/boundary workers remove-worker-tags -id w_1234567890 -tag key=val1`
 
>   Tags:
>     Configuration:
>       type: ["dev" "local"]
>     Api:
>       key: ["val2"]
>       key2: ["val"]
>     Canonical:
>       key: ["val2"]
>       key2: ["val"]
>       type: ["dev" "local"]

setting all tags to nil:

 `bin/boundary workers set-worker-tags -id w_1234567890 -tag null`
 
>    Tags:
>     Configuration:
>       type: ["dev" "local"]
>     Canonical:
>       type: ["dev" "local"]


A-440Hz:
just the string parsing/initial flag stuff for now

irenarindos:
Can you add the new commands to internal/cmd/commands.go?

internal/tests/api/workers/worker_test.go
```diff
@@ -0,0 +1,55 @@
...
+	tc := controller.NewTestController(t, nil)
+	defer tc.Shutdown()
+
+	client := tc.Client()
+	token := tc.Token()
+	client.SetToken(token.Token)
+
+	tw := worker.NewTestWorker(t, &worker.TestWorkerOpts{
+		WorkerAuthKms:    tc.Config().WorkerAuthKms,
+		InitialUpstreams: tc.ClusterAddrs()})
```
irenarindos, line 25: Just a small gofmt thing:
```suggestion
		InitialUpstreams: tc.ClusterAddrs(),
	})
```

internal/tests/api/workers/worker_test.go
```diff
@@ -0,0 +1,55 @@
...
+	defer tw.Shutdown()
+
+	workerClient := workers.NewClient(client)
+
+	wcr, err := workerClient.CreateWorkerLed(tc.Context(), tc.Token().Token, scope.Global.String())
```
irenarindos, line 30: It looks like this is failing because the test needs to generate worker credentials for an authorization request and pass them where tc.Token().Token is now; I'll help you with that so we can get this test working.

A-440Hz: thank you!! seems like I was missing a lot

internal/tests/api/workers/worker_test.go
```diff
@@ -0,0 +1,105 @@
...
+	inputTags := map[string][]string{"key": {"value"}}
+	wcr, err = workerClient.AddWorkerTags(tc.Context(), wcr.Item.Id, wcr.Item.Version, inputTags)
+	require.NoError(err)
+	require.NotNil(wcr)
+	for k, v := range wcr.Item.ApiTags {
+		for expK, expV := range inputTags {
+			assert.Equal(k, expK)
+			assert.ElementsMatch(v, expV)
+		}
+	}
```
irenarindos, line 77: Can you add some error cases to AddWorkerTags and ensure it errors? (Tag has a comma, adding the same KV twice, etc)

internal/tests/api/workers/worker_test.go
```diff
@@ -0,0 +1,105 @@
...
+	wcr, err = workerClient.SetWorkerTags(tc.Context(), wcr.Item.Id, wcr.Item.Version, inputTags)
+	require.NoError(err)
+	require.NotNil(wcr)
+	for k, v := range wcr.Item.ApiTags {
+		for expK, expV := range inputTags {
+			assert.Equal(k, expK)
+			assert.ElementsMatch(v, expV)
+		}
+	}
+
```
irenarindos, line 89: Can you also add some error cases for Set and Remove tags?

irenarindos (commented):
Looks great, just a suggestion around testing some error cases

irenarindos:
Looks like we're going to hold off on merging this until @jefferai is back and can take a look

internal/tests/api/workers/worker_test.go
A-440Hz: The tag commas are checked when the string from the CLI is first parsed as a flag value, so the test format here will not be able to reach it

internal/cmd/base/flags.go
```diff
@@ -682,6 +682,112 @@ func mapToKV(m map[string]string) string {
...
+}
+
+func (s *stringSliceMapValue) Set(val string) error {
+	kv := strings.TrimSpace(val)
+	if kv == "null" {
+		s.target = nil
+		return nil
+	}
+
+	if *s.target == nil {
```
jefferai, line 728: This will panic if passed two values, where the first is `null`:

```
boundary workers set-worker-tags -id foobar -tag null -tag foo=bar
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x119b0a1]

goroutine 1 [running]:
github.com/hashicorp/boundary/internal/cmd/base.(*stringSliceMapValue).Set(0xc000c422b0, {0x7ffeb550c1f8, 0x7})
        /home/jeff/hc/boundary/internal/cmd/base/flags.go:728 +0xa1
flag.(*FlagSet).parseOne(0xc000bc8720)
        /usr/lib/go/src/flag/flag.go:1087 +0x643
flag.(*FlagSet).Parse(0xc000bc8720, {0xc00012e030?, 0xc0001a5a68?, 0x417f53?})
        /usr/lib/go/src/flag/flag.go:1106 +0x4f
github.com/hashicorp/boundary/internal/cmd/base.(*FlagSets).Parse(...)
        /home/jeff/hc/boundary/internal/cmd/base/base.go:495
github.com/hashicorp/boundary/internal/cmd/commands/workerscmd.(*Command).Run(0xc00069ef30, {0xc00012e030, 0x6, 0x6})
        /home/jeff/hc/boundary/internal/cmd/commands/workerscmd/workers.gen.go:136 +0x15c
github.com/mitchellh/cli.(*CLI).Run(0xc0004b2140)
        /home/jeff/go/pkg/mod/github.com/mitchellh/cli@v1.1.2/cli.go:262 +0x5f8
github.com/hashicorp/boundary/internal/cmd.RunCustom({0xc00012e010?, 0x60?, 0x0?}, 0x0?)
        /home/jeff/hc/boundary/internal/cmd/main.go:186 +0xa16
github.com/hashicorp/boundary/internal/cmd.Run(...)
        /home/jeff/hc/boundary/internal/cmd/main.go:92
main.main()
        /home/jeff/hc/boundary/cmd/boundary/main.go:13 +0xc7
```

The reason is you set the pointer to the map to `nil` above, but here you are checking if the map itself is nil and creating it if so. You can fix this either by adding an additional check on `s.target` or by setting `*s.target = nil` in the if clause above.

internal/cmd/base/flags.go
```diff
@@ -682,6 +682,112 @@ func mapToKV(m map[string]string) string {
...
+	*target = def
+	return &stringSliceMapValue{
+		hidden: hidden,
+		target: target,
+	}
+}
+
+func (s *stringSliceMapValue) Set(val string) error {
+	kv := strings.TrimSpace(val)
+	if kv == "null" {
```
jefferai, line 723: This is sort of an in-between pattern, between the add/set/remove we have for e.g. grants on roles, and attributes/secrets like we have on host catalogs. The host catalogs have two flags: `attr` or `secret` for individual values, or `attributes` and `secrets` for setting the full value at once. Array fields like grants and principals on roles allow `null` only in a `set` call.

Based on that prior art I think we don't need a `tags` field right now, but I do think we should limit `null` to a `set` call, and track its usage to ensure that if `null` is passed during `set` that it's the only value passed.

A-440Hz: I understand limiting `set-worker-tags` to `null` only, based on the array fields precedent, but don't we need the `tags` field to allow `add-worker-tags` and `remove-worker-tags` to work?

internal/cmd/base/flags.go
A-440Hz: thanks for the catch!

internal/cmd/commands/workerscmd/funcs.go
```diff
@@ -25,12 +54,103 @@ func (c *Command) extraHelpFunc(helpMap map[string]func() string) string {
...
 		helpStr = helpMap[c.Func]()
 	}
 	return helpStr + c.Flags().Help()
 }
 
+func extraFlagsFuncImpl(c *Command, _ *base.FlagSets, f *base.FlagSet) {
+	for _, name := range flagsMap[c.Func] {
+		switch name {
+		case "tag":
+			var nullCheckFn func() bool = nil
```
jefferai, line 103: You can just have your function check whether c.Func starts with `set-`,

internal/cmd/commands/workerscmd/funcs.go
```diff
@@ -25,12 +54,103 @@ func (c *Command) extraHelpFunc(helpMap map[string]func() string) string {
...
+				NullCheck: nullCheckFn,
+				Usage:     "The api tag resources to add, remove, or set.",
+			})
+		}
+	}
+}
+
+func extraFlagsHandlingFuncImpl(c *Command, _ *base.FlagSets, opts *[]workers.Option) bool {
+	switch c.Func {
+	case "add-worker-tags", "remove-worker-tags":
```
jefferai, line 121: This is missing a `set-worker-tags` case.

internal/cmd/config/config.go
```diff
@@ -546,13 +546,19 @@ func Parse(d string) (*Config, error) {
 			if !strutil.Printable(k) {
 				return nil, fmt.Errorf("Tag key %q contains non-printable characters", k)
 			}
+			if strings.Contains(k, ",") {
```
jefferai, line 549: Not that this is necessarily wrong, but what's the reason for this?

internal/cmd/commands/workerscmd/funcs.go
A-440Hz: Do we want an empty `-tag` entry to convert to nil for set-worker-tags? I thought it was pretty nifty when `boundary set-worker-tags -id w_idxxxxxxxx` cleared the tags

internal/cmd/config/config.go
A-440Hz: Since no commas are allowed thru the CLI entry, this applies the same standards to tag entries from the config file. Otherwise you can input keys such as "k3,k4" if you input them this way

>   tags {
    key = ["val1", "val2"],
    key2 = ["val1", "val2"],
    "k3,k4" = ["vvv"],
  }

internal/cmd/commands/workerscmd/funcs.go
A-440Hz: I added a `set-worker-tags` case with `No tag supplied via -tag. Please use "-tag null" to clear all tags` as the UI error when there is no -tag input

A-440Hz: Actually yeah I'm not sure what to do here since if I check whether `c.FlagTags` is empty, it overrides the "null" case--where I set target to `nil` for a `stringSliceMapValue` in `flags.go`--and returns an error. 

I don't want to set a token entry for the `map[string][]string` to represent that its value should be nil, since that limits a valid key from use (even if it's a mapping like `null=true`). I don't like the idea of using an invalid key with commas like some form of `,,null,,` since I think that's a bit confusing and unintuitive.

The most clear way forward might be creating another/using another existing bool to distinguish when the user inputs a "null" entry for `set-worker-tags`..

However, do I have to put a `set-worker-tags` case here? Is it alright for the user to be able to null the entry via `boundary set-worker-tags -id w_idxxxxxxxx` and `boundary set-worker-tags -id w_idxxxxxxxx -tag null`?

PR #2282, opened 2022-07-22
[QTI-220] Add infrastructure integration testing

ryancragun:
On pull requests that originate from branches of the `hashicorp/boundary`
repo, we now conditionally dispatch infrastructure integration tests using
the Enos framework. The `integration` Enos scenario includes two `builder`
variants which allow developers or CI to deploy a Boundary cluster in AWS
using either a local `boundary` install bundle or that which is produced
in the `build` workflow for CRT. The `test` variants determine which type
of infrastructure integration test to run. At present those options are
`cli_ui` (Bats CLI tests) or `smoke` (creates target instances and a
target host catalog and verifies that `boundary` works).

In addition to the new workflow and quality scenarios, we simplified the
some `build` workflow steps and added Go module caching to speed up the
build pipeline.

See the [enos/README.md](https://github.com/hashicorp/boundary/blob/main/enos/README.md)
for a detailed explanation on how to configure and execute Enos scenarios
from your machine.

Signed-off-by: Ryan Cragun <me@ryan.ec>
Co-authored-by: Rebecca Willett <rwillett@hashicorp.com>
Co-authored-by: Josh Brand <jbrand@hashicorp.com>

### DIFF
--- a/.github/workflows/enos-fmt.yml
+++ b/.github/workflows/enos-fmt.yml
@@ -10,7 +10,7 @@ jobs:
   fmt_check:
     # Only run this workflow on pull requests from hashicorp/boundary branches
     # as we need secrets to install enos.
-    if: ${{ github.event.pull_request.head.repo.full_name == github.repository && github.actor != 'dependabot[bot]'}}
+    if: "! github.event.pull_request.head.repo.fork"
     runs-on: ubuntu-latest
     env:
       GITHUB_TOKEN: ${{ secrets.SERVICE_USER_GITHUB_TOKEN }}
--- a/.github/workflows/enos-fmt.yml
+++ b/.github/workflows/enos-fmt.yml
@@ -1,0 +1,27 @@
+---
+name: enos_fmt
+
+on:
+  pull_request:
+    paths:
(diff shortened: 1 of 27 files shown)


tmessi (approved):
Looks good to me, but I defer to @kpenfound or @claire-labry for changes to `.github/workflows/build.yml`

.github/workflows/enos-fmt.yml
```diff
@@ -0,0 +1,27 @@
...
+on:
+  pull_request:
+    paths:
+      - enos/**
+
+jobs:
+  fmt_check:
+    # Only run this workflow on pull requests from hashicorp/boundary branches
+    # as we need secrets to install enos.
+    if: ${{ github.event.pull_request.head.repo.full_name == github.repository && github.actor != 'dependabot[bot]'}}
```
ryancragun, line 13: We should simplify this as we have in the build workflow.

jeanneryan (approved):
Had a question about the gpg key and I was told this is a throwaway secret that is only needed for boundary's default keychain. It's not used for package signing.

PR #2285, opened 2022-07-22
fix: Respect context when retrying

johanbrandhorst:
This PR refactors the existing retry code to use a well known community library for retries, while maintaining roughly the same behavior as before. This new library is configured to respect the cancellation of the context being passed in, preventing us from retrying events that are doomed to fail.

The refactor isn't _really_ necessary, so if we want I can tear that out again and just add the logic for respecting the context cancelation.


jimlambrt:
I suggest adding:

```
ATTEMPTS:
	for attempts := uint(1); ; attempts++ {
		select {
		case <-ctx.Done():
			break ATTEMPTS
		default:
                     // existing code...
```

johanbrandhorst:
> I suggest adding:
> 
> ```
> ATTEMPTS:
> 	for attempts := uint(1); ; attempts++ {
> 		select {
> 		case <-ctx.Done():
> 			break ATTEMPTS
> 		default:
>                      // existing code...
> ```

You mean, instead of the refactor? I can do that.

jimlambrt (approved):
Ty.

PR #2289, opened 2022-07-25
When allowed_origins is "*" use that in response

jefferai:


internal/daemon/controller/handler.go
```diff
@@ -419,6 +419,9 @@ func wrapHandlerWithCors(h http.Handler, props HandlerProperties) http.Handler {
 
 		case len(allowedOrigins) == 1 && allowedOrigins[0] == "*":
 			valid = true
+			// When allowed origins is "*" we want to return that rather than
+			// round-tripping any user-specified value
+			origin = "*"
```
johanbrandhorst, line 424: Can we add a test for this to prevent regressions?

jefferai: Sorry, my bad -- I should have marked as Draft. This was intended to get a look by a specific audience but I didn't make it clear it wasn't ready for a general review yet!

johanbrandhorst (approved):

PR #2292, opened 2022-07-25
OSS side of second downstream connections (see enterprise PR 69)

talanknight:


internal/daemon/controller/kms_conn_router.go
```diff
@@ -0,0 +1,139 @@
...
+// KMS key. This means that KMS access is a requirement, and simple replay
+// itself is not sufficient.
+func startKmsConnRouter(
+	c *Controller,
+	baseLn net.Listener,
+	authedListener,
+	reverseGrpcListener *nodeenet.MultiplexingListener,
+) {
+	const op = "controller.startKmsAuthRouter"
+
```
irenarindos, line 50: Check for nil params?

irenarindos (approved):
One small thing but looks good

PR #2296, opened 2022-07-27
[CI-only] Build updates

mdeggies:
This PR has a few small changes:

- Triggers the build workflow to run on every commit. This change is not required, but may help catch build failures on PR's before they're merged into `main`. Most projects have move towards building on every commit. 
- Sets the go version used for building boundary in a single place, `.go-version`, which is used by all downstream CI jobs. Users of goenv who are building boundary locally using the make targets will also benefit from this
- Updates the build jobs themselves to use the `make build` target, instead of the inline `go build` commands. The makefile targets will become the source of truth for building boundary.
- Adds some additional logging to the build script which is useful during debugging.

Testing:

- The artifacts built from the first commit on this branch are available here to be tested: https://github.com/hashicorp/boundary/actions/runs/2749912094. You can view the full successful pipeline for that commit here: https://github.com/hashicorp/boundary/runs/7550183842. I performed basic validations locally on the darwin and linux_amd64 artifacts- ideally enos will be able to help with the rest!


.release/ci.hcl
```diff
@@ -13,6 +13,7 @@ project "boundary" {
 
     release_branches = [
       "main",
+      "release-mods"
```
mdeggies, line 16: This will be removed before merging; it's just needed to test the full pipeline e2e

scripts/build.sh
```diff
@@ -71,4 +75,13 @@ if [ "${BOUNDARY_INSTALL_BINARY}x" != "x" ]; then
...
 
 # Done!
-echo "==> Done!"
+echo "==> Results:"
+ls -hl bin/
+
+# Print the version output for just linux_amd64.
+# Since we run this script in CI, and our CI build jobs run on linux runners, 
+# We can only check the version output for a subset of builds
+if [ "${GOOS}" == "linux" ] && [ "${GOARCH}" == "amd64" ]; then
```
mdeggies, line 84: @tmessi I added your suggestion to check the version output here. Unfortunately, this has some limitations since ubuntu runners can only run linux_amd64 artifacts, and seg faults on all others.. but it does provide a little bit more visibility

tmessi: Makes sense

tmessi (approved):

sarahethompson (approved):

PR #2298, opened 2022-07-28
chore: update go-dbw dependencies 

jimlambrt:
This update included an update of the upstream gorm dependency.

In the latest update, gorm appears to have become more sensitive when
a slice is passed into operations.  The ptr to slice must now
be a ptr to slice of ptrs.  This change required an update to how
boundary handles slices of PrincipalRole, which now must return a ptr
to PrincipalRole.   There were also some tests that needed to be
updated because of this change.


go.mod
```diff
@@ -69,25 +69,25 @@ require (
...
-	gorm.io/driver/postgres v1.2.2
-	gorm.io/gorm v1.22.3
+	gorm.io/driver/postgres v1.3.8
+	gorm.io/gorm v1.23.8
 	mvdan.cc/gofumpt v0.3.1
 	nhooyr.io/websocket v1.8.7
 )
 
-require github.com/hashicorp/go-dbw v0.0.0-20220412153211-c470aec9369f // this is a branch and should be updated before merging
+require github.com/hashicorp/go-dbw v0.0.0-20220725170111-b7cb3aa3d628 // this is a branch and should be updated before merging
```
johanbrandhorst, line 90: So, judging by the fact that this comment was already in main, I assume it's inaccurate?

johanbrandhorst (approved):

go.mod
jimlambrt: ty!

johanbrandhorst (approved):

PR #2299, opened 2022-07-28
Add methods to get attributes more easily in Go API

jefferai:


johanbrandhorst (commented):
Approach looks reasonable to me, provided that it works 😛.

jefferai:
@johanbrandhorst Did some manual testing and it seems fine. Added an extra check, see waht you think?

johanbrandhorst (approved):
Ooh that type check is nice

jefferai:
TY!

PR #2308, opened 2022-07-29
Remove auth.DisabledAuthTestContext from session and target service tests

tmessi:
Disabling the auth was resulting in some unrealistic tests setup and 
expectations. It also makes it difficult to refactor any behavior in the 
authn/authz flow since this option can bypass most of it.

### DIFF
--- a/internal/daemon/controller/handlers/sessions/session_service_test.go
+++ b/internal/daemon/controller/handlers/sessions/session_service_test.go
@@ -19,6 +19,7 @@ import (
 	"github.com/hashicorp/boundary/internal/host/static"
 	"github.com/hashicorp/boundary/internal/iam"
 	"github.com/hashicorp/boundary/internal/kms"
+	"github.com/hashicorp/boundary/internal/requests"
 	"github.com/hashicorp/boundary/internal/server"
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):
Awesome job!

internal/daemon/controller/handlers/sessions/session_service_test.go
```diff
@@ -270,7 +292,8 @@ func TestList(t *testing.T) {
 	tarOther := tcp.TestTarget(ctx, t, conn, pWithOtherSessions.GetPublicId(), "test", target.WithHostSources([]string{hsOther.GetPublicId()}))
 
 	var wantSession []*pb.Session
-	var totalSession []*pb.Session
+	var wantOtherSession []*pb.Session
+	// var totalSession []*pb.Session
```
hugoamvieira, line 296: Presumably we don't need this anymore?

hugoamvieira (approved):

internal/daemon/controller/handlers/sessions/session_service_test.go
tmessi: yep

hugoamvieira (approved):

PR #2311, opened 2022-07-30
Update grpc-gateway dependency

johanbrandhorst:
This fixes an issue which would allow an attacker to trigger a
(recovered) panic in the gateway handler.


internal/gen/controller.swagger.json
```diff
@@ -3729,6 +3729,33 @@
         ]
       }
     },
+    "/v1/workers/:create:worker-led": {
```
johanbrandhorst, line 3732: This path changed 🤔

johanbrandhorst: This change appears to be limited to the openapiv2 generator, the handler is still doing the same thing.

johanbrandhorst: I'm pretty sure this is caused by https://github.com/grpc-ecosystem/grpc-gateway/pull/2727

jefferai: Can we wait to merge this until that PR is pulled in too?

sdk/pbs/controller/api/resources/groups/group.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.27.1
+// 	protoc-gen-go v1.28.1
```
jefferai, line 3: Can you update the tools file if we're going to standardize on this for the next release?

jefferai (changes requested):

internal/gen/controller.swagger.json
johanbrandhorst: I've got a fix in https://github.com/grpc-ecosystem/grpc-gateway/pull/2825.

sdk/pbs/controller/api/resources/groups/group.pb.go
johanbrandhorst: The tools file doesn't contain versions, only modules, right? The version is already controlled by `go.mod`, which is updated by this PR.

johanbrandhorst:
I don't think we need to get this in for 0.10.0. The impact of this bug is that an attacker can basically spam the logs of the controller with panic messages.

internal/gen/controller.swagger.json
johanbrandhorst: I can confirm that this PR fixes the issue for us. I would like to wait for one of the other gateway contributors to comment on it before merging though.

johanbrandhorst:
This is ready for review again, we can merge once 0.10.0 is out.

PR #2312, opened 2022-08-02
feat(workers): Add ability to read and reinitialize cert authority

irenarindos:
Addition of domain function ReinitializeRoots
API implementation of ReinitializeCertificateAuthority and ListCertificateAuthority


internal/proto/controller/api/resources/workers/v1/worker.proto
```diff
@@ -85,3 +85,17 @@ message Worker {
...
+
+  // Output only. The public key of the cert authority
+  bytes public_key = 20; // @gotags: `class:"public"`
+
+  // Output only. The time before which this CA is invalid
+  google.protobuf.Timestamp not_before_time = 30; // @gotags: `class:"public"`
+
+  // Output only. The time after which this CA is invalid
+  google.protobuf.Timestamp not_after_time = 40; // @gotags: `class:"public"`
+}
```
irenarindos, line 101: I'd like feedback on these fields when returning a CertAuthority.
Initially had the cert DER- is that needed?

internal/server/service_rotate_roots.go
```diff
@@ -18,7 +19,15 @@ func RotateRoots(ctx context.Context, workerAuthRepo *WorkerAuthRepositoryStorag
...
-	_, err := rotation.RotateRootCertificates(ctx, workerAuthRepo, opt...)
+	// Get current root certs so we can pass along the version
+	rootCerts := &types.RootCertificates{Id: CA_ID}
+	err := workerAuthRepo.Load(ctx, rootCerts)
+	if err != nil && err != nodeenrollment.ErrNotFound {
+		return errors.Wrap(ctx, err, op)
+	}
+	opt = append(opt, nodeenrollment.WithState(rootCerts.GetState()))
+
+	_, err = rotation.RotateRootCertificates(ctx, workerAuthRepo, opt...)
```
jefferai, line 30: `rotation.RotateRootCertificates` will return the new root certificates on success. It might be nice to return those from here, so that up in the service you don't need to rotate followed by a lookup.

internal/daemon/controller/handlers/workers/worker_service.go
```diff
@@ -408,6 +416,58 @@ func (s Service) RemoveWorkerTags(ctx context.Context, req *pbs.RemoveWorkerTags
...
+	if authResults.Error != nil {
+		return nil, authResults.Error
+	}
+
+	repo, err := s.workerAuthFn()
+	if err != nil {
+		return nil, err
+	}
+
+	err = server.ReinitializeRoots(ctx, repo)
```
jefferai, line 457: See comment below, but I think this function could return the new roots (by having the function it chains to return them) so you don't need an additional lookup just after this.

internal/daemon/controller/handlers/workers/worker_service.go
```diff
@@ -935,3 +1034,26 @@ func validateRemoveTagsRequest(req *pbs.RemoveWorkerTagsRequest) error {
 	}
 	return nil
 }
+
+func validateReadCaRequest(req *pbs.ReadCertificateAuthorityRequest) error {
+	badFields := map[string]string{}
+	if req.GetScopeId() != scope.Global.String() {
+		badFields["scope_id"] = "Must be 'global' when listing."
```
jefferai, line 1041: "when reading"?

internal/proto/controller/api/services/v1/worker_service.proto
```diff
@@ -123,6 +123,26 @@ service WorkerService {
       summary: "Removes api tags from an existing Worker."
     };
   }
+
+  // ReadCertificateAuthority returns the current and next set of root certificates
+  rpc ReadCertificateAuthority(ReadCertificateAuthorityRequest) returns (ReadCertificateAuthorityResponse) {
+    option (google.api.http) = {
+      get: "/v1/workers:read-cas"
```
jefferai, line 130: While some equate one root == one certificate authority in my mind the authority is the thing responsible for the roots and using them. So I think this would be singular -- `read-ca` and `reinitialize-ca`.

Also there is a mismatch here between the action on the API and the action in the grants, which are `read-certificate-authority` and `reinitialize-certificate-authority`, that needs to be solved since the two need to match. Personally I lean towards the latter, we have tended to favor explicitness over brevity, but I could be convinced either way. Maybe poll the BE channel?

internal/server/root_certificate.go
```diff
@@ -12,8 +12,8 @@ import (
 )
 
 // The CertificateAuthority id will always be set to "roots".
-// The const ca_id contains this value
-const ca_id = "roots"
+// The const CA_ID contains this value
+const CA_ID = "roots"
```
jefferai, line 16: Curious about this change. Our normal style would actually be `caId` so to me this reads as a change from one form that isn't our normal form to a different form that isn't our normal form :-)

internal/server/service_reinitialize_roots.go
```diff
@@ -0,0 +1,38 @@
...
+		return errors.New(ctx, errors.InvalidParameter, op, "missing workerAuthRepo")
+	}
+
+	// Delete roots from workerAuthRepo
+	rootCerts := &types.RootCertificates{Id: CA_ID}
+	err := workerAuthRepo.Load(ctx, rootCerts)
+	if err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+	err = workerAuthRepo.Remove(ctx, rootCerts)
```
jefferai, line 27: I think we should fold this functionality into an option in RotateRoots (and from there an option into nodeenrollment). By doing so if our storage is transactional, the removal and rotation happens in the same transaction. It'd be simple to implement in nodeenrollment, it just calls Remove on the existing values at the start of the function if the option is set.

jefferai: In fact if you do this I think you can get rid of this function entirely in favor of calling RotateRoots with an option indicating to clear storage beforehand.

internal/types/action/action.go
```diff
@@ -8,110 +8,114 @@ import (
...
+	RemoveCredentialSources          Type = 41
+	AddHostSources                   Type = 42
+	SetHostSources                   Type = 43
+	RemoveHostSources                Type = 44
+	CreateWorkerLed                  Type = 45
+	AddWorkerTags                    Type = 46
+	SetWorkerTags                    Type = 47
+	RemoveWorkerTags                 Type = 48
+	ReinitializeCertificateAuthority Type = 49
+	ReadCertificateAuthority         Type = 50
```
jefferai, line 61: You didn't follow the comments below! :-)

jefferai (changes requested):
See comments -- this mostly looks great, just some inconsistencies to clear up and some cases where I think it could be simplified.

internal/types/action/action.go
irenarindos: I think it's asking for an update to the wrong test:
https://github.com/hashicorp/boundary/blob/main/internal/perms/acl_test.go#L391

vs what I think it should be, TestActions
https://github.com/hashicorp/boundary/blob/main/internal/types/action/action_test.go#L9

I had looked at Test_AnonRestrictions and assumed I was covered since I thought it should fail for the anon user. 
(Please let me know if I'm misunderstanding anything)

internal/server/service_rotate_roots.go
irenarindos: thanks!

internal/server/root_certificate.go
irenarindos: I'll update it to CaId- I needed the const to be exported

internal/server/service_reinitialize_roots.go
irenarindos: I'll update nodeenrollment to support the new option

internal/types/action/action.go
jefferai: It's missing from TestActions too! But no, it's correct about Test_AnonRestrictions. In the for loop in the test, it iterates based on the numeric value of the action. Right now it maxes out at RemoveHostSources; it needs to be upped to max out at ReadCertificateAuthority.

(There's an additional bug in that test in that the resource wasn't updated to max out at Credential instead of CredentialLibrary).

irenarindos: Ohhh I see that now. I'll fix those.

internal/server/service_reinitialize_roots.go
irenarindos: https://github.com/hashicorp/nodeenrollment/pull/28

internal/daemon/controller/handlers/workers/worker_service.go
```diff
@@ -612,6 +668,45 @@ func (s Service) authResult(ctx context.Context, id string, a action.Type) auth.
...
+		return nil, err
+	}
+
+	return certs, nil
+}
+
+func certificateAuthorityToProto(in *types.RootCertificates) *pb.CertificateAuthority {
+	certs := make([]*pb.Certificate, 2)
+
+	next := in.GetNext()
```
jefferai, line 689: Any reason you return Next first? I would think we'd want to go in chronological order, which would mean returning Current first.

irenarindos: No reason! I'll swap it so current is returned first.

internal/daemon/controller/handlers/workers/worker_service_test.go
```diff
@@ -1791,3 +1850,157 @@ func TestService_RemoveWorkerTags(t *testing.T) {
...
+				if tc.wantErrIs != nil {
+					assert.ErrorIs(err, tc.wantErrIs)
+				}
+				if tc.wantErrContains != "" {
+					assert.Contains(err.Error(), tc.wantErrContains)
+				}
+				return
+			}
+			require.NoError(err)
+			require.NotNil(got)
```
jefferai, line 2003: It'd probably be good to read and store the public keys from before the reinitialize call (just by running a Read on the CA), then compare to ensure that both public keys are now different.

jefferai: Oh, I see that in the test below. Never mind!

jefferai (approved):

PR #2315, opened 2022-08-03
Don't set up Go for enos-run workflow

ryancragun:
Signed-off-by: Ryan Cragun <me@ryan.ec>


moduli (changes requested):
I don't think we want this anymore? The new end-to-end test suite that is kicked off by enos requires Go.

PR #2322, opened 2022-08-04
Add Boundary ALB module

manthonygfp:
This change adds a module that can add an ALB to a VPC for use with communicating with Boundary over the internet.

### DIFF
--- a/enos/modules/alb_boundary/main.tf
+++ b/enos/modules/alb_boundary/main.tf
@@ -16,6 +16,10 @@ resource "aws_security_group" "boundary_ingress_sg" {
     protocol    = "-1"
     cidr_blocks = ["0.0.0.0/0"]
   }
+
+  tags = {
+    Name = var.cluster_name
+  }
 }
 
 resource "aws_lb" "boundary_clients_ingress" {
@@ -24,6 +28,9 @@ resource "aws_lb" "boundary_clients_ingress" {
   load_balancer_type = "application"
   security_groups    = [aws_security_group.boundary_ingress_sg.id]
   subnets            = [for k, _ in var.vpc_subnets : k]
+  tags = {
+    Name = var.cluster_name
+  }
 }
 
 resource "aws_lb_listener" "boundary_listener" {
@@ -38,6 +45,10 @@ resource "aws_lb_listener" "boundary_listener" {
         arn = aws_lb_target_group.boundary_clients.arn
       }
     }
+  }
+
+  tags = {
+    Name = var.cluster_name
   }
 }
 
@@ -54,6 +65,10 @@ resource "aws_lb_target_group" "boundary_clients" {
     # Mark healthy if redirected
     matcher = "200,301,302"
   }
+
+  tags = {
+    Name = var.cluster_name
+  }
 }
 
 resource "aws_lb_target_group_attachment" "boundary_clients" {
@@ -66,4 +81,8 @@ resource "aws_lb_target_group_attachment" "boundary_clients" {
 
 output "alb_address" {
   value = "http://${aws_lb.boundary_clients_ingress.dns_name}:${local.boundary_port}"
+}
+
+output "cluster_name_tag" {
+  value = var.cluster_name
 }
--- a/enos/modules/alb_boundary/main.tf
+++ b/enos/modules/alb_boundary/main.tf
@@ -18,20 +18,12 @@ resource "aws_security_group" "boundary_ingress_sg" {
   }
 }
 
-# data "aws_subnets" "default_subnets" {
-#   filter {
-#     name   = "vpc-id"
-#     values = [var.vpc_id]
-#   }
-# }
-
 resource "aws_lb" "boundary_clients_ingress" {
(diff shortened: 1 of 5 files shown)


enos/modules/alb_boundary/main.tf
```diff
@@ -0,0 +1,69 @@
...
+      target_group {
+        arn = aws_lb_target_group.boundary_clients.arn
+      }
+    }
+  }
+}
+
+resource "aws_lb_target_group" "boundary_clients" {
+  name     = "nomad-clients"
+  port     = 80
```
joshbrand, line 46: the targets are gonna use 9200 by default, maybe we should just use that everywhere instead of 80? or both?

enos/modules/alb_boundary/main.tf
```diff
@@ -0,0 +1,69 @@
...
+  protocol = "HTTP"
+  vpc_id   = var.vpc_id
+
+  # TODO: what should the health check be?
+  health_check {
+    port = 80
+    path = "/"
```
joshbrand, line 53: this -should- be okay. there's also a "ops" listener on port 9203 that has a /health path

joshbrand (approved):

enos/modules/alb_boundary/main.tf
joshbrand: https://www.boundaryproject.io/docs/oss/operations/health

enos/modules/alb_boundary/main.tf
```diff
@@ -0,0 +1,69 @@
...
+  egress {
+    from_port   = 0
+    to_port     = 0
+    protocol    = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+}
+
+resource "aws_lb" "boundary_clients_ingress" {
+  name               = "boundary-ingress-alb"
```
mikebaum, line 22: does this name need to be unique. Like what if there is more than one deployed boundary cluster?

enos/modules/alb_boundary/main.tf
mikebaum: yeah I was thinking the same, might just use 9200 all the way down 🐢 🐢 🐢

enos/modules/alb_boundary/main.tf
```diff
@@ -0,0 +1,69 @@
...
+    matcher = "200,301,302"
+  }
+}
+
+resource "aws_lb_target_group_attachment" "boundary_clients" {
+  count = length(var.instance_ids)
+
+  target_group_arn = aws_lb_target_group.boundary_clients.arn
+  target_id        = var.instance_ids[count.index]
+  port             = 80
```
mikebaum, line 64: considering we're not using TLS, can we limit access to the ALB. Not sure it's possible? @joshbrand Any ideas, or is this okay.

mikebaum (commented):
just a few questions.

enos/modules/alb_boundary/main.tf
manthonygfp: (josh and I talked about this in Slack. We decided to go with 9200 all the way through and not use the ops listener 9203 healthcheck for the 9200 routes that include 9200.

enos/modules/alb_boundary/main.tf
manthonygfp:
```
ingress {
    from_port   = 80
    to_port     = 80
    protocol    = "tcp"
    cidr_blocks = var.cidr_blocks
  }
```

the cidr_blocks whitelisted come from something that only gets your local IP earlier in the Enos scenario.

manthonygfp: Reference ⬇️
https://github.com/hashicorp/boundary/blob/manthony/nomad-deployment/enos/enos.hcl#L343

https://github.com/hashicorp/boundary/blob/manthony/nomad-deployment/enos/enos.hcl#L109

https://github.com/hashicorp/boundary/blob/manthony/nomad-deployment/enos/modules/aws_nomad/outputs.tf#L27

https://github.com/hashicorp/boundary/blob/manthony/nomad-deployment/enos/modules/aws_nomad/locals.tf#L7

https://github.com/hashicorp/boundary/blob/manthony/nomad-deployment/enos/modules/aws_nomad/main.tf#L12

Original source ⬆️

mikebaum: right, of course, missed that... awesome.

enos/modules/alb_boundary/main.tf
manthonygfp: The ALB should have an ID that makes it distinguishable. Do you think we should also add a random resource to additionally name this in a unique way?

manthonygfp: We decided to tag all resources that could be, and this has been done in the new commit.

enos/modules/alb_boundary/main.tf
manthonygfp: I would like to come back after this PR and add the better healthcheck for bonus points.

mikebaum (approved):
looks good to me.

PR #2331, opened 2022-08-07
Add support for SSH private key passphrases

jefferai:


jefferai:
(Note: hitting a flaky cluster test at times, which can be ignored for now.)

louisruch (approved):
LGTM, I found a couple of minor issues but added a commit instead of a review. I think it would be good for @jimlambrt to review as an outside observer before we merge.

jefferai:
Changes look good, I forgot to back out those extra field names, thanks for catching that!

internal/db/schema/migrations/oss/postgres/42/01_ssh_private_key_passphrase.up.sql
```diff
@@ -0,0 +1,101 @@
+begin;
+
+alter table credential_static_ssh_private_key_credential
+  add column private_key_passphrase_encrypted bytea;
+
+alter table credential_static_ssh_private_key_credential
+  add column private_key_passphrase_hmac bytea;
```
mgaffney, line 7: * Can these be combined into a single `alter table statement`?
* If there is no password, can we use a sentinel value instead of a null?
* Can we add a constraint check to make sure both values are set or neither value is set?

mgaffney (changes requested):
I only reviewed the sql changes so far

ddebko (approved):
LGTM

internal/db/schema/migrations/oss/postgres/42/01_ssh_private_key_passphrase.up.sql
mgaffney, line 7:
```suggestion
alter table credential_static_ssh_private_key_credential
  add column private_key_passphrase_encrypted bytea,
  add column private_key_passphrase_hmac bytea,
  add constraint private_key_passphrase_both_null_or_not_null
    check (
      (private_key_passphrase_encrypted is null and private_key_passphrase_hmac is null)
      or
      (private_key_passphrase_encrypted is not null and private_key_passphrase_hmac is not null)
    )
  ;
```

mgaffney (approved):

mgaffney (approved):

PR #2335, opened 2022-08-08
feat(perms): Derive permissions list from user grants

hugoamvieira:
These changes are a building block to enable us to improve Boundary resource listing performance.

### DIFF
--- a/internal/perms/acl.go
+++ b/internal/perms/acl.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/hashicorp/boundary/internal/types/action"
 	"github.com/hashicorp/boundary/internal/types/resource"
+	"github.com/hashicorp/boundary/sdk/pbs/controller/api/resources/scopes"
 )
 
 const AnonymousUserId = "u_anon"
@@ -25,6 +26,18 @@ type ACLResults struct {
 
 	// This is included but unexported for testing/debugging
 	scopeMap map[string][]Grant
+}
+
+// Permission provides information about the specific
+// resources that a user has been granted access to for a given scope, resource, and action.
+type Permission struct {
+	ScopeId  string // The scope id for which the permission applies.
+	Resource resource.Type
+	Action   action.Type
+
+	ResourceIds []string // Any specific resource ids that have been referred in the grant's `id` field, if applicable.
+	OnlySelf    bool     // The grant only allows actions against the user's own resources.
+	All         bool     // We got a wildcard in the grant string's `id` field.
 }
 
 // Resource defines something within boundary that requires authorization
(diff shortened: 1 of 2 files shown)


internal/perms/acl.go
```diff
@@ -27,6 +28,16 @@ type ACLResults struct {
...
+// ListPermission objects are built off of a set of Boundary grants, and
+// contain the required data to enable database query optimization,
+// specifically for Boundary resource listing.
+type ListPermission struct {
+	ScopeId     string   // The scope id for which the permission applies.
+	ResourceIds []string // Any specific resource ids that have been referred in the grant's `id` field, if applicable.
+	OnlySelf    bool     // The grant only allows actions against the user's own resources.
+	All         bool     // We got a wildcard in the grant string's `id` field.
+}
+
```
tmessi, line 40: I am thinking that maybe something like this would make sense:

```suggestion
// Permission provides information about the specific
// resources that a user has been granted access to for a given scope, resource, and action.
type Permission struct {
	ScopeId     string   // The scope id for which the permission applies.
	Resource resrouce.Type
	Action action.Type

	ResourceIds []string // Any specific resource ids that have been referred in the grant's `id` field, if applicable.
	OnlySelf    bool     // The grant only allows actions against the user's own resources.
	All         bool     // We got a wildcard in the grant string's `id` field.
}
```

internal/perms/acl.go
```diff
@@ -209,6 +220,69 @@ func (a ACL) Allowed(r Resource, aType action.Type, userId string, opt ...Option
 	return
 }
 
+// AllowedListConditions builds a set of user permissions (according to their grants)
+// for resource listing ahead of DB access.
+// This enables list query optimization: instead of having to query all data and then drop
+// what the requestor is not authorized to see, we can use the data in the Permission objects
+// to build a more efficient DB query.
+// It does not (and is not meant to) override current authorization mechanisms.
+func (a ACL) AllowedListConditions(scopes map[string]*scopes.ScopeInfo, r resource.Type, aSet action.ActionSet) []ListPermission {
```
tmessi, line 229: Then this could be:

```suggestion
// ListPermissions builds a set of Permissions based on the grants in the ACL for the List action for the given scpoes and resource type. 
func (a ACL) ListPermissions(scopes map[string]*scopes.ScopeInfo, r resource.Type, aSet action.ActionSet) []ListPermission {
```

tmessi (commented):
Looks good, added some thoughts on naming.

internal/perms/acl.go
hugoamvieira: Yep, sounds good! [bb61302](https://github.com/hashicorp/boundary/pull/2335/commits/bb61302c23afbdd45259a44f39711299a2fc450e)

internal/perms/acl.go
hugoamvieira: Changed in [bb61302](https://github.com/hashicorp/boundary/pull/2335/commits/bb61302c23afbdd45259a44f39711299a2fc450e)

internal/perms/acl.go
```diff
@@ -209,6 +222,64 @@ func (a ACL) Allowed(r Resource, aType action.Type, userId string, opt ...Option
 	return
 }
 
+// ListPermissions builds a set of Permissions based on the grants in the ACL for the List action, for the requested scopes and resource type.
+func (a ACL) ListPermissions(requestedScopes map[string]*scopes.ScopeInfo, requestedType resource.Type, requestedActions action.ActionSet) []Permission {
+	perms := make([]Permission, 0, len(requestedScopes))
+	for scopeId := range requestedScopes {
+		p := Permission{ScopeId: scopeId}
```
tmessi, line 229:
```suggestion
		p := Permission{
			ScopeId: scopeId,
			Resource: requestedType,
			Action: action.List,
		}
```

tmessi (commented):
one thing, otherwise looks good

internal/perms/acl.go
hugoamvieira: Ha, good catch, thanks! Addressed in [e854b6b](https://github.com/hashicorp/boundary/pull/2335/commits/e854b6beaff280ea7e1db104a0f6da8a01aceb89)

tmessi (approved):

PR #2336, opened 2022-08-08
feat(auth): Get all authorised scope info for a given list request

hugoamvieira:
These changes are a building block to enable us to improve Boundary resource listing performance.

### DIFF
--- a/internal/daemon/controller/auth/auth.go
+++ b/internal/daemon/controller/auth/auth.go
@@ -13,6 +13,7 @@ import (
 	"github.com/hashicorp/boundary/internal/errors"
 	authpb "github.com/hashicorp/boundary/internal/gen/controller/auth"
 	"github.com/hashicorp/boundary/internal/gen/controller/tokens"
+	"github.com/hashicorp/boundary/internal/iam"
 
 	"github.com/hashicorp/boundary/internal/daemon/controller/common"
 	"github.com/hashicorp/boundary/internal/daemon/controller/handlers"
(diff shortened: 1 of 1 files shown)


internal/daemon/controller/common/scopeids/scope_ids.go
```diff
@@ -353,3 +353,143 @@ func GetListingScopeIds(
...
+
+// AuthorizedForList retrieves and returns all scopes where a user is authorized
+// to perform a *list* action on. It looks recursively from `rootScopeId`.
+func AuthorizedForList(
+	ctx context.Context, // The context to use when listing in the DB, if required
+	iamRepoFn common.IamRepoFactory, // An IAM repo function to use for a scope listing call
+	authResults auth.VerifyResults, // The original auth results from the list command
+	rootScopeId string, // The scope ID to use, or the starting point for a recursive search
+	resourceType resource.Type, // The type of resource being listed
+	recursive bool, // Whether the search is recursive.
```
tmessi, line 365: I think we should drop this, and instead any calling code should just not call this function if it is a non-recursive call.

internal/daemon/controller/common/scopeids/scope_ids.go
```diff
@@ -353,3 +353,143 @@ func GetListingScopeIds(
...
+
+	// This will be used to save scope info so we can put the right scope
+	// info for each returned value
+	scopeResourceMap := make(map[string]*scopes.ScopeInfo)
+
+	// Base case: if not recursive, return the scope we were given and the
+	// already-looked-up info.
+	if !recursive {
+		return map[string]*scopes.ScopeInfo{authResults.Scope.Id: authResults.Scope}, nil
+	}
```
tmessi, line 391: likewise this can be dropped if we drop the `recursive` arg

internal/daemon/controller/common/scopeids/scope_ids.go
```diff
@@ -353,3 +353,143 @@ func GetListingScopeIds(
...
+				return nil, errors.New(ctx, errors.Internal, op, "unexpected action in set")
+			}
+			if scopeResourceMap[scpId] == nil {
+				scopeResourceMap[scpId] = &scopes.ScopeInfo{
+					Id:            scp.GetPublicId(),
+					Type:          scp.GetType(),
+					Name:          scp.GetName(),
+					Description:   scp.GetDescription(),
+					ParentScopeId: scp.GetParentId(),
+				}
```
tmessi, line 439: We can come back to this, but just to make a note of it, I think ideally this transformation would happen elsewhere within the application services, and this function would instead return a `map[string]*iam.Scope`

internal/daemon/controller/common/scopeids/scope_ids.go
```diff
@@ -353,3 +353,143 @@ func GetListingScopeIds(
 	}
 	return scopeResourceInfo.ScopeIds, scopeMap, nil
 }
+
+// AuthorizedForList retrieves and returns all scopes where a user is authorized
+// to perform a *list* action on. It looks recursively from `rootScopeId`.
+func AuthorizedForList(
+	ctx context.Context, // The context to use when listing in the DB, if required
+	iamRepoFn common.IamRepoFactory, // An IAM repo function to use for a scope listing call
+	authResults auth.VerifyResults, // The original auth results from the list command
```
tmessi, line 362: I am wondering if this function should just be a method on `auth.VerifyResults`. I don't know if that causes some import issues, but I think it would work since, an `auth.VerifyResults` already has a `common.IamRepoFactory` via the [verifier](https://github.com/hashicorp/boundary/blob/a21ff8b00476fad4238e98295673a70ca59e3532/internal/daemon/controller/auth/auth.go#L94). That would turn this into something like:

```go
func (r *VerifyResults) ScopesAuthorizedForList(ctx context.Context, rootScopeId string, resourceType resource.Type) (map[string]*scopes.ScopeInfo, error)
```

internal/daemon/controller/common/scopeids/scope_ids.go
```diff
@@ -353,3 +353,143 @@ func GetListingScopeIds(
 	}
 	return scopeResourceInfo.ScopeIds, scopeMap, nil
 }
+
+// AuthorizedForList retrieves and returns all scopes where a user is authorized
+// to perform a *list* action on. It looks recursively from `rootScopeId`.
+func AuthorizedForList(
```
tmessi, line 359: Adding a note for us to follow up with some unit tests for this prior to merging the llb

internal/daemon/controller/common/scopeids/scope_ids.go
hugoamvieira: I shall give this a try!

hugoamvieira: Works great, and it's actually more helpful because what we have in the service functions is a `VerifyResults` object - [771086a](https://github.com/hashicorp/boundary/pull/2336/commits/771086aaacde57756d84ea6bfe19f917ee4fb42c)

internal/daemon/controller/common/scopeids/scope_ids.go
hugoamvieira: Yep, addressed in [9fd0fe9](https://github.com/hashicorp/boundary/pull/2336/commits/9fd0fe983420986c11fe4364a2ff7a0862284608)

internal/daemon/controller/common/scopeids/scope_ids.go
hugoamvieira: https://github.com/hashicorp/boundary/pull/2336#discussion_r941161611

internal/daemon/controller/common/scopeids/scope_ids.go
hugoamvieira: After merge here into the LLB, I'll link these comments over 👍

internal/daemon/controller/common/scopeids/scope_ids.go
hugoamvieira: Sure, I'll link this comment over to the LLB when this branch is merged

internal/daemon/controller/auth/auth.go
```diff
@@ -694,3 +695,123 @@ func GetTokenFromRequest(ctx context.Context, kmsCache *kms.Kms, req *http.Reque
 
 	return publicId, encryptedToken, uint32(receivedTokenType)
 }
+
+// ScopesAuthorizedForList retrieves and returns all scopes where a user is authorized
+// to perform a *list* action on. It looks recursively from `rootScopeId`.
+func (r *VerifyResults) ScopesAuthorizedForList(ctx context.Context, rootScopeId string, resourceType resource.Type) (map[string]*scopes.ScopeInfo, error) {
+	const op = "scopeids.AuthorizedForList"
```
tmessi, line 702:
```suggestion
	const op = "auth.(VerifyResults).ScopesAuthorizedForList"
```

tmessi (commented):
one thing, otherwise looks good

internal/daemon/controller/auth/auth.go
hugoamvieira: There's always something I miss 😆, thanks for spotting, addressed in [8d45a6a](https://github.com/hashicorp/boundary/pull/2336/commits/8d45a6a21cf97b37362066fadcfbc8380019df3c)

tmessi (approved):

tmessi:
This commit should probably be `feat(auth)` now

hugoamvieira:
Yep good point, I'll do a reword with the soon 👍

PR #2337, opened 2022-08-08
Remove static names from module

manthonygfp:
* Fix .gitignore to standard for Enos and TF

### DIFF
--- a/enos/.gitignore
+++ b/enos/.gitignore
@@ -1,16 +1,3 @@
-# Binaries for programs and plugins
-*.exe
-*.exe~
-*.dll
-*.so
-*.dylib
-
-# Test binary, built with `go test -c`
-*.test
-
-# Output of the go coverage tool, specifically when used with LiteIDE
-*.out
-
 # Local .terraform directories
 **/.terraform/*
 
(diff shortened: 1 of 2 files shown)


enos/.gitignore
```diff
@@ -1,40 +1,43 @@
...
+# Local Enos directories
+**/.enos/*
 
-# Enos vars files
-enos*.vars.hcl
+# Exclude all .vars.hcl files, which are likely to contain sensitive data, such as
+# password, private keys, and other secrets. These should not be part of version 
+# control as they are data points which are potentially sensitive and subject 
+# to change depending on the environment.
+*.vars.hcl
```
manthonygfp, line 43: This .gitignore was not standard for TF/Enos for some reason.

enos/modules/alb_boundary/main.tf
```diff
@@ -1,5 +1,4 @@
 resource "aws_security_group" "boundary_ingress_sg" {
-  name   = "${var.lb_name_suffix}-server-lb"
```
manthonygfp, line 2: Names like these are not required and the provider will assign random, unique names for them. Other values here had sane defaults and their declaration can be removed from our Terraform module.

joshbrand (approved):

PR #2341, opened 2022-08-09
Remove unused variable

manthonygfp:
This PR is for cosmetic changes. The ALB module had an unused variable and the Nomad job file had extra newlines.


manthonygfp:
I have also now added a change that adds a restart stanza and count. Previously, Boundary was only deployed to one node.

joshbrand (approved):

joshbrand (approved):

PR #2319, opened 2022-08-04
Target List request performance optimisation

hugoamvieira:
This PR is the `Targets` counterpart of https://github.com/hashicorp/boundary/pull/2381 and contains a set of commits designed to improve Boundary's performance when listing Target resources.

PR #2342, opened 2022-08-09
Boundary Resource listing Performance

hugoamvieira:
This LLB contains a set of commits geared towards improving Boundary's performance when listing resources.


hugoamvieira:
https://github.com/hashicorp/boundary/pull/2336#discussion_r940506075

tmessi:
@jefferai this contains the commits from #2381 (which you already reviewed) as well as the commits from #2319 (which I don't think you have gotten to yet).

PR #2353, opened 2022-08-11
Fix error validation code in attr parsing

jefferai:
For a flag type used for attributes/secrets, values have always been
able to be used with env:// or file:// format, but the error check to
see whether the value was successfully parsed was doing a simple string
equality rather than `errors.Is`. As a result a value containing a colon
would trip up this check and cause an error, rather than allowing it to
be used as a bare value.

Fixes ICU-5629


johanbrandhorst (approved):
_nice_

PR #2354, opened 2022-08-11
Enos: break out init job, cleanup

joshbrand:
* Move jq deps to our earlier dependency installation
* Add more guards around installation
* Try and make job module more generic-er for the future
* Make controller group count configurable


manthonygfp (approved):
hell yeah brother

PR #2357, opened 2022-08-12
feature (downstreams): Add hooks for optional downstream workers/router

jimlambrt:
This includes the required interfaces, factories, error codes, etc


jimlambrt:
need to fix something before re-opening this

PR #2360, opened 2022-08-12
Add Rotate Keys Endpoint to Scopes Api

DanielleMiu:
this PR is reliant on [boundry#2371](https://github.com/hashicorp/boundary/pull/2371)

add the rotate keys endpoint for generating new key versions via scope id


internal/proto/controller/api/services/v1/scope_service.proto
```diff
@@ -99,6 +99,20 @@ service ScopeService {
...
   }
+
+  // RotateKeys rotates and optionally rewraps all the keys found in the
+  // scope specified. If the scope is not found an error is returned. If
+  // the scope is empty, the global scope is used.
+  rpc RotateKeys(RotateKeysRequest) returns (RotateKeysResponse) {
+    option (google.api.http) = {
+      post: "/v1/scopes:rotate-keys"
+      body: "*"
+      // response_body: "item" // does this need to be here for an empty response?
```
johanbrandhorst, line 110: No, this is for when you want your response body to be unwrapped from the wrapper response message. It does not apply in this case.

internal/db/schema/migrations/oss/postgres/42/05_key_rewrap.up.sql
```diff
@@ -0,0 +1,30 @@
...
+
+
+-- we need to make the key and version columns mutable in order to support
+-- rewrapping the root key versions.
+drop trigger kms_immutable_columns on kms_root_key_version;
+
+create trigger kms_immutable_columns
+before
+update on kms_root_key_version
+  for each row execute procedure kms_immutable_columns('private_id', 'root_key_id', 'create_time');
```
mgaffney, line 17: Please reformat to follow our [sql style guide](https://github.com/hashicorp/boundary/blob/main/sql-style-guide.md#triggers).

internal/db/schema/migrations/oss/postgres/42/05_key_rewrap.up.sql
```diff
@@ -0,0 +1,30 @@
...
+
+
+-- we need to make the key and version columns mutable in order to support
+-- rewrapping the data key version.
+drop trigger kms_immutable_columns on kms_data_key_version;
+
+create trigger kms_immutable_columns
+before
+update on kms_data_key_version
+  for each row execute procedure kms_immutable_columns('private_id', 'data_key_id', 'root_key_version_id', 'create_time');
```
mgaffney, line 28: Please reformat to follow our [sql style guide](https://github.com/hashicorp/boundary/blob/main/sql-style-guide.md#triggers).

mgaffney (changes requested):

internal/db/schema/migrations/oss/postgres/42/05_key_rewrap.up.sql
DanielleMiu: this file is from [go-kms-wrapping](https://github.com/hashicorp/go-kms-wrapping/blob/main/extras/kms/migrations/postgres/05_key_rewrap.up.sql) and should probably not be modified unless necessary

internal/db/schema/migrations/oss/postgres/42/05_key_rewrap.up.sql
```diff
@@ -0,0 +1,26 @@
+
+-- make the required schema changes to upgrade the dependency:
```
mgaffney, line 2: * This migration should be in a new migration directory since directory 42 has already been released. 
* The number in the filename needs to be changed if the new migration directory already has an '05' migration file

mgaffney (changes requested):

internal/db/schema/migrations/oss/postgres/42/05_key_rewrap.up.sql
johanbrandhorst: We accidentally missed this when rebasing, but we're going to make this change in a separate PR instead since it should be fixed in `main`, not just on this branch.

mgaffney: Why would this change already be on main?

johanbrandhorst: Sorry, I mean that it _should_ be fixed in `main`, not that it is already fixed on `main`

johanbrandhorst: https://github.com/hashicorp/boundary/pull/2371 will track the merging, it's just blocked right now

internal/daemon/controller/handlers/scopes/scope_service.go
```diff
@@ -412,6 +412,26 @@ func (s Service) ListKeys(ctx context.Context, req *pbs.ListKeysRequest) (*pbs.L
...
+	authResults := s.authResult(ctx, req.GetId(), action.RotateScopeKeys)
+	if authResults.Error != nil {
+		return nil, authResults.Error
+	}
+
+	if err := s.kmsRepo.RotateKeys(ctx, req.GetId(), kms.WithRewrap(req.GetRewrap())); err != nil {
+		return nil, err
+	}
+
+	return nil, nil
```
johanbrandhorst, line 432: I think this may panic the gRPC handles, you want to return an empty message even if there is nothing in it
```suggestion
	return &pbs.RotateKeysResponse{}, nil
```

internal/daemon/controller/handlers/scopes/scope_service_test.go
```diff
@@ -1727,3 +1727,137 @@ func TestListKeys(t *testing.T) {
...
+
+	org.Name = "defaultOrg"
+	org.Description = "defaultOrg"
+	org, _, err := tc.IamRepo().UpdateScope(context.Background(), org, 1, []string{"Name", "Description"})
+	require.NoError(t, err)
+
+	// Add new role for listing+rotating keys in org
+	rotateKeysRole := iam.TestRole(t, tc.DbConn(), org.PublicId)
+	_, err = tc.IamRepo().AddRoleGrants(context.Background(), rotateKeysRole.PublicId, 1, []string{"id=*;type=*;actions=rotate-keys,list-keys"})
+	require.NoError(t, err)
```
johanbrandhorst, line 1782: I know this is probably copied from the ListKeys tests, but could you look into using `iam.TestRoleGrant` instead to assign grants to roles? I think it should make this code a little easier to read, and I didn't know about it when I wrote the ListKeys tests.

internal/daemon/controller/handlers/scopes/scope_service_test.go
```diff
@@ -1727,3 +1727,137 @@ func TestListKeys(t *testing.T) {
...
+		{
+			name:    "successfully Rotate keys in org",
+			req:     &pbs.RotateKeysRequest{Id: org.GetPublicId(), Rewrap: false},
+			authCtx: privCtx,
+		},
+		{
+			name:    "successfully Rotate keys in project",
+			req:     &pbs.RotateKeysRequest{Id: proj.GetPublicId(), Rewrap: true},
+			authCtx: privCtx,
+		},
```
johanbrandhorst, line 1826: Maybe split this out into (org, rewrap=false), (org, rewrap=true), (proj, rewrap=false), (proj, rewrap=true)?

internal/proto/controller/api/services/v1/scope_service.proto
```diff
@@ -153,3 +166,10 @@ message ListKeysRequest {
 message ListKeysResponse {
   repeated resources.scopes.v1.Key items = 1;
 }
+
+message RotateKeysRequest {
+  string id = 1;
```
johanbrandhorst, line 171: Maybe make this `scope_id` since it's going into the request body rather than the path?

johanbrandhorst (changes requested):

internal/daemon/controller/handlers/scopes/scope_service.go
DanielleMiu: leaving as-is to keep parity with DeleteScope endpoint

johanbrandhorst (approved):
LGTM

PR #2363, opened 2022-08-15
Ensure that "null" can only be specified in "set-" commands

jefferai:
This allows a check function to be set to determine whether it is
allowed to be set. This also ensures that if null is allowed it is
either the only value or not set.

### DIFF
--- a/internal/cmd/base/flags.go
+++ b/internal/cmd/base/flags.go
@@ -548,7 +548,11 @@ func appendDurationSuffix(s string) string {
 	return s + "s"
 }
 
-// -- StringSliceVar and stringSliceValue
+// StringSliceVar reads in parameters from the same flag into a string array.
+// Setting NullCheck enables the following behavior: the function will be run
+// whenever a "null" is seen in the input to determine whether or not it is
+// allowed (and erroring if not); and even if allowed generally, the flag will
+// ensure that "null" is the only value passed.
 type StringSliceVar struct {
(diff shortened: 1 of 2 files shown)


internal/cmd/base/flags.go
```diff
@@ -581,25 +586,53 @@ func (f *FlagSet) StringSliceVar(i *StringSliceVar) {
...
+			// error; we don't check if _all_ are "null" since presumably this
+			// function prevents there being more values than just "null" if
+			// that's specified
+			if len(*s.target) > 0 && (*s.target)[0] != "null" {
+				return fmt.Errorf(`"null" cannot be combined with other values`)
+			}
+			// Set the target to only contain "null" and return
+			*s.target = []string{"null"}
+			return nil
+			// Something came in that isn't "null" but we already have "null"
```
A-440Hz, line 629: nit: can we move this comment back one tab or start it on a new line? (unless this way is already an established precedent)

A-440Hz (approved):
lgtm!

internal/cmd/base/flags.go
jefferai: Yeah, I had done that and didn't notice that autoformatting pulled it forward. I moved it down into the if clause itself.

mgaffney (approved):
Nice!

jefferai:
Merging for now; this can be rolled out in bits.

PR #2366, opened 2022-08-16
Add internal LB routing

manthonygfp:
With this change, traffic follows the following pattern:

Client -> ALB -> Træfik -> Boundary

This allows external requests to Boundary to be routed to any Boundary container, no matter where it is currently scheduled.

Of note, this change involves the following:
* Boundary containers are now port mapped from a random ip outside the container to a static IP of 9200 in the container
* One instance Traefik is now running on the Nomad cluster
* The ALB routes traffic to the active Traefik node which then internally routes the request to a Boundary container that it looked up as in the Consul catalog of services


enos/modules/alb_boundary/main.tf
```diff
@@ -2,11 +2,15 @@ resource "aws_security_group" "boundary_ingress_sg" {
   vpc_id = var.vpc_id
 
   # Boundary
-  ingress {
-    from_port   = local.boundary_port
-    to_port     = local.boundary_port
-    protocol    = "tcp"
-    cidr_blocks = var.cidr_blocks
+  dynamic "ingress" {
```
joshbrand, line 5: NICE

joshbrand (approved):
gimmiegimmiegimmie

joshbrand:
might want to do an `enos fmt` before merge (see failed check)

joshbrand:
oh, and was there a scenario change at all?

mikebaum:
I tried running the scenario and it complained about missing some required input vars for the `provision_nomd_cluster` step.

scotthain:
This is done

manthonygfp:
@joshbrand integrated this into a different repo

PR #2370, opened 2022-08-17
Update Dockerfile to use new PRODUCT_VERSION arg

johanbrandhorst:
This is necessary since the merge of https://github.com/hashicorp/actions-docker-build/pull/25

Note that our `build-docker` job uses the `version` input correctly, I think: https://github.com/hashicorp/boundary/blob/369f818f4c39f303e4d41af8130982b765b1c18d/.github/workflows/build.yml#L316-L344.


mdeggies (approved):
This looks great. Thanks Johan!

PR #2371, opened 2022-08-17
Add SQL Migration 05 From go-kms-wrapping

DanielleMiu:
add in the [missing sql migration file](https://github.com/hashicorp/go-kms-wrapping/blob/main/extras/kms/migrations/postgres/05_key_rewrap.up.sql) to boundary to enable key rewrapping in the kms and scopes repositories.


johanbrandhorst (changes requested):
Just making sure we don't merge this until main is unfozen

internal/db/schema/migrations/oss/postgres/44/05_key_rewrap.up.sql
```diff
@@ -0,0 +1,23 @@
+-- make the required schema changes to upgrade the dependency:
```
mgaffney, line 1: Given that there are no other migrations in `44`, the filename should be `01_key_rewrap.up.sql`.

DanielleMiu: fixed

johanbrandhorst (approved):
Migration code freeze is over

internal/db/schema/migrations/oss/postgres/44/01_key_rewrap.up.sql
```diff
@@ -0,0 +1,23 @@
+-- make the required schema changes to upgrade the dependency:
```
mgaffney, line 1: File needs to be moved to migration directory `47`

mgaffney (changes requested):

internal/db/schema/migrations/oss/postgres/44/01_key_rewrap.up.sql
DanielleMiu: done

johanbrandhorst:
This should be good to merge again, assuming we've fixed the migration issue with 0.10.2

mgaffney (approved):

PR #2374, opened 2022-08-18
fix: Move enos job attribute into enos-run workflow

johanbrandhorst:
Maybe this will fix it? This doesn't make any sense.

Apparently the last fix was wrong too: https://github.com/hashicorp/boundary/actions/runs/2879194545. I think this should be it!


louisruch (approved):
Seems sane…

PR #2375, opened 2022-08-18
Key revocation prototype

johanbrandhorst:


internal/kms/rewrapping.go
```diff
@@ -0,0 +1,68 @@
...
+		on u.constraint_catalog = fk.unique_constraint_catalog
+			and u.constraint_schema = fk.unique_constraint_schema
+			and u.constraint_name = fk.unique_constraint_name
+	inner join information_schema.key_column_usage        r
+		on r.constraint_catalog = fk.constraint_catalog
+			and r.constraint_schema = fk.constraint_schema
+			and r.constraint_name = fk.constraint_name
+	where
+		u.column_name = 'private_id' and
+		u.table_name = 'kms_data_key_version'
```
johanbrandhorst, line 47: Technically this "belongs" to the extras/kms/v2 library that we import, so I don't know if we should be using it literally here or expose it from the library. Thoughts?

internal/kms/kms.go
```diff
@@ -436,38 +437,42 @@ func (k *Kms) RunKeyRevocation(ctx context.Context, keyRevocation *KeyRevocation
...
-	select {
-	case <-time.After(3 * time.Second):
-	case <-ctx.Done():
-		fmt.Println("canceled")
-		return ctx.Err()
+	switch {
+	case strings.HasPrefix(keyRevocation.KeyId, "krkv"):
+		// Root key, rewrap keys in scope
+		// TODO: A better way to get scope_id from key version private id
+		// kms doesn't export the type.
```
johanbrandhorst, line 444: I'm leaning towards exporting `RootKeyVersion` from extras/kms/v2 to be able to scan this easily.

internal/auth/oidc/rewrapping.go
```diff
@@ -0,0 +1,86 @@
...
+				if _, err := repo.writer.Update(ctx, am, allAuthMethodFields, nil); err != nil {
+					return err
+				}
+				// TODO: Should we add a sleep here to pace the DB updates?
+			}
+			// TODO: Should we sleep here to pace the DB load?
+		}
+		return nil
+	}
+}
```
johanbrandhorst, line 86: This shows a sample rewrapping function, we'd have to write one for each table that depends on the data keys. The current list is:
```
"auth_oidc_method",
"credential_static_ssh_private_key_credential",
"credential_static_username_password_credential",
"credential_vault_client_certificate",
"credential_vault_token",
"host_plugin_catalog_secret",
"session_credential",
"worker_auth_authorized",
"worker_auth_ca_certificate",
```

internal/daemon/controller/kms_revocation_registration_test.go
```diff
@@ -0,0 +1,26 @@
...
+// so any packages that will register a rewrapping fn should have been imported here.
+func TestAllDataKeyReferencersHaveARewrappingFn(t *testing.T) {
+	ctx := context.Background()
+	tc := controller.NewTestController(t, nil)
+	t.Cleanup(tc.Shutdown)
+	tableNames, err := tc.Kms().ListDataKeyReferencers(ctx)
+	require.NoError(t, err)
+	registeredTableNames := kms.ListTablesSupportingRewrapping()
+	require.Empty(t, cmp.Diff(tableNames, registeredTableNames, cmpopts.SortSlices(func(i, j string) bool { return i < j })), "At least one table referencing a data key does not have a rewrapping function registered")
+}
```
johanbrandhorst, line 26: This test should start failing if we add a new data key dependency without a corresponding key rewrapping method

internal/kms/rewrapping.go
```diff
@@ -0,0 +1,68 @@
...
+func RegisterTableRewrappingFn(tableName string, rewrappingFn RewrappingFn) {
+	if _, ok := tableNameToRewrappingFn[tableName]; ok {
+		panic(fmt.Sprintf("rewrapping function for table name %q already exists", tableName))
+	}
+	tableNameToRewrappingFn[tableName] = rewrappingFn
+}
+
+// ListTablesSupportingRewrapping lists all the table names registered with a rewrapping function
+func ListTablesSupportingRewrapping() []string {
+	return maps.Keys(tableNameToRewrappingFn)
```
johanbrandhorst, line 26: Generics 💥

internal/auth/oidc/rewrapping.go
johanbrandhorst: This is missing 3 tables that do not have the foreign key relationships set up:
```
"auth_token",
"auth_password_argon2_cred",
"session",
```

internal/kms/kms.go
jimlambrt: I would hesitate to export this type... since it's not exported in any of that pkgs operations.   You could just Kms.ListKeys(...) which lists the keys for a scope and then find the one with the matching id in the list?

johanbrandhorst:
Closing pending redesign

PR #2377, opened 2022-08-18
feat(workers): Return Release Version on worker list and read

irenarindos:


internal/db/schema/migrations/oss/postgres/99/01_server_worker_release_version.up.sql
```diff
@@ -0,0 +1,53 @@
...
+  w.update_time,
+  w.version,
+  w.last_status_time,
+  w.type,
+  w.release_version,
+  cc.count as active_connection_count,
+  -- keys and tags can be any lowercase printable character so use uppercase characters as delimitors.
+  wt.tags as api_tags,
+  ct.tags as worker_config_tags
+from server_worker w
```
jefferai, line 43: The formatting in this section varies from the version in 34/04. It doesn't seem intentional, so calling it out.

internal/db/schema/migrations/oss/postgres/99/01_server_worker_release_version.up.sql
```diff
@@ -0,0 +1,53 @@
+begin;
+
+  alter table server_worker
+    add column release_version text;
```
jefferai, line 4: You may want to check in with @mgaffney around `not null` here -- I agree that we need to handle unspecified but there may be a preference for a sentinel value, which is already a type that exits.

internal/proto/controller/storage/servers/store/v1/worker.proto
```diff
@@ -54,6 +54,10 @@ message Worker {
   // The type of the worker, denoted by how it authenticates: pki or kms.
   // @inject_tag: `gorm:"not_null"`
   string type = 130;
+
+  // The version of Boundary the worker binary is running
+  // @inject_tag: `gorm:"default:null"`
```
jefferai, line 59: If the type changes to not null the tag here should change too.

internal/server/repository_worker.go
```diff
@@ -292,7 +292,7 @@ func (r *Repository) UpsertWorkerStatus(ctx context.Context, worker *Worker, opt
 				// "description" since we want description changes for PKI-based
 				// workers to come via API only. We can't really guard on this
 				// in the DB so we need to be sure to not include it here.
-				n, err := w.Update(ctx, workerClone, []string{"address"}, nil)
+				n, err := w.Update(ctx, workerClone, []string{"address", "ReleaseVersion"}, nil)
```
jefferai, line 295: It seems odd that here it would be `ReleaseVersion` and below `release_version`. I realize gorm probably ends up making it the same but feels like it should be consistent.

jefferai (commented):
Looking great! Just a few things to look into.

internal/server/repository_worker.go
irenarindos: It is odd but Update fieldMasks and the create on conflict db.SetColumns accept different column name formats. Update will not accept release_version and will error and inversely SetColumns would err on ReleaseVersion. 
It looks like SetColumns expect the actual column name and Update converts field mask paths into column names in go.dbw

A-440Hz (commented):
lgtm

internal/db/schema/migrations/oss/postgres/99/01_server_worker_release_version.up.sql
irenarindos: Thanks! I checked with Mike and he thinks null is appropriate for this instead of using a sentinel.

internal/db/schema/migrations/oss/postgres/99/01_server_worker_release_version.up.sql
irenarindos: Thanks for catching that!

PR #2381, opened 2022-08-22
Improve session list performance

tmessi:


internal/session/repository.go
```diff
@@ -23,36 +29,89 @@ type Repository struct {
...
+
+		where = append(where, fmt.Sprintf("(%s)", strings.Join(clauses, " and ")))
+	}
+	return where, args
+}
+
 // list will return a listing of resources and honor the WithLimit option or the
 // repo defaultLimit.  Supports WithOrder option.
+//
+// TODO: (tmessi) Remove this, it is only used by a test.
```
mgaffney, line 114: Can this TODO be completed?

tmessi: It might need to be a follow up. I am not sure of the best way to refactor the [specific test](https://github.com/hashicorp/boundary/blob/16da8d2388269cbf915f4c7d37787a5874440cf6/internal/session/job_session_cleanup_test.go#L305-L330). It looks like that test was exploiting the fact that this `list` method would allow any interface to be passed in, and that the interface was then being passed ultimately into gorm, so the test's call to `repo.list` was actually populating `[]*session.Connection` which it then used for its assertions.

mgaffney (approved):

hugoamvieira (approved):

internal/session/repository.go
tmessi: I figured this out: [671328a](https://github.com/hashicorp/boundary/pull/2381/commits/671328a4d8497ff851ee0d580eab88461a11a1d1)

jefferai (approved):
This looks absolutely amazing. Wonderful work.

PR #2382, opened 2022-08-22
Enos: add additional guards before authentication

joshbrand:
Previously, we would see failures when the ALB in front of a cluster was alive and passing healthchecks, but the database behind had not been initialized yet.  This adds a check with an unauth'd call to `auth-methods list` to validate the DB is up before auth.

### DIFF
--- a/enos/templates/get-token.sh
+++ b/enos/templates/get-token.sh
@@ -21,11 +21,8 @@ function retry {
   return 0
 }
 
-# make sure the ALB is up and passing healthchecks
-retry 10 curl -s -o /dev/null ${BOUNDARY_ADDR}
-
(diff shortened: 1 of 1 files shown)


ryancragun (approved):
I love the smell of a good race condition in the morning

johanbrandhorst (approved):
This LGTM, but would appreciate thoughts from @hugoamvieira and @talanknight on whether it would be more appropriate to use the health check endpoint for this?

joshbrand:
why-not-both-why-not
healthcheck updates in: https://github.com/hashicorp/terraform-enos-aws-boundary/pull/21, will get pulled in once merged

tmessi (approved):

PR #2384, opened 2022-08-23
Add worker list, read, and update cli ui tests.

talanknight:
This is the first round of worker bats tests.
Follow up PRs will add tests for worker creation, delete, and API tag management.

### DIFF
--- a/internal/tests/cli/boundary/workers.bats
+++ b/internal/tests/cli/boundary/workers.bats
@@ -15,43 +15,43 @@ export NEW_UPDATED_WORKER='newtest'
 }
 
 @test "boundary/workers: can create $NEW_WORKER worker" {
-	run create_worker $NEW_WORKER
+  run create_worker $NEW_WORKER
   echo "$output"
-	[ "$status" -eq 0 ]
+  [ "$status" -eq 0 ]
 }
 
 @test "boundary/workers: can not create already created $NEW_WORKER worker" {
-	run create_worker $NEW_WORKER
+  run create_worker $NEW_WORKER
   echo "$output"
-	[ "$status" -eq 1 ]
+  [ "$status" -eq 1 ]
 }
 
 @test "boundary/workers: can read $NEW_WORKER worker" {
   local wid=$(worker_id $NEW_WORKER)
-	run read_worker $wid
+  run read_worker $wid
   echo "$output"
-	[ "$status" -eq 0 ]
+  [ "$status" -eq 0 ]
 }
 
 @test "boundary/workers: can list workers" {
-	run list_workers
+  run list_workers
   echo "$output"
-	[ "$status" -eq 0 ]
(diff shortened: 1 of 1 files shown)


internal/tests/cli/boundary/workers.bats
```diff
@@ -0,0 +1,44 @@
...
+  echo "$output"
+	[ "$status" -eq 0 ]
+}
+
+@test "boundary/workers: can update worker's name" {
+  local wid=$(worker_id)
+  local name="test"
+    run update_worker $wid $name
+  echo "$output"
+	[ "$status" -eq 0 ]
```
irenarindos, line 26: It looks like this is failing with "cannot update a KMS worker"

irenarindos: I didn't run it locally- it failed on Enos: https://github.com/hashicorp/boundary/runs/7982789871?check_suite_focus=true

internal/tests/cli/boundary/workers.bats
```diff
@@ -0,0 +1,63 @@
...
+export NEW_UPDATED_WORKER='newtest'
+
+@test "boundary/login: can login as default user" {
+  run login $DEFAULT_LOGIN
+  echo "$output"
+  [ "$status" -eq 0 ]
+}
+
+@test "boundary/workers: can create $NEW_WORKER worker" {
+	run create_worker $NEW_WORKER
```
moduli, line 18: nit: Is the indentation in this file intended to be this way?

internal/tests/cli/boundary/workers.bats
```diff
@@ -0,0 +1,63 @@
...
+	[ "$status" -eq 0 ]
+}
+
+@test "boundary/workers: can delete $NEW_UPDATED_WORKER worker" {
+  local wid=$(worker_id $NEW_UPDATED_WORKER)
+  run delete_worker $wid
+  echo "$output"
+  run has_status_code "$output" "204"
+  [ "$status" -eq 0 ]
+}
```
moduli, line 63: I've seen some other bats tests that check that it can't read the deleted object. Not sure what your thoughts are there. 

On the flip side, I *might* be seeing some flakiness in that kind of test. A theory is that there's some sort of timing thing where boundary hadn't finished completed deleting the entity before reading it again.

internal/tests/cli/boundary/_workers.bash
```diff
@@ -0,0 +1,44 @@
...
+  boundary workers create controller-led \
+    -name $1 \
+    -description 'test worker'
+}
+
+function update_worker() {
+  boundary workers update -id $1 -name $2 -version 0
+}
+
+function add_worker_tag() {
```
moduli, line 17: Is this being used anywhere?

moduli (approved):

PR #2387, opened 2022-08-24
feat(cli): Read and reinitialize worker certificate authority

irenarindos:
Addition of CLI functions for the worker certificate authority read and reinitialize methods.


internal/api/genapi/input.go
```diff
@@ -782,6 +782,10 @@ var inputStructs = []*structInfo{
 		fieldFilter:         []string{"private_key"},
 		recursiveListing:    true,
 	},
+	{
+		inProto: &workers.Certificate{},
+		outFile: "workers/certificate.gen.go",
+	},
```
jefferai, line 788:
```suggestion
	},
		{
		inProto: &workers.CertificateAuthority{},
		outFile: "workers/certificate_authority.gen.go",
	},
```

internal/cmd/commands/workerscmd/certificate-authority.go
```diff
@@ -0,0 +1,220 @@
...
+			"",
+			"  Reinitialize the certificate authority used to authorize Boundary workers:",
+			"",
+			`    $ boundary workers certificate-authority reinitialize`,
+			"",
+			"",
+		}) + c.Flags().Help()
+	}
+	return base.WrapForHelpText([]string{
+		"Usage: boundary workers certificate-authority [options]",
```
jefferai, line 68: See internal/cmd/common/help.go  line 44: the general pattern here is to say something high level, then have an example command. It's a subtle difference but one that makes it clear that the user didn't run the command the wrong way just missing `read`, that `read` is simply _one_ example of something that can be done. Then the "Please see the {{hyphentype}}s subcommand..." bit makes it clear to look at the available subcommands.

jefferai (approved):
Looks really good! Just one wording suggestion.

internal/cmd/commands/workerscmd/certificate-authority.go
irenarindos: Thanks, I've updated this!

PR #2388, opened 2022-08-24
bug(workers): Create worker-led panics on invalid token

irenarindos:
Fix panic if invalid token is passed:
```
% boundary workers create worker-led -worker-generated-auth-token=bogus
[signal SIGSEGV: segmentation violation code=0x2 addr=0x8 pc=0x10533ce24]

goroutine 1 [running]:
github.com/hashicorp/boundary/internal/cmd/commands/workerscmd.executeExtraWorkerLedActionsImpl(0x14000b6cfc0?, 0x0?, 0x10535e532?, {0x0?, 0x0?}, 0x18?, 0xc0fb18?, {0x0?, 0x14000c0fb88?, 0x1044c4f20?})
	/boundary/internal/cmd/commands/workerscmd/worker-led_funcs.go:67 +0x84
```


louisruch (commented):
Mind updating the same error here https://github.com/hashicorp/boundary/blob/main/internal/cmd/commands/authmethodscmd/oidc_funcs.go#L294

louisruch (approved):
LGTM

PR #2389, opened 2022-08-24
Fix incorrect use of loop variable in parallel tests

renatolabs:
This fixes occurences of a loop variable being captured in parallel
tests. With the previous code, only the last test case is actually
exercised. To work around this problem, we create a local copy of the
range variable before the parallel test, as suggested in the
documentation for the `testing` package:

https://pkg.go.dev/testing#hdr-Subtests_and_Sub_benchmarks

Issues were found automatically using the `loopvarcapture` linter.


johanbrandhorst (approved):
LGTM! Thanks for catching these

PR #2390, opened 2022-08-24
Fix validation function panic

johanbrandhorst:
The logic would panic if no attributes were specified. Ensure that attributes are non-nil before checking any sub fields.

### DIFF
--- a/internal/daemon/controller/handlers/managed_groups/managed_group_service.go
+++ b/internal/daemon/controller/handlers/managed_groups/managed_group_service.go
@@ -601,12 +601,13 @@ func validateCreateRequest(req *pbs.CreateManagedGroupRequest) error {
 			attrs := req.GetItem().GetOidcManagedGroupAttributes()
 			if attrs == nil {
 				badFields[globals.AttributesField] = "Attribute fields is required."
-			}
-			if attrs.Filter == "" {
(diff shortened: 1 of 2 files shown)


internal/daemon/controller/handlers/managed_groups/managed_group_service.go
```diff
@@ -601,12 +601,13 @@ func validateCreateRequest(req *pbs.CreateManagedGroupRequest) error {
 			attrs := req.GetItem().GetOidcManagedGroupAttributes()
 			if attrs == nil {
 				badFields[globals.AttributesField] = "Attribute fields is required."
-			}
-			if attrs.Filter == "" {
-				badFields[attrFilterField] = "This field is required."
 			} else {
-				if _, err := bexpr.CreateEvaluator(attrs.Filter); err != nil {
-					badFields[attrFilterField] = fmt.Sprintf("Error evaluating submitted filter expression: %v.", err)
+				if attrs.Filter == "" {
```
louisruch, line 605: Could also make use of the `GetFilter()` method that should protect against `nil`

louisruch (approved):
It seems we have a mixture of using the `Getters` in a number of places and others do not, either way LGTM

PR #2395, opened 2022-08-25
Add Rotate Keys CLI Command

DanielleMiu:
Add key rotation CLI command with optional rewrap

### DIFF
--- a/internal/cmd/commands/scopescmd/rotate_keys.go
+++ b/internal/cmd/commands/scopescmd/rotate_keys.go
@@ -29,7 +29,7 @@ func (c *RotateKeysCommand) Help() string {
 	return base.WrapForHelpText([]string{
 		"Usage: boundary scopes rotate-keys [args]",
 		"",
-		"  Generate new keys for a scope and all nested scopes. Example:",
+		"  Generate a new root key and new DEKs for a scope. Example:",
 		"",
 		`    $ boundary scopes rotate-keys -scope-id global`,
 		"",
@@ -38,7 +38,6 @@ func (c *RotateKeysCommand) Help() string {
 }
 
 func (c *RotateKeysCommand) Flags() *base.FlagSets {
-	// what are these and what do they do??
 	set := c.FlagSet(base.FlagSetHTTP | base.FlagSetClient | base.FlagSetOutputFormat)
 	f := set.NewFlagSet("Command Options")
 
@@ -51,7 +50,7 @@ func (c *RotateKeysCommand) Flags() *base.FlagSets {
 	f.BoolVar((&base.BoolVar{
 		Name:   "rewrap",
 		Target: &c.FlagRewrap,
-		Usage:  "Whether or not child keys should be re-encrypted by the newly generated keys",
+		Usage:  "Whether or not to re-encrypt DEKs with the new KEK",
 	}))
 
 	return set
--- a/internal/cmd/commands/scopescmd/rotate_keys.go
+++ b/internal/cmd/commands/scopescmd/rotate_keys.go
@@ -31,7 +31,7 @@ func (c *RotateKeysCommand) Help() string {
 		"",
 		"  Generate new keys for a scope and all nested scopes. Example:",
 		"",
-		`    $ boundary scopes rotate-keys -scope-id global`, // how to show rewrap?
+		`    $ boundary scopes rotate-keys -scope-id global`,
 		"",
 		"",
 	}) + c.Flags().Help()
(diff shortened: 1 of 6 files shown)


internal/cmd/commands/scopescmd/rotate_keys.go
```diff
@@ -0,0 +1,118 @@
...
+
+func (c *RotateKeysCommand) Synopsis() string {
+	return wordwrap.WrapString("Rotate the keys within a scope", base.TermWidth)
+}
+
+func (c *RotateKeysCommand) Help() string {
+	return base.WrapForHelpText([]string{
+		"Usage: boundary scopes rotate-keys [args]",
+		"",
+		"  Generate new keys for a scope and all nested scopes. Example:",
```
johanbrandhorst, line 32: Is it true that it generates new keys for nested scopes too? I'm not sure?

internal/cmd/commands/scopescmd/rotate_keys.go
```diff
@@ -0,0 +1,118 @@
...
+		"",
+		`    $ boundary scopes rotate-keys -scope-id global`,
+		"",
+		"",
+	}) + c.Flags().Help()
+}
+
+func (c *RotateKeysCommand) Flags() *base.FlagSets {
+	// what are these and what do they do??
```
johanbrandhorst, line 41: Do you want to remove this comment now?

internal/cmd/commands/scopescmd/rotate_keys.go
```diff
@@ -0,0 +1,118 @@
...
+	f.StringVar(&base.StringVar{
+		Name:   "scope-id",
+		Target: &c.FlagScopeId,
+		Usage:  "The id of the scope in which to rotate keys",
+	})
+
+	f.BoolVar((&base.BoolVar{
+		Name:   "rewrap",
+		Target: &c.FlagRewrap,
+		Usage:  "Whether or not child keys should be re-encrypted by the newly generated keys",
```
johanbrandhorst, line 54: I feel like this could be worded a bit differently. How about
```suggestion
		Usage:  "Whether or not to re-encrypt DEKs with the new KEK",
```
Looks like this is the terminology we use externally: https://www.boundaryproject.io/docs/concepts/security/data-encryption#the-root-kms-key-and-per-scope-kek-deks-oss-only

internal/cmd/commands/scopescmd/rotate_keys.go
```diff
@@ -0,0 +1,118 @@
...
+			c.PrintApiError(apiErr, "Error from controller when rotating scope keys")
+			return base.CommandApiError
+		}
+		c.PrintCliError(fmt.Errorf("Error trying to rotate scope keys: %w", err))
+		return base.CommandCliError
+	}
+
+	switch base.Format(c.UI) {
+	case "json":
+		if ok := c.PrintJsonItem(result.GetResponse()); !ok {
```
johanbrandhorst, line 109: Does this actually print anything? Maybe we should just not output anything here.

internal/cmd/commands/scopescmd/rotate_keys.go
DanielleMiu: ah yeah it looks like I missed one of my confusion comments

internal/cmd/commands/scopescmd/rotate_keys.go
DanielleMiu: I thought it was but I don't think it is anymore. the new comment is mostly taken from the description given by [go-kms-wrapping](https://github.com/hashicorp/go-kms-wrapping/blob/main/extras/kms/kms.go#L436-L437), so should be accurate

internal/cmd/commands/scopescmd/rotate_keys.go
```diff
@@ -0,0 +1,117 @@
...
+
+func (c *RotateKeysCommand) Synopsis() string {
+	return wordwrap.WrapString("Rotate the keys within a scope", base.TermWidth)
+}
+
+func (c *RotateKeysCommand) Help() string {
+	return base.WrapForHelpText([]string{
+		"Usage: boundary scopes rotate-keys [args]",
+		"",
+		"  Generate a new root key and new DEKs for a scope. Example:",
```
johanbrandhorst, line 32: I wonder if we should just double down on the `KEK` terminology as well
```suggestion
		"  Generate a new KEK and new DEKs for a scope. Example:",
```
What do you think?

johanbrandhorst (approved):

PR #2397, opened 2022-08-25
docs(workers): add documentation for pki worker cred encryption

irenarindos:
Added documentation for PKI workers around the use of encrypted cred storage

### DIFF
--- a/website/content/docs/configuration/worker/pki-worker.mdx
+++ b/website/content/docs/configuration/worker/pki-worker.mdx
@@ -22,6 +22,18 @@ worker {
 }
 ```
 
+## KMS Configuration
+PKI Workers credentials can be encrypted by including an optional KMS stanza with the purpose `worker-auth-storage`.
+
+```hcl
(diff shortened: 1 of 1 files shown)


website/content/docs/configuration/worker/pki-worker.mdx
```diff
@@ -22,6 +22,18 @@ worker {
 }
 ```
 
+## KMS Configuration
+PKI Workers credentials can be encrypted by including an optional KMS stanza with the purpose `worker-auth-storage`.
+
+```hcl
```
jefferai, line 28: Here and below can you add a comment making it clear that this is an example only and production scenarios should use a different KMS option?

jefferai (approved):
Looks good, please just add that one comment

website/content/docs/configuration/worker/pki-worker.mdx
irenarindos: Done!

PR #2398, opened 2022-08-26
Merge BYOW GA branch to main

irenarindos:


internal/db/schema/migrations/oss/postgres/48/01_server_worker_release_version.up.sql
```diff
@@ -0,0 +1,53 @@
+begin;
+
+  alter table server_worker
+    add column release_version text;
```
mgaffney, line 4: * Can `release_version` be `null`?
* Is an empty string allowed?
* Are there any restrictions (min/max length) `release_version`?

irenarindos: Release version will be null in the case where a worker has not reported status yet (or if a user is using an older worker that does not include release version in its status).

If version is provided, it should start with `Boundary`; I can add that as a min length restriction if it's not null.

Issue #2349, opened 2022-08-11
database migration failed


hijak:
Hi i updated (0.10.0) via yum but am unable to start the service without a db migration which fails:
```
[root@bastion01 ~]# boundary database migrate -config /etc/boundary.hcl 
{"id":"2wDjWNR1F5","source":"https://hashicorp.com/boundary/bastion01.prd.TLD/boundary-database-migrate","specversion":"1.0","type":"system","data":{"version":"v0.1","op":"github.com/hashicorp/boundary/internal/observability/event.(*HclogLoggerAdapter).writeEvent","data":{"@original-log-level":"none","@original-log-name":"aws","msg":"configuring client automatic mTLS"}},"datacontentype":"application/cloudevents","time":"2022-08-11T12:12:27.01721969+02:00"}
{"id":"zmK1HzTdgk","source":"https://hashicorp.com/boundary/bastion01.prd.TLD/boundary-database-migrate","specversion":"1.0","type":"system","data":{"version":"v0.1","op":"github.com/hashicorp/boundary/internal/observability/event.(*HclogLoggerAdapter).writeEvent","data":{"@original-log-level":"none","@original-log-name":"aws","args":["/tmp/198135332/boundary-plugin-host-aws.gz"],"msg":"starting plugin","path":"/tmp/198135332/boundary-plugin-host-aws.gz"}},"datacontentype":"application/cloudevents","time":"2022-08-11T12:12:27.040366907+02:00"}
{"id":"CDl5AezqRk","source":"https://hashicorp.com/boundary/bastion01.prd.TLD/boundary-database-migrate","specversion":"1.0","type":"system","data":{"version":"v0.1","op":"github.com/hashicorp/boundary/internal/observability/event.(*HclogLoggerAdapter).writeEvent","data":{"@original-log-level":"none","@original-log-name":"aws","msg":"plugin started","path":"/tmp/198135332/boundary-plugin-host-aws.gz","pid":2843753}},"datacontentype":"application/cloudevents","time":"2022-08-11T12:12:27.041835792+02:00"}
[...]

Issue #1001, opened 2021-03-15
Feature: Postgres Connection Pooling Control


dylanrhysscott:
**Is your feature request related to a problem? Please describe.**

I'm using a managed postgres database on Digitalocean which has pro rata connection limits via connection pooling - When running Boundary I don't see an option for controlling the number of connections made. A Boundary replica sometimes fails to start due to connection pool exhaustion. There are 3 controllers running each appearing to make 3 connections taking 9 connections total

**Describe the solution you'd like**

It would be really good to be able to specify the number of max connections a given controller can make to Postgres so I can control my usage and prevent Boundary from exhausting available connections. This could be added to Postgres configuration like below or via an env var:

```
controller {
...
  database {
    url = "postgresql://<username>:<password>@10.0.0.1:5432/<database_name>"
    max_connections = 1
  }
}

```

**Describe alternatives you've considered**

* I have considered scaling up my DO postgres database to increase possible connections which has a cost implication (This is only a pet project) 
* Scaling down my replicas
* Running a single controller pod and multiple workers

**Explain any additional use-cases**

I'm currently running Boundary on Kubernetes deployment with 3 replicas. Each replica runs as a controller and worker

**Additional context**
* My connection pool is a session pool - I had issues with transactional pools
* I am running a $15 DO Postgres instance
* Link to DO Postgres docs [https://www.digitalocean.com/docs/databases/postgresql/how-to/manage-connection-pools/](https://www.digitalocean.com/docs/databases/postgresql/how-to/manage-connection-pools/)

dylanrhysscott:
Just to add more info:

Looking here I see Boundary uses Gorm  https://github.com/hashicorp/boundary/blob/main/internal/db/db.go#L31 - I think I'm running into issues with hitting the MaxIdleConnections. If the DB object returned on open follows the generic interface it should be possible to set some additional options around numbers of connections. [https://gorm.io/docs/generic_interface.html#Connection-Pool](https://gorm.io/docs/generic_interface.html#Connection-Pool)

malnick:
Thanks for posting @dylanrhysscott - I'm going to escalate this to @jimlambrt.

jimlambrt:
Ty for the suggestion and we can definitely add a configuration setting to override the max idle connections.  It's in our backlog now and we'll work on prioritizing it.

dylanrhysscott:
Thanks @jimlambrt  @malnick Looking at the change I would be happy to have a go at this?

[...]

PR #2407, opened 2022-08-29
fix(db): update host_plugin_set to set project_id column value

ddebko:
### Summary:

We had a user report a [issue](https://github.com/hashicorp/boundary/issues/2349#issuecomment-1229953874) where an error occurred when applying the 0.10.2 patch. Details of the error being thrown: `project_id must not be empty: not null constraint violated: integrity violation: error #1001`. 

The cause of this error was due to a bug in this sql statement. The where clause is not comparing the correct columns to set the project_id value, which caused all the rows to have a null value for the project_id column:
```update host_plugin_set
     set (project_id) =
         (select project_id
            from host_set
           where host_set.public_id = host_plugin_set.catalog_id
         )
  ;
```
  
The fix:
```update host_plugin_set
     set (project_id) =
         (select project_id
            from host_set
           where host_set.public_id = host_plugin_set.public_id
         )
  ;
```

### Validations:

I created a aws host plugin set in both 0.9.2 & 0.10.1 and applied the patch with no issues.

### DIFF
--- a/internal/db/schema/migrations/oss/postgres/46/02_hosts.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/46/02_hosts.up.sql
@@ -88,7 +88,7 @@ begin;
      set (project_id) =
          (select project_id
             from host_set
-           where host_set.public_id = host_plugin_set.catalog_id
+           where host_set.public_id = host_plugin_set.public_id
          )
   ;
 


.release/ci.hcl
```diff
@@ -13,7 +13,7 @@ project "boundary" {
 
     release_branches = [
       "main",
-      "release/0.10.2",
+      "release/0.10.3",
```
louisruch, line 16:
```suggestion
      "release/0.10.x",
```

CHANGELOG.md
```diff
@@ -2,8 +2,23 @@
 
 Canonical reference for changes, improvements, and bugfixes for Boundary.
 
+## 0.10.3 (2022/08/30)
+
+### Bug Fixes
+
+* db: Fix an issue with migrations failing due to not updating the project_id value for the host plugin set.
```
louisruch, line 9:
```suggestion
* db: Fix an issue with migrations failing due to not updating the project_id value for the host plugin set
```

louisruch (changes requested):

mgaffney (approved):

PR #2412, opened 2022-08-30
Add key_id, scope_id to oplog_entry table, add kms_key_destruction table

johanbrandhorst:
## [feat(oplog): Store key_id and scope_id with oplog entries](https://github.com/hashicorp/boundary/commit/3eb709e66801262e5f5b2a4305477cb4fca78a8c)

This migration truncates all existing oplog entries, as
migrating existing data would have been complex and
likely unnecessary.

## [feat(kms): Add kms_key_destruction table](https://github.com/hashicorp/boundary/pull/2412/commits/7e7eef3affa45cb706d5dfafbdf3be50aa0e4181)

This is used for creating, listing and updating key destructions.


internal/db/read_writer_ext_test.go
```diff
@@ -16,8 +16,8 @@ import (
 
 func TestDb_Create_OnConflict(t *testing.T) {
 	ctx := context.Background()
-	wrapper := db.TestWrapper(t)
 	conn, _ := db.TestSetup(t, "postgres")
+	wrapper := db.TestDBWrapper(t, conn, "oplog")
```
johanbrandhorst, line 20: A lot of tests that previously were happy enough to use an in-memory wrapper now require a full DB wrapper, since oplog entries require a key_id that exists.

internal/iam/repository_test.go
```diff
@@ -160,6 +160,7 @@ func Test_Repository_delete(t *testing.T) {
 		err = db.TestVerifyOplog(t, rw, s.PublicId, db.WithOperation(oplog.OpType_OP_TYPE_DELETE), db.WithCreateNotBefore(5*time.Second))
 		require.NoError(err)
 	})
+	oplog.TestOplogDeleteAllEntries(t, db.New(conn).UnderlyingDB())
```
johanbrandhorst, line 163: This is now needed in some integration tests whenever we want to delete all KMS keys, to avoid foreign key reference errors (the oplog has `restrict` on delete).

internal/db/schema/migrations/oss/postgres/48/03_create_kms_key_destructions.up.sql
```diff
@@ -0,0 +1,40 @@
...
+    ('pending'),
+    ('running'),
+    ('completed'),
+    ('failed');
+
+  create trigger 
+    immutable_columns
+  before
+  update on kms_key_destruction_status_enm
+    for each row execute procedure immutable_columns('name');
```
mgaffney, line 22:
```suggestion
   create trigger immutable_columns before update on kms_key_destruction_status_enm
    for each row execute procedure immutable_columns('name');
```
Please reformat to follow our [sql style guide](https://github.com/hashicorp/boundary/blob/main/sql-style-guide.md#triggers).

internal/db/schema/migrations/oss/postgres/48/03_create_kms_key_destructions.up.sql
```diff
@@ -0,0 +1,40 @@
...
+
+  create table kms_key_destruction (
+    private_id wt_private_id primary key,
+    key_id wt_private_id not null, -- Note: not a foreign key, outlives original reference
```
mgaffney, line 26: Can the same `key_id` be in this table multiple times?

internal/db/schema/migrations/oss/postgres/48/03_create_kms_key_destructions.up.sql
```diff
@@ -0,0 +1,40 @@
...
+    activation_time wt_timestamp not null,
```
mgaffney, line 27: What does `activation_time` mean? Is it the time the key was activated or the time the destruction of the key was activated? Is the value related to `state`?

internal/db/schema/migrations/oss/postgres/48/03_create_kms_key_destructions.up.sql
```diff
@@ -0,0 +1,40 @@
...
+    deactivation_time wt_timestamp not null,
```
mgaffney, line 28: What does `deactivation_time` mean? How does it relate to `activation_time`? Can `activation_time` be greater than `deactivation_time`? How does it relate to the `_time` columns? Does it have any relationship with the `state` column?

internal/db/schema/migrations/oss/postgres/48/03_create_kms_key_destructions.up.sql
```diff
@@ -0,0 +1,40 @@
...
+  );
+  comment on table kms_key_destruction_status_enm is
+    'Table holding valid statuses a key destruction can take';
+
+  insert into kms_key_destruction_status_enm (name)
+  values
+    ('pending'),
+    ('running'),
+    ('completed'),
+    ('failed');
```
mgaffney, line 16: What happens in the `failed` state?

internal/db/schema/migrations/oss/postgres/48/03_create_kms_key_destructions.up.sql
```diff
@@ -0,0 +1,40 @@
...
+    for each row execute procedure immutable_columns('name');
+
+  create table kms_key_destruction (
+    private_id wt_private_id primary key,
+    key_id wt_private_id not null, -- Note: not a foreign key, outlives original reference
+    activation_time wt_timestamp not null,
+    deactivation_time wt_timestamp not null,
+    destruction_start_time wt_timestamp not null,
+    destruction_end_time wt_timestamp null default null,
+    status text not null
```
mgaffney, line 31: * What does it mean if the `status` is `completed` and the `destruction_end_time` is null? 
* What does it mean if the `status` is `pending`, `running`, or `failed` and the `destruction_end_time` is not null?

mgaffney (changes requested):

internal/db/schema/migrations/oss/postgres/48/03_create_kms_key_destructions.up.sql
johanbrandhorst: It is the time the key was originally activated, i.e. when the key was created. This is meant to be used by a person who wants to track exactly from and to when a key was in the various states, `active`, `deactivated` and `destroyed`.

internal/db/schema/migrations/oss/postgres/48/03_create_kms_key_destructions.up.sql
johanbrandhorst: `deactivation_time` is the time the key was deactivated, i.e. when it was rotated and no longer used to encrypt new data, only decrypt historical data. It will always be greater than `activation_time`, as keys can only progress from `deactivated` after being `active`. It does not have any relationship to the `state` column.

internal/db/schema/migrations/oss/postgres/48/03_create_kms_key_destructions.up.sql
johanbrandhorst: If the state is `failed`, the process has finished unsuccessfully. It will not be automatically retried. The key material will not have been destroyed, through some data in the database may have been successfully rewrapped already.

internal/db/schema/migrations/oss/postgres/48/03_create_kms_key_destructions.up.sql
johanbrandhorst: > What does it mean if the status is `completed` and the destruction_end_time is null?

This is an invalid state. Do you know any way we could enforce this in the schema?

> What does it mean if the status is `pending`, `running`, or `failed` and the destruction_end_time is not null?

For statuses `pending`, and `running` this is also an invalid state. For `failed`, this would be expected to be set (same as for `completed`).

johanbrandhorst:
Going back to the drawing board on this

PR #2417, opened 2022-09-01
feat(worker): Support reloading of initial upstreams on SIGHUP

tmessi:
On a SIGHUP, if the worker's `initial_upstreams` config option has changed,
it will use the new values as the address to send its status. Any
previously discovered upstream addresses from previous status requests are
dropped. Once a successful status request is made to the new addresses, the
worker will resume its normal behavior of using the addresses from the
response to the status request.

This allows a worker to be deployed in a more stable way in a dynamic
environment where controller IPs could change. In particular this helps
with a case where all controllers are replaced with new controllers running
on new IPs.

To illustrate with an example, suppose there is an instance group of
controllers instances. And there are worker instances that have their
config rendered from consul-template that sets `initial_upstreams` to the
IPs of the controller instances. If all controller instances are restarted
or replaced at the same time and assigned new IPs, a new config will be
rendered for the worker. Previously the only way for the worker to use the
new value would be for the worker process to be restarted. Otherwise the
worker would only know about the previous IPs and would fail to connect to
the new controllers. However, a restart would mean any active sessions and
session connections on the worker instance would be terminated. With this
update, consul-template can send a SIGHUP to the worker, the worker will
connect to the new controllers and resume normal status reporting, allowing
the sessions and session connections to remain.

The same principles would apply for other examples of dynamic environments
such as nomad and kubernetes.

See:
  https://github.com/hashicorp/consul-template
  https://www.nomadproject.io/docs/job-specification/template

### DIFF
--- a/internal/cmd/commands/server/worker_initial_upstreams_reload_test.go
+++ b/internal/cmd/commands/server/worker_initial_upstreams_reload_test.go
@@ -1,0 +1,141 @@
+//go:build !hsm
+
+package server
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/boundary/internal/server"
+	"github.com/hashicorp/boundary/testing/controller"
+	"github.com/mitchellh/cli"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/atomic"
+)
+
+const initialUpstreamConfig = `
+disable_mlock = true
+
+kms "aead" {
+	purpose = "worker-auth"
+	aead_type = "aes-gcm"
+	key = "%s"
+	key_id = "global_worker-auth"
+}
+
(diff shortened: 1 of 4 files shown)


internal/cmd/commands/server/worker_initial_upstreams_reload_test.go
```diff
@@ -0,0 +1,128 @@
+//go:build !hsm
+// +build !hsm
```
jefferai, line 2: I don't think we need this second line given our minimum Go version?

jefferai (approved):

internal/cmd/commands/server/worker_initial_upstreams_reload_test.go
tmessi: Yep, fixed. At some point we should probably remove the remaining `+build` lines like the one I copied this from.

PR #2421, opened 2022-09-02
Add Staticcheck to Build Pipeline

DanielleMiu:
runs [staticcheck](https://staticcheck.io/) via [golangci-lint](https://golangci-lint.run/) on pull requests and available via the command line using `make lint` or `make lint-diff`

silences errors `ST1005` (error message formatting) and `ST1003` (variable name formatting) due to current boundary conventions differing from the formatting expected from these style checks


.github/workflows/staticcheck.yml
```diff
@@ -0,0 +1,19 @@
+name: "staticcheck"
+on: ["push", "pull_request"]
```
johanbrandhorst, line 2: Looks like this is causing it to run twice for our normal workflow, maybe we just need `push` like our other CI workflows (https://github.com/hashicorp/boundary/blob/main/.github/workflows/build.yml#L3-L6)?

.github/workflows/staticcheck.yml
```diff
@@ -0,0 +1,19 @@
...
+    name: "Run Staticcheck"
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 1
+    - uses: WillAbides/setup-go-faster@v1.7.0
+      with:
+        go-version: 1.19.x
+    # - run: "go vet ./..."
```
johanbrandhorst, line 15: If this is being made part of the same workflow eventually, we should call this job `"Run Linters"` and rename this workflow and file. Maybe it should just be a separate workflow?

johanbrandhorst: Also, we can reenable this and just set `continue-on-error: true`: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

.github/workflows/staticcheck.yml
```diff
@@ -0,0 +1,19 @@
+name: "staticcheck"
+on: ["push", "pull_request"]
+
+jobs:
+  ci:
+    name: "Run Staticcheck"
+    runs-on: ubuntu-latest
```
johanbrandhorst, line 7: I think we want `continue-on-error: true` on this level too: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error

johanbrandhorst (changes requested):

johanbrandhorst:
Only 230 errors left 😂

tmessi:
Can we use [golangci-lint](https://github.com/golangci/golangci-lint) instead of calling individual linters directly? This would make it easier to introduce additional linters as we see fix. It also provides a nice option to only detect new errors relative to a git revision. So we can check just new code in a PR without needing to fix all existing code. This makes it easier to add a CI check and have it be green.

I have a (rather old) branch that was a prove of concept on this [here](https://github.com/hashicorp/boundary/tree/tmessi-linters).

.github/workflows/linting.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          go-version: "${{ steps.get-go-version.outputs.go-version }}"
+      - name: Install Dependencies
+        # if we really need to we can update this to run `make tools`
+        # later but its just not necessary to only run linters
+        run: |
+          make golangci-lint
+      - name: Running Linters
+        run: |
+          LINT_DIFF_BRANCH=origin/${GITHUB_BASE_REF} make lint-diff
```
tmessi, line 30: In order for `GITHUB_BASE_REF` to be set, I think you need to change this to run on pull-requests instead of pushes, ie:

```
on: ["pull_request"]
```

internal/session/job_session_cleanup_test.go
```diff
@@ -39,7 +39,7 @@ func TestSessionConnectionCleanupJob(t *testing.T) {
 	kms := kms.TestKms(t, conn, wrapper)
 	serversRepo, err := server.NewRepository(rw, rw, kms)
 	require.NoError(err)
-	sessionRepo, err := NewRepository(ctx, rw, rw, kms)
```
tmessi, line 42: Probably should resolve this by adding a check against this error, ie

```go
require.NoError(err)
```

internal/session/repository_session_test.go
```diff
@@ -1563,7 +1563,7 @@ func TestRepository_deleteTargetFKey(t *testing.T) {
 	iamRepo := iam.TestRepo(t, conn, wrapper)
 	kms := kms.TestKms(t, conn, wrapper)
 	ctx := context.Background()
-	repo, err := NewRepository(ctx, rw, rw, kms)
```
tmessi, line 1566: Same here, add a `require.NoError(err)` instead of ignoring this error.

tmessi (changes requested):
looks pretty good. Just one request around how you addressed some of the lint errors.

Makefile
```diff
@@ -19,11 +20,21 @@ cli:
...
 	go install github.com/mfridman/tparse@v0.10.3
 
+# golangci-lint recommends installing the binary directly, instead of using go get
+# See the note: https://golangci-lint.run/usage/install/#install-from-source
+.PHONY: golangci-lint
+golangci-lint:
+	$(eval GOLINT_INSTALLED := $(shell which golangci-lint))
+
+	if [ "$(GOLINT_INSTALLED)" = "" ]; then \
+		curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(GO_PATH)/bin v1.50.1; \
```
johanbrandhorst, line 35: Is this really the best way of doing this? I'm a bit worried about introducing something like this in our CI pipeline. Instead, could we copy the contents of this (or at least the important parts) to a local script that we control, instead of one that could be modified in a supply chain attack?

tmessi (approved):

johanbrandhorst (approved):
LGTM, I'm very excited about this! Thanks for bringing it over the line!

PR #2423, opened 2022-09-06
feat: static json credentials

ddebko:
### Summary:

This PR provides the user the ability to utilize json credentials. The json credentials are stored within Boundary's static credential store. The user should only be able to associate a json credential to a target with a brokered credential purpose. Trying to associate a json credential with an injected application credential purpose should result in an error.  JSON credentials can be identified by the prefix `credjson`. When updating a json credential's object, it will not append the new object to existing values, but it will completely overwrite the old value with the new value.

### DB Changes:

- a new table was created called credential_static_json_credential, which stores the encrypted value of the json object into the column `object_encrypted` as bytes.

### Example CLI Commands:

Create json credential using flags: string-kv, object, bool-kv, num-kv
```
boundary credentials create json \
    -credential-store-id csst_uR6vghv9w1 \
    -name "json credential 1" \
    -description "this is a json object" \
    -string-kv "username=admin" \
    -kv "password=1234567890" \
    -bool-kv "secret=true" \
    -num-kv "pin=1234"
```
Output:
```
Credential information:
  Created Time:          Tue, 06 Sep 2022 10:23:27 PDT
  Credential Store ID:   csst_uR6vghv9w1
  Description:           this is a key value secret
  ID:                    credjson_Q75DAHgqFq
  Name:                  json credential 1
  Type:                  json
  Updated Time:          Tue, 06 Sep 2022 10:23:27 PDT
  Version:               1

  Scope:
    ID:                  p_1234567890
    Name:                Generated project scope
    Parent Scope ID:     o_1234567890
    Type:                project

  Authorized Actions:
    no-op
    read
    update
    delete

  Attributes:
    object_hmac:        yvo-Z46zGEN2VbCz54l2wekUduB4H3S1qfIutbD-N0c
```

Create json credential using json file input:
```
boundary credentials create json \
    -credential-store-id csst_uR6vghv9w1 \
    -name "json credential 2" \
    -description "this is a json object using a file input" \
    -secrets file:///real/path/to/secrets
```
Output:
```
Credential information:
  Created Time:          Tue, 06 Sep 2022 10:36:27 PDT
  Credential Store ID:   csst_uR6vghv9w1
  Description:           this is a json object using a file input
  ID:                    credjson_WxUp1eQJpg
  Name:                  json credential 2
  Type:                  json
  Updated Time:          Tue, 06 Sep 2022 10:36:27 PDT
  Version:               1

  Scope:
    ID:                  p_1234567890
    Name:                Generated project scope
    Parent Scope ID:     o_1234567890
    Type:                project

  Authorized Actions:
    no-op
    read
    update
    delete

  Attributes:
    object_hmac:        awql3hdWvcPS2OjVGZHh7XnuOGqG_LXpHB38qqzNx9E
```

Updating the name, description, and object values:
```
boundary credentials update json \
    -id credjson_Q75DAHgqFq \
    -name "updated json credential 1" \
    -description "this is a update test for json credentials" \
    -string-kv "username=default" \
    -string-kv "password=qwertyuiop"
```
Output:
```
Credential information:
  Created Time:          Tue, 06 Sep 2022 10:23:27 PDT
  Credential Store ID:   csst_uR6vghv9w1
  Description:           this is a update test for json credentials
  ID:                    credjson_Q75DAHgqFq
  Name:                  updated json credential 1
  Type:                  json
  Updated Time:          Tue, 06 Sep 2022 10:46:35 PDT
  Version:               2

  Scope:
    ID:                  p_1234567890
    Name:                Generated project scope
    Parent Scope ID:     o_1234567890
    Type:                project

  Authorized Actions:
    no-op
    read
    update
    delete

  Attributes:
    object_hmac:        KDwBzXhlQiYMNQ3s-g6gi21XrP7l4KKIPqQlqfWDwfY
```

Reading a json credential:
```
boundary credentials read -id credjson_Q75DAHgqFq 
```
Output:
```
Credential information:
  Created Time:          Tue, 06 Sep 2022 10:23:27 PDT
  Credential Store ID:   csst_uR6vghv9w1
  Description:           this is a update test for json credentials
  ID:                    credjson_Q75DAHgqFq
  Name:                  updated json credential 1
  Type:                  json
  Updated Time:          Tue, 06 Sep 2022 10:46:35 PDT
  Version:               2

  Scope:
    ID:                  p_1234567890
    Name:                Generated project scope
    Parent Scope ID:     o_1234567890
    Type:                project

  Authorized Actions:
    no-op
    read
    update
    delete

  Attributes:
    object_hmac:        KDwBzXhlQiYMNQ3s-g6gi21XrP7l4KKIPqQlqfWDwfY
```

Listing json credentials:
```
boundary credentials list -credential-store-id csst_uR6vghv9w1
```
Output:
```
Credential information:
  ID:                    credjson_Q75DAHgqFq
    Version:             2
    Type:                json
    Name:                updated json credential 1
    Description:         this is a update test for json credentials
    Authorized Actions:
      no-op
      read
      update
      delete

  ID:                    credjson_WxUp1eQJpg
    Version:             1
    Type:                json
    Name:                json credential 2
    Description:         this is a key value object using a json file
    Authorized Actions:
      no-op
      read
      update
      delete
```

Deleting a json credential:
```
boundary credentials delete -id credjson_WxUp1eQJpg
```
Output:
```
The delete operation completed successfully.
```

Associating a json credential to a target with brokered credential purpose:
```
boundary targets add-credential-sources \
  -id ttcp_1234567890 \
  -brokered-credential-source credjson_Q75DAHgqFq
```
Output:
```
Target information:
  Created Time:               Tue, 06 Sep 2022 10:06:16 PDT
  Description:                Provides an initial target in Boundary
  ID:                         ttcp_1234567890
  Name:                       Generated target
  Session Connection Limit:   -1
  Session Max Seconds:        28800
  Type:                       tcp
  Updated Time:               Tue, 06 Sep 2022 10:55:51 PDT
  Version:                    3

  Scope:
    ID:                       p_1234567890
    Name:                     Generated project scope
    Parent Scope ID:          o_1234567890
    Type:                     project

  Authorized Actions:
    no-op
    read
    update
    delete
    add-host-sets
    set-host-sets
    remove-host-sets
    add-host-sources
    set-host-sources
    remove-host-sources
    add-credential-sources
    set-credential-sources
    remove-credential-sources
    authorize-session

  Host Sources:
    Host Catalog ID:          hcst_1234567890
    ID:                       hsst_1234567890

  Brokered Credential Sources:
    Credential Store ID:      csst_uR6vghv9w1
    ID:                       credjson_Q75DAHgqFq

  Attributes:
    Default Port:             22
```

Connecting to Target w/ json brokered credential:
```
boundary connect -target-id ttcp_1234567890
```
Output:
```
Proxy listening information:
  Connection Limit:    -1
  Expiration:          Tue, 06 Sep 2022 18:57:23 PDT
  Port:                62500
  Protocol:            tcp
  Session ID:          s_rNbfi96YZo

  Credentials:
    Credential Source Description: this is a update test for json credentials
    Credential Source ID:          credjson_Q75DAHgqFq
    Credential Source Name:        updated json credential 1
    Credential Store ID:           csst_uR6vghv9w1
    Credential Store Type:         static
    Credential Type:               json
    Secret:
          "password": "qwertyuiop",
          "username": "default"
```

Trying to associate a json credential with injected application purpose returns an error:
```
boundary targets add-credential-sources \
  -id ttcp_1234567890 \
  --injected-application-credential-source credjson_Q75DAHgqFq
```
Output:
```
Error from controller when performing add-credential-sources on target

Error information:
  Kind:                InvalidArgument
  Message:             Errors in provided fields.
  Status:              400
  context:             Error from controller when performing add-credential-sources on target

  Field-specific Errors:
    Name:              -injected-application-credential-source-ids
      Error:           Incorrectly formatted credential source identifier "credjson_Q75DAHgqFq".
```


internal/api/genapi/input.go
```diff
@@ -520,6 +520,17 @@ var inputStructs = []*structInfo{
...
+	{
+		inProto:     &credentials.KvAttributes{},
+		outFile:     "credentials/kv_attributes.gen.go",
+		subtypeName: "KvCredential",
+		fieldOverrides: []fieldInfo{
+			{
+				Name:        "Secrets",
+				SkipDefault: true,
+			},
+		},
```
louisruch, line 532: I would need to dig into the usage within the gen path but this does not include some of the fields in other credentials:
```
parentTypeName: "Credential",
templates: []*template.Template{
	mapstructureConversionTemplate,
},
```

internal/daemon/controller/handlers/credentials/credential_service.go
```diff
@@ -685,6 +777,18 @@ func validateCreateRequest(req *pbs.CreateCredentialRequest) error {
 				}
 			}
 
+		case credential.KvSubtype.String():
+			secrets := req.GetItem().GetKvAttributes().GetSecrets()
+			if secrets == nil && len(secrets.AsMap()) <= 0 {
+				badFields[secretsField] = "Field required for creating an kv credential."
```
johanbrandhorst, line 783:
```suggestion
				badFields[secretsField] = "Field required for creating a kv credential."
```

internal/daemon/controller/handlers/credentials/credential_service.go
```diff
@@ -685,6 +777,18 @@ func validateCreateRequest(req *pbs.CreateCredentialRequest) error {
...
+			}
+			secretBytes, err := json.Marshal(secrets)
+			if err != nil {
+				badFields[secretsField] = "Unable to parse given secret kv value"
+			} else if len(secretBytes) <= 0 {
+				badFields[secretsField] = "Field required for creating an kv credential."
```
johanbrandhorst, line 789:
```suggestion
				badFields[secretsField] = "Field required for creating a kv credential."
```

internal/daemon/controller/handlers/targets/credentials.go
```diff
@@ -162,13 +162,27 @@ func staticToWorkerCredential(ctx context.Context, cred credential.Static) (sess
 			},
 		}
 
+	case *credstatic.KvCredential:
+		secrets := structpb.Struct{}
+		err := json.Unmarshal(c.GetSecrets(), &secrets)
```
johanbrandhorst, line 167: Should this be using `jsonpb` since this is a protobuf type?

internal/db/schema/migrations/oss/postgres/48/01_static_kv_creds.up.sql
```diff
@@ -0,0 +1,81 @@
...
+    update_time wt_timestamp,
+    version wt_version,
+
+    secrets_encrypted bytea not null
+      constraint secrets_encrypted_must_not_be_empty
+        check(length(secrets_encrypted) > 0),
+    secrets_hmac bytea not null
+      constraint secrets_hmac_must_not_be_empty
+        check(length(secrets_hmac) > 0),
+    key_id text not null
```
johanbrandhorst, line 40: This type should probably be `kms_private_id` as that is the type of the referenced column: https://github.com/hashicorp/boundary/blob/d7c4c648ec89c98d05cbbd6fee3603f9b78fab67/internal/db/schema/migrations/oss/postgres/30/04_kms_keys.up.sql#L83

johanbrandhorst (changes requested):
How does parsing secrets from a JSON file work? Does it require the JSON file to be a specific format, or is the JSON file used as a value in a named key?

internal/db/schema/migrations/oss/postgres/48/01_static_kv_creds.up.sql
```diff
@@ -0,0 +1,81 @@
...
+  -- Add new constraint that only allows known types
+  -- This replaces the constraint defined in 39/01_static_ssh_private_key_creds.up.sql
+  alter table credential_type_enm
+    add constraint only_predefined_credential_types_allowed
+      check (
+        name in (
+          'unspecified',
+          'username_password',
+          'ssh_private_key',
+          'kv'
```
mgaffney, line 16: Why does `kv` need to be added to the credential types?

ddebko: I see, credential types are only used as a foreign key constraint for the credential library

internal/daemon/controller/handlers/targets/credentials.go
ddebko: I read the package doc for jsonpb & I think you are right, because it says 
```
Do not rely on the default behavior of the standard encoding/json package when called on generated message types as it does not operate correctly.
```

But it also states that jsonpb is `Deprecated` and I should use this package instead "google.golang.org/protobuf/encoding/protojson", is that cool with you?

internal/db/schema/migrations/oss/postgres/39/01_static_ssh_private_key_creds.up.sql
```diff
@@ -6,6 +6,7 @@ begin;
 
   -- Add new constraint that only allows known types
   -- This replaces the constraint defined in 36/01_credential_type_up
+  -- Replaced in 42/01_static_kv_creds.up.sql
```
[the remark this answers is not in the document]
ddebko: I will remove this comment since `kv` is not a required value in credential types

internal/daemon/controller/handlers/targets/credentials.go
johanbrandhorst: Yes, definitely use the new version 👍🏻.

ddebko:
@johanbrandhorst The JSON parsing does not require any specific file extension. An error will be returned to the user if the provided file is not valid json, example error output:
```
Error evaluating secret flags to: error parsing secret flag as JSON: invalid character 'i' looking for beginning of object key string
```
The secrets flags should behave similarly to how host set secrets work because they share the same generated code.

internal/api/genapi/input.go
ddebko: good catch, I will take a look at this & check if the api tests still pass

ddebko: I added this change and the tests are still passing. I will added more checks to validate the generated functions.

internal/credential/static/kv_credential.go
```diff
@@ -0,0 +1,138 @@
...
+	secrets credential.KvSecret,
+	opt ...Option,
+) (*KvCredential, error) {
+	const op = "static.NewKvCredential"
+	var jsonSecrets []byte
+	var err error
+
+	// since the secret is an unordered map of dynamically typed values, the hmac value will not be consistent.
+	// In order to calculate a consistent hmac value, the secret must have a ordered map
+	// which is done by marshalling the secret.
```
johanbrandhorst, line 41: Nit picking:
```suggestion
	// Since the secret is an unordered map of dynamically typed values, the hmac value will not be consistent.
	// In order to calculate a consistent hmac value, the input must be deterministic,
	// which is done by marshalling the secret.
```

internal/credential/static/repository_credential.go
```diff
@@ -485,8 +582,137 @@ func (r *Repository) UpdateSshPrivateKeyCredential(ctx context.Context,
...
+	}
+
+	// Clear secret fields, only SecretsHmac should be returned
+	returnedCredential.SecretsEncrypted = nil
+	returnedCredential.Secrets = nil
+
+	return returnedCredential, rowsUpdated, nil
+}
+
+// ListCredentials returns a slice of UsernamePasswordCredentials, SshPrivateKeyCredentials, & KVCredentials
```
johanbrandhorst, line 714:
```suggestion
// ListCredentials returns a slice of UsernamePasswordCredentials, SshPrivateKeyCredentials and KVCredentials
```

johanbrandhorst (approved):
Just minor things from me!

internal/cmd/commands/credentialscmd/kv_credentials_funcs.go
```diff
@@ -0,0 +1,33 @@
...
+			`    $ boundary credentials create kv -credential-store-id csvlt_1234567890 -secret file:///home/user/secret`,
+			"",
+			"",
+		})
+
+	case "update":
+		helpStr = base.WrapForHelpText([]string{
+			"Usage: boundary credentials update kv [options] [args]",
+			"",
+			"  Update an kv credential given its ID. Example:",
```
jefferai, line 25:
```suggestion
			"  Update a kv credential given its ID. Example:",
```

internal/cmd/gencli/templates.go
```diff
@@ -385,10 +385,11 @@ func (c *{{ camelCase .SubActionPrefix }}Command) Run(args []string) int {
 		c.FlagSecrets,
 		c.FlagScrts,
 		func() {
-			opts = append(opts, {{ .Pkg }}.DefaultSecrets())
+			{{ if (and (eq .ResourceType "host-catalog") (eq .SubActionPrefix "plugin")) }} opts = append(opts, {{ .Pkg }}.DefaultSecrets()){{ end }}
```
jefferai, line 388: I haven't gotten to the API side of things here yet (going in order on the PR) but this change strikes me as a bit eyebrow-raising -- we have avoided putting type-specific items into the templates so far...

internal/cmd/gencli/templates.go
```diff
@@ -385,10 +385,11 @@ func (c *{{ camelCase .SubActionPrefix }}Command) Run(args []string) int {
...
 		},
 		func(in map[string]interface{}) {
-			opts = append(opts, {{ .Pkg }}.WithSecrets(in))
+			{{ if (and (eq .ResourceType "host-catalog") (eq .SubActionPrefix "plugin")) }}opts = append(opts, {{ .Pkg }}.WithSecrets(in)){{ end }}
+			{{ if (and (eq .ResourceType "credential") (eq .SubActionPrefix "kv")) }}opts = append(opts, {{ .Pkg }}.WithKvCredentialSecrets(in)){{ end }}
```
jefferai, line 392: In particular, it's not clear to me (yet) why this needs a special `WithKvCredentialSecrets`

jefferai: Went back and looked at what we do elsewhere and it seems like this should just be `WithSecrets`

internal/daemon/controller/handlers/targets/credentials.go
```diff
@@ -218,6 +233,27 @@ func staticToSessionCredential(ctx context.Context, cred credential.Static) (*pb
...
+			return nil, errors.New(ctx, errors.InvalidParameter, op, "unmarshalling kv secret")
+		}
+		kvSecret, err := structpb.NewStruct(s)
+		credData, err = handlers.ProtoToStruct(
+			&pb.KVCredential{
+				Secret: kvSecret,
+			},
+		)
+		secret = map[string]interface{}{
+			"secret": s,
```
jefferai, line 251: Shouldn't `secret` be assigned to `s` here? For the other types, the credential values are top-level, but here they're being put a level down.

internal/cmd/gencli/templates.go
jefferai: If we want to keep it typed, please break up the HasGenericSecrets in some way to compose what is needed -- don't put type checks in the templates

internal/proto/controller/api/resources/targets/v1/target.proto
```diff
@@ -334,3 +334,8 @@ message SshPrivateKeyCredential {
   // The optional passphrase of the private_key
   string private_key_passphrase = 3; // @gotags: `class:"secret"`
 }
+
+// The layout of the struct for "credential" field in SessionCredential for a kv credential type.
+message KVCredential {
+  google.protobuf.Struct secret = 1; // @gotags: `class:"secret"`
```
jefferai, line 340: The pluralization here does not match the credential service. Here it's singular `secret`, there it's plural `secrets`

internal/proto/controller/storage/credential/static/store/v1/static.proto
```diff
@@ -209,3 +209,66 @@ message SshPrivateKeyCredential {
...
+  string store_id = 6;
+
+  // version allows optimistic locking of the resource.
+  // @inject_tag: `gorm:"default:null"`
+  uint32 version = 7;
+
+  // secrets is the plain-text of the secrets associated with the
+  // credential. We are not storing this plain-text secret in the database.
+  // @inject_tag: `gorm:"-" wrapping:"pt,secrets"`
+  bytes secrets = 8 [(custom_options.v1.mask_mapping) = {
```
jefferai, line 252: I've been wondering as I go through this whether `secrets` is the correct name, as opposed to `kv`. Pros and cons to both.

internal/cmd/gencli/templates.go
ddebko: I agree, I don't like having this conditional check too. 

When generating the code within the api package, the DefaultSecrets() function is not generated for kv credentials because secrets is a required field. The reason behind this conditional check in the template is because dynamic host catalog secrets & kv credentials share some code when it comes to the api and cli. 
For the dynamic host catalog, secrets is an optional field, which is why it has the DefaultSecrets function. 

I could change the kv credential api input so that a default value function is created, but then it would be an anti-pattern for how our other credentials required fields behave, such as the password field in a static username & password credential.

internal/proto/controller/storage/credential/static/store/v1/static.proto
ddebko: The reason why the field was labeled secrets is so that we can reuse the code from the dynamic host catalog secrets. For example, this [function](https://github.com/hashicorp/boundary/blob/039be6d8e40d25db68dd9bb07ead2fc00389daf5/api/credentials/option.gen.go#L168) was generated by reusing an input flag that was created for generic secrets. 
If I change the label to kv, then there will be an inconsistency between labels in storage & the api. In order to maintain naming consistency through the repo, I would be required  to adjust the templates with more conditional checks for generic secrets, which would make the templating more fragile.

internal/daemon/controller/handlers/targets/credentials.go
ddebko: Sorry, I don't understand the intentions for flattening the map. I believe that the fields provided by the user in the variable s will never be accessed within the code. Kv secrets are only intended for passing along json information, never to validate its' content for a specific structure. I also think that the variable s would never be written to by boundary in the future, so flattening the map would not help.

internal/cmd/gencli/templates.go
ddebko: The function WithKvCredentialSecrets is generated from the api package and it following the same naming patterns as the other credential types. For example: `WithSshPrivateKeyCredentialUsername`

internal/daemon/controller/handlers/targets/credentials.go
ddebko: I do think it is reasonable to remove the nested object when displaying the kv secret during a session.

internal/cmd/commands/connect/credentials.go
```diff
@@ -25,6 +25,12 @@ type sshPrivateKey struct {
 	consumed bool
 }
 
+type kv struct {
```
louisruch, line 28: I think this can be dropped now

internal/daemon/controller/handlers/targets/credentials.go
```diff
@@ -162,13 +163,27 @@ func staticToWorkerCredential(ctx context.Context, cred credential.Static) (sess
 			},
 		}
 
+	case *credstatic.KvCredential:
```
louisruch, line 166: KV can never be a worker credential, we should drop this `case` and return an error in `default`

internal/daemon/controller/handlers/targets/credentials.go
```diff
@@ -218,6 +233,27 @@ func staticToSessionCredential(ctx context.Context, cred credential.Static) (*pb
...
+		var err error
+		credType = string(credential.KvType)
+		s := map[string]interface{}{}
+		err = json.Unmarshal(c.GetSecrets(), &s)
+		if err != nil {
+			return nil, errors.New(ctx, errors.InvalidParameter, op, "unmarshalling kv secret")
+		}
+		kvSecret, err := structpb.NewStruct(s)
```
louisruch, line 244: check err

internal/daemon/controller/handlers/targets/credentials.go
```diff
@@ -218,6 +233,27 @@ func staticToSessionCredential(ctx context.Context, cred credential.Static) (*pb
...
+		credData, err = handlers.ProtoToStruct(
+			&pb.KVCredential{
+				Secret: kvSecret,
+			},
+		)
+		secret = map[string]interface{}{
+			"secret": s,
+		}
+		if err != nil {
```
louisruch, line 253: I think this error check should be before the assignment of `secret`

louisruch (commented):
Couple of quick comments will still look more

internal/proto/controller/servers/services/v1/credential.proto
```diff
@@ -35,3 +38,9 @@ message SshPrivateKey {
   // The optional passphrase of the private_key
   string private_key_passphrase = 30; // @gotags: `class:"secret"`
 }
+
+// Kv is a credential containing a set of key value pairs.
+message Kv {
```
louisruch, line 43: This can be dropped since we do not need to ever send KV credential to the worker

internal/credential/credential.go
```diff
@@ -134,3 +141,9 @@ type SshPrivateKey interface {
 	PrivateKey() PrivateKey
 	PrivateKeyPassphrase() []byte
 }
+
+// Kv is a credential containing a set of key-value pairs.
+type Kv interface {
```
louisruch, line 146: Is this interface used anywhere

internal/cmd/gencli/templates.go
ddebko: I replaced the string comparison conditions with the variable IsPluginType.

internal/daemon/controller/handlers/targets/credentials.go
ddebko: I removed the nested secrets field and assigned `s` to `secret`

internal/credential/credential.go
ddebko: I didn't use the interface anywhere, so I will drop it.

internal/cmd/gencli/templates.go
ddebko: I replaced the string comparison conditionals with IsPluginType & IsKvCredentialType

mgaffney (approved):

johanbrandhorst (approved):

louisruch (approved):
This is great well done!

internal/proto/controller/api/resources/credentials/v1/credential.proto
```diff
@@ -161,24 +161,24 @@ message SshPrivateKeyAttributes {
...
-      that: "Secrets"
+      this: "attributes.object"
+      that: "Object"
     }
   ]; // @gotags: `class:"secret"`
 
-  // Output only. The hmac value of the secrets.
-  string secrets_hmac = 20 [
-    json_name = "secrets_hmac",
+  // Output only. The hmac value of the secret.
```
johanbrandhorst, line 176: object?

internal/proto/controller/storage/credential/static/store/v1/static.proto
```diff
@@ -246,25 +246,25 @@ message KvCredential {
...
+  // object_encrypted is the ciphertext of the secret. It is stored in
   // the database.
-  // @inject_tag: `gorm:"column:secrets_encrypted;not_null" wrapping:"ct,secrets"`
-  bytes secrets_encrypted = 9;
+  // @inject_tag: `gorm:"column:object_encrypted;not_null" wrapping:"ct,object"`
+  bytes object_encrypted = 9;
 
-  // secrets_hmac is a sha256-hmac of the unencrypted secrets. It is recalculated
-  // everytime the secrets is updated.
+  // object_hmac is a sha256-hmac of the unencrypted secret. It is recalculated
```
johanbrandhorst, line 262: object?

internal/proto/controller/storage/credential/static/store/v1/static.proto
```diff
@@ -246,25 +246,25 @@ message KvCredential {
...
+  // everytime the secret is updated.
```
johanbrandhorst, line 263: object?

internal/tests/cli/boundary/credentials.bats
```diff
@@ -91,9 +91,9 @@ export NEW_KV_CREDENTIAL='test-kv'
...
 }
 
-@test "boundary/credential-stores: can not read deleted $NEW_KV_CREDENTIAL credential" {
+@test "boundary/credential-stores: can not read deleted $NEW_JSON_CREDENTIAL credential" {
   local csid=$(credential_store_id $NEW_STORE $DEFAULT_P_ID)
-  local cid=$(credential_id $NEW_KV_CREDENTIAL $csid)
+  local cid=$(credential_id $NEW_JSON_CREDENTIAL $csid)
   run read_credential $cid
   echo "$output"
   [ "$status" -eq 1 ]
```
johanbrandhorst, line 99: Will we add a test for setting a `-object` and `-kv` flag at the same time separately?

johanbrandhorst (approved):
Nice!

mgaffney (approved):

internal/credential/credential.go
```diff
@@ -119,6 +121,11 @@ type Password string
 // PrivateKey represents a secret private key.
 type PrivateKey []byte
 
+// JsonObject represents a JSON object that is serialized.
+type JsonObject struct {
```
jefferai, line 125: It seems like some of the code later could be slightly simplified by doing `type JsonObject structpb.Struct` instead of embedding but I'm not sure it makes any true difference.

jefferai (approved):

internal/credential/credential.go
ddebko: I decided to use struct composition instead so I can use some of the methods from structpb.Struct. For example, I use the AsMap() function.

PR #2425, opened 2022-09-06
Add "Connect to Target" tests to new end-to-end test suite

moduli:
## Description
This PR adds two tests to a new end-to-end test suite that utilizes [enos](https://github.com/hashicorp/Enos-Doc) to spin up the desired infrastructure. In this enos scenario, it creates a boundary cluster + target and invokes the following tests...
- The first test uses the boundary cli to create a number of domain objects (org, project, host catalog, etc.) and connect to a target. The test executes `hostname -i` on the connected host and asserts that the output matches what we expect (the IP address of the host)
- The second test does similar operations, but uses the boundary go api. It does not connect to a target due to the underlying complexity of the `boundary connect` cli command.

Tests were written in Go due to the following reasons
- Go is the primary language used in the application
- Tests can be written against the boundary go api
 
The following diagram highlights the primary flow when executed.
```mermaid
flowchart TD

A[enos/enos-scenario-e2e-target.hcl] --> B
B[enos/modules/test_e2e/main.tf] --> C
C[testing/internal/e2e/target/target_test.go]
```

To run these tests...
1. Set up your `enos.vars.hcl` file
2. Run `enos scenario run e2e_target builder:local`


moduli:
Looks like we may need to adjust the `Makefile`. The `Run Acceptance Tests` job is failing because it's running these new tests

testing/internal/e2e/target/target_test.go
```diff
@@ -0,0 +1,260 @@
...
+	})
+	require.NoError(t, output.Err, string(output.Stderr))
+	var newProjectResult scopes.ScopeCreateResult
+	err = json.Unmarshal(output.Stdout, &newProjectResult)
+	require.NoError(t, err)
+	newProject := newProjectResult.Item
+	t.Cleanup(func() {
+		e2e.RunCommand([]string{"boundary", "scopes", "delete", "-id", newProject.Id})
+	})
+	t.Logf("Created Project Id: %s", newProject.Id)
```
johanbrandhorst, line 64: I give it 1 more of these tests before you want to write `createProject` etc helpers 😁.

enos/modules/test_e2e/main.tf
```diff
@@ -0,0 +1,62 @@
...
+    BOUNDARY_ADDR                             = var.alb_boundary_api_addr,
+    BOUNDARY_AUTHMETHOD_ID                    = var.auth_method_id,
+    BOUNDARY_AUTHENTICATE_PASSWORD_LOGIN_NAME = var.auth_login_name,
+    BOUNDARY_AUTHENTICATE_PASSWORD_PASSWORD   = var.auth_password,
+    BOUNDARY_E2E_TARGET_IP                    = var.target_ips[0],
+    BOUNDARY_E2E_SSH_USER                     = "ubuntu"
+    BOUNDARY_E2E_SSH_KEY_PATH                 = local.aws_ssh_private_key_path,
+  }
+
+  inline = ["PATH=\"${var.local_boundary_dir}:$PATH\" go test -v ${var.test_package}"]
```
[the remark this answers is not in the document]
moduli: Good question... I believe it currently uses whatever is on your machine.

testing/internal/e2e/target/target_test.go
moduli: Agreed. Didn't want to abstract things too much yet. I imagine we'll see some more refactoring once we start work on the next set of tests.

enos/modules/test_e2e/main.tf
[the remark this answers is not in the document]
moduli: Not sure if this answers your question, but after talking with the enos team, their plan is to use this github action: https://github.com/actions/setup-go and supply `go.mod` to determine the version. 

Example in vault: https://github.com/hashicorp/vault-enterprise/blob/main/.github/workflows/build.yml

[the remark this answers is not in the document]
johanbrandhorst: I mostly asked the question to make you think about it 😁. Specifying the version via the action is fine. Note that setting the Go version in `go.mod` is only used to specify the _minimum_ supported Go version in the module, it doesn't manage the Go installation for you. I expect in reality this will just be updated together with all the other version strings when we update our Go version.

testing/internal/e2e/target/target_test.go
```diff
@@ -0,0 +1,280 @@
...
+		"-id", newHostSet.Id,
+		"-host", newHost.Id,
+	)
+	require.NoError(t, err, string(output.Stderr))
+
+	// Create a target
+	output, err = e2e.RunCommand(
+		"boundary", "targets", "create", "tcp",
+		"-scope-id", newProject.Id,
+		"-default-port", "22",
```
tmessi, line 140: You might want to make this configurable via an env var. I was able to run these tests against a docker setup but my ssh server was running on `2222` so I needed to change this port. (The test still failed but that was due to a the `hostname -i` command returning a docker ip instead of the address from the host, which is a separate problem that this normally will not encounter).

tmessi (changes requested):
Awesome, couple minor comments but otherwise this is working well.

testing/internal/e2e/target/target_test.go
moduli: Should we think of an alternative to checking `hostname -i` so that it would work in this docker environment?

testing/internal/e2e/boundary/boundary.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/hashicorp/boundary/api/authmethods"
+	"github.com/hashicorp/boundary/testing/internal/e2e"
+	"github.com/kelseyhightower/envconfig"
+)
+
+type config struct {
+	Address            string `envconfig:"BOUNDARY_ADDR" default:"http://127.0.0.1:9200"`
+	AuthMethodId       string `envconfig:"E2E_PASSWORD_AUTH_METHOD_ID" default:"ampw_1234567890"`
+	AdminLoginName     string `envconfig:"E2E_PASSWORD_ADMIN_LOGIN_NAME" default:"admin"`
+	AdminLoginPassword string `envconfig:"E2E_PASSWORD_ADMIN_PASSWORD"`
```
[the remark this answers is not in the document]
tmessi: Ah, no that makes sense.

testing/internal/e2e/README.md
```diff
@@ -0,0 +1,28 @@
...
+enos scenario launch e2e_{scenario} builder:local  // runs and keeps infra online
+enos scenario output // displays any defined enos output
+
+enos scenario destroy // destroys infra
+```
+
+Enos scenarios set up the infrastructure, set the appropriate environment variables, and run the selected tests. Folders in this directory correspond to an enos scenario (e.g. `enos/enos-scenario-e2e-target.hcl` runs tests in `testing/e2e/target`)
+
+#### Local
+Set the appropriate environment variables, then run...
```
johanbrandhorst, line 22: Can you list the environment variables needed?

testing/internal/e2e/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+```shell
+go test github.com/hashicorp/boundary/testing/e2e/target // run target tests
```
johanbrandhorst, line 25: If running from the current folder, you can use a relative path too
```suggestion
go test ./target/ // run target tests
```

testing/internal/e2e/helpers.go
```diff
@@ -0,0 +1,53 @@
...
+	var exitCode int
+	if errors.As(err, &ee) {
+		exitCode = ee.ExitCode()
+	}
+
+	return &CommandResult{
+		Stdout:   outbuf.Bytes(),
+		Stderr:   errbuf.Bytes(),
+		ExitCode: exitCode,
+	}, err
```
johanbrandhorst, line 42: It is unusual to return an error and a result at the same time. Could we add a comment that the returned `*CommandResult` is always valid, even if there is an error, to the function comment?

testing/internal/e2e/helpers.go
```diff
@@ -0,0 +1,53 @@
...
+
+// TestInit is a helper method to set up individual tests.
+func TestInit(t *testing.T) {
+	if _, ok := os.LookupEnv(envToCheck); !ok {
+		t.Skip(fmt.Sprintf(
+			"Skipping test because environment variable '%s' is not set. This is needed for e2e tests.",
+			envToCheck,
+		))
+	}
+}
```
johanbrandhorst, line 53: I think this function might harm more than it helps. My biggest concern is that when calling this function in a test it obscures the fact that we might entirely skip the test, especially since `t.Skip` already messes with control flow magically. Could we inline this function wherever we're using it, instead of using this helper?

testing/internal/e2e/target/target_test.go
```diff
@@ -0,0 +1,280 @@
...
+	require.NoError(t, err, string(output.Stderr))
+
+	parts := strings.Fields(string(output.Stdout))
+	hostIp := parts[len(parts)-1]
+	assert.Equal(t, c.TargetIp, hostIp, "SSH session did not return expected output")
+	t.Log("Successfully connected to target")
+}
+
+// TestCreateTargetApi uses the boundary go api to create a number of supporting objects
+// to connect to a target.
```
johanbrandhorst, line 186: Maybe add a note here that this doesn't attempt to connect to the target because it's complicated to do without the CLI?

johanbrandhorst: Personally I'd be curious to see how hard it would be too 😁, but it can wait for another PR, if ever.

johanbrandhorst (changes requested):

testing/internal/e2e/boundary/boundary.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/hashicorp/boundary/api"
+	"github.com/hashicorp/boundary/api/authmethods"
+	"github.com/hashicorp/boundary/testing/internal/e2e"
+	"github.com/kelseyhightower/envconfig"
+)
+
+type config struct {
+	Address            string `envconfig:"BOUNDARY_ADDR" default:"http://127.0.0.1:9200"`
+	AuthMethodId       string `envconfig:"E2E_PASSWORD_AUTH_METHOD_ID" default:"ampw_1234567890"`
```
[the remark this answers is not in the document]
moduli: I agree it would be nice to reduce the number of environment variables that need to be set. For now, though, I would like to be explicit and use the auth method id that enos gives. We can maybe revisit this at some point in the future.

testing/internal/e2e/helpers.go
moduli: Curious, what if we renamed the method to something like `CheckIfTestShouldBeSkipped`?

johanbrandhorst: `MaybeSkipTest`?

moduli: Works for me.

testing/internal/e2e/target/target_test.go
moduli: Keeping the `hostname` check as is for now. Will revisit in the future.

tmessi (approved):

johanbrandhorst (approved):

joshbrand (approved):

PR #2430, opened 2022-09-07
Rename key_id column to key_version_id

johanbrandhorst:
This makes the relationship of encrypted data to a key version more explicit.

Just opening this PR to see how much work this would be - I'm particularly worried about all the existing views and triggers we have that reference key_id. There's also the fact that we have a `key_id` field in one of the worker gRPC message definitions - it's safe to change this if we only care about wire compatibility, but it's annoying all the same.


johanbrandhorst:
Probably not proceeding with this

PR #2431, opened 2022-09-07
Store key_id and scope_id with oplog entries, re-type columns referencing data key version

johanbrandhorst:
## [fix(db): Correct types of data key version referencers](https://github.com/hashicorp/boundary/pull/2431/commits/d62687138c87ca541c0c34148ffdfc41eb071abb)

The referencing views have to be dropped and recreated after the column type change

## [feat(oplog): Store key_id and scope_id with oplog entries](https://github.com/hashicorp/boundary/commit/ef398dd76bd37b9ea0393195274862f78bd777c9)

This migration truncates all existing oplog entries, as migrating existing data would have been complex and likely unnecessary.

The migrations have been reorganized to let the 48002 migration tests use the test helpers. Most of the logic changes are necessary because a lot of tests were using the oplog without a DB kms available.


internal/proto/controller/storage/oplog/store/v1/oplog.proto
```diff
@@ -34,6 +34,14 @@ message Entry {
...
   // @inject_tag: gorm:"-" wrapping:"pt,entry_data"
   bytes data = 8;
+
+  // the private id of the data key version used to encrypt the data.
+  // @inject_tag: gorm:"not_null"
+  string key_id = 9;
+
+  // the scope id associated with the key used to encrypt the data.
+  // @inject_tag: gorm:"not_null"
+  string scope_id = 10;
```
mgaffney, line 44: Is this needed since it is filled in by a trigger?

mgaffney (changes requested):

internal/proto/controller/storage/oplog/store/v1/oplog.proto
johanbrandhorst: It's for reading the scope ID out of the entry, not for setting it. As you notice, setting the scope ID on insert will always be overwritten by the trigger.

johanbrandhorst: I'm happy to remove this if we don't think we'll need it

mgaffney: I think it is fine to leave it

johanbrandhorst:
So the tests are failing with errors like:
```
iam.(Repository).CreateScope: for o_IFws3MZYtz: db.DoTx: iam.(Repository).CreateScope: error creating scope: db.Create: dbw.Create: error after write: db.afterFnSingleItem: db.addOplog: unable to write oplog entry: oplog.(Entry).WriteEntryWith: error writing data to storage: wt_scope_id_check constraint failed: check constraint violated: integrity violation: error #1000
```

It seems we're somehow in violation of the `wt_scope_id` check constraint. The constraint is
```
create domain wt_scope_id as text
check(
  length(trim(value)) > 10 or value = 'global'
);
```

However, in my short attempt at debugging this, it seems like the scope_id should be set to the appropriate scope. Here's a diff that gives some debugging output when running
```
$ go test ./internal/host/
```

```diff
diff --git a/internal/oplog/oplog.go b/internal/oplog/oplog.go
index 63050fa00..98838d829 100644
--- a/internal/oplog/oplog.go
+++ b/internal/oplog/oplog.go
@@ -54,6 +54,7 @@ func NewEntry(ctx context.Context, aggregateName string, metadata Metadata, ciph
 	if err != nil {
 		return nil, errors.Wrap(ctx, err, op)
 	}
+	fmt.Println(keyId)
 	entry := Entry{
 		Entry: &store.Entry{
 			AggregateName: aggregateName,
@@ -340,7 +341,18 @@ func (e *Entry) WriteEntryWith(ctx context.Context, tx *Writer, ticket *store.Ti
 			return errors.Wrap(ctx, err, op)
 		}
 	}
+	keyId, _ := e.Cipherer.KeyId(ctx)
 	rw := dbw.New(tx.DB)
+	rows, _ := rw.Query(ctx, `select scope_id
+	from kms_root_key
+	inner join kms_data_key on kms_root_key.private_id = kms_data_key.root_key_id
+	inner join kms_data_key_version on kms_data_key.private_id = kms_data_key_version.data_key_id
+	where kms_data_key_version.private_id = $1`, []interface{}{keyId})
+	rows.Next()
+	var scopeId string
+	rows.Scan(&scopeId)
+	rows.Close()
+	fmt.Println(scopeId, *e)
 	if err := rw.Create(ctx, e); err != nil {
 		return errors.Wrap(ctx, err, op, errors.WithMsg("error writing data to storage"))
 	}
```

I would appreciate any help debugging this @mgaffney. I am probably missing something obvious.

internal/proto/controller/storage/oplog/store/v1/oplog.proto
```diff
@@ -34,6 +34,14 @@ message Entry {
   // we are NOT storing this plain-text entry data in the db
   // @inject_tag: gorm:"-" wrapping:"pt,entry_data"
   bytes data = 8;
+
+  // the private id of the data key version used to encrypt the data.
+  // @inject_tag: gorm:"not_null"
+  string key_id = 9;
+
+  // the scope id associated with the key used to encrypt the data.
+  // @inject_tag: gorm:"not_null"
```
mgaffney, line 43:
```suggestion
  // @inject_tag: gorm:"default:null"
```
This should fix the problem with the tests. `gorm:"not_null"` causes gorm to set the value of `scope_id` to any empty string on insert. In Postgres, domain type constraints are checked before any triggers are run and '' is not allowed for `wt_scope_id` so the insert fails.

johanbrandhorst:
Tests are passing 🎉

internal/iam/repository_test.go
```diff
@@ -160,6 +160,7 @@ func Test_Repository_delete(t *testing.T) {
 		err = db.TestVerifyOplog(t, rw, s.PublicId, db.WithOperation(oplog.OpType_OP_TYPE_DELETE), db.WithCreateNotBefore(5*time.Second))
 		require.NoError(err)
 	})
+	oplog.TestOplogDeleteAllEntries(t, db.New(conn).UnderlyingDB())
 	kms.TestKmsDeleteAllKeys(t, conn)
```
mgaffney, line 164: There seems to be a tight coupling between these two functions. `oplog.TestOplogDeleteAllEntries` is always called right before `kms.TestKmsDeleteAllKeys`. I commented out the call to `oplog.TestOplogDeleteAllEntries` which caused an error in `kms.TestKmsDeleteAllKeys`. Given this coupling, should we just call `oplog.TestOplogDeleteAllEntries` from `kms.TestKmsDeleteAllKeys`?

mgaffney (changes requested):

internal/iam/repository_test.go
johanbrandhorst: That's a great idea

internal/db/testing.go
```diff
@@ -128,6 +129,20 @@ func TestWrapper(t testing.TB) wrapping.Wrapper {
 	return root
 }
 
+// TestDBWrapper initializes a DB wrapper for testing
+func TestDBWrapper(t testing.TB, conn *DB, purpose kms.KeyPurpose) wrapping.Wrapper {
```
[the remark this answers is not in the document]
johanbrandhorst: It's used in the `db_test` package external tests. We could handwrite this function in the test package instead if you like? I think it's a generically useful test method.

mgaffney (approved):

PR #2432, opened 2022-09-08
Minor documentation fixes for controller-led

irenarindos:


internal/cmd/commands/workerscmd/controller-led_funcs.go
```diff
@@ -19,7 +19,7 @@ func (c *ControllerLedCommand) extraControllerLedHelpFunc(helpMap map[string]fun
 			"",
 			"  Create a worker using the controller-led approach, receiving an activation token for a worker. Example:",
 			"",
-			`    $ boundary workers create controller-led -name us-east-1-1 "`,
+			`    $ boundary workers create controller-led -name us-east-1-1 `,
```
jefferai, line 22: There also weirdly seems to be an extra space at the end of the line, can you remove that too?

jefferai (approved):
Thank you!

internal/cmd/commands/workerscmd/controller-led_funcs.go
irenarindos: Done!

PR #2435, opened 2022-09-08
Use an atomic.Pointer to store wrapped DB objects

jefferai:
This allows atomically swapping to the new object. A timeout is introduced so that if the object is swapped after the existing one has been looked up it isn't immediately invalidated. The timeout is arbitrary but has been set to five minutes as anything running that long probably _should_ be interrupted.

This could be done with mutexes to avoid the arbitrary timeout, but that carries its own risk: if a write lock is attempted to be acquired while a query is holding a read lock, the write lock will block, and any subsequent read locks will also block since write locks take priority. As a result, if a query is long-running or timing out, all other queries in the system will become blocked. Our default user request timeout is 90 seconds and some jobs have no timeout at all, so this is not a purely theoretical concern, even if it's unlikely. So for the moment, using an atomic pointer feels gross but safer.

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -68,15 +68,34 @@ func (db *DB) Swap(ctx context.Context, newDB *DB) (closeDbFn, error) {
 	oldDbw := db.wrapped.Swap(newDB.wrapped.Load())
 	closeOldDbFn := func(ctx context.Context) {
 		go func() {
-			t := time.NewTimer(CloseSwappedDbDuration)
-			select {
-			case <-t.C:
-				if err := oldDbw.Close(ctx); err != nil {
-					event.WriteError(ctx, op, errors.Wrap(ctx, err, op))
+			maxTime := time.Now().Add(CloseSwappedDbDuration)
+			t := time.NewTicker(time.Second)
+			var done bool
+			for {
+				select {
+				case <-t.C:
+					if time.Now().After(maxTime) || done {
+						t.Stop()
+						if err := oldDbw.Close(ctx); err != nil {
+							event.WriteError(ctx, op, errors.Wrap(ctx, err, op))
+						}
+						return
+					}
+					sqlDb, err := oldDbw.SqlDB(ctx)
+					if err != nil {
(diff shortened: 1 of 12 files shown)


internal/db/db.go
```diff
@@ -42,42 +44,49 @@ func StringToDbType(dialect string) (DbType, error) {
...
-	db.mu.Unlock()
+	oldDbw := db.wrapped.Swap(newDB.wrapped.Load())
+	closeOldDbFn := func(ctx context.Context) {
+		go func() {
+			t := time.NewTimer(CloseSwappedDbDuration)
+			select {
+			case <-t.C:
+				if err := oldDbw.Close(ctx); err != nil {
+					event.WriteError(ctx, op, errors.Wrap(ctx, err, op))
+				}
```
hugoamvieira, line 76: This is my only concern, because it means that for any swap operation, we'll have 2  connections (per controller) that will just sit there for whatever's set in `CloseSwappedDbDuration`.

We've already got a significant problem with database connections and this will only push our average connection count up over time on deployments that make use of this feature.

As it stands, if we change the database credentials and switch over, we'd have a minimum of 5 connections per controller (assuming we lock the DB) for whatever is in `CloseSwappedDbDuration`, after which will revert to the normal 3.

image

I don't particularly have a code solution for this, and I understand why we need to do this, I just thought it was something that we should be aware of. My only idea at the moment is that perhaps on deployments that make use of this feature we could tweak `CloseSwappedDbDuration` to minimise any impact this could have.

hugoamvieira (approved):
Thanks Jeff! I've run some manual testing as well and it all seems to work perfectly 😄 

I only have one concern ⬇️

tmessi (approved):
awesome

internal/db/db.go
tmessi: We could try examining the [sql.DBStats](https://pkg.go.dev/database/sql#DBStats) to see if there is only idle connections in the pool and close instead of waiting the full time.

tmessi (approved):

PR #2436, opened 2022-09-08
Correct types of data key version referencers

johanbrandhorst:
The referencing views have to be dropped and recreated after the type change, or the migration fails.


johanbrandhorst:
Closing in favor of #2431

PR #2437, opened 2022-09-09
Adapt shared lock skip logic to dbswap branch

jefferai:


internal/cmd/commands/server/server.go
```diff
@@ -887,7 +894,7 @@ func (c *Command) reloadControllerDatabase(newConfig *config.Config) error {
 	}
 
 	// Acquire new lock on the new database and verify that it's in a good state to be used.
-	newDbSchemaManager, err := acquireDatabaseSharedLock(c.Context, newDb)
+	newDbSchemaManager, err := acquireSchemaManager(c.Context, newDb, c.Config.Controller.Database.SkipSharedLockAcquisition)
```
hugoamvieira, line 897:
```
if c.Config.Controller.Database.SkipSharedLockAcquisition {
			if err := c.schemaManager.Close(c.Context); err != nil {
				c.UI.Error(fmt.Errorf("Unable to release shared lock to the database: %w", err).Error())
				return base.CommandCliError
			}
		}
```

We need to do this for `newDbSchemaManager` too after we verify its state

jefferai: Good catch!

internal/db/schema/manager.go
```diff
@@ -6,13 +6,16 @@ import (
...
+	"sync"
 
 	"github.com/hashicorp/boundary/internal/db/schema/internal/edition"
 	"github.com/hashicorp/boundary/internal/db/schema/internal/log"
 	"github.com/hashicorp/boundary/internal/db/schema/internal/postgres"
 	"github.com/hashicorp/boundary/internal/db/schema/internal/provider"
 	"github.com/hashicorp/boundary/internal/db/schema/migration"
 	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/observability/event"
+	"go.uber.org/atomic"
```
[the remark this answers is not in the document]
jefferai: ah no I meant to use the new stdlib stuff...either is honestly fine but no reason to use this, will fix!

tmessi (approved):

hugoamvieira (approved):
Did a bunch more testing this morning, all looks great, thank you!

PR #2438, opened 2022-09-09
fix(worker): Improper reload when running as controller and worker

tmessi:
In cases where the server was started with both a controller and worker 
stanza, the reload of the config was not properly setting the initial 
upstream value on the new config. In this case the value is derived from 
other config values and not necessarily set directly.

This now properly initializes the new config so that initial upstream is 
set in the same manner as the first load.

Blame: 5d0cdf680f83bfc44f3c68237f087526911c944d


jefferai:
@tmessi To make reviewing easier, can you just say whether or not you changed the logic that you moved or refactored into functions? That'll just save me doing a line-by-line comparison to see if anything changed :-)

tmessi:
> @tmessi To make reviewing easier, can you just say whether or not you changed the logic that you moved or refactored into functions? That'll just save me doing a line-by-line comparison to see if anything changed :-)

The 2 refact commits are refactors, so no logic changed, the fix commit is the only new functionality.

jefferai (approved):
Thanks!!!

hugoamvieira (commented):
Looks good! 

`TestServer_ReloadWorkerTags` is failing at the moment though - I re-ran the suite and it doesn't seem due to flakiness

tmessi:
> Looks good!
> 
> `TestServer_ReloadWorkerTags` is failing at the moment though - I re-ran the suite and it doesn't seem due to flakiness

Fixed in [cb45d84](https://github.com/hashicorp/boundary/pull/2438/commits/cb45d84673a9d0fec6c35d77de0db29112a37aab), `SetupControllerPublicClusterAddress` sets `c.Controller` if it is nil, which was causing the upstream to not change because it hit the case where there was a controller and worker, but then the upstreams did not match the cluster adder, so an error happened during the reload.

PR #2442, opened 2022-09-13
Add health endpoint for worker

talanknight:
Health endpoints for workers and controllers now have an optional query parameter "worker_info".  If a worker is running in the process a new key is added to the returned json containing the worker info.  The worker info contains:
* Operation State of the worker
* How many sessions are being handled by the worker
* How many connections each session has on the worker

When a worker is running but a controller is not it'll always return a 200.  When a controller and worker are both running the status code set by the controller is used (sometimes a 503).


hugoamvieira (approved):

internal/daemon/controller/handlers/health/health_test.go
```diff
@@ -24,20 +39,50 @@ func TestGetHealth(t *testing.T) {
...
 		{
 			name:                 "service unavailable reply",
 			ctx:                  grpc.NewContextWithServerTransportStream(context.Background(), &testServerTransportStream{expHttpCode: "503"}),
 			serviceUnavailable:   true,
+			request:              &services.GetHealthRequest{},
 			expGetHealthResponse: &services.GetHealthResponse{},
 			expErr:               false,
 		},
+		{
+			name:                 "service unavailable withworker info reply",
```
irenarindos, line 73:
```suggestion
			name:                 "service unavailable with worker info reply",
```

irenarindos (approved):

PR #2443, opened 2022-09-13
feat(scheduler): set intervals from config

louisruch:
Run a manual test with the following config locally:
```
controller {
  name = "d1"
  description = "A controller for a demo 1!"

  database {
      url = "postgresql://127.0.0.1/boundary?sslmode=disable"
  }

  scheduler {
    job_run_interval = "1s"
    monitor_interval = "1s"
  }
}
```

I also added a log every-time the ticker was reset and got an event each second from both loops

### DIFF
--- a/internal/cmd/config/config.go
+++ b/internal/cmd/config/config.go
@@ -132,10 +132,11 @@ type Config struct {
 }
 
 type Controller struct {
-	Name              string    `hcl:"name"`
-	Description       string    `hcl:"description"`
-	Database          *Database `hcl:"database"`
-	PublicClusterAddr string    `hcl:"public_cluster_addr"`
+	Name              string     `hcl:"name"`
+	Description       string     `hcl:"description"`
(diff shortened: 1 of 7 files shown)


internal/cmd/config/config.go
```diff
@@ -387,6 +397,24 @@ func Parse(d string) (*Config, error) {
 			result.Controller.GracefulShutdownWaitDuration = t
 		}
 
+		if result.Controller.Scheduler != nil {
+			if result.Controller.Scheduler.JobRunInterval != "" {
```
ddebko, line 401: does the hcl parser guarantee that the interface will not be a nil value, and default it to an empty string?

ddebko (approved):
LGTM

internal/cmd/config/config.go
louisruch: It seems like it must since other examples do this and the test with a scheduler block and no intervals does parse correctly.

PR #2444, opened 2022-09-13
fix: return an error for vault tokens that will expire before the scheduler can run

ddebko:
### Summary:

Currently the Vault credential store API accepts any token TTL, we validate the TTL and store the next renewal time during create and update. We should return an error if that TTL cannot be honored due to the scheduler interval for running jobs.

### Fix:

Now we compare the ttl value to the runJobsInterval value and return an error to the user when we cannot honor the token renewal.

### Error Output:
`scheduler interval must be greater than token ttl. scheduler jobs interval value: 2s`

### DIFF
--- a/internal/credential/vault/repository_credential_store.go
+++ b/internal/credential/vault/repository_credential_store.go
@@ -3,6 +3,7 @@ package vault
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/hashicorp/boundary/internal/db"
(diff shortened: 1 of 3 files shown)


internal/credential/vault/repository_credential_store.go
```diff
@@ -116,6 +117,11 @@ func (r *Repository) CreateCredentialStore(ctx context.Context, cs *CredentialSt
 		return nil, err
 	}
 
+	runJobsInterval := r.scheduler.GetRunJobsInterval()
+	if token.expiration <= runJobsInterval {
+		return nil, errors.Wrap(ctx, fmt.Errorf("scheduler interval must be greater than token ttl. scheduler jobs interval value: %s seconds", strconv.FormatFloat(runJobsInterval.Seconds(), 'f', -1, 64)), op)
```
louisruch, line 122: Do we have other examples in Boundary where we print a duration? Getting it in seconds is probably fine, but we could just print out using golang's default duration print.

louisruch (commented):
LGTM, I did have one question about the duration format for the error message

internal/credential/vault/repository_credential_store.go
ddebko: I found one example where we print the time.Duration variable using the String() func, which I just tested out and it does look nicer.

louisruch (approved):
LGTM!

PR #2446, opened 2022-09-13
fix: show user warning log when token period is less than jobRunInterval

ddebko:
### Summary:

The scheduler by default only runs every minute, if a credential is Issued by Vault that has a lower TTL of 1 minute or what every the scheduler interval is set to, we should warn the user that this credential will expire before we can renew it.

### Solution:

Check the metadata from the vault secret to validate the token period is greater than the runJobsInterval. When we cannot renew the token in time, we log the issue.

### DIFF
--- a/internal/credential/vault/repository_credentials.go
+++ b/internal/credential/vault/repository_credentials.go
@@ -2,11 +2,13 @@ package vault
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/hashicorp/boundary/internal/credential"
 	"github.com/hashicorp/boundary/internal/db"
 	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/observability/event"
 )
 
 var _ credential.Issuer = (*Repository)(nil)
@@ -33,11 +35,22 @@ func (r *Repository) Issue(ctx context.Context, sessionId string, requests []cre
 
 	var creds []credential.Dynamic
 	var minLease time.Duration
+	runJobsInterval := r.scheduler.GetRunJobsInterval()
 	for _, lib := range libs {
 		cred, err := lib.retrieveCredential(ctx, op, sessionId)
 		if err != nil {
 			return nil, err
 		}
+
+		if cred.getExpiration() < runJobsInterval {
+			event.WriteError(ctx, op,
+				fmt.Errorf("WARNING: credential will expire before job scheduler can run"),
+				event.WithInfo("credential_public_id", cred.GetPublicId()),
+				event.WithInfo("credential_library_public_id", lib.GetPublicId()),
+				event.WithInfo("runJobsInterval", runJobsInterval),
+			)
+		}
+
 		if minLease > cred.getExpiration() {
 			minLease = cred.getExpiration()
 		}


internal/credential/vault/repository_credential_store.go
```diff
@@ -121,6 +122,14 @@ func (r *Repository) CreateCredentialStore(ctx context.Context, cs *CredentialSt
 		return nil, errors.Wrap(ctx, fmt.Errorf("scheduler interval must be greater than token ttl. scheduler jobs interval value: %s", runJobsInterval.String()), op)
 	}
 
+	period, err := parseutil.ParseDurationSecond(tokenLookup.Data["period"])
```
louisruch, line 125: I think we are safe on the credential store with the previous PR that covers the token TTL because we always renew the token to the period when we create/update the token in the store.

This warning should be logged when a credential is issued from a credential library [here](https://github.com/hashicorp/boundary/blob/main/internal/credential/vault/repository_credentials.go#L41)

louisruch (changes requested):
Sorry if the details in the ticket was not clear see inline comment

internal/credential/vault/repository_credentials.go
```diff
@@ -38,6 +41,16 @@ func (r *Repository) Issue(ctx context.Context, sessionId string, requests []cre
 		if err != nil {
 			return nil, err
 		}
+
+		period, err := parseutil.ParseDurationSecond(cred.getExpiration())
```
louisruch, line 45: Do we need to parse this, can we not just do `if cred.getExpiration() < r.scheduler.GetRunJobsInterval()`

internal/credential/vault/repository_credentials.go
```diff
@@ -38,6 +41,16 @@ func (r *Repository) Issue(ctx context.Context, sessionId string, requests []cre
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to parse vault token period")
+		}
+		runJobsInterval := r.scheduler.GetRunJobsInterval()
+		if period < runJobsInterval {
+			event.WriteError(ctx, op, fmt.Errorf("WARNING: token ttl expires before job scheduler can run"), event.WithInfo("runJobsInterval", runJobsInterval))
```
louisruch, line 51: This is calling out a warning on the token ttl but should probably be `WARNING: credential expires before...`

internal/credential/vault/repository_credentials.go
ddebko: good point

internal/credential/vault/repository_credentials.go
```diff
@@ -38,6 +40,12 @@ func (r *Repository) Issue(ctx context.Context, sessionId string, requests []cre
 		if err != nil {
 			return nil, err
 		}
+
+		runJobsInterval := r.scheduler.GetRunJobsInterval()
+		if cred.getExpiration() < runJobsInterval {
+			event.WriteError(ctx, op, fmt.Errorf("WARNING: credential expires before job scheduler can run"), event.WithInfo("runJobsInterval", runJobsInterval))
```
louisruch, line 46: Sorry :)
So I am only catching this now, but I noticed this is in a loop and if multiple credentials have expiration lower than the scheduler run interval we will spam the same message. I think multiple events are fine, but perhaps we should add the credential library public id in the `withInfo`

louisruch: We might also not want this to be an error, perhaps `WriteSysEvent` or `WriteObservation` I would scan the code/look at the eventer documentation and validate which is better for a warning

internal/credential/vault/repository_credentials.go
```diff
@@ -35,15 +35,16 @@ func (r *Repository) Issue(ctx context.Context, sessionId string, requests []cre
...
 	for _, lib := range libs {
 		cred, err := lib.retrieveCredential(ctx, op, sessionId)
 		if err != nil {
 			return nil, err
 		}
 
-		runJobsInterval := r.scheduler.GetRunJobsInterval()
 		if cred.getExpiration() < runJobsInterval {
-			event.WriteError(ctx, op, fmt.Errorf("WARNING: credential expires before job scheduler can run"), event.WithInfo("runJobsInterval", runJobsInterval))
+			msgs = append(msgs, event.WithInfo("credential_public_id", cred.GetPublicId()))
```
louisruch, line 47: We should add info about the credential library as well since that contains the information needed for the admin to update Vault to set an appropriate TTL

louisruch (approved):
One comment about adding library public id but otherwise LGTM

louisruch (approved):
This is great thanks!!

PR #2449, opened 2022-09-13
[docs] Address reported bugs

yhyakuna:
This PR address the reported issues:

- Remove [broken link](https://www.boundaryproject.io/docs/concepts/filtering/events) -- since there is a link to [common sink parameter]() doc page, I removed the hyperlink on `sink`
   image

-  For now, made the `https://hashicorp.com/boundary.io/ServerName/Path/FileName` to be a text (not a hyperlink) --> we probably need little more description though...
- Add missing title on [AEAD KMS](https://www.boundaryproject.io/docs/configuration/kms/aead) doc page
- Flip the order of KMS worker and PKI worker


stellarsquall (approved):
👍

PR #2451, opened 2022-09-14
cli(ssh): Do not print private key if consumed

louisruch:
Came up during a demo, I see we mark the copied cred as consumed but never set it in the returned creds


ddebko (approved):
LGTM

PR #2453, opened 2022-09-14
address op state review feedback

irenarindos:
Add view migration comment
Only use active workers for the target service; allow workers in an unknown state if they have no release version for backwards compatibility.

### DIFF
--- a/internal/server/repository_worker.go
+++ b/internal/server/repository_worker.go
@@ -200,7 +200,7 @@ func (r *Repository) ListWorkers(ctx context.Context, scopeIds []string, opt ...
 	if opts.withExcludeShutdownWorkers {
 		// If a worker has a version, it must be active. If it's an older worker (no release version), we'll include it for
 		// backwards compatibilty
-		where = append(where, fmt.Sprintf("(operational_state = '%s' or (release_version = '') IS NOT FALSE)", ActiveOperationalState.String()))
+		where = append(where, fmt.Sprintf("(operational_state = '%s' or length(trim(release_version)) = 0 or release_version is null)", ActiveOperationalState.String()))
 	}
 
 	limit := r.defaultLimit
--- a/internal/server/repository_worker.go
+++ b/internal/server/repository_worker.go
@@ -198,8 +198,9 @@ func (r *Repository) ListWorkers(ctx context.Context, scopeIds []string, opt ...
 	}
 
 	if opts.withExcludeShutdownWorkers {
-		where = append(where, "operational_state not in (?)")
-		whereArgs = append(whereArgs, ShutdownOperationalState.String())
+		// If a worker has a version, it must be active. If it's an older worker (no release version), we'll include it for
+		// backwards compatibilty
+		where = append(where, fmt.Sprintf("(operational_state = '%s' or length(trim(release_version)) = 0)", ActiveOperationalState.String()))
 	}
 
 	limit := r.defaultLimit
--- a/internal/server/repository_worker.go
+++ b/internal/server/repository_worker.go
@@ -200,7 +200,7 @@ func (r *Repository) ListWorkers(ctx context.Context, scopeIds []string, opt ...
 	if opts.withExcludeShutdownWorkers {
 		// If a worker has a version, it must be active. If it's an older worker (no release version), we'll include it for
 		// backwards compatibilty
-		where = append(where, fmt.Sprintf("(operational_state = '%s' or length(trim(release_version)) = 0)", ActiveOperationalState.String()))
+		where = append(where, fmt.Sprintf("(operational_state = '%s' or (release_version = '') IS NOT FALSE)", ActiveOperationalState.String()))
 	}
 
 	limit := r.defaultLimit
--- a/internal/server/repository_worker.go
+++ b/internal/server/repository_worker.go
@@ -164,7 +164,7 @@ func lookupWorker(ctx context.Context, reader db.Reader, id string) (*Worker, er
 // then the last status update time is ignored.
 // If WithLimit < 0, then unlimited results are returned. If WithLimit == 0, then
 // default limits are used for results.
-// Also supports: WithWorkerType, WithExcludeShutdownWorkers
+// Also supports: WithWorkerType, WithActiveWorkers
 func (r *Repository) ListWorkers(ctx context.Context, scopeIds []string, opt ...Option) ([]*Worker, error) {
 	const op = "server.(Repository).ListWorkers"
 	switch {
@@ -197,10 +197,9 @@ func (r *Repository) ListWorkers(ctx context.Context, scopeIds []string, opt ...
 		return nil, errors.New(ctx, errors.InvalidParameter, op, fmt.Sprintf("unknown worker type %v", opts.withWorkerType))
 	}
 
-	if opts.withExcludeShutdownWorkers {
-		// If a worker has a version, it must be active. If it's an older worker (no release version), we'll include it for
-		// backwards compatibilty
-		where = append(where, fmt.Sprintf("(operational_state = '%s' or length(trim(release_version)) = 0 or release_version is null)", ActiveOperationalState.String()))
+	if opts.withActiveWorkers {
+		where = append(where, "operational_state = ?")
+		whereArgs = append(whereArgs, ActiveOperationalState.String())
 	}
 
 	limit := r.defaultLimit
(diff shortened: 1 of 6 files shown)


internal/server/repository_worker.go
```diff
@@ -198,8 +198,9 @@ func (r *Repository) ListWorkers(ctx context.Context, scopeIds []string, opt ...
 	}
 
 	if opts.withExcludeShutdownWorkers {
-		where = append(where, "operational_state not in (?)")
-		whereArgs = append(whereArgs, ShutdownOperationalState.String())
+		// If a worker has a version, it must be active. If it's an older worker (no release version), we'll include it for
+		// backwards compatibilty
+		where = append(where, fmt.Sprintf("(operational_state = '%s' or length(trim(release_version)) = 0)", ActiveOperationalState.String()))
```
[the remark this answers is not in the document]
irenarindos: My bad, I've changed this to be `or (release_version = '') IS NOT FALSE` so it works for null or empty strings

internal/server/repository_worker.go
```diff
@@ -198,8 +198,9 @@ func (r *Repository) ListWorkers(ctx context.Context, scopeIds []string, opt ...
 	}
 
 	if opts.withExcludeShutdownWorkers {
-		where = append(where, "operational_state not in (?)")
-		whereArgs = append(whereArgs, ShutdownOperationalState.String())
+		// If a worker has a version, it must be active. If it's an older worker (no release version), we'll include it for
+		// backwards compatibilty
+		where = append(where, fmt.Sprintf("(operational_state = '%s' or (release_version = '') IS NOT FALSE)", ActiveOperationalState.String()))
```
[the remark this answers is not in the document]
mgaffney: Can we be explicit about nulls by adding another `or` statement? 

```go
where = append(where, fmt.Sprintf("(operational_state = '%s' or length(trim(release_version)) = 0 or release_version is null)", ActiveOperationalState.String()))
```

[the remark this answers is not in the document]
irenarindos: Done!

internal/server/repository_worker_test.go
```diff
@@ -691,12 +695,24 @@ func TestListWorkers_WithExcludeShutdown(t *testing.T) {
...
 			server.WithAddress(worker3.GetAddress()),
+			server.WithReleaseVersion("Boundary v.0.11"),
 			server.WithOperationalState(server.ShutdownOperationalState.String())),
 		server.WithPublicId(worker3.GetPublicId()))
 	require.NoError(err)
 	result, err = serversRepo.ListWorkers(ctx, []string{scope.Global.String()}, server.WithExcludeShutdownWorkers(true))
 	require.NoError(err)
 	require.Len(result, 0)
+
+	// Upsert without a release version or state and expect to get a hit- test backwards compatibility
```
mgaffney, line 706: Can we add tests to cover each of the `or` parts of the where clause?

mgaffney (changes requested):

internal/server/repository_worker_test.go
irenarindos: The existing test should cover the nil version; I've added one for an empty string version. Thanks!

internal/server/repository_worker_test.go
```diff
@@ -691,12 +695,36 @@ func TestListWorkers_WithExcludeShutdown(t *testing.T) {
...
+
+	// Upsert with an empty string release version and no state
+	_, err = serversRepo.UpsertWorkerStatus(ctx,
+		server.NewWorker(scope.Global.String(),
+			server.WithName(worker3.GetName()),
+			server.WithAddress(worker3.GetAddress()),
+			server.WithReleaseVersion("")),
+		server.WithPublicId(worker3.GetPublicId()))
+	require.NoError(err)
+	result, err = serversRepo.ListWorkers(ctx, []string{scope.Global.String()}, server.WithExcludeShutdownWorkers(true))
```
mgaffney, line 725: I'm not sure we are testing all of the combinations in the `where` clause. These are the ones I came up with, but please let me know if I'm wrong:

(`¯\_(ツ)_/¯` represents a `null` value, and the big blank space is a string with spaces)

```
test case │ operational_state │ release_version │ return?
══════════╪═══════════════════╪═════════════════╪═══════════
     1    │ active            │ Boundary v.0.11 │  True
     2    │ active            │ ¯\_(ツ)_/¯      │  True
     3    │ active            │                 │  True
     4    │ shutdown          │ Boundary v.0.11 │  False
     5    │ shutdown          │ ¯\_(ツ)_/¯      │  True
     6    │ shutdown          │                 │  True
```

I queried the test database right before each call to `serversRepo.ListWorkers` and got these results:

```
select public_id, operational_state, release_version from server_worker;

At line 663:
  public_id   │ operational_state │ release_version    test case
══════════════╪═══════════════════╪══════════════════
 w_13VWxoTRI1 │ active            │ Boundary v0.11.0   1
 w_lsFI8gEYUC │ active            │ Boundary v0.11.0   1
 w_BOns6xVfAW │ active            │ Boundary v0.11.0   1

At line 678:
  public_id   │ operational_state │ release_version
══════════════╪═══════════════════╪══════════════════
 w_B1LF31hSjA │ active            │ Boundary v0.11.0   1
 w_wJmlVvxp8v │ active            │ Boundary v0.11.0   1
 w_lS4zt6IwAR │ shutdown          │ Boundary v.0.11    4

At line 690:
  public_id   │ operational_state │ release_version
══════════════╪═══════════════════╪══════════════════
 w_i6wxYZZcjY │ active            │ Boundary v0.11.0   1
 w_bKS63cjh6T │ shutdown          │ Boundary v.0.11    4
 w_v5gYvY3nAz │ shutdown          │ Boundary v.0.11    4

At line 702:
  public_id   │ operational_state │ release_version
══════════════╪═══════════════════╪═════════════════
 w_x8iaBdptE9 │ shutdown          │ Boundary v.0.11    4
 w_Re8qJcxg7i │ shutdown          │ Boundary v.0.11    4
 w_vO92o9ikUp │ shutdown          │ Boundary v.0.11    4

At line 713:
  public_id   │ operational_state │ release_version
══════════════╪═══════════════════╪═════════════════
 w_x8iaBdptE9 │ shutdown          │ Boundary v.0.11    4
 w_Re8qJcxg7i │ shutdown          │ Boundary v.0.11    4
 w_vO92o9ikUp │ active            │ ¯\_(ツ)_/¯         2

At line 725:
  public_id   │ operational_state │ release_version
══════════════╪═══════════════════╪═════════════════
 w_KiJjHwNUDl │ shutdown          │ Boundary v.0.11    4
 w_Oh5yP6wVhv │ shutdown          │ Boundary v.0.11    4
 w_9BbPwH6zPk │ active            │ ¯\_(ツ)_/¯         2
```
So we have tests for test cases 1, 2, and 4, but not for 3, 5, and 6.

irenarindos: Ahh thanks! Old workers (pre 0.11.0) won't report operational state or release version and will default to Active, so cases 5 and 6 can't happen but that lead me to look at other cases... 

Old and new workers will have an Unknown state with no version on Create and will only be updated after a status update. New workers will report state/ version on status and old workers will default to active after status. A controller-led worker would be in an Unknown state/ without version until a user starts it & and it sends status, so we wouldn't want to send it work.

I've updated the where-clause to only include active workers and updated the test cases to cover these.

mgaffney: I think it would be helpful to convert the existing tests to a table-driven test structure. That would allow each test to have a name and it would make it easier to add new ones. What do you think?

irenarindos: Sure, I'll do that!

internal/server/repository_worker_test.go
```diff
@@ -662,41 +664,106 @@ func TestListWorkers_WithExcludeShutdown(t *testing.T) {
...
+			},
+			wantCnt:   0,
+			wantState: server.UnknownOperationalState.String(),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			worker, err := tt.upsertFn()
+			require.NoError(err)
+			got, err := serversRepo.ListWorkers(ctx, []string{scope.Global.String()}, server.WithActiveWorkers(true))
```
mgaffney, line 759: Does the test function name still make sense `TestListWorkers_WithExcludeShutdown`?

mgaffney (changes requested):
Can we rename the test function since the option has changed?

mgaffney (approved):

PR #2455, opened 2022-09-14
Worker graceful shutdown

irenarindos:
- Changed the shutdown channel so that it stays open to listen for more interrupts
- Addition of worker graceful shutdown- on first ctrl+c, worker will start graceful shutdown (waiting for active connections to drain). Second will cause the worker to move on to shutdown. Third will force shutdown immediately.

### DIFF
--- a/internal/cmd/commands/dev/dev.go
+++ b/internal/cmd/commands/dev/dev.go
@@ -903,7 +903,7 @@ func (c *Command) Run(args []string) int {
 		}
 	}
 
-	for !errorEncountered.Load() && (!workerShutdownDone.Load() && !controllerShutdownDone.Load()) {
+	for !errorEncountered.Load() && (!workerShutdownDone.Load() || !controllerShutdownDone.Load()) {
 		select {
 		case <-c.ServerSideShutdownCh:
 			c.UI.Output("==> Boundary dev environment self-terminating")
@@ -919,10 +919,8 @@ func (c *Command) Run(args []string) int {
 			n := runtime.Stack(buf[:], true)
 			event.WriteSysEvent(context.TODO(), op, "goroutine trace", "stack", string(buf[:n]))
 
-		case <-time.After(10 * time.Millisecond):
-			if workerShutdownDone.Load() && controllerShutdownDone.Load() {
-				break
-			}
+		case <-c.Context.Done():
+			break
 		}
 	}
 
--- a/internal/cmd/commands/dev/dev.go
+++ b/internal/cmd/commands/dev/dev.go
@@ -870,8 +870,8 @@ func (c *Command) Run(args []string) int {
 		count := atm.LoadUint32(&shutdownTriggerCount)
 		switch {
 		case count == 1:
+			c.ContextCancel()
 			go func() {
-				c.ContextCancel()
 				if c.Config.Controller != nil {
 					c.opsServer.WaitIfHealthExists(c.Config.Controller.GracefulShutdownWaitDuration, c.UI)
 				}
--- a/internal/cmd/commands/dev/dev.go
+++ b/internal/cmd/commands/dev/dev.go
@@ -870,15 +870,19 @@ func (c *Command) Run(args []string) int {
 				}
 
 				if !c.flagControllerOnly {
-					result := workerGracefulShutdownFunc()
-					if result == nil {
-						workerShutdownDone.Store(true)
-					}
+					go func() {
+						result := workerGracefulShutdownFunc()
+						if result == nil {
+							workerShutdownDone.Store(true)
+						}
+					}()
 				}
 
-				if err := controllerShutdownFunc(); err == nil {
-					controllerShutdownDone.Store(true)
-				}
+				go func() {
+					if err := controllerShutdownFunc(); err == nil {
+						controllerShutdownDone.Store(true)
+					}
+				}()
 
 				err := c.opsServer.Shutdown()
 				if err != nil {
@@ -888,14 +892,20 @@ func (c *Command) Run(args []string) int {
 		case count == 2 && !c.flagControllerOnly:
 			go func() {
 				if !c.flagControllerOnly && !workerShutdownDone.Load() {
-					if err := workerShutdownFunc(); err == nil {
-						workerShutdownDone.Store(true)
-					}
-				} else if !controllerShutdownDone.Load() {
-					if err := controllerShutdownFunc(); err == nil {
-						controllerShutdownDone.Store(true)
-					}
-				} else {
+					go func() {
+						if err := workerShutdownFunc(); err == nil {
+							workerShutdownDone.Store(true)
+						}
+					}()
+				}
+				if !controllerShutdownDone.Load() {
+					go func() {
+						if err := controllerShutdownFunc(); err == nil {
+							controllerShutdownDone.Store(true)
+						}
+					}()
+				}
+				if workerShutdownDone.Load() && controllerShutdownDone.Load() {
 					c.UI.Error("Forcing shutdown")
 					os.Exit(base.CommandUserError)
 				}
--- a/internal/cmd/commands/dev/dev.go
+++ b/internal/cmd/commands/dev/dev.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"runtime"
 	"strings"
+	sync "sync/atomic"
 	"time"
 
 	"github.com/hashicorp/boundary/globals"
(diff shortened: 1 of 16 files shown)


internal/cmd/commands/dev/dev.go
```diff
@@ -824,20 +822,27 @@ func (c *Command) Run(args []string) int {
...
-			signal.Notify(shutdownCh, os.Interrupt, syscall.SIGTERM)
 			for {
 				select {
-				case <-shutdownCh:
+				case <-c.ShutdownCh:
+					if !workerShutdownTriggered && !c.flagControllerOnly {
+						workerShutdownTriggered = true
+						if err := c.worker.Shutdown(); err != nil {
+							c.UI.Error(fmt.Errorf("Error shutting down worker: %w", err).Error())
+						}
```
jefferai, line 844: Keeping in mind that earlier the shutdown channel would be closed -- which is why it created a new channel here to listen for a second interrupt -- there is now a race condition between readers of `c.ShutdownCh`...depending on which one gets it first you may hit this case, or you may hit the case below and create a new goroutine running this code. Additionally when this is first run by the first struct down `c.ShutdownCh`, this case won't execute until a second interrupt comes in -- not sure if that's what you wanted or not.

There is also a data race on `workerShutdownTriggered` since more than one goroutine here can be launched concurently.

Additionally once `gracefulShutdownTriggered` in the case below the `for` loop will exit so you'll immediately return from this function.

Given the changes to `c.ShutdownCh`, I think what you may want to do is move the `gracefulShutdownTriggered` logic into `runShutdownLogic` so the for loop below is just a signal handler. You could instantiate an atomic uint32 that keeps track of how many interrupts have been received and is atomically incremented before `runShutdownLogic` is called; the function then switches logic based on that value. The first two interrupts trigger goroutines to either do a graceful shutdown or a non-graceful shutdown; the third runs `os.Exit`. Spitballing a bit there but I think something like that may be easier to reason about and avoid some of these signal handler issues.

internal/cmd/commands/dev/dev.go
```diff
@@ -824,73 +823,84 @@ func (c *Command) Run(args []string) int {
...
+				if err != nil {
+					c.UI.Error(fmt.Errorf("Failed to shutdown ops listeners: %w", err).Error())
 				}
-			}
-		}()
+				shutdownDone = true
+			}()
+		}
+
+		if sync.LoadUint32(&shutdownTriggerCount) == 2 && !c.flagControllerOnly {
```
jefferai, line 863: These if conditions would probably be a bit cleaner by assigning the value of load to a local var at the start of the function and then using a `switch` statement.

internal/cmd/commands/dev/dev.go
```diff
@@ -824,73 +823,84 @@ func (c *Command) Run(args []string) int {
...
+				}
+
+				if !c.flagControllerOnly {
+					if !c.flagWorkerAuthStorageSkipCleanup && c.worker.WorkerAuthStorage != nil {
+						c.worker.WorkerAuthStorage.Cleanup()
+					}
+					if err := c.worker.GracefulShutdown(); err != nil {
+						c.UI.Error(fmt.Errorf("Error shutting down worker gracefully: %w", err).Error())
+					}
+					if err := c.worker.Shutdown(); err != nil {
```
jefferai, line 846: I'm not sure if this was possible before or not but we do run a risk here of calling `c.worker.Shutdown` more than once concurrently. That may not really be an issue, any errors that arise will be logged but probably be okay.

internal/cmd/commands/dev/dev.go
```diff
@@ -824,73 +823,84 @@ func (c *Command) Run(args []string) int {
...
-		go func() {
-			shutdownCh := make(chan os.Signal, 4)
-			signal.Notify(shutdownCh, os.Interrupt, syscall.SIGTERM)
-			for {
-				select {
-				case <-shutdownCh:
-					c.UI.Error("Forcing shutdown")
-					os.Exit(base.CommandUserError)
+		if sync.LoadUint32(&shutdownTriggerCount) == 1 {
+			go func() {
```
jefferai, line 834: The logic in here currently makes an assumption that `c.controller.Shutdown()` will only be called during the first SIGINT and that we will actually reach it -- but if we're stuck waiting on graceful worker shutdown we may not. We may want the second SIGINT to also trigger an immediate controller shutdown, and save the `os.Exit` for the third in all cases.

There would be some knock-on effects. Probably there would need to be separate goroutines for worker shutdown and controller shutdown to run concurrently to ensure we're not blocking controller shutdown at that point on worker shutdown -- something we _don't_ want in the graceful case! And that also means that `shutdownDone` becomes harder to understand -- right now it's set when the functions return, but what if we end up running controller shutdown concurrently and one ends up returning immediately with an error? Possibly the thing to do would be to only set `shutdownDone` if the shutdown func returns with no error.

internal/cmd/commands/dev/dev.go
```diff
@@ -824,73 +823,84 @@ func (c *Command) Run(args []string) int {
...
-	// Wait for shutdown
-	shutdownTriggered := false
-
-	// Add a force-shutdown goroutine to consume more interrupts
-	abortForceShutdownCh := make(chan struct{})
-	defer close(abortForceShutdownCh)
+	// On first interrupt, enter graceful shutdown for controllers and workers (workers wait for connections to drain)
+	// On second interrupt, force controller shutdown and send workers into shutdown (stop all connections)
+	// On third+ interrupt, force worker shutdown
+	shutdownDone := false
```
jefferai, line 829: This is read and written across goroutines so needs to be an atomic value.

internal/cmd/commands/dev/dev.go
```diff
@@ -824,73 +823,84 @@ func (c *Command) Run(args []string) int {
...
-
-			if err := c.controller.Shutdown(); err != nil {
-				c.UI.Error(fmt.Errorf("Error shutting down controller: %w", err).Error())
-			}
-
-			err := c.opsServer.Shutdown()
-			if err != nil {
-				c.UI.Error(fmt.Errorf("Failed to shutdown ops listeners: %w", err).Error())
+		default:
+			if shutdownDone {
```
jefferai, line 902: This will cause very rapid spinning in the for loop because it will keep selecting this case over and over until one of the others is true. What I think you should do instead of `default` is do something like `time.After(10*time.Millisecond)` so that we don't end up creating a spin loop but also can react quickly when it's done.

internal/cmd/commands/dev/dev.go
irenarindos: The first check in worker.Shutdown() is whether or not shutdown has been initiated so we should be ok
```
	if !w.started.Load() {
		event.WriteSysEvent(w.baseContext, op, "already shut down, skipping")
		return nil
	}
```

internal/cmd/commands/dev/dev.go
irenarindos: Thanks!

internal/cmd/commands/server/worker_initial_upstreams_reload_test.go
```diff
@@ -135,7 +135,6 @@ pollSecondController:
 		}
 	}
 
-	close(cmd.ShutdownCh)
-
+	cmd.ShutdownCh <- struct{}{}
```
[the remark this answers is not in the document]
irenarindos: I think this is possibly taken care of by dev.go and server.go [using this](https://github.com/hashicorp/boundary/blob/e690ac6048d76f49a4c4729bb288d1e9157f3346/internal/cmd/base/base.go#L151-L160)

Tagging @jefferai for discussion since he suggested changing this to a counting chan

internal/cmd/commands/server/server.go
```diff
@@ -667,69 +666,113 @@ func (c *Command) StartWorker() error {
...
+				if c.Config.Worker != nil && !workerShutdownDone.Load() {
+					go func() {
+						if err := workerShutdownFunc(); err == nil {
+							workerShutdownDone.Store(true)
+						}
+					}()
+				}
+				if !controllerShutdownDone.Load() {
+					go func() {
+						if err := controllerShutdownFunc(); err == nil {
```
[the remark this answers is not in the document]
irenarindos: thank you!

internal/cmd/commands/dev/dev.go
```diff
@@ -824,73 +824,104 @@ func (c *Command) Run(args []string) int {
...
-			if err := c.controller.Shutdown(); err != nil {
-				c.UI.Error(fmt.Errorf("Error shutting down controller: %w", err).Error())
-			}
 
-			err := c.opsServer.Shutdown()
-			if err != nil {
-				c.UI.Error(fmt.Errorf("Failed to shutdown ops listeners: %w", err).Error())
+		case <-time.After(10 * time.Millisecond):
+			if workerShutdownDone.Load() && controllerShutdownDone.Load() {
+				break
```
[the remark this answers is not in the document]
irenarindos: The forloop also checks !errorEncountered.Load() which will prevent the for loop from terminating

internal/cmd/commands/server/server.go
```diff
@@ -667,69 +667,98 @@ func (c *Command) StartWorker() error {
...
+
+		case count >= 2:
+			c.UI.Error("Forcing shutdown")
+			os.Exit(base.CommandCliError)
+		}
 		shutdownTriggered = true
 	}
 
-	for !shutdownTriggered {
+	for !shutdownTriggered { //!workerShutdownDone.Load() && !controllerShutdownDone.Load() {
```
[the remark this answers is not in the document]
irenarindos: thanks!

internal/cmd/commands/dev/dev.go
```diff
@@ -824,73 +823,121 @@ func (c *Command) Run(args []string) int {
...
```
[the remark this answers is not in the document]
irenarindos: I've switched to this, thanks!

internal/daemon/worker/worker.go
```diff
@@ -423,6 +423,38 @@ func (w *Worker) Start() error {
...
+// Graceful shutdown sets the worker state to "shutdown" and will wait to return until there
+// are no longer any active connections.
+func (w *Worker) GracefulShutdown() error {
+	const op = "worker.(Worker).GracefulShutdown"
+	event.WriteSysEvent(w.baseContext, op, "worker entering graceful shutdown")
+	w.operationalState.Store(server.ShutdownOperationalState)
+
+	// Wait for connections to drain
+	for w.hasActiveConnection() {
+		time.Sleep(time.Second)
```
jefferai, line 451: Assuming that GetLocalConnections doesn't make a database call, I think you could actually do this more often, maybe 250ms. But we can always tweak later!

internal/cmd/commands/server/server.go
```diff
@@ -667,69 +667,100 @@ func (c *Command) StartWorker() error {
...
 func (c *Command) WaitForInterrupt() int {
 	const op = "server.(Command).WaitForInterrupt"
-	// Wait for shutdown
-	shutdownTriggered := false
 
-	// Add a force-shutdown goroutine to consume another interrupt
-	abortForceShutdownCh := make(chan struct{})
-	defer close(abortForceShutdownCh)
+	var shutdownCompleted atm.Bool
+	shutdownCompleted.Store(false)
```
jefferai, line 672: The zero value is false per the docs so you don't need the Store(false) here or below

internal/cmd/commands/dev/dev.go
```diff
@@ -824,74 +824,103 @@ func (c *Command) Run(args []string) int {
...
-
-			if err := c.controller.Shutdown(); err != nil {
-				c.UI.Error(fmt.Errorf("Error shutting down controller: %w", err).Error())
-			}
-
-			err := c.opsServer.Shutdown()
-			if err != nil {
-				c.UI.Error(fmt.Errorf("Failed to shutdown ops listeners: %w", err).Error())
-			}
+		case <-c.Context.Done():
```
jefferai, line 922: See comment in server.go

jefferai (commented):
Looking good! Just a couple of small things.

internal/cmd/commands/dev/dev.go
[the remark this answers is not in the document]
jefferai: I was commenting concurrently and I don't think there needs to be another context, it will just complicate things. See my comment...

internal/cmd/commands/server/server.go
```diff
@@ -667,69 +667,100 @@ func (c *Command) StartWorker() error {
...
+				if c.Config.Worker != nil {
+					c.UI.Output("==> Boundary server graceful shutdown triggered, interrupt again to enter shutdown")
+					workerGracefulShutdownFunc()
+				} else {
+					c.UI.Output("==> Boundary server shutdown triggered, interrupt again to force")
+				}
 
-		shutdownTriggered = true
+				if c.Config.Controller != nil {
+					controllerOnce.Do(controllerShutdownFunc)
```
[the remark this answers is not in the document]
jefferai: I don't think this is true? `sync.Once.Do` is does not spawn a goroutine.

internal/cmd/commands/server/server.go
```diff
@@ -667,69 +667,100 @@ func (c *Command) StartWorker() error {
...
+				}
+				shutdownCompleted.Store(true)
+			}()
+
+		case count == 2 && c.Config.Worker != nil:
+			go func() {
+				if c.Config.Worker != nil && !workerShutdownDone.Load() {
```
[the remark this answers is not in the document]
irenarindos: Thanks, I've simplified the shutdown logic!

internal/cmd/commands/dev/dev.go
```diff
@@ -824,74 +824,93 @@ func (c *Command) Run(args []string) int {
...
-	abortForceShutdownCh := make(chan struct{})
-	defer close(abortForceShutdownCh)
+	var workerShutdownOnce sync.Once
+	workerShutdownFunc := func() {
+		if err := c.worker.Shutdown(); err != nil {
+			c.UI.Error(fmt.Errorf("Error shutting down worker: %w", err).Error())
+		}
+	}
+	workerGracefulShutdownFunc := func() {
+		if !c.flagWorkerAuthStorageSkipCleanup && c.worker.WorkerAuthStorage != nil {
```
jefferai, line 837: I would put this storage cleanup in `workerShutdownFunc` after the `Shutdown` call. That way in case something tries to access the files while in the process of shutting down we won't have unexpected behavior.

internal/cmd/commands/server/server.go
```diff
@@ -667,69 +667,84 @@ func (c *Command) StartWorker() error {
...
 func (c *Command) WaitForInterrupt() int {
 	const op = "server.(Command).WaitForInterrupt"
-	// Wait for shutdown
-	shutdownTriggered := false
 
-	// Add a force-shutdown goroutine to consume another interrupt
-	abortForceShutdownCh := make(chan struct{})
-	defer close(abortForceShutdownCh)
+	var shutdownCompleted atm.Bool
+	shutdownTriggerCount := uint32(0)
```
jefferai, line 672: With all of the more recent changes I do not think it's necessary for shutdownTriggerCount to be an atomic any more, because the goroutines are kicked off _after_ reading the value from the main loop.

If you want to keep it atomic, note that you shouldn't just declare a uint32 and run atomic functions on it anymore in Go 1.19; use `atomic.NewUint32` instead.

jefferai (commented):
Two comments, looking good!

internal/cmd/commands/dev/dev.go
irenarindos: Done, thanks!

internal/cmd/commands/server/server.go
irenarindos: Done!

PR #2456, opened 2022-09-14
enos(smoke): Test brokered credential

louisruch:
enos(smoke): Test brokered credential consumption instead of manually
passing private key to boundary connect


louisruch:
@ryancragun @ddebko PTAL
I noticed that we could easily extend the brokered consumption behavior of ssh private keys recently added to Boundary when I had a transient failure on this path.

johanbrandhorst (approved):
This is awesome!

louisruch:
@moduli Might be worth updating the e2e test connect using brokered credential consumption at some point as well

moduli:
> @moduli Might be worth updating the e2e test connect using brokered credential consumption at some point as well

Yup! We're actively working on a new test that uses Vault to provide credentials. We'll also later add a test that uses boundary's built-in credential store (we'll use this as a model).

PR #2459, opened 2022-09-15
scheduler: log details when starting scheduler

louisruch:
We log when the scheduling and monitor loop go down so it makes sense to log when they start. Also since these values are configurable it is good to log:
```
{
  "id": "AguTiArdEX",
  "source": "https://hashicorp.com/boundary/louisruch-C02DF0BSML85/controller+worker",
  "specversion": "1.0",
  "type": "system",
  "data": {
    "version": "v0.1",
    "op": "scheduler.(Scheduler).monitorJobs",
    "data": {
      "interrupt threshold": "5m0s",
      "monitor interval": "30s",
      "msg": "monitor loop running",
      "server id": "dev-controller"
    }
  },
  "time": "2022-09-15T08:29:42.078339-07:00"
}
{
  "id": "LvcWbEfftP",
  "source": "https://hashicorp.com/boundary/louisruch-C02DF0BSML85/controller+worker",
  "specversion": "1.0",
  "type": "system",
  "data": {
    "version": "v0.1",
    "op": "scheduler.(Scheduler).start",
    "data": {
      "msg": "scheduling loop running",
      "run interval": "1m0s",
      "run limit": -1,
      "server id": "dev-controller"
    }
  },
  "time": "2022-09-15T08:29:42.078515-07:00"
}
```

### DIFF
--- a/internal/scheduler/scheduler.go
+++ b/internal/scheduler/scheduler.go
@@ -184,12 +184,17 @@ func (s *Scheduler) RunNow() {
 
 func (s *Scheduler) start(ctx context.Context) {
 	const op = "scheduler.(Scheduler).start"
+	event.WriteSysEvent(ctx, op, "scheduling loop running",
+		"server id", s.serverId,
+		"run interval", s.runJobsInterval.String(),
+		"run limit", s.runJobsLimit)
 	timer := time.NewTimer(0)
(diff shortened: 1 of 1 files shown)


ddebko (approved):
LGTM

PR #2462, opened 2022-09-16
chore(cli): Run bats tests against boundary dev in CI

tmessi:
Several of these tests are being skipped in the enos run. This seems to be
due to differences in the setup that is performed. When these tests were
run manually, `boundary dev` was used which performs some setup of
resources automatically. However since enos uses a non-dev controller, a
similar setup is performed via the enos modules.

Until the discrepancies between the setup can be investigated, this adds a
separate CI job to run the tests against `boundary dev`. These tests are
all passing, without any being skipped. This test suite can still be useful
even if the enos version is fixed, since this test suite will run much
faster, and therefore identify failures sooner.


johanbrandhorst (approved):
LGTM

PR #2472, opened 2022-09-19
update domain, api, and cli to all include key versions in outputs

DanielleMiu:


internal/daemon/controller/handlers/scopes/scope_service.go
```diff
@@ -573,6 +573,13 @@ func sortKeys(keys []*pb.Key) {
 	sort.SliceStable(keys, func(i, j int) bool {
 		return keys[i].GetId() < keys[j].GetId()
 	})
+	// we also want to sort the key versions by version id so they are in
+	// descending order (newest first)
```
johanbrandhorst, line 577: Nit: If we want to sort by newest first, do we not want the sort to be on the `CreateTime` 😛. Should this say "sort based on highest version"?

internal/db/schema/migrations/oss/postgres/47/01_key_rewrap.up.sql
```diff
@@ -17,6 +17,7 @@ create trigger kms_immutable_columns before update on kms_root_key_version
 -- rewrapping the data key version.
 drop trigger kms_immutable_columns on kms_data_key_version;
 
+-- trigger updated again in migration 51 file 03
```
johanbrandhorst, line 20: This is 54 now I think?

internal/db/schema/migrations/oss/postgres/51/03_mutable_root_key_version.up.sql
```diff
@@ -0,0 +1,13 @@
...
+-- https://github.com/hashicorp/go-kms-wrapping/blob/main/extras/kms/migrations/postgres/07_mutable_root_key_version.up.sql
+
+-- we need to make the root_key_version_id mutable in order to support
+-- rewrapping the data key version.
+drop trigger kms_immutable_columns on kms_data_key_version;
+
+create trigger kms_immutable_columns before update on kms_data_key_version
+  for each row execute procedure kms_immutable_columns('private_id', 'data_key_id', 'create_time');
+
+commit;
```
johanbrandhorst, line 13: This file should be moved to 54 I think?

internal/proto/controller/api/resources/scopes/v1/scope.proto
```diff
@@ -86,27 +86,35 @@ message Scope {
   map<string, google.protobuf.ListValue> authorized_collection_actions = 310 [json_name = "authorized_collection_actions"];
 }
 
+// KeyVersion describes a specific version of a key and holds the actual key material
+message KeyVersion {
+  // The ID of the key version (kdkv/krkv).
```
johanbrandhorst, line 91: I don't think we want to add the prefix here, they are subject to change
```suggestion
  // The ID of the key version.
```

johanbrandhorst (changes requested):

internal/daemon/controller/handlers/scopes/scope_service.go
DanielleMiu: what we really want here is "sort based on order in which the version(s) were in use for encryption" but that seems kind of verbose and I think the point come across by just saying newest first

johanbrandhorst (approved):
Nice!

internal/db/schema/migrations/oss/postgres/55/03_mutable_root_key_version.up.sql
```diff
@@ -0,0 +1,13 @@
...
+-- https://github.com/hashicorp/go-kms-wrapping/blob/main/extras/kms/migrations/postgres/07_mutable_root_key_version.up.sql
+
+-- we need to make the root_key_version_id mutable in order to support
+-- rewrapping the data key version.
+drop trigger kms_immutable_columns on kms_data_key_version;
+
+create trigger kms_immutable_columns before update on kms_data_key_version
+  for each row execute procedure kms_immutable_columns('private_id', 'data_key_id', 'create_time');
+
+commit;
```
johanbrandhorst, line 13: This will need to be 55/04 now since I merged the other DB changes as 55/03.

johanbrandhorst (changes requested):
Please rebase on the llb again 😁

johanbrandhorst (approved):

PR #2475, opened 2022-09-20
chore(build): Allow for downloading pre-built admin ui assets

tmessi:
The admin ui assets are built via a github workflow and published as
github a artifact. The artifact can be downloaded, avoiding the need to
build the assets again locally, and eliminating the need for front-end
build tools like yarn and npm.

This updates the `build-ui` target to use a new script that will attempt
to download the pre-built assets. Note that this script requires that
the `gh` cli is installed since an authenticated request is necessary to
download the workflow artifacts. If the artifact cannot be downloaded
either due to the lack of `gh` or because the artifact cannot be found,
it will fallback to using `git clone` (and will require yarn to build
the assets).

See:
    https://cli.github.com/
    https://docs.github.com/en/rest/actions/artifacts
    https://github.com/actions/upload-artifact/issues/51


randallmorey (changes requested):
This is amazing!  Thanks for putting this together.  Should there be a note in the README regarding the UI artefacts and prereqs (namely, `gh`)?

tmessi:
> This is amazing! Thanks for putting this together. Should there be a note in the README regarding the UI artefacts and prereqs (namely, `gh`)?

Good idea, I updated the readme.

randallmorey (approved):

PR #2479, opened 2022-09-21
Add schema for new kms destruction jobs

johanbrandhorst:
The schema lets us manage destruction jobs per-table while providing a guarantee that only one run is running at a time.

### DIFF
--- a/internal/db/schema/migrations/oss/postgres/55/03_create_kms_key_destructions.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/55/03_create_kms_key_destructions.up.sql
@@ -1,0 +1,109 @@
+begin;
+
+  create table kms_data_key_version_destruction_job (
+    key_id kms_private_id primary key -- one job per key version
+      references kms_data_key_version (private_id)
+        on delete cascade -- Note: instead of restrict, we cascade the delete to this job when the key version is deleted
+        on update cascade,
+    create_time wt_timestamp not null
+  );
+  comment on table kms_data_key_version_destruction_job is
+    'Table holding running and pending data key version destruction jobs';
+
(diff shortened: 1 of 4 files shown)


internal/db/schema/migrations/oss/postgres/55/03_create_kms_key_destructions.up.sql
```diff
@@ -0,0 +1,109 @@
...
+    for each row execute procedure immutable_columns('key_id', 'create_time');
+
+
+  create table kms_data_key_version_destruction_job_run (
+    key_id kms_private_id not null
+      references kms_data_key_version_destruction_job (key_id)
+        on delete cascade
+        on update cascade,
+    table_name text not null,
+    total_count int not null
```
mgaffney, line 23: Would it be helpful if a trigger checked (or populated) `total_count`?

mgaffney (changes requested):

internal/db/schema/migrations/oss/postgres/55/03_create_kms_key_destructions.up.sql
johanbrandhorst: I think we should keep doing this in the application, as it lets us avoid adding rows that have no rows that need rewrapping.

johanbrandhorst: If we wanted to check that total count is accurate, we would need to use postgres' eval function to count rows in the table name (since it is a parameter).

mgaffney (approved):

jimlambrt (approved):
Ty!

PR #2486, opened 2022-09-22
Add infrastructure tags to Enos scenarios

jaymalasinha:
Signed-off-by: Jaymala Sinha <jaymala@hashicorp.com>

### DIFF
--- a/enos/enos-variables.hcl
+++ b/enos/enos-variables.hcl
@@ -52,6 +52,18 @@ variable "alb_listener_api_port" {
   default     = 9200
 }
 
+variable "project_name" {
+  description = "The description of the project"
+  type        = string
+  default     = "boundary-enos-integration"
+}
+
+variable "tags" {
+  description = "Tags that will be applied to infrastructure resources that support tagging"
+  type        = map(string)
+  default     = null
+}
+
 variable "target_instance_type" {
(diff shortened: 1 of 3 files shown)


enos/enos-variables.hcl
```diff
@@ -52,6 +52,18 @@ variable "alb_listener_api_port" {
...
+variable "project_name" {
+  description = "The description of the project"
+  type        = string
+  default     = "boundary-enos-integration"
+}
+
+variable "tags" {
+  description = "Tags that will be applied to infrastructure resources that support tagging"
+  type        = map(string)
+  default     = null
```
mikebaum, line 64: I wonder if this default value should rather be: `{}`. Not sure if merge function works well with a `null` argument for any of its args. Maybe you could just double check that it works with null.

mikebaum (commented):
Just one question about the default value for the tags variable, otherwise everything looks good.

enos/enos-variables.hcl
jaymalasinha: it does seem to work with default `null`, as we have the same setup in Vault scenarios.

mikebaum (approved):
Looks good to me.

joshbrand (approved):

PR #2492, opened 2022-09-23
fix(genapi): uint64 fields failing to be unmarshalled

hugoamvieira:
uint64 fields get marshalled into json by protobuf as strings, so we have to tell the json parser to look for strings when unmarshalling.


tmessi (approved):

jefferai:
Can you explain the issue a bit more?

hugoamvieira:
@jefferai Sure! With the work to have bytes up and down actually work, we'll now have that data in a Session Read response (Inside the `Connection` object [here](https://github.com/hashicorp/boundary/blob/main/api/sessions/connection.gen.go#L9-L10))

When we receive the json payload, eventually we'll try to unmarshal the `"connections"` json array into a `[]Connection` object. Right now, for `BytesUp` and `BytesDown`, the unmarshaller expects to see some sort of number type to match the `uint64` (not sure which one actually), however we use protobuf, and it [marshals `uint64` fields into json as strings](https://developers.google.com/protocol-buffers/docs/proto3#json) so there's a type mismatch for those fields.

Everything works now because all these fields are `omitempty` and because we don't actually have values, they're always zero, so they're never sent, but when we actually start sending data, we'd expect to see a json unmarshal error.

johanbrandhorst (approved):
LGTM! My only concern is whether we actually need to make this this powerful - could a "marshalAsString" option or something be less powerful and solve the same problem? What if someone specifies an invalid JSON tag with this option? We won't find out until runtime, right?

hugoamvieira:
@johanbrandhorst Well it'd have to pass code review, but yeah I think the fields within the json tag aren't evaluated until they need to be, so unless it breaks Go's tag syntax, we'd only find out at runtime. At the moment, we only need that `string`, so perhaps it'd be better to have a more specific method to avoid any user error.

The *slight* downside there is if we need this for any other purpose in the future we'd have to add that in.

WDYT @tmessi ?

PR #2495, opened 2022-09-26
use prior and current worker auth keys

irenarindos:


mgaffney (approved):

internal/db/schema/migrations/oss/postgres/55/01_worker_auth_create_time.up.sql
```diff
@@ -0,0 +1,138 @@
...
+      references worker_auth_authorized_state_enm(state)
+      on delete restrict
+      on update cascade,
+  add constraint worker_auth_authorized_worker_id_state_uq
+    unique(worker_id, state)
+;
+
+-- The worker_auth_authorized table may contain multiple rows for the same
+-- worker_id. For each worker_id in the table, we need to pick one row and set
+-- it's state to 'current'. There is no way to be sure that we will pick the
```
jefferai, line 56:
```suggestion
-- its state to 'current'. There is no way to be sure that we will pick the
```

internal/server/repository_workerauth.go
```diff
@@ -879,22 +919,55 @@ func decrypt(ctx context.Context, value []byte, wrapper wrapping.Wrapper) ([]byt
...
+	var currentWorkerAuth *WorkerAuth
+
+	var workerAuths []*WorkerAuth
+	if err := r.reader.SearchWhere(ctx, &workerAuths, "worker_id = ?", []interface{}{workerId}); err != nil {
 		return nil, errors.Wrap(ctx, err, op)
 	}
 
-	return worker, nil
+	workerAuthsFound := len(workerAuths)
+	if workerAuthsFound == 0 {
```
jefferai, line 938: This would be cleaner as a `switch` function on `workerAuthsFound`, with the `default` being the >2 case.

jefferai (approved):
One minor cleanup suggestion, looks good otherwise!

internal/server/repository_workerauth.go
irenarindos: thanks!

PR #2497, opened 2022-09-26
cli(ssh): Use type instead of SubtypeFromId

louisruch:
The OSS Boundary CLI cannot get the SubtypeFromId for non tcp targets


ddebko (approved):
LGTM

PR #2474, opened 2022-09-19
Add vault test to e2e test suite

moduli:
This PR adds a new enos scenario and two new tests that use Vault as a credential store. 
- The first test uses the boundary and vault clis to set up vault as a credential store, create a set of credentials, attach those credentials to a target, and connect to that target using what was provided by boundary/vault.
- The second test is similar but uses the bounday go api

As a design philosophy, we wanted to minimize the amount of setup done in enos in order to maintain as much flexibility as possible with regards to how the infrastructure gets created.

We opted not to use the vault go api for the following reasons
- the vault cli is more readable than the vault api
- it enables flexibility in using the corresponding vault cli version with the vault server version
- boundary itself has a dependency on the vault api

Additionally, some refactoring was done to create methods for commonly used boundary operations.

```
❯ enos scenario run e2e_vault builder:local
Scenario: e2e_vault [builder:local]
  Generate: ✅
  Init: ✅
  Validate: ✅
  Plan: ✅
  Apply: ✅
  Destroy: ✅
```
```
❯ go test -v github.com/hashicorp/boundary/testing/internal/e2e/credential/vault
=== RUN   TestCreateVaultCredentialStoreCli
    vault_test.go:103: Created Vault Credential
    vault_test.go:118: Created Vault Cred Store Token
    vault_test.go:125: Created Org Id: o_ySVmbNaxxr
    vault_test.go:129: Created Project Id: p_Wzs62xiIHR
    vault_test.go:143: Created Credential Store: csvlt_mVToVFzxYy
    vault_test.go:158: Created Credential Library: clvlt_srq6V83oG8
    vault_test.go:171: Created Host Catalog: hcst_nVsO8t6jxX
    vault_test.go:184: Created Host Set: hsst_SGDm2eM1eU
    vault_test.go:198: Created Host: hst_AgdajjZYxo
    vault_test.go:219: Created Target: ttcp_w2srYLuknu
    vault_test.go:250: Successfully retrieved credentials for target
    vault_test.go:279: Successfully connected to target
--- PASS: TestCreateVaultCredentialStoreCli (6.15s)
=== RUN   TestCreateVaultCredentialStoreApi
    vault_test.go:314: Created Vault Credential
    vault_test.go:329: Created Vault Cred Store Token
    vault_test.go:345: Created Org Id: o_GMdO5I3DaO
    vault_test.go:351: Created Project Id: p_ZAATn1fCNt
    vault_test.go:361: Created Credential Store: csvlt_zVLSIqdjwA
[...]

PR #2498, opened 2022-09-27
Add ability to list key version destruction jobs, recurring jobs and destroying key versions

johanbrandhorst:
## [feat(kms): Add ability to list data key version destruction jobs](https://github.com/hashicorp/boundary/commit/b1f2e2f2683cb0891489934e7c53966df47e2fe4)

Adds the domain part of this new method

## [feat(scopes): Add ListKeyVersionDestructionJobs](https://github.com/hashicorp/boundary/commit/5c8e5f2a7d08014f1724058562ca64aa3afb2602)

Adds the API part of this new method

## [feat(cli): Add list-key-version-destruction-jobs](https://github.com/hashicorp/boundary/commit/7f4ffa69069811b67f9503416ea8d8bfcdf8a782)

Adds the CLI part of this new method

## [feat(kms): Add recurring job that performs table rewrapping](https://github.com/hashicorp/boundary/commit/c4f7fb875b0201276c9de6f313b3bd43470a6ee9)

The new recurring job runs for each job table with a registered
rewrapping callback. It attempts to become the running run and
start its rewrapping, gracefully handling sudden interruptions
by resuming it's work if it finds evidence of sudden interruption.

## [feat(kms): Add recurring job for destroying key versions](https://github.com/hashicorp/boundary/commit/9d8eada4bdd05c0091eb32399a727757d659b853)

The new job monitors the database for data key version destruction
jobs that have finished rewrapping all its data, performing
the final data key version revocation.

## [feat(kms): Add DestroyKeyVersion](https://github.com/hashicorp/boundary/pull/2498/commits/31f0e8d576d83c4887fe7f2dcb39842fde3a0dd0)

DestroyKeyVersion immediately destroys a key version
if it is a root key version or a data key version
which encrypts no data. If the data key version
encrypts data which needs to be rewrapped, it
queues an asynchronous job to complete the rewrapping
operation.

## [feat(cli): Add scopes destroy-key-version CLI command](https://github.com/hashicorp/boundary/pull/2498/commits/ce458b2c27f9b68a933c2f3d557b8c3a798878fc)


internal/cmd/commands/scopescmd/list_key_version_destruction_jobs.go
```diff
@@ -0,0 +1,156 @@
...
+			output = append(output,
+				fmt.Sprintf("    Rows completed:  %d", item.CompletedCount),
+			)
+		}
+		if true {
+			output = append(output,
+				fmt.Sprintf("    Total rows:      %d", item.TotalCount),
+			)
+		}
+	}
```
johanbrandhorst, line 102: This is a proposal for how to display these on the CLI. This is what it looks like on the CLI:

```
$ boundary scopes list-key-version-destruction-jobs -scope-id global

Key version destruction job information:
  Key version ID:    kdkv_OJz1sJIgbZ
    Scope ID:        global
    Started:         Mon, 26 Sep 2022 17:23:07 PDT
    Status:          pending
    Rows completed:  0
    Total rows:      300
```

internal/api/genapi/input.go
```diff
@@ -167,6 +167,18 @@ var inputStructs = []*structInfo{
...
+		inProto:     &scopes.KeyVersionDestructionJob{},
+		outFile:     "scopes/key_version_destruction_job.gen.go",
+		skipOptions: true,
+		fieldOverrides: []fieldInfo{
+			// int64 fields get marshalled by protobuf as strings, so we have
+			// to tell the json parser that their json representation is a
+			// string but they go into Go int64 types.
+			{Name: "CompletedCount", JsonTags: []string{"string"}},
+			{Name: "TotalCount", JsonTags: []string{"string"}},
+		},
```
johanbrandhorst, line 180: Note that this is taking advantage of a feature of the CLI generation that will likely change soon - it was merged for 0.11.0 to get support for stringified large integers into 0.11.0 but longer term this will likely be implicit behavior when using large integers in the protobuf type. CC @tmessi @hugoamvieira

internal/cmd/commands/scopescmd/list_key_version_destruction_jobs.go
```diff
@@ -0,0 +1,156 @@
...
+	return wordwrap.WrapString("List all pending key version destruction jobs within a scope", base.TermWidth)
+}
+
+func (c *ListKeyVersionDestructionJobsCommand) Help() string {
+	return base.WrapForHelpText([]string{
+		"Usage: boundary scopes list-key-version-destruction-jobs [args]",
+		"",
+		"  List all pending key version destruction jobs within a scope. A key version",
+		"  destruction job asynchronously re-encrypts any existing data with the latest",
+		"  key version before destroying the key version. Example:",
```
johanbrandhorst, line 35: I don't know if this is the right place to explain this sort of thing, but I figured it couldn't hurt, right?

internal/daemon/controller/handlers/scopes/scope_service_test.go
```diff
@@ -1950,3 +1956,226 @@ func TestRotateKeys(t *testing.T) {
...
+		_, err = sqldb.ExecContext(context.Background(), "insert into kms_data_key_version_destruction_job_run(key_id, table_name, total_count) values ($1, 'auth_token', 100)", kvToDestroy.Id)
+		require.NoError(t, err)
+		_, err = sqldb.ExecContext(context.Background(), "insert into kms_data_key_version_destruction_job_run(key_id, table_name, total_count) values ($1, 'auth_oidc_method', 200)", kvToDestroy.Id)
+		require.NoError(t, err)
+		_, err = sqldb.ExecContext(context.Background(), "update kms_data_key_version_destruction_job_run set completed_count=100 where key_id=$1 and table_name='auth_token'", kvToDestroy.Id)
+		require.NoError(t, err)
+		t.Cleanup(func() {
+			_, err = sqldb.ExecContext(context.Background(), "delete from kms_data_key_version_destruction_job where key_id=$1", kvToDestroy.Id)
+			require.NoError(t, err)
+		})
```
johanbrandhorst, line 2052: It's a layer violation to have the scope service test reach right into the database like this, but we don't have much choice until the real DestroyKeyVersion is available.

internal/proto/controller/storage/kms/store/v1/data_key_version_destruction_job_progress.proto
```diff
@@ -0,0 +1,34 @@
...
+  // @inject_tag: `gorm:"not_null"`
+  string status = 50;
+
+  // The number of rows that have been rewrapped
+  // @inject_tag: `gorm:"not_null"`
+  int64 completed_count = 60;
+
+  // The total number of rows that need rewrapping
+  // @inject_tag: `gorm:"not_null"`
+  int64 total_count = 70;
```
johanbrandhorst, line 33: The use of `bigint` in our database schema necessitates this large integer

internal/tests/cli/boundary/scopes.bats
```diff
@@ -24,6 +24,12 @@ export NEW_PROJECT='test_project'
 	[ "$status" -eq 0 ]
 }
 
+@test "boundary/scopes: can list key version destruction jobs at global scope" {
+	run list_key_version_destruction_jobs $DEFAULT_GLOBAL
+  echo "$output"
+	[ "$status" -eq 0 ]
+}
```
johanbrandhorst, line 31: These tests will become more interesting once we can request a key destruction

testing/internal/e2e/scopes/scope_test.go
```diff
@@ -0,0 +1,49 @@
...
+	client, err := boundary.NewApiClient()
+	require.NoError(t, err)
+
+	ctx := context.Background()
+
+	scopeClient := scopes.NewClient(client)
+	keyVersionDestructionJobListResult, err := scopeClient.ListKeyVersionDestructionJobs(ctx, "global")
+	require.NoError(t, err)
+	assert.Empty(t, keyVersionDestructionJobListResult.Items)
+}
```
johanbrandhorst, line 49: CC @moduli let me know if its premature to add tests to this framework.

moduli: Nope! I think it should be ok. Two things...
1. We'll need to add an enos scenario for this. Based on what I see here, it looks like it will just be a duplicate of `enos-scenario-e2e-target.hcl`, but you can get rid of the `create_target` step and update `test_package` to point to this
2. There's a slight change to `RunCommand` coming in https://github.com/hashicorp/boundary/pull/2474 that will make it look something like this
```
output = e2e.RunCommand("boundary", e2e.WithArgs("scopes", "list-key-version-destruction-jobs",
		"-scope-id", "global",
		"-format", "json",
	))
```

moduli: Also, maybe nit, but I might suggest changing the directory name from `scopes` to `scope`

testing/internal/e2e/scopes/scope_test.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/hashicorp/boundary/api/scopes"
+	"github.com/hashicorp/boundary/testing/internal/e2e"
+	"github.com/hashicorp/boundary/testing/internal/e2e/boundary"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestListKeyVersionDestructionJobsCli uses the boundary CLI to list
+// pending key version destruction jobs.
+func TestListKeyVersionDestructionJobsCli(t *testing.T) {
```
tmessi, line 17: Is there a workflow that can be tested here? Generally I think these tests shouldn't just test a single cli command, you already added tests for that in the cli (bats) tests.

johanbrandhorst: There will be a workflow once we have the ability to trigger a key destruction. I could remove these tests until we can do them all together. I'm also not sure what the overlap between bats and these are, IMO we could replace the bats tests with this new framework.

tmessi: My understanding of the differences between the tests is that the bats tests are for testing the cli ui, so testing individual commands and getting expected outputs, while theses tests are for testing end-to-end admin workflows that will often involve multiple commands, and could be accomplished via the cli/api/admin ui.

I agree that the cli ui tests *could* be implemented using some of this code, instead of using bats, but I don't think there are any current plans to do so.

testing/internal/e2e/scopes/scope_test.go
johanbrandhorst: I removed these for now, will readd once we have key destruction commands.

internal/daemon/controller/handlers/scopes/scope_service_test.go
DanielleMiu: this will probably be added later, but wanted to call out we don't test completed (or running) jobs here

DanielleMiu (approved):

api/scopes/custom.go
```diff
@@ -105,3 +118,43 @@ func (c *Client) RotateKeys(ctx context.Context, scopeId string, rewrapKeys bool
 	}
 	return target, nil
 }
+
+func (c *Client) ListKeyVersionDestructionJobs(ctx context.Context, scopeId string, opt ...Option) (*KeyVersionDestructionJobListResult, error) {
+	if scopeId == "" {
+		return nil, fmt.Errorf("empty scopeId value passed into ListKeyVersionDestructionJobs request")
+	}
```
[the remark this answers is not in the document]
johanbrandhorst: I think in the CLI we consistently error when one isn't set while on the API we consistently infer global if one isn't set 😁.

testing/internal/e2e/scope/scope_test.go
```diff
@@ -0,0 +1,231 @@
+package scope_test
```
johanbrandhorst, line 1: Both these tests still need to be rewritten to be independent of previous runs. Will probably run in a temporary scope instead of "global".

internal/db/schema/migrations/oss/postgres/56/03_create_kms_key_destructions.up.sql
```diff
@@ -18,7 +18,8 @@ begin;
     key_id kms_private_id not null
       references kms_data_key_version_destruction_job (key_id)
         on delete cascade
-        on update cascade,
+        on update cascade
+        deferrable initially deferred,
```
johanbrandhorst, line 22: This lets us try to create the job_runs before the job, which means we don't have to create the job and then delete it if there were no runs.

internal/db/sqltest/tests/kms/kms_data_key_destruction_job.sql
```diff
@@ -26,14 +26,6 @@ begin;
       ('kdkv_______colors');
   select lives_ok('insert_new_key_id', 'insert of valid key_id in kms_data_key_version_destruction_job failed');
 
-  -- Should fail when inserting an unknown key_id
-  prepare run_insert_unknown_key_id as
-    insert into kms_data_key_version_destruction_job_run
-      (key_id, table_name, total_count)
-    values
-      ('kdkv___unknown', 'auth_token', 100);
-  select throws_ok('run_insert_unknown_key_id', '23503', null, 'insert of unknown key_id in kms_data_key_version_destruction_job_run succeeded');
```
johanbrandhorst, line 35: It wasn't possible to trigger this _inside_ of a transaction anymore as the consistency check is done at transaction commit time. I couldn't find an easy way to run this test at commit time so I removed it.

internal/kms/job/job.go
```diff
@@ -0,0 +1,33 @@
...
+	const op = "kmsjob.RegisterJobs"
+
+	dataKeyVersionDestructionMonitorJob, err := newDataKeyVersionDestructionMonitorJob(ctx, kmsRepo)
+	if err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+	if err := s.RegisterJob(ctx, dataKeyVersionDestructionMonitorJob); err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+	for _, tableName := range kms.ListTablesSupportingRewrap() {
```
johanbrandhorst, line 22: Note: this lists the tables that are registered via `init` functions, not all tables referencing the key_id column. What guarantees that those lists are the same is a test that runs both and compares the results, which is not part of this PR.

testing/internal/e2e/scope/scope_test.go
```diff
@@ -0,0 +1,322 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestKeyDestructionCli uses the boundary CLI to test key destruction.
+func TestKeyDestructionCli(t *testing.T) {
+	e2e.MaybeSkipTest(t)
+
+	output := boundary.AuthenticateCli()
+	require.NoError(t, output.Err, string(output.Stderr))
```
moduli, line 22: When you rebase with main, you would need to change this to...
```suggestion
        boundary.AuthenticateCli(t)
```

testing/internal/e2e/scope/scope_test.go
```diff
@@ -0,0 +1,322 @@
+package scope_test
```
moduli, line 1: We'll have to think about file organization here. For now, I might suggest to add `scope_test.go` to the `static_test` package in `e2e/host/static` (you might not see this until you rebase with main). That way, it will automatically get run in the existing enos scenario. The alternative is to add another enos scenario, which would add some additional overhead in CI.

testing/internal/e2e/scope/scope_test.go
```diff
@@ -0,0 +1,322 @@
...
+	var scopeReply scopes.ScopeCreateResult
+	require.NoError(t, json.Unmarshal(output.Stdout, &scopeReply))
+	t.Cleanup(func() {
+		output = e2e.RunCommand(
+			"boundary", "scopes", "delete",
+			"-id", scopeReply.Item.Id,
+			"-format", "json",
+		)
+		require.NoError(t, output.Err, string(output.Stderr))
+	})
```
moduli, line 40: When you rebase with main, you could change this to
```suggestion
	scopeId := boundary.CreateNewOrgCli(t)
```

testing/internal/e2e/scope/scope_test.go
```diff
@@ -0,0 +1,322 @@
...
+	output = e2e.RunCommand(
+		"boundary", "scopes", "list-keys",
+		"-scope-id", scopeReply.Item.Id,
+		"-format", "json",
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+	var keys scopes.KeyListResult
+	err := json.Unmarshal(output.Stdout, &keys)
+	require.NoError(t, err)
+	assert.Len(t, keys.Items, 7)
```
moduli, line 51: What does the `7` represent here? Why do we expect 7 items?

testing/internal/e2e/scope/scope_test.go
```diff
@@ -0,0 +1,322 @@
...
+		"-format", "json",
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+	var amReply authmethods.AuthMethodCreateResult
+	require.NoError(t, json.Unmarshal(output.Stdout, &amReply))
+	t.Cleanup(func() {
+		output = e2e.RunCommand(
+			"boundary", "auth-methods", "delete",
+			"-id", amReply.Item.Id,
+		)
```
moduli, line 72: nit: If you wanted to, you could change this to one line since it's shorter. I'm not sure what the preferred styling is here.
```suggestion
		output = e2e.RunCommand("boundary", "auth-methods", "delete", "-id", amReply.Item.Id)
```

testing/internal/e2e/scope/scope_test.go
```diff
@@ -0,0 +1,322 @@
...
+		"-scope-id", scopeReply.Item.Id,
+		"-format", "json",
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+	var keys scopes.KeyListResult
+	err := json.Unmarshal(output.Stdout, &keys)
+	require.NoError(t, err)
+	assert.Len(t, keys.Items, 7)
+	for _, key := range keys.Items {
+		assert.Len(t, key.Versions, 1)
```
moduli, line 53: Do we care what the version value is here? Is there a default value?

testing/internal/e2e/scope/scope_test.go
moduli: Separate question, if we're deleting the scope/org in cleanup, do we need to explicitly clean up other things in that scope/org?

testing/internal/e2e/scope/scope_test.go
```diff
@@ -0,0 +1,322 @@
...
+		require.NoError(t, output.Err, string(output.Stderr))
+	})
+
+	output = e2e.RunCommand(
+		"boundary", "scopes", "rotate-keys",
+		"-scope-id", scopeReply.Item.Id,
+		"-format", "json",
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+	var rotate scopes.KeysRotateResult
```
moduli, line 82: Do we use `rotate` anywhere or is this simply just to ensure that the output from the command can be appropriately parsed/unmarshaled?

testing/internal/e2e/scope/scope_test.go
```diff
@@ -0,0 +1,322 @@
...
+
+	ctx := context.Background()
+	sc := scopes.NewClient(client)
+
+	scope, err := sc.Create(ctx, "global", scopes.WithName("testscope"))
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		_, err := sc.Delete(context.Background(), scope.Item.Id)
+		require.NoError(t, err)
+	})
```
moduli, line 220: When you rebase with main, this could be changed to
```suggestion
	scopeId := boundary.CreateNewOrgApi(t, ctx, client)
```

testing/internal/e2e/scope/scope_test.go
```diff
@@ -0,0 +1,322 @@
...
+
+	keys, err := sc.ListKeys(ctx, scope.Item.Id)
+	require.NoError(t, err)
+	assert.Len(t, keys.Items, 7)
+	// Assert that all keys have the same number of versions
+	numVersions := len(keys.Items[0].Versions)
```
moduli, line 226: Just pointing out that this was done a little differently in the cli test (it was hard-coding numbers). I assume now that the tests are in different scopes, you can still assume there should only be 1 version here?

internal/kms/kms.go
```diff
@@ -269,6 +273,245 @@ func (k *Kms) RotateKeys(ctx context.Context, scopeId string, opt ...Option) err
...
+			}
+			defer rows.Close()
+			var numRows uint
+			for rows.Next() {
+				if err := r.ScanRows(ctx, rows, &numRows); err != nil {
+					errors.Wrap(ctx, err, op, errors.WithMsg("failed to scan number of rows for %q", table.GetTableName()))
+				}
+			}
+			if numRows == 0 {
+				// No rows to rewrap 🎉
```
DanielleMiu, line 482: lol love this

internal/kms/kms.go
```diff
@@ -269,6 +273,245 @@ func (k *Kms) RotateKeys(ctx context.Context, scopeId string, opt ...Option) err
...
+			var cancel context.CancelFunc
+			ctx, cancel = context.WithTimeout(context.Background(), 5*time.Second)
+			defer cancel()
+		}
+		// Update the progress of this run
+		args := []interface{}{
+			sql.Named("key_id", run.KeyId),
+			sql.Named("table_name", tableName),
+		}
+		if _, err := k.writer.Exec(ctx, fmt.Sprintf(updateCompletedCountQueryTemplate, tableName), args); err != nil {
```
DanielleMiu, line 355: so this query will updated the completed count to an accurate figure even if a previous run was killed and unable to update the completed count. however, this will leave an inaccurate completed count throughout the duration of this run's execution. I assume that's known + expected + and we don't really care since we only update after a run finishes anyway?

internal/tests/api/scopes/scope_test.go
```diff
@@ -152,3 +155,102 @@ func TestErrors(t *testing.T) {
...
+	assert.Equal(t, "pending", result.State)
+
+	jobs, err = sc.ListKeyVersionDestructionJobs(ctx, "global")
+	require.NoError(t, err)
+	assert.Len(t, jobs.Items, 1)
+
+	// The configured scheduler monitoring interval is 1 second. The jobs become available 1
+	// second after the last successful run. We need to re-encrypt data in 4 different tables,
+	// and then we need to destroy the key. This job will take between 4 and 5 seconds to run,
+	// depending on the timing of the first started run.
```
DanielleMiu, line 228: 1. only 4 tables?
2. is there any data to rewrap in tests? I didn't think that the test setup inserted any default data into the tables, and I didn't see you insert any in the test itself

internal/tests/api/scopes/scope_test.go
```diff
@@ -152,3 +155,102 @@ func TestErrors(t *testing.T) {
...
+		}
+	}
+
+	result, err = sc.DestroyKeyVersion(ctx, "global", destroyKeyVersion.Id)
+	require.NoError(t, err)
+	assert.Equal(t, "pending", result.State)
+
+	jobs, err = sc.ListKeyVersionDestructionJobs(ctx, "global")
+	require.NoError(t, err)
+	assert.Len(t, jobs.Items, 1)
```
DanielleMiu, line 223: if you didn't insert any data for rewrapping, wouldn't destroy return a completed status? per [this logic from destroy key version](https://github.com/hashicorp/boundary/pull/2498/files#diff-65fa95abdd78accdc6b1f3ac98751fe4700c8abc1ee696b8d6ae7e5403af0831R511-R517)?

testing/internal/e2e/scope/scope_test.go
johanbrandhorst: We have 7 different key purposes in each scope, so we get 7 keys per scope.

testing/internal/e2e/scope/scope_test.go
johanbrandhorst: I like to be explicit about it, even though it probably doesn't need to be done 😄

testing/internal/e2e/scope/scope_test.go
johanbrandhorst: The version is a struct which has ID, CreateTime on it, so it's not easy to compare against something.

DanielleMiu (changes requested):

testing/internal/e2e/scope/scope_test.go
johanbrandhorst: Hm, you're right, we can probably just remove this

testing/internal/e2e/scope/scope_test.go
johanbrandhorst: Good shout, I missed changing this back

internal/daemon/controller/handlers/scopes/scope_service.go
```diff
@@ -432,6 +434,76 @@ func (s Service) RotateKeys(ctx context.Context, req *pbs.RotateKeysRequest) (*p
...
+	if err != nil {
+		if errors.Match(errors.T(errors.KeyNotFound), err) {
+			return nil, handlers.NotFoundErrorf("unknown key_version_id %q", req.KeyVersionId)
+		}
+		return nil, err
+	}
+	state := "completed"
+	if !destroyed {
+		state = "pending"
+	}
```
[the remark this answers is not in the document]
johanbrandhorst: I don't think there's much reason to introduce an enum for something as simple as this. It could be a boolean, but this was the API definition we agreed on, which I think is nicer than a boolean and certainly more flexible.

internal/tests/api/scopes/scope_test.go
johanbrandhorst: Yep! There's 8 pieces of encrypted data created by the test controller setup (creating some tokens, users, auth methods, etc) spread across 4 tables in the global scope.

internal/tests/api/scopes/scope_test.go
johanbrandhorst: It would, yep! But since there's default data it doesn't.

internal/kms/kms.go
johanbrandhorst: Indeed, we don't do any mid-run updates, only at the end of the run. Since we break up the rewrapping per-table I think that's still okay as the user will get updates every time a table finishes. We could implement mid-run updates in a future update.

internal/kms/kms.go
```diff
@@ -269,6 +273,245 @@ func (k *Kms) RotateKeys(ctx context.Context, scopeId string, opt ...Option) err
...
+func (k *Kms) ListDataKeyVersionDestructionJobs(ctx context.Context, scopeId string, _ ...Option) ([]*DataKeyVersionDestructionJobProgress, error) {
+	var jobs []*DataKeyVersionDestructionJobProgress
+	if err := k.reader.SearchWhere(ctx, &jobs, "scope_id=?", []interface{}{scopeId}); err != nil {
+		return nil, err
+	}
+	return jobs, nil
+}
+
+// MonitorTableRewrappingRuns checks for pending rewrapping runs for the
+// specified table name, and attempts to become the running run and
```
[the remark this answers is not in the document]
johanbrandhorst: The terminology is a bit of a struggle 😅. I'll update it.

internal/kms/kms.go
```diff
@@ -269,6 +273,245 @@ func (k *Kms) RotateKeys(ctx context.Context, scopeId string, opt ...Option) err
...
+	run := allocDataKeyVersionDestructionJobRun()
+	err := k.reader.LookupWhere(ctx, &run, "is_running=true and table_name!=?", []interface{}{tableName})
+	switch {
+	case err == nil:
+		// Another run was already running
+		return nil
+	case errors.Match(errors.T(errors.RecordNotFound), err):
+		// Great, no running run, lets continue
+	default:
+		return errors.Wrap(ctx, err, op, errors.WithMsg("failed to find any running runs"))
```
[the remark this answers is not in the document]
johanbrandhorst: Nah, this is for an unexpected error like no database connection or something like that.

testing/internal/e2e/scope/scope_test.go
johanbrandhorst: I'll move it

internal/db/sqltest/tests/kms/kms_data_key_destruction_job.sql
johanbrandhorst: No, since it's a direct database schema test, we don't really have another place to test this, and frankly I'm not sure it's necessary.

internal/db/schema/migrations/oss/postgres/56/03_create_kms_key_destructions.up.sql
johanbrandhorst: It happens in the same transaction as the job runs in `DestroyKeyVersion`: https://github.com/hashicorp/boundary/pull/2498/files#diff-65fa95abdd78accdc6b1f3ac98751fe4700c8abc1ee696b8d6ae7e5403af0831R504. What do you mean that they will be deleted? It will error at commit and rollback if these constraints are not met, it won't delete anything silently.

internal/daemon/controller/handlers/scopes/scope_service.go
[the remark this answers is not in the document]
johanbrandhorst: I assume by enum you mean something like
```go
type KeyDestructionState int

const (
	KeyDestructionStateUnspecified KeyDestructionState = iota
	KeyDestructionStatePending
	KeyDestructionStateCompleted
)

func (k KeyDestructionState) String() string {
	switch k {
	case KeyDestructionStatePending:
		return "pending"
	case KeyDestructionStateCompleted:
		return "completed"
	default:
		return "unspecified"
	}
}
```

It's quite a bit of overhead for little gain here, I think, and it actually introduces the issue of using the zero value of the type by mistake.

internal/kms/kms.go
```diff
@@ -269,6 +273,254 @@ func (k *Kms) RotateKeys(ctx context.Context, scopeId string, opt ...Option) err
...
+		for _, table := range tables {
+			rows, err := r.Query(ctx, fmt.Sprintf(findAffectedRowsForKeyQueryTemplate, table.GetTableName()), []interface{}{keyVersionId})
+			if err != nil {
+				errors.Wrap(ctx, err, op, errors.WithMsg("failed to get affected rows for %q", table.GetTableName()))
+			}
+			defer rows.Close()
+			var numRows uint
+			for rows.Next() {
+				if err := r.ScanRows(ctx, rows, &numRows); err != nil {
+					errors.Wrap(ctx, err, op, errors.WithMsg("failed to scan number of rows for %q", table.GetTableName()))
```
DanielleMiu, line 480: these two errors are not returned?

johanbrandhorst: Hm, oops!

PR #2501, opened 2022-09-27
feat(worker): net.Conn implementation to count Bytes Read and Written

hugoamvieira:
This commit introduces a new net.Conn implementation (countingConn) that will keep track of bytes read and written for a particular client connection. This is then implemented into the worker code using interface composition.


johanbrandhorst (approved):
_nice_

PR #2502, opened 2022-09-27
Extend enos-run.yml to cover E2E testing

manthonygfp:
This PR is related to adding a GitHub Action workflow that executes the Enos End to End scenario.

We have chosen to add on to the existing Enos Run workflow that also runs integration tests. In the process, we have inverted some logic that previously checked the affirmative case of "a" and now checks the negative case of "b". This was necessary to add more than 2 types of filters to the workflow.

Additionally, another deviation from previous implementations is that we will only run the Destroy step if either the Run or its retry step have failed. This is because `enos scenario run` will execute a destroy action if it completes its apply action successfully.

Other than that, the only other change is that we have added a new filter to the existing scenario in order to execute this new scenario.

### DIFF
--- a/.github/workflows/enos-run.yml
+++ b/.github/workflows/enos-run.yml
@@ -25,8 +25,11 @@ jobs:
       fail-fast: false # don't fail as that can skip required cleanup steps for jobs
       matrix:
         include:
-          - test: smoke
-          - test: cli_ui
+          - filter: 'integration test:smoke'
+          - filter: 'integration test:cli_ui'
+          - filter: 'e2e_credential_vault'
+          - filter: 'e2e_host_aws'
+          - filter: 'e2e_host_static'
     runs-on: ubuntu-latest
     env:
       GITHUB_TOKEN: ${{ secrets.SERVICE_USER_GITHUB_TOKEN }}
@@ -44,14 +47,12 @@ jobs:
           # it changes the output to text when we expect it to be JSON.
           terraform_wrapper: false
       - name: Import GPG key for Boundary pass keystore
-        if: matrix.test == 'cli_ui'
         id: import_gpg
         uses: crazy-max/ghaction-import-gpg@v5
         with:
           gpg_private_key: ${{ secrets.ENOS_GPG_PRIVATE_KEY }}
           passphrase: ${{ secrets.ENOS_GPG_PASSPHRASE }}
       - name: Trust the pass keystore GPG key
-        if: matrix.test == 'cli_ui'
         id: trust_gpg
         run: |
           gpg -a --encrypt -r ${{ secrets.ENOS_GPG_UID }} --trust-model always
@@ -75,47 +76,57 @@ jobs:
           mkdir -p ./enos/support
           echo "${{ secrets.ENOS_CI_SSH_KEY }}" > ./enos/support/private_key.pem
           chmod 600 ./enos/support/private_key.pem
-      - name: Set up Bats CLI UI tests dependency cache
-        if: matrix.test == 'cli_ui'
+      - name: Set up dependency cache
         id: dep-cache
         uses: actions/cache@v3
         with:
-          path: /tmp/bats-cli-ui-deps
-          key: enos-bats-cli-ui-deps-jq-1.6-password-store-1.7.4
+          path: /tmp/test-deps
+          key: enos-test-deps-jq-1.6-password-store-1.7.4-vault-1.11.4
+      - name: Debug dep-cache
+        run: |
+          mkdir -p /tmp/test-deps
+          ls -la /tmp/test-deps
       - name: Set up Node for Bats install
-        if: matrix.test == 'cli_ui'
+        if: matrix.filter == 'integration test:cli_ui'
         uses: actions/setup-node@v3
         with:
           node-version: 16
       - name: Install Bats via NPM
-        if: matrix.test == 'cli_ui'
+        if: matrix.filter == 'integration test:cli_ui'
         # Use npm so this workflow is portable on multiple runner distros
         run: npm install --location=global bats
       - name: Download jq for Bats CLI UI tests
-        if: matrix.test == 'cli_ui' && steps.dep-cache.outputs.cache-hit != 'true'
+        if: steps.dep-cache.outputs.cache-hit != 'true'
         # NOTE: if you update the jq version make sure to update the dep cache key
         run: |
-          mkdir -p /tmp/bats-cli-ui-deps
-          wget https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 -O /tmp/bats-cli-ui-deps/jq-bin
+          mkdir -p /tmp/test-deps
+          wget https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 -O /tmp/test-deps/jq-bin
       - name: Install jq for Bats CLI UI tests
-        if: matrix.test == 'cli_ui'
+        if: matrix.filter == 'integration test:cli_ui'
         run: |
-          chmod +x /tmp/bats-cli-ui-deps/jq-bin
-          sudo cp /tmp/bats-cli-ui-deps/jq-bin /usr/local/bin/jq
+          chmod +x /tmp/test-deps/jq-bin
+          sudo cp /tmp/test-deps/jq-bin /usr/local/bin/jq
       - name: Download and unzip pass for Boundary keyring
-        if: matrix.test == 'cli_ui' && steps.dep-cache.outputs.cache-hit != 'true'
+        if: steps.dep-cache.outputs.cache-hit != 'true'
         # NOTE: if you update the password store version make sure to update the dep cache key
         run: |
-          mkdir -p /tmp/bats-cli-ui-deps/pass
-          wget https://git.zx2c4.com/password-store/snapshot/password-store-1.7.4.tar.xz -O /tmp/bats-cli-ui-deps/pass/pass.tar.xz
-          cd /tmp/bats-cli-ui-deps/pass
+          mkdir -p /tmp/test-deps/pass
+          wget https://git.zx2c4.com/password-store/snapshot/password-store-1.7.4.tar.xz -O /tmp/test-deps/pass/pass.tar.xz
+          cd /tmp/test-deps/pass
           tar -xvf pass.tar.xz
       - name: Install pass for Boundary keyring
-        if: matrix.test == 'cli_ui'
         run: |
-          cd /tmp/bats-cli-ui-deps/pass/password-store-1.7.4
+          cd /tmp/test-deps/pass/password-store-1.7.4
           sudo make install
           pass init ${{ secrets.ENOS_GPG_UID }}
+      - name: Download Vault AMD64 binary for integration testing
+        if: steps.dep-cache.outputs.cache-hit != 'true'
+        run: |
+          wget https://releases.hashicorp.com/vault/1.11.4/vault_1.11.4_linux_amd64.zip -O /tmp/test-deps/vault.zip
+      - name: Install Vault for integration testing
+        if: matrix.filter == 'e2e_credential_vault'
+        run: |
+          unzip /tmp/test-deps/vault.zip -d /usr/local/bin
       - name: Download Linux AMD64 Boundary bundle
         id: download
         uses: actions/download-artifact@v3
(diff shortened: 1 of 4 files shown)


.github/workflows/enos-run.yml
```diff
@@ -37,21 +38,22 @@ jobs:
...
+        run: |
+          echo "${{ inputs.go-version }}"
       - name: Set up Terraform
         uses: hashicorp/setup-terraform@v2
         with:
           # the terraform wrapper will break Terraform execution in enos because
           # it changes the output to text when we expect it to be JSON.
           terraform_wrapper: false
       - name: Import GPG key for Boundary pass keystore
-        if: matrix.test == 'cli_ui'
```
jaymalasinha, line 47: if you need this step for both `cli_ui` and `e2e` scenarios but not for `smoke` you might be able to update this condition to `if: matrix.test != 'smoke'`

.github/workflows/enos-run.yml
```diff
@@ -157,9 +156,9 @@ jobs:
           ENOS_VAR_skip_failing_bats_tests: "true"
         run: |
           export ENOS_VAR_enos_user=$GITHUB_ACTOR && \
-          enos scenario run --timeout 60m0s --chdir ./enos integration test:${{ matrix.test}} builder:crt
+          enos scenario run --timeout 60m0s --chdir ./enos ${{ matrix.filter}} builder:crt
       - name: Destroy Enos scenario
-        if: ${{ always() }}
+        if: ${{ failure() && steps.run.conclusion == 'failure' || failure() && steps.run_retry.conclusion == 'failure' }}
```
jaymalasinha, line 161: It seems even though the destroy step failed before it did not error out in GHA with Enos version 0.0.10 and before, but now failure in destroy step does return error with Enos 0.0.11 .  May be we can add `continue-on-error: true` as a temporary workaround to this step with a comment, as the failure of destroy step does not determine the failure of Enos run

jaymalasinha (commented):
Just left couple comments, otherwise looks great 👍

.github/workflows/enos-run.yml
manthonygfp: (Not replying to Jaymala here. We talked about this in Slack. This comment is for posterity.)

In the case of continue-on-error, there will still be an icon indicating this step failed if the runs were successful. We know that's okay, but that isn't immediately clear to a random user that may see this.

continue-on-error is present for the whole job which already has this suggested functionality present, but my desire is to not have the red X indicating failure on this step if the step didn't need to run in the first place.

This solution keeps the step from running when it doesn't need to run and shows that it was appropriately skipped. This allows the user to just glance at it and not need to investigate further or have tribal knowledge about the workflow.

.github/workflows/enos-run.yml
```diff
@@ -142,7 +141,7 @@ jobs:
         run: |
           mkdir -p ./enos/terraform-plugin-cache
           export ENOS_VAR_enos_user=$GITHUB_ACTOR && \
-          enos scenario run --timeout 60m0s --chdir ./enos integration test:${{ matrix.test}} builder:crt
+          enos scenario run --timeout 60m0s --chdir ./enos ${{ matrix.filter}} builder:crt
```
manthonygfp, line 144:
```suggestion
          enos scenario run --timeout 60m0s --chdir ./enos ${{ matrix.filter }} builder:crt
```

.github/workflows/enos-run.yml
```diff
@@ -157,9 +156,9 @@ jobs:
           ENOS_VAR_skip_failing_bats_tests: "true"
         run: |
           export ENOS_VAR_enos_user=$GITHUB_ACTOR && \
-          enos scenario run --timeout 60m0s --chdir ./enos integration test:${{ matrix.test}} builder:crt
+          enos scenario run --timeout 60m0s --chdir ./enos ${{ matrix.filter}} builder:crt
```
manthonygfp, line 159:
```suggestion
          enos scenario run --timeout 60m0s --chdir ./enos ${{ matrix.filter }} builder:crt
```

.github/workflows/enos-run.yml
```diff
@@ -171,7 +170,7 @@ jobs:
           ENOS_VAR_skip_failing_bats_tests: "true"
         run: |
           export ENOS_VAR_enos_user=$GITHUB_ACTOR && \
-          enos scenario destroy --timeout 60m0s --chdir ./enos integration test:${{ matrix.test}} builder:crt
+          enos scenario destroy --timeout 60m0s --chdir ./enos ${{ matrix.filter}} builder:crt
```
manthonygfp, line 173:
```suggestion
          enos scenario destroy --timeout 60m0s --chdir ./enos ${{ matrix.filter }} builder:crt
```

.github/workflows/enos-run.yml
moduli: I'm a little confused.

The previous steps are executing `enos scenario run` which, based on my understanding, includes a `destroy`. Previously, this step executes another destroy. Now, this change is suggesting to only run this destroy if `run` or `run_retry` failed (whether it failed when running or destroying? Is that correct?

I'm not clear on what `continue-on-error` would do here.

manthonygfp: Addendum to the last bit, for posterity: Jaymala suggested that this could and probably should be handled in Enos itself. Terraform _also_ is okay with saying "There is nothing to destroy. My job is done." I'm inclined to agree that a destroy that does not need to destroy anything should exit 0. Hopefully this is temporary, and we will add functionality in Enos to allow this to execute without condition.

.github/workflows/enos-run.yml
```diff
@@ -75,43 +77,40 @@ jobs:
...
         # NOTE: if you update the jq version make sure to update the dep cache key
         run: |
           mkdir -p /tmp/bats-cli-ui-deps
           wget https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 -O /tmp/bats-cli-ui-deps/jq-bin
       - name: Install jq for Bats CLI UI tests
-        if: matrix.test == 'cli_ui'
+        if: matrix.filter == 'integration test:cli_ui' && steps.dep-cache.outputs.cache-hit != 'true'
         run: |
           chmod +x /tmp/bats-cli-ui-deps/jq-bin
           sudo cp /tmp/bats-cli-ui-deps/jq-bin /usr/local/bin/jq
```
moduli, line 104: In one of the new scenarios I'm adding, it will also need `jq`. I presume we'll need to add another block here for that?

.github/workflows/enos-run.yml
```diff
@@ -37,21 +38,22 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version: ${{ inputs.go-version }}
+      - name: Log Go version
+        run: |
+          echo "${{ inputs.go-version }}"
```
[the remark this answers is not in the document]
manthonygfp: Oh great catch- this was a debug statement that was there to ensure that we were in fact using the same version of Go as the .go-version file (which it was, of course). I'll remove it.

.github/workflows/enos-run.yml
manthonygfp: @moduli That's correct. The destroy part of the lifecycle of `enos scenario run` only executes if the apply part of it was successful. This discrete call to `enos scenario destroy` ensures that the infrastructure is always destroyed even in the case of the run or run-retry step's failure.

With the previous logic, if the destroy step ran after the run was successful, you would have an icon on the workflow run page that indicates the failure of a step. This change makes it to where the step is explicitly skipped in the event that the run (or retry of course) step was successful.

But, because of the discussion in this PR, Ryan is making a change to Enos that will make the `enos scenario destroy` action not return a non-zero exit code if it simply has nothing to destroy. As a result, we will likely move the conditional from this step before it is merged.

.github/workflows/enos-run.yml
manthonygfp: You make a good point implicitly that this could grow to be unwieldy. I think one approach would be for us to make the inversion of this logic instead, but maybe we could more simply _always_ install `jq` as it won't harm scenarios that _don't_ use it.

manthonygfp:
One of my recent commits has an inaccurate commit message, but these will be squashed so I'm not going to go through the effort of amending it.

.github/workflows/enos-run.yml
manthonygfp:
```suggestion
        run: |
          chmod +x /tmp/bats-cli-ui-deps/jq-bin
          sudo cp /tmp/bats-cli-ui-deps/jq-bin /usr/local/bin/jq
```

What do we think about removing it?

enos/enos-variables.hcl
```diff
@@ -79,6 +79,7 @@ variable "target_count" {
 variable "local_boundary_dir" {
   description = "Path to local boundary executable"
   type        = string
+  default     = "/tmp"
```
manthonygfp, line 82: We could maybe do away with this as it was part of trying to figure out a previous issue. It's not harming anything to be here.
```suggestion
```

.github/workflows/enos-run.yml
manthonygfp, line 43:
```suggestion
```

.github/workflows/enos-run.yml
moduli: Ah. Got it.

enos/enos-variables.hcl
moduli: I would lean towards removing it if it's not used for anything.

.github/workflows/enos-run.yml
moduli: Seems ok? Reminder: we'd have to remove it in the download step above, too.

Alternatively, would it make sense for me to handle this in the scenario? Would that even be possible?

.github/workflows/enos-run.yml
manthonygfp: Ryan has updated Enos- I will remove the condition altogether 👏

.github/workflows/enos-run.yml
manthonygfp: @jaymalasinha do you think that ensuring `jq` is present should be handled in the GHA workflow or Enos scenario?

.github/workflows/enos-run.yml
manthonygfp: Removing the condition as Ryan has a change to Enos that will let destroy run and exit cleanly if it has nothing to destroy.

.github/workflows/enos-run.yml
```diff
@@ -157,9 +153,9 @@ jobs:
           ENOS_VAR_skip_failing_bats_tests: "true"
         run: |
           export ENOS_VAR_enos_user=$GITHUB_ACTOR && \
-          enos scenario run --timeout 60m0s --chdir ./enos integration test:${{ matrix.test}} builder:crt
+          enos scenario run --timeout 60m0s --chdir ./enos ${{ matrix.filter }} builder:crt
       - name: Destroy Enos scenario
-        if: ${{ always() }}
+        if: ${{ failure() && steps.run.conclusion == 'failure' || failure() && steps.run_retry.conclusion == 'failure' }}
```
manthonygfp, line 158:
```suggestion
```

.github/workflows/enos-run.yml
manthonygfp:
```suggestion
```

.github/workflows/enos-run.yml
jaymalasinha: `jq` install can be done within the Enos scenario modules

manthonygfp: Hm but do you think we should remove it here and edit all of the scenarios that need it to include it...

or should we delegate that task to the workflow?

I guess my question is that we don't really want to be doing this in different ways that get stuffed into the same workflow (some installing jq with Enos, some doing it with the GHA workflow).

manthonygfp: I talked with Michael about this out of the comments- we're going to install it as part of the workflow as we already assume other local deps (like Go) are installed when this is running on someone's MacBook.

.github/workflows/enos-run.yml
```diff
@@ -75,43 +74,40 @@ jobs:
...
         uses: actions/setup-node@v3
         with:
           node-version: 16
       - name: Install Bats via NPM
-        if: matrix.test == 'cli_ui'
         # Use npm so this workflow is portable on multiple runner distros
         run: npm install --location=global bats
       - name: Download jq for Bats CLI UI tests
-        if: matrix.test == 'cli_ui' && steps.dep-cache.outputs.cache-hit != 'true'
+        if: matrix.filter == 'integration test:cli_ui' && steps.dep-cache.outputs.cache-hit != 'true'
```
manthonygfp, line 92:
```suggestion
```

manthonygfp, line 92:
```suggestion
        if: steps.dep-cache.outputs.cache-hit != 'true'
```

manthonygfp:
I think we're good to go, waiting on the changes to Enos we had talked about to be merged before merging the PR.

.github/workflows/enos-run.yml
```diff
@@ -75,43 +74,40 @@ jobs:
...
         uses: actions/setup-node@v3
         with:
           node-version: 16
       - name: Install Bats via NPM
-        if: matrix.test == 'cli_ui'
         # Use npm so this workflow is portable on multiple runner distros
         run: npm install --location=global bats
       - name: Download jq for Bats CLI UI tests
-        if: matrix.test == 'cli_ui' && steps.dep-cache.outputs.cache-hit != 'true'
+        if: steps.dep-cache.outputs.cache-hit != 'true'
```
moduli, line 92: nit: For these updated steps that aren't checking for `cli_ui` anymore, we may want to update the `name` so that it's not referring to `Bats CLI UI tests`

moduli:
Also, just noticed that the PR status check for this is failing? What's going on there?

jaymalasinha (approved):
LGTM

jaymalasinha (approved):

moduli (approved):

joshbrand (approved):
🎉 🎉 🎉

PR #2503, opened 2022-09-28
Recording and Persisting Session Connection's Bytes Up and Down

hugoamvieira:
This PR contains a set of commits that enable the collection of the amount bytes received and sent over a worker's connection between a client and target.


internal/db/schema/migrations/oss/postgres/56/01_session_connection_bytes_up_down_trigger.up.sql
```diff
@@ -0,0 +1,23 @@
...
+comment on function prevent_bytes_update_on_closed_connection is
+  'The last update of bytes_up and bytes_down for any session_connection should
+  be the one that happens as the consequence of a connection closure
+  (when closed_reason is also set). This function ensures those fields cannot
+  be updated past connection closure';
+
+create trigger update_connection_bytes before update of bytes_up, bytes_down on session_connection
+  for each row execute procedure prevent_bytes_update_on_closed_connection();
+
+commit;
```
hugoamvieira, line 23: Do we want this? 

Introducing this check in the database acts as a second layer of defense against potentially bad state in the db (the first being the [update query itself](https://github.com/hashicorp/boundary/pull/2503/files#diff-249aec8bcdb2e6880afd550f5cdb35d4026fedf391dd5425f600babc63bf3438R107)), however it can also be confusing from a design perspective as to why we carved out those two fields specifically, considering that no field should really be updated after a connection is marked as closed.

cc @mgaffney @tmessi

internal/tests/cluster/worker_bytesupdown_test.go
```diff
@@ -0,0 +1,131 @@
...
+	pl, err := net.Listen("tcp", "localhost:0")
+	require.NoError(err)
+	c1 := controller.NewTestController(t, &controller.TestControllerOpts{
+		Config:                    conf,
+		InitialResourcesSuffix:    "1234567890",
+		Logger:                    logger.Named("c1"),
+		PublicClusterAddr:         pl.Addr().String(),
+		StatusGracePeriodDuration: helper.DefaultGracePeriod,
+	})
+	defer c1.Shutdown()
```
johanbrandhorst, line 40: Nit: use `t.Cleanup` for this sort of thing in tests, and don't forget to check the errors!

internal/tests/cluster/worker_bytesupdown_test.go
```diff
@@ -0,0 +1,131 @@
...
+	// Wire up the testing proxies
+	require.Len(c1.ClusterAddrs(), 1)
+	proxy, err := dawdle.NewProxy("tcp", "", c1.ClusterAddrs()[0],
+		dawdle.WithListener(pl),
+		dawdle.WithRbufSize(256),
+		dawdle.WithWbufSize(256),
+	)
+	require.NoError(err)
+	defer proxy.Close()
+	require.NotEmpty(t, proxy.ListenerAddr())
```
johanbrandhorst, line 53: Any way we can count the actual bytes going up and down this and compare it against the reported bytes?

internal/cmd/commands/sessionscmd/funcs.go
```diff
@@ -230,6 +232,24 @@ func printItemTable(item *sessions.Session, resp *api.Response) string {
...
+		cm := map[string]interface{}{
+			"Client Address":   net.JoinHostPort(sc.ClientTcpAddress, strconv.FormatUint(uint64(sc.ClientTcpPort), 10)),
+			"Endpoint Address": net.JoinHostPort(sc.EndpointTcpAddress, strconv.FormatUint(uint64(sc.EndpointTcpPort), 10)),
+		}
+		if sc.BytesUp > 0 {
+			cm["Bytes Up"] = sc.BytesUp
+		}
+		if sc.BytesDown > 0 {
+			cm["Bytes Down"] = sc.BytesDown
+		}
```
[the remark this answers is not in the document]
hugoamvieira: Good point!

internal/tests/cluster/worker_bytesupdown_test.go
hugoamvieira: I don't see any methods within `dawdle.Proxy` to allow us to do this - We'd likely have to wrap the `net.Listener` with something to count `Read` and `Write` calls like we did for this task

internal/tests/cluster/worker_bytesupdown_test.go
hugoamvieira: Fixed in [88742e5](https://github.com/hashicorp/boundary/pull/2503/commits/88742e510ec70d1934256b6323dbdd667ed28e09)

johanbrandhorst (approved):
LGTM pending decision on trigger, I'm on favor of removing it

internal/db/schema/migrations/oss/postgres/56/01_session_connection_bytes_up_down_trigger.up.sql
mgaffney: I don't like this. It feels like we are moving domain logic into the database.

internal/session/service_worker_status_report.go
```diff
@@ -25,24 +25,34 @@ type StateReport struct {
...
 	}
 	if len(closed) > 0 {
 		event.WriteSysEvent(ctx, op, "marked unclaimed connections as closed", "controller_id", workerId, "count", len(closed))
 	}
-	return notActive, err
+
+	err = connRepo.updateBytesUpBytesDown(ctx, reportedConnections...)
+	if err != nil {
+		return notActive, errors.New(ctx, errors.Internal, op, fmt.Sprintf("failed to update bytes up and down for worker reported connections: %v", err))
+	}
```
mgaffney, line 55: I'm concerned that updates to the bytes up and bytes down values might be lost because of a potentially unrelated error being returned from either `checkIfNoLongerActive` or `closeOrphanedConnections`. Can we prevent this from happening?

mgaffney (changes requested):
I think we should remove the trigger

internal/session/repository_connection.go
```diff
@@ -81,6 +82,45 @@ func (r *ConnectionRepository) list(ctx context.Context, resources interface{},
...
+			for _, c := range conns {
+				_, err := w.Update(
+					ctx,
+					&Connection{PublicId: c.PublicId, BytesUp: c.BytesUp, BytesDown: c.BytesDown},
+					updateMask,
+					nil,
+					// The last update to these two fields should come from our
+					// connection closure logic, so we don't update here if the
+					// connection has already been closed. There's also a
+					// trigger on the database itself to prevent this.
```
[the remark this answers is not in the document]
hugoamvieira: Addressed in [ac9b087](https://github.com/hashicorp/boundary/pull/2503/commits/ac9b087b0c553e3516db4bc54a0dbd7bbdbc7c6b)

internal/db/schema/migrations/oss/postgres/56/01_session_connection_bytes_up_down_trigger.up.sql
hugoamvieira: Addressed in [ac9b087](https://github.com/hashicorp/boundary/pull/2503/commits/ac9b087b0c553e3516db4bc54a0dbd7bbdbc7c6b)

internal/session/service_worker_status_report.go
hugoamvieira: Good catch - I noticed that none of the calls in that function actually depend on each other, so I changed it to use `multierror` and just return error or nil at the end. [c2bf0f3](https://github.com/hashicorp/boundary/pull/2503/commits/c2bf0f309456ba769136c9b27f81c00ecb5f2d89)

internal/session/service_worker_status_report.go
```diff
@@ -6,43 +6,60 @@ import (
...
 )
 
 // StateReport is used to report on the state of a Session.
 type StateReport struct {
-	SessionId     string
-	Status        Status
-	ConnectionIds []string
+	SessionId   string
+	Status      Status
+	Connections []Connection
```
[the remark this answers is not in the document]
hugoamvieira: Well we're actually instantiating those status report objects (and their connections) from data on other objects [here](https://github.com/hashicorp/boundary/blob/main/internal/daemon/cluster/handlers/worker_service.go#L145-L173), not assigning them from another object, so I suppose it doesn't really matter whether it's a pointer or not here?

internal/daemon/worker/session/session.go
```diff
@@ -351,6 +382,19 @@ func (s *sess) CancelAllLocalConnections() []string {
 	return closedIds
 }
 
+func (s *sess) ApplyConnectionCounterCallbacks(connId string, bytesUp func() int64, bytesDown func() int64) error {
```
mgaffney, line 385: Can we add a doc string since this is exported?

internal/daemon/cluster/handlers/worker_service.go
```diff
@@ -158,14 +158,18 @@ func (ws *workerServiceServer) Status(ctx context.Context, req *pbs.StatusReques
 			}
 
 			sr := session.StateReport{
-				SessionId:     si.GetSessionId(),
-				ConnectionIds: make([]string, 0, len(si.GetConnections())),
+				SessionId:   si.GetSessionId(),
+				Connections: make([]session.Connection, 0, len(si.GetConnections())),
```
mgaffney, line 162:
```suggestion
				Connections: make([]*session.Connection, 0, len(si.GetConnections())),
```

mgaffney (changes requested):
Overall, looks good. Just needs some small changes as we discussed

internal/session/service_worker_status_report.go
```diff
@@ -6,43 +6,60 @@ import (
...
+// WorkerStatusReport is a domain service function that, given a Worker's
+// session state reports, performs a few tasks:
+//  1. Updates the bytes up and down statistics for each reported connection.
+//  2. Compares the state of sessions and connections as reported by a Worker,
+//     to the known state in the repositories. It returns a StateReport object
+//     for each session that is in the canceling or terminated state.
+//  3. Checks for any orphaned connections, which is defined as a connection
+//     that is in an active state, but was not reported by worker. Any orphaned
+//     connections will be marked as closed.
 func WorkerStatusReport(ctx context.Context, repo *Repository, connRepo *ConnectionRepository, workerId string, report []StateReport) ([]StateReport, error) {
```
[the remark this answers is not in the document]
hugoamvieira: Addressed in [1f2d0a8](https://github.com/hashicorp/boundary/pull/2503/commits/1f2d0a8f6e410ea3e542f8439c417a917fd9c2e8)

internal/session/service_worker_status_report.go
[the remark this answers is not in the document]
hugoamvieira: Addressed in [1f2d0a8](https://github.com/hashicorp/boundary/pull/2503/commits/1f2d0a8f6e410ea3e542f8439c417a917fd9c2e8)

internal/session/repository_connection.go
```diff
@@ -81,6 +81,49 @@ func (r *ConnectionRepository) list(ctx context.Context, resources interface{},
...
+	if len(conns) == 0 {
+		return nil
+	}
+
+	updateMask := []string{"BytesUp", "BytesDown"}
+	_, err := r.writer.DoTx(
+		ctx,
+		db.StdRetryCnt,
+		db.ExpBackoff{},
+		func(read db.Reader, w db.Writer) error {
```
mgaffney, line 95:
```suggestion
		func(_ db.Reader, w db.Writer) error {
```
nit: the `read` parameter is not used so replacing it with the blank identifier makes that clearer.

PR #2505, opened 2022-09-28
chore: update sdk/api deps

louisruch:


louisruch:
@johanbrandhorst @irenarindos PTAL
If we do push a new 0.11.1 we should add this to it as well

irenarindos (approved):

johanbrandhorst (approved):

PR #2509, opened 2022-09-30
feat(worker): Report bytes up/down on Worker <-> Client connection closure

hugoamvieira:
Introduces data about bytes read and written to the connection close payload on the Worker's side. No changes are required in the Controller because the RPC messages already contained the required fields. As of this commit, bytes_up and bytes_down columns in the session_connection table are now populated when a connection closes.

### DIFF
--- a/internal/daemon/worker/countingconn.go
+++ b/internal/daemon/worker/countingconn.go
@@ -5,6 +5,10 @@ import (
 	"sync"
 )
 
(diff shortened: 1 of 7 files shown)


johanbrandhorst (approved):
LGTM, I'm mildly surprised this turn into the production solution for this problem as it was something I hacked together for the prototype, but it looks good!

tmessi (approved):

PR #2511, opened 2022-10-03
Add interfaces and error codes for processing downstream router connections

talanknight:


internal/daemon/controller/controller.go
```diff
@@ -49,6 +49,11 @@ type downstreamRouter interface {
 	// StartRouteMgmtTicking starts a ticker which manages the router's
 	// connections.
 	StartRouteMgmtTicking(context.Context, func() string, int) error
+
+	// ProcessPendingConnections starts a function that continually processes
+	// incoming client connections and processes them. This never returns until
```
johanbrandhorst, line 54: Nit: this reads a bit weird 😁.
```suggestion

	// ProcessPendingConnections starts a function that continually processes
	// incoming client connections. This never returns until
```

internal/daemon/worker/worker.go
```diff
@@ -46,6 +46,11 @@ type downstreamRouter interface {
 	// StartRouteMgmtTicking starts a ticker which manages the router's
 	// connections.
 	StartRouteMgmtTicking(context.Context, func() string, int) error
+
+	// ProcessPendingConnections starts a function that continually processes
+	// incoming client connections and processes them. This never returns until
```
johanbrandhorst, line 51:
```suggestion
	// ProcessPendingConnections starts a function that continually processes
	// incoming client connections. This never returns until
```

internal/errors/code.go
```diff
@@ -45,6 +45,8 @@ const (
 	CycleFound                         = 121 // CycleFound represents an error when a cycle is found between a parent and child worker
 	WorkerConnNotFound                 = 122 // WorkerConnNotFound represents an error when a connection to a worker is not found
 	KmsWorkerUnsupportedOperation      = 123 // KmsWorkerUnsupportedOperation represents an error when a KMS worker is not supported for an operation
+	RetryLimitExceeded                 = 124 // RetryLimitExceeded represents an error when a retry limit is exceeded
+	QueueIsFull                        = 125 // QueueIsFull results in attempting to add an item to a queue which is full
```
johanbrandhorst, line 49: Where are these used?

johanbrandhorst (changes requested):

internal/errors/code.go
johanbrandhorst: Could we add a comment? I think I see the practical reason for putting this here but it might still be confusing to see unused constants.

irenarindos (approved):

PR #2513, opened 2022-10-03
Add test helper for creating a multihop worker setup

talanknight:


internal/daemon/worker/testing.go
```diff
@@ -367,3 +375,84 @@ func (tw *TestWorker) AddClusterWorkerMember(t testing.TB, opts *TestWorkerOpts)
...
+	childPkiWorkerReq := new(types.FetchNodeCredentialsRequest)
+	require.NoError(proto.Unmarshal(reqBytes, childPkiWorkerReq))
+	_, err = serversRepo.CreateWorker(controllerContext, &server.Worker{
+		Worker: &store.Worker{
+			ScopeId: scope.Global.String(),
+		},
+	}, server.WithFetchNodeCredentialsRequest(childPkiWorkerReq))
+	require.NoError(err)
+
+	// Sleep so that DAG gets updated
```
irenarindos, line 454: There's no DAG in OSS; maybe we should add an option so that we skip the sleep on OSS but enable it on ENT?

irenarindos (changes requested):
Minor suggestion but looks good otherwise

irenarindos (approved):

PR #2515, opened 2022-10-03
Disconnect PKI workers from upstreams when they are no longer authorized

talanknight:
With this PR the worker and controller intermittently check if the downstream pki workers connected to them are still authorized in the boundary system.  If a connected pki worker is found to no longer be authorized the underlying connection is closed.

For controllers this happens every 3 seconds, for workers this happens in every status update request.


internal/proto/controller/servers/services/v1/server_coordination_service.proto
```diff
@@ -124,6 +128,10 @@ message StatusResponse {
   // The ID of the worker which made the request. The worker can send this value in subsequent requests so the
   // controller does not need to do a database lookup for the id using the name field.
   string worker_id = 40; // @gotags: `class:"public"`
+
+  // Of the worker key identifiers provided in the request, these are the ones
+  // which are authorized to remain connected.
+  repeated string authroized_worker_key_identifiers = 50;
```
irenarindos, line 134:
```suggestion
  repeated string authorized_worker_key_identifiers = 50;
```

internal/daemon/controller/tickers.go
```diff
@@ -155,3 +157,31 @@ func (c *Controller) startCloseExpiredPendingTokens(cancelCtx context.Context) {
 		}
 	}
 }
+
+func (c *Controller) startWorkerConnectionMaintenanceTicking(cancelCtx context.Context, m *cluster.DownstreamManager) {
+	const op = "controller.(downstreamManager).startManagingTicker"
```
irenarindos, line 162:
```suggestion
	const op = "controller.(Controller).startWorkerConnectionMaintenanceTicking"
```

irenarindos, line 162: Check for nil DownstreamManager?

internal/daemon/controller/tickers.go
```diff
@@ -13,8 +14,9 @@ import (
 
 // In the future we could make this configurable
 const (
-	statusInterval      = 10 * time.Second
-	terminationInterval = 1 * time.Minute
+	workerConnectionMaintenanceInterval = 3 * time.Second
```
irenarindos, line 17: Do we want a bigger time interval to ease db load/ should it at least be as long as the status interval?

internal/daemon/controller/tickers.go
```diff
@@ -155,3 +157,31 @@ func (c *Controller) startCloseExpiredPendingTokens(cancelCtx context.Context) {
 		}
 	}
 }
+
+func (c *Controller) startWorkerConnectionMaintenanceTicking(cancelCtx context.Context, m *cluster.DownstreamManager) {
+	const op = "controller.(downstreamManager).startManagingTicker"
+	timer := time.NewTimer(0)
```
irenarindos, line 163: Can you add the getRandomInterval  that other tickers in this file use so that controllers aren't hitting the db at the same time?

irenarindos (changes requested):

internal/server/repository_workerauth.go
```diff
@@ -520,38 +520,25 @@ func (r *WorkerAuthRepositoryStorage) findCertBundles(ctx context.Context, worke
...
+	const op = "server.(WorkerAuthRepositoryStorage).FilterToAuthorizedWorkerKeyIds"
 	if len(workerKeyIds) == 0 {
 		return nil, nil
 	}
-	rows, err := r.reader.Query(ctx, authorizedWorkerQuery, []interface{}{workerKeyIds})
-	if err != nil {
+	var bundles []*WorkerCertBundle
+	if err := r.reader.SearchWhere(ctx, &bundles, "worker_key_identifier in (?)", []interface{}{workerKeyIds}, db.WithLimit(-1)); err != nil {
 		return nil, errors.Wrap(ctx, err, op)
 	}
```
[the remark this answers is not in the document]
tmessi: I took "custom query" to mean crafting a query that gets passed directly to [Query](https://github.com/hashicorp/boundary/blob/51de0878c37658203f64c8080f04815ba4b317f9/internal/db/read_writer.go#L236-L251) or [Exec](https://github.com/hashicorp/boundary/blob/51de0878c37658203f64c8080f04815ba4b317f9/internal/db/read_writer.go#L221-L234) instead of calling things like [SearchWhere](https://github.com/hashicorp/boundary/blob/51de0878c37658203f64c8080f04815ba4b317f9/internal/db/read_writer.go#L497-L517) that construct a query. There are many examples that fall into this category, generally they are in a `query.go` file in the domain packages:

- [session/query.go](https://github.com/hashicorp/boundary/blob/51de0878c37658203f64c8080f04815ba4b317f9/internal/session/query.go#L8-L425)
- [iam/query.go](https://github.com/hashicorp/boundary/blob/51de0878c37658203f64c8080f04815ba4b317f9/internal/iam/query.go#L6-L111)
- [target/query.go](https://github.com/hashicorp/boundary/blob/51de0878c37658203f64c8080f04815ba4b317f9/internal/target/query.go#L3-L58)
- [server/query.go](https://github.com/hashicorp/boundary/blob/51de0878c37658203f64c8080f04815ba4b317f9/internal/server/query.go#L3-L33)
- [credential/vault/query.go](https://github.com/hashicorp/boundary/blob/51de0878c37658203f64c8080f04815ba4b317f9/internal/credential/vault/query.go#L3-L189)

internal/daemon/controller/tickers.go
jefferai: A reminder that this will change a bit with https://github.com/hashicorp/boundary/pull/2578

jefferai: Not status interval, grace and such. But just want to surface it in case it matters somewhere in here.

jefferai (approved):
Great job!!!

PR #2528, opened 2022-10-10
Delete the base auth account when a concrete auth account is deleted

talanknight:
This deletes records from auth_account if the public id doesn't exist in either auth_password_account or auth_oidc_account.  

It also creates a trigger to delete records from auth_account when the cooresponding record in auth_password_account or auth_oidc_account is deleted.  This behavior is already in place for auth_method.

This fixes Jira task https://hashicorp.atlassian.net/browse/ICU-4377

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,12 @@
 
 Canonical reference for changes, improvements, and bugfixes for Boundary.
 
- ## Next
+## Next
+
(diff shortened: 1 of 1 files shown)


mgaffney (changes requested):
Can we add tests to the `sqltest` suite?

mgaffney (approved):

PR #2531, opened 2022-10-11
Filter out null named workers when looking up the worker id in bats

talanknight:


internal/tests/cli/boundary/_workers.bash
```diff
@@ -24,7 +24,7 @@ function list_workers() {
 
 function worker_id() {
   local id=$1
-  strip $(list_workers | jq -c ".items[] | select(.name | contains(\"$id\")) | .[\"id\"]")
+  strip $(list_workers | jq -c ".items[] | select(.name != null) | select(.name | contains(\"$id\")) | .[\"id\"]")
```
moduli, line 27: Curious, where are the `null` workers coming from?

moduli (approved):

PR #2532, opened 2022-10-11
Rewrap Functions for Encrypted Tables

DanielleMiu:
- [x] ALL 12 REWRAPS DONE
- [x] update sql files to include `updated in 56/05_mutable_ciphertext_columns.up.sql` message
- [x] fix root cert test

root cert pk change has been moved to [boundary#2542](https://github.com/hashicorp/boundary/pull/2542)

### DIFF
--- a/internal/auth/oidc/rewrapping.go
+++ b/internal/auth/oidc/rewrapping.go
@@ -1,0 +1,65 @@
+package oidc
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/kms"
+)
+
+func init() {
+	kms.RegisterTableRewrapFn(defaultAuthMethodTableName, authMethodRewrapFn)
+}
+
+func authMethodRewrapFn(ctx context.Context, dataKeyVersionId string, reader db.Reader, writer db.Writer, kmsRepo *kms.Kms) error {
+	const op = "oidc.authMethodRewrapFn"
+	repo, err := NewRepository(ctx, reader, writer, kmsRepo)
+	if err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+	rows, err := repo.reader.Query(ctx, fmt.Sprintf(`select distinct scope_id from %q where key_id=?`, defaultAuthMethodTableName), []interface{}{dataKeyVersionId})
+	if err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+	var scopeIds []string
+	for rows.Next() {
+		var scopeId string
+		if err := rows.Scan(&scopeId); err != nil {
+			_ = rows.Close()
+			return errors.Wrap(ctx, err, op)
+		}
+		scopeIds = append(scopeIds, scopeId)
+	}
+	if err := rows.Err(); err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+	for _, scopeId := range scopeIds {
+		var authMethods []*AuthMethod
+		if err := repo.reader.SearchWhere(ctx, &authMethods, "scope_id=? and key_id=?", []interface{}{scopeId, dataKeyVersionId}, db.WithLimit(-1)); err != nil {
+			return errors.Wrap(ctx, err, op)
+		}
+		wrapper, err := repo.kms.GetWrapper(ctx, scopeId, kms.KeyPurposeDatabase)
+		if err != nil {
+			return errors.Wrap(ctx, err, op)
+		}
+		newKeyVersionId, err := wrapper.KeyId(ctx)
+		if err != nil {
+			return errors.Wrap(ctx, err, op)
+		}
+		for _, am := range authMethods {
+			if err := am.decrypt(ctx, wrapper); err != nil {
+				return errors.Wrap(ctx, err, op)
+			}
+			if err := am.encrypt(ctx, wrapper); err != nil {
+				return errors.Wrap(ctx, err, op)
+			}
+			am.KeyId = newKeyVersionId
+			if _, err := repo.writer.Update(ctx, am, []string{CtClientSecretField, KeyIdField}, nil); err != nil {
+				return errors.Wrap(ctx, err, op)
+			}
+		}
+	}
+	return nil
+}
--- a/internal/auth/oidc/rewrapping.go
+++ b/internal/auth/oidc/rewrapping.go
@@ -2,7 +2,6 @@ package oidc
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/hashicorp/boundary/internal/db"
 	"github.com/hashicorp/boundary/internal/errors"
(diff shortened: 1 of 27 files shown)


internal/server/query.go
```diff
@@ -23,4 +23,15 @@ const (
...
+	workerAuthRewrapQuery = `
+		select distinct
+		auth.worker_key_identifier,
+		auth.controller_encryption_priv_key,
+		auth.key_id
+		from server_worker worker
+		inner join worker_auth_authorized auth
+		on auth.worker_id = worker.public_id
+		and auth.key_id = ?
+		where worker.scope_id = ?
```
johanbrandhorst, line 35: I don't know if this matters, but it's weird to me that the `key_id` comparison is part of the join and not the where clause. What do you think about
```suggestion
		on auth.worker_id = worker.public_id
		where worker.scope_id = ? and auth.key_id = ?
```

Also consider using indentation to make this a bit easier to read, something like
```go
workerAuthRewrapQuery = `
		select distinct
			auth.worker_key_identifier,
			auth.controller_encryption_priv_key,
			auth.key_id
		from
			server_worker 		worker
		inner join
			worker_auth_authorized 	auth
		on
			auth.worker_id = worker.public_id
		where
			worker.scope_id = ? and auth.key_id = ?
`
```

internal/server/rewrapping.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	for _, cert := range certs {
+		privateKey, err := decrypt(ctx, cert.PrivateKey, wrapper)
+		if err != nil {
+			return errors.Wrap(ctx, err, op)
+		}
+		if cert.PrivateKey, err = encrypt(ctx, privateKey, wrapper); err != nil {
+			return errors.Wrap(ctx, err, op)
+		}
+		cert.KeyId = newKeyVersionId
```
johanbrandhorst, line 39: We should probably refactor this to be a `encrypt` and `decrypt` method on the `RootCertificate` type instead. What do you think? I think it would also mean moving these types to use `structwrapping` like the other encrypt and decrypt methods do.

internal/server/rewrapping.go
```diff
@@ -0,0 +1,92 @@
...
+		return errors.Wrap(ctx, err, op)
+	}
+	for rows.Next() {
+		workerAuth := allocWorkerAuth()
+		if err := rows.Scan(
+			&workerAuth.WorkerKeyIdentifier,
+			&workerAuth.ControllerEncryptionPrivKey,
+			&workerAuth.KeyId,
+		); err != nil {
+			_ = rows.Close()
```
johanbrandhorst, line 62: This is fine, but the general convention is to just defer this call. I only mention it because it's worth having consistency. You can keep this if you like, but maybe check other places in this package where we iterate rows and see what's done there.

internal/server/rewrapping.go
```diff
@@ -0,0 +1,92 @@
...
+	wrapper, err := kmsRepo.GetWrapper(ctx, scopeId, kms.KeyPurposeDatabase)
+	if err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+	newKeyVersionId, err := wrapper.KeyId(ctx)
+	if err != nil {
+		return errors.Wrap(ctx, err, op)
+	}
+	for _, workerAuth := range auths {
+		privateKey, err := decrypt(ctx, workerAuth.ControllerEncryptionPrivKey, wrapper)
```
johanbrandhorst, line 79: Re: refactor to `encrypt`, `decrypt` method.

internal/session/credential.go
```diff
@@ -12,10 +12,11 @@ import (
...
 type credential struct {
-	SessionId    string
-	KeyId        string
-	Credential   []byte `gorm:"-" wrapping:"pt,credential_data"`
-	CtCredential []byte `gorm:"column:credential" wrapping:"ct,credential_data"`
+	SessionId        string `gorm:"index:,unique,composite:session_credential_session_id_credential_sha256_uq"`
+	KeyId            string
+	Credential       []byte `gorm:"-" wrapping:"pt,credential_data"`
+	CtCredential     []byte `gorm:"column:credential" wrapping:"ct,credential_data"`
+	CredentialSha256 []byte `gorm:"index:,unique,composite:session_credential_session_id_credential_sha256_uq"`
```
johanbrandhorst, line 19: Should we make this a generated type?

internal/session/query.go
```diff
@@ -407,6 +407,25 @@ and
...
+select distinct
+  cred.session_id,
+  cred.key_id,
+  cred.credential,
+  cred.credential_sha256
+from session
+  inner join session_credential cred
+    on cred.session_id = session.public_id
+      and cred.key_id = ?
+where session.project_id = ?
```
johanbrandhorst, line 421: Re: should this comparison be part of the `where`?
```suggestion
where session.project_id = ? and cred.key_id = ?
```

internal/session/query.go
```diff
@@ -407,6 +407,25 @@ and
...
+`
+	sessionCredentialRewrapUpdate = `
+update session_credential
+	set credential = ?,
+		key_id = ?
+where session_id = ?
+	and credential_sha256 = ?;
```
johanbrandhorst, line 428: Can this be done with a gorm type update instead? This was the table that didn't have a primary key right?

johanbrandhorst (commented):
This looks great, good job on all these changes! I have some scattered thoughts but overall this is perfect.

internal/auth/oidc/rewrapping.go
```diff
@@ -0,0 +1,38 @@
...
+		return errors.Wrap(ctx, err, op)
+	}
+	for _, am := range authMethods {
+		if err := am.decrypt(ctx, wrapper); err != nil {
+			return errors.Wrap(ctx, err, op)
+		}
+		if err := am.encrypt(ctx, wrapper); err != nil {
+			return errors.Wrap(ctx, err, op)
+		}
+		if _, err := writer.Update(ctx, am, []string{CtClientSecretField, ClientSecretHmacField, KeyIdField}, nil); err != nil {
```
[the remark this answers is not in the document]
DanielleMiu: so I did make an assumption here that because the wrapper is passed into the hmac function (it's called cipher within the encrypt func), that the current encrypting key is used to calculate it, but I will double check this and get back to you regarding the consistency of calculated hmacs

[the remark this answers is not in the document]
DanielleMiu: just confirmed, the [underlying hmac function](https://github.com/hashicorp/boundary/blob/main/internal/libs/crypto/hmac_sha256.go#L52-L56) and [bytes assigned to the reader](https://github.com/hashicorp/boundary/blob/main/internal/libs/crypto/derived_reader.go#L45) passed to the ed25519 key generator use the cipher bytes from the kms wrapper, so as long as the key version used to generate the hmac stays the same, the generated hmac will also stay the same.

-----
tested:
```
assert.NoError(t, got.decrypt(ctx, kmsWrapper2))
fmt.Printf("decrypted: %x\n", got.PasswordHmac)
assert.NoError(t, got.encrypt(ctx, kmsWrapper2))
fmt.Printf("encrypted: %x\n", got.PasswordHmac)
assert.NoError(t, got.decrypt(ctx, kmsWrapper2))
fmt.Printf("decrypted: %x\n", got.PasswordHmac)
assert.NoError(t, got.encrypt(ctx, kmsWrapper2))
fmt.Printf("encrypted: %x\n", got.PasswordHmac)
```
output
```
decrypted: 773b143a91bb516482497e8510c0a560c6fea6d7c146cc214166d2f9356ba5a5
encrypted: 773b143a91bb516482497e8510c0a560c6fea6d7c146cc214166d2f9356ba5a5
decrypted: 773b143a91bb516482497e8510c0a560c6fea6d7c146cc214166d2f9356ba5a5
encrypted: 773b143a91bb516482497e8510c0a560c6fea6d7c146cc214166d2f9356ba5a5
```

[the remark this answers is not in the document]
DanielleMiu: so, to answer your question. we're updating the key version used to encrypt the fields, which updates the wrapper used by the hmac function, which updates key bytes passed to the ed25519 key generator. all of which results in a new hmac

internal/authtoken/rewrapping.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+func init() {
+	kms.RegisterTableRewrapFn(defaultAuthTokenTableName, authTokenRewrapFn)
+}
+
+func authTokenRewrapFn(ctx context.Context, dataKeyVersionId, scopeId string, reader db.Reader, writer db.Writer, kmsRepo *kms.Kms) error {
+	const op = "authtoken.authTokenRewrapFn"
+	var credentials []*AuthToken
+	// indexes on public id and token, and the only reference to public id is session (unreliable join). this is the fastest query
```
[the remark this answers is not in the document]
DanielleMiu: I was trying to find a succinct way of saying, there's only one reference and there may not be sessions for all of the rows we're trying to pull, so we can't reliably use that index. also I was using the phrase "this is the fastest query" to check that I had updated all the functions lol

internal/auth/oidc/rewrapping.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+func init() {
+	kms.RegisterTableRewrapFn(defaultAuthMethodTableName, authMethodRewrapFn)
+}
+
+func authMethodRewrapFn(ctx context.Context, dataKeyVersionId, scopeId string, reader db.Reader, writer db.Writer, kmsRepo *kms.Kms) error {
+	const op = "oidc.authMethodRewrapFn"
+	var authMethods []*AuthMethod
+	// there are indexes on (scope id, <other>), so we can query on scope and refine via key id. this is the fastest query
```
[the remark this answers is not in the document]
DanielleMiu: dupe

internal/server/query.go
DanielleMiu: so the way this was explained to me is that adding the the comparison on the join yields a better query plan, but I double checked that the query plans are the same, so I'll update the query and add the indents.

internal/server/rewrapping.go
DanielleMiu: I completely agree, but I don't think that is in the context of this ticket or this PR

internal/server/rewrapping.go
DanielleMiu: wouldn't deferring it wait until the function exits? seems like we would want to close this before we move onto the rewrapping loop

internal/server/rewrapping.go
johanbrandhorst: Could we pick it up as a quick follow up to this PR if this PR specifically? I think there's a lot of value in making sure we don't just leave it for the next person to be confused.

internal/session/credential.go
DanielleMiu: agree it would be best but again not sure it belongs in the context of this ticket/pr

internal/server/rewrapping.go
johanbrandhorst: It would indeed, but it's only important to call in the error cases, in the cases where the loop exists and rows.Err doesn't error, it's not important to have it called at any particular time.

internal/session/query.go
DanielleMiu: I will check to see if this can be done with a db.withwhere

johanbrandhorst: I don't think that's what I mean, If we end up adding primary keys, we shouldn't need any `WithWhere`. We should just try to make this table easier to use with our types in general.

internal/server/rewrapping.go
DanielleMiu: would depend on what the close func does. if we don't call it do we leave the connection open the whole time? might take up (admittedly, a small amount of) db resources deferring it

johanbrandhorst: The behavior of the Close function is well documented, which is why I can state this confidently 😛: https://pkg.go.dev/database/sql#Rows.Close.

internal/session/query.go
DanielleMiu: yeah, if we end up adding a pk to this table then it would do this automatically

internal/server/rewrapping.go
DanielleMiu: super cool, I'll update it then

internal/session/query.go
DanielleMiu: doesn't look like updates can be performed via gorm using `withWhere` without a primary key

internal/server/rewrapping.go
DanielleMiu: being moved to separate task

internal/proto/controller/storage/servers/store/v1/root_certificate.proto
```diff
@@ -34,7 +34,7 @@ message RootCertificate {
   storage.timestamp.v1.Timestamp not_valid_after = 40;
 
   // The public key associated with this certificate
-  // @inject_tag: `gorm:"not_null"`
+  // @inject_tag: `gorm:"primary_key"`
```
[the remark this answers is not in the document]
DanielleMiu: [boundary#2542](https://github.com/hashicorp/boundary/pull/2542)

johanbrandhorst (approved):
LGTM pending discussion on updating hmacs without updating underlying secret

johanbrandhorst (approved):

PR #2534, opened 2022-10-11
Add grant parsing fuzz test

johanbrandhorst:
Adds fuzz tests for the grant parser. Running this I found two issues:

* We were not consistently handling invalid unicode. I fixed this by always replacing invalid unicode in inputs with the unicode replacement handler. This is what marshaling to JSON already does.
* We were not handling empty output fields consistently. `output_fields=,` was allowed while `output_fields=` was not. Now the latter is allowed and the former is equal to the latter.

### DIFF
--- a/internal/perms/grants.go
+++ b/internal/perms/grants.go
@@ -47,7 +47,7 @@ type Grant struct {
 	actions map[action.Type]bool
 
 	// The set of output fields granted
-	OutputFields OutputFieldsMap
+	OutputFields *OutputFields
 
 	// This is used as a temporary staging area before validating permissions to
 	// allow the same validation code across grant string formats
@@ -105,11 +105,12 @@ func (g Grant) clone() *Grant {
 			ret.actions[action] = true
 		}
 	}
-	if g.OutputFields != nil {
-		ret.OutputFields = make(OutputFieldsMap, len(g.OutputFields))
-		for k, v := range g.OutputFields {
-			ret.OutputFields[k] = v
+	if outFields := g.OutputFields.Fields(); outFields != nil {
+		fieldsToAdd := make([]string, 0, len(outFields))
+		for _, v := range outFields {
+			fieldsToAdd = append(fieldsToAdd, v)
 		}
+		ret.OutputFields = ret.OutputFields.AddFields(fieldsToAdd)
 	}
 	return ret
 }
@@ -135,8 +136,8 @@ func (g Grant) CanonicalString() string {
 		builder = append(builder, fmt.Sprintf("actions=%s", strings.Join(actions, ",")))
 	}
 
-	if len(g.OutputFields) > 0 {
-		builder = append(builder, fmt.Sprintf("output_fields=%s", strings.Join(g.OutputFields.Fields(), ",")))
+	if outFields := g.OutputFields.Fields(); outFields != nil {
+		builder = append(builder, fmt.Sprintf("output_fields=%s", strings.Join(outFields, ",")))
 	}
 
 	return strings.Join(builder, ";")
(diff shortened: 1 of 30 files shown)


johanbrandhorst:
I added some more roundtrip tests, and it discovered the following:

```
grants_test.go:788: grant JSON roundtrip failed, input "output_fields=\xf1", output "output_fields=�"
```

I don't know if we care about this, but maybe we want to do unicode normalization as part of our `CanonicalString` method?

johanbrandhorst:
Went down a bit of a rabbit hole, actually this is not unicode normalization, it's about replacing invalid unicode in inputs with the unicode replacement character, which is what JSON does automatically.

johanbrandhorst:
Fuzzing did discover a fun one: `output_fields=,` was allowed.

jefferai:
> Fuzzing did discover a fun one: output_fields=, was allowed.

Correct! This is allowed!

johanbrandhorst:
> > Fuzzing did discover a fun one: output_fields=, was allowed.
> 
> Correct! This is allowed!

Interesting! The error happened because this input creates a canonical string of `output_fields=`, which when roundtripped back into the parser is _not_ allowed. Should the canonical string of `output_fields=,` be `output_fields=b`?

johanbrandhorst:
I removed the "fix" for `output_fields=,` and started allowing `output_fields=`, with the same behavior.

jefferai:
Yes, I think in general any trailing comma on a string should be stripped for parsing and for the canonical representation

jefferai:
Meant to follow up on this: yes, `output_fields=` means "no output fields allowed". While it doesn't have a lot of use cases -- and if it's confusing we could think about taking it out -- the idea is for things like services where you want to perform an action periodically and the only thing you need or want the service to know about is the HTTP status code, and don't want to divulge any other data.

johanbrandhorst:
> Meant to follow up on this: yes, `output_fields=` means "no output fields allowed". While it doesn't have a lot of use cases -- and if it's confusing we could think about taking it out -- the idea is for things like services where you want to perform an action periodically and the only thing you need or want the service to know about is the HTTP status code, and don't want to divulge any other data.

OK so it sounds like the only real change here (accepting `output_fields=`) is what we want, should we merge this then?

internal/perms/grants.go
```diff
@@ -405,7 +407,7 @@ func Parse(scopeId, grantString string, opt ...Option) (Grant, error) {
 		}
 		// Set but empty output fields...
 		if grant.OutputFields != nil && len(grant.OutputFields) == 0 {
-			return Grant{}, errors.NewDeprecated(errors.InvalidParameter, op, "parsed grant string has output_fields set but empty")
+			grant.OutputFields = OutputFieldsMap{"": true}
```
jefferai, line 410: What's the purpose of setting `"": true` here?

johanbrandhorst: This error was the source of the roundtrip inconsistency I believe. Adding the empty string here lets us reproduce the `output_fields=` in the string representation (see `CanonicalString`).

internal/perms/output_fields.go
```diff
@@ -6,68 +6,105 @@ import (
...
 
-// Fields returns an alphabetical string slice of the fields in the map
-func (o OutputFieldsMap) Fields() (ret []string) {
-	if o == nil {
+// Fields returns an alphabetical string slice of the fields in the map. The
+// return value will be nil if fields are unset (e.g. we'd use the defaults in
+// SelfOrDefaults), and non-nil but empty if no fields are allowed. It is safe
+// to call this on a nil object; it will return a nil slice.
+func (o *OutputFields) Fields() (ret []string) {
+	if o == nil || o.fields == nil {
```
johanbrandhorst, line 70: I can get behind using the difference between a `nil` map and an empty slice internally, but should we really expose this as our interface? It's almost never an important distinction in Go. I'd like this to be more explicit, maybe
```go
func (o *OutputFields) Fields() (bool, []string) {
```
and have the `bool` be `false` if the fields are unset?

jefferai: Sure.

PR #2535, opened 2022-10-11
Rewrite interface{} to any

johanbrandhorst:
Go 1.18 introduced generics, but it also introduced an alias for `interface{}` called `any`. It's shorter, more descriptive and already adopted [across the standard library](https://github.com/golang/go/issues/49884). We should probably also use it.

These changes were created using

```
$ gofmt -w -r 'interface{} -> any' . && make gen
```


johanbrandhorst:
This has 4 approvals and zero rejections so I'm going to merge it :)

PR #2538, opened 2022-10-13
feat(dbtest): Allow host to be configurable

tmessi:
Normally the tests are run on a host machine with a docker container for
the database. The docker container should be accessible via 127.0.0.1.
However there might be cases where the test execution is run in a docker
container, or the postgres server is configured to use a different
docker network, in which case the container would be accessible via some
other dns name or ip. In these cases, users can now set the TEST_DB_HOST
environment variable so the tests can connect to the postgres server.

### DIFF
--- a/testing/dbtest/template.go
+++ b/testing/dbtest/template.go
@@ -16,10 +16,10 @@ import (
 	_ "github.com/jackc/pgx/v4/stdlib"
 )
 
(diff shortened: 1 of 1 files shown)


johanbrandhorst (approved):

ddebko (approved):
LGTM

PR #2539, opened 2022-10-14
Add scopeId to RewrapFn

johanbrandhorst:
This makes it much easier for rewrap functions to look up the correct wrapper and limit the number of rows searched.


DanielleMiu (approved):
noice

PR #2542, opened 2022-10-18
Update Root Cert Proto Definition

DanielleMiu:
update root cert proto definition to properly identify public_key as the table primary key

the [pk already exists](https://github.com/hashicorp/boundary/blob/main/internal/db/schema/migrations/oss/postgres/34/03_worker_authentication.up.sql#L53) in sql, just not in the proto file


johanbrandhorst (approved):
CC @irenarindos

irenarindos (approved):

PR #2555, opened 2022-10-21
Rewrapping Registered Tables Test

DanielleMiu:
- [added new rewrapping test to check that all encrypted tables have assigned rewrapping functions](https://github.com/hashicorp/boundary/pull/2555/files#diff-deb54d5881ef6b4457f12626325966134e5902bb01da879c15278be1ec0be50b)
- [added ListDataKeyVersionReferencers function to kms](https://github.com/hashicorp/boundary/pull/2555/files#diff-65fa95abdd78accdc6b1f3ac98751fe4700c8abc1ee696b8d6ae7e5403af0831)
- added new rewrapping functions on [`credential_static_json_credential`](https://github.com/hashicorp/boundary/pull/2555/files#diff-231e7485c6e362d7d68a2bd4c5ba36aa93f2b77923cb51c656b1ecbb2d728f01R102) ([test](https://github.com/hashicorp/boundary/pull/2555/files#diff-af94db4dbe7299f4cf0e078eab23a2b9a10abf0007a9d1926589483efc4d2543)) and [`worker_auth_server_led_activation_token`](https://github.com/hashicorp/boundary/pull/2555/files#diff-fea1b2d017b7f403426fa6d599f13ff1887d46be49590033ab4a9edd5d5ca7d3) ([test](https://github.com/hashicorp/boundary/pull/2555/files#diff-43bdf027bdb2e76c372553b3c3c9e94924f7c389a70c3570e88c6197c77d27bf))
- [modified existing rewrap functions on `credential_static_username_password_credential` and `credential_static_ssh_private_key_credential` to use faster retrieval queries](https://github.com/hashicorp/boundary/pull/2555/files#diff-fe0cea8b0c88467266daedbb720cc258f14eff2392a485224d62b3798b75e42d) (found while making json cred rewrapfn)

## psql output for `credential_static_store` showing references for credential lookups:
```
boundary_test_gfmaerwwxwhtgcmv=# \d+ credential_static_store
                                Table "public.credential_static_store"
   Column    |      Type      | Collation | Nullable | Default | Storage  | Stats target | Description 
-------------+----------------+-----------+----------+---------+----------+--------------+-------------
 public_id   | wt_public_id   |           | not null |         | extended |              | 
 project_id  | wt_scope_id    |           | not null |         | extended |              | 
 name        | wt_name        |           |          |         | extended |              | 
 description | wt_description |           |          |         | extended |              | 
 create_time | wt_timestamp   |           |          |         | plain    |              | 
 update_time | wt_timestamp   |           |          |         | plain    |              | 
 version     | wt_version     |           |          |         | plain    |              | 
Indexes:
    "credential_static_store_pkey" PRIMARY KEY, btree (public_id)
    "credential_static_store_project_id_name_uq" UNIQUE CONSTRAINT, btree (project_id, name)
Foreign-key constraints:
    "credential_store_fkey" FOREIGN KEY (project_id, public_id) REFERENCES credential_store(project_id, public_id) ON UPDATE CASCADE ON DELETE CASCADE
Referenced by:
    TABLE "credential_static_username_password_credential" CONSTRAINT "credential_static_store_fkey" FOREIGN KEY (store_id) REFERENCES credential_static_store(public_id) ON UPDATE CASCADE ON DELETE CASCADE
    TABLE "credential_static_ssh_private_key_credential" CONSTRAINT "credential_static_store_fkey" FOREIGN KEY (store_id) REFERENCES credential_static_store(public_id) ON UPDATE CASCADE ON DELETE CASCADE
    TABLE "credential_static_json_credential" CONSTRAINT "credential_static_store_fkey" FOREIGN KEY (store_id) REFERENCES credential_static_store(public_id) ON UPDATE CASCADE ON DELETE CASCADE
Triggers:
    default_create_time_column BEFORE INSERT ON credential_static_store FOR EACH ROW EXECUTE PROCEDURE default_create_time()
    delete_credential_store_subtype AFTER DELETE ON credential_static_store FOR EACH ROW EXECUTE PROCEDURE delete_credential_store_subtype()
    immutable_columns BEFORE UPDATE ON credential_static_store FOR EACH ROW EXECUTE PROCEDURE immutable_columns('public_id', 'project_id', 'create_time')
    insert_credential_store_subtype BEFORE INSERT ON credential_static_store FOR EACH ROW EXECUTE PROCEDURE insert_credential_store_subtype()
    update_time_column BEFORE UPDATE ON credential_static_store FOR EACH ROW EXECUTE PROCEDURE update_time_column()
    update_version_column AFTER UPDATE ON credential_static_store FOR EACH ROW EXECUTE PROCEDURE update_version_column()
```

## Query plan(s) for old vs new credential lookup queries
new plan:
```
                                                                          QUERY PLAN                                                                          
--------------------------------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=4.31..21.86 rows=1 width=280)
   ->  Bitmap Heap Scan on credential_static_store store  (cost=4.16..9.50 rows=2 width=32)
         Recheck Cond: ((project_id)::text = ''::text)
         ->  Bitmap Index Scan on credential_static_store_project_id_name_uq  (cost=0.00..4.16 rows=2 width=0)
               Index Cond: ((project_id)::text = ''::text)
   ->  Index Scan using credential_static_json_credential_store_id_public_id_uq on credential_static_json_credential json  (cost=0.15..6.17 rows=1 width=280)
         Index Cond: ((store_id)::text = (store.public_id)::text)
         Filter: ((key_id)::text = ''::text)
(8 rows)
```
old plan:
```
                                     QUERY PLAN                                     
------------------------------------------------------------------------------------
 Seq Scan on credential_static_json_credential  (cost=0.00..13.25 rows=1 width=280)
   Filter: ((key_id)::text = ''::text)
(2 rows)
```

### DIFF
--- a/internal/daemon/controller/rewrapping_test.go
+++ b/internal/daemon/controller/rewrapping_test.go
@@ -1,0 +1,58 @@
+package controller
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/kms"
+	"github.com/stretchr/testify/assert"
+)
+
+func SliceContains(slice []string, item string) bool {
+	for _, i := range slice {
+		if item == i {
+			return true
+		}
+	}
+	return false
+}
+
+func Test_AllRewrapTablesRegistered(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+	conn, _ := db.TestSetup(t, "postgres")
+	extWrapper := db.TestWrapper(t)
+	kmsCache := kms.TestKms(t, conn, extWrapper)
+
+	tables, err := kmsCache.ListDataKeyVersionReferencers(ctx)
+	assert.NoError(t, err)
+
+	var filteredTables []string
+	// there are some tables we know won't be rewrapped
+	for _, table := range tables {
+		// oplog entries cannot be rewrapped at this time
+		if table == "oplog_entry" {
+			continue
+		}
+		// kms_data_key_version_destruction_job does not contain encrypte data, only a reference to the key versions
+		if table == "kms_data_key_version_destruction_job" {
+			continue
+		}
+
+		// all other tables should be registered
+		filteredTables = append(filteredTables, table)
+	}
+
+	registeredTables := kms.ListTablesSupportingRewrap()
+
+	// first things first, make sure we have the same number of expected values.
+	assert.Equal(t, len(filteredTables), len(registeredTables), "the number of registered rewrap functions does not match the number of tables that contain encrypted data.")
+
+	// then make sure that all of the tables that need rewrap funcs registered actually have them.
+	for _, table := range filteredTables {
+		// do this in a loop so that we can explicitly call out which tables are missing, if any are.
+		assert.True(t, SliceContains(registeredTables, table), fmt.Sprintf("%s doesn't have a rewrap function registered!", table))
+	}
+}
--- a/internal/daemon/controller/rewrapping_test.go
+++ b/internal/daemon/controller/rewrapping_test.go
@@ -2,22 +2,14 @@ package controller
 
 import (
 	"context"
-	"fmt"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/hashicorp/boundary/internal/db"
 	"github.com/hashicorp/boundary/internal/kms"
 	"github.com/stretchr/testify/assert"
 )
-
-func SliceContains(slice []string, item string) bool {
-	for _, i := range slice {
-		if item == i {
-			return true
-		}
-	}
-	return false
-}
 
 func Test_AllRewrapTablesRegistered(t *testing.T) {
 	t.Parallel()
@@ -36,7 +28,7 @@ func Test_AllRewrapTablesRegistered(t *testing.T) {
 		if table == "oplog_entry" {
 			continue
 		}
-		// kms_data_key_version_destruction_job does not contain encrypte data, only a reference to the key versions
+		// kms_data_key_version_destruction_job does not contain encrypted data, only a reference to the key versions
 		if table == "kms_data_key_version_destruction_job" {
 			continue
 		}
@@ -45,14 +37,5 @@ func Test_AllRewrapTablesRegistered(t *testing.T) {
 		filteredTables = append(filteredTables, table)
 	}
 
-	registeredTables := kms.ListTablesSupportingRewrap()
-
-	// first things first, make sure we have the same number of expected values.
-	assert.Equal(t, len(filteredTables), len(registeredTables), "the number of registered rewrap functions does not match the number of tables that contain encrypted data.")
-
-	// then make sure that all of the tables that need rewrap funcs registered actually have them.
-	for _, table := range filteredTables {
-		// do this in a loop so that we can explicitly call out which tables are missing, if any are.
-		assert.True(t, SliceContains(registeredTables, table), fmt.Sprintf("%s doesn't have a rewrap function registered!", table))
-	}
+	assert.Empty(t, cmp.Diff(filteredTables, kms.ListTablesSupportingRewrap(), cmpopts.SortSlices(func(i, j string) bool { return i < j })), "At least one table referencing a data key does not have a rewrapping function registered")
 }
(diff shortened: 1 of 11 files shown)


internal/daemon/controller/rewrapping_test.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+func SliceContains(slice []string, item string) bool {
+	for _, i := range slice {
+		if item == i {
+			return true
+		}
+	}
+	return false
+}
```
johanbrandhorst, line 20: Nit: you can use https://pkg.go.dev/golang.org/x/exp/slices#Index and check that `i > 0`.

johanbrandhorst: :doh: there's also just https://pkg.go.dev/golang.org/x/exp/slices#Contains

internal/daemon/controller/rewrapping_test.go
```diff
@@ -0,0 +1,58 @@
...
+	registeredTables := kms.ListTablesSupportingRewrap()
+
+	// first things first, make sure we have the same number of expected values.
+	assert.Equal(t, len(filteredTables), len(registeredTables), "the number of registered rewrap functions does not match the number of tables that contain encrypted data.")
+
+	// then make sure that all of the tables that need rewrap funcs registered actually have them.
+	for _, table := range filteredTables {
+		// do this in a loop so that we can explicitly call out which tables are missing, if any are.
+		assert.True(t, SliceContains(registeredTables, table), fmt.Sprintf("%s doesn't have a rewrap function registered!", table))
+	}
```
johanbrandhorst, line 57: Just use `cmp.Diff`, it'll do what you want:
```go
require.Empty(t, cmp.Diff(tableNames, registeredTableNames, cmpopts.SortSlices(func(i, j string) bool { return i < j })), "At least one table referencing a data key does not have a rewrapping function registered")
```

internal/db/schema/migrations/oss/postgres/56/05_mutable_ciphertext_columns.up.sql
```diff
@@ -32,4 +32,11 @@ begin;
   create trigger immutable_columns before update on worker_auth_authorized
     for each row execute function immutable_columns('worker_key_identifier', 'worker_id', 'worker_signing_pub_key', 'worker_encryption_pub_key', 'nonce', 'create_time');
 
+
+  -- With the addition of the registered tables test, we found one more table that needs its encrypted column made mutable
```
johanbrandhorst, line 36: 😂. These comments are fun, but they are also a bit distracting tbh, maybe we should just remove them?

internal/daemon/controller/rewrapping_test.go
DanielleMiu: yknow. I actually looked for something like this, and didn't find anything. sighhhhhhhhh

internal/kms/kms.go
```diff
@@ -269,6 +269,26 @@ func (k *Kms) RotateKeys(ctx context.Context, scopeId string, opt ...Option) err
...
+// ListDataKeyVersionReferencers will list the names of all tables
+// referencing the private_id column of the data key version table.
+// Supported options:
+//   - WithTx
+func (k *Kms) ListDataKeyVersionReferencers(ctx context.Context, opt ...Option) ([]string, error) {
+	const op = "kms.(Kms).ListDataKeyVersionReferencers"
+
+	opts := getOpts(opt...)
+	kmsOpts := []wrappingKms.Option{
+		// underlying key version refs function supports WithReaderWriter option, but the interfaces don't match and can't be passed
```
johanbrandhorst, line 281: Minor point: can't we convert the option like we do in RotateKeys above? I'd argue we don't need to support any options at all here until we actually have a need for them, so feel free to just remove this too.

internal/daemon/controller/rewrapping_test.go
DanielleMiu: I remember seeing this in other tests and I intentionally didn't use this because if one is missing (likely not at the end of the list) it'll say that more than just the one missing is incorrect. doing it this way will explicitly state which function is missing and needs to be implemented

johanbrandhorst: Alright, fair enough 👍🏻

DanielleMiu: this is seemingly no longer the case, so I can switch to cmp.diff

internal/daemon/controller/rewrapping_test.go
DanielleMiu: moot point with cmp.diff

johanbrandhorst (approved):
Nice!

PR #2556, opened 2022-10-22
Ensure all secret updates include key ID

johanbrandhorst:
When updating a secret, it is paramount that the key ID is also updated, as it is the only means we have of ensuring that we only destroy keys once there is no more data associated with the key.

I used three different methods for finding instances of secret updates:

1. Audit all uses of `(db.Writer).Update` manually for updates of secrets.
2. (As a backup) audit all uses of `kms.GetWrapper(ctx, <scopeId>, kms.KeyPurpose)` for all key purposes.
3. (As a backup) search for all instances of `update xxx set` to find any extra SQL literal statements.

The test I've added shows a scenario where we would end up with irrecoverable data without these fixes.

### DIFF
--- a/internal/tests/api/credentials/credentials_test.go
+++ b/internal/tests/api/credentials/credentials_test.go
@@ -4,13 +4,23 @@ import (
 	"fmt"
 	"net/http"
 	"testing"
+	"time"
 
 	"github.com/hashicorp/boundary/api"
 	"github.com/hashicorp/boundary/api/credentials"
 	"github.com/hashicorp/boundary/api/credentialstores"
+	"github.com/hashicorp/boundary/api/hostcatalogs"
+	"github.com/hashicorp/boundary/api/hosts"
+	"github.com/hashicorp/boundary/api/hostsets"
+	"github.com/hashicorp/boundary/api/scopes"
+	"github.com/hashicorp/boundary/api/targets"
 	"github.com/hashicorp/boundary/internal/credential"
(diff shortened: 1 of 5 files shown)


johanbrandhorst:
The tests are failing because of a problem with the key_id stored with each session row. I dropped a quick explanation in Slack, but I plan to prepare a separate PR with a fix.

internal/tests/api/credentials/credentials_test.go
```diff
@@ -384,3 +394,112 @@ func TestErrors(t *testing.T) {
...
+			destroyKeyVersion = key.Versions[1]
+			break
+		}
+	}
+
+	// Destroy the older key version
+	result, err := scopesClient.DestroyKeyVersion(ctx, proj.PublicId, destroyKeyVersion.Id)
+	require.NoError(err)
+	// Should start asynchronous rewrapping of the encrypted JSON credential
+	assert.Equal("pending", result.State)
```
DanielleMiu, line 487: what is the purpose of destroying the key version here? could we not just pull this row from the db and ensure that the key id has been updated? perhaps attempt decrypting with that new keyid?

johanbrandhorst: The reason we're destroying the key version here is because I wanted to create a test case that could realistically be reproduced by a user. This is an integration test, the test oracle for this isn't "is the DB row updated?" it's "does the product behave correctly under this specific scenario using its public API endpoints?". This way, this test doesn't have to know anything about our internal implementation details, just like the user doesn't know about it.

PR #2557, opened 2022-10-24
add new encrypt/decrypt funcs and made necessary proto changes

DanielleMiu:
- [add encrypt/decrypt functions for worker auth and root cert](https://github.com/hashicorp/boundary/commit/7e78e842d2d00ebb9e23c77d7a36983ac7317b08)
- [add parameter checking to rewrap functions](https://github.com/hashicorp/boundary/pull/2557/commits/ebb7ddb7b26c5c7e9a01551c1a1ea4ce35e443ed)


internal/proto/controller/storage/servers/store/v1/root_certificate.proto
```diff
@@ -37,10 +37,15 @@ message RootCertificate {
...
+  // in the database.
+  // @inject_tag: `gorm:"-"` wrapping:"pt,private_key"`
+  bytes private_key = 60;
+
   // The private key associated with this certificate
   // This is a ciphertext field
-  // @inject_tag: `gorm:"not_null"`
-  bytes private_key = 60;
+  // @inject_tag: `gorm:"column:private_key;not_null"` wrapping:"ct,private_key"`
+  bytes ct_private_key = 65;
```
johanbrandhorst, line 48: Nit: Just make it 61, I think. AFAIK the splitting of these numbers is to allow us to add more related fields, and this is very very related to the above.

internal/server/repository_workerauth.go
```diff
@@ -134,7 +134,7 @@ func StoreNodeInformationTx(ctx context.Context, writer db.Writer, databaseWrapp
 	if err != nil {
 		return errors.Wrap(ctx, err, op)
 	}
-	nodeAuth.ControllerEncryptionPrivKey, err = encrypt(ctx, node.ServerEncryptionPrivateKeyBytes, databaseWrapper)
+	nodeAuth.CtControllerEncryptionPrivKey, err = encrypt(ctx, node.ServerEncryptionPrivateKeyBytes, databaseWrapper)
```
johanbrandhorst, line 137: Should this not be a `nodeAuth.encrypt`?

internal/server/repository_workerauth.go
```diff
@@ -236,12 +236,9 @@ func (r *WorkerAuthRepositoryStorage) convertRootCertificate(ctx context.Context
...
 		return nil, errors.Wrap(ctx, err, op)
 	}
-	rootCert.KeyId, err = databaseWrapper.KeyId(ctx)
-	if err != nil {
-		return nil, errors.Wrap(ctx, err, op)
-	}
-	rootCert.PrivateKey, err = encrypt(ctx, cert.PrivateKeyPkcs8, databaseWrapper)
-	if err != nil {
+
+	rootCert.PrivateKey = cert.PrivateKeyPkcs8
```
johanbrandhorst, line 240: What do you think about moving this above the GetWrapper call with the other assignments?

internal/server/repository_workerauth.go
```diff
@@ -398,7 +395,7 @@ func (r *WorkerAuthRepositoryStorage) loadNodeInformation(ctx context.Context, n
 	if err != nil {
 		return errors.Wrap(ctx, err, op)
 	}
-	node.ServerEncryptionPrivateKeyBytes, err = decrypt(ctx, authorizedWorker.ControllerEncryptionPrivKey, databaseWrapper)
+	node.ServerEncryptionPrivateKeyBytes, err = decrypt(ctx, authorizedWorker.CtControllerEncryptionPrivKey, databaseWrapper)
```
johanbrandhorst, line 398: Should this be `node.decrypt`?

internal/server/repository_workerauth.go
```diff
@@ -538,7 +535,7 @@ func (r *WorkerAuthRepositoryStorage) findPriorEncryptionKey(ctx context.Context
 	// Create the EncryptionKey using the prior worker auth record
 	priorKey := &types.EncryptionKey{
 		KeyId:           workerAuthSet.Previous.WorkerKeyIdentifier,
-		PrivateKeyPkcs8: workerAuthSet.Previous.ControllerEncryptionPrivKey,
+		PrivateKeyPkcs8: workerAuthSet.Previous.CtControllerEncryptionPrivKey,
```
johanbrandhorst, line 538: Should this be using the encrypted value?

internal/server/repository_workerauth.go
```diff
@@ -599,7 +596,7 @@ func (r *WorkerAuthRepositoryStorage) loadRootCertificates(ctx context.Context,
 		if err != nil {
 			return errors.Wrap(ctx, err, op)
 		}
-		rootCert.PrivateKeyPkcs8, err = decrypt(ctx, rootCertificate.PrivateKey, databaseWrapper)
+		rootCert.PrivateKeyPkcs8, err = decrypt(ctx, rootCertificate.CtPrivateKey, databaseWrapper)
```
johanbrandhorst, line 599: Should this be `rootCert.decrypt`?

internal/server/rewrapping_test.go
```diff
@@ -45,12 +45,12 @@ func TestRewrap_workerAuthCertRewrapFn(t *testing.T) {
 	assert.NoError(t, err)
 
 	// decrypt with the new key version and check to make sure things match
-	decryptedGotPrivKey, err := decrypt(ctx, got.GetPrivateKey(), kmsWrapper2)
+	decryptedGotPrivKey, err := decrypt(ctx, got.GetCtPrivateKey(), kmsWrapper2)
```
johanbrandhorst, line 48: Should we eventually be able to remove these `encrypt` `decrypt`methods altogether?

internal/server/root_certificate.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 
 	"github.com/hashicorp/boundary/internal/errors"
+	wrapping "github.com/hashicorp/go-kms-wrapping/v2"
+	"github.com/hashicorp/go-kms-wrapping/v2/extras/structwrapping"
 
 	timestamp "github.com/hashicorp/boundary/internal/db/timestamp"
```
johanbrandhorst, line 11: Not your fault but I think we should merge these import groups, we generally only have two.

internal/server/worker_auth.go
```diff
@@ -57,7 +83,7 @@ func newWorkerAuth(ctx context.Context, workerKeyIdentifier, workerId string, op
 		l.WorkerEncryptionPubKey = opts.withWorkerKeys.workerEncryptionPubKey
 	}
 	if &opts.withControllerEncryptionPrivateKey != nil {
-		l.ControllerEncryptionPrivKey = opts.withControllerEncryptionPrivateKey
+		l.CtControllerEncryptionPrivKey = opts.withControllerEncryptionPrivateKey
```
johanbrandhorst, line 86: Should this be the ciphertext field?

internal/server/worker_auth.go
```diff
@@ -96,7 +122,7 @@ func (w *WorkerAuth) ValidateNewWorkerAuth(ctx context.Context) error {
 	if w.WorkerEncryptionPubKey == nil {
 		return errors.New(ctx, errors.InvalidParameter, op, "missing WorkerEncryptionPubKey")
 	}
-	if w.ControllerEncryptionPrivKey == nil {
+	if w.CtControllerEncryptionPrivKey == nil {
```
johanbrandhorst, line 125: Ditto

internal/server/repository_workerauth.go
DanielleMiu: I'm not 100% sure, but from what I could tell, in both structs, the field was generally only used as a ciphertext field. may need to do a little more digging to determine for sure

DanielleMiu: in this case, there is no decryption happening in `FindWorkerAuthByWorkerId`, the function pulling the worker auths, so the value would be encrypted from the db

internal/server/rewrapping_test.go
DanielleMiu: yes, but I was hesitant to do so immediately so that we're not using the encrypt/decrypt functions to test to make sure the encrypt/decrypt functions work. leaving the test with these funcs gives us an end to end test to make sure they're working as intended

internal/server/worker_auth.go
DanielleMiu: this one is definitely the case, as this is the validate function, it's called before storing the value in the database. since we're not storing the plaintext field, it doesn't need to be checked!

DanielleMiu: (I will, however, update the error string, as it's somewhat misleading as-is)

internal/server/rewrapping_test.go
johanbrandhorst: hm, can we define the functions in the test file then so they're inaccessible from the other package?

internal/server/worker_auth.go
johanbrandhorst: Should we move it to one of the test files?

DanielleMiu: not sure that's necessary, it is a `new` function after all, just that this is how its used. maybe worth while adding a comment that it assigns to the ciphertext field

DanielleMiu: the more I thought about this, the less I liked leaving this as-is. leaving as is makes it so that (if it's ever used outside of tests) the default encrypt/decrypt functions are required to be used to create a new worker auth. I've updated new worker auth to assign to plaintext and removed the `withKeyId` option as well as updated the tests and test helpers that use this func to encrypt the worker auth afterwards to correctly assign the keyid and ciphertext fields

internal/server/rewrapping_test.go
DanielleMiu: I think this depends on if we feel encrypt/decrypt func tests are necessary. I don't see them anywhere else in the codebase (maybe because its mostly handled by structwrapping?) if that's the case we can just switch this to use the worker auth decrypt and potentially remove the other funcs

johanbrandhorst: I would like to make it so that using the `.encrypt` and `.decrypt` methods is the obvious (ideally only) way to do this operation. Leaving these functions behind may mean someone uses them in the future, which means we'll have multiple ways of doing a single operation, which is error prone 😬.

johanbrandhorst (approved):
Nice!

johanbrandhorst (approved):

PR #2565, opened 2022-10-26
Add some common account bits and some getOpts -> GetOpts changes

jefferai:
This is a precursor to another upcoming PR that will make use of these changes, but it makes that PR more manageable.

### DIFF
--- a/internal/authtoken/options.go
+++ b/internal/authtoken/options.go
@@ -3,7 +3,9 @@ package authtoken
 import (
 	"time"
 
+	"github.com/hashicorp/boundary/internal/auth/password"
 	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/iam"
 )
 
 var (
@@ -31,6 +33,8 @@ type options struct {
 	withLimit                    int
 	withStatus                   Status
 	withPublicId                 string
+	withPasswordOptions          []password.Option
+	withIamOptions               []iam.Option
 }
 
 func getDefaultOptions() options {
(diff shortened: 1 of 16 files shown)


internal/authtoken/options.go
```diff
@@ -91,3 +95,21 @@ func WithPublicId(id string) Option {
...
+}
+
+// WithIamOptions allows passing through options for the IAM package. This is
+// useful for things like tests where you may use testing helper functions in
+// this package but they also create IAM resources.
+func WithIamOptions(with ...iam.Option) Option {
+	return func(o *options) {
+		o.withIamOptions = with
+	}
+}
```
johanbrandhorst, line 115: Should these go in `testing.go`?

internal/daemon/controller/handlers/targets/registry.go
```diff
@@ -39,7 +39,7 @@ type registryEntry struct {
 }
 
 type registry struct {
-	sync.Map
+	*sync.Map
```
johanbrandhorst, line 42: 😱 was this buggy the whole time?

> A Map must not be copied after first use

https://pkg.go.dev/sync#Map

johanbrandhorst (approved):

internal/daemon/controller/handlers/targets/registry.go
jefferai: Yep! It was!

internal/authtoken/options.go
jefferai: The reason I didn't -- and maybe this is wrong -- is that they do go into normal non-testing functions, even if they're mostly used for tests. So I didn't want to have e.g. `NewAccount(..., ..., ..., password.TestOptions...)`. If these were only ever used in actual testing functions I would definitely put them in testing.go and label them e.g. `WithTestIamOption` and `WithTestPasswordOptions`.

WDYT?

jefferai: We can always move it later; resolving for now.

PR #2573, opened 2022-10-26
Disconnect PKI workers from upstreams when they are no longer authorized

talanknight:
With this PR the worker and controller intermittently check if the downstream pki workers connected to them are still authorized in the boundary system. If a connected pki worker is found to no longer be authorized the underlying connection is closed.

For controllers this happens every 3 seconds, for workers this happens in every status update request.

### DIFF
--- a/globals/globals.go
+++ b/globals/globals.go
@@ -6,6 +6,10 @@ const (
 	TcpProxyV1     = "boundary-tcp-proxy-v1"
 	ServiceTokenV1 = "s1"
 	SessionPrefix  = "s_"
+
+	AnyAuthenticatedUserId = "u_auth"
+	AnonymousUserId        = "u_anon"
+	RecoveryUserId         = "u_recovery"
 )
 
 type (
--- a/internal/auth/additional_verification_test.go
+++ b/internal/auth/additional_verification_test.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/hashicorp/boundary/api/authmethods"
 	"github.com/hashicorp/boundary/api/authtokens"
+	"github.com/hashicorp/boundary/globals"
 	"github.com/hashicorp/boundary/internal/auth/password"
 	"github.com/hashicorp/boundary/internal/authtoken"
 	"github.com/hashicorp/boundary/internal/daemon/controller"
@@ -131,7 +132,7 @@ func TestRecursiveListingDifferentOutputFields(t *testing.T) {
 	// will expect the defaults.
 	globalRole := iam.TestRole(t, conn, scope.Global.String())
 	iam.TestUserRole(t, conn, globalRole.PublicId, token.UserId)
-	iam.TestUserRole(t, conn, globalRole.PublicId, auth.AnonymousUserId)
+	iam.TestUserRole(t, conn, globalRole.PublicId, globals.AnonymousUserId)
 	iam.TestRoleGrant(t, conn, globalRole.PublicId, "id=*;type=auth-method;actions=list,no-op")
 
 	// Create some users at the org level, and some role grants for them
--- a/internal/auth/oidc/service_callback_test.go
+++ b/internal/auth/oidc/service_callback_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/hashicorp/boundary/globals"
 	"github.com/hashicorp/boundary/internal/auth/oidc/store"
 	authStore "github.com/hashicorp/boundary/internal/auth/store"
 	"github.com/hashicorp/boundary/internal/authtoken"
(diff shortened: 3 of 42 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
   ([PR](https://github.com/hashicorp/boundary/pull/2544))
 * workers: Fixed a panic that can happen in certain situations
   ([PR](https://github.com/hashicorp/boundary/pull/2553))
+* workers: PKI Worker daemons now get disconnected from upstreams when their
+  corresponding resource is deleted ([PR](https://github.com/hashicorp/boundary/pull))
```
irenarindos, line 19:
```suggestion
  corresponding resource is deleted ([PR](https://github.com/hashicorp/boundary/pull/2573))
```

internal/daemon/cluster/downstreammanager.go
```diff
@@ -0,0 +1,67 @@
...
+			r = append(r, k)
+		}
+	}
+	return r
+}
+
+// DisconnectUnauthorized calls disconnects for all ids which are present in
+// the connected but not in the authorized slice of key ids.
+func DisconnectUnauthorized(dm *DownstreamManager, connected, authorized []string) {
+	am := make(map[string]struct{}, len(authorized))
```
irenarindos, line 58: Check for nil DownstreamManager?

internal/server/repository_workerauth.go
```diff
@@ -517,6 +517,33 @@ func (r *WorkerAuthRepositoryStorage) findWorkerAuth(ctx context.Context, node *
 	return worker, nil
 }
 
+// FilterToAuthorizedWorkerKeyIds returns all the worker key identifiers that
+// are authorizable from the slice of key identifiers provided to the function.
```
irenarindos, line 521:
```suggestion
// are authorized from the slice of key identifiers provided to the function.
```

internal/server/repository_workerauth.go
```diff
@@ -517,6 +517,33 @@ func (r *WorkerAuthRepositoryStorage) findWorkerAuth(ctx context.Context, node *
...
+func (r *WorkerAuthRepositoryStorage) FilterToAuthorizedWorkerKeyIds(ctx context.Context, workerKeyIds []string) ([]string, error) {
+	const op = "server.(WorkerAuthRepositoryStorage).FilterToAuthorizedWorkerKeyIds"
+	if len(workerKeyIds) == 0 {
+		return nil, nil
+	}
+	rows, err := r.reader.Query(ctx, authorizedWorkerQuery, []interface{}{workerKeyIds})
+	if err != nil {
```
irenarindos, line 528: Do something with the error here?

irenarindos (changes requested):

internal/daemon/controller/controller.go
```diff
@@ -419,10 +423,9 @@ func (c *Controller) Start() error {
 		defer c.tickerWg.Done()
 		c.startCloseExpiredPendingTokens(c.baseContext)
 	}()
-	go func() {
-		defer c.tickerWg.Done()
-		c.started.Store(true)
-	}()
+	if err := c.startWorkerConnectionMaintenanceTicking(c.baseContext, c.tickerWg, c.pkiConnManager); err != nil {
```
jefferai, line 426: The pattern for the rest of the tickers is to start it in a goroutine here. Any reason to deviate?

internal/server/repository_workerauth.go
```diff
@@ -517,6 +517,34 @@ func (r *WorkerAuthRepositoryStorage) findWorkerAuth(ctx context.Context, node *
...
+	}
+	defer rows.Close()
+
+	type rowsResult struct {
+		WorkerKeyIdentifier string
+	}
+	var ret []string
+	for rows.Next() {
+		var result rowsResult
+		err = r.reader.ScanRows(ctx, rows, &result)
```
jefferai, line 539: Can you not just scan directly into a string?

internal/server/query.go
```diff
@@ -23,4 +23,12 @@ const (
 		delete from worker_auth_ca_certificate
  		where state = @state;
 	`
+
+	authorizedWorkerQuery = `
+		select distinct w.worker_key_identifier 
+		from 
+			worker_auth_certificate_bundle as w
```
jefferai, line 30: I thought with the workers API you could delete a KMS worker which would disconnect it (although it could potentially auth again -- but maybe not if it's lost its KMS access). This query only handles PKI workers though right? Are KMS workers included in the rollup (seems like they should be) and would they end up being disconnected because of this?

internal/server/repository_workerauth_test.go
```diff
@@ -444,6 +444,49 @@ func TestStoreNodeInformationTx(t *testing.T) {
 	}
 }
 
+func TestFilterToAuthorizedWorkerKeyIds(t *testing.T) {
```
jefferai, line 447: Similar to my point above, there are no KMS workers in this test...

jefferai (commented):
Before continuing on with a review I think we need to discuss KMS behavior here.

PR #2576, opened 2022-10-27
enable recovery user to list targets and sessions

irenarindos:


internal/perms/acl.go
```diff
@@ -243,6 +244,10 @@ func (a ACL) ListPermissions(requestedScopes map[string]*scopes.ScopeInfo, reque
 			Action:   action.List,
 			OnlySelf: true, // default to only self to be restrictive
 		}
+		if userId == "u_recovery" {
+			perms = append(perms, p)
```
jefferai, line 248: I think you need to unset `OnlySelf` in this case

jefferai: And I think you also need to set `p.All`

internal/daemon/controller/auth/testing.go
```diff
@@ -20,7 +21,7 @@ func DisabledAuthTestContext(iamRepoFn common.IamRepoFactory, scopeId string, op
 	}
 	reqInfo.UserIdOverride = opts.withUserId
 	if reqInfo.UserIdOverride == "" {
-		reqInfo.UserIdOverride = "u_auth"
+		reqInfo.UserIdOverride = perms.AdminUserId
```
jefferai, line 24: This doesn't seem right. There's no such thing as an "admin user", at least in terms of built-ins.

internal/daemon/controller/handlers/groups/group_service.go
```diff
@@ -710,7 +710,7 @@ func validateAddGroupMembersRequest(req *pbs.AddGroupMembersRequest) error {
 			badFields["member_ids"] = fmt.Sprintf("Must only contain valid user ids but found %q.", id)
 			break
 		}
-		if id == "u_recovery" {
+		if id == perms.RecoveryUserId {
```
jefferai, line 713: This should really be in the `globals` package as there are other places that use `u_recovery`. I'd hold off switching things for a separate PR though as it will add a lot of extra files/changes to this PR.

internal/iam/repository_scope.go
```diff
@@ -85,9 +86,9 @@ func (r *Repository) CreateScope(ctx context.Context, s *Scope, userId string, o
 	var adminRoleRaw any
 	switch {
 	case userId == "",
-		userId == "u_anon",
-		userId == "u_auth",
-		userId == "u_recovery",
+		userId == perms.AnonymousUserId,
+		userId == perms.AdminUserId,
```
jefferai, line 90: I think you meant for this to be `perms.AnyAuthenticatedUserId` or `perms.AnyAuthedUserId`

internal/perms/acl.go
```diff
@@ -234,7 +238,9 @@ func (a ACL) Allowed(r Resource, aType action.Type, userId string, opt ...Option
 // or for action.All in order for a Permission to be created for the scope.
 // The set of "id actions" is resource dependant, but will generally include all
 // actions that can be taken on an individual resource.
-func (a ACL) ListPermissions(requestedScopes map[string]*scopes.ScopeInfo, requestedType resource.Type, idActions action.ActionSet) []Permission {
+func (a ACL) ListPermissions(requestedScopes map[string]*scopes.ScopeInfo, requestedType resource.Type,
+	idActions action.ActionSet, userId string,
```
jefferai, line 242: Nit: rather than break it in the middle can you just have each parameter on its own line?

jefferai (changes requested):
A couple of things to change.

Also, the test changes don't actually exercise the changes to `auth.go`. I would suggest modifying one of the recursive listing tests in sessions or targets to have a test case that uses the recovery user.

internal/daemon/controller/handlers/sessions/session_service_test.go
```diff
@@ -294,6 +294,9 @@ func TestList(t *testing.T) {
 	o, pWithSessions := iam.TestScopes(t, iamRepo)
 	oOther, pWithOtherSessions := iam.TestScopes(t, iamRepo)
 
+	fmt.Println(fmt.Sprintf("First scope is: %v", o))
```
jefferai, line 297: Left in some debug here

irenarindos: D'oh, thanks!

internal/daemon/controller/handlers/groups/group_service.go
irenarindos: I'll do this in a follow-up PR

jefferai (approved):
Thank you!

PR #2577, opened 2022-10-27
test(e2e): Add more checks in credential store tests

moduli:
This PR does the following
- Adds a workflow to `static/credential_store_test.go` to create a username/password credential
- Adds a check in `static/credential_store_test.go` and `static_with_vault/credential_store_test.go` to validate that `authorize-session` does not return any credentials when no credentials are associated with a target
- Adds various `connect` tests in `static_with_vault` to ensure that users can connect to targets using the various options when credentials are stored by vault
- Reorganizes the test directory to better align with enos scenarios and have more clarity for where to add tests
- Fixes a flake in `static/session_cancel_user_test.go` due to a misunderstanding in how goroutines work. Instead, the goroutine will now authenticate via token instead of utilizing the shared keyring.

### DIFF
--- a/testing/internal/e2e/tests/static/connect_ssh_test.go
+++ b/testing/internal/e2e/tests/static/connect_ssh_test.go
@@ -15,10 +15,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// TestConnectTargetWithSshCli uses the boundary cli to create a credential using boundary's
+// TestCliConnectTargetWithSsh uses the boundary cli to create a credential using boundary's
 // built-in credential store. The test attaches that credential to a target and attempts to connect
 // to that target using those credentials.
-func TestConnectTargetWithSshCli(t *testing.T) {
+func TestCliConnectTargetWithSsh(t *testing.T) {
 	e2e.MaybeSkipTest(t)
 	c, err := loadConfig()
 	require.NoError(t, err)
@@ -81,7 +81,7 @@ func TestConnectTargetWithSshCli(t *testing.T) {
 	require.Equal(t, string(k), retrievedKey)
 	t.Log("Successfully retrieved credentials for target")
 
-	// Connect to target and print host's IP address using retrieved credentials
+	// Connect to target using ssh option
 	output = e2e.RunCommand(ctx, "boundary",
 		e2e.WithArgs(
 			"connect", "ssh",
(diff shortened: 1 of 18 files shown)


testing/internal/e2e/tests/static/connect_ssh_test.go
```diff
@@ -0,0 +1,96 @@
...
+	retrievedUser := fmt.Sprintf("%s", newSessionAuthorization.Credentials[0].Credential["username"])
+	retrievedKey := fmt.Sprintf("%s\n", newSessionAuthorization.Credentials[0].Credential["private_key"])
+	assert.Equal(t, c.TargetSshUser, retrievedUser)
+
+	k, err := os.ReadFile(c.TargetSshKeyPath)
+	require.NoError(t, err)
+	require.Equal(t, string(k), retrievedKey)
+	t.Log("Successfully retrieved credentials for target")
+
+	// Connect to target and print host's IP address using retrieved credentials
```
johanbrandhorst, line 84: This doesn't print the hosts IP address, is the test wrong or is the comment wrong?

testing/internal/e2e/tests/static/session_cancel_user_test.go
```diff
@@ -122,6 +123,7 @@ func TestSessionCancelUserCli(t *testing.T) {
 				"{{boundary.ip}}",
 				"hostname -i; sleep 60",
 			),
+			e2e.WithEnv("E2E_AUTH_TOKEN", token),
```
johanbrandhorst, line 126: _nice_

testing/internal/e2e/tests/static_with_vault/connect_ssh_test.go
```diff
@@ -0,0 +1,154 @@
...
+	retrievedUser := fmt.Sprintf("%s", newSessionAuthorization.Credentials[0].Credential["username"])
+	retrievedKey := fmt.Sprintf("%s", newSessionAuthorization.Credentials[0].Credential["private_key"])
+	assert.Equal(t, c.TargetSshUser, retrievedUser)
+
+	k, err := os.ReadFile(c.TargetSshKeyPath)
+	require.NoError(t, err)
+	require.Equal(t, string(k), retrievedKey)
+	t.Log("Successfully retrieved credentials for target")
+
+	// Connect to target and print host's IP address using retrieved credentials
```
johanbrandhorst, line 142: Same here

testing/internal/e2e/tests/static_with_vault/env_test.go
```diff
@@ -0,0 +1,21 @@
...
+}
+
+func loadConfig() (*config, error) {
+	var c config
+	err := envconfig.Process("", &c)
+	if err != nil {
+		return nil, err
+	}
+
+	return &c, err
```
johanbrandhorst, line 20: Nit: `err` is always nil here, it's conventional to just return `nil` in these cases
```suggestion
	return &c, nil
```

johanbrandhorst (approved):
🚢

PR #2578, opened 2022-10-27
Split grace period into multiple config values

jefferai:
This splits a misleadingly-named env var for status grace period into separate timers that are more targeted and can be set via config or env.

Note: an almost identical version of this was previously reviewed by Todd and Jim, very little has changed other than rebasing on 0.11 instead of 0.10.5.


internal/daemon/controller/controller.go
```diff
@@ -98,6 +99,10 @@ type Controller struct {
 	// Used for testing and tracking worker health
 	workerStatusUpdateTimes *sync.Map
 
+	// Timing variables
+	workerStatusGracePeriod *atomic.Int64
+	livenessTimeToStale     *atomic.Int64
```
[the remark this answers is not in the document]
jefferai: It's for future `SIGHUP` support which would be easy but I didn't do here. I will document it.

internal/daemon/worker/status.go
```diff
@@ -117,18 +117,23 @@ func (w *Worker) sendWorkerStatus(cancelCtx context.Context, sessionManager sess
 	var activeJobs []*pbs.JobStatus
 
 	// Range over known sessions and collect info
+	var connectionCount uint64
```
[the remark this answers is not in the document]
jefferai: Done.

internal/daemon/worker/worker.go
```diff
@@ -117,6 +118,10 @@ type Worker struct {
 	downstreamWorkers downstreamers
 	downstreamRoutes  downstreamRouter
 
+	// Timing variables
+	successfulStatusGracePeriod *atomic.Int64
```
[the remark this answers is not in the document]
jefferai: Done.

internal/cmd/config/config.go
```diff
@@ -583,6 +636,41 @@ func Parse(d string) (*Config, error) {
...
+		}
+		if successfulStatusGracePeriod != nil {
+			t, err := parseutil.ParseDurationSecond(successfulStatusGracePeriod)
+			if err != nil {
+				return result, err
+			}
+			result.Worker.SuccessfulStatusGracePeriodDuration = t
+		}
+
+		switch {
```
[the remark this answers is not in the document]
jefferai: Nice catch, done.

internal/cmd/config/config.go
```diff
@@ -498,6 +527,36 @@ func Parse(d string) (*Config, error) {
 			}
 		}
 
+		workerStatusGracePeriod := result.Controller.WorkerStatusGracePeriod
+		if workerStatusGracePeriod == nil {
```
jimlambrt, line 531: should this be a call to `util.IsNil(...)` since it's an interface?

jimlambrt: This comment sort of applies to all these new fields which are empty interfaces.

internal/daemon/cluster/handlers/worker_service.go
```diff
@@ -48,14 +50,16 @@ func NewWorkerServiceServer(
...
-		connectionRepoFn: connectionRepoFn,
-		updateTimes:      updateTimes,
-		kms:              kms,
+		serversRepoFn:       serversRepoFn,
+		workerAuthRepoFn:    workerAuthRepoFn,
+		sessionRepoFn:       sessionRepoFn,
+		connectionRepoFn:    connectionRepoFn,
+		updateTimes:         updateTimes,
+		kms:                 kms,
+		livenessTimeToStale: livenessTimeToStale,
```
jimlambrt, line 62: if `livenessTimeToStale` is nil, there's a possible panic when this is accessed.

internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -147,15 +151,16 @@ func NewService(
...
+		repoFn:                  repoFn,
+		iamRepoFn:               iamRepoFn,
+		serversRepoFn:           serversRepoFn,
+		sessionRepoFn:           sessionRepoFn,
+		pluginHostRepoFn:        pluginHostRepoFn,
+		staticHostRepoFn:        staticHostRepoFn,
+		vaultCredRepoFn:         vaultCredRepoFn,
+		staticCredRepoFn:        staticCredRepoFn,
+		kmsCache:                kmsCache,
+		workerStatusGracePeriod: workerStatusGracePeriod,
```
jimlambrt, line 163: same question about a possible panic if this is nil.

internal/daemon/worker/session/manager.go
```diff
@@ -5,11 +5,19 @@ import (
...
 	"github.com/hashicorp/boundary/internal/errors"
 	pbs "github.com/hashicorp/boundary/internal/gen/controller/servers/services"
 )
 
+// FIXME: This is really ugly -- but not as ugly as plumbing this value into the
+// interface. We should figure out something better. For now, it will at least
+// keep sync with any changes to the value once it's initialized as the worker
+// will set it in its new function; it can also be overridden in tests if
+// desired with normal global variable caveats.
+var CloseCallTimeout = new(atomic.Int64)
```
jimlambrt, line 19: I like to use set/getter funcs for this sort of global vs just exporting it.  It's not a lot better... but it sure feels better.

internal/daemon/worker/worker.go
```diff
@@ -185,6 +193,21 @@ func New(conf *Config) (*Worker, error) {
...
+		w.successfulStatusGracePeriod.Store(int64(conf.RawConfig.Worker.SuccessfulStatusGracePeriodDuration))
+	}
+	switch conf.RawConfig.Worker.StatusCallTimeoutDuration {
+	case 0:
+		w.statusCallTimeoutDuration.Store(int64(common.DefaultStatusTimeout))
+	default:
+		w.statusCallTimeoutDuration.Store(int64(conf.RawConfig.Worker.StatusCallTimeoutDuration))
+	}
+	// FIXME: This is really ugly, but works.
+	session.CloseCallTimeout = w.statusCallTimeoutDuration
```
jimlambrt, line 209: yuck.

internal/session/jobs.go
```diff
@@ -13,7 +14,7 @@ import (
 const deleteTerminatedThreshold = time.Hour
 
 // RegisterJobs registers session related jobs with the provided scheduler.
-func RegisterJobs(ctx context.Context, scheduler *scheduler.Scheduler, w db.Writer, r db.Reader, k *kms.Kms, gracePeriod time.Duration) error {
+func RegisterJobs(ctx context.Context, scheduler *scheduler.Scheduler, w db.Writer, r db.Reader, k *kms.Kms, gracePeriod *atomic.Int64) error {
```
jimlambrt, line 17: perhaps check for nil gracePeriod?

jimlambrt (approved):
Couple pretty minor comments; feel free to ignore or address.

internal/cmd/config/config.go
jefferai: Done

internal/daemon/cluster/handlers/worker_service.go
jefferai: It's true. I didn't want to change this to return an error right now; I'm relying on the fact that we know it will not be nil for any real controller/worker (since it's explicitly set in the New functions) so only tests might panic and then we can and should fix the test.

internal/daemon/worker/session/manager.go
jefferai: Fair, but will leave as an exercise for another time.

internal/daemon/worker/worker.go
jefferai: ikr

internal/session/jobs.go
jefferai: Done

PR #2579, opened 2022-10-28
test(e2e): Add `ssh localhost` test

moduli:
This PR adds another test case that starts a `boundary connect` followed by a `ssh localhost -p PORT` to connect to the target. This is another common way that users access targets.

### DIFF
--- a/testing/internal/e2e/tests/static/connect_localhost_test.go
+++ b/testing/internal/e2e/tests/static/connect_localhost_test.go
@@ -1,0 +1,119 @@
+package static_test
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/cenkalti/backoff/v4"
+	"github.com/hashicorp/boundary/api/sessions"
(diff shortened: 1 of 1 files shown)


testing/internal/e2e/tests/static/connect_localhost_test.go
```diff
@@ -0,0 +1,119 @@
...
+	ctxCancel, cancel := context.WithCancel(context.Background())
+	port := "12345"
+	cmdChan := make(chan *e2e.CommandResult)
+	go func() {
+		t.Log("Starting session...")
+		cmdChan <- e2e.RunCommand(ctxCancel, "boundary",
+			e2e.WithArgs(
+				"connect",
+				"-target-id", newTargetId,
+				"-listen-port", port,
```
moduli, line 45: I couldn't figure out a way to access the output of this command without canceling it (in order to get the port that gets generated), so I had to specify a port here.

johanbrandhorst: If you wanted to, I think you could do this with a pipe, but this seems fine and much less complicated.

johanbrandhorst (approved):

PR #2580, opened 2022-10-28
move user variables into globals

irenarindos:


johanbrandhorst (approved):
Very nice!

jefferai (changes requested):
Looks great except for the tmp test files you committed :-D

PR #2583, opened 2022-11-01
Store cert private key, encrypt tofu token

johanbrandhorst:
Previously, we would derive a session certificate private key from the session key used in each project, and not store the key. We would also encrypt the tofu token with the database key but not store a reference to this key in the database.

This change fixes this by replacing our key derivation with a key generation step, and instead store the generated key, encrypted, in the database. We also ensure that any new tofu tokens are encrypted with the same key.

To handle existing sessions, we lazily rewrite the sessions whenever a user lists them. There is a minor risk that a user could end up destroying the database key that encrypts the tofu token before that session has been rewrapped, but this is going to be rare and temporary.

### DIFF
--- a/internal/daemon/cluster/handlers/worker_service.go
+++ b/internal/daemon/cluster/handlers/worker_service.go
@@ -349,6 +349,7 @@ func (ws *workerServiceServer) LookupSession(ctx context.Context, req *pbs.Looku
 		Authorization: &targets.SessionAuthorizationData{
 			SessionId:   sessionInfo.GetPublicId(),
 			Certificate: sessionInfo.Certificate,
+			PrivateKey:  sessionInfo.CertificatePrivateKey,
 		},
 		Status:          sessionInfo.States[0].Status.ProtoVal(),
 		Version:         sessionInfo.Version,
@@ -365,18 +366,6 @@ func (ws *workerServiceServer) LookupSession(ctx context.Context, req *pbs.Looku
 	}
 	if resp.ConnectionsLeft != -1 {
 		resp.ConnectionsLeft -= int32(authzSummary.CurrentConnectionCount)
-	}
-
-	wrapper, err := ws.kms.GetWrapper(ctx, sessionInfo.ProjectId, kms.KeyPurposeSessions, kms.WithKeyId(sessionInfo.KeyId))
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "Error getting sessions wrapper: %v", err)
-	}
-
-	// Derive the private key, which should match. Deriving on both ends allows
-	// us to not store it in the DB.
-	_, resp.Authorization.PrivateKey, err = session.DeriveED25519Key(ctx, wrapper, sessionInfo.UserId, sessionInfo.GetPublicId())
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "Error deriving session key: %v", err)
 	}
 
 	return resp, nil
--- a/internal/db/schema/migrations/oss/postgres_56_02_test.go
+++ b/internal/db/schema/migrations/oss/postgres_56_02_test.go
@@ -9,9 +9,8 @@ import (
 	"github.com/hashicorp/boundary/internal/db"
 	"github.com/hashicorp/boundary/internal/db/common"
 	"github.com/hashicorp/boundary/internal/db/schema"
-	"github.com/hashicorp/boundary/internal/iam"
 	"github.com/hashicorp/boundary/internal/kms"
-	"github.com/hashicorp/boundary/internal/session"
+	"github.com/hashicorp/boundary/internal/types/scope"
 	"github.com/hashicorp/boundary/testing/dbtest"
 	"github.com/stretchr/testify/require"
 )
@@ -65,12 +64,9 @@ func TestMigrations_NewForeignKeys(t *testing.T) {
 
 	// Create project
 	wrapper := db.TestWrapper(t)
-	iamRepo := iam.TestRepo(t, conn, wrapper)
 	kmsCache := kms.TestKms(t, conn, wrapper)
-
-	// Create a session
-	sess := session.TestSession(t, conn, wrapper, session.TestSessionParams(t, conn, wrapper, iamRepo))
-	_ = sess
+	err = kmsCache.CreateKeys(ctx, scope.Global.String())
+	require.NoError(t, err)
 
 	// Create a password account
 	pwam := password.TestAuthMethod(t, conn, "global")
--- a/internal/session/options.go
+++ b/internal/session/options.go
@@ -138,6 +138,8 @@ func WithPermissions(p *perms.UserPermissions) Option {
 
 // WithIgnoreDecryptionFailures is used to ignore decryption
 // failures when doing lookups. This should be used sparingly.
+// It is currently only used to allow a user to cancel a session
+// in the presence of a undecryptable TOFU token.
 func WithIgnoreDecryptionFailures(ignoreFailures bool) Option {
 	return func(o *options) {
 		o.withIgnoreDecryptionFailures = ignoreFailures
--- a/internal/session/options.go
+++ b/internal/session/options.go
@@ -1,6 +1,8 @@
 package session
 
 import (
+	"crypto/rand"
+	"io"
 	"time"
 
 	"github.com/hashicorp/boundary/internal/db"
@@ -35,11 +37,13 @@ type options struct {
 	withTerminated               bool
 	withPermissions              *perms.UserPermissions
 	withIgnoreDecryptionFailures bool
+	withRandomReader             io.Reader
 }
 
 func getDefaultOptions() options {
 	return options{
 		withWorkerStateDelay: 10 * time.Second,
+		withRandomReader:     rand.Reader,
 	}
 }
 
@@ -137,5 +141,13 @@ func WithPermissions(p *perms.UserPermissions) Option {
 func WithIgnoreDecryptionFailures(ignoreFailures bool) Option {
 	return func(o *options) {
 		o.withIgnoreDecryptionFailures = ignoreFailures
+	}
+}
+
+// WithRandomReader is used to configure the random source
+// to use when generating secrets. Defaults to crypto/rand.Reader.
+func WithRandomReader(rand io.Reader) Option {
+	return func(o *options) {
+		o.withRandomReader = rand
 	}
 }
--- a/internal/session/options.go
+++ b/internal/session/options.go
@@ -22,18 +22,19 @@ type Option func(*options)
 
 // options = how options are represented
 type options struct {
-	withLimit             int
-	withOrderByCreateTime db.OrderBy
-	withProjectIds        []string
-	withUserId            string
-	withExpirationTime    *timestamp.Timestamp
-	withTestTofu          []byte
-	withListingConvert    bool
-	withSessionIds        []string
-	withDbOpts            []db.Option
-	withWorkerStateDelay  time.Duration
-	withTerminated        bool
-	withPermissions       *perms.UserPermissions
+	withLimit                    int
+	withOrderByCreateTime        db.OrderBy
+	withProjectIds               []string
+	withUserId                   string
+	withExpirationTime           *timestamp.Timestamp
+	withTestTofu                 []byte
+	withListingConvert           bool
+	withSessionIds               []string
+	withDbOpts                   []db.Option
+	withWorkerStateDelay         time.Duration
+	withTerminated               bool
+	withPermissions              *perms.UserPermissions
+	withIgnoreDecryptionFailures bool
 }
 
 func getDefaultOptions() options {
@@ -128,5 +129,13 @@ func WithTerminated(withTerminated bool) Option {
 func WithPermissions(p *perms.UserPermissions) Option {
 	return func(o *options) {
 		o.withPermissions = p
+	}
+}
+
+// WithIgnoreDecryptionFailures is used to ignore decryption
+// failures when doing lookups. This should be used sparingly.
+func WithIgnoreDecryptionFailures(ignoreFailures bool) Option {
+	return func(o *options) {
+		o.withIgnoreDecryptionFailures = ignoreFailures
 	}
 }
--- a/internal/session/query.go
+++ b/internal/session/query.go
@@ -443,7 +443,7 @@ values
 `
 )
 
-func batchInsertsessionCredentialDynamic(creds []*DynamicCredential) (string, []interface{}, error) {
+func batchInsertSessionCredentialDynamic(creds []*DynamicCredential) (string, []interface{}, error) {
 	if len(creds) <= 0 {
 		return "", nil, fmt.Errorf("empty slice of DynamicCredential, cannot build query")
 	}
--- a/internal/session/repository.go
+++ b/internal/session/repository.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"io"
 	"sort"
 	"strings"
 
@@ -14,6 +15,7 @@ import (
 	"github.com/hashicorp/boundary/internal/perms"
 	"github.com/hashicorp/boundary/internal/types/action"
 	"github.com/hashicorp/boundary/internal/types/resource"
+	"github.com/hashicorp/boundary/internal/util"
 )
 
 // Clonable provides a cloning interface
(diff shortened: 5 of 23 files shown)


internal/daemon/cluster/handlers/worker_service.go
```diff
@@ -351,18 +352,6 @@ func (ws *workerServiceServer) LookupSession(ctx context.Context, req *pbs.Looku
...
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "Error getting sessions wrapper: %v", err)
-	}
-
-	// Derive the private key, which should match. Deriving on both ends allows
-	// us to not store it in the DB.
-	_, resp.Authorization.PrivateKey, err = session.DeriveED25519Key(ctx, wrapper, sessionInfo.UserId, sessionInfo.GetPublicId())
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "Error deriving session key: %v", err)
-	}
```
johanbrandhorst, line 364: There's no need for this logic anymore as the `LookupSession` call above will derive and populate the private key for older sessions.

internal/db/schema/migrations/oss/postgres_56_02_test.go
```diff
@@ -65,12 +64,9 @@ func TestMigrations_NewForeignKeys(t *testing.T) {
 
 	// Create project
 	wrapper := db.TestWrapper(t)
-	iamRepo := iam.TestRepo(t, conn, wrapper)
 	kmsCache := kms.TestKms(t, conn, wrapper)
-
-	// Create a session
-	sess := session.TestSession(t, conn, wrapper, session.TestSessionParams(t, conn, wrapper, iamRepo))
-	_ = sess
```
johanbrandhorst, line 73: This became a pain now since we have to do it manually, so I removed it. It doesn't add that much confidence that I think it's worth trying to figure out doing it manually, but I can spend the time if we insist.

internal/session/repository.go
```diff
@@ -134,40 +134,36 @@ func (r *Repository) convertToSessions(ctx context.Context, sessionList []*sessi
...
-					databaseWrapper, err := r.kms.GetWrapper(ctx, workingSession.ProjectId, kms.KeyPurposeDatabase)
-					if err != nil {
-						return nil, errors.Wrap(ctx, err, op, errors.WithMsg("unable to get database wrapper"))
-					}
-					if err := workingSession.decrypt(ctx, databaseWrapper); err != nil {
-						return nil, errors.Wrap(ctx, err, op, errors.WithMsg("cannot decrypt session value"))
-					}
-				} else {
-					workingSession.CtTofuToken = nil
+				if err := decryptAndMaybeUpdateSession(ctx, r.kms, workingSession, r.writer); err != nil {
```
johanbrandhorst, line 165: This will lazily update any old style sessions on ListSession, which should migrate sessions very quickly.

internal/session/repository_session_test.go
```diff
@@ -1206,22 +1207,6 @@ func TestRepository_CancelSessionViaFKNull(t *testing.T) {
...
-					Scope: &iamStore.Scope{
-						PublicId: s.ProjectId,
-					},
-				}
-				return cancelFk{
-					s:      s,
-					fkType: t,
-				}
-			}(),
-		},
```
johanbrandhorst, line 1224: I removed this test as it was no longer possible to delete a scope without a foreign key reference error from the key_id column in the session when trying to cascade delete the keys in the scope. This limitation is in place to avoid deleting keys that are still encrypting data.

internal/session/rewrapping_test.go
```diff
@@ -45,38 +46,38 @@ func TestRewrap_sessionCredentialRewrapFn(t *testing.T) {
 	}
 
 	kmsWrapper, err := kmsCache.GetWrapper(context.Background(), prj.PublicId, kms.KeyPurposeDatabase)
-	assert.NoError(t, err)
+	require.NoError(t, err)
```
johanbrandhorst, line 49: I updated all of these to `require` as we don't want to test to continue on any of these errors (since it may invalidate assumptions about values used in the rest of the test, leading to confusing panics).

internal/session/session.go
```diff
@@ -381,7 +391,7 @@ func newCert(ctx context.Context, wrapper wrapping.Wrapper, userId, jobId string
 	if len(addresses) == 0 {
 		return nil, nil, errors.New(ctx, errors.InvalidParameter, op, "missing addresses")
 	}
-	pubKey, privKey, err := DeriveED25519Key(ctx, wrapper, userId, jobId)
+	pubKey, privKey, err := ed25519.GenerateKey(nil)
```
johanbrandhorst, line 394: Using a `nil` random reader defaults the reader to `crypto/rand.Reader`

internal/daemon/cluster/handlers/worker_service.go
jefferai: 👍

internal/session/session.go
jefferai: Correct, but don't do this. :-D Pass in the SecureRandomReader from the base server (Controller in this case), which will require a bit of plumbing, but it's a forward-looking thing.

jefferai (commented):
One comment, but looks amazing!

internal/session/session.go
johanbrandhorst: Will do

internal/session/repository_session.go
```diff
@@ -680,3 +674,54 @@ func fetchConnections(ctx context.Context, r db.Reader, sessionId string, opt ..
...
+		// New-style session with private key stored in DB, just decrypt.
+		if err := session.decrypt(ctx, sessionWrapper); err != nil {
+			return errors.Wrap(ctx, err, op, errors.WithMsg("unable to decrypt session value"))
+		}
+		return nil
+	}
+	// No certificate private key present, this is a legacy session with a
+	// private key derived from the key. Derive it again and store it back
+	// in the DB.
+	_, session.CertificatePrivateKey, err = DeriveED25519Key(ctx, sessionWrapper, session.UserId, session.GetPublicId())
```
jimlambrt, line 698: Do we care if session.UserId or session.GetPublicId() are empty?  Derivation will succeed even if the salt and/or info are provided.

internal/session/repository_session.go
```diff
@@ -680,3 +674,54 @@ func fetchConnections(ctx context.Context, r db.Reader, sessionId string, opt ..
...
+	if len(session.CtCertificatePrivateKey) > 0 {
+		// New-style session with private key stored in DB, just decrypt.
+		if err := session.decrypt(ctx, sessionWrapper); err != nil {
+			return errors.Wrap(ctx, err, op, errors.WithMsg("unable to decrypt session value"))
+		}
+		return nil
+	}
+	// No certificate private key present, this is a legacy session with a
+	// private key derived from the key. Derive it again and store it back
+	// in the DB.
```
jimlambrt, line 697: I don't see coverage for anything below here.... is that intentional?

internal/session/rewrapping.go
```diff
@@ -95,17 +95,17 @@ func sessionRewrapFn(ctx context.Context, dataKeyVersionId string, scopeId strin
...
 			return errors.Wrap(ctx, err, op, errors.WithMsg("failed to fetch kms wrapper for rewrapping"))
 		}
-		if err := session.decrypt(ctx, wrapper); err != nil {
-			return errors.Wrap(ctx, err, op, errors.WithMsg("failed to decrypt session"))
-		}
 		if err := session.encrypt(ctx, wrapper); err != nil {
 			return errors.Wrap(ctx, err, op, errors.WithMsg("failed to re-encrypt session"))
 		}
-		if _, err := writer.Update(ctx, session, []string{"CtTofuToken", "KeyId"}, nil); err != nil {
+		if _, err := writer.Update(ctx, session, []string{"CtTofuToken", "CtCertificatePrivateKey", "KeyId"}, nil); err != nil {
```
jimlambrt, line 108: Doesn't decryptAndMaybeUpdateSession update these fields as well?

internal/session/session.go
```diff
@@ -367,7 +377,7 @@ func contains(ss []string, t string) bool {
 }
 
 // newCert creates a new session certificate. If addresses are supplied, they will be parsed and added to IP or DNS SANs as appropriate.
-func newCert(ctx context.Context, wrapper wrapping.Wrapper, userId, jobId string, addresses []string, exp time.Time) (ed25519.PrivateKey, []byte, error) {
+func newCert(ctx context.Context, wrapper wrapping.Wrapper, userId, jobId string, addresses []string, exp time.Time, rand io.Reader) (ed25519.PrivateKey, []byte, error) {
```
jimlambrt, line 380: perhaps test rand for IsNil(...) or are you relying on ed25519.GenerateKey(...) to handle nil properly?

jimlambrt: perhaps test rand for IsNil(...)?  I'm not sure if CreateCertificate(...) does the right thing.

internal/session/session.go
```diff
@@ -429,8 +439,32 @@ func newCert(ctx context.Context, wrapper wrapping.Wrapper, userId, jobId string
 
 func (s *Session) encrypt(ctx context.Context, cipher wrapping.Wrapper) error {
```
jimlambrt, line 440: should we test if IsNil cipher?

internal/session/session.go
```diff
@@ -442,32 +476,59 @@ func (s *Session) encrypt(ctx context.Context, cipher wrapping.Wrapper) error {
 
 func (s *Session) decrypt(ctx context.Context, cipher wrapping.Wrapper) error {
```
jimlambrt, line 477: same question about chipher is nil...

internal/daemon/cluster/handlers/worker_service.go
```diff
@@ -333,6 +333,7 @@ func (ws *workerServiceServer) LookupSession(ctx context.Context, req *pbs.Looku
 		Authorization: &targets.SessionAuthorizationData{
 			SessionId:   sessionInfo.GetPublicId(),
 			Certificate: sessionInfo.Certificate,
+			PrivateKey:  sessionInfo.CertificatePrivateKey,
```
jimlambrt, line 336: do we want to assert this in any of the unit tests?

jimlambrt (changes requested):
Overall looks great.  Some minor questions.

internal/session/query.go
```diff
@@ -443,7 +443,7 @@ values
 `
 )
 
-func batchInsertsessionCredentialDynamic(creds []*DynamicCredential) (string, []interface{}, error) {
+func batchInsertSessionCredentialDynamic(creds []*DynamicCredential) (string, []interface{}, error) {
```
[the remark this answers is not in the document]
johanbrandhorst: The LLB is currently targeting 0.11.x which doesn't have the `interface{}` -> `any` rewrite applied to it, but I will update these when rebasing the LLB on `main`.

internal/session/options.go
```diff
@@ -130,3 +135,19 @@ func WithPermissions(p *perms.UserPermissions) Option {
 		o.withPermissions = p
 	}
 }
+
+// WithIgnoreDecryptionFailures is used to ignore decryption
+// failures when doing lookups. This should be used sparingly.
```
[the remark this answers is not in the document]
johanbrandhorst: I added a comment and a test for these two, thank you!

internal/session/repository.go
```diff
@@ -69,6 +73,7 @@ func NewRepository(ctx context.Context, r db.Reader, w db.Writer, kms *kms.Kms,
 		kms:          kms,
 		defaultLimit: opts.withLimit,
 		permissions:  opts.withPermissions,
+		randomReader: opts.withRandomReader,
```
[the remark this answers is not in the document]
johanbrandhorst: I added a test for this an the other options, thanks

internal/session/repository.go
```diff
@@ -134,40 +139,36 @@ func (r *Repository) convertToSessions(ctx context.Context, sessionList []*sessi
...
+				Version:                 sv.Version,
+				Endpoint:                sv.Endpoint,
+				ConnectionLimit:         sv.ConnectionLimit,
+				KeyId:                   sv.KeyId,
 			}
 			if opts.withListingConvert {
-				workingSession.CtTofuToken = nil // CtTofuToken should not returned in lists
-				workingSession.TofuToken = nil   // TofuToken should not returned in lists
-				workingSession.KeyId = ""        // KeyId should not be returned in lists
+				workingSession.CtCertificatePrivateKey = nil // CtCertificatePrivateKey should not returned in lists
```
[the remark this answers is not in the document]
johanbrandhorst: Added a test!

internal/session/repository_session.go
```diff
@@ -20,62 +19,62 @@ import (
...
 	}
 	if newSession.AuthTokenId == "" {
-		return nil, nil, errors.New(ctx, errors.InvalidParameter, op, "missing auth token id")
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing auth token id")
 	}
 	if newSession.ProjectId == "" {
-		return nil, nil, errors.New(ctx, errors.InvalidParameter, op, "missing project id")
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing project id")
 	}
 	if newSession.CtTofuToken != nil {
```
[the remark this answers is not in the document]
johanbrandhorst: The only low hanging fruit here was the expiration time, so I added a test for that. It doesn't look like the newSession fields were ever tested.

internal/session/repository_session.go
```diff
@@ -680,3 +674,54 @@ func fetchConnections(ctx context.Context, r db.Reader, sessionId string, opt ..
 	}
 	return connections, nil
 }
+
+// decryptAndMaybeUpdateSession switches between the database key and session key based on whether
+// the session uses a legacy private key or not. It also updates the encrypted session
+// in the database (if necessary). Eventually we should be able to remove this function
+// and use the session key unconditionally.
+func decryptAndMaybeUpdateSession(ctx context.Context, kmsRepo *kms.Kms, session *Session, writer db.Writer) error {
+	const op = "session.decryptAndMaybeUpdateSession"
```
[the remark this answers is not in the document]
johanbrandhorst: I've added validation for these parameters, good spot.

internal/session/repository_session.go
```diff
@@ -680,3 +674,54 @@ func fetchConnections(ctx context.Context, r db.Reader, sessionId string, opt ..
...
+	sessionWrapper, err := kmsRepo.GetWrapper(ctx, session.ProjectId, kms.KeyPurposeSessions, kms.WithKeyId(session.KeyId))
```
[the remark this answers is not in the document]
johanbrandhorst: Yep, that's an error, I added some validation.

internal/session/repository_session.go
johanbrandhorst: Yep, I've added some validation that covers this.

internal/session/repository_session.go
johanbrandhorst: I've added some tests for this function, good call!

internal/session/rewrapping.go
johanbrandhorst: Only in the case of old style sessions. We still need to do this update for any new style sessions (with the session cert private key in).

internal/session/session.go
johanbrandhorst: I looked at this method with a fresh set of eyes and realised I could remove half the parameters, so now it's both tested and much slimmer. Good shout!

internal/session/session.go
johanbrandhorst: Good idea, added a check.

internal/session/session.go
johanbrandhorst: Good idea, added a check.

internal/daemon/cluster/handlers/worker_service.go
johanbrandhorst: Good spot, I added asserts for this value.

internal/session/repository.go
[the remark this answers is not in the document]
jimlambrt: ty for adding the test.  I'm not sure that additional test covers my question here.  I was asking if we should test that the factory function actually honors the option passed in by setting `randomReader` appropriately based on the opts?

internal/session/repository_session.go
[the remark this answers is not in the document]
jimlambrt: I was hoping to convince you to add the missing unit tests for newSession fields that had been added post orig implementation.

internal/session/session.go
jimlambrt: Much better!!!!!

jimlambrt (approved):
Looks great!

PR #2584, opened 2022-11-01
fix(test): Retry websocket connection due to flaky failure

moduli:
We have observed several unit tests occasionally fail when trying to read a handshake due to the following error:
```
worker_shutdown_reload_test.go:148: 
         Error Trace:	testing_helper.go:138
                                 testing_helper.go:312
                                 worker_shutdown_reload_test.go:148
                        failed to read protobuf message: failed to get reader: received close frame: status = StatusProtocolError and reason = "received header with unexpected rsv bits set: false:false:true"
         Test:       	TestServer_ShutdownWorker
--- FAIL: TestServer_ShutdownWorker (1.78s)
```
This also occurs in `TestWorkerSessionProxyMultipleConnections` and `TestSessionCleanup/default/multi_controller`.

After an investigation, it's still unclear how this happens. The `wspb.Write` seemed to work ok, but the `Read` contained some unexpected data.

We are opting to retry the websocket connection to mitigate the flakiness.

### DIFF
--- a/internal/tests/helper/testing_helper.go
+++ b/internal/tests/helper/testing_helper.go
@@ -111,14 +111,14 @@ func (s *TestSession) connect(ctx context.Context, t *testing.T) net.Conn {
 
 	var conn *websocket.Conn
 	var resp *http.Response
-	var err error
 	var handshakeResult proxy.HandshakeResult
 	// A retry was added here to mitigate some flakiness.
 	// Occasionally, `wspb.Read` would throw an error due to "received header with unexpected
 	// rsv bits set: false:false:true". It was unclear what the cause of this was, so we
 	// resorted to retrying the connection.
-	err = backoff.RetryNotify(
+	err := backoff.RetryNotify(
 		func() error {
+			var err error
 			conn, resp, err = websocket.Dial(
 				ctx,
 				fmt.Sprintf("wss://%s/v1/proxy", s.WorkerAddr),
@@ -131,12 +131,6 @@ func (s *TestSession) connect(ctx context.Context, t *testing.T) net.Conn {
 			)
 			if err != nil {
 				return backoff.Permanent(err)
-			}
-			if conn == nil {
-				return backoff.Permanent(errors.New("conn is nil"))
-			}
-			if resp == nil {
-				return backoff.Permanent(errors.New("resp is nil"))
 			}
 			if resp.Header.Get("Sec-WebSocket-Protocol") != globals.TcpProxyV1 {
 				return backoff.Permanent(errors.New(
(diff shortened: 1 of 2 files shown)


internal/tests/helper/testing_helper.go
```diff
@@ -107,34 +108,67 @@ func NewTestSession(
...
-	require.NotNil(resp)
-	require.Equal(resp.Header.Get("Sec-WebSocket-Protocol"), globals.TcpProxyV1)
 
-	// Send the handshake.
-	if s.tofuToken == "" {
-		s.tofuToken, err = base62.Random(20)
-	}
+	var conn *websocket.Conn
+	var resp *http.Response
+	var err error
```
johanbrandhorst, line 114: We probably don't need this variable to be declared here

internal/tests/helper/testing_helper.go
```diff
@@ -107,34 +108,67 @@ func NewTestSession(
...
+			conn, resp, err = websocket.Dial(
+				ctx,
+				fmt.Sprintf("wss://%s/v1/proxy", s.WorkerAddr),
+				&websocket.DialOptions{
+					HTTPClient: &http.Client{
+						Transport: s.Transport,
+					},
+					Subprotocols: []string{globals.TcpProxyV1},
+				},
+			)
```
johanbrandhorst, line 131: You can declare the error here to limit the scope.
```suggestion
			var err error
			conn, resp, err = websocket.Dial(
				ctx,
				fmt.Sprintf("wss://%s/v1/proxy", s.WorkerAddr),
				&websocket.DialOptions{
					HTTPClient: &http.Client{
						Transport: s.Transport,
					},
					Subprotocols: []string{globals.TcpProxyV1},
				},
			)
```

internal/tests/helper/testing_helper.go
```diff
@@ -107,34 +108,67 @@ func NewTestSession(
...
+			if err != nil {
+				return backoff.Permanent(err)
+			}
+			if conn == nil {
+				return backoff.Permanent(errors.New("conn is nil"))
+			}
+			if resp == nil {
+				return backoff.Permanent(errors.New("resp is nil"))
+			}
```
johanbrandhorst, line 140: I don't think we need to check these two, I know they were tested before, but this is generally a bit excessive, worst case the test will crash 🤷🏻.

internal/tests/helper/testing_helper.go
```diff
@@ -107,34 +108,67 @@ func NewTestSession(
...
-	// Receive/check the handshake
-	var handshakeResult proxy.HandshakeResult
-	err = wspb.Read(ctx, conn, &handshakeResult)
+			err = wspb.Write(ctx, conn, &handshake)
+			if err != nil {
+				return backoff.Permanent(err)
+			}
+
+			err = wspb.Read(ctx, conn, &handshakeResult)
+			return err
```
johanbrandhorst, line 165: If we're not wrapping the error, this simplifies to
```suggestion
			return wspb.Read(ctx, conn, &handshakeResult)
```

internal/tests/helper/testing_helper.go
```diff
@@ -107,34 +108,67 @@ func NewTestSession(
...
+		},
+		backoff.WithMaxRetries(backoff.NewConstantBackOff(1*time.Second), 1),
```
johanbrandhorst, line 167:
```suggestion
		backoff.WithMaxRetries(backoff.NewConstantBackOff(time.Second), 1),
```

johanbrandhorst (approved):

PR #2587, opened 2022-11-02
Inject Custom Response Headers By Status

DanielleMiu:
relies on [go-secure-stdlib#49](https://github.com/hashicorp/go-secure-stdlib/pull/49)


internal/daemon/controller/handler.go
```diff
@@ -57,18 +58,113 @@ type HandlerProperties struct {
...
+		}
+	}
+
+	// Setting the default headers first
+	if val, ok := sch["default"]; ok {
+		setter(val)
+	}
+
+	// Then setting the generic hundred-level headers
+	d := fmt.Sprintf("%vxx", statusCode/100)
```
johanbrandhorst, line 113: Use `%d` here to ensure we're working with integers. `%v` allows any input value.
```suggestion
	d := fmt.Sprintf("%dxx", statusCode/100)
```

johanbrandhorst (commented):
Nice start!

johanbrandhorst (commented):
Thanks for the extensive docs, I'm tagging @Dan-Heath for a look at this as well

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -21,6 +21,21 @@ The `listener` stanza may be specified more than once to make Boundary listen on
 multiple interfaces; however, only one listener marked for `cluster` purpose is
 allowed.
 
+## Listener's custom response headers
+
+Boundary supports defining custom HTTP response headers for all requests on any Boundary controller.
+Headers are defined based on the returned status code. For example, a user can define a list of
+custom response headers for the `200` status code, and another list of custom response headers for
+the `307` status code, and so on. Users can also define headers based on the hundred-level status
+code. For example, a list of headers applied to the `4xx` code will be applied to 400, 401, 404 and
```
Dan-Heath, line 30:
```suggestion
code. For example, a list of headers applied to the `4xx` code will be applied to 400, 401, 404, and
```

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -21,6 +21,21 @@ The `listener` stanza may be specified more than once to make Boundary listen on
 multiple interfaces; however, only one listener marked for `cluster` purpose is
 allowed.
 
+## Listener's custom response headers
+
+Boundary supports defining custom HTTP response headers for all requests on any Boundary controller.
+Headers are defined based on the returned status code. For example, a user can define a list of
```
Dan-Heath, line 27:
```suggestion
Headers are defined based on the returned status code. For example, you can define a list of
```

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -21,6 +21,21 @@ The `listener` stanza may be specified more than once to make Boundary listen on
...
+custom response headers for the `200` status code, and another list of custom response headers for
+the `307` status code, and so on. Users can also define headers based on the hundred-level status
```
Dan-Heath, line 29:
```suggestion
the `307` status code, and so on. You can also define headers based on the hundred-level status
```

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -21,6 +21,21 @@ The `listener` stanza may be specified more than once to make Boundary listen on
...
+code. For example, a list of headers applied to the `4xx` code will be applied to 400, 401, 404 and
+all other 400-level status codes. The more specific the status, the higher priority it will have.
+Default headers are overwritten by hundred-level headers, which are overwritten by status-specific
+headers.
```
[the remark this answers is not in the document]
Dan-Heath: I like Johan's suggestion above, but I would change the sentence a bit to make it more active:

> If a header is desired for all responses, _you must explicitly add it to all configured levels_.

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -84,6 +99,42 @@ allowed.
   default, such as `"Content-Type"`, `"X-Requested-With"`, and
   `"Authorization"`.
 
+### `custom_api_response_headers` Parameters
+
+- `default` `(key-value-map: {})` - A map of string header names to an array of
+  string values. The default headers are set on all endpoints regardless of
+  the status code value. For an example, please refer to the
```
Dan-Heath, line 106:
```suggestion
  the status code value. For an example, refer to the
```

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -177,6 +228,69 @@ listener "tcp" {
 }
 ```
 
+### Configuring custom http response headers
+
+Note: Requires Boundary version 0.12 or newer. This example shows configuring custom http response headers.
```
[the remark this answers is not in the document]
Dan-Heath: We do have versioning built in. If this is for the 0.12 release, we will only publish it to that branch (and going forward). So I think Johan's right that we don't need the note.

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -177,6 +228,69 @@ listener "tcp" {
...
+Operators can configure `"custom_api_response_headers"` and `"custom_ui_response_headers"` sub-stanzas in
+the listener stanza to set custom http headers appropriate to their applications. Examples of such headers
```
Dan-Heath, line 235:
```suggestion
the listener stanza to set custom http headers that are appropriate to their applications. Examples of such headers
```

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -177,6 +228,69 @@ listener "tcp" {
...
+are `"Strict-Transport-Security"` and `"Content-Security-Policy"` which are known HTTP headers, and could
+be configured to harden the security of an application communicating with the Boundary endpoints. Note that
+vulnerability scans often examine such security related HTTP headers. In addition, application specific
+custom headers can also be configured. For example, `"X-Custom-Header"` has been configured in the example
```
Dan-Heath, line 239:
```suggestion
vulnerability scans often examine such security related HTTP headers. In addition, you can configure application-specific custom headers. For example, `"X-Custom-Header"` has been configured in the example
```

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -177,6 +228,69 @@ listener "tcp" {
...
+    "301" = {
+      "Strict-Transport-Security" = ["max-age=31536000"],
+      "Content-Security-Policy" = ["default-src 'none'; script-src 'none'; connect-src 'none'"],
+    },
+  }
+}
+```
+
+In situations where a header is defined under several status code subsections,
+the header matching the most specific response code will be returned. For example,
```
Dan-Heath, line 274:
```suggestion
Boundary returns the header matching the most specific response code. For example,
```

Dan-Heath (commented):
Thank you for the documentation! I think it reads well and makes sense. I added some minor suggestions for style and grammar. Please let me know if you have any questions about them.

website/content/docs/configuration/listener/tcp.mdx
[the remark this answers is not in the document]
DanielleMiu: do they need to, though? wouldn't they just set it under default? unless you mean specifically that they want different custom values _for each response code_, then yes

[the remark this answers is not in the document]
johanbrandhorst: What I mean is that we should make it explicit that there's no inheritance from more generic to more specific codes.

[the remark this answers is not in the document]
johanbrandhorst: Apparently I misunderstood how this worked, so carry on!

internal/daemon/controller/handler.go
```diff
@@ -57,18 +57,36 @@ type HandlerProperties struct {
...
+
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		if isUiRequest(req) {
+			uiHandler.ServeHTTP(w, req)
+		} else {
+			grpcGwMux.ServeHTTP(w, req)
+		}
+	}), nil
+}
+
```
tmessi, line 82: I don't follow this change that moves away from calling `mux.Handle` and instead adds this `isUiRequest` method.

tmessi: After digging into https://github.com/hashicorp/go-secure-stdlib/pull/49 I think I see what is going on here. It still seems odd to introduce this `!strings.HasPrefix` check with `isUiRequest` instead of leveraging the `mux.Handle` and allowing the mux to do more efficient matching, but I am not seeing a way to do that given the implementation in listenerutil.

tmessi (approved):

internal/daemon/controller/handler.go
DanielleMiu: [latest commit](https://github.com/hashicorp/boundary/pull/2587/commits/a5263b3ec52e4f4be733bc3b58f677fb7a209ede) swaps `strings.HasPrefix` for [`mux.Handler`](https://pkg.go.dev/net/http?utm_source=gopls#ServeMux.Handler), which is used internally by `mux.ServeHTTP`, so we can go back to using mux's logic for serving as well

internal/daemon/controller/handler.go
```diff
@@ -57,18 +57,34 @@ type HandlerProperties struct {
...
+	}
+
+	mux := http.NewServeMux()
+	mux.Handle("/v1/", grpcGwMux)
+	mux.Handle("/", handleUi(c))
+
+	isUiRequest := func(req *http.Request) bool {
+		_, p := mux.Handler(req)
+		// check to see if the matched pattern is for the ui
+		return p == "/"
```
tmessi, line 75: I would suggest you mane "/" a const to make sure these never get out of sync for some reason, ie:

```go
const uiPath = "/"
```

tmessi (commented):
One minor thing, otherwise looks good to me

PR #2588, opened 2022-11-02
test(e2e): Check that an active session ends if the user is deleted

moduli:
This PR adds a test to the e2e test suite that does the following...
- Start a session as a user
- Delete the user with the active session
- Check that the session is automatically ended

```
=== RUN   TestCliSessionEndWhenUserIsDeleted
    scope.go:68: Created Org Id: o_OIEuW2gkIL
    scope.go:91: Created Project Id: p_WpjOpdXQnl
    host.go:79: Created Host Catalog: hcst_KmKLcK9vMS
    host.go:99: Created Host Set: hsst_wtKRjm46h9
    host.go:121: Created Host: hst_wfdSLWIPnk
    target.go:59: Created Target: ttcp_N4kcy0sIWP
    account.go:73: Created Account: acctpw_oGa6iPHkQs
    user.go:54: Created User: u_2U5qCF8o5Y
    role.go:30: Created Role: r_b9G0wRWFtc
    session.go:18: Waiting for session to be active...
    session.go:71: No items are appearing in the session list. Retrying...
    session_end_delete_user_test.go:47: Starting session as user...
    session.go:40: Found 1 session(s)
    session_end_delete_user_test.go:76: Waiting for session to be canceling/terminated...
    session_end_delete_user_test.go:108: Session successfully ended after user was deleted
--- PASS: TestCliSessionEndWhenUserIsDeleted (10.79s)
```

### DIFF
--- a/testing/internal/e2e/boundary/user.go
+++ b/testing/internal/e2e/boundary/user.go
@@ -48,10 +48,9 @@ func CreateNewUserCli(t testing.TB, ctx context.Context, scopeId string) string 
 	newUserId := newUserResult.Item.Id
 	t.Cleanup(func() {
 		AuthenticateAdminCli(t, context.Background())
-		output := e2e.RunCommand(ctx, "boundary",
+		e2e.RunCommand(ctx, "boundary",
 			e2e.WithArgs("users", "delete", "-id", newUserId),
 		)
-		require.NoError(t, output.Err, string(output.Stderr))
 	})
 	t.Logf("Created User: %s", newUserId)
 
--- a/testing/internal/e2e/boundary/user.go
+++ b/testing/internal/e2e/boundary/user.go
@@ -48,9 +48,21 @@ func CreateNewUserCli(t testing.TB, ctx context.Context, scopeId string) string 
 	newUserId := newUserResult.Item.Id
 	t.Cleanup(func() {
 		AuthenticateAdminCli(t, context.Background())
-		e2e.RunCommand(ctx, "boundary",
-			e2e.WithArgs("users", "delete", "-id", newUserId),
+		output := e2e.RunCommand(ctx, "boundary",
+			e2e.WithArgs("users", "delete", "-id", newUserId, "-format", "json"),
 		)
+		// Only allow an error if it's due to a Resource Not Found (404)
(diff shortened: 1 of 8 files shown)


testing/internal/e2e/boundary/user.go
```diff
@@ -48,10 +47,9 @@ func CreateNewUserCli(t testing.TB, ctx context.Context, scopeId string) string
 	newUserId := newUserResult.Item.Id
 	t.Cleanup(func() {
 		AuthenticateAdminCli(t, context.Background())
-		output := e2e.RunCommand(ctx, "boundary",
+		e2e.RunCommand(ctx, "boundary",
```
moduli, line 50: Since the test was deleting the user, this check was failing because the user was already deleted. I'm thinking about potentially removing these checks in any of the cleanup steps. Thoughts?

johanbrandhorst: Should we not do this on a case by case basis? Sure it makes sense to expect an error when trying to delete a user that was successfully deleted as part of the test, but you could just write this logic to ignore that error specifically and still error on other errors.

moduli: Mm. I did think about that. I'll explore it further.

moduli: Pushed up a change

testing/internal/e2e/boundary/user.go
```diff
@@ -49,9 +49,20 @@ func CreateNewUserCli(t testing.TB, ctx context.Context, scopeId string) string
...
+		// This was needed since there's a test that already deletes the user and would run into an
+		// error during cleanup
+		if output.Err != nil {
+			var response e2e.CliError
+			err := json.Unmarshal(output.Stderr, &response)
+			require.NoError(t, err)
+			statusCode := response.Status
+			require.Condition(t, func() bool {
+				return statusCode == 404
+			})
```
johanbrandhorst, line 64: Testify is a bit particular about types when using literals, this should work
```suggestion
			require.Equal(t, int(response.Status), 404)
```

johanbrandhorst (approved):

PR #2590, opened 2022-11-03
Add logging to journal for systemd

stephan242:
Right now no logs are visible for a boundary instance run via systemd. This PR adds 2 lines to the unit file to log stdout and stderr to journal, so that the logs can be seen with e.g.

```
journalctl -u boundary@pki-worker.service
```

That is, if one manually implements the PR https://github.com/hashicorp/boundary/pull/2268 which btw really should be merged asap, since right now boundary via systemd is essentially not working, unless one puts the config file one wants to run to `/etc/boundary.d/.hcl`, which would be a dangerous botch job.

### DIFF
--- a/.release/linux/package/usr/lib/systemd/system/boundary.service
+++ b/.release/linux/package/usr/lib/systemd/system/boundary.service
@@ -18,6 +18,8 @@ Restart=on-failure
 RestartSec=5
 TimeoutStopSec=30
 LimitMEMLOCK=infinity
+StandardOutput=journal
+StandardError=journal
 
 [Install]
 WantedBy=multi-user.target


omkensey:
From what I understand, stdout should be piped to the journal by default when started as a systemd unit and journald is running, and that's the behavior I've observed previously when running Boundary via a unit file.  What distro are you using?

stephan242:
```
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.1 LTS"

$ systemctl --version
systemd 249 (249.11-0ubuntu3.4)
+PAM +AUDIT +SELINUX +APPARMOR +IMA +SMACK +SECCOMP +GCRYPT +GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS +FIDO2 +IDN2 -IDN +IPTC +KMOD +LIBCRYPTSETUP -LIBFDISK +PCRE2 -PWQUALITY -P11KIT -QRENCODE +BZIP2 +LZ4 +XZ +ZLIB +ZSTD -XKBCOMMON +UTMP +SYSVINIT default-hierarchy=unified

```

PR #2595, opened 2022-11-03
test(e2e): Update test to also use a username/password credential in vault

moduli:
This test extends an existing Vault e2e test to also create a username/password credential in vault and attach that credential to a target.


testing/internal/e2e/vault/vault.go
```diff
@@ -3,17 +3,20 @@ package vault
...
-	"runtime"
 	"testing"
 
 	"github.com/hashicorp/boundary/testing/internal/e2e"
+	"github.com/hashicorp/go-secure-stdlib/base62"
 	"github.com/kelseyhightower/envconfig"
 	"github.com/stretchr/testify/require"
 )
 
+//go:embed boundary-controller-policy.hcl
```
moduli, line 17: @johanbrandhorst I tried using the embed here. It wasn't working with a `testdata` path, presumably because this particular package isn't a test package?

testing/internal/e2e/tests/static_with_vault/credential_store_test.go
```diff
@@ -138,14 +165,20 @@ func TestCliVaultCredentialStore(t *testing.T) {
 	require.NoError(t, err)
 
 	newSessionAuthorization := newSessionAuthorizationResult.Item
-	retrievedUser := fmt.Sprintf("%s", newSessionAuthorization.Credentials[0].Credential["username"])
-	retrievedKey := fmt.Sprintf("%s", newSessionAuthorization.Credentials[0].Credential["private_key"])
+	retrievedUser := newSessionAuthorization.Credentials[0].Credential["username"].(string)
```
johanbrandhorst, line 168: What do you think about adding a len check before doing the index lookup? This would avoid a panic if we get an unexpected result.
```suggestion
	require.Len(t, newSessionAuthorization.Credentials, 2)
	retrievedUser := newSessionAuthorization.Credentials[0].Credential["username"].(string)
```

testing/internal/e2e/vault/vault.go
johanbrandhorst: If it's in testdata, you may need to reference testdata explicitly
```suggestion
//go:embed testdata/boundary-controller-policy.hcl
```

This would also affect the path when calling `ReadFile`

johanbrandhorst (approved):

testing/internal/e2e/vault/vault.go
moduli: Yup. I was missing the `ReadFile` part when I was trying it earlier.

testing/internal/e2e/tests/static_with_vault/credential_store_test.go
```diff
@@ -138,14 +165,26 @@ func TestCliVaultCredentialStore(t *testing.T) {
...
-	assert.Equal(t, c.TargetSshUser, retrievedUser)
+	require.Len(t, newSessionAuthorization.Credentials, 2)
 
-	k, err := os.ReadFile(c.TargetSshKeyPath)
-	require.NoError(t, err)
-	require.Equal(t, string(k), retrievedKey)
-	t.Log("Successfully retrieved credentials for target")
+	for _, v := range newSessionAuthorization.Credentials {
+		if v.CredentialSource.CredentialType == "ssh_private_key" {
+			retrievedUser := newSessionAuthorization.Credentials[0].Credential["username"].(string)
```
johanbrandhorst, line 172: If you're going to use a loop, use the loop variable too 😛 
```suggestion
			retrievedUser := v.Credential["username"].(string)
```

johanbrandhorst (changes requested):

testing/internal/e2e/tests/static_with_vault/credential_store_test.go
```diff
@@ -138,14 +165,26 @@ func TestCliVaultCredentialStore(t *testing.T) {
...
-	retrievedUser := fmt.Sprintf("%s", newSessionAuthorization.Credentials[0].Credential["username"])
-	retrievedKey := fmt.Sprintf("%s", newSessionAuthorization.Credentials[0].Credential["private_key"])
-	assert.Equal(t, c.TargetSshUser, retrievedUser)
+	require.Len(t, newSessionAuthorization.Credentials, 2)
 
-	k, err := os.ReadFile(c.TargetSshKeyPath)
-	require.NoError(t, err)
-	require.Equal(t, string(k), retrievedKey)
-	t.Log("Successfully retrieved credentials for target")
+	for _, v := range newSessionAuthorization.Credentials {
```
moduli, line 170: @johanbrandhorst FYI: After I added the length check above (probably coincidentally), I started seeing some occasional failures where the credentials weren't in this exact order. Added this loop to check the credential type.

testing/internal/e2e/tests/static_with_vault/credential_store_test.go
moduli: Oops

testing/internal/e2e/vault/vault.go
```diff
@@ -31,71 +34,108 @@ func loadConfig() (*config, error) {
...
-		require.NoError(t, output.Err, string(output.Stderr))
-	})
+	// Set up boundary policy
+	data, err := boundaryPolicyFile.ReadFile("testdata/boundary-controller-policy.hcl")
+	require.NoError(t, err)
+	boundaryPolicyFilePath := fmt.Sprintf("%s/%s", t.TempDir(), "boundary-controller-policy.hcl")
+	f, err := os.Create(boundaryPolicyFilePath)
+	require.NoError(t, err)
+	_, err = f.WriteString(string(data))
+	require.NoError(t, err)
```
johanbrandhorst, line 49: Looks like we can avoid this whole write-out-to-tmp-file-dance by just sending the policy contents to stdin: https://developer.hashicorp.com/vault/docs/commands/policy/write. Would be nice to avoid all the file system operations that can fail. What do you think?

moduli: Can give it a shot

moduli: Do you know if we can do the same for boundary/postgres? My initial search didn't turn up with anything, which means we would have to do this dance in other places.

moduli: @johanbrandhorst After thinking about it some more, I think I'd like to leave it as is since this aligns with the what our tutorials do (creating a file).

johanbrandhorst: Hm, in that case, maybe we should just not bother embedding and read straight from the testdata directory?

moduli: I don't mind keeping the embedding in there as it helped with the issue of needing to use `runtime.Caller` to get the path of this file since you could have different callers here. Is there another solution?

johanbrandhorst: Hm, I guess maybe the Vault command doesn't know to look in the testdata folder... alright sure!

johanbrandhorst (approved):

johanbrandhorst (approved):

PR #2600, opened 2022-11-09
Add KMS rotation/destruction feature to the changelog and docs

johanbrandhorst:
This adds a note to the changelog about the new functionality, referring to the docs page. The docs changes give a simple run through of the usage and meaning of each command.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,14 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
     string after a specified substring; this is useful for pulling an user/account name from an email address. In the following example it uses the account email can be any other parameter:
     
     * `{{ truncateFrom .Account.Email "@" }}`: this would turn `foo@example.com` into `foo`
+* Per-Scope Key lifecycle management: Both Key Encryption Keys (KEKs) and Data Encryption
+  Keys (DEKs) can now have their lifecycle managed by the user through the new
+  key rotation and key destruction functionality. To learn more about this new feature,
+  please visit the [docs
+  site](https://developer.hashicorp.com/boundary/docs/concepts/security/data-encryption).
+
+  Upgrade notice: If the Database purpose DEK for a scope is destroyed, sessions that
+  predate the upgrade need to be canceled via the API.
 
 * workers: PKI Worker daemons now get disconnected from upstreams when their
   corresponding resource is deleted ([PR](https://github.com/hashicorp/boundary/pull/2573))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,14 +30,14 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
     string after a specified substring; this is useful for pulling an user/account name from an email address. In the following example it uses the account email can be any other parameter:
     
     * `{{ truncateFrom .Account.Email "@" }}`: this would turn `foo@example.com` into `foo`
-* Per-Scope Key lifecycle management: Both Key Encryption Keys (KEKs) and Data Encryption
-  Keys (DEKs) can now have their lifecycle managed by the user through the new
-  key rotation and key version destruction functionality. To learn more about this new feature,
-  please visit the [docs
-  site](https://developer.hashicorp.com/boundary/docs/concepts/security/data-encryption).
+* Per-scope key lifecycle management: You can now manage the lifecycles of both Key
+  Encryption Keys (KEKs) and Data Encryption Keys (DEKs) using the new key rotation
+  and key version destruction functionality. To learn more about this new feature,
+  refer to the
+  [documentation](https://developer.hashicorp.com/boundary/docs/concepts/security/data-encryption).
 
-  Upgrade notice: If the Database purpose DEK for a scope is destroyed, sessions that
-  predate the upgrade need to be canceled via the API.
+  Upgrade notice: If the Database purpose DEK for a scope is destroyed, you must use
+  the API to cancel any sessions that predate the upgrade.
   ([PR](https://github.com/hashicorp/boundary/pull/2477))
   
 * workers: PKI Worker daemons now get disconnected from upstreams when their
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,13 +32,14 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
     * `{{ truncateFrom .Account.Email "@" }}`: this would turn `foo@example.com` into `foo`
 * Per-Scope Key lifecycle management: Both Key Encryption Keys (KEKs) and Data Encryption
   Keys (DEKs) can now have their lifecycle managed by the user through the new
-  key rotation and key destruction functionality. To learn more about this new feature,
+  key rotation and key version destruction functionality. To learn more about this new feature,
   please visit the [docs
   site](https://developer.hashicorp.com/boundary/docs/concepts/security/data-encryption).
 
   Upgrade notice: If the Database purpose DEK for a scope is destroyed, sessions that
   predate the upgrade need to be canceled via the API.
-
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -30,6 +30,14 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
     string after a specified substring; this is useful for pulling an user/account name from an email address. In the following example it uses the account email can be any other parameter:
     
     * `{{ truncateFrom .Account.Email "@" }}`: this would turn `foo@example.com` into `foo`
+* Per-Scope Key lifecycle management: Both Key Encryption Keys (KEKs) and Data Encryption
+  Keys (DEKs) can now have their lifecycle managed by the user through the new
+  key rotation and key destruction functionality. To learn more about this new feature,
+  please visit the [docs
+  site](https://developer.hashicorp.com/boundary/docs/concepts/security/data-encryption).
```
jimlambrt, line 37: This dev guide states that `Boundary does not currently support rotation of its internal keys` which is confusing given the capability being added :)

jimlambrt, line 37: This dev guide states that `Boundary does not currently support rotation of its internal keys` which is confusing given the capability being added :)

jimlambrt (changes requested):
I'm not sure this PR is properly rebased given that it contains 30 commits.

jefferai (changes requested):
Lotta conflicts for a changelog entry buddy

CHANGELOG.md
johanbrandhorst: May as well commandeer this PR for the docs updates too, good point.

johanbrandhorst:
Converted this to draft so it doesn't block LLB merging, lets wait with this until we have the LLB in the release branch.

website/content/docs/concepts/security/data-encryption.mdx
```diff
@@ -13,39 +13,42 @@ Systems](/docs/configuration/kms) that protect the base encryption keys used
...
-
 ## The `worker-auth-storage` KMS Key
 
 The `worker-auth-storage` KMS key is used by a [PKI
 Worker](/docs/configuration/worker/pki-worker) for storage of authentication
 keys. It is optional for PKI workers; if not specified the authentication keys
 will not be encrypted on disk. This is not used by KMS workers.
 
-## The `root` KMS Key and Per-Scope KEK/DEKs <sup>OSS Only</sup>
+## The `rootKey` KMS Key and Per-Scope KEK/DEKs <sup>OSS Only</sup>
```
johanbrandhorst, line 23: This is the purpose name we use internally, so it's what's visible to the user on the API.

CHANGELOG.md
```diff
@@ -30,7 +30,16 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
     string after a specified substring; this is useful for pulling an user/account name from an email address. In the following example it uses the account email can be any other parameter:
     
     * `{{ truncateFrom .Account.Email "@" }}`: this would turn `foo@example.com` into `foo`
-
+* Per-Scope Key lifecycle management: Both Key Encryption Keys (KEKs) and Data Encryption
```
moduli, line 33: I'm not sure I follow `Per-Scope Key` (is that a common term?). Does that just mean these keys are managed/organized by scope? 

EDIT: Seems like that's what it means. Maybe I was being thrown off by the capitalization in Scope? (Per-Scope -> Per-scope)

website/content/docs/concepts/security/data-encryption.mdx
```diff
@@ -13,39 +13,42 @@ Systems](/docs/configuration/kms) that protect the base encryption keys used
...
 
 Following best practices of using different encryption keys for different
 purposes, Boundary has a number of encryption keys generated within each scope.
+Each key further contains versions, which hold the key material used to encrypt
```
moduli, line 27: I'm slightly confused about this line. Is this saying that "versions" hold the key material? For some reason, when I first read it, it didn't make sense that "versions" itself holds key material, that it's just a version number.

website/content/docs/concepts/security/data-encryption.mdx
```diff
@@ -55,6 +58,53 @@ provided in this section is purely for informational purposes.
...
+API using the key endpoints under the scopes section. To rotate all the keys
+in a scope, use the `rotate-keys` endpoint:
+
+```shell-session
+$ boundary scopes rotate-keys -scope-id p_A4jfDjZ9jf
+```
+
+This creates new key versions for all keys in the scope `p_A4jfDjZ9jf`, and
+makes these key versions the active key versions for encrypting new data. The
+previous key version(s) will continue to be used for decrypting existing data.
```
moduli, line 73: I'm sure that has been talked about in the past, but is that expected behavior from users to only encrypt new data with the new key after you do a rotation? Was it mostly a time consideration not needing to re-encrypt all of the existing data?

Dan-Heath (commented):
I added some grammar/style suggestions, but overall this reads well and makes sense to me. If you do write the blog post that contains some of the "whys" for this feature, we can always update this later. Please let me know if you have any questions about my suggestions. Thanks!

website/content/docs/concepts/security/data-encryption.mdx
johanbrandhorst: Yep, the versions of a key hold the key material, so that each "key" is more conceptual and the key version actually maps more 1:1 to an encryption key. Do you think I could rewrite this to clarify this better?

website/content/docs/concepts/security/data-encryption.mdx
johanbrandhorst: My understand is that this sort of process is pretty standard - when rotating a key, it doesn't actually re-encrypt all existing data, it just means new data uses the new key. Re-encrypting all existing data could be a very expensive process. In the future we may do it in the background to automatically destroy key versions that are old, but we're not doing it right now, and I don't think we'll ever do it synchronously with a key rotation. With that in mind, do you think this sentence still makes sense?

CHANGELOG.md
```diff
@@ -30,7 +30,16 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
...
+  Keys (DEKs) can now have their lifecycle managed by the user through the new
+  key rotation and key version destruction functionality. To learn more about this new feature,
+  please visit the [docs
+  site](https://developer.hashicorp.com/boundary/docs/concepts/security/data-encryption).
```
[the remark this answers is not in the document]
johanbrandhorst: The intent with `Per-Scope` was to use `Scope` as a proper noun. Note how the title of this section uses this capitalization too. At the same time, I'm happy to change it here and just leave it in the title 👍🏻.

website/content/docs/concepts/security/data-encryption.mdx
```diff
@@ -55,6 +58,53 @@ provided in this section is purely for informational purposes.
...
+The `-rewrap` flag can be used to immediately rewrap all DEK versions with the
+new KEK version. Otherwise the DEK versions will remain encrypted by the KEK
+version created when they were.
```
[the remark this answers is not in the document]
johanbrandhorst: Looks good to me!

CHANGELOG.md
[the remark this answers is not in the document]
Dan-Heath: I _think_ that's right per the standards. I believe feature names are supposed to be [lowercase](https://github.com/hashicorp/engineering-docs/blob/main/writing/style-guide.md#capitalization-and-proper-nouns), but titles have first initials capitalized. So this looks good to me!

Dan-Heath (approved):
Thanks Johan! LGTM. I'm happy to take another look if any other changes come out of those last couple open questions.

CHANGELOG.md
johanbrandhorst: This was fixed :)

PR #2607, opened 2022-11-11
Key Rotation/Destruction (#2477)

johanbrandhorst:
This cherry-picks the KMS key rotation/destruction feature into `main`. There have been some minor changes made as part of the rebase:

1. The migrations are now in migration number 58 (was 56).
2. All `interface{}` types have been rewritten to `any`.
3. The e2e tests have been moved and rewritten with the new helpers.


testing/internal/e2e/tests/static/key_destruction_test.go
```diff
@@ -0,0 +1,344 @@
...
+		switch key.Purpose {
+		case "rootKey", "database":
+			assert.Len(t, key.Versions, 1)
+		default:
+			assert.Len(t, key.Versions, 2)
+		}
+	}
+
+	t.Logf("Successfully destroyed a root key and data key")
+}
```
johanbrandhorst, line 344: @moduli please have a look at the tests.

testing/internal/e2e/tests/static/key_destruction_test.go
```diff
@@ -0,0 +1,344 @@
...
+			"scopes", "list-keys",
+			"-scope-id", scopeReply.Item.Id,
+			"-format", "json",
+		),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+	var keys scopes.KeyListResult
+	err := json.Unmarshal(output.Stdout, &keys)
+	require.NoError(t, err)
+	assert.Len(t, keys.Items, len(kms.ValidDekPurposes())+1)
```
moduli, line 58: Also, would it be worth listing out the keys on failure to assist with debugging?

testing/internal/e2e/tests/static/key_destruction_test.go
```diff
@@ -0,0 +1,344 @@
...
+		e2e.WithArgs(
+			"scopes", "list-keys",
+			"-scope-id", scopeReply.Item.Id,
+			"-format", "json",
+		),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+	err = json.Unmarshal(output.Stdout, &keys)
+	require.NoError(t, err)
+	assert.Len(t, keys.Items, len(kms.ValidDekPurposes())+1)
```
moduli, line 106: Is this to ensure that the number of keys don't change after a rotation?

testing/internal/e2e/tests/static/key_destruction_test.go
```diff
@@ -0,0 +1,344 @@
+package static_test
```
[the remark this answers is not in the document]
johanbrandhorst: There is, should be fixed by https://github.com/hashicorp/boundary/pull/2615, which I've ported to this PR.

testing/internal/e2e/tests/static/key_destruction_test.go
```diff
@@ -0,0 +1,344 @@
...
+
```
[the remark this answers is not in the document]
johanbrandhorst: The test failures are because of a change to the oplog, I'll have to manually rewrite some tests.

testing/internal/e2e/tests/static/key_destruction_test.go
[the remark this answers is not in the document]
johanbrandhorst: We want to see all DEKs and also the root key, so it's all DEKs + 1. I've added a comment.

johanbrandhorst: Sure, I've added a debug message, though I don't know how useful it actually be since `keys.Items` is a pretty complex structure.

testing/internal/e2e/tests/static/key_destruction_test.go
johanbrandhorst: Yes, the number of keys always stays the same, we only get new versions.

PR #2608, opened 2022-11-11
feat(cli): Implement address flag for boundary targets create <target_type>

hugoamvieira:
This commit adds CLI support for addresses to be attached directly to a Target, for all current types of Target (ssh and tcp).

It also clarifies the relevant documentation regarding the autogen of API option functionality.


ddebko (approved):
LGTM

PR #2609, opened 2022-11-11
Update Boundary network requirements

Dan-Heath:
Per the following Slack conversation, the Boundary Network Requirements are incorrect:

https://hashicorp.slack.com/archives/C016ZKNM05T/p1668000335382949

This pull request:

- Removes redundant entries for ports 9200 and 9201
    - The Slack conversation mentioned that the client -> controller info re. port 9200 is repeated, but I think this is redundant as well: `:9201 must be open between Worker and Controller`. Please correct me if I am wrong.
- Adds the Client to Worker traffic with a required port of 9202
- Fixes the direction of the traffic flow from Worker -> Controller

Asana task: https://app.asana.com/0/563192436488770/1203340202571850/f

### DIFF
--- a/website/content/docs/getting-started/installing/production.mdx
+++ b/website/content/docs/getting-started/installing/production.mdx
@@ -12,9 +12,8 @@ Installing Boundary in a production setting requires prerequisits for infrastruc
 ## Network Requirements
 
 - Client -> Controller port is :9200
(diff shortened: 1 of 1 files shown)


stellarsquall (approved):
looks great 👍 thanks!

omkensey (approved):

PR #2610, opened 2022-11-11
Fix pull-down placement

Dan-Heath:
Per a Slack conversation:
https://hashicorp.slack.com/archives/C8DK5PR0B/p1668182567733029

Some of the pages in the Boundary docs have content above the H1 heading. Per Kyle McDonald, this is causing the version pull-down to overlap content and there is an upcoming rule that will warn that all pages must begin with an H1. 

image

This PR moves the H1 heading above any other content. The following pages in the Filtering section all have this issue:

- [OIDC managed groups](https://developer.hashicorp.com/boundary/docs/concepts/filtering/oidc-managed-groups)
- [Resource Listing](https://developer.hashicorp.com/boundary/docs/concepts/filtering/resource-listing)
- [Worker Tags](https://developer.hashicorp.com/boundary/docs/concepts/filtering/worker-tags)
- [Events](https://developer.hashicorp.com/boundary/docs/concepts/filtering/events)

### DIFF
--- a/website/content/docs/concepts/filtering/events.mdx
+++ b/website/content/docs/concepts/filtering/events.mdx
@@ -5,11 +5,9 @@ description: |-
   How to filter events emitted by Boundary.
 ---
 
-This page describes how to filter events written to Boundary event sinks.
-
-~> **Note:** This feature was introduced in Boundary 0.5.0.
-
 # Event Filtering
+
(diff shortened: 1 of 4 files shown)


stellarsquall (approved):
thanks for addressing this 🚀

PR #2612, opened 2022-11-11
Check that the returned resources before referencing them

talanknight:
Fixes https://hashicorp.atlassian.net/browse/ICU-6912

### DIFF
--- a/internal/daemon/cluster/handlers/worker_service.go
+++ b/internal/daemon/cluster/handlers/worker_service.go
@@ -295,21 +295,21 @@ func (ws *workerServiceServer) LookupSession(ctx context.Context, req *pbs.Looku
 	if sessionInfo.WorkerFilter != "" {
 		if req.WorkerId == "" {
 			event.WriteError(ctx, op, errors.New("worker filter enabled for session but got no id information from worker"))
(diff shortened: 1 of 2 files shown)


internal/daemon/cluster/handlers/worker_service.go
```diff
@@ -306,7 +306,10 @@ func (ws *workerServiceServer) LookupSession(ctx context.Context, req *pbs.Looku
 		if err != nil {
 			event.WriteError(ctx, op, err, event.WithInfoMsg("error looking up worker", "worker_id", req.WorkerId))
 			return &pbs.LookupSessionResponse{}, status.Errorf(codes.Internal, "Error looking up worker: %v", err)
-
+		}
+		if w == nil {
+			event.WriteError(ctx, op, err, event.WithInfoMsg("error looking up worker", "worker_id", req.WorkerId))
+			return &pbs.LookupSessionResponse{}, status.Errorf(codes.Internal, "Worker not found")
```
johanbrandhorst, line 312: Ugh this is such a dangerous API, to return `nil, nil` when the worker could not be found. I know this isn't the only place we do this but if this bug is any indicator it's not something we should continue doing.

johanbrandhorst (approved):

johanbrandhorst (commented):
Can we create some tests that exercises these issues?

johanbrandhorst (approved):
Thank you!

PR #2614, opened 2022-11-14
feat(cli): Implement `set-static-address` and `remove-static-address` actions to `boundary targets` subcommand

hugoamvieira:
Adds CLI support for `set-static-address` and `remove-static-address` actions to `boundary targets` subcommand. Also wires everything up from the CLI to the API. Because of this, this PR also adds some API stub methods that will be implemented later.

Sidenote for reviewers: I recommend reviewing each commit individually due to potentially confusing `*.gen.go` code reordering in [193252b](https://github.com/hashicorp/boundary/pull/2614/commits/193252b2f954c72ec49b99bedbdec7fd74333eea).


internal/cmd/commands/targetscmd/funcs.go
```diff
@@ -76,6 +79,16 @@ func extraSynopsisFuncImpl(c *Command) string {
...
+	case "set-static-address", "remove-static-address":
+		var in string
+		switch {
+		case strings.HasPrefix(c.Func, "set"):
+			in = "Set the static address for"
+		case strings.HasPrefix(c.Func, "remove"):
+			in = "Remove the static address from"
+		}
+		return wordwrap.WrapString(fmt.Sprintf("%s a target.", in), base.TermWidth)
+
```
johanbrandhorst, line 91: What do you think about
```suggestion
	case "set-static-address":
		return wordwrap.WrapString("Set the static address for a target.", base.TermWidth)
	case "remove-static-address":
		return wordwrap.WrapString("Remove the static address from a target.", base.TermWidth)

```

internal/cmd/commands/targetscmd/funcs.go
```diff
@@ -285,6 +323,15 @@ func extraFlagsFuncImpl(c *Command, _ *base.FlagSets, f *base.FlagSet) {
 			Usage:      "Target scope name, if authorizing the session via scope parameters and target name. Mutually exclusive with -scope-id.",
 		})
 	}
+
+	if c.Func == "set-static-address" {
+		flagsMap[c.Func] = append(flagsMap[c.Func], "address")
+		f.StringVar(&base.StringVar{
+			Name:   "address",
+			Target: &c.flagAddress,
+			Usage:  "Optionally, a valid network address to connect to for this target. Can not be used alongside host sources.",
```
johanbrandhorst, line 332: I believe this should be `cannot`.
```suggestion
			Usage:  "Optionally, a valid network address to connect to for this target. Cannot be used alongside host sources.",
```

internal/cmd/commands/targetscmd/funcs.go
```diff
@@ -383,6 +430,17 @@ func extraFlagsHandlingFuncImpl(c *Command, _ *base.FlagSets, opts *[]targets.Op
...
 		}
+
+	case "set-static-address":
+		if len(c.flagAddress) == 0 {
+			c.UI.Error("No Target static address supplied. Please supply it using -address.")
+			return false
+		}
+		if c.flagAddress == "null" {
+			*opts = append(*opts, targets.DefaultAddress())
+		} else {
```
johanbrandhorst, line 441: I do really dislike this but I think you already made the case that this is consistent with other flags?

johanbrandhorst (commented):
I want to get @talanknight or @jefferai to just have a quick look at the new API here before we merge it. It _is_ a change in convention, having this new command for setting a single field, vs `add-host-sources` which are appending or removing from a collection. Personally I prefer this to the use of a `null` literal, but I just wanted to flag it and get consensus.

hugoamvieira:
As discussed, we're not doing this anymore, instead opting to ensure consistency with existing behaviour (using `update` to set and remove)

PR #2615, opened 2022-11-15
Unset session key ID when deleting the scope, ignore decryption when session has no project or user

johanbrandhorst:
Unsetting the key ID allows the session to live beyond the lifetime of the scope, while allowing the scope to cascade delete its keys.

We also stop decrypting session fields if the session is missing a project or user, since decryption cannot always succeed in these cases.

Note that removing the relationship to the key means that the encrypted data in the (now canceled) session may become unrecoverable. This is an acceptable consequence of this change.

### DIFF
--- a/internal/db/schema/migrations/oss/postgres/56/02_add_data_key_foreign_key_references.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/56/02_add_data_key_foreign_key_references.up.sql
@@ -31,6 +31,35 @@ begin;
         on delete restrict
         on update cascade;
 
+  -- Replaces trigger from 44/04_session.up.sql
+  -- Adding a foreign key reference to the kms key means we have
+  -- to set the key_id to null when the scope_id is set to null,
+  -- otherwise we can't delete scopes with sessions, since deleting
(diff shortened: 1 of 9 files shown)


internal/db/schema/migrations/oss/postgres/56/02_add_data_key_foreign_key_references.up.sql
```diff
@@ -30,6 +30,17 @@ begin;
...
+  -- to delete the session when the scope is deleted, or else
+  -- we can't cascade scope deletions to the key referenced here.
+  alter table session
+    drop constraint iam_scope_project_fkey;
+  alter table session
+    add constraint iam_scope_project_fkey
+      foreign key (project_id)
+        references iam_scope_project (scope_id)
+        on delete cascade
+        on update cascade;
```
johanbrandhorst, line 43: I added this to an existing migration file rather creating a completely new one, since the change above this is what necessitates this change. Let me know if you'd rather I move this to a separate file.

mgaffney (approved):

DanielleMiu (approved):

PR #2616, opened 2022-11-15
refactor(db): apply consistent naming conventions

A-440Hz:
applies formatting and renames views and constraints as needed


A-440Hz:
first commit should be named `
apply consistent standards to 'comment on' statements`

johanbrandhorst (approved):
LGTM, though I fear it will be difficult to maintain this consistency. Did you use a script of some sort to do this or was it all manual work?

A-440Hz:
I did it manually. I don't know how to write a script to do this which doesn't rewrite every line in the sql files, messing up the git history

PR #2619, opened 2022-11-15
refactor(db): formatting for 'create trigger' statements

A-440Hz:


internal/db/schema/migrations/oss/postgres/24/01_session_valid_state.up.sql
```diff
@@ -23,68 +23,68 @@ set termination_reason='canceled' where termination_reason is null;
...
-            ),
-    current_state text
-        references session_state_enm(name)
-            on delete restrict
-            on update cascade,
-        constraint current_state_session_state_enm_fkey
-        check (
-            current_state in ('pending', 'active', 'canceling', 'terminated')
+  prior_state text
+      references session_state_enm(name)
```
johanbrandhorst, line 27: Should this be indented 2 spaces instead of 4? There are a few places in this PR where we sometimes indent by 4 instead of 2. I can leave comments on all of them if you want.

internal/db/schema/migrations/oss/postgres/56/01_auth.up.sql
```diff
@@ -3,11 +3,11 @@ begin;
 -- This deletes the base type if the concrete types exist anymore.
 delete from auth_account
 where public_id not in
-      (select public_id
-       from auth_oidc_account)
-and public_id not in
     (select public_id
-     from auth_password_account);
+     from auth_oidc_account)
+and public_id not in
```
johanbrandhorst, line 8: If you're going to change this, I think we want the `and` indented 2 spaces.

johanbrandhorst (commented):
There are a lot of whitespace changes here, if we do want to do this, please consider adding the commit (once merged) to https://github.com/hashicorp/boundary/blob/main/.git-blame-ignore-revs so it doesn't mess too much with attributing changes in history.

Also, I think the focus of this work should be on naming things consistently, not so much the formatting of the files themselves, that's going to be mostly impossible unless we get an automatic formatter. If these changes make it easier to do the naming that's great, but lets not spend too much time trying to sort out all of the formatting, that's going to be a lot of work for little gain.

mgaffney (changes requested):
Can you split this PR up? It would be helpful if the simple one line reformattings (like the `create trigger` statements) where separated from the multiline reformattings.

A-440Hz:
> There are a lot of whitespace changes here, if we do want to do this, please consider adding the commit (once merged) to https://github.com/hashicorp/boundary/blob/main/.git-blame-ignore-revs so it doesn't mess too much with attributing changes in history.
> 
> Also, I think the focus of this work should be on naming things consistently, not so much the formatting of the files themselves, that's going to be mostly impossible unless we get an automatic formatter. If these changes make it easier to do the naming that's great, but lets not spend too much time trying to sort out all of the formatting, that's going to be a lot of work for little gain.

Closing this PR based on Johan's feedback.

PR #2620, opened 2022-11-16
refactor(db): formatting consistency for 'create function' statements

A-440Hz:


johanbrandhorst (approved):
Like I said in the other PR, these sort of whitespace changes are probably mostly going to be a waste of time since we can't statically enforce the style. Don't worry too much about this, just making sure everything is named consistently is the primary goals of this work.

PR #2621, opened 2022-11-16
[QT-396] Add Enos CI bootstrapping scenarios

mikebaum:
As part of migrating the CI to use the newly created `boundary_ci` AWS account, a set of Terraform modules have been created to simplify bootstrapping the repo to use the account. The module `service-user-iam` has been created to manage setup the service user IAM policy/role. The module `bootstrap_ci` has been created to manage the resources that are created for the `boundary_ci` account.


ryancragun:
My feedback here directly maps to what I left on the Vault PR which does nearly the same thing. Perhaps we ought to update that first and do the same changes here and then revisit reviewing this one.

.github/workflows/enos-ci-bootstrap-ent.yml
```diff
@@ -0,0 +1,44 @@
+name: enos-ci-bootstrap-ent
```
ryancragun, line 1: We don't need to include ent or hcp stuff in the OSS repo. Better to merge the OSS stuff in here and create another PR for the ent/hcp stuff.

.github/workflows/enos-ci-bootstrap-ent.yml
```diff
@@ -0,0 +1,44 @@
...
+  workflow_call:
+
+jobs:
+  bootstrap-ci-ent:
+    env:
+      TF_WORKSPACE: "${{ github.event.repository.name }}-ci-enos-bootstrap"
+      TF_VAR_repository: ${{ github.event.repository.name }}
+      TF_VAR_aws_ssh_public_key: ${{ secrets.SSH_KEY_PUBLIC_CI }}
+      TF_TOKEN_app_terraform_io: ${{ secrets.TF_API_TOKEN }}
+    runs-on: ubuntu-latest
```
ryancragun, line 18: I know we'll be moving this file to another PR, but we'll need to change the runner to use self-hosted runners for private/internal accounts.

.github/workflows/enos-ci-bootstrap-ent.yml
```diff
@@ -0,0 +1,44 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Terraform
+        uses: hashicorp/setup-terraform@v2
+      - name: Assume service user role via Doormat
+        uses: hashicorp/doormat-action@main
+        with:
+          # This role, its associated IAM policy, the allowed Github workflow
```
ryancragun, line 26: Aren't we moving this into this repo with this PR?

.github/workflows/enos-ci-bootstrap.yml
```diff
@@ -0,0 +1,26 @@
...
+      - enos/ci/**
+      - .github/workflows/enos-ci-bootstrap*
+
+jobs:
+  bootstrap-ci-oss:
+    if: ${{ github.event.repository.name == 'boundary' }}
+    uses: ./.github/workflows/enos-ci-bootstrap-oss.yml
+    secrets: inherit
+
+  bootstrap-ci-ent:
```
ryancragun, line 23: I think we should move the ent job to another workflow in ent repo only.

ryancragun (commented):
Looking good! I think it just need to be broken into OSS and ENT/HCP only PRs.

.github/workflows/enos-ci-bootstrap-ent.yml
mikebaum: I removed the enterprise workflow

.github/workflows/enos-ci-bootstrap.yml
mikebaum: Done.

enos/ci/service-user-iam/service-quotas.tf
```diff
@@ -0,0 +1,60 @@
+locals {
+  subnets_per_vpc          = "L-F678F1CE"
```
ryancragun, line 2: Can we get some comments that explain what these are and why they're static?

ryancragun (approved):

PR #2630, opened 2022-11-17
fix(session): Always invoke all parts of session cancel trigger

johanbrandhorst:
The session cancelation trigger would not set the key_id to null appropriately if another one of the fields on the session was already null, since that case statement would be matched first. The new structure matches all statements, in case any of them have special logic (such as in the project case).


moduli (approved):

ddebko (approved):
LGTM

moduli:
Test results from `main` look good: https://github.com/hashicorp/boundary/pull/2631 👍

PR #2654, opened 2022-11-28
feat(targets): Addition of egress and ingress worker filters

irenarindos:
Partner PR to https://github.com/hashicorp/boundary-enterprise/pull/164


internal/db/schema/migrations/oss/postgres/59/01_target_ingress_egress_worker_filters.up.sql
```diff
@@ -0,0 +1,137 @@
...
+
+end;
+$$ language plpgsql;
+
+create function validate_filter_values_on_update() returns trigger
+as $$
+begin
+  if new.egress_worker_filter is not null then
+    if new.worker_filter = old.worker_filter then
+      new.worker_filter = null;
```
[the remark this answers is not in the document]
irenarindos: We went back and forth on it but decided to clear the old filter instead of making them clear it manually, some of that discussion is in the thread here: https://hashicorp.slack.com/archives/C038L63F67L/p1667242927405219

internal/tests/api/targets/classification_test.go
```diff
@@ -64,7 +64,7 @@ func TestTargets(t *testing.T) {
 					},
 					SessionMaxSeconds:           &wrapperspb.UInt32Value{Value: 0},
 					SessionConnectionLimit:      &wrapperspb.Int32Value{Value: 0},
-					WorkerFilter:                &wrapperspb.StringValue{Value: "worker-filter"},
+					EgressWorkerFilter:          &wrapperspb.StringValue{Value: "egress-worker-filter"},
```
jefferai, line 67: Do we actually want to remove WorkerFilter? Aren't we still returning that for a while?

internal/db/schema/migrations/oss/postgres/59/01_target_ingress_egress_worker_filters.up.sql
```diff
@@ -0,0 +1,137 @@
...
+  return new;
+
+end;
+$$ language plpgsql;
+
+create function validate_filter_values_on_update() returns trigger
+as $$
+begin
+  if new.egress_worker_filter is not null then
+    if new.worker_filter = old.worker_filter then
```
jefferai, line 30: What happens if these _aren't_ equivalent, e.g. if they do an update and both add ingress/egress fields but also update the existing worker filter? I think right now it will just accept that?

irenarindos: If a user tries to update both, they'll get a warning that 'cannot set worker_filter and ingress_filter; they are mutually exclusive fields' and nothing will happen (this is checked lower in the trigger func). The old filter can only be cleared out if they *only* update egress/ ingress without updating the worker filter

PR #2662, opened 2022-11-29
Onboard boundary to use new version/VERSION file and reproducible actions-go-build

emilymianeil:
# PR Description
This PR introduces the new [actions-set-product-version](https://github.com/hashicorp/actions-set-product-version), a tiny, but mighty, github action that acts as a bridge between the product repo and our new CRT feature: [automated version bumping](https://github.com/hashicorp/bob/commit/6813d9757c644679193a0af317e99570ac8cc848). This PR also introduces automatic version patch bumping via the bump-version {todo: enter link when merged} workflow.

tl;dr automated version bumping has a new command (bob update version) in the [bob CLI](https://github.com/hashicorp/bob) that automatically bumps the version to a new patch. This automation has been introduced to crt-workflows-common as a [new workflow](https://github.com/hashicorp/crt-workflows-common/pull/233) (with the new bob command)  and handles version bumping at the end of the release pipeline (after being released to production); for example, `1.0.0→1.0.1` and `1.0.0-dev→1.0.0`. Bumping the minor version (ie `1.0.x→1.1.0`) is only supported manually via `bob update version -bump minor`, but not supported in CRT (this work is upcoming).  This is made possible by adding the new event “bump-version-patch” in the ci.hcl file in this PR. 

What this small action does: 

-  Allows for the static version string from the `.release/VERSION` file to be read by the new CRT workflow and automagically be bumped to the next version (whether it be a minor, or patch, or major version bump). 
- Outputs an error if there’s no `VERSION` file in the `.release/ dir`
- Outputs an error if there’s no version string in the VERSION file
- Is able to parse `product_version` if it is `1.3.0-alpha1` as `1.3.0` (example: when `product_version = 1.3.0-alpha1`, `base_version = 1.3.0`)
- Is able to parse prerelease product versions such as `alpha1` (example `prerelease_product_version = alpha1`) in the statement above.  

In other words, the action grabs the version from the version file at the beginning of the build, then passes those versions (along with metadata, where necessary) to the new version bump automation CRT workflow at the end of the release pipeline to bump the product’s version. One thing to note is that the `version.go` and `base_version.go` files are used for local development and will be now overwritten by the ld-flags with the actual version that are being called in the go builds. 

This PR also introduces [actions-go-build](https://github.com/hashicorp/actions-go-build), an action that dynamically takes care of building and packaging go binaries. Through this action, you can define the go build, assert that it’s reproducible (optional*) and then use the artifacts in the workflow. (Once the build action run is finished, you can check out it the builds are reproducible by scrolling down on the summary page, [example](https://github.com/hashicorp/packer/actions/runs/3587803756#summary-9808859635)) 

*NOTE:  if [assert is enabled](https://github.com/hashicorp/actions-go-build#reproducibility-assertions), this could potentially create some disruptions in workflows as the action tries to verify two builds (primary build and ui builds) are the same (size, shasums) as ui builds might have non-deterministic compilation.

Passing tests ✔️  Check out these examples:

- https://github.com/hashicorp/consul-enterprise/actions/runs/3623272718

This PR also introduces a bump to the version tags for `actions/checkout`, `actions/setup-go`, and `actions/upload-artifact` (setup-go version tag update was required by the introduction of actions-go-build)-- however, if the Packer team feels like these should be isolated in their own PRs for easier/broader visibility, super happy to raise separate PRs! 

Finally, as a part of this effort to onboard new CRT tooling, we have implemented the new `prepare` workflow in the `ci.hcl` config file. This workflow simply wraps all workflows in the pre-stable channel in one workflow, to make your ci.hcl config file more readable for your team. The `prepare` runs `upload-dev`, `security-scan`, `notarize`, `sign`, `verify`, `promote-dev-docker`, and `fossa-scan` (depending on your team’s current needs) under the hood - no need to worry about these workflows anymore!

# TODO Before Merging
 - [x] Update version/VERSION to latest version
 - [x] Remove `set-product-version` branch from `build.yml` workflow dispatch list
 - [x] Remove `set-product-version` branch from `ci.hcl` allowlist


.release/VERSION
```diff
@@ -0,0 +1 @@
+0.0.0-dev
```
emilymianeil, line 1: TODO (before merge): Update with actual current version on main

internal/cmd/commands/version/version.go
```diff
@@ -67,9 +67,6 @@ func (c *Command) Run(args []string) int {
 	if verInfo.Version != "" {
 		nonAttributeMap["Version Number"] = verInfo.VersionNumber()
 	}
-	if verInfo.VersionMetadata != "" {
```
sarahethompson, line 70: I think we want to keep `VersionMetadata` here?

tmessi (changes requested):

emilymianeil:
### Manual Test for Prepare Workflow: PASSED ✅ 
We need to verify the most critical parts of the prepare workflow: notarization, codesigning (running in circle-codesign CI, and verified through verifying correct notarization jobs were run), and checksums file.
1. Verify prepare workflow has passed:
      - [X] Prepare: [passed](https://github.com/hashicorp/crt-workflows-common/actions/runs/3643308670) 
2. Verify correct notarization has run + passed:
      - [X] Darwin AMD64: [passed](https://github.com/hashicorp/crt-workflows-common/actions/runs/3643308670/jobs/6151590245) 
      - [X] Darwin ARM64: [passed](https://github.com/hashicorp/crt-workflows-common/actions/runs/3643308670/jobs/6151590644)
      - [X] Windows 386: [passed](https://github.com/hashicorp/crt-workflows-common/actions/runs/3643308670/jobs/6151590416) 
      - [X] Windows AMD64: [passed](https://github.com/hashicorp/crt-workflows-common/actions/runs/3643308670/jobs/6151590510) 
3. Verify the checksums file to ensure all binaries are present
      - [X] Download SHASUMS from stable bucket: [passed](https://artifactory.hashicorp.engineering/artifactory/hashicorp-crt-stable-local/boundary/0.12.0/5fd4446af771ae6df5284b112c07fabe8e558a46/boundary_0.12.0_SHA256SUMS)
4. Verify version from downloaded binary:
```
➜  Downloads ./boundary\ 3 version

Version information:
  Git Revision:        07a823c5202eb9ab9f9fcb4f47d46b82b0f5fcc9
  Version Number:      0.12.0
```

scripts/build.sh
```diff
@@ -64,8 +64,15 @@ echo "==> Building into bin/ for ${GOOS}_${GOARCH}..."
...
-    -o "bin/${BINARY_NAME}" \
+    -trimpath \
+    -buildvcs=false \
+    -ldflags "
+      -X github.com/hashicorp/boundary/version.GitCommit=${GIT_COMMIT}${GIT_DIRTY}
+      -X 'github.com/hashicorp/boundary/version.Version=$PRODUCT_VERSION'
+      -X 'github.com/hashicorp/boundary/version.VersionPrerelease=$PRERELEASE_PRODUCT_VERSION'
+      -X 'github.com/hashicorp/boundary/version.VersionMetadata=$METADATA_PRODUCT_VERSION'
+      " \
+    -o "$BIN_PATH" \
```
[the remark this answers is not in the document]
emilymianeil: It's an environment variable in the `actions-go-build` action - I can definitely set a default though to make it work locally!

.github/workflows/build.yml
```diff
@@ -7,13 +7,26 @@ on:
 
 env:
   PKG_NAME: "boundary"
+  PRODUCT_EDITION: "oss"
```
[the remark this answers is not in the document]
emilymianeil: We've decided to implement the `make edition` (and using our `set-product-version` action).

scripts/build.sh
```diff
@@ -60,8 +60,18 @@ rm -f bin/*
 mkdir -p bin/
```
tmessi, line 60: These `rm` and `mkdir` should be updated to use `${BIN_PARENT_DIR}`, so you might want to move that logic up.

scripts/build.sh
```diff
@@ -60,8 +60,18 @@ rm -f bin/*
...
 # Build!
-echo "==> Building into bin/ for ${GOOS}_${GOARCH}..."
 BINARY_NAME="boundary${BINARY_SUFFIX}"
+
+if ["${BIN_PATH}" == ""]; then
+    echo "==> Setting binary path to bin/${BINARY_NAME}"
+    BIN_PATH=bin/${BINARY_NAME}
+else
+    echo "==> Binary path set to ${BIN_PATH}"
+fi
```
tmessi, line 70: This can be done via a default assignment, ie:

```bash
BIN_PATH=${BIN_PATH:=bin/${BINARY_NAME}}
```

enos/modules/build_local/templates/build.sh
```diff
@@ -9,7 +9,6 @@ npm install --global yarn || true
 # Go to the root of the boundary repo
 root_dir="$(git rev-parse --show-toplevel)"
 pushd "${root_dir}" > /dev/null
-
 make build-ui
```
tmessi, line 12: looks like something is still changing here that doesnt need to.

version/version_base.go
```diff
@@ -8,6 +8,7 @@ var (
 	// Whether cgo is enabled or not; set at build time
 	CgoEnabled bool
 
+
```
tmessi, line 11: This line should be removed and is causing a CI failure due to a `gofumpt` diff. You can just run `make fmt` to fix it.

tmessi (changes requested):
Looking good, couple minor things and a CI failure to fix though.

tmessi (approved):

PR #2658, opened 2022-11-29
Backport of Document vault credential library templating into release/0.11.x

hc-github-team-secure-boundary:
## Backport

This PR is auto-generated from #2635 to be assessed for backporting due to the inclusion of the label backport/0.11.x.

The below text is copied from the body of the original PR.

---

Adds information about the Vault credential parameter templating function being added to Boundary in v 0.11.1.

~~[View the preview environment here.](https://boundary-8yzjtehvr-hashicorp.vercel.app/boundary/docs/concepts/domain-model/credential-libraries#vault-credential-library-parameter-templating)~~

[View the new preview here.](https://boundary-55mew97ph-hashicorp.vercel.app/boundary/docs/concepts/domain-model/credential-libraries#vault-credential-library-parameter-templating)

PR #2635, opened 2022-11-21
Document vault credential library templating

Dan-Heath:
Adds information about the Vault credential parameter templating function being added to Boundary in v 0.11.1.

~~[View the preview environment here.](https://boundary-8yzjtehvr-hashicorp.vercel.app/boundary/docs/concepts/domain-model/credential-libraries#vault-credential-library-parameter-templating)~~

[View the new preview here.](https://boundary-55mew97ph-hashicorp.vercel.app/boundary/docs/concepts/domain-model/credential-libraries#vault-credential-library-parameter-templating)

PR #2663, opened 2022-11-29
Document vault credential library templating (#2635)

Dan-Heath:
Recreating PR #2658 but to target the `main` branch.


Dan-Heath:
Opened in error

PR #2664, opened 2022-11-29
Document vault credential library templating (#2635)

Dan-Heath:
Cherry pick to add backport to the current latest branch (since it was published while we were trying to fix the backport).


Dan-Heath:
No longer necessary

PR #2661, opened 2022-11-29
Revert "feat(metric): adds active connections gauge metric for controller and worker clusters"

talanknight:
Reverts hashicorp/boundary#2593

PR #2668, opened 2022-12-02
feat(metric): adds accepted/closed connections counters to worker and controller servers

A-440Hz:
reverts the #2661 revert


A-440Hz:
the bug was in L218 of the worker listeners.go file: the listener in `httpServer.Serve()` has to return a *tls.Conn

A-440Hz:
[diff.txt](https://github.com/hashicorp/boundary/files/10143476/diff.txt)

PR #2674, opened 2022-12-05
test(e2e): Minor clerical updates to key destruction tests

moduli:
This PR updates the test names for the key destruction tests to match the current convention as well as adds some logging to help with debugging and clarity that something is happening when running locally.


johanbrandhorst (approved):
Thanks!

PR #2675, opened 2022-12-05
fix(cmd): boundary connect SSH depending on a host id for ssh aliasing

hugoamvieira:
Targets are not guaranteed to have Host Ids anymore, so we now use Target Id when a Host Id is not present.


johanbrandhorst (commented):
Could we add a e2e test that exercises this?

ddebko (approved):
LGTM

hugoamvieira:
> Could we add a e2e test that exercises this?

Yep! This will be done as part of our End-to-End suite of tasks :)

PR #2627, opened 2022-11-17
test(e2e): Add database migration test

moduli:
This PR adds an e2e test to validate database migrations, which does the following...
- Sets up a boundary database using the latest released version of Boundary
- Populates the boundary database with a number of resources (org, project, host catalog, host set, target, static credentials, an AWS dynamic host catalog, and some vault credentials)
- Uses the newest version of Boundary to migrate the database

It accomplishes this through the use of docker containers via the `dockertest` library. This PR also adds a corresponding enos scenario so that it can be run in CI. 

## Testing Notes
- [x] Confirm that the test runs successfully locally
- [x] Confirm that the test runs successfully in CI
- [x] Confirm that if migration fails, an error is returned and the test would report it.
<details>
    <summary>Passing Log</summary>

```
❯ go test -v github.com/hashicorp/boundary/testing/internal/e2e/tests/database -count=1
=== RUN   TestDatabaseMigration
    docker.go:25: Starting postgres database...
    docker.go:169: Starting openssh-server to serve as target...
    docker.go:76: Initializing postgres database...
    docker.go:103: Starting Boundary...
    migration_test.go:105: Using Boundary Version: 
        Version information:
          Git Revision:        e4372a7e1c7bb16e72239460d12492a560e602c8
          Version Number:      0.11.1
        
    scope.go:57: Created Org Id: o_tCF02PGzxc
    scope.go:80: Created Project Id: p_XnSccDSc2I
    host.go:82: Created Host Catalog: hcst_P6HQCsnW6B
    host.go:102: Created Host Set: hsst_GR42AV2kpW
    host.go:124: Created Host: hst_VDkKEiV2nU
    target.go:59: Created Target: ttcp_5nCl1IgTsc
    host.go:159: Created Host Catalog: hcplg_sxoU0nUvg3
    host.go:181: Created Host Set: hsplg_RpBUl0il0W
    host.go:188: Looking for items in the host set...
    host.go:219: No items are appearing in the host set. Retrying...
    host.go:214: Found 1 hosts
    credential.go:64: Created Credential Store: csst_yM57cPeotd
    credential.go:111: Created Username/Password Credentials: credup_fbNKeHRDxL
    credential.go:133: Created Username/Password Credentials: credjson_k2QzOGTcQS
    credential.go:87: Created SSH Private Key Credentials: credspk_KfYxNiYvfn
    migration_test.go:177: Created Vault Credential
    migration_test.go:195: Created Vault Cred Store Token
[...]

PR #2676, opened 2022-12-06
refact(e2e): Changes to support an automated migration test

moduli:
This is the 1st of 2 PRs to support the new automated migration test (A draft PR containing all of the changes can be found here: https://github.com/hashicorp/boundary/pull/2627). The migration test intends to do the following
- Set up a boundary database using the latest released version of Boundary
- Populate the boundary database with a number of resources (org, project, host catalog, host set, target, static credentials, an AWS dynamic host catalog, and some vault credentials)
- Use the newest version of Boundary to migrate the database (i.e. `boundary database migrate`)

This PR does some refactoring to make creating the migration test easier. These changes were split off to simplify the review process. The changes include the following
- Cleanup calls were moved out of methods and into each test. This makes it clearer where cleanup steps are happening when you're reading a test (as opposed to it being hidden in a method call).
- Additional methods were created for commonly used actions to help reduce the amount of code in each test. These methods will be used in the migration test to help populate the database with entries
- Previously, we were checking the environment variable `E2E_PASSWORD_AUTH_METHOD_ID` was set in order to determine if the e2e tests should run. The migration test will not have this variable set because it doesn't need a boundary instance to be spun up beforehand, so we needed a different way to denote an e2e test.

### DIFF
--- a/enos/modules/test_e2e/main.tf
+++ b/enos/modules/test_e2e/main.tf
@@ -99,7 +99,7 @@ locals {
 
 resource "enos_local_exec" "run_e2e_test" {
   environment = {
-    E2E_TESTS                     = true,
+    E2E_TESTS                     = "true",
     BOUNDARY_ADDR                 = var.alb_boundary_api_addr,
     E2E_PASSWORD_AUTH_METHOD_ID   = var.auth_method_id,
     E2E_PASSWORD_ADMIN_LOGIN_NAME = var.auth_login_name,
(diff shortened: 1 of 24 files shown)


enos/modules/test_e2e/main.tf
```diff
@@ -98,6 +99,7 @@ locals {
 
 resource "enos_local_exec" "run_e2e_test" {
   environment = {
+    E2E_TESTS                     = true,
```
johanbrandhorst, line 102: What does it mean to set this to a boolean value? Should this be
```suggestion
    E2E_TESTS                     = "true",
```

johanbrandhorst (approved):

PR #2678, opened 2022-12-06
Refactor proxy handler to separate connection and protocol handling

talanknight:
This updates the AuthorizeConnection response to include protocol information specific to the connection being created.
TCP protocol proxying does not require any additional information since it is a simple pass through proxy.  

This PR also passes in the client originating connection into the protocol handler and a dialer that can be used to dial down to the target's host source address which ideally maintains a separation of logic between establishing the proxy connections and the logic associated with doing protocol intelligent things in the protocol handlers.


irenarindos (approved):

louisruch (approved):
Great work :shipit:

internal/daemon/worker/handler.go
```diff
@@ -258,35 +247,66 @@ func (w *Worker) handleProxy(listenerCfg *listenerutil.ListenerConfig, sessionMa
...
+		if err != nil {
+			event.WriteError(ctx, op, err, event.WithInfoMsg("worker failed to parse target endpoint", "endpoint", sess.GetEndpoint()))
+			if err = conn.Close(websocket.StatusProtocolError, "unable to parse endpoint"); err != nil {
 				event.WriteError(ctx, op, err, event.WithInfoMsg("error closing client connection"))
 			}
 			return
 		}
+		protocolCtx := ci.GetProtocolContext()
+		if protocolCtx == nil {
+			// TODO: Remove this if block once pre v0.12.0 controllers are no longer supported.
```
jefferai, line 260: Might be ideal to put in a task for 0.15 or so in to remind us to remove this.

PR #2683, opened 2022-12-06
feat: update target domain to support direct address association

ddebko:
Summary:

The following changes enable the user to attach a network address directly to a target resource. A network address can be directly assigned to a target when initially creating the resource or it can be set later by calling the Update method in the target repository.

A target to host source relationship is mutually exclusive from a target to network address association. When trying to assign a network address directly to a target that has an existing host source resource, an error should be returned to the user. When trying to assign a host source to a target (set or add methods) that has an existing network address association, an error should be returned to the user.


internal/target/tcp/repository_tcp_target_test.go
```diff
@@ -201,9 +219,51 @@ func TestRepository_UpdateTcpTarget(t *testing.T) {
...
+				fieldMaskPaths: []string{"Address"},
+				ProjectId:      proj.PublicId,
+				address:        "null",
+			},
+			newProjectId:    proj.PublicId,
+			newTargetOpts:   []target.Option{target.WithAddress("8.8.8.8")},
+			wantErr:         false,
+			wantRowsUpdate:  1,
+			wantHostSources: false,
+		},
```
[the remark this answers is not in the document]
ddebko: This test case should validate that the address association was deleted from a Target.

internal/target/repository.go
```diff
@@ -493,6 +464,113 @@ func (r *Repository) CreateTarget(ctx context.Context, target Target, opt ...Opt
...
+		db.ExpBackoff{},
+		func(reader db.Reader, w db.Writer) error {
+			hostSources, err := fetchHostSources(ctx, reader, target.GetPublicId())
+			if err != nil {
+				return errors.Wrap(ctx, err, op)
+			}
+			if len(hostSources) > 0 {
+				return errors.New(ctx, errors.Forbidden, op, "unable to set address because one or more host sources is assigned to the given target")
+			}
+			if address.GetAddress() == "null" {
```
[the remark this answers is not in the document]
ddebko: You are correct, the code is wrong. I fixed this issue and added a test to validate both setting the address to null and the ability to delete an address.

internal/target/repository.go
```diff
@@ -537,28 +630,115 @@ func (r *Repository) UpdateTarget(ctx context.Context, target Target, version ui
...
+		switch {
+		case strings.EqualFold("Address", f):
+		default:
+			filteredDbMask = append(filteredDbMask, f)
+		}
+	}
+	for _, f := range nullFields {
+		switch {
+		case strings.EqualFold("Address", f):
+			address = "null"
```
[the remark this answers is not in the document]
ddebko: I changed this behavior to set two different boolean flags instead of setting the address variable to "null". The boolean flags are set when the Address field appears in the nullField or the maskField. Do you like this approach or do you prefer something else?

internal/target/tcp/repository_host_source_test.go
```diff
@@ -461,6 +486,17 @@ func TestRepository_SetTargetHostSets(t *testing.T) {
...
+			name: "address-mutually-exclusive-relationship",
+			args: args{
+				target:               tcp.TestTarget(ctx, t, conn, proj.PublicId, "invalid-host-source", target.WithAddress("8.8.8.8")),
+				targetVersion:        2, // yep, since setupFn will increment it to 2
+				hostSourceIds:        []string{testHostSetIds[0], testHostSetIds[1]},
+				addToOrigHostSources: true,
+			},
+			wantErr:    true,
+			wantErrMsg: "unable to set host sources because a network address is directly assigned to the given target",
+		},
```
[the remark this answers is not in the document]
ddebko: I added a test case for this situation.

internal/target/repository.go
```diff
@@ -336,80 +363,6 @@ func (r *Repository) DeleteTarget(ctx context.Context, publicId string, _ ...Opt
...
-			}
-			return nil
-		},
-	)
-	if err != nil {
-		return nil, nil, nil, db.NoRowsAffected, errors.Wrap(ctx, err, op)
-	}
-	return returnedTarget.(Target), hostSources, credSources, rowsUpdated, nil
-}
-
```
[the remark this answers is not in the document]
ddebko: This function was removed after a refactor to flatten the the nested functions inside of the db transactions in the UpdateTarget function.

internal/target/address.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/oplog"
+	"github.com/hashicorp/boundary/internal/target/store"
+)
+
+const (
+	DefaultTargetAddressTableName = "target_address"
+)
+
+// A Address represents the relationship between a target and a
```
johanbrandhorst, line 17: Nit 😁 
```suggestion
// An Address represents the relationship between a target and a
```

internal/target/repository_address.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+func fetchAddress(ctx context.Context, r db.Reader, targetId string) (*Address, error) {
+	const op = "target.fetchAddress"
+	var address *Address
+	if err := r.SearchWhere(ctx, &address, "target_id = ?", []any{targetId}); err != nil {
+		return nil, errors.Wrap(ctx, err, op)
+	}
+	if address.TargetAddress == nil {
+		return nil, nil
```
johanbrandhorst, line 17: I know this pattern is pervasive in our code, but I think it's demonstrably dangerous (message me for details), and I would much sooner we return an error in these cases. What do you think?

johanbrandhorst (changes requested):

internal/target/repository.go
```diff
@@ -504,18 +475,30 @@ func (r *Repository) UpdateTarget(ctx context.Context, target Target, version ui
...
-	if target.GetPublicId() == "" {
-		return nil, nil, nil, db.NoRowsAffected, errors.New(ctx, errors.InvalidParameter, op, "missing target public id")
+	projectId := target.GetProjectId()
+	if projectId == "" {
+		t := allocTargetView()
+		t.PublicId = target.GetPublicId()
+		if err := r.reader.LookupByPublicId(ctx, &t); err != nil {
+			return nil, nil, nil, db.NoRowsAffected, errors.Wrap(ctx, err, op, errors.WithMsg(fmt.Sprintf("lookup failed for %s", t.PublicId)))
+		}
+		projectId = t.ProjectId
```
hugoamvieira, line 498: We should keep the API consistent with the other methods. We only need this projectId to grab the oplog, but if you check for example `CreateTarget`, we error if there's no projectId straight away (like we did here before this change), and then we just grab the project id from the target and pass it onto the kms wrapper func.

```go
	oplogWrapper, err := r.kms.GetWrapper(ctx, target.GetProjectId(), kms.KeyPurposeOplog)
	if err != nil {
		return nil, nil, nil, errors.Wrap(ctx, err, op, errors.WithMsg("unable to get oplog wrapper"))
	}
```

internal/target/repository.go
```diff
@@ -537,28 +523,130 @@ func (r *Repository) UpdateTarget(ctx context.Context, target Target, version ui
...
+				return errors.Wrap(ctx, err, op, errors.WithMsg("unable to get ticket"))
+			}
+
+			var tOplogMsg oplog.Message
+			rowsUpdated, err = w.Update(ctx, t, filteredDbMask, filteredNullFields, db.NewOplogMsg(&tOplogMsg), db.WithVersion(&version))
 			if err != nil {
 				return errors.Wrap(ctx, err, op)
 			}
+			if rowsUpdated > 1 {
+				return errors.New(ctx, errors.MultipleRecords, op, fmt.Sprintf("updated target version and %d rows updated", rowsUpdated))
```
hugoamvieira, line 592: Here the transaction will rollback and nothing will actually be updated, so it's better to just return something like `more than 1 resource would have been updated`

internal/target/repository.go
```diff
@@ -537,28 +523,130 @@ func (r *Repository) UpdateTarget(ctx context.Context, target Target, version ui
...
+				var err error
+				address = allocTargetAddress()
+				address.TargetAddress.TargetId = t.GetPublicId()
+				rowsUpdated, err = w.Delete(ctx, address, db.NewOplogMsg(&aOplogMsg))
+				if err != nil {
+					return errors.Wrap(ctx, err, op, errors.WithMsg("unable to delete target address"))
+				}
+				if rowsUpdated > 1 {
+					return errors.New(ctx, errors.MultipleRecords, op, fmt.Sprintf("delete target address and %d rows deleted", rowsUpdated))
+				}
```
hugoamvieira, line 632: Same here

hugoamvieira (commented):
Looks awesome, just a few comments from me 👇

internal/target/repository.go
ddebko: Yea, this makes sense to me because the scopeId is a required field so it shouldn't be missing. I will update the code based on your suggestion.

internal/target/repository.go
ddebko: I like your suggested error message a lot more, I will update the code.

internal/target/repository.go
ddebko: One issue, there is a unit test that actually expects no error message back when the projectId is missing and that's because we fetch the projectId value. The unit test is called `TestRepository_UpdateTcpTarget/empty-project-id-with-name-mask`. I will dig into this and figure out if this unit test case is outdated and can be removed.

ddebko: I went ahead and deleted the unit test because I think it was added by mistake and overlooked in the original PR.

hugoamvieira: Sounds good!

hugoamvieira (approved):

PR #2684, opened 2022-12-06
feat: update session domain to support using address directly associated to a target

ddebko:
Summary:

A host_id and host_set_id are no longer required arguments when authorizing a session. The session endpoint is now derived from either a host or a direct network address association to a Target resource. If a network address association is deleted from a Target, then all sessions using the same Target will be canceled. The behavior of canceling a session when a resource is deleted follows the existing pattern for host and host set. A session will not be canceled if the network address value is changed.

### DIFF
--- a/internal/session/repository_session.go
+++ b/internal/session/repository_session.go
@@ -34,20 +34,17 @@ func (r *Repository) CreateSession(ctx context.Context, sessionWrapper wrapping.
 	if newSession.TargetId == "" {
 		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing target id")
 	}
-	if newSession.HostId == "" {
-		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing host id")
-	}
 	if newSession.UserId == "" {
 		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing user id")
-	}
-	if newSession.HostSetId == "" {
-		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing host set id")
 	}
 	if newSession.AuthTokenId == "" {
 		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing auth token id")
 	}
 	if newSession.ProjectId == "" {
 		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing project id")
+	}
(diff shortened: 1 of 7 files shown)


internal/session/repository_session.go
```diff
@@ -197,6 +214,13 @@ func (r *Repository) LookupSession(ctx context.Context, sessionId string, opt ..
 				session.StaticCredentials = staticCreds
 			}
 
+			sessionHostSetHost := AllocSessionHostSetHost()
+			if err := read.LookupWhere(ctx, sessionHostSetHost, "session_id = ?", []interface{}{sessionId}); err != nil && !errors.IsNotFoundError(err) {
```
johanbrandhorst, line 218:
```suggestion
			if err := read.LookupWhere(ctx, sessionHostSetHost, "session_id = ?", []any{sessionId}); err != nil && !errors.IsNotFoundError(err) {
```

internal/session/session_host_set_host.go
```diff
@@ -0,0 +1,68 @@
...
+	s.tableName = n
+}
+
+// AllocSessionHostSet will allocate a SessionHostSetHost
+func AllocSessionHostSetHost() *SessionHostSetHost {
+	return &SessionHostSetHost{}
+}
+
+// Clone creates a clone of the SessionHostSetHost
+func (s *SessionHostSetHost) Clone() interface{} {
```
johanbrandhorst, line 61:
```suggestion
func (s *SessionHostSetHost) Clone() any {
```

internal/session/session_target_address.go
```diff
@@ -0,0 +1,62 @@
...
+	s.tableName = n
+}
+
+// AllocSessionTargetAddress will allocate a SessionTargetAddress
+func AllocSessionTargetAddress() *SessionTargetAddress {
+	return &SessionTargetAddress{}
+}
+
+// Clone creates a clone of the SessionTargetAddress
+func (s *SessionTargetAddress) Clone() interface{} {
```
johanbrandhorst, line 56:
```suggestion
func (s *SessionTargetAddress) Clone() any {
```

hugoamvieira (approved):
LGTM!

johanbrandhorst (approved):

PR #2685, opened 2022-12-06
feat: update session & target handlers to support direct address association

ddebko:
Summary:

The controller handlers have been updated to support returning an Address value if it is directly associated with a Target resource when returning a Target Item for the following actions: create, list, read, & update. The target update handler allows the user to set an address value or delete it by setting the value to "null". The behavior of using the value "null" on an argument flag for an update action follows existing patterns in the code, such as the vault credential library override mappings.


internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -766,25 +770,32 @@ func (s Service) AuthorizeSession(ctx context.Context, req *pbs.AuthorizeSession
...
-		if t.GetDefaultPort() == 0 {
-			return nil, errors.New(ctx, errors.InvalidParameter, op, fmt.Sprintf("neither the selected host %q nor the target provides a port to use", chosenEndpoint.HostId))
+	if chosenEndpoint != nil {
+		hostId = chosenEndpoint.HostId
+		hostSetId = chosenEndpoint.SetId
+		h, p, err = net.SplitHostPort(chosenEndpoint.Address)
+		switch {
+		case err != nil && strings.Contains(err.Error(), missingPortErrStr):
+			if t.GetDefaultPort() == 0 {
+				return nil, errors.New(ctx, errors.InvalidParameter, op, fmt.Sprintf("neither the selected host %q nor the target provides a port to use", chosenEndpoint.HostId))
```
johanbrandhorst, line 785: Could this error be encountered by a user? If so, maybe we should use `handlers.ApiErrorWithCodeAndMessage`?

internal/proto/controller/api/resources/targets/v1/target.proto
```diff
@@ -188,8 +188,15 @@ message Target {
   // Output only. The available actions on this resource for this user.
   repeated string authorized_actions = 300 [json_name = "authorized_actions"]; // @gotags: `class:"public"`
 
-  // Optionally, a valid network address to connect to for this target. Can not be used alongside host sources.
-  string address = 540 [(custom_options.v1.generate_sdk_option) = true]; // @gotags: `class:"public"`
+  // Optional boolean expression to filter the workers that are allowed to satisfy this request.
```
johanbrandhorst, line 191: Pretty sure this comment is wrong

johanbrandhorst (changes requested):

internal/daemon/controller/handlers/targets/target_service.go
ddebko: This error message was not written by me, I only reorganized the code to better handle when we need to parse from a chosen host resource or a network address that is directly associated to a target. I'm happy to change it though.

internal/proto/controller/api/resources/targets/v1/target.proto
ddebko: correct the description to " Optional string value that represents a network resource and is used when establishing a session."

johanbrandhorst (approved):
Would this be a good feature for a new e2e test perhaps? CC @moduli

moduli:
> Would this be a good feature for a new e2e test perhaps? CC @moduli

I believe Hugo and Damian have tasks for that? Correct me if I'm wrong.

hugoamvieira (commented):
Looks good! I have one last file to take a look at, but in the meantime I just had a Q ⬇️

hugoamvieira:
@moduli @johanbrandhorst Yep! Our end-to-end tasks cover creating, removing, updating and the mutual exclusivity aspect of this :)

internal/daemon/controller/handlers/sessions/session_service.go
```diff
@@ -235,6 +235,8 @@ func (s Service) CancelSession(ctx context.Context, req *pbs.CancelSessionReques
 	if ses == nil {
 		return nil, handlers.NotFoundErrorf("Session %q doesn't exist.", req.GetId())
 	}
+	hostId := ses.HostId
+	hostSetId := ses.HostSetId
```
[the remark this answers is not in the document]
ddebko: I believe the context of the session host id and host set id are lost when canceling because the values are now stored in session_host_set_host. I will double check if there is something missing when calling cancel twice with the skip flag to will help us avoid setting these values manually.

johanbrandhorst (approved):

internal/daemon/controller/handlers/sessions/session_service.go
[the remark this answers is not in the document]
ddebko: I think that I found an alternative solution, there is a function called `updateState` in the session repository that mutates the Session object when updating the row in the sessions table. This is when we lost the values for host_id and host_set_id. I can do a fetch those values in updateState.

PR #2688, opened 2022-12-07
fix(db): Reorder migrations

johanbrandhorst:
Migration numbers need to be consistent between version upgrades. The migrations for the key lifecycle management were merged into v0.11.1 as number 56, but are currently 58 in main. This reorders them down to 56, fixing an issue when migrating from 0.11.1 to main.

I've manually verified this fixes the issue, and it should be further verified by the tests introduced in #2627.


internal/db/schema/migrations/oss/postgres_56_01_test.go
```diff
@@ -2,26 +2,23 @@ package oss_test
...
 	"github.com/stretchr/testify/require"
 )
 
-func TestMigrations_DeleteOrphanedAccounts(t *testing.T) {
+func TestMigrations_NewForeignKeys(t *testing.T) {
 	const (
-		priorMigration   = 55001
-		currentMigration = 56001
+		priorMigration   = 56001
+		currentMigration = 56002
```
moduli, line 21: Excuse my lack of understanding here, but why wouldn't this be `55001 -> 56001`?

johanbrandhorst: Discussed on call, this is correct.

moduli (approved):

PR #2695, opened 2022-12-08
refactor(metric): adds exclude and include maps to metric init function 

A-440Hz:
* adds exclude and include filters to initialize grpc metric function
* ensures that the same metrics belong to the same subsystems after proto file revert in commit 0f89b1

### DIFF
--- a/internal/daemon/metric/initialize_metrics.go
+++ b/internal/daemon/metric/initialize_metrics.go
@@ -33,7 +33,18 @@ var (
 
 /* The following methods are used to initialize Prometheus histogram vectors for gRPC connections. */
 
-func rangeProtoFiles(m map[string][]string, fd protoreflect.FileDescriptor) bool {
+func contains(r []string, n string) bool {
+	for _, e := range r {
+		if n == e {
+			return true
+		}
+	}
+	return false
+}
+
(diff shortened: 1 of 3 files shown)


A-440Hz:
[diff.txt](https://github.com/hashicorp/boundary/files/10181563/diff.txt)
diff shows no changes

internal/daemon/metric/initialize_metrics.go
```diff
@@ -33,7 +33,18 @@ var (
 
 /* The following methods are used to initialize Prometheus histogram vectors for gRPC connections. */
 
-func rangeProtoFiles(m map[string][]string, fd protoreflect.FileDescriptor) bool {
+func contains(r []string, n string) bool {
```
louisruch, line 36: You could also make use of the go-secure-stdlibs we have created for this https://github.com/hashicorp/go-secure-stdlib/blob/main/strutil/strutil.go#L26

A-440Hz: ah thanks, I was wondering if there was a better way to do this

A-440Hz:
[diff.txt](https://github.com/hashicorp/boundary/files/10190848/diff.txt)

PR #2699, opened 2022-12-09
Upgrade Go

johanbrandhorst:
Go version 1.19.4 contains a security fix that affects all HTTP/2 servers


tmessi (approved):
IDK if I would consider the first commit a "fix", since it isn't fixing a bug, its just a nice improvement.

PR #2705, opened 2022-12-12
Dheath-fix-kms-notes

Dan-Heath:
* Add initial changelog entry

* feat(docs): Add initial stab at docs for KMS rotation/destruction

* fixup! Add initial changelog entry

* Review comments


Dan-Heath:
Opened in error.

PR #2707, opened 2022-12-12
docs: Delete roadmap topic

Dan-Heath:
This roadmap topic was one of the topics we thought was hidden. Although it doesn't appear on the TOC, users could still find this via search. Since we don't want anyone to see this outdated topic, this PR deletes it.

I don't see any internal links to this topic in the `boundary` repo. I created a redirect, however, just in case someone has a direct URL to access the roadmap page. For lack of a similar roadmap topic, I redirected the URL to the **_What is Boundary?_** topic.

[View the preview environment here.](https://boundary-phosn40jz-hashicorp.vercel.app/boundary/docs)

NOTE: Not sure why, but the roadmap topic is still showing up in search in the preview environment. Clicking it here invokes the redirect (so that's working properly):

image


stellarsquall (approved):
this looks great!

I think there is some caching server-side of recent search results, so I wouldn't worry about that for now. Hopefully that will catch up a little while after the changes are published and the cache is cleaned

Dan-Heath:
> this looks great!
> 
> I think there is some caching server-side of recent search results, so I wouldn't worry about that for now. Hopefully that will catch up a little while after the changes are published and the cache is cleaned

Makes sense. Thanks Robin!!

PR #2689, opened 2022-12-07
test(e2e): Add tests to validate that session ends after host source is deleted

moduli:
This PR adds two more e2e tests to validate that an active session ends when the corresponding host set or host is deleted. The idea for this came about during the key destruction work. There will be future PRs to add additional tests for deleting other resource types.

PR #2709, opened 2022-12-12
test(e2e): Add tests to validate that session ends after target is deleted

moduli:
This PR is a continuation of this one: https://github.com/hashicorp/boundary/pull/2689

It adds another test to validate that an active session ends when the corresponding target is deleted. The idea for this came about during the key destruction work.


testing/internal/e2e/tests/static/session_end_delete_project_test.go
```diff
@@ -0,0 +1,133 @@
...
+		// `boundary connect` returns a 255 when cancelled
+		require.Equal(t, output.ExitCode, 255, string(output.Stdout), string(output.Stderr))
+	case <-time.After(time.Second * 5):
+		t.Fatal("Timed out waiting for session command to exit")
+	}
+
+	err = backoff.RetryNotify(
+		func() error {
+			// Check that session cannot be read
+			output = e2e.RunCommand(ctx, "boundary",
```
moduli, line 110: Note: When testing, the session can no longer be read once the project is deleted, so instead, this test checks that you get a 403 here.

johanbrandhorst: I believe the session should outlive the project 🤔. It should be canceled, naturally, but it should outlive it. Maybe worth investigating more?

johanbrandhorst: If you're getting 403, is it because you're logged in as a user that was within the project that got deleted?

johanbrandhorst (approved):

testing/internal/e2e/tests/static/session_end_delete_project_test.go
moduli: I'll investigate more. I thought I tried it locally with the same result (ensuring that I was logged in as an admin). I wanted to double check what you thought was the expected behavior.

moduli: Turns out there may be an issue here. Reverting the test back to its original form.

PR #2716, opened 2022-12-13
docs: Fix redirect syntax

Dan-Heath:
A recently merged redirect is not working and I believe it is because I left a slash out. This PR adds the slash.


Dan-Heath:
Opened in error.

PR #2717, opened 2022-12-13
docs: Remove "Overview" titles

Dan-Heath:
There are several topics titled "Overview" in the Boundary navigation. This is a somewhat unhelpful title for users. This PR attempts to provide more meaningful topic titles.


Dan-Heath:
I spoke to Yoko about this, and she pointed out that this is a standard across the docs, due to the need for an index file in each folder grouping. I am going to close this PR for now, while I research how index files work and whether they are still necessary.

PR #2719, opened 2022-12-14
chore(e2e): Add tparse to summarize test output

moduli:
This PR adds the tparse tool to include a table summary of test results at the end. This is the same tool used by unit tests.

Here's an example log: https://github.com/hashicorp/boundary/actions/runs/3699246273/jobs/6266548822

The standard format didn't look nice...
Screen Shot 2022-12-14 at 5 57 15 PM

... so this PR is currently opting to use the plain format
Screen Shot 2022-12-14 at 5 57 05 PM

### DIFF
--- a/.github/workflows/enos-run.yml
+++ b/.github/workflows/enos-run.yml
@@ -40,6 +40,8 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version: ${{ inputs.go-version }}
+      - name: Install tools to get tparse
+        run: make tools
       - name: Set up Terraform
         uses: hashicorp/setup-terraform@v2
         with:
--- a/enos/modules/test_e2e/main.tf
+++ b/enos/modules/test_e2e/main.tf
@@ -122,7 +122,7 @@ resource "enos_local_exec" "run_e2e_test" {
     E2E_AWS_HOST_SET_IPS2         = local.aws_host_set_ips2
   }
 
-  inline = ["PATH=\"${var.local_boundary_dir}:$PATH\" go test -v ${var.test_package} -count=1"]
+  inline = ["PATH=\"${var.local_boundary_dir}:$PATH\" go test -v ${var.test_package} -count=1 -json | tparse -follow -format plain"]
 }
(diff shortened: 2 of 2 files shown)


.github/workflows/enos-run.yml
```diff
@@ -40,6 +40,8 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version: ${{ inputs.go-version }}
+      - name: Install tparse
+        run: go install github.com/mfridman/tparse@v0.10.3
```
johanbrandhorst, line 44: If we use `make tools` here instead we don't have to version this tool in two places.

enos/modules/test_e2e/main.tf
```diff
@@ -122,7 +122,7 @@ resource "enos_local_exec" "run_e2e_test" {
     E2E_AWS_HOST_SET_IPS2         = local.aws_host_set_ips2
   }
 
-  inline = ["PATH=\"${var.local_boundary_dir}:$PATH\" go test -v ${var.test_package} -count=1"]
+  inline = ["PATH=\"${var.local_boundary_dir}:$PATH\" go test -v ${var.test_package} -count=1 -json | tparse -follow -format plain"]
```
johanbrandhorst, line 125: Weird that the format looks ugly in github actions when it looks fine in CircleCI, maybe worth raising an issue? Also looks like the markdown format should work particularly well for github actions: https://github.com/mfridman/tparse/issues/57?

moduli: I'll try it and see what it looks like

moduli: Screen Shot 2022-12-14 at 8 25 26 PM
https://github.com/hashicorp/boundary-enterprise/blob/main/.github/workflows/build.yml#L392-L396

moduli: Created an issue here: https://github.com/mfridman/tparse/issues/85

johanbrandhorst (approved):

PR #2721, opened 2022-12-15
Add the ability for CLI map values to contain keys only, which map to JSON nulls

jefferai:


internal/cmd/common/flags_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/hashicorp/boundary/internal/cmd/base"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/types/known/wrapperspb"
 )
 
 // TestPopulateAttrFlags tests common patterns we'll actually be using. Note
```
tmessi, line 15: can you add some new tests in this file to cover the case where just a key is set?

tmessi (changes requested):

internal/cmd/common/flags_test.go
jefferai: Yep

jefferai: Done!

internal/cmd/common/flags.go
```diff
@@ -263,76 +264,88 @@ func HandleAttributeFlags(c *base.Command, suffix, fullField string, sepFields [
...
+			}
 
 		case "bool-" + suffix:
-			switch field.Value {
-			case "true":
-				val = true
-			case "false":
-				val = false
+			switch {
+			case field.Value == nil:
```
tmessi, line 293: Thinking about this, and looking at the new tests, is this what we want? This seems to turn a `boolean` into a ternary or a "[NullBool](https://pkg.go.dev/database/sql#NullBool)". I would think for this, and probably all of the type-prefixed ones, we should either error or force the value to be the specified type.

jefferai: That's fair. I can change it back. For the same reason string-attr should not be key only and num-attr, yes?

tmessi: > For the same reason string-attr should not be key only and num-attr, yes?

Yea, I think so.

jefferai: Done!

internal/cmd/common/flags.go
```diff
@@ -264,8 +265,10 @@ func HandleAttributeFlags(c *base.Command, suffix, fullField string, sepFields [
 		// First, perform any needed parsing if we are given the type
 		switch field.Name {
 		case "num-" + suffix:
+			if field.Value == nil {
+				return fmt.Errorf("num-%s flag does not support key-only values", suffix)
```
tmessi, line 269: "key-only values" sounds odd to me, perhaps something like "num-%s flag requires a value"?

jefferai: Fixed!

tmessi (approved):

PR #2723, opened 2022-12-15
fix(kms/test): Fix flake in Test_RotateKeys

moduli:
We observed a flake in `Test_RotateKeys` due to the following reason
```
--- FAIL: Test_RotateKeys/success (0.02s)
                                             
    kms_test.go:225: 
        	Error Trace:	/home/circleci/boundary/internal/kms/kms_test.go:225
        	Error:      	Not equal: 
        	            	expected: 21
        	            	actual  : 14
        	Test:       	Test_RotateKeys/success
```

The theory is that there may be some other tests impacting the number of keys, so this PR prevents this test from running in parallel.

### DIFF
--- a/internal/kms/kms_test.go
+++ b/internal/kms/kms_test.go
@@ -202,7 +202,6 @@ func Test_RotateKeys(t *testing.T) {
 	rw := db.New(conn)
 
 	t.Run("success", func(t *testing.T) {
-		t.Parallel()
 		// arrange
 		// we're not trying to test the ListKeys function, although we need to use it to validate the rotation
 		keys, err := kmsCache.ListKeys(testCtx, "global")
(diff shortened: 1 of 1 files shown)


internal/kms/kms_test.go
```diff
@@ -192,7 +192,6 @@ func Test_ListKeys(t *testing.T) {
 }
 
 func Test_RotateKeys(t *testing.T) {
-	t.Parallel()
```
johanbrandhorst, line 195: I see what's happening here, we're using the same `kmsCache` between multiple concurrently running tests. We can leave this one in place, but we should either give each subtest its own "kmsCache", or remove t.Parallel() from inside each subtest.

johanbrandhorst (changes requested):
Thanks a lot for this, I think I can see what the issue is.

internal/kms/kms_test.go
moduli: I tried doing the `kmsCache` route, but was getting failures/doing something wrong. Pushed up a change for doing the `t.Parallel` route.

johanbrandhorst (approved):

PR #2724, opened 2022-12-16
feature (auth/ldap): add Repository.CreateAuthMethod(...) and Repository.DeleteAuthMethod(...)

jimlambrt:


internal/auth/ldap/repository_auth_method_create.go
```diff
@@ -0,0 +1,149 @@
...
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/kms"
+	"github.com/hashicorp/boundary/internal/oplog"
+)
+
+func (r *Repository) CreateAuthMethod(ctx context.Context, am *AuthMethod, _ ...Option) (*AuthMethod, error) {
```
irenarindos, line 13: Can this be documented?

internal/auth/ldap/auth_method.go
```diff
@@ -175,7 +183,7 @@ func (am *AuthMethod) convertUrls(ctx context.Context) ([]any, error) {
 		if err != nil {
 			return nil, errors.Wrap(ctx, err, op)
 		}
-		obj, err := NewUrl(ctx, am.PublicId, priority, parsed)
+		obj, err := NewUrl(ctx, am.PublicId, priority+1, parsed)
```
irenarindos, line 186: Can you document that priority is incremented by 1 in this function/ why?

internal/auth/ldap/repository_auth_method_create.go
jimlambrt: ty!

internal/auth/ldap/auth_method.go
jimlambrt: The `NewUrl(...)` documentation states that `connectionPriority cannot be less than one`.  That isn't sufficient, given that the range index will start at zero?

irenarindos: That's fair! I was just thinking it'd be a little less digging for someone in the future

internal/auth/ldap/repository_auth_method_create.go
```diff
@@ -0,0 +1,157 @@
...
+	)
+	if err != nil {
+		return nil, errors.Wrap(ctx, err, op)
+	}
+	found, err := r.getAuthMethods(ctx, am.GetPublicId(), nil)
+	switch {
+	case err != nil:
+		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("failed to lookup created auth method %q", am.GetPublicId()))
+	case len(found) != 1:
+		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("found %q auth method (%q) and expected 1", len(found), am.GetPublicId()))
```
johanbrandhorst, line 153: `%q` is here used with an integer, but is only supported with strings and byte slices according to https://pkg.go.dev/fmt#hdr-Printing. I think you want `%d`?.

johanbrandhorst (changes requested):

internal/auth/ldap/repository_auth_method_create.go
jimlambrt: ty!

irenarindos (approved):

johanbrandhorst (approved):

PR #2725, opened 2022-12-16
Use primary auth method if no auth method ID is provided

malnick:
Enable faster authentication workflows on the Boundary CLI. This PR allows short-hand usage of `boundary authenticate` with a primary auth method is set in the global scopes. This PR also modifies the password sub command to read in the login name. This is a requirement in order to get around flag rewriting between `boundary authenticate` and the `password` and `oidc` sub commands. 

Here are a few examples:
### Using the Primary Auth Method in Global Scope
```
boundary authenticate
```

### Using the Primary Auth Method in a passed in scope:
```
boundary authenticate password -scope-id o_1234567890
```

### Using flags to set the password login name and password (traditional)
```
boundary authenticate password -login-name foo -password $Up3r$3cure
```

### Using a Specific Auth Method ID (traditional)
Password:
```
boundary authenticate password -login-name user -auth-method-id ampw_1234567890
```
OIDC:
```
boundary authenticate oidc -auth-method-id amoidc_1234567890
```

A few other notes:
- If both auth method and scope ID flags are used, the auth method is preferred.
-  Flag usage is not allowed with `boundary authenticate`: if you prefer to pass your login name and/or password using flags for the `password` sub command, use `boundary authenticate password ...`
- These changes are compatible with existing CLI usage, there are no breaking changes to existing usage.

### DIFF
--- a/internal/cmd/commands/authenticate/authenticate.go
+++ b/internal/cmd/commands/authenticate/authenticate.go
@@ -9,6 +9,7 @@ import (
 	"github.com/hashicorp/boundary/internal/auth/oidc"
 	"github.com/hashicorp/boundary/internal/auth/password"
 	"github.com/hashicorp/boundary/internal/cmd/base"
+	"github.com/hashicorp/boundary/internal/cmd/common"
 	"github.com/hashicorp/boundary/internal/types/scope"
 	"github.com/mitchellh/cli"
 	"github.com/mitchellh/go-wordwrap"
@@ -47,7 +48,21 @@ func (c *Command) Help() string {
 		"      $ boundary authenticate oidc -auth-method-id amoidc_1234567890",
 		"",
 		"  Please see the auth method subcommand help for detailed usage information.",
+	}) + c.Flags().Help()
+}
+
+func (c *Command) Flags() *base.FlagSets {
+	set := c.FlagSet(base.FlagSetHTTP | base.FlagSetClient | base.FlagSetOutputFormat)
+	f := set.NewFlagSet("Command Options")
+
+	f.StringVar(&base.StringVar{
+		Name:   "scope-id",
+		EnvVar: "BOUNDARY_SCOPE_ID",
+		Target: &c.FlagScopeId,
+		Usage:  "The scope to use for the operation",
 	})
+
+	return set
 }
 
 func (c *Command) Run(args []string) int {
@@ -64,6 +79,13 @@ func (c *Command) Run(args []string) int {
 		return base.CommandCliError
 	}
 
+	f := c.Flags()
+
+	if err := f.Parse(args); err != nil {
+		c.PrintCliError(err)
+		return base.CommandUserError
+	}
+
 	// Lookup the primary auth method ID in the global scope
 	aClient := authmethods.NewClient(client)
 	pri, err := getPrimaryAuthMethodId(c.Context, aClient, scope.Global.String(), "")
@@ -76,15 +98,15 @@ func (c *Command) Run(args []string) int {
 
 	switch {
 	case strings.HasPrefix(c.FlagAuthMethodId, password.AuthMethodPrefix):
-		cmd := PasswordCommand{Command: c.Command}
+		cmd := PasswordCommand{Command: c.Command, Opts: []common.Option{common.WithSkipScopeIdFlag(true)}}
 		cmd.Run([]string{})
 
 	case strings.HasPrefix(c.FlagAuthMethodId, oidc.AuthMethodPrefix):
-		cmd := OidcCommand{Command: c.Command}
+		cmd := OidcCommand{Command: c.Command, Opts: []common.Option{common.WithSkipScopeIdFlag(true)}}
 		cmd.Run([]string{})
 
 	default:
-		c.PrintCliError(fmt.Errorf("The primary auth method was of an unsupported type. Only 'ampw' (password) and 'amoidc' (OIDC) auth method prefixes are supported.", c.FlagAuthMethodId))
+		c.PrintCliError(fmt.Errorf("The primary auth method was of an unsupported type. The given ID was %s; only 'ampw' (password) and 'amoidc' (OIDC) auth method prefixes are supported.", c.FlagAuthMethodId))
 		return cli.RunResultHelp
 	}
 
--- a/internal/cmd/commands/authenticate/oidc.go
+++ b/internal/cmd/commands/authenticate/oidc.go
@@ -11,6 +11,7 @@ import (
 	"github.com/hashicorp/boundary/api"
 	"github.com/hashicorp/boundary/api/authmethods"
 	"github.com/hashicorp/boundary/internal/cmd/base"
+	"github.com/hashicorp/boundary/internal/cmd/common"
 	"github.com/hashicorp/boundary/internal/types/scope"
 	"github.com/hashicorp/cap/util"
 	"github.com/mitchellh/cli"
@@ -25,6 +26,9 @@ var (
 
 type OidcCommand struct {
 	*base.Command
+
+	Opts       []common.Option
+	parsedOpts *common.Options
 }
 
 func (c *OidcCommand) Synopsis() string {
@@ -54,12 +58,14 @@ func (c *OidcCommand) Flags() *base.FlagSets {
 		Usage:  "The auth-method resource to use for the operation",
 	})
 
-	f.StringVar(&base.StringVar{
-		Name:   "scope-id",
-		EnvVar: "BOUNDARY_SCOPE_ID",
-		Target: &c.FlagScopeId,
-		Usage:  "The scope ID to use for the operation.",
-	})
+	if c.parsedOpts == nil || !c.parsedOpts.WithSkipScopeIdFlag {
+		f.StringVar(&base.StringVar{
+			Name:   "scope-id",
+			EnvVar: "BOUNDARY_SCOPE_ID",
+			Target: &c.FlagScopeId,
+			Usage:  "The scope ID to use for the operation.",
+		})
+	}
 
 	return set
 }
@@ -73,6 +79,13 @@ func (c *OidcCommand) AutocompleteFlags() complete.Flags {
 }
 
 func (c *OidcCommand) Run(args []string) int {
+	opts, err := common.GetOpts(c.Opts...)
+	if err != nil {
+		c.PrintCliError(err)
+		return base.CommandCliError
+	}
+	c.parsedOpts = opts
+
 	f := c.Flags()
 
 	if err := f.Parse(args); err != nil {
--- a/internal/cmd/commands/authenticate/password.go
+++ b/internal/cmd/commands/authenticate/password.go
@@ -9,6 +9,7 @@ import (
 	"github.com/hashicorp/boundary/api"
 	"github.com/hashicorp/boundary/api/authmethods"
 	"github.com/hashicorp/boundary/internal/cmd/base"
+	"github.com/hashicorp/boundary/internal/cmd/common"
 	"github.com/hashicorp/go-secure-stdlib/parseutil"
 	"github.com/hashicorp/go-secure-stdlib/password"
 	"github.com/mitchellh/cli"
@@ -31,6 +32,9 @@ type PasswordCommand struct {
 
 	flagLoginName string
 	flagPassword  string
+
+	Opts       []common.Option
+	parsedOpts *common.Options
 }
 
 func (c *PasswordCommand) Synopsis() string {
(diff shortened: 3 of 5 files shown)


internal/cmd/commands/authenticate/oidc.go
```diff
@@ -72,12 +72,6 @@ func (c *OidcCommand) Run(args []string) int {
 		return base.CommandUserError
 	}
 
-	switch {
-	case c.FlagAuthMethodId == "":
-		c.PrintCliError(errors.New("Auth method ID must be provided via -auth-method-id"))
-		return base.CommandUserError
-	}
-
```
johanbrandhorst, line 80: Do we not still need this for OIDC?

internal/cmd/commands/authenticate/password.go
```diff
@@ -74,6 +74,13 @@ func (c *PasswordCommand) Flags() *base.FlagSets {
 		Usage:  "The auth-method resource to use for the operation.",
 	})
 
+	f.StringVar(&base.StringVar{
+		Name:   "scope-id",
+		EnvVar: "BOUNDARY_SCOPE_ID",
+		Target: &c.FlagScopeId,
+		Usage:  "The scope ID to use for the operation.",
+	})
```
johanbrandhorst, line 82: I would think with the addition of this we should also add an error if the auth-method-id is not in the scope-id specified, since it may confuse users who provide conflicting information. What do you think?

malnick: That's a great point, given the order of operations:
1. If the scope id and AM id are not present, it defaults to global scope and if there is no primary in global scope, we should have a specific error for that.
2. If the scope id is pass, and no primary is found, an error specific to no primary being found in that scope id should be emitted.

internal/cmd/commands/authenticate/oidc.go
malnick: I was waiting on the first pass on password auth method before doing OIDC. If you think the pattern in password is ok, I can get that same implemented in OIDC next.

internal/cmd/commands/authenticate/password.go
johanbrandhorst: What if there is a primary auth method ID but it's not a password auth method 🤔?

malnick: I'm not sure what that would do, but the PR here wouldn't change the behavior since that could occur today.

malnick: I.e., you could do a `boundary authenticate password` and pass a OIDC AM ID to the -auth-method-id flag.

johanbrandhorst: True, I guess I only think we need a new error for when the scope-id and auth-method-id conflict. I'm not sure if you agreed in your first reply, but I think it's important to avoid ambiguity for the user.

internal/cmd/commands/authenticate/password.go
```diff
@@ -67,12 +67,32 @@ func (c *PasswordCommand) Flags() *base.FlagSets {
 		Usage:  "The password associated with the login name. If blank, the command will prompt for the password to be entered interactively in a non-echoing way. Otherwise, this can refer to a file on disk (file://) from which a password will be read or an env var (env://) from which the password will be read.",
 	})
 
-	f.StringVar(&base.StringVar{
-		Name:   "auth-method-id",
-		EnvVar: "BOUNDARY_AUTH_METHOD_ID",
-		Target: &c.FlagAuthMethodId,
-		Usage:  "The auth-method resource to use for the operation.",
-	})
```
malnick, line 75: @jimlambrt @johanbrandhorst (FYI @jefferai ) Ideas on this issue I ran into starting on line 76?

internal/cmd/commands/authenticate/authenticate.go
```diff
@@ -35,5 +48,42 @@ func (c *Command) Help() string {
...
+	if err != nil {
+		c.PrintCliError(fmt.Errorf("Error creating API client: %w", err))
+		return base.CommandCliError
+	}
+
+	// Lookup the primary auth method ID in the global scope
+	aClient := authmethods.NewClient(client)
+	pri, err := getPrimaryAuthMethodId(c.Context, aClient, "global", "")
+	if err != nil {
+		c.PrintCliError(errors.New("Error looking up primary auth method ID for global scope. Try setting primary auth method for global scope, or use a auth method subcommand (see 'boundary authenticate -h' for available sub command usage)."))
```
johanbrandhorst, line 68: Should we not be using the definite article for the global scope?
```suggestion
		c.PrintCliError(errors.New("Error looking up primary auth method ID for the global scope. Try setting a primary auth method for the global scope, or use an auth method subcommand (see 'boundary authenticate -h' for available sub command usage)."))
```

internal/cmd/commands/authenticate/authenticate.go
```diff
@@ -35,5 +48,42 @@ func (c *Command) Help() string {
...
+	}
+
+	c.FlagAuthMethodId = pri
+
+	switch {
+	case strings.HasPrefix(c.FlagAuthMethodId, "ampw"):
+		cmd := PasswordCommand{Command: c.Command}
+		cmd.Run([]string{})
+
+	case strings.HasPrefix(c.FlagAuthMethodId, "amoidc"):
```
johanbrandhorst, line 79: I was under the impression that we wanted to remove these prefixes from IDs at some point, would it not be better to check the type of the primary auth method directly?

internal/cmd/commands/authenticate/authenticate.go
```diff
@@ -35,5 +48,42 @@ func (c *Command) Help() string {
...
+		cmd := OidcCommand{Command: c.Command}
+		cmd.Run([]string{})
+
+	default:
+		c.PrintCliError(fmt.Errorf("Invalid auth method ID prefix '%s'. Only 'ampw' (password) and 'amoidc' (OIDC) auth method prefixes are supported.", c.FlagAuthMethodId))
```
johanbrandhorst, line 84: This error is probably confusing to the user, as they didn't specify the auth method ID (we inferred it from the primary auth method, right?). When we add LDAP this could be happening to users on old CLIs, so maybe we should make this say something like "The primary auth method was of an unsupported type, please upgrade your Boundary version" or something. What do you think?

internal/cmd/commands/authenticate/funcs.go
```diff
@@ -100,3 +102,26 @@ func saveAndOrPrintToken(c *base.Command, result *authmethods.AuthenticateResult
...
+	}
+	authMethodListResult, err := client.List(ctx, scopeId)
+	if err != nil {
+		return "", err
+	}
+
+	for _, m := range authMethodListResult.GetItems() {
+		if m.IsPrimary {
+			if !strings.HasPrefix(m.Id, amType) {
+				return "", fmt.Errorf("Error looking up primary auth method in scope '%s': got '%s' but the command requires a auth method prefix of '%s'. Make sure the sub command you're using matches the primary auth method type in the scope being used. For example, if using the password sub command the primary auth method must have prefix 'ampw'.\n\nSee 'boundary authenticate -h' for available sub command usage.", scopeId, m.Id, amType)
```
johanbrandhorst, line 120:
```suggestion
				return "", fmt.Errorf("Error looking up primary auth method in scope '%s': got '%s' but the command requires an auth method prefix of '%s'. Make sure the sub command you're using matches the primary auth method type in the scope being used. For example, if using the password sub command the primary auth method must have a prefix of 'ampw'.\n\nSee 'boundary authenticate -h' for available sub command usage.", scopeId, m.Id, amType)
```

internal/cmd/commands/authenticate/funcs.go
```diff
@@ -100,3 +102,26 @@ func saveAndOrPrintToken(c *base.Command, result *authmethods.AuthenticateResult
...
+			}
+
+			return m.Id, nil
+		}
+	}
+	return "", fmt.Errorf("Primary auth method not found for scope ID: %s. Please set a primary auth method on this scope or pass one explicitly using an authenticate sub command (see 'boundary authenticate -h') along with the -auth-method-id flag.", scopeId)
```
johanbrandhorst, line 126: Should we quote the scope ID here like we do in other errors?
```suggestion
	return "", fmt.Errorf("Primary auth method not found for scope ID: '%s'. Please set a primary auth method on this scope or pass one explicitly using an authenticate sub command (see 'boundary authenticate -h') along with the -auth-method-id flag.", scopeId)
```

internal/cmd/commands/authenticate/authenticate.go
```diff
@@ -35,5 +48,42 @@ func (c *Command) Help() string {
...
+		}()
+	}
+	if err != nil {
+		c.PrintCliError(fmt.Errorf("Error creating API client: %w", err))
+		return base.CommandCliError
+	}
+
+	// Lookup the primary auth method ID in the global scope
+	aClient := authmethods.NewClient(client)
+	pri, err := getPrimaryAuthMethodId(c.Context, aClient, "global", "")
```
johanbrandhorst, line 66: Nit: the scope package has a string for this
```suggestion
	pri, err := getPrimaryAuthMethodId(c.Context, aClient, scope.Global.String(), "")
```

internal/cmd/commands/authenticate/oidc.go
```diff
@@ -92,6 +93,23 @@ func (c *OidcCommand) Run(args []string) int {
 	}
 
 	aClient := authmethods.NewClient(client)
+
+	// if auth method ID isn't passed on the CLI, try looking up the primary auth method ID
+	if c.FlagAuthMethodId == "" {
+		// if flag for scope is empty try looking up global
+		if c.FlagScopeId == "" {
+			c.FlagScopeId = "global"
```
johanbrandhorst, line 101: Nit
```suggestion
			c.FlagScopeId = scope.Global.String()
```

johanbrandhorst (commented):
Should we add some e2e tests for this? CC @moduli

moduli:
> Should we add some e2e tests for this? CC @moduli

Yea. That seems like a good idea.

internal/cmd/commands/authenticate/authenticate.go
malnick: Ah, good catch, we can definitely change to verifying by type. I'll get that taken care of.

internal/cmd/commands/authenticate/authenticate.go
malnick: 👍

internal/cmd/commands/authenticate/authenticate.go
malnick: I didn't realize I had this option, thanks for cluing me in!

internal/cmd/commands/authenticate/authenticate.go
jefferai: That said it'd be better to use the various consts -- and in some of those cases perhaps moving those consts to the globals package.

internal/cmd/commands/authenticate/authenticate.go
jefferai: +1. Also maybe say "...are `currently` supported"

internal/cmd/commands/authenticate/authenticate.go
```diff
@@ -35,5 +48,42 @@ func (c *Command) Help() string {
...
+		}()
+	}
+	if err != nil {
+		c.PrintCliError(fmt.Errorf("Error creating API client: %w", err))
+		return base.CommandCliError
+	}
+
+	// Lookup the primary auth method ID in the global scope
+	aClient := authmethods.NewClient(client)
+	pri, err := getPrimaryAuthMethodId(c.Context, aClient, scope.Global.String(), "")
```
jefferai, line 66: Can we allow a `-scope-id` flag here to override this if the auth method ID isn't known but scope ID is?

internal/cmd/commands/authenticate/authenticate.go
malnick: 👍 Fixed

internal/cmd/commands/authenticate/authenticate.go
malnick: We decided to restrict the flag scope to a sub command due to limitations in the flag library: the -scope-id flag would move up in precedence and that would modify the CLI usage, and CLI usage changes of existing commands was out of scope for this PR. To maintain the existing CLI usage and have the flag remain the precedence in the command it is for password and OIDC, we could not use it at both the `boundary authenticate` level and the `boundary authenticate <auth method>` level. Fun times.

internal/cmd/commands/authenticate/authenticate.go
malnick: 👍 Fixed

internal/cmd/commands/authenticate/authenticate.go
```diff
@@ -47,7 +48,21 @@ func (c *Command) Help() string {
...
 		"",
 		"  Please see the auth method subcommand help for detailed usage information.",
+	}) + c.Flags().Help()
+}
+
+func (c *Command) Flags() *base.FlagSets {
+	set := c.FlagSet(base.FlagSetHTTP | base.FlagSetClient | base.FlagSetOutputFormat)
+	f := set.NewFlagSet("Command Options")
+
+	f.StringVar(&base.StringVar{
```
malnick, line 58: Should we update the help text with an example of this flag usage?

johanbrandhorst (approved):
Excited for this! Do we have a plan for updating tutorials and docs with the new inferred behavior?

malnick:
We are, but since the old functionality hasn't changed, it'll be a best effort to get those changes in before our next release, but we'll work towards updating all inferences over time. Thank you for raising this!

PR #2726, opened 2022-12-16
chore(e2e): Update to use Vault 1.12.2

moduli:
This PR updates the e2e test suite to use the latest version of Vault (version 1.12.2).

### DIFF
--- a/.github/workflows/enos-run.yml
+++ b/.github/workflows/enos-run.yml
@@ -124,7 +124,7 @@ jobs:
       - name: Download Vault AMD64 binary for integration testing
         if: steps.dep-cache.outputs.cache-hit != 'true'
         run: |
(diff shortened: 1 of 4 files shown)


johanbrandhorst (commented):
Is there an easier way for us to manage this than you (or something else) manually upgrading 3 different places?

moduli:
> Is there an easier way for us to manage this than you (or something else) manually upgrading 3 different places?

Maybe worth a chat to brainstorm some ideas. We could consolidate the changes on the enos side by using an enos variable. I'm not sure how we also tie in the github workflow side.

johanbrandhorst:
2 sources of truth is still better than 3 I think, lets do it :).

johanbrandhorst (approved):

PR #2731, opened 2022-12-19
test(e2e): Add test for when an incorrect tag is used in a dynamic host catalog

moduli:
This PR adds a test to the Dynamic Host Catalog scenario that verifies the following
- Create an AWS Dynamic Host Catalog
- Create a host set with an invalid tag
- Verify that no hosts appear in the host set
- Attach the empty host set to a target
- Confirm that connecting to the target fails

This PR also contains some fixes where type assertions should have been used.

### DIFF
--- a/testing/internal/e2e/tests/aws/env_test.go
+++ b/testing/internal/e2e/tests/aws/env_test.go
@@ -1,0 +1,25 @@
+package aws_test
+
+import "github.com/kelseyhightower/envconfig"
+
+type config struct {
+	AwsAccessKeyId     string `envconfig:"E2E_AWS_ACCESS_KEY_ID" required:"true"`
+	AwsSecretAccessKey string `envconfig:"E2E_AWS_SECRET_ACCESS_KEY" required:"true"`
+	AwsHostSetFilter1  string `envconfig:"E2E_AWS_HOST_SET_FILTER" required:"true"`  // e.g. "tag:testtag=true"
+	AwsHostSetIps1     string `envconfig:"E2E_AWS_HOST_SET_IPS" required:"true"`     // e.g. "[\"1.2.3.4\", \"2.3.4.5\"]"
+	AwsHostSetFilter2  string `envconfig:"E2E_AWS_HOST_SET_FILTER2" required:"true"` // e.g. "tag:testtagtwo=test"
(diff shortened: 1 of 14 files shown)


testing/internal/e2e/tests/aws/env_test.go
```diff
@@ -0,0 +1,25 @@
...
+}
+
+func loadConfig() (*config, error) {
+	var c config
+	err := envconfig.Process("", &c)
+	if err != nil {
+		return nil, err
+	}
+
+	return &c, err
```
johanbrandhorst, line 24: It's convention to return `nil` when we know `err == nil`
```suggestion
	return &c, nil
```

testing/internal/e2e/tests/static/connect_authz_token_test.go
```diff
@@ -52,8 +52,8 @@ func TestCliConnectTargetWithAuthzToken(t *testing.T) {
 	require.NoError(t, err)
 
 	newSessionAuthorization := newSessionAuthorizationResult.Item
-	retrievedUser := fmt.Sprintf("%s", newSessionAuthorization.Credentials[0].Credential["username"])
-	retrievedKey := fmt.Sprintf("%s\n", newSessionAuthorization.Credentials[0].Credential["private_key"])
+	retrievedUser := newSessionAuthorization.Credentials[0].Credential["username"].(string)
```
johanbrandhorst, line 55: FYI these are called `type assertions` and can fail (with a panic) at runtime. I think we're confident of these types in this case, but if you're doing this and you're not _sure_ it's the right type, you can use an `_, ok` pattern to handle the error without panicing.
```go
	retrievedUser, ok := newSessionAuthorization.Credentials[0].Credential["username"].(string)
```

johanbrandhorst (approved):

testing/internal/e2e/tests/static/connect_authz_token_test.go
moduli: To confirm, it is preferred to use a `type assertion` here, but we should do error checking here to prevent panics?

moduli: Chatted with Johan. We're good to go.

PR #2732, opened 2022-12-20
add end2end enos test for target address

ddebko:
### Summary:

The following changes adds a new end2end test in enos that validates the latest feature for assigning a network address directly to a Target resource. The test validates the following logic:

- connecting to a target with a host source
- expected error when trying to assign a network address to a Target that has a host source
- assigning a network address to a Target after clearing all host source associations
- connecting to a target with a direct network address association
- expected error when trying to set & add a host source to a Target with a network address assigned to it
- remove the network address from a Target and reassigned a host source to it

### DIFF
--- a/testing/internal/e2e/boundary/target.go
+++ b/testing/internal/e2e/boundary/target.go
@@ -61,10 +61,60 @@ func CreateNewTargetCli(t testing.TB, ctx context.Context, projectId string, def
 	return newTargetId
 }
 
-// AddHostSourceToTargetCli uses the cli to add a host source (host set or host) to a target
-func AddHostSourceToTargetCli(t testing.TB, ctx context.Context, targetId string, hostSourceId string) {
+// CreateNewAddressTargetCli uses the cli to create a new target using an
+// address in boundary.
+// Returns the id of the new target.
+func CreateNewAddressTargetCli(t testing.TB, ctx context.Context, projectId string, defaultPort string, address string) string {
+	output := e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"targets", "create", "tcp",
+			"-scope-id", projectId,
+			"-default-port", defaultPort,
+			"-name", "e2e Target",
+			"-address", address,
+			"-format", "json",
+		),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+	var newTargetResult targets.TargetCreateResult
+	err := json.Unmarshal(output.Stdout, &newTargetResult)
+	require.NoError(t, err)
+	newTargetId := newTargetResult.Item.Id
+	t.Logf("Created Target: %s", newTargetId)
+
+	return newTargetId
+}
+
+// AddHostSourceToTargetCli uses the cli to add a host source (host set or host)
+// to a target.
+// Boundary's `add-host-sources` functionality appends a new host source to the
+// existing set of host sources in the target.
+func AddHostSourceToTargetCli(t testing.TB, ctx context.Context, targetId, hostSourceId string) {
 	output := e2e.RunCommand(ctx, "boundary",
 		e2e.WithArgs("targets", "add-host-sources", "-id", targetId, "-host-source", hostSourceId),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+}
+
+// SetHostSourceToTargetCli uses the cli to set a host source (host set or host)
+// to a target.
+// Boundary's `set-host-sources` functionality replaces all existing host sets
+// on a target with the provided one.
+func SetHostSourceToTargetCli(t testing.TB, ctx context.Context, targetId, hostSourceId string) {
+	output := e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs("targets", "set-host-sources", "-id", targetId, "-host-source", hostSourceId),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+}
+
+// RemoveHostSourceFromTargetCli uses the cli to remove a host source (host set or host) to a target
+func RemoveHostSourceFromTargetCli(t testing.TB, ctx context.Context, targetId, hostSourceId string) {
+	output := e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"targets", "remove-host-sources",
+			"-id", targetId,
+			"-host-source", hostSourceId,
+		),
 	)
 	require.NoError(t, output.Err, string(output.Stderr))
 }
--- a/testing/internal/e2e/tests/static/target_address_test.go
+++ b/testing/internal/e2e/tests/static/target_address_test.go
@@ -1,0 +1,340 @@
+package static_test
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/hashicorp/boundary/testing/internal/e2e"
+	"github.com/hashicorp/boundary/testing/internal/e2e/boundary"
+	"github.com/stretchr/testify/require"
+)
+
+// TestCliCreateUpdateTargetAddress asserts that creating and updating a target
+// using a valid address is successful.
+func TestCliCreateUpdateTargetAddress(t *testing.T) {
+	e2e.MaybeSkipTest(t)
+	c, err := loadConfig()
+	require.NoError(t, err)
+
+	ctx := context.Background()
+	boundary.AuthenticateAdminCli(t, ctx)
+	newOrgId := boundary.CreateNewOrgCli(t, ctx)
+	t.Cleanup(func() {
+		ctx := context.Background()
+		boundary.AuthenticateAdminCli(t, ctx)
+		output := e2e.RunCommand(ctx, "boundary", e2e.WithArgs("scopes", "delete", "-id", newOrgId))
+		require.NoError(t, output.Err, string(output.Stderr))
+	})
+	newProjectId := boundary.CreateNewProjectCli(t, ctx, newOrgId)
+	newTargetId := boundary.CreateNewAddressTargetCli(t, ctx, newProjectId, c.TargetPort, c.TargetIp)
+
+	// Connect to target and print host's IP address
+	output := e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"connect",
+			"-target-id", newTargetId,
+			"-exec", "/usr/bin/ssh", "--",
+			"-l", c.TargetSshUser,
+			"-i", c.TargetSshKeyPath,
+			"-o", "UserKnownHostsFile=/dev/null",
+			"-o", "StrictHostKeyChecking=no",
+			"-o", "IdentitiesOnly=yes", // forces the use of the provided key
+			"-p", "{{boundary.port}}", // this is provided by boundary
+			"{{boundary.ip}}",
+			"hostname", "-i",
+		),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+
+	parts := strings.Fields(string(output.Stdout))
+	hostIp := parts[len(parts)-1]
+	require.Equal(t, c.TargetIp, hostIp, "SSH session did not return expected output")
+	t.Log("Successfully connected to target")
+
+	// Remove target address. We have now no address or host sources.
+	output = e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"targets", "update", "tcp",
+			"-id", newTargetId,
+			"-address", "null",
+			"-format", "json",
+		),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+
+	// Connect to target and print host's IP address - should error because
+	// there's no place to connect to.
+	output = e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"connect",
+			"-target-id", newTargetId,
+			"-format", "json",
+			"-exec", "/usr/bin/ssh", "--",
+			"-l", c.TargetSshUser,
+			"-i", c.TargetSshKeyPath,
+			"-o", "UserKnownHostsFile=/dev/null",
+			"-o", "StrictHostKeyChecking=no",
+			"-o", "IdentitiesOnly=yes", // forces the use of the provided key
+			"-p", "{{boundary.port}}", // this is provided by boundary
+			"{{boundary.ip}}",
+			"hostname", "-i",
+		),
+	)
+	require.Error(t, output.Err, string(output.Stderr))
+
+	var response boundary.CliError
+	require.NoError(t, json.Unmarshal(output.Stderr, &response))
+	require.Equal(t, http.StatusNotFound, response.Status, "Expected error when connecting to target with no host sources or address")
+
+	// Add a valid address back in.
+	output = e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"targets", "update", "tcp",
+			"-id", newTargetId,
+			"-address", c.TargetIp,
+			"-format", "json",
+		),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+
+	// Connect to target and print host's IP address.
+	output = e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"connect",
+			"-target-id", newTargetId,
+			"-exec", "/usr/bin/ssh", "--",
+			"-l", c.TargetSshUser,
+			"-i", c.TargetSshKeyPath,
+			"-o", "UserKnownHostsFile=/dev/null",
+			"-o", "StrictHostKeyChecking=no",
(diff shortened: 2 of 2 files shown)


testing/internal/e2e/tests/static/target_address_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"testing"
+
+	"github.com/hashicorp/boundary/testing/internal/e2e"
+	"github.com/hashicorp/boundary/testing/internal/e2e/boundary"
+	"github.com/stretchr/testify/require"
+)
+
+// TestTargetAddress uses the boundary cli to create a number of supporting objects
+// to connect to a target. It then attempts to connect to that target using a host source
+// and verifies that the connection was successful. Afterwards, we validate that a address
```
hugoamvieira, line 15:
```suggestion
// and verifies that the connection was successful. Afterwards, we validate that an address
```

testing/internal/e2e/tests/static/target_address_test.go
```diff
@@ -0,0 +1,91 @@
...
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+
+	parts := strings.Fields(string(output.Stdout))
+	hostIp := parts[len(parts)-1]
+	require.Equal(t, c.TargetIp, hostIp, "SSH session did not return expected output")
+	t.Log("Successfully connected to target")
+
+	_, err = boundary.SetTargetAddressCli(t, ctx, newTargetId, c.TargetIp)
+	require.Error(t, err, "Target update did not return expected error")
```
hugoamvieira, line 66: We should probably assert that this error is what we expect (we could string match part of it or something like that)

testing/internal/e2e/tests/static/target_address_test.go
```diff
@@ -0,0 +1,91 @@
...
+			"-o", "UserKnownHostsFile=/dev/null",
+			"-o", "StrictHostKeyChecking=no",
+			"-o", "IdentitiesOnly=yes", // forces the use of the provided key
+			"-p", "{{boundary.port}}", // this is provided by boundary
+			"{{boundary.ip}}",
+			"hostname", "-i",
+		),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+}
```
hugoamvieira, line 91: At the end here, now that the target uses an address, we should also validate that we can't add or set host sources on it.

hugoamvieira (commented):
Just some comments on the test logic ⬇️

testing/internal/e2e/boundary/target.go
```diff
@@ -62,13 +62,25 @@ func CreateNewTargetCli(t testing.TB, ctx context.Context, projectId string, def
...
+func RemoveHostSourceFromTargetCli(t testing.TB, ctx context.Context, targetId, hostSourceId string) (string, error) {
+	output := e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"targets", "remove-host-sources",
+			"-id", targetId,
+			"-host-source", hostSourceId,
+		),
+	)
+	return string(output.Stdout), output.Err
+}
```
johanbrandhorst, line 82: Should this take a `t testing.TB` if it's returning the error? I'd say it should either take a `t testing.TB` and then check for the error with `require`, or return the error and no accept a `t`. What do you think? It looks like we could just check the error in this helper function instead?

testing/internal/e2e/boundary/target.go
```diff
@@ -81,3 +93,15 @@ func AddCredentialSourceToTargetCli(t testing.TB, ctx context.Context, targetId
...
+// SetTargetAddressCli uses the cli to assign a network address value to a target
+func SetTargetAddressCli(ctx context.Context, targetId, address string) (string, error) {
+	output := e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"targets", "update", "tcp",
+			"-id", targetId,
+			"-address", address,
+		),
+	)
+	return string(output.Stdout), output.Err
```
johanbrandhorst, line 106: I see this function doesn't take a `t` at all, but is there a reason not to just check the error in here? I see that we sometimes want to check the exact error, but in those cases, do we even need this helper? We can just use `e2e.RunCommand` directly.

testing/internal/e2e/tests/static/target_address_test.go
johanbrandhorst: If we want to do any sort of asserting about the content of this error, I think it should be checking what the status code was. The string text could change but the status code should be stable, and should be what customers use. @moduli can correct me but I think we have some utility for parsing out the status code from the error already.

moduli: I agree. We should validate the status code here. You can parse errors using `boundary.CliError`.

Example:
```
var response boundary.CliError
err = json.Unmarshal(output.Stderr, &response)
require.NoError(t, err)
require.Equal(t, response.Status, 404, "Expected to error when connecting to a target with zero hosts")
```

testing/internal/e2e/boundary/target.go
moduli: Agreed. The pattern we've used so far is to do the error checking in the helper function. You can see `AddHostSourceToTargetCli` as an example.

testing/internal/e2e/tests/static/target_address_test.go
```diff
@@ -0,0 +1,154 @@
...
+// to connect to a target. It then attempts to connect to that target using a host source
+// and verifies that the connection was successful. Afterwards, we validate that an address
+// cannot be assigned directly to the target that has an existing host source association.
+// The host sources are removed from the same target and a network address is assigned
+// directly to the target. Then it attempts to connect to that target using
+// the direct network address association. Afterwards the test validates that a host source
+// cannot be added or set to a target with an existing network address assigned to it.
+// Finally, the target network address is removed and the host source is added back to the
+// target to attempt a final connection attempt.
+func TestTargetAddress(t *testing.T) {
```
moduli, line 23: We may want to update the name of this test to be more specific, given that we know we have at least 3 tests already defined.

testing/internal/e2e/tests/static/target_address_test.go
ddebko: That sounds like a good idea, I will update the test to make checks based on the status code instead of error message.

testing/internal/e2e/boundary/target.go
ddebko: I was debating this in my head too, so I will go ahead and remove the helper function and use e2e.RunCommand directly.

testing/internal/e2e/tests/static/target_address_test.go
ddebko: Yea, I was having a hard time naming this one, maybe one the following names could be a good replacement:
- TestTargetAddressHostSourceRelationship()
- TestTargetAddressToHostSourceMutuallyExclusiveRelationship()
- TestTargetMutuallyExclusiveRelationship()

I wasn't sure what to pick because I didn't want the function name to feel like I was reading out a sentence, which is why I added a detailed comment to the test.

testing/internal/e2e/boundary/target.go
ddebko: I agree, we should keep the pattern in the helper functions consistent.

testing/internal/e2e/tests/static/target_address_test.go
moduli: There were three tasks identified on the spreadsheet. Is that still the plan? Does this test cover the first and third items or am I mistaken? Does it make sense to keep these tests separate to help with readability?

Something like...
```
TestCliTargetAddressCreateAndUpdate??
TestCliTargetAddressToHostSource
TestCliTargetHostSourceToAddress
```

testing/internal/e2e/tests/static/target_address_test.go
```diff
@@ -0,0 +1,181 @@
...
+	require.NoError(t, output.Err, string(output.Stderr))
+
+	output = e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"targets", "set-host-sources",
+			"-id", newTargetId,
+			"-host-source", newHostSetId,
+		),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
```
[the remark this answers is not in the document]
ddebko: Both set and add need to validate that the target doesn't have an existing network address associated to it. The handlers go down different code paths, so we need to validate that both paths will return the expected error.

testing/internal/e2e/tests/static/target_address_test.go
```diff
@@ -0,0 +1,184 @@
...
+			"-address", c.TargetIp,
+			"-format", "json",
+		),
+	)
+	require.Error(t, output.Err, "Target update did not return expected error")
+
+	var response boundary.CliError
+	err = json.Unmarshal(output.Stderr, &response)
+	require.NoError(t, err)
+	require.Equal(t, http.StatusForbidden, response.Status, "Expected error when setting address to target with an host source")
```
hugoamvieira, line 83:
```suggestion
	require.Equal(t, http.StatusForbidden, response.Status, "Expected error when setting address to target with a host source")
```

testing/internal/e2e/tests/static/target_address_test.go
```diff
@@ -0,0 +1,184 @@
...
+			"targets", "set-host-sources",
+			"-id", newTargetId,
+			"-host-source", newHostSetId,
+			"-format", "json",
+		),
+	)
+	require.Error(t, output.Err, "Target set host sources did not return expected error")
+	err = json.Unmarshal(output.Stderr, &response)
+	require.NoError(t, err)
+	require.Equal(t, http.StatusForbidden, response.Status, "Expected error when setting host source to target with an network address")
```
hugoamvieira, line 138:
```suggestion
	require.Equal(t, http.StatusForbidden, response.Status, "Expected error when setting host source to target with an address")
```

hugoamvieira:
@ddebko wondering if we should put Boundary's returned error message on our `require.Errorf` calls to allow for easier debugging. If you notice for the current [github action](https://github.com/hashicorp/boundary/actions/runs/3747590566/jobs/6364148541), it failed with a 500 but other than that, we don't really know what happened

testing/internal/e2e/tests/static/target_address_test.go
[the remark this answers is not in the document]
hugoamvieira: `add` _appends_ a new host source to a target
`set` _replaces_ all existing host source(s) with a new one

testing/internal/e2e/tests/static/target_address_test.go
hugoamvieira: I just did this - let me know what you think 😃

testing/internal/e2e/boundary/target.go
```diff
@@ -61,14 +61,58 @@ func CreateNewTargetCli(t testing.TB, ctx context.Context, projectId string, def
...
+	require.NoError(t, err)
+	newTargetId := newTargetResult.Item.Id
+	t.Logf("Created Target: %s", newTargetId)
+
+	return newTargetId
+}
+
 // AddHostSourceToTargetCli uses the cli to add a host source (host set or host) to a target
-func AddHostSourceToTargetCli(t testing.TB, ctx context.Context, targetId string, hostSourceId string) {
+func AddHostSourceToTargetCli(t testing.TB, ctx context.Context, targetId, hostSourceId string) {
```
moduli, line 89: Is this a general pattern that is used? Does this mean that `targetId` and `hostSourceId` are strings?

testing/internal/e2e/tests/static/target_address_test.go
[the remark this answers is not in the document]
moduli: > `add` _appends_ a new host source to a target `set` _replaces_ all existing host source(s) with a new one

Could we add this to the comments for the respective methods?

testing/internal/e2e/tests/static/target_address_test.go
```diff
@@ -0,0 +1,282 @@
...
+	"testing"
+
+	"github.com/hashicorp/boundary/testing/internal/e2e"
+	"github.com/hashicorp/boundary/testing/internal/e2e/boundary"
+	"github.com/stretchr/testify/require"
+)
+
+// TestCreateUpdateTargetAddress asserts that creating and updating a target
+// using a valid address is successful.
+func TestCreateUpdateTargetAddress(t *testing.T) {
```
moduli, line 17:
```suggestion
func TestCliCreateUpdateTargetAddress(t *testing.T) {
```

Can we add `Cli` to the names here to denote that these are tests for the CLI? Additionally, you could look into a similar test using the Go Api (without needing to test the `connect` pieces).

testing/internal/e2e/tests/static/target_address_test.go
```diff
@@ -0,0 +1,282 @@
...
+	parts := strings.Fields(string(output.Stdout))
+	hostIp := parts[len(parts)-1]
+	require.Equal(t, c.TargetIp, hostIp, "SSH session did not return expected output")
+	t.Log("Successfully connected to target")
+
+	output = e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"targets", "update", "tcp",
+			"-id", newTargetId,
+			"-address", "null", // TODO: update to different host and connect (requires having new host to connect to first)
```
moduli, line 61: @hugoamvieira We had talked about this over Slack, but I had another thought. Instead of spinning up a second host, when if we keep this update to `null` here, confirm that connecting will fail, and then update it back to the original address?

testing/internal/e2e/boundary/target.go
ddebko: Yea, If you have a function signature with multiple arguments of the same type, you don't have to repeat yourself and write the type of every argument.

testing/internal/e2e/tests/static/target_address_test.go
```diff
@@ -0,0 +1,342 @@
...
+		),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+
+	parts := strings.Fields(string(output.Stdout))
+	hostIp := parts[len(parts)-1]
+	require.Equal(t, c.TargetIp, hostIp, "SSH session did not return expected output")
+	t.Log("Successfully connected to target")
+
+	// Attempt add-host-sources, which appends a new host source to an existing host sources.
```
moduli, line 176: Sorry, could we move the comments I suggested to the methods in `target.go`? (`AddHostSourceToTargetCli`, `SetHostSourceToTargetCli`)

testing/internal/e2e/tests/static/target_address_test.go
```diff
@@ -0,0 +1,342 @@
...
+			"targets", "update", "tcp",
+			"-id", newTargetId,
+			"-address", "null",
+			"-format", "json",
+		),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+
+	boundary.AddHostSourceToTargetCli(t, ctx, newTargetId, newHostSetId)
+	boundary.SetHostSourceToTargetCli(t, ctx, newTargetId, newHostSetId)
```
moduli, line 219: Just curious, why do we do both `add` and `set` here? You just want to confirm that they perform successfully?

moduli (commented):
Couple of minor comments, but overall it's looking good!

testing/internal/e2e/tests/static/target_address_test.go
hugoamvieira: @moduli Correct. `add` and `set` are two entirely different Boundary operations that have their own codepaths, and we added stuff on both, so we have to test both.

moduli: Do we need to confirm that it was `added` correctly? (outside of just verifying that the command didn't error, that the `add` operation correctly added the host source.

hugoamvieira: That's we try a `connect` afterwards. If it wasn't added or set correctly, the connect won't find any host source to connect to 🙂

hugoamvieira: Ah! I see what you mean now. Yeah, the set will replace what the add did, so here we're relying on an error to be given back from the CLI call. We can always run 2 boundary connects, one after adding, then again after setting 😄

testing/internal/e2e/tests/static/target_address_test.go
hugoamvieira: Done with [13c23e7](https://github.com/hashicorp/boundary/pull/2732/commits/13c23e7de21767dbd32f0f8a1889d585e9eb18d7)

moduli:
FYI: Looks like there are some failing unit tests.

hugoamvieira:
@moduli Yep, been working on that, just pushed up a fix. These commits really shouldn't be on this branch, I'm going to do some git-fu to clean this branch up a little now

moduli (approved):

PR #2733, opened 2022-12-20
Expose State() *structpb.Struct interface through the metric connecti… …on trackers

irenarindos:
Ensuring change as part of this merge is in OSS: https://github.com/hashicorp/boundary-enterprise/commit/444e861cc3a250feeb1d25b96962f26b83a0215f

### DIFF
--- a/internal/daemon/metric/instrument_handlers.go
+++ b/internal/daemon/metric/instrument_handlers.go
@@ -15,6 +15,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/stats"
 	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/structpb"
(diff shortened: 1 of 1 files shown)


internal/daemon/metric/instrument_handlers.go
```diff
@@ -121,6 +128,25 @@ func NewConnectionTrackingListener(l net.Listener, ac prometheus.Counter, cc pro
 	return &connectionTrackingListener{Listener: l, acceptedConns: ac, closedConns: cc}
 }
 
+type stateProvidingConn interface {
+	net.Conn
+	State() *structpb.Struct
```
johanbrandhorst, line 133: Is the structure of this always a `WorkerConnectionInfo`? If so, do we need it to be this generic?

PR #2736, opened 2022-12-20
SSH Proxy support

irenarindos:
Partner PR for https://github.com/hashicorp/boundary-enterprise/pull/194


internal/daemon/worker/proxy/proxydialer.go
```diff
@@ -78,9 +78,9 @@ func (d *ProxyDialer) LastConnectionAddr() *proxyAddr {
 // Dial uses the provided dial function to get a net.Conn and record its
 // net.Addr information.  The returned net.Addr should contain the information
 // for the endpoint that is being proxied to.
-func (d *ProxyDialer) Dial(ctx context.Context) (net.Conn, error) {
+func (d *ProxyDialer) Dial(ctx context.Context, opt ...Option) (net.Conn, error) {
```
jefferai, line 81: Generally we call out in comments the available options...I realize here they're only passed through but could do e.g. `WithXYZ (passed through to the dial function)` or so

PR #2738, opened 2022-12-20
feat(cmd): Create target using address on boundary dev / database init

hugoamvieira:
Creates a target using an address for boundary dev and when first init-ing boundary (boundary database init).

It also makes this newly created target the default one (ttcp_1234567890). The target using host sources is now a secondary target (ttcp_0987654321).

### DIFF
--- a/internal/cmd/commands/dev/dev.go
+++ b/internal/cmd/commands/dev/dev.go
@@ -66,6 +66,7 @@ type Command struct {
 	flagUnprivilegedLoginName        string
 	flagUnprivilegedPassword         string
 	flagIdSuffix                     string
+	flagSecondaryIdSuffix            string
 	flagHostAddress                  string
 	flagTargetDefaultPort            int
 	flagTargetSessionMaxSeconds      int
@@ -141,7 +142,15 @@ func (c *Command) Flags() *base.FlagSets {
 		Target:  &c.flagIdSuffix,
 		Default: "1234567890",
 		EnvVar:  "BOUNDARY_DEV_ID_SUFFIX",
-		Usage:   `If set, auto-created resources will use this value for their identifier (along with their resource-specific prefix). Must be 10 alphanumeric characters. As an example, if this is set to "1234567890", the generated password auth method ID will be "ampw_1234567890", the generated TCP target ID will be "ttcp_1234567890", and so on.`,
+		Usage:   `If set, auto-created resources will use this value for their identifier (along with their resource-specific prefix). Must be 10 alphanumeric characters. As an example, if this is set to "1234567890", the generated password auth method ID will be "ampw_1234567890", the generated TCP target ID will be "ttcp_1234567890", and so on. Must be different from -secondary-id-suffix (BOUNDARY_DEV_SECONDARY_ID_SUFFIX).`,
+	})
+
+	f.StringVar(&base.StringVar{
+		Name:    "secondary-id-suffix",
+		Target:  &c.flagSecondaryIdSuffix,
+		Default: "0987654321",
+		EnvVar:  "BOUNDARY_DEV_SECONDARY_ID_SUFFIX",
+		Usage:   `If set, secondary auto-created resources will use this value for their identifier (along with their resource-specific prefix). Must be 10 alphanumeric characters. Currently only used for the target resource. Must be different from -id-suffix (BOUNDARY_DEV_ID_SUFFIX).`,
 	})
 
 	f.StringVar(&base.StringVar{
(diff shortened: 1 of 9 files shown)


internal/cmd/commands/dev/dev.go
```diff
@@ -141,7 +142,15 @@ func (c *Command) Flags() *base.FlagSets {
 		Target:  &c.flagIdSuffix,
 		Default: "1234567890",
 		EnvVar:  "BOUNDARY_DEV_ID_SUFFIX",
-		Usage:   `If set, auto-created resources will use this value for their identifier (along with their resource-specific prefix). Must be 10 alphanumeric characters. As an example, if this is set to "1234567890", the generated password auth method ID will be "ampw_1234567890", the generated TCP target ID will be "ttcp_1234567890", and so on.`,
+		Usage:   `Must be different from -secondary-id-suffix (BOUNDARY_DEV_SECONDARY_ID_SUFFIX). If set, auto-created resources will use this value for their identifier (along with their resource-specific prefix). Must be 10 alphanumeric characters. As an example, if this is set to "1234567890", the generated password auth method ID will be "ampw_1234567890", the generated TCP target ID will be "ttcp_1234567890", and so on.`,
```
johanbrandhorst, line 145: Can you put the `Must be different...` detail at the end of the usage? I think it would make it a little easier to read.

internal/cmd/commands/dev/dev.go
```diff
@@ -141,7 +142,15 @@ func (c *Command) Flags() *base.FlagSets {
...
+	})
+
+	f.StringVar(&base.StringVar{
+		Name:    "secondary-id-suffix",
+		Target:  &c.flagSecondaryIdSuffix,
+		Default: "0987654321",
+		EnvVar:  "BOUNDARY_DEV_SECONDARY_ID_SUFFIX",
+		Usage:   `Must be different from -id-suffix (BOUNDARY_DEV_ID_SUFFIX). If set, secondary auto-created resources will use this value for their identifier (along with their resource-specific prefix). Must be 10 alphanumeric characters. Currently only used for the target resource.`,
```
johanbrandhorst, line 153: Ditto

internal/cmd/commands/dev/dev.go
```diff
@@ -427,14 +436,18 @@ func (c *Command) Run(args []string) int {
 	c.Config.Plugins.ExecutionDir = c.flagPluginExecutionDir
 	c.Config.Worker.AuthStoragePath = c.flagWorkerAuthStorageDir
 	if c.flagIdSuffix != "" {
```
johanbrandhorst, line 438: We're not validating the secondary id suffix if the first one isn't set.

hugoamvieira: Good catch, completely missed this top-level if statement!

internal/cmd/commands/dev/dev.go
hugoamvieira: Sure! [13c93ca](https://github.com/hashicorp/boundary/pull/2738/commits/13c93cabcaff097a4fd5f1b33303a468fe366695)

internal/cmd/commands/dev/dev.go
hugoamvieira: [13c93ca](https://github.com/hashicorp/boundary/pull/2738/commits/13c93cabcaff097a4fd5f1b33303a468fe366695)

internal/cmd/commands/dev/dev.go
hugoamvieira: Fixed with [e26ee9b](https://github.com/hashicorp/boundary/pull/2738/commits/e26ee9bde08e04a2e66ef5a981a0e1b895e29e5a)

internal/cmd/commands/dev/dev.go
```diff
@@ -427,11 +436,11 @@ func (c *Command) Run(args []string) int {
...
-		if len(c.flagIdSuffix) != 10 {
+		if len(c.flagIdSuffix) != 10 || len(c.flagSecondaryIdSuffix) != 10 {
```
johanbrandhorst, line 439: This will not report an error about the primary id suffix when the secondary id suffix is wrong. Do we not want to remove these checks from this if statement?

internal/cmd/commands/dev/dev.go
```diff
@@ -450,6 +459,24 @@ func (c *Command) Run(args []string) int {
...
+		}
+		c.DevSecondaryTargetId = fmt.Sprintf("%s_%s", tcp.TargetPrefix, c.flagSecondaryIdSuffix)
+	}
+
+	if c.flagIdSuffix != "" && c.flagSecondaryIdSuffix != "" {
+		if strings.EqualFold(c.flagIdSuffix, c.flagSecondaryIdSuffix) {
+			c.UI.Error("Primary and secondary ID suffixes are equal, must be distinct")
+			return base.CommandUserError
+		}
+	}
```
johanbrandhorst, line 479: We could just write this as a single statement, no?
```suggestion
	if c.flagIdSuffix != "" && c.flagSecondaryIdSuffix != "" &&
		strings.EqualFold(c.flagIdSuffix, c.flagSecondaryIdSuffix) {
		c.UI.Error("Primary and secondary ID suffixes are equal, must be distinct")
		return base.CommandUserError
	}
```

internal/cmd/commands/dev/dev.go
hugoamvieira: Ugh.. I must've hit undo or something. Thanks for spotting this. Fixed with [0c70f05](https://github.com/hashicorp/boundary/pull/2738/commits/0c70f05ddfe65ab9fdb9d9bfd0b19bf5708b0308)

internal/cmd/commands/dev/dev.go
hugoamvieira: Yep - I just thought it read a little clearer with separate statements, addressed in [0c70f05](https://github.com/hashicorp/boundary/pull/2738/commits/0c70f05ddfe65ab9fdb9d9bfd0b19bf5708b0308)

johanbrandhorst (approved):

ddebko (approved):
LGTM

PR #2739, opened 2022-12-21
feature (auth/ldap): add Repository.UpdateAuthMethod(...)

jimlambrt:


internal/auth/ldap/repository_auth_method_update.go
```diff
@@ -0,0 +1,563 @@
...
+	addUrls, deleteUrls, err := valueObjectChanges(ctx, origAm.PublicId, UrlVO, am.Urls, origAm.Urls, dbMask, nullFields)
+	if err != nil {
+		return nil, db.NoRowsAffected, errors.Wrap(ctx, err, op)
+	}
+	addCerts, deleteCerts, err := valueObjectChanges(ctx, origAm.PublicId, CertificateVO, am.Certificates, origAm.Certificates, dbMask, nullFields)
+	if err != nil {
+		return nil, db.NoRowsAffected, errors.Wrap(ctx, err, op)
+	}
+	var addUserSearchConf, deleteUserSearchConf any
+	if strutil.StrListContains(dbMask, UserDnField) || strutil.StrListContains(dbMask, UserAttrField) || strutil.StrListContains(dbMask, UserFilterField) {
```
johanbrandhorst, line 125: This function does a linear search every time. Maybe this logic could benefit from a `StrListContainsAny(list []string, needles ...[]string)`? It could create a set of needles for O(1) lookup if there are enough needles.

johanbrandhorst: There's also https://pkg.go.dev/golang.org/x/exp/slices now which has a `Contains` method.

johanbrandhorst (approved):

internal/auth/ldap/repository_auth_method_update.go
```diff
@@ -0,0 +1,563 @@
...
+		u, err := url.Parse(fmt.Sprintf("%s", i))
+		if err != nil {
+			return nil, errors.Wrap(ctx, err, "ldap.urlFactory")
+		}
+		return NewUrl(ctx, publicId, idx+1, u)
+	},
+}
+
+// valueObjectChanges takes the new and old list of VOs (value objects) and
+// using the dbMasks/nullFields it will return lists of VOs where need to be
```
irenarindos, line 490:
```suggestion
// using the dbMasks/nullFields it will return lists of VOs which need to be
```

internal/auth/ldap/repository_auth_method_update.go
```diff
@@ -0,0 +1,563 @@
...
+func valueObjectChanges(
+	ctx context.Context,
+	publicId string,
+	valueObjectName voName,
+	newVOs,
+	oldVOs,
+	dbMask,
+	nullFields []string,
+) (add []any, del []any, e error) {
+	const op = "valueObjectChanges"
```
irenarindos, line 501:
```suggestion
	const op = "ldap.valueObjectChanges"
```

internal/auth/ldap/repository_auth_method_update.go
```diff
@@ -0,0 +1,563 @@
...
+			}
+			obj, err := factory(ctx, publicId, i, v)
+			if err != nil {
+				return nil, nil, errors.Wrap(ctx, err, op)
+			}
+			adds = append(adds, obj)
+			delete(foundVOs, v)
+		}
+	}
+	if len(foundVOs) > 0 {
```
irenarindos, line 552: I think you can remove this since the for won't range over anything for an empty map

internal/auth/ldap/repository_auth_method_update.go
jimlambrt: ty!

internal/auth/ldap/repository_auth_method_update.go
jimlambrt: ty!

internal/auth/ldap/repository_auth_method_update.go
jimlambrt: ty!

internal/auth/ldap/repository_auth_method_update.go
jimlambrt: I'll wait for the exp pkg to move to the stdlib, but I'm looking fwd to it.

internal/auth/ldap/repository_auth_method_update.go
jimlambrt: I'm not sure that it's empty.  If the VO isn't in the nullFields and it's in the dbMask, but doesn't match an existing oldVO then it will still be in the foundVOs map.

internal/auth/ldap/repository_auth_method_update.go
jimlambrt: pushed a fixup, better?

johanbrandhorst (approved):
LGTM

irenarindos (approved):

PR #2747, opened 2023-01-03
Add check to ensure milestone

jefferai:
Stolen from https://github.com/hashicorp/vault/pull/18406

### DIFF
--- a/.github/workflows/milestone-checker.yml
+++ b/.github/workflows/milestone-checker.yml
@@ -14,7 +14,7 @@ on:
 jobs:
   # checks that a milestone entry is present for a PR
   milestone-check:
-    # If there  a `pr/no-milestone` label we ignore this check
+    # If there is a `pr/no-milestone` label we ignore this check
     if: "!contains(github.event.pull_request.labels.*.name, 'pr/no-milestone')"
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/milestone-checker.yml
+++ b/.github/workflows/milestone-checker.yml
@@ -23,7 +23,7 @@ jobs:
         with:
           repository: "grafana/grafana-github-actions"
           path: ./actions
-          ref: main
+          ref: be89ad434792280ebaa4d982ac72ba548b6f7095
       - name: Install Actions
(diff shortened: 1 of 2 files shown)


.github/workflows/milestone-checker.yml
```diff
@@ -0,0 +1,33 @@
+# This workflow checks that there is either a 'pr/no-milestone' label applied
+to a PR # or there is a milestone associated with a PR
```
johanbrandhorst, line 2: This looks like it's going to cause a yaml parse error

jefferai: Huh, didn't realize it auto wrapped on paste, thanks!

jefferai:
Note: Since the script uses the config json from main it won't pass until it's merged.

.github/workflows/milestone-checker.yml
```diff
@@ -0,0 +1,33 @@
...
+    types: [opened, synchronize, labeled, unlabeled]
+    # Runs on PRs to main and release branches
+    branches:
+      - main
+      - release/**
+
+jobs:
+  # checks that a milestone entry is present for a PR
+  milestone-check:
+    # If there  a `pr/no-milestone` label we ignore this check
```
johanbrandhorst, line 17: Nit
```suggestion
    # If there is a `pr/no-milestone` label we ignore this check
```

.github/workflows/milestone-checker.yml
```diff
@@ -0,0 +1,33 @@
...
+    if: "!contains(github.event.pull_request.labels.*.name, 'pr/no-milestone')"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Actions
+        uses: actions/checkout@v2
+        with:
+          repository: "grafana/grafana-github-actions"
+          path: ./actions
+          ref: main
```
johanbrandhorst, line 26: Can we lock this to a ref to prevent supply chains attack in the grafana github actions?

johanbrandhorst (approved):
🙏🏻 lets hope it builds on `main` 😄

PR #2756, opened 2023-01-05
Create issuingCredentialLibrary Interface

DanielleMiu:
- refactor vault.(issueCredentialLibrary).retrieveCredential to no longer use session id
- remove `dynamicCred`'s `insertQuery` and `updateSessionQuery` functions
- update `TestRepository_IssueCredentials` to perform db lookup for issued credentials
- rename `issueCredentialLibrary` to `genericIssuingCredentialLibrary`
- create new `privateCredentialLibraryAllTypes` struct to handle db operations and decryption
- add `issuingCredentialLibrary` interface to handle credential retrieval in a generic way in preparation for new credential types

### DIFF
--- a/internal/credential/vault/private_library.go
+++ b/internal/credential/vault/private_library.go
@@ -33,6 +33,7 @@ func (bc *baseCred) Secret() credential.SecretData { return bc.secretData }
 func (bc *baseCred) Library() credential.Library   { return bc.lib }
 func (bc *baseCred) Purpose() credential.Purpose   { return bc.lib.Purpose }
 func (bc *baseCred) getExpiration() time.Duration  { return bc.expiration }
+func (bc *baseCred) getCredential() *Credential    { return bc.Credential }
 
 // convert converts bc to a specific credential type if bc is not
 // UnspecifiedType.
@@ -284,15 +285,14 @@ func (pl *issueCredentialLibrary) client(ctx context.Context) (vaultClient, erro
 type dynamicCred interface {
 	credential.Dynamic
 	getExpiration() time.Duration
-	insertQuery() (query string, queryValues []any)
-	updateSessionQuery(purpose credential.Purpose) (query string, queryValues []any)
+	getCredential() *Credential
 }
 
 // retrieveCredential retrieves a dynamic credential from Vault for the
 // given sessionId.
 //
 // Supported options: credential.WithTemplateData
-func (pl *issueCredentialLibrary) retrieveCredential(ctx context.Context, op errors.Op, sessionId string, opt ...credential.Option) (dynamicCred, error) {
+func (pl *issueCredentialLibrary) retrieveCredential(ctx context.Context, op errors.Op, opt ...credential.Option) (dynamicCred, error) {
 	opts, err := credential.GetOpts(opt...)
 	if err != nil {
 		return nil, errors.Wrap(ctx, err, op)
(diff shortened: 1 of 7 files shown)


internal/credential/vault/private_library.go
```diff
@@ -429,9 +402,157 @@ func (r *Repository) getIssueCredLibraries(ctx context.Context, requests []crede
...
+		TokenKeyId:                    pl.TokenKeyId,
+		ClientCert:                    append(pl.ClientCert[:0:0], pl.ClientCert...),
+		ClientKey:                     append(pl.ClientKey[:0:0], pl.ClientKey...),
+		CtClientKey:                   append(pl.CtClientKey[:0:0], pl.CtClientKey...),
+		ClientKeyId:                   pl.ClientKeyId,
+		Purpose:                       pl.Purpose,
+	}
+}
+
+func (pl *privateCredentialLibraryAllTypes) toTypedIssuingCredentialLibrary() *genericIssuingCredentialLibrary {
```
tmessi, line 520: This should return `issuingCredentialLibrary` so that it can be expanded to return other typed issuing credential libraries in the future.

internal/credential/vault/private_library.go
```diff
@@ -429,9 +402,157 @@ func (r *Repository) getIssueCredLibraries(ctx context.Context, requests []crede
...
+		TokenHmac:                     append(pl.TokenHmac[:0:0], pl.TokenHmac...),
+		Token:                         append(pl.Token[:0:0], pl.Token...),
+		CtToken:                       append(pl.CtToken[:0:0], pl.CtToken...),
+		TokenKeyId:                    pl.TokenKeyId,
+		ClientCert:                    append(pl.ClientCert[:0:0], pl.ClientCert...),
+		ClientKey:                     append(pl.ClientKey[:0:0], pl.ClientKey...),
+		CtClientKey:                   append(pl.CtClientKey[:0:0], pl.CtClientKey...),
+		ClientKeyId:                   pl.ClientKeyId,
+		Purpose:                       pl.Purpose,
+	}
```
tmessi, line 555: I think this can just set the fields, it doesn't need to "clone" them, so it doesnt need to call `proto.Clone` and doesnt need to do the `append(a[:0:0], a)`

tmessi (changes requested):
just a couple minor things

internal/credential/vault/private_library.go
DanielleMiu: that's completely true, I just copied the clone function last night. it makes more sense not to clone it

internal/credential/vault/repository_credentials.go
```diff
@@ -54,8 +87,28 @@ func (r *Repository) Issue(ctx context.Context, sessionId string, requests []cre
...
+		switch c := cred.(type) {
+		case *usrPassCred:
+			underlyingCred = c.Credential
+		case *sshPrivateKeyCred:
+			underlyingCred = c.Credential
+		case *baseCred:
+			underlyingCred = c.Credential
+		default:
+			return nil, errors.New(ctx, errors.InvalidDynamicCredential, op, fmt.Sprintf("vault dynamicCred (typeof %T) could not be mapped back to vault.Credential", c))
+		}
```
tmessi, line 101: I think we can eliminate the need for this type switch by expanding the `dynamicCred` interface. There are a few ways this could be done. The main motivation would be so that when new implementations of `dynamicCred` are added, we won't need to remember to add a case to this switch. Instead if the interface isn't implemented completely, we would get a compile-time error.

One way to do this would be to update `dynamicCred` to have:

```go
type dynamicCred interface {
  // other methods
  getCredential() *Credential
}
```

Then update `baseCred` to implement this method:

```go
func (bc *baseCred) getCredential() *Credential    { return bc.Credential }
```

Then this switch can be replaced with:

```go
underlyingCred := cred.getCredential()
```

tmessi (approved):

PR #2759, opened 2023-01-05
Fix string regression in attribute parsing

jefferai:
In #2721, the ability for key-only values was added via using a wrapperspb.StringValue that could be null. This introduced a bug because originally the default case for `-attr` was to get the string value but instead it used the incoming value as-is to allow for nulls; the null check was moved up later but the original behavior was not restored. As a result, non-explicitly-quoted strings carried through the wrapperspb.StringValue, resulting in a map for the value in JSON instead of a string.

This also fixes explicitly quoted string handling.


tmessi (approved):
Looks good. You will want to reword your commit when you go to merge, since this now adds official support for explicitly quoted strings, rather than removing it.

jefferai:
Yep -- updated the PR description already, and will make sure I update the commit message. Thanks for the reminder.

PR #2760, opened 2023-01-05
feature (auth/ldap): add managed groups

jimlambrt:
I believe this PR may finish up the repository portion for the LDAP auth method. 

@mgaffney -> you're welcome to review the entire PR, but I'd be very happy if you only had time for the [SQL bits](https://github.com/hashicorp/boundary/pull/2760/files#diff-17b5a87e45311928aa43aba376e1dc155f08c370308d2540c03eb5333ad70977)


internal/auth/ldap/repository_managed_group_test.go
```diff
@@ -0,0 +1,1224 @@
...
+			want: &ManagedGroup{
+				ManagedGroup: &store.ManagedGroup{
+					AuthMethodId: testAuthMethod.PublicId,
+					GroupNames:   string(TestEncodedGrpNames(t, notTestGroupName)),
+				},
+			},
+			wantMgmAcctIds: []string{acct3.PublicId},
+		},
+		{
+			name:    "duplicate-name", // must follow "with-valid-description" test
```
irenarindos, line 214:
```suggestion
			name:    "duplicate-name", // must follow "valid-with-description" test
```

internal/proto/controller/storage/auth/ldap/store/v1/ldap.proto
```diff
@@ -473,3 +473,42 @@ message AccountAttributeMap {
...
+
+  // @inject_tag: `gorm:"default:null"`
+  uint32 version = 60;
+
+  // auth_method_id is the fk to the account's auth method.
+  // @inject_tag: `gorm:"not_null"`
+  string auth_method_id = 70;
+
+  // groups is json marshalled list of groups that make up the ManagedGroup
+  string group_names = 80;
```
irenarindos, line 513: Missing inject_tag?

internal/auth/ldap/managed_group.go
```diff
@@ -0,0 +1,112 @@
...
+	return mg, nil
+}
+
+// validate the Managed Group.  On success, it will return nil.
+func (mg *ManagedGroup) validate(ctx context.Context, caller errors.Op) error {
+	if mg.AuthMethodId == "" {
+		return errors.New(ctx, errors.InvalidParameter, caller, "missing auth method id")
+	}
+	if len(mg.GroupNames) == 0 {
+		return errors.New(ctx, errors.InvalidParameter, caller, "missing filter")
```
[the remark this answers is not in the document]
jimlambrt: ty!

internal/auth/ldap/repository_managed_group_test.go
jimlambrt: ty!

internal/proto/controller/storage/auth/ldap/store/v1/ldap.proto
jimlambrt: ty!

internal/auth/testing.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	return "auth_managed_group_member_account"
+}
+
+// TestSortManagedGroupMemberAccounts simply sorts them by public id to make
+// comparisons a bit easier.
+func TestSortManagedGroupMemberAccounts(t testing.TB, m []*ManagedGroupMemberAccount) {
+	sort.Slice(m, func(a, b int) bool {
+		return m[a].MemberId < m[b].MemberId
+	})
```
johanbrandhorst, line 42: Only noting this because you've used the `slices` package in other places, maybe you want https://pkg.go.dev/golang.org/x/exp/slices#Sort?

johanbrandhorst (approved):

internal/auth/testing.go
jimlambrt: the first clone is needed to ensure that `mg is not changed` (part of the contract) before we add the public id.    This clone is needed to ensure that retried transactions have an changed copy to work on (since create could change it).

internal/auth/ldap/repository_managed_group.go
```diff
@@ -0,0 +1,261 @@
...
+
+	oplogMetadata, err := mg.oplog(ctx, oplog.OpType_OP_TYPE_CREATE, scopeId)
+	if err != nil {
+		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("unable to generate managed group oplog metadata"))
+	}
+
+	var newManagedGroup *ManagedGroup
+	_, err = r.writer.DoTx(ctx, db.StdRetryCnt, db.ExpBackoff{},
+		func(_ db.Reader, w db.Writer) error {
+			newManagedGroup = mg.clone()
```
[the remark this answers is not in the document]
jimlambrt: the first clone is needed to ensure that mg is not changed (part of the contract) before we add the public id. This clone is needed to ensure that retried transactions have an changed copy to work on (since create could change it).

internal/auth/testing.go
jimlambrt: We're using `sort.Slice(...)` in 14 places already in this pkg.     I think, slice.SortFunc(...) would be the equivalent and I'm not sure why I'd chose one over the other.... and I don't think any of our "sort" use cases warrant benchmarks to prove which is better.   For now, it's more consistent to stick with sort.Slice but in all honesty is "potato" vs "potato" for now until there's benchmark data.

jimlambrt:
@mgaffney -> I'm happy to make post merge changes if you request them for the [sql bits](https://github.com/hashicorp/boundary/pull/2760/files#diff-17b5a87e45311928aa43aba376e1dc155f08c370308d2540c03eb5333ad70977)

internal/auth/testing.go
johanbrandhorst: Totally fair, I will just note that I think `slices.Sort` is strictly better because it provides static type safety while `sort.Slice` has to accept an `any`. You also don't need to specify a sort function if you're using the basic comparison of the type. I don't know about any performance differences. Definitely worth staying consistent in this package though 👍🏻.

PR #2765, opened 2023-01-09
feat(vault): domain methods for ssh certificate credential library

A-440Hz:


internal/credential/vault/repository_ssh_certificate_credential_library.go
```diff
@@ -0,0 +1,344 @@
...
+//
+// l must contain a valid PublicId. Only Name, Description, VaultPath,
+// HttpMethod, HttpRequestBody, and MappingOverride can be updated. If
+// l.Name is set to a non-empty string, it must be unique within l.StoreId.
+//
+// An attribute of l will be set to NULL in the database if the attribute
+// in l is the zero value and it is included in fieldMaskPaths except for
+// HttpMethod.  If HttpMethod is in the fieldMaskPath but l.HttpMethod
+// is not set it will be set to the value "GET".  If storage has a value
+// for HttpRequestBody when l.HttpMethod is set to GET the update will fail.
```
tmessi, line 119: This comment needs updating since it references fields that are not a part of a SSHCerticiateCredentialLibrary.

internal/credential/vault/ssh_certificate_credential_library.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+const (
+	KeyTypeEcdsa   = "ecdsa"
+	KeyTypeEd25519 = "ed25519"
+	KeyTypeRsa     = "rsa"
+)
+
+// A CredentialLibrary contains a Vault path and is owned by a credential
+// store.
```
tmessi, line 17: This comment needs updating

```suggestion
// SSHCertificateCredentialLibrary is a credential library that issues
// ssh certificate using the vault ssh secret engine.
// See: https://developer.hashicorp.com/vault/api-docs/secret/ssh#sign-ssh-key
```

internal/credential/vault/ssh_certificate_credential_library.go
```diff
@@ -0,0 +1,98 @@
...
+type SSHCertificateCredentialLibrary struct {
+	*store.SSHCertificateCredentialLibrary
+	tableName string `gorm:"-"`
+}
+
+// NewSSHCertificateCredentialLibrary creates a new in memory CredentialLibrary
+// for a Vault backend at vaultPath assigned to storeId.
+// Name, description, method, request body, credential type, and mapping
+// override are the only valid options. All other options are ignored.
```
tmessi, line 26: This also needs updating.

internal/credential/vault/repository_ssh_certificate_credential_library_test.go
```diff
@@ -0,0 +1,1294 @@
...
+				},
+			},
+			wantErr: errors.InvalidParameter,
+		},
+		{
+			name: "invalid-no-vault-path",
+			in: &SSHCertificateCredentialLibrary{
+				SSHCertificateCredentialLibrary: &store.SSHCertificateCredentialLibrary{
+					StoreId:   cs.GetPublicId(),
+					VaultPath: "/some/path",
```
tmessi, line 78: This looks like it is setting a vault path, despite the test name being "invalid-no-vault-path"

internal/credential/vault/repository_ssh_certificate_credential_library_test.go
```diff
@@ -0,0 +1,1294 @@
...
+			},
+			want: &SSHCertificateCredentialLibrary{
+				SSHCertificateCredentialLibrary: &store.SSHCertificateCredentialLibrary{
+					StoreId:     cs.GetPublicId(),
+					KeyType:     "ed25519",
+					Description: "test-description-repo",
+					VaultPath:   "/some/path",
+					Username:    "admin",
+				},
+			},
```
tmessi, line 142: Can you add some additional test cases, specifically around:
- `KeyType` and `KeyBits` both valid and invalid combinations
- CriticalOptions
- Extensions

internal/credential/vault/repository_ssh_certificate_credential_library_test.go
```diff
@@ -0,0 +1,1294 @@
...
+	}
+}
+
+func TestRepository_UpdateSSHCertificateCredentialLibrary(t *testing.T) {
+	t.Parallel()
+	conn, _ := db.TestSetup(t, "postgres")
+	rw := db.New(conn)
+	wrapper := db.TestWrapper(t)
+
+	// TODO: add more field tests
```
tmessi, line 450: I think this todo is still needed? Specifically around changes to combinations of KeyType and KeyBits.

tmessi (changes requested):

A-440Hz:
I am having trouble setting `key bits` to 0 in an update call.

I have two testcases:

1. change `key bits` and `key type` in the same update call
    - afaik this should error, and the preferred user behavior is to update the `key bits` to 0 first and then change `key type` to the desired type, and then update the correct `key bits` afterward if necessary. 
        - It also shouldn't be that bad to check if both fields are present in an update call, and automatically do things in that order if needed. I haven't seen that pattern before in other areas though, so I'll hold off indefinitely on adding it.

2. update the `key bits` to 0 first and then change `key type` to the desired type, and then update the correct `key bits` afterward if necessary.
    - this testcase should be the default (only) valid path for a user trying to switch `key type` and select an appropriate `key bits`. Please correct me if I'm wrong.

When I first ran testcase 2, I got a "violates not null constraint" error when I updated the `key bits` field from an "int > 0" value to 0.

I then tried changing the `key bits` schema constraint from `not null` to `null` (which doesn't seem like the right thing to do, since int 0 shouldn't count as null anyway), but this caused a segfault in testcase 1 after running  `repo.UpdateSSHCertificateCredentialLibrary` and attempting to access it in the db with `dbassert`. I don't know why this happens.

I noticed a `wrapperpb.UInt32` type being used in similar places, so I tried reverting `key bits` back to `not null`, and then switching `key bits` from uint32 to wrapperspb.UInt32, thinking it might fix the "0 to null" schema conversion.  Currently I am getting the error

> db.DoTx: db.Create: unknown, unknown: error #0: dbw.Create: create failed: invalid field found for struct github.com/hashicorp/boundary/internal/credential/vault.SSHCertificateCredentialLibrary's field KeyBits: define a valid foreign key for relations or implement the Valuer/Scanner interface

when I run the tests. It seems like I am missing something on the schema side to be able to use this data type.

internal/credential/vault/repository_ssh_certificate_credential_library.go
```diff
@@ -0,0 +1,341 @@
...
+			usernameField:        l.Username,
+			keyTypeField:         l.KeyType,
+			keyBitsField:         l.KeyBits,
+			ttlField:             l.Ttl,
+			keyIdField:           l.KeyId,
+			criticalOptionsField: l.CriticalOptions,
+			extensionsField:      l.Extensions,
+		},
+		fieldMaskPaths,
+		nil,
```
tmessi, line 168: You need to add `keyBitsField` here so it is allowed to be `0`. Othwerwise `dbw.BuildUpdatePaths` will convert the `0` to a `null`, which is what is causing the null constraint error when trying to update KeyBits to 0.

```suggestion
[]string{keyBitsField},
```

tmessi (changes requested):
> 1. change `key bits` and `key type` in the same update call
>        
>    * afaik this should error, and the preferred user behavior is to update the `key bits` to 0 first and then change `key type` to the desired type, and then update the correct `key bits` afterward if necessary.

I don't think this is correct. You should be able to change both `key type` and `key bits` at the same time. As long as the new values are a valid combination it should allow the update. So the test that you have that starts with `ecdsa 256` and updates to `rsa 2048` should *not* error.

> When I first ran testcase 2, I got a "violates not null constraint" error when I updated the `key bits` field from an "int > 0" value to 0.

This is due to not marking that key bits can be allowed to be zero when building the dbMask and nullFields (see inline comment below).

internal/credential/vault/repository_ssh_certificate_credential_library.go
A-440Hz: Thanks. I'm sad I didn't pursue this. I saw the other credential library didn't use this field and I assumed this was not the problem

PR #2767, opened 2023-01-09
Add default client port to targets and use in connect command

jefferai:


doratias18:
excited about this

jefferai:
@doratias18 It needs some rework so won't be in 0.12.0 but it's coming eventually. Glad you're excited :-)

doratias18:
@jefferai great to hear! Do you know if it will be supported in the UI as well once this PR is merged and released?

internal/cmd/commands/targetscmd/ssh_funcs.go
```diff
@@ -130,6 +137,19 @@ func extraSshFlagsHandlingFuncImpl(c *SshCommand, _ *base.FlagSets, opts *[]targ
 		*opts = append(*opts, targets.WithSshTargetDefaultPort(uint32(port)))
 	}
 
+	switch c.flagDefaultClientPort {
+	case "":
+	case "null":
+		*opts = append(*opts, targets.DefaultSshTargetDefaultClientPort())
+	default:
+		port, err := strconv.ParseUint(c.flagDefaultClientPort, 10, 32)
```
johanbrandhorst, line 145: Very much nit but aren't port numbers limited to 16 bit unsigned integers?

johanbrandhorst (approved):

internal/cmd/commands/targetscmd/ssh_funcs.go
jefferai: Yes, but the smallest int size in protobuf is 32 bits. We could technically do translation after generation so that our API uses 16-bit unsigned ints but we didn't in the past for other port values so I didn't here 🤷

johanbrandhorst (approved):

jimlambrt (approved):
Looks great... I've seen most of this before 😄

PR #2793, opened 2023-01-15
fix (controller): prevent panic when controller stops when there's no listener

jimlambrt:


jefferai:
Should this be 0.11.x? Or is this new?

jimlambrt:
I'm okay with 0.12.x since I don't think it's urgent and is likely unreachable unless there's a defect elsewhere.

PR #2803, opened 2023-01-18
feat(wh): Support vault ssh certificate credentials

tmessi:
This expands the wh_credential_dimension to include fields for the new
vault ssh certificate credential library.


internal/db/schema/migrations/oss/postgres/62/02_wh_ssh_cert_library.up.sql
```diff
@@ -0,0 +1,221 @@
...
+              end                                      as credential_library_username,
+              case
+                when   vcl.public_id is not null then 'Not Applicable'
+                when vsccl.public_id is not null then
+                  case
+                    when vsccl.key_type = 'ed25519' then vsccl.key_type
+                    else vsccl.key_type || '-' || vsccl.key_bits::text
+                  end
+                else 'Unknown'
+              end                                      as credential_library_key_type_and_bits,
```
mgaffney, line 73: Noting this is the area we discussed in our zoom call

mgaffney (changes requested):

internal/db/schema/migrations/oss/postgres/62/01_credential_vault_ssh_cert_library.up.sql
```diff
@@ -155,8 +149,29 @@ begin;
...
+        new.key_bits = 256;
+      when new.key_type = 'rsa' then
+        new.key_bits = 2048;
+      else
+        new.key_bits = 0;
+      end case;
+    end if;
+    return new;
+  end;
+  $$ language plpgsql;
```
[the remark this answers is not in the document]
tmessi: Yep, for now I just removed this from the migration, I will review the logic we have in the domain layer and add any changes if needed.

mgaffney (approved):
The SQL is very readable. Great job!

internal/credential/vault/private_library_test.go
```diff
@@ -1333,15 +1333,6 @@ func TestRepository_sshCertIssuingCredentialLibrary_retrieveCredential(t *testin
...
 		},
-		{
-			name:     "vault sign boundary ec(224) invalid key",
-			username: "username-11-lucky-number",
-			expected: map[string]string{
-				"error": "invalid KeyBits. when KeyType=ecdsa, KeyBits must be one of: 256, 384, or 521",
-			},
-			vaulthPath: "ssh/sign/boundary",
-			opts:       []Option{WithKeyType(KeyTypeEcdsa), WithKeyBits(224)},
-		},
```
tmessi, line 1344: This test was removed since the setup is no longer possible, the schema changes in this PR make it so a library can not be created with `224` for key bits.

internal/credential/vault/ssh_certificate_credential_library.go
```diff
@@ -11,6 +11,16 @@ const (
...
+
+	KeyBitsDefault = 0
+
+	KeyBitsEcdsa256 = 256
+	KeyBitsEcdsa384 = 384
+	KeyBitsEcdsa521 = 521
+
+	KeyBitsRsa2048 = 2048
+	KeyBitsRsa3072 = 3072
+	KeyBitsRsa4096 = 4096
```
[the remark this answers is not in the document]
tmessi: We should update the checks to use this consts, I thought I caught them all, but I can update the ones in the validation checks.

internal/credential/vault/repository_ssh_certificate_credential_library.go
```diff
@@ -150,6 +164,20 @@ func (r *Repository) UpdateSSHCertificateCredentialLibrary(ctx context.Context,
 			return nil, db.NoRowsAffected, errors.New(ctx, errors.InvalidFieldMask, op, f)
 		}
 	}
+
+	if keyTypeChange && l.KeyType == "" {
+		l.KeyType = KeyTypeEd25519
+	}
+
+	if keyTypeChange && keyBitChangeDefault {
```
DanielleMiu, line 172: `keyTypeChange` required in this conditional? keybits could change to default but keytype doesn't resulting in invalid keybits

internal/daemon/controller/handlers/credentiallibraries/credentiallibrary_service.go
```diff
@@ -1066,12 +1066,12 @@ func validateMapping(badFields map[string]string, credentialType credential.Type
...
+	case vault.KeyBitsDefault:
+	case vault.KeyBitsEcdsa256, vault.KeyBitsEcdsa384, vault.KeyBitsEcdsa521:
 		if keyType != "" && keyType != "ecdsa" {
 			badFields[keyBitsField] = fmt.Sprintf("Invalid bit size %d for key type %s", keyBits, keyType)
 		}
-	case 2048, 3072, 4096:
+	case vault.KeyBitsRsa2048, vault.KeyBitsRsa3072, vault.KeyBitsRsa4096:
 		if keyType != "" && keyType != "rsa" {
 			badFields[keyBitsField] = fmt.Sprintf("Invalid bit size %d for key type %s", keyBits, keyType)
 		}
```
DanielleMiu, line 1077: `"ecdsa"` and `"rsa"` should probably use consts `vault.KeyTypeEcdsa` and `vault.KeyTypeRsa`

internal/credential/vault/private_library_test.go
DanielleMiu: the setup is technically not possible, but the validation still exists in the function, and this test still correctly checks that the validation works. imo this test should not be removed

tmessi: Well, this test was failing during the setup phase due to it not being possible. I suppose we could implement the setup another way...

DanielleMiu: oh gotcha, you're right those tests work by going through the db which wouldn't be possible with your changes. I think it's ok in that case. important part is that all code paths from db are checked

internal/daemon/controller/handlers/credentiallibraries/credentiallibrary_service.go
tmessi: good catch

internal/credential/vault/repository_ssh_certificate_credential_library.go
tmessi: Yep, good catch. I will add some logic if only the key bits change. In which case we need to look at the keytype from `origLib` below

tmessi: New tests for this case ecc3e7dcb and implementation to make those test pass 1ae357271

DanielleMiu (approved):
looks good!

PR #2814, opened 2023-01-19
Add automated cleanup for CI resources

joshbrand:


.github/workflows/test-ci-cleanup.yml
```diff
@@ -0,0 +1,127 @@
+name: test-ci-cleanup
+on:
+  push:
```
mikebaum, line 3: don't forget to delete this.

.github/workflows/test-ci-cleanup.yml
```diff
@@ -0,0 +1,127 @@
...
+    - cron:  '05 02 * * *'
+
+jobs:
+  setup:
+    runs-on: ubuntu-latest
+    outputs:
+      regions: ${{steps.setup.outputs.regions}}
+    steps:
+      - name: Assume service user role via Doormat
+        uses: hashicorp/doormat-action@main
```
mikebaum, line 18: The workflow fails, claiming this action does not exist, maybe that's since this is the OSS repo. So I'm not sure we can use this pattern, smart though.

mikebaum: maybe use the pattern you used in the next job when assuming the creds.

mikebaum (commented):
Just a couple of comments. I know it's draft :)

.github/workflows/test-ci-cleanup.yml
joshbrand: this was the end result of trying lots of things to make the if statement work, they all inevitably failed trying to fetch the action (which is private.) So I'll be continuing with your pattern from before.

mikebaum (approved):
Looks good to me

mikebaum (approved):

PR #2823, opened 2023-01-20
feat(credential): adds service crudl for ssh cert credential library

A-440Hz:


internal/daemon/controller/handlers/credentiallibraries/credentiallibrary_service.go
```diff
@@ -372,26 +448,47 @@ func (s Service) updateInRepo(
...
 	if err != nil {
 		return nil, errors.Wrap(ctx, err, op)
 	}
-	out, rowsUpdated, err := repo.UpdateCredentialLibrary(ctx, projId, cl, item.GetVersion(), dbMasks)
-	if err != nil {
-		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("unable to update credential library"))
-	}
-	if rowsUpdated == 0 {
-		return nil, handlers.NotFoundErrorf("Credential Library %q doesn't exist or incorrect version provided.", id)
+	switch subtypes.SubtypeFromType(domain, item.GetType()) {
```
tmessi, line 457: An update request wont have the type provided, instead this should switch on `subtypes.SubtypeFromId(domain, id)`

A-440Hz: thanks for catching!

internal/daemon/controller/handlers/credentiallibraries/credentiallibrary_service_test.go
```diff
@@ -32,6 +32,7 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 	"google.golang.org/protobuf/types/known/wrapperspb"
+	"gopkg.in/square/go-jose.v2/json"
```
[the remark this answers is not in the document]
A-440Hz: The unmarshal in the CL service file seems to be this import too. Should I change it to encoding/json as well?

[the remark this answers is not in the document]
tmessi: yea

internal/daemon/controller/handlers/credentiallibraries/credentiallibrary_service_test.go
```diff
@@ -101,12 +121,6 @@ func TestList(t *testing.T) {
 			res:     &pbs.ListCredentialLibrariesResponse{Items: wantLibraries[1:2]},
 			anonRes: &pbs.ListCredentialLibrariesResponse{Items: wantLibraries[1:2]},
 		},
-		{
-			name:    "Filter on Attribute",
-			req:     &pbs.ListCredentialLibrariesRequest{CredentialStoreId: store.GetPublicId(), Filter: fmt.Sprintf(`"/item/attributes/path"==%q`, wantLibraries[2].GetVaultCredentialLibraryAttributes().GetPath().Value)},
-			res:     &pbs.ListCredentialLibrariesResponse{Items: wantLibraries[2:3]},
-			anonRes: &pbs.ListCredentialLibrariesResponse{}, // anonymous user does not have access to attributes
-		},
```
tmessi, line 109: Why was this removed?

tmessi (changes requested):

internal/daemon/controller/handlers/credentiallibraries/credentiallibrary_service_test.go
A-440Hz: I remember now. I moved the filter on attributes tests to [TestList_Attributes](https://github.com/hashicorp/boundary/blob/45a51ee23c33cb379065569f26a5ae8c693b007a/internal/daemon/controller/handlers/credentiallibraries/credentiallibrary_service_test.go#L1773) because something was messing up the filter order when I put both library types in the same list/store. TestList_Attributes is exactly the same as TestList, but the different CL types are in separate credential stores.

internal/daemon/controller/handlers/credentiallibraries/credentiallibrary_service.go
```diff
@@ -670,17 +950,29 @@ func validateCreateRequest(req *pbs.CreateCredentialLibraryRequest) error {
...
 
 func validateUpdateRequest(req *pbs.UpdateCredentialLibraryRequest, currentCredentialType credential.Type) error {
+	prefix := ""
+	st := subtypes.SubtypeFromId(domain, req.GetId())
+	switch st {
+	case vault.GenericLibrarySubtype:
+		prefix = vault.CredentialLibraryPrefix
+	case vault.SSHCertificateLibrarySubtype:
+		prefix = vault.SSHCertificateCredentialLibraryPrefix
+	}
```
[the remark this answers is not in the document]
A-440Hz: I dont think so, unless I change the function parameters, since we are individually returning `prefix` at the end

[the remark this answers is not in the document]
tmessi: ah I see it now :+1:

A-440Hz:
latest merge fixes options.go since I forgot to push the change I made yesterday

internal/credential/vault/options.go
```diff
@@ -182,42 +182,46 @@ func WithMappingOverride(m MappingOverride) Option {
...
 		o.withKeyBits = b
 	}
 }
 
-// WithTtl provides an optional vault token ttl.
+// WithTtl provides an optional requested time to live for a generated ssh certificate.
 func WithTtl(t string) Option {
 	return func(o *options) {
 		o.withTtl = t
+		o.withTtl = t
```
tmessi, line 204: looks like this line got added twice

PR #2829, opened 2023-01-23
docs: Remove links from release note headings - 0.11

Dan-Heath:
This PR removes the links from the H1 headings in the release notes.  From the web team:

> We're trying to remove instances of links in heading elements, as it can cause difficulties for users who rely on screen readers and voice-based assistive tech, and it can lead to issues when we automatically generation anchor links.


stellarsquall (approved):
great work 👍

PR #2842, opened 2023-01-24
feat(cli): Add vault ssh certificate credential library

tmessi:


internal/tests/cli/Makefile
```diff
@@ -1,5 +1,36 @@
...
+TEST_VAULT_PORT ?= 8200
+TEST_VAULT_CONTAINER_NAME ?= boundary-cli-tests-vault
+
+VAULT_ADDR ?= "http://127.0.0.1:$(TEST_VAULT_PORT)"
+VAULT_TOKEN ?= boundarytok
+
+.PHONY: test-vault-up
+test-vault-up:
+	@echo "Using image: $(TEST_VAULT_IMAGE)"
+	@echo " vault addr: $(VAULT_ADDR)"
```
moduli, line 15: Extra space?
```suggestion
	@echo "vault addr: $(VAULT_ADDR)"
```

internal/tests/cli/boundary/_vault.bash
```diff
@@ -0,0 +1,47 @@
...
+    -policy="boundary-controller" \
+    -orphan=true \
+    -period=2m \
+    -renewable=true | \
+    jq -r '.auth.client_token'
+}
+
+function skip_if_no_vault() {
+  if [[ -z $VAULT_ADDR ]] || [[ -z $VAULT_TOKEN ]]; then
+    skip "vault environment variables  \$VAULT_ADDR and \$VAULT_TOKEN not set"
```
moduli, line 45: Extra space?
```suggestion
    skip "vault environment variables \$VAULT_ADDR and \$VAULT_TOKEN not set"
```

.github/workflows/build.yml
```diff
@@ -479,6 +486,13 @@ jobs:
...
+          echo "vault version:"
+          vault version
       - name: Run cli bats tests
         run: |
+          export VAULT_ADDR=http://127.0.0.1:8200
+          export VAULT_TOKEN=boundarytok
+          make -C internal/tests/cli test-vault-up
+          until vault status; do docker container inspect boundary-cli-tests-vault &> /dev/null || exit -1; sleep 1; done
           make test-cli
+          make -C internal/tests/cli test-vault-down
```
moduli, line 498: Does this run if a bats test fails?

internal/tests/cli/Makefile
tmessi: This is purposeful so the echos all align on the `:`.

.github/workflows/build.yml
tmessi: Split this out into a separate step that will run even on failure.

.github/workflows/build.yml
```diff
@@ -453,6 +453,13 @@ jobs:
           cd /tmp/bats-cli-ui-deps/pass/password-store-1.7.4
           sudo make install
           pass init ${{ secrets.ENOS_GPG_UID }}
+      - name: Download Vault AMD64 binary for integration testing
+        if: steps.dep-cache.outputs.cache-hit != 'true'
+        run: |
+          wget https://releases.hashicorp.com/vault/1.12.2/vault_1.12.2_linux_amd64.zip -O /tmp/bats-cli-ui-deps/vault.zip
```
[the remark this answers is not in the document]
tmessi: I like that idea, any thoughts on the best way to do that?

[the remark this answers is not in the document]
moduli: Hm... I was looking at this: https://docs.github.com/en/actions/learn-github-actions/variables#defining-environment-variables-for-a-single-workflow. Hopefully we can insert that into the `wget` url?

A-440Hz (approved):

.github/workflows/build.yml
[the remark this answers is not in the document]
tmessi: Somehow that also needs to work its way into the key used for the caching. I think I will look into this as a follow up.

internal/tests/cli/boundary/credential_libraries.bats
```diff
@@ -0,0 +1,181 @@
...
+  [ "$status" -eq 0 ]
+}
+
+@test "boundary/credential-libraries: can not create already created $NEW_VAULT_LIB vault-ssh-certificate libary" {
+  skip_if_no_vault
+
+  local csid=$(credential_store_id $NEW_STORE $DEFAULT_P_ID)
+  run create_vault_ssh_certificate_library \
+    -name $NEW_VAULT_LIB -credential-store-id $csid \
+    -vault-path ssh/sign/foo \
```
moduli, line 142: Was it intentional to not have a leading slash here?

moduli (approved):

DanielleMiu (approved):

PR #2846, opened 2023-01-25
Fix target port handling

jefferai:
This fixes two issues that compounded on each other (see the Changelog update for more information):

* The verification logic for hosts was not correct for update operations (in multiple ways) which meant that a host could be updated after creation to have a port. Targets had a previously fixed bug where they did not require a default port, which meant that ports could be used from hosts
* A recent (unreleased) change had prioritized any port coming from the host over the default port, which would mean there was no way if a host had a port specified to use it in multiple targets.

This fixes the update verification logic, and strikes a middle ground between breaking things and not by allowing existing addresses with ports to be used with targets but ignoring that port, instead requiring targets to have default port set at authorize time (currently there is backwards compat that does not require this due to the original optional port bug).

### DIFF
--- a/internal/daemon/controller/handlers/hosts/host_service.go
+++ b/internal/daemon/controller/handlers/hosts/host_service.go
@@ -643,15 +643,16 @@ func validateCreateRequest(req *pbs.CreateHostRequest) error {
 					len(attrs.GetAddress().GetValue()) < static.MinHostAddressLength ||
 					len(attrs.GetAddress().GetValue()) > static.MaxHostAddressLength {
 					badFields["attributes.address"] = fmt.Sprintf("Address length must be between %d and %d characters.", static.MinHostAddressLength, static.MaxHostAddressLength)
-				}
-				_, _, err := net.SplitHostPort(attrs.GetAddress().GetValue())
-				switch {
-				case err == nil:
-					badFields["attributes.address"] = "Address for static hosts does not support a port."
-				case strings.Contains(err.Error(), "missing port in address"):
-					// Bare hostname, which we want
-				default:
-					badFields["attributes.address"] = fmt.Sprintf("Error parsing address: %v.", err)
+				} else {
+					_, _, err := net.SplitHostPort(attrs.GetAddress().GetValue())
+					switch {
+					case err == nil:
+						badFields["attributes.address"] = "Address for static hosts does not support a port."
+					case strings.Contains(err.Error(), "missing port in address"):
+						// Bare hostname, which we want
+					default:
+						badFields["attributes.address"] = fmt.Sprintf("Error parsing address: %v.", err)
+					}
 				}
 			}
 		case plugin.Subtype:
@@ -668,17 +669,26 @@ func validateUpdateRequest(req *pbs.UpdateHostRequest) error {
 		case static.Subtype:
 			if req.GetItem().GetType() != "" && req.GetItem().GetType() != static.Subtype.String() {
 				badFields[globals.TypeField] = "Cannot modify the resource type."
+			}
+			if handlers.MaskContains(req.GetUpdateMask().GetPaths(), "attributes.address") {
 				attrs := req.GetItem().GetStaticHostAttributes()
-
-				if handlers.MaskContains(req.GetUpdateMask().GetPaths(), "attributes.address") {
-					switch {
-					case attrs == nil:
-						badFields["attributes"] = "Attributes field not supplied request"
-					default:
-						if attrs.GetAddress() == nil ||
-							len(strings.TrimSpace(attrs.GetAddress().GetValue())) < static.MinHostAddressLength ||
-							len(strings.TrimSpace(attrs.GetAddress().GetValue())) > static.MaxHostAddressLength {
-							badFields["attributes.address"] = fmt.Sprintf("Address length must be between %d and %d characters.", static.MinHostAddressLength, static.MaxHostAddressLength)
+				switch {
+				case attrs == nil:
+					badFields["attributes"] = "Attributes field not supplied request"
+				default:
+					if attrs.GetAddress() == nil ||
+						len(strings.TrimSpace(attrs.GetAddress().GetValue())) < static.MinHostAddressLength ||
+						len(strings.TrimSpace(attrs.GetAddress().GetValue())) > static.MaxHostAddressLength {
+						badFields["attributes.address"] = fmt.Sprintf("Address length must be between %d and %d characters.", static.MinHostAddressLength, static.MaxHostAddressLength)
+					} else {
+						_, _, err := net.SplitHostPort(attrs.GetAddress().GetValue())
+						switch {
+						case err == nil:
+							badFields["attributes.address"] = "Address for static hosts does not support a port."
+						case strings.Contains(err.Error(), "missing port in address"):
+							// Bare hostname, which we want
+						default:
+							badFields["attributes.address"] = fmt.Sprintf("Error parsing address: %v.", err)
 						}
 					}
 				}
--- a/internal/daemon/controller/handlers/hosts/host_service.go
+++ b/internal/daemon/controller/handlers/hosts/host_service.go
@@ -637,21 +637,21 @@ func validateCreateRequest(req *pbs.CreateHostRequest) error {
 			attrs := req.GetItem().GetStaticHostAttributes()
 			switch {
 			case attrs == nil:
-				badFields["attributes"] = "This is a required field."
+				badFields[globals.AttributesField] = "This is a required field."
 			default:
 				if attrs.GetAddress() == nil ||
 					len(attrs.GetAddress().GetValue()) < static.MinHostAddressLength ||
 					len(attrs.GetAddress().GetValue()) > static.MaxHostAddressLength {
-					badFields["attributes.address"] = fmt.Sprintf("Address length must be between %d and %d characters.", static.MinHostAddressLength, static.MaxHostAddressLength)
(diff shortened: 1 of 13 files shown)


internal/daemon/controller/handlers/hosts/host_service.go
```diff
@@ -668,17 +669,26 @@ func validateUpdateRequest(req *pbs.UpdateHostRequest) error {
...
-					case attrs == nil:
-						badFields["attributes"] = "Attributes field not supplied request"
-					default:
-						if attrs.GetAddress() == nil ||
-							len(strings.TrimSpace(attrs.GetAddress().GetValue())) < static.MinHostAddressLength ||
-							len(strings.TrimSpace(attrs.GetAddress().GetValue())) > static.MaxHostAddressLength {
-							badFields["attributes.address"] = fmt.Sprintf("Address length must be between %d and %d characters.", static.MinHostAddressLength, static.MaxHostAddressLength)
+				switch {
+				case attrs == nil:
+					badFields["attributes"] = "Attributes field not supplied request"
```
johanbrandhorst, line 677: Not your change, but this text reads a bit strange?
```suggestion
					badFields["attributes"] = "Attributes field not supplied in request"
```

internal/daemon/controller/handlers/hosts/host_service_test.go
```diff
@@ -1232,23 +1287,27 @@ func TestUpdate_Static(t *testing.T) {
...
-				assert.NotNilf(tc.res, "Expected UpdateHost response to be nil, but was %v", got)
-				gotUpdateTime := got.GetItem().GetUpdatedTime().AsTime()
-				// Verify it is a set updated after it was created
-				// TODO: This is currently failing.
-				assert.True(gotUpdateTime.After(hCreated), "Updated set should have been updated after it's creation. Was updated %v, which is after %v", gotUpdateTime, hCreated)
+			assert.NotNilf(tc.res, "Expected UpdateHost response to be nil, but was %v", got)
+			gotUpdateTime := got.GetItem().GetUpdatedTime().AsTime()
+			// Verify it is a set updated after it was created
+			// TODO: This is currently failing.
+			assert.True(gotUpdateTime.After(hCreated), "Updated set should have been updated after it's creation. Was updated %v, which is after %v", gotUpdateTime, hCreated)
```
johanbrandhorst, line 1306: Sorry, also not you.
```suggestion
			assert.True(gotUpdateTime.After(hCreated), "Updated set should have been updated after its creation. Was updated %v, which is after %v", gotUpdateTime, hCreated)
```

internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -745,87 +749,103 @@ func (s Service) AuthorizeSession(ctx context.Context, req *pbs.AuthorizeSession
...
+	}
+
+	// Ensure we don't have a port from the address, which would be unexpected
+	// FIXME: We've decided to hold off on making this an error until 0.14. In
+	// the meantime, ignore any port coming from the host address.
+	hostWithoutPort, _, err := net.SplitHostPort(h)
+	switch {
+	case err != nil && strings.Contains(err.Error(), missingPortErrStr):
+		// This is what we expect
+		_ = hostWithoutPort
```
johanbrandhorst, line 838: What is the meaning of this assignment? I think we can remove it

internal/daemon/controller/handlers/hosts/host_service.go
```diff
@@ -643,15 +643,16 @@ func validateCreateRequest(req *pbs.CreateHostRequest) error {
...
-					badFields["attributes.address"] = "Address for static hosts does not support a port."
-				case strings.Contains(err.Error(), "missing port in address"):
-					// Bare hostname, which we want
-				default:
-					badFields["attributes.address"] = fmt.Sprintf("Error parsing address: %v.", err)
+				} else {
+					_, _, err := net.SplitHostPort(attrs.GetAddress().GetValue())
+					switch {
+					case err == nil:
+						badFields["attributes.address"] = "Address for static hosts does not support a port."
```
jimlambrt, line 650: perhaps we can define a const for `attributes.address` since it's referenced 5 times in this small change.  Might be a good idea to define one from `attributes` as well, but that's a bigger change.

internal/daemon/controller/handlers/hosts/host_service.go
```diff
@@ -643,15 +643,16 @@ func validateCreateRequest(req *pbs.CreateHostRequest) error {
...
+					case strings.Contains(err.Error(), "missing port in address"):
```
jimlambrt, line 651: perhaps we can define a const for this important error msg as well.  It's used like 8 times on our code base.  Even a helper func that returns if it's actual this error like: `func missingPortError(err) bool`.  It's just an idea.

internal/daemon/controller/handlers/hosts/host_service_test.go
```diff
@@ -868,22 +888,21 @@ func TestUpdate_Static(t *testing.T) {
...
 	cases := []struct {
-		name string
-		req  *pbs.UpdateHostRequest
-		res  *pbs.UpdateHostResponse
-		err  error
+		name            string
+		req             *pbs.UpdateHostRequest
+		res             *pbs.UpdateHostResponse
+		err             error
+		wantErrContains string
```
jimlambrt, line 900: ty!

internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -745,87 +749,103 @@ func (s Service) AuthorizeSession(ctx context.Context, req *pbs.AuthorizeSession
...
+	case err != nil:
+		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("error when parsing the chosen endpoint host address"))
+	case err == nil:
+		h = hostWithoutPort
+		// Use below logic for 0.14:
+		/*
+			return nil, handlers.ApiErrorWithCodeAndMessage(
+				codes.FailedPrecondition,
+				"Address specified for use unexpectedly contains a port.")
+		*/
```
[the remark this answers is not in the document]
jimlambrt: @talanknight -> ty for mentioning this.  I had not seen these feature capabilities yet.

jimlambrt (changes requested):
Looks great.  Just a few minor suggestions.

internal/daemon/controller/handlers/hosts/host_service.go
johanbrandhorst: The target package has this constant, I don't know if we need to define it more centrally than that, but it could be a constant in this package too.

internal/daemon/controller/handlers/hosts/host_service_test.go
```diff
@@ -1232,23 +1287,27 @@ func TestUpdate_Static(t *testing.T) {
...
+			require.NoError(gErr)
+			require.NotNil(got)
 
-			if got != nil {
-				assert.NotNilf(tc.res, "Expected UpdateHost response to be nil, but was %v", got)
-				gotUpdateTime := got.GetItem().GetUpdatedTime().AsTime()
-				// Verify it is a set updated after it was created
-				// TODO: This is currently failing.
-				assert.True(gotUpdateTime.After(hCreated), "Updated set should have been updated after it's creation. Was updated %v, which is after %v", gotUpdateTime, hCreated)
+			assert.NotNilf(tc.res, "Expected UpdateHost response to be nil, but was %v", got)
```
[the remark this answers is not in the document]
jefferai: ✅

internal/daemon/controller/handlers/hosts/host_service_test.go
```diff
@@ -1232,23 +1287,27 @@ func TestUpdate_Static(t *testing.T) {
...
+			gotUpdateTime := got.GetItem().GetUpdatedTime().AsTime()
+			// Verify it is a set updated after it was created
+			// TODO: This is currently failing.
```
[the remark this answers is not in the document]
jefferai: ✅

internal/daemon/controller/handlers/targets/target_service.go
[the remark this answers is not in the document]
jefferai: @talanknight I like the idea but don't want to spend the time ramping up on how to do that properly given the feature freeze and time constraints. I'd be happy if you did a follow-up PR to add that...

internal/daemon/controller/handlers/hosts/host_service.go
jefferai: ✅

jefferai: I changed "attributes" in a few places (we already had a globals), but a future PR can just sweep through and update.

We may also want to consider moving things out of globals into a fields subpackge, so instead of e.g. "globals.AttributesField" we can just have "fields.Attributes". Maybe.

internal/daemon/controller/handlers/hosts/host_service.go
jefferai: ✅

internal/daemon/controller/handlers/targets/target_service.go
johanbrandhorst: Did you mean to leave this around?

johanbrandhorst (approved):

jimlambrt (approved):

tmessi (approved):

internal/daemon/controller/handlers/targets/target_service.go
jefferai: Nope, thanks!

PR #2856, opened 2023-01-29
feature (ldap/cli) add ldap accounts and managed groups CRUDL commands

jimlambrt:


internal/cmd/commands/accountscmd/ldap_funcs.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (c *LdapCommand) extraLdapHelpFunc(helpMap map[string]func() string) string {
+	var helpStr string
+	switch c.Func {
+	case "create":
+		helpStr = base.WrapForHelpText([]string{
+			"Usage: boundary accounts create ldap [options] [args]",
+			"",
+			"  Create a ldap-type account. Example:",
```
johanbrandhorst, line 35: Nit:
```suggestion
			"  Create an ldap-type account. Example:",
```

internal/cmd/commands/accountscmd/ldap_funcs.go
```diff
@@ -0,0 +1,86 @@
...
+			`    $ boundary accounts create ldap -login-name prodops -description "ldap account for ProdOps"`,
+			"",
+			"",
+		})
+
+	case "update":
+		helpStr = base.WrapForHelpText([]string{
+			"Usage: boundary accounts update ldap [options] [args]",
+			"",
+			"  Update a ldap-type account given its ID. Example:",
```
johanbrandhorst, line 46:
```suggestion
			"  Update an ldap-type account given its ID. Example:",
```

internal/cmd/commands/managedgroupscmd/ldap_funcs.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func (c *LdapCommand) extraLdapHelpFunc(helpMap map[string]func() string) string {
+	var helpStr string
+	switch c.Func {
+	case "create":
+		helpStr = base.WrapForHelpText([]string{
+			"Usage: boundary managed-groups create ldap [options] [args]",
+			"",
+			"  Create a ldap-type managed group. Example:",
```
johanbrandhorst, line 38:
```suggestion
			"  Create an ldap-type managed group. Example:",
```

internal/cmd/commands/managedgroupscmd/ldap_funcs.go
```diff
@@ -0,0 +1,83 @@
...
+			`    $ boundary managed-groups create ldap -group-names admin -description "Ldap managed group for ProdOps"`,
+			"",
+			"",
+		})
+
+	case "update":
+		helpStr = base.WrapForHelpText([]string{
+			"Usage: boundary managed-groups update ldap [options] [args]",
+			"",
+			"  Update a ldap-type managed group given its ID. Example:",
```
johanbrandhorst, line 49:
```suggestion
			"  Update an ldap-type managed group given its ID. Example:",
```

internal/cmd/commands/managedgroupscmd/ldap_funcs.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+	}
+}
+
+func extraLdapFlagsHandlingFuncImpl(c *LdapCommand, _ *base.FlagSets, opts *[]managedgroups.Option) bool {
+	switch {
+	case len(c.flagGroupNames) == 0:
+	case len(c.flagGroupNames) == 1 && c.flagGroupNames[0] == "null":
+		c.UI.Error(fmt.Sprintf("There must be at least one %q", groupNamesFlagName))
+		return false
```
johanbrandhorst, line 77: Two questions:
1. If the error is true, why is `len(c.flagGroupNames) == 0` a success case?
2. A bit ridicolous perhaps, but what's stopping someone from specifying 2 null valued groups? If we don't bother checking for that, why are we checking for the single case?

internal/cmd/commands/accountscmd/ldap_funcs.go
jimlambrt: ty!

internal/cmd/commands/accountscmd/ldap_funcs.go
jimlambrt: ty!

internal/cmd/commands/managedgroupscmd/ldap_funcs.go
jimlambrt: ty!

internal/cmd/commands/managedgroupscmd/ldap_funcs.go
jimlambrt: ty!

internal/cmd/commands/managedgroupscmd/ldap_funcs.go
jimlambrt: I'm confused by the "error is true" portion of question #1.    If there's no group names, then it's a no-op.

Our CLI is consistent across several use-cases that in order to set a multi-value flag to `null` there for likely deleting all of current values for the flag, the user just needs to set the first value for the flag to `null`

johanbrandhorst: I see, that makes sense! Apologies for the confusion, my comment was regarding the `There must be at least one ...`, I was saying, "if that message is true, why is it ok to add 0?". Are you saying that it shouldn't be possible to unset group names once one has been set (which is the case the error is handling)?

jimlambrt: Ah... yes, we should return and error here.  Ty!

jimlambrt: Wait... you need to be able to update Name and Description and leave the existing group names as is.

johanbrandhorst: Sounds good to me

johanbrandhorst (approved):

irenarindos (approved):

PR #2866, opened 2023-01-31
Add job run cleaner

johanbrandhorst:
HCP Boundary reports that the job_runs table is filling up with data, and this is something that will be affecting customers too. Create a cleaner job that cleans up old job runs.

### DIFF
--- a/internal/daemon/controller/controller.go
+++ b/internal/daemon/controller/controller.go
@@ -29,6 +29,7 @@ import (
 	"github.com/hashicorp/boundary/internal/observability/event"
 	"github.com/hashicorp/boundary/internal/plugin/host"
 	"github.com/hashicorp/boundary/internal/scheduler"
+	"github.com/hashicorp/boundary/internal/scheduler/cleaner"
 	"github.com/hashicorp/boundary/internal/scheduler/job"
 	"github.com/hashicorp/boundary/internal/server"
 	serversjob "github.com/hashicorp/boundary/internal/server/job"
@@ -513,6 +514,9 @@ func (c *Controller) registerJobs() error {
 		return err
 	}
 	if err := kmsjob.RegisterJobs(c.baseContext, c.scheduler, c.kms); err != nil {
+		return err
+	}
+	if err := cleaner.RegisterJob(c.baseContext, c.scheduler, rw); err != nil {
 		return err
 	}
 
--- a/internal/daemon/controller/controller.go
+++ b/internal/daemon/controller/controller.go
@@ -28,7 +28,6 @@ import (
 	kmsjob "github.com/hashicorp/boundary/internal/kms/job"
 	"github.com/hashicorp/boundary/internal/observability/event"
 	"github.com/hashicorp/boundary/internal/plugin/host"
-	hostplugin "github.com/hashicorp/boundary/internal/plugin/host"
 	"github.com/hashicorp/boundary/internal/scheduler"
 	"github.com/hashicorp/boundary/internal/scheduler/job"
 	"github.com/hashicorp/boundary/internal/server"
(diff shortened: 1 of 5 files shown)


internal/daemon/controller/controller.go
```diff
@@ -28,8 +28,8 @@ import (
 	kmsjob "github.com/hashicorp/boundary/internal/kms/job"
 	"github.com/hashicorp/boundary/internal/observability/event"
 	"github.com/hashicorp/boundary/internal/plugin/host"
-	hostplugin "github.com/hashicorp/boundary/internal/plugin/host"
```
louisruch, line 31: Nice catch

internal/db/schema/migrations/oss/postgres/62/01_job_run_index.up.sql
```diff
@@ -0,0 +1,7 @@
+begin;
+
+  delete from job_run where status='completed';
+
+  create index on job_run (status);
```
louisruch, line 5: Most of the index's we create, apart from 1 that I found create a named index. https://github.com/hashicorp/boundary/blob/main/internal/db/schema/migrations/oss/postgres/50/01_session_list_index.up.sql Often we also add a comment on the index https://github.com/hashicorp/boundary/blob/main/internal/db/schema/migrations/oss/postgres/10/04_vault_credential.up.sql#L182

internal/scheduler/cleaner/cleaner_job.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func newCleanerJob(w db.Writer) *cleanerJob {
+	return &cleanerJob{
+		w: w,
+	}
+}
+
+// Status reports the job’s current status. We never change these values as
+// this job never finishes.
```
louisruch, line 23: Not sure I follow the `this job never finishes` comment. Though I am fine with not reporting status here

internal/scheduler/cleaner/cleaner.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"context"
+
+	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/scheduler"
+)
+
+// RegisterJob registers the cleaner job with the provided scheduler.
+func RegisterJob(ctx context.Context, s *scheduler.Scheduler, w db.Writer) error {
```
louisruch, line 12: This func is never tested

internal/scheduler/cleaner/cleaner.go
```diff
@@ -0,0 +1,23 @@
...
+	const op = "cleaner.RegisterJob"
+	if s == nil {
```
louisruch, line 14: Might want to nil check the writer

louisruch (commented):
Thanks for getting onto this!

internal/db/schema/migrations/oss/postgres/62/01_job_run_index.up.sql
johanbrandhorst: Good shout, I added a comment and a name.

internal/scheduler/cleaner/cleaner_job.go
johanbrandhorst: I removed this part of the comment, it was just confusing.

internal/scheduler/cleaner/cleaner.go
johanbrandhorst: I added a test for this, thanks!

internal/scheduler/cleaner/cleaner.go
johanbrandhorst: Good idea, done!

louisruch (approved):
:shipit:

PR #2871, opened 2023-01-31
docs: Update credentials docs for 0.12.0

Dan-Heath:
This pull request contains updates for 2 enhancements that affected the Boundary credentials documentation:

- JSON credentials were added in 0.11.0, but were never documented. This PR adds them to the **Credentials** page. View the updates in the preview deployment:
    - [Credentials](https://boundary-83oe0erah-hashicorp.vercel.app/boundary/docs/concepts/domain-model/credentials)

- Vault SSH certificate injection is being introduced in 0.12.0. This PR adds information about the new credential type and credential library. View the updates in the preview deployment:
    - [Credentials](https://boundary-83oe0erah-hashicorp.vercel.app/boundary/docs/concepts/domain-model/credentials)
    - [Credential Libraries](https://boundary-83oe0erah-hashicorp.vercel.app/boundary/docs/concepts/domain-model/credential-libraries)


website/content/docs/concepts/domain-model/credential-libraries.mdx
```diff
@@ -7,39 +7,70 @@ description: |-
...
+- `key_bits` - The number of bits used to generate the SSH private key.
+The number of bits depends on the `key_type` value you select:
+  - For an `ed25519` key type, you should not set the `key_bits` value.
+  - For an `ecdsa` key type, you can select either `256`, `384`, or `521`.
+  - For an `rsa` key type, you can select either `2048`, `3072`, or `4096`.
+
+- `TTL` - The SSH certificate's time-to-live (TTL).
+Note that the certificate is issued for the entire session, so if the `TTL` value is shorter than the target's `session_max_seconds` value, later connections may fail.
+To prevent failures, you should ensure that the `TTL` value is equal to or longer than the target's `session_max_seconds`.
+Alternatively, you could set the `session_connection_limit` to `1` for any targets that use the credential library.
```
Dan-Heath, line 63: I am not sure if this caveat is better in context with the `TTL` field, or if it would be better in the introduction section _before_ the bulleted list.

ddebko (approved):
JSON credentials docs looks good to me.

website/content/docs/concepts/domain-model/credential-libraries.mdx
```diff
@@ -7,39 +7,70 @@ description: |-
 
 # Credential Libraries
 
-A credential library is a resource
-that provides [credentials][]
-of the same type and same access level
-from a single [credential store][].
+A credential library is a resource that provides [credentials][] of the same type and same access level from a single [credential store][].
```
tmessi, line 10: Can we keep the [semantic line breaks](https://sembr.org/)? It makes reading future diffs a lot easier.

website/content/docs/concepts/domain-model/credential-libraries.mdx
```diff
@@ -7,39 +7,70 @@ description: |-
...
-A credential library has the following configurable attributes:
+A credential library has the following optional configurable attributes:
 
-- `name` - (optional)
-  If set, the `name` must be unique within the credential library's parent
-  [credential store][].
+- `name` - If you set this attribute, the `name` must be unique within the credential library's parent [credential store][].
 
-- `description` - (optional)
+- `description` - A user-defined description of the credential library for identification purposes.
```
tmessi, line 18: Looking at other pages in the domain model section, we are being explicit with the `(optional)` after each optional field. I think it would be good to keep that here to be consistent with the other docs.

website/content/docs/concepts/domain-model/credential-libraries.mdx
```diff
@@ -7,39 +7,70 @@ description: |-
...
-  The body of the HTTP request the library sends to Vault when requesting credentials.
-  Only valid if `http_method` is set to `POST`.
+- `http_request_body` - The body of the HTTP request the library sends to Vault when requesting credentials.
+Only valid if `http_method` is set to `POST`.
 
-### Vault Credential Library Parameter Templating
+### Vault SSH certificate credential library attributes
 
-In some cases it can be useful to provide information about a Boundary user or account when making a call to Vault. For example, this can allow picking the correct role when asking for database credentials (if roles are separated per-user), or providing a value to encode in an X.509 certificate generated by Vault. As of Boundary 0.11.1, you can template user and account information into either the path in Vault, the `POST` request body, or both.
+As of Boundary 0.12.0, you can configure SSH credential injection using Vault.
```
tmessi, line 35: I think this need to be rephrased slightly. Earlier versions support SSH credential injection using vault, but only via the generic credential library and only using `username_password` or `ssh_private_key` credential types.

The new thing in 0.12.0 is a new credential library that enables SSH credential injection using Vault's SSH secret engine to create SSH Certificate credentials.

website/content/docs/concepts/domain-model/credential-libraries.mdx
tmessi: I think either before or after the bulleted list as a "note". Looking at some other pages I think it would be something like:

```
~> **Note:** The certificate is issued for the entire session, so if the `TTL` value is shorter than the target's `session_max_seconds` value, later connections may fail.
To prevent failures, you should ensure that the `TTL` value is equal to or longer than the target's `session_max_seconds`.
Alternatively, you could set the `session_connection_limit` to `1` for any targets that use the credential library.
```

website/content/docs/concepts/domain-model/credential-libraries.mdx
```diff
@@ -7,39 +7,70 @@ description: |-
...
+- `TTL` - The SSH certificate's time-to-live (TTL).
+Note that the certificate is issued for the entire session, so if the `TTL` value is shorter than the target's `session_max_seconds` value, later connections may fail.
+To prevent failures, you should ensure that the `TTL` value is equal to or longer than the target's `session_max_seconds`.
+Alternatively, you could set the `session_connection_limit` to `1` for any targets that use the credential library.
+
+- `key_id` - The key ID for the created SSH certificate.
+
+- `critical_options` - Any critical options that the certificate should be signed for.
+
+- `extensions` - Any extensions that the certificate should be signed for.
```
tmessi, line 69: It might be useful to have another note here linking to some possible critical options or extensions, ie the ones openssh supports: https://github.com/openssh/openssh-portable/blob/5f93c4836527d9fda05de8944a1c7b4a205080c7/PROTOCOL.certkeys#L221-L319

In addition I think we should have a note for `extensions` noting that `permit-pty` should be set for an interactive shell to function properly, since this is probably the most common use case.

website/content/docs/concepts/domain-model/credential-libraries.mdx
```diff
@@ -7,39 +7,70 @@ description: |-
...
+The key type is either `ed25519`, `ecdsa`, or `rsa`.
+The default key type is `ed25519`.
+
+- `key_bits` - The number of bits used to generate the SSH private key.
+The number of bits depends on the `key_type` value you select:
+  - For an `ed25519` key type, you should not set the `key_bits` value.
+  - For an `ecdsa` key type, you can select either `256`, `384`, or `521`.
+  - For an `rsa` key type, you can select either `2048`, `3072`, or `4096`.
+
+- `TTL` - The SSH certificate's time-to-live (TTL).
```
tmessi, line 60:
```suggestion
- `ttl` - The SSH certificate's time-to-live (TTL).
```

website/content/docs/concepts/domain-model/credential-libraries.mdx
```diff
@@ -7,39 +7,70 @@ description: |-
...
+Unlike SSH key pairs, SSH certificates are short-lived and self-destructive.
+
+A Vault SSH certificate credential library has the following additional attributes:
+
+- `path` - (required) The path in Vault to request credentials from.
+
+- `username` - (required) The username to use with the SSH certificate.
+You can create a template for this value using [Vault credential library parameter templating](#vault-credential-library-parameter-templating).
+
+- `key_type` - The type of key to use for the generated SSH private key.
```
tmessi, line 50: I think to be consistent with other pages, each of these optional attributes should have `(optional)` before the description.

tmessi (changes requested):

website/content/docs/concepts/domain-model/credential-libraries.mdx
Dan-Heath: Sorry about that! Will do in the future.

website/content/docs/concepts/domain-model/credential-libraries.mdx
Dan-Heath: I added these back in for all the optional values.

website/content/docs/concepts/domain-model/credential-libraries.mdx
Dan-Heath: This is great info, thanks! I added links underneath each of the respective options. The Tech Docs team is really trying to limit the number of **Notes** we use in topics, b/c studies show they derail users reading patterns. So I added these as regular sentences in the bullets. I think it's legible like that, but let me know if you disagree.

website/content/docs/concepts/domain-model/credential-libraries.mdx
Dan-Heath: Good catch! Thanks.

website/content/docs/concepts/domain-model/credential-libraries.mdx
```diff
@@ -7,39 +7,78 @@ description: |-
...
+The key type is either `ed25519`, `ecdsa`, or `rsa`.
+The default key type is `ed25519`.
+
+- `key_bits` - (optional) The number of bits used to generate the SSH private key.
+The number of bits depends on the `key_type` value you select:
+  - For an `ed25519` key type, you should not set the `key_bits` value.
+  - For an `ecdsa` key type, you can select either `256`, `384`, or `521`.
+  - For an `rsa` key type, you can select either `2048`, `3072`, or `4096`.
+
+- `TTL` - (optional) The SSH certificate's time-to-live (TTL).
```
tmessi, line 68: This got capitialized again. I think when it is monspaced/inline-code it should be all lowercase, so both here and in the note above.

website/content/docs/concepts/domain-model/credential-libraries.mdx
tmessi: Yea I think this is fine as a regular sentence in those bullets.

website/content/docs/concepts/domain-model/credential-libraries.mdx
Dan-Heath: Oops, thank you for catching that. That makes sense, I updated it in both places.

tmessi (approved):

PR #2873, opened 2023-01-31
fix(vault): Only persist revokable credentials

tmessi:
When issuing credentials from vault credential libraries, only persist
to the `credential_vault_credential` table if it is revokable. A
credential that is revokable has a valid lease id that can be used to
revoke the lease in vault. If the credential is not revokable it is 
also not renewable since renewing also requires a lease id.

Note that previously we would persist to this table with a sentinel
value for the external id (aka the lease id), and a status of
"unknown". This meant that row was (correctly) ignored by the renewal
and revoke job.

See: https://developer.hashicorp.com/vault/docs/concepts/lease

### DIFF
--- a/internal/credential/vault/repository_credentials.go
+++ b/internal/credential/vault/repository_credentials.go
@@ -75,6 +75,12 @@ func (r *Repository) Issue(ctx context.Context, sessionId string, requests []cre
 			return nil, err
 		}
 
+		creds = append(creds, cred)
+		if !cred.isRevokable() {
+			// No need to persist since the credential cannot be revoked nor renewed
+			continue
+		}
+
 		if cred.getExpiration() < runJobsInterval {
 			event.WriteError(ctx, op,
 				fmt.Errorf("WARNING: credential will expire before job scheduler can run"),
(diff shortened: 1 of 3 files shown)


internal/credential/vault/repository_credentials.go
```diff
@@ -75,53 +75,55 @@ func (r *Repository) Issue(ctx context.Context, sessionId string, requests []cre
...
 
-		if cred.getExpiration() < runJobsInterval {
-			event.WriteError(ctx, op,
-				fmt.Errorf("WARNING: credential will expire before job scheduler can run"),
-				event.WithInfo("credential_public_id", cred.GetPublicId()),
-				event.WithInfo("credential_library_public_id", lib.GetPublicId()),
-				event.WithInfo("runJobsInterval", runJobsInterval.String()),
-			)
-		}
+		if !cred.isRevokable() {
```
louisruch, line 78: This should be `if cred.isRevokable() {`

Also minor formatting comment, you could remove a bunch of this nested if if you do something like:
```
if !cred.isRevokable() {
  // Amazing comment explaining that we don't need to persist the cred so we can continue early
  creds = append(creds, cred)
  continue
```

tmessi: Yea, I went back and forth between that, which is how I ended up with the incorrect check here :facepalm:

louisruch (approved):
LGTM

PR #2879, opened 2023-02-01
chore: Update deprecated set-output command

moduli:
This PR updates some GitHub action workflows because of an upcoming deprecation of the `set-output` command. We were seeing warnings here: https://github.com/hashicorp/boundary/actions/runs/4060061351
> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

A number of warnings have been addressed, going from 89 to 9. The remaining warnings will require some updates to `hashicorp/actions-docker-build` ~and `hashicorp/action-setup-enos`~.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,7 @@ jobs:
         # version, because "goenv" can react to it automatically.
         run: |
           echo "Building with Go $(cat .go-version)"
-          echo "::set-output name=go-version::$(cat .go-version)"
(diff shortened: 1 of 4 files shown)


modrake (approved):
Thanks for taking care of this; to the best of my knowledge this mimics the pattern that we've used in similar migrations.

PR #2887, opened 2023-02-01
fix(cli): Status code on api error

tmessi:
When using the json output of the cli (`-format=json`), a successful call
to the controller API would place the HTTP status code in the JSON as
`status_code`. However, if there was an error from the API the status code
would be placed in the JSON as `status`. In order to be consistent, this
adds a `status_code` field to the error response case. For backwards
compatibility it still populates `status` in the error case.

This inconsistency was missed partially due to a bug in the CLI tests. The
assertion on status code was not valid and would always be successful. Thus
this also fixes the assertions and tests.

### DIFF
--- a/internal/tests/cli/boundary/_helpers.bash
+++ b/internal/tests/cli/boundary/_helpers.bash
@@ -31,3 +31,11 @@ function has_status_code() {
 diag() {
     echo "$@" | sed -e 's/^/# /' >&3 ;
 }
+
+function field_eq() {
+    local json=$1
+    local field=$2
+    local expected=$3
+    echo "checking $field == $expected in $json"
+    echo "$json" | jq -e "$field == $expected"
+}
--- a/internal/tests/cli/boundary/_helpers.bash
+++ b/internal/tests/cli/boundary/_helpers.bash
@@ -24,9 +24,8 @@ function strip_all() {
 function has_status_code() {
   local json=$1
   local code=$2
-  if [ echo "$json"|jq -c ".status_code == $code" ]; then
-    return 1
-  fi
+  echo "checking .status_code == $code in $json"
+  echo "$json" | jq -e ".status_code == $code"
 }
 
 diag() {
--- a/internal/tests/cli/boundary/credential_libraries.bats
+++ b/internal/tests/cli/boundary/credential_libraries.bats
@@ -162,6 +162,417 @@ export NEW_VAULT_LIB="test_vault"
   [ "$status" -eq 0 ]
 }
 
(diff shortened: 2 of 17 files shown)


internal/tests/cli/boundary/_helpers.bash
```diff
@@ -24,11 +24,18 @@ function strip_all() {
 function has_status_code() {
   local json=$1
   local code=$2
-  if [ echo "$json"|jq -c ".status_code == $code" ]; then
-    return 1
-  fi
+  echo "checking .status_code == $code in $json"
+  echo "$json" | jq -e ".status_code == $code"
```
johanbrandhorst, line 28: Wouldn't it be more correct to use `.status` here?

johanbrandhorst (commented):
Shouldn't we just bite the bullet and remove `status_code`? Or at least deprecate it so we can remove it in a later release and move usage to `status`?

EDIT: I mean the other way around, deprecating the old one.

internal/tests/cli/boundary/_helpers.bash
tmessi: This is moving the other way, from `status` to `status_code`

johanbrandhorst: Doh, thanks

internal/tests/cli/boundary/credential_libraries.bats
```diff
@@ -155,6 +162,417 @@ export NEW_VAULT_LIB="test_vault"
...
+  echo "$output"
+  [ "$status" -eq 0 ]
+
+  run read_credential_library $clid
+  echo "$output"
+  [ "$status" -eq 0 ]
+  got=$(echo "$output")
+
+  run field_eq "$got" ".item.attributes.key_type" '"ecdsa"'
+  [ "$status" -eq 0 ]
```
moduli, line 180: Do you want an `echo "$output"` here to get debug info from `field_eq`?

tmessi: I am not sure how useful it is. The message from `jq` isn't the greatest. The output from the `read_credential_library` should already show the full json that is being used for the comparison.

internal/cmd/commands/credentiallibrariescmd/funcs.go
```diff
@@ -140,6 +140,16 @@ func printItemTable(item *credentiallibraries.CredentialLibrary, resp *api.Respo
 		nonAttributeMap["Type"] = item.Type
 	}
 
+	var keySubstMap map[string]string
+	switch item.Type {
+	case "vault":
+		fallthrough
+	case "vault-generic":
```
johanbrandhorst: What do you think about just
```suggestion
	case "vault", "vault-generic":
```

PR #2890, opened 2023-02-02
feature (cli/ldap/authenticate): use primary auth method if none is provided

jimlambrt:
Also includes the overlooked ability to set the `state` of an ldap auth method, which is required for this qol cli feature.


irenarindos (approved):

internal/cmd/commands/authenticate/authenticate.go
```diff
@@ -109,6 +110,10 @@ func (c *Command) Run(args []string) int {
 		cmd := OidcCommand{Command: c.Command, Opts: []common.Option{common.WithSkipScopeIdFlag(true)}}
 		cmd.Run([]string{})
 
+	case strings.HasPrefix(c.FlagAuthMethodId, ldap.AuthMethodPrefix):
+		cmd := LdapCommand{Command: c.Command, Opts: []common.Option{common.WithSkipScopeIdFlag(true)}}
+		cmd.Run([]string{})
+
 	default:
 		c.PrintCliError(fmt.Errorf("The primary auth method was of an unsupported type. The given ID was %s; only 'ampw' (password) and 'amoidc' (OIDC) auth method prefixes are supported.", c.FlagAuthMethodId))
```
johanbrandhorst, line 118: This error will need updating now that we support LDAP too.

johanbrandhorst (changes requested):

johanbrandhorst (approved):

PR #2892, opened 2023-02-02
fix (handlers/authmethods): fix ldap authorized actions

jimlambrt:


johanbrandhorst (approved):
I'm not sure I understand the context of this change, is the user not allowed to change the state once it's been set?

jimlambrt:
@johanbrandhorst the user can change the state through a "normal" update... there is not special change-state API/CLI for LDAP since there is no "validation" step required like OIDC has.

johanbrandhorst (approved):
Thanks for the explanation, that makes sense

PR #2895, opened 2023-02-02
Update README.md

stevenzamborsky:
Fix image names in Docker run commands to include `hashicorp/` prefix.  Without this we see the following errors:

```sh
$ docker run \
  --network host \
  -e 'BOUNDARY_POSTGRES_URL=postgresql://postgres:postgres@0.0.0.0:5432/postgres?sslmode=disable' \
  boundary database init -config /boundary/config.hcl

Unable to find image 'boundary:latest' locally
docker: Error response from daemon: pull access denied for boundary, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.
```


moduli:
This will be addressed by https://github.com/hashicorp/boundary/pull/3405

PR #2903, opened 2023-02-03
fix(handlers): return error when updating json cred w/ empty obj

ddebko:
### Summary:

The API was returning a 200 when updating a json credential with an empty object secret. No changes were applied to the credential. The expected behavior is to return an error to the user that the object field cannot be empty. This bug only appears in the API, not the CLI.

### Steps to Reproduce Bug:

1. Launch `boundary dev`
2. Authenticate
```
curl \
  --header "Content-Type: application/json" \
  --request POST \
  --data '{"attributes":{"login_name": "admin","password": "password"},"type": "token"}' \
  http://localhost:9200/v1/auth-methods/ampw_1234567890:authenticate | export TOKEN=$(jq -r '.attributes.token')
```
3. Create a Static Credential Store
```
curl \
  --header "Content-Type: application/json" \
  --header "Authorization: Bearer $TOKEN" \
  --request POST \
  --data '{"scope_id":"p_1234567890","type": "static"}' \
  http://localhost:9200/v1/credential-stores | export STORE_ID=$(jq -r '.id')
```
4. Create a JSON Credential. Save the output of the resource id.
```
curl \
  --header "Content-Type: application/json" \
  --header "Authorization: Bearer $TOKEN" \
  --request POST \
  --data "{\"credential_store_id\":\"$STORE_ID\",\"type\":\"json\",\"attributes\":{\"object\":{\"test\":\"value\"}}}" \
  http://localhost:9200/v1/credentials | jq
```
5. Update JSON Credential w/ Empty Object
```
curl \
  --header "Content-Type: application/json" \
  --header "Authorization: Bearer $TOKEN" \
  --request PATCH \
  --data '{"credential_store_id":"csst_CNSHyoQx8u","version":1,"description":"test update","type":"json","attributes":{"object":{}}}' \
  http://localhost:9200/v1/credentials/credjson_tq7j9hiaYz
```

### Fix:

Expected return response in step 5:
```
{
  "kind": "InvalidArgument",
  "message": "Error in provided request.",
  "details": {
    "request_fields": [
      {
        "name": "attributes.object",
        "description": "This is a required field and cannot be set to empty."
      }
    ]
  }
}
```

### DIFF
--- a/internal/daemon/controller/handlers/credentials/credential_service.go
+++ b/internal/daemon/controller/handlers/credentials/credential_service.go
@@ -849,11 +849,14 @@ func validateUpdateRequest(req *pbs.UpdateCredentialRequest) error {
 				if err != nil {
 					badFields[objectField] = "Unable to parse given json value"
 				}
-				if !handlers.MaskContainsSubString(req.GetUpdateMask().GetPaths(), objectField) || len(object.AsMap()) <= 0 {
+				if !handlers.MaskContainsPrefix(req.GetUpdateMask().GetPaths(), objectField) {
+					badFields[objectField] = "This is a required field and cannot be set to empty."
+				}
+				if len(object.AsMap()) == 0 {
 					badFields[objectField] = "This is a required field and cannot be set to empty."
 				}
 			}
-			if handlers.MaskContainsSubString(req.GetUpdateMask().GetPaths(), objectField) && object == nil {
+			if handlers.MaskContainsPrefix(req.GetUpdateMask().GetPaths(), objectField) && object == nil {
 				badFields[objectField] = "This is a required field and cannot be set to empty."
 			}
 
--- a/internal/daemon/controller/handlers/credentials/credential_service.go
+++ b/internal/daemon/controller/handlers/credentials/credential_service.go
@@ -845,15 +845,15 @@ func validateUpdateRequest(req *pbs.UpdateCredentialRequest) error {
 		case credential.JsonSubtype:
 			object := req.GetItem().GetJsonAttributes().GetObject()
 			if object != nil {
-				objectBytes, err := json.Marshal(object.AsMap())
+				_, err := json.Marshal(object.AsMap())
 				if err != nil {
 					badFields[objectField] = "Unable to parse given json value"
 				}
-				if handlers.MaskContains(req.GetUpdateMask().GetPaths(), objectField) && len(objectBytes) <= 0 {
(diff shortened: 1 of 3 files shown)


internal/daemon/controller/handlers/mask_manager.go
```diff
@@ -117,3 +117,12 @@ func MaskContains(paths []string, s string) bool {
 	}
 	return false
 }
+
+func MaskContainsSubString(paths []string, s string) bool {
+	for _, p := range paths {
+		if strings.Contains(p, s) {
```
louisruch, line 123: Should be not be doing a `MaskContainsPrefix` and validate the field mask has the prefix of `attributes.object`

internal/daemon/controller/handlers/credentials/credential_service.go
```diff
@@ -845,15 +845,15 @@ func validateUpdateRequest(req *pbs.UpdateCredentialRequest) error {
 		case credential.JsonSubtype:
 			object := req.GetItem().GetJsonAttributes().GetObject()
 			if object != nil {
-				objectBytes, err := json.Marshal(object.AsMap())
+				_, err := json.Marshal(object.AsMap())
 				if err != nil {
 					badFields[objectField] = "Unable to parse given json value"
 				}
-				if handlers.MaskContains(req.GetUpdateMask().GetPaths(), objectField) && len(objectBytes) <= 0 {
+				if !handlers.MaskContainsSubString(req.GetUpdateMask().GetPaths(), objectField) || len(object.AsMap()) <= 0 {
```
louisruch, line 852: I wonder if it would make the code easier to maintain and provide a more clear error message to the user if we break this into two checks. The first checks if the mask was provided and the second checks if the len is == 0.

Also nit `len(object.AsMap()) == 0` instead of the `<=`

louisruch (commented):
Thanks for jumping onto this so quickly!

internal/daemon/controller/handlers/mask_manager.go
ddebko: I like the name `MaskContainsPrefix`. I will apply the change.

calcaide:
Tested passing an empty secret from the UI `{}` and now it returns an error.

calcaide (approved):
LGTM! thanks for the fix!

internal/daemon/controller/handlers/credentials/credential_service.go
ddebko: I agree, this will make it easier to read the conditional statement.

johanbrandhorst (approved):

PR #2806, opened 2023-01-19
docs: Update multi-hop docs

Dan-Heath:
This PR udpates the attribute descriptions for the ingress/egress filters with information about what happens when the optional settings are not configured.

You can view these updates in the preview deployment on the [Targets](https://boundary-oh1jyj5l6-hashicorp.vercel.app/boundary/docs/concepts/domain-model/targets) page.

Additionally, on the [PKI Worker Configuration page](https://boundary-oh1jyj5l6-hashicorp.vercel.app/boundary/docs/configuration/worker/pki-worker), I changed this link:

image

To go directly to the [Target Worker Filter](https://boundary-oh1jyj5l6-hashicorp.vercel.app/boundary/docs/concepts/filtering/worker-tags#target-worker-filtering) section, instead of to the top of the [Worker Tags](https://boundary-oh1jyj5l6-hashicorp.vercel.app/boundary/docs/concepts/filtering/worker-tags) page (same page, but deeper link).

PR #2772, opened 2023-01-10
Multihop workers doc

irenarindos:
To go on top of PR https://github.com/hashicorp/boundary/pull/2771

@Dan-Heath I've included an image here- feel free to clean it up for publication if you don't think it's good enough as is.

PR #2771, opened 2023-01-10
Update documentation to add ingress and egress filters

irenarindos:

PR #2906, opened 2023-02-03
docs: remove multihop from 0.12 branch

Dan-Heath:
The Boundary OSS binaries will be available to the public on Feb 8, but HCP features will not be available until the following week. We already merged several PRs (#2771 , #2772 , #2806) with multi-hop (HCP-only feature) documentation to the `0.12.x `release branch. That content needs to be deleted, so that it does not go live on the 8th. 

This PR removes the multi-hop documentation from the `0.12.x` release branch. We will have to create a new PR or revert this PR to add this content back to the `0.12.x` and `stable-website` branches, when the HCP 0.12.0 release goes live.


website/content/docs/concepts/domain-model/targets.mdx
```diff
@@ -51,63 +51,6 @@ TCP targets have the following additional attributes:
...
-  The default is -1.
-  The value must be greater than 0 or exactly -1.
-
-- `session_max_seconds` - (required)
-  The maximum duration of an individual session between the user and the target.
-  All connections for a session are closed
-  and the session is terminated
-  when a session reaches the maximum duration.
-  The default is 8 hours (28800 seconds).
-  This value must be greater than 0.
```
stellarsquall, line 80: Egress worker filters are in OSS as well, using `egress_worker_filter` as a replacement to the original `worker_filter`, not previously documented on this page. It should just be `ingress_worker_filter` that is HCP-only

```
- `egress_worker_filter` - (optional)
  A boolean expression to [filter][] which egress workers can handle sessions
  for this target. Replaces `worker_filter` used prior to Boundary 0.12.0.
  Egress worker filters determine which workers are used to access targets.
```

stellarsquall: this can include the original multi-hop content after next week

website/content/docs/concepts/filtering/worker-tags.mdx
```diff
@@ -140,13 +140,6 @@ Once workers have tags, you can use these tags to control which
 workers are allowed to manage a given session by specifying optional worker filter attributes
 when you configure [targets](/boundary/docs/concepts/domain-model/targets).
 
-The `egress_worker_filter` attribute controls which workers are used for egress to a target. This is the worker
-that accesses the target.
-
-The `ingress_worker_filter`<sup>HCP Only</sup> attribute controls which workers are used for ingress to a target.
-This is the worker a client connects to when initiating a connection to a target.
```
stellarsquall, line 147: I believe this should also keep the `egress_worker_filter` bits for the OSS release

```
The `egress_worker_filter` attribute controls which workers are used for egress to a target. This is the worker
that accesses the target.
```

website/content/docs/concepts/filtering/worker-tags.mdx
```diff
@@ -140,13 +140,6 @@ Once workers have tags, you can use these tags to control which
...
-
-
 # Vault Worker Filtering <sup>HCP Only</sup>
 Tags are used to control which [PKI workers] can manage Vault requests by specifying
 a `worker_filter`attribute when configuring [credential stores].
```
stellarsquall, line 145: check my math here, it's only the targets that have a new `egress_worker_filter` attribute, right? I think credential stores still use the `worker_filter` parameter. 

So this should be fine, I just want to make sure I didn't miss something here

stellarsquall (changes requested):

website/content/docs/concepts/filtering/worker-tags.mdx
Dan-Heath: Good question. Yes, that's correct as far as I understand too. This enhancement shouldn't have affected the credential stores parameter at all. I just rechecked the PRD/RFC and I don't see any reference to an impact to credential stores either.

Dan-Heath:
Thanks for the review @stellarsquall ! Good catch on the egress filter content. I added it all back.

website/content/docs/concepts/domain-model/targets.mdx
```diff
@@ -58,56 +58,6 @@ TCP targets have the following additional attributes:
   You can configure an egress filter to enable [multi-hop](/boundary/docs/configuration/worker/pki-worker#multi-hop-workershcp-only) connections.
   If you do not configure an egress filter, then Boundary uses a single worker to connect to the controller.
```
stellarsquall, line 59: this section contains references to multi-hop and HCP, do you want to leave it in?

Dan-Heath: Oops, definitely not. That link won't be good until the 14th. I removed both of these sentences. Thanks for catching this!

stellarsquall (approved):

PR #2909, opened 2023-02-03
test(e2e): Update migration test to use vault docker container

moduli:
This PR updates the e2e migration test to use a vault docker container as opposed to an instance of vault created by enos. This helps to reduce the amount of time this test takes to run as well as reduces cost by decreasing the number of items enos needs to create.

Some sample times from the GitHub Action that runs the enos database scenario
```
# Before Change
6m 11
6m 31
6m 37
7m 1

# After Change
5m 34
5m 51
5m 13
4m 52
```

### DIFF
--- a/enos/modules/test_e2e/main.tf
+++ b/enos/modules/test_e2e/main.tf
@@ -123,7 +123,7 @@ resource "enos_local_exec" "run_e2e_test" {
     E2E_AWS_HOST_SET_IPS2         = local.aws_host_set_ips2
   }
 
-  inline = ["PATH=\"${var.local_boundary_dir}:$PATH\" go test -v ${var.test_package} -count=1 -json | tparse -follow -format plain 2>&1 | tee ${path.module}/../../test-e2e-${local.package_name}.out"]
+  inline = ["set -o pipefail; PATH=\"${var.local_boundary_dir}:$PATH\" go test -v ${var.test_package} -count=1 -json | tparse -follow -format plain 2>&1 | tee ${path.module}/../../test-e2e-${local.package_name}.out"]
 }
 
 output "test_results" {
--- a/testing/internal/e2e/infra/docker.go
+++ b/testing/internal/e2e/infra/docker.go
@@ -97,7 +97,8 @@ func InitBoundaryDatabase(t testing.TB, pool *dockertest.Pool, network *dockerte
 }
 
 // GetDbInitInfoFromContainer extracts info from calling `boundary database init` in the specified
-// container. Returns a struct containing the generated info.
+// container.
+// Returns a struct containing the generated info.
 func GetDbInitInfoFromContainer(t testing.TB, pool *dockertest.Pool, container *Container) boundary.DbInitInfo {
 	_, err := pool.Client.WaitContainer(container.Resource.Container.ID)
 	require.NoError(t, err)
@@ -122,8 +123,8 @@ func GetDbInitInfoFromContainer(t testing.TB, pool *dockertest.Pool, container *
 }
 
 // StartBoundary starts a boundary container and spins up an instance of boundary using the
-// specified database at postgresURI. Returns information about the container.
-// Returns information about the container
+// specified database at postgresURI.
+// Returns information about the container.
 func StartBoundary(t testing.TB, pool *dockertest.Pool, network *dockertest.Network, postgresURI string) *Container {
 	t.Log("Starting Boundary...")
 
--- a/testing/internal/e2e/infra/docker.go
+++ b/testing/internal/e2e/infra/docker.go
@@ -161,6 +161,39 @@ func StartBoundary(t testing.TB, pool *dockertest.Pool, network *dockertest.Netw
 	}
 }
 
+// StartVault starts a vault container.
+// Returns information about the container.
+func StartVault(t testing.TB, pool *dockertest.Pool, network *dockertest.Network) *Container {
+	t.Log("Starting Vault...")
+
+	vaultToken := "boundarytok"
+	resource, err := pool.RunWithOptions(&dockertest.RunOptions{
+		Repository: "hashicorp/vault",
+		Tag:        "latest",
+		Env: []string{
+			"VAULT_DEV_ROOT_TOKEN_ID=" + vaultToken,
+		},
+		Name:         "vault",
+		Networks:     []*dockertest.Network{network},
+		ExposedPorts: []string{"8200"},
+		PortBindings: map[docker.Port][]docker.PortBinding{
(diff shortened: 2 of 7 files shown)


testing/internal/e2e/infra/docker.go
```diff
@@ -118,12 +119,16 @@ func GetDbInitInfoFromContainer(t testing.TB, pool *dockertest.Pool, container *
 	err = json.Unmarshal(buf.Bytes(), &dbInitInfo)
 	require.NoError(t, err, buf.String())
 
+	os.Setenv("E2E_PASSWORD_AUTH_METHOD_ID", dbInitInfo.AuthMethod.AuthMethodId)
+	os.Setenv("E2E_PASSWORD_ADMIN_LOGIN_NAME", dbInitInfo.AuthMethod.LoginName)
+	os.Setenv("E2E_PASSWORD_ADMIN_PASSWORD", dbInitInfo.AuthMethod.Password)
```
johanbrandhorst, line 124: I'm confused by this change. Why is this being set here? If we need to set these values, I would expect that to be done in the function that's calling whatever it is that needs these in the environment. Do we need to do this at all though?

testing/internal/e2e/infra/docker.go
```diff
@@ -153,13 +158,49 @@ func StartBoundary(t testing.TB, pool *dockertest.Pool, network *dockertest.Netw
 	})
 	require.NoError(t, err)
 
+	uriLocalhost := "http://localhost:9200"
+	os.Setenv("BOUNDARY_ADDR", uriLocalhost)
```
johanbrandhorst, line 162: Ditto

testing/internal/e2e/infra/docker.go
```diff
@@ -153,13 +158,49 @@ func StartBoundary(t testing.TB, pool *dockertest.Pool, network *dockertest.Netw
...
+		PortBindings: map[docker.Port][]docker.PortBinding{
+			"8200/tcp": {{HostIP: "localhost", HostPort: "8200/tcp"}},
+		},
+		CapAdd: []string{"IPC_LOCK"},
+	})
+	require.NoError(t, err)
+
+	uriLocalhost := "http://localhost:8200"
+	os.Setenv("VAULT_ADDR", uriLocalhost)
+	os.Setenv("VAULT_TOKEN", vaultToken)
```
johanbrandhorst, line 195: Ditto, and also, we probably want to return the token from this.

testing/internal/e2e/tests/database/migration_test.go
```diff
@@ -150,16 +175,12 @@ func setupEnvironment(t testing.TB, ctx context.Context, c *config) TestEnvironm
...
-func populateBoundaryDatabase(t testing.TB, ctx context.Context, c *config, te TestEnvironment) {
-	os.Setenv("BOUNDARY_ADDR", te.Boundary.UriLocalhost)
-	os.Setenv("E2E_PASSWORD_AUTH_METHOD_ID", te.DbInitInfo.AuthMethod.AuthMethodId)
-	os.Setenv("E2E_PASSWORD_ADMIN_LOGIN_NAME", te.DbInitInfo.AuthMethod.LoginName)
-	os.Setenv("E2E_PASSWORD_ADMIN_PASSWORD", te.DbInitInfo.AuthMethod.Password)
-
+func populateBoundaryDatabase(t testing.TB, ctx context.Context, c *config, e TestEnvironment) {
 	// Create resources for target. Uses the local CLI so that these methods can be reused.
 	// While the CLI version used won't necessarily match the controller version, it should be (and is
 	// supposed to be) backwards ompatible
```
johanbrandhorst, line 186: Can't we pass in this auth information to `AuthenticateAdminCli`?

testing/internal/e2e/tests/database/migration_test.go
```diff
@@ -150,16 +175,12 @@ func setupEnvironment(t testing.TB, ctx context.Context, c *config) TestEnvironm
...
 	}
 }
 
-func populateBoundaryDatabase(t testing.TB, ctx context.Context, c *config, te TestEnvironment) {
-	os.Setenv("BOUNDARY_ADDR", te.Boundary.UriLocalhost)
-	os.Setenv("E2E_PASSWORD_AUTH_METHOD_ID", te.DbInitInfo.AuthMethod.AuthMethodId)
-	os.Setenv("E2E_PASSWORD_ADMIN_LOGIN_NAME", te.DbInitInfo.AuthMethod.LoginName)
-	os.Setenv("E2E_PASSWORD_ADMIN_PASSWORD", te.DbInitInfo.AuthMethod.Password)
-
+func populateBoundaryDatabase(t testing.TB, ctx context.Context, c *config, e TestEnvironment) {
```
johanbrandhorst, line 183: Why did this variable name change? `te` makes sense to me.

moduli: I had a thought that `e` was simpler, but I'll revert the change.

testing/internal/e2e/infra/docker.go
moduli: I was originally thinking that creating the docker image, you may want to have these automatically set, but I was going back and forth on it. I'll move it back for clarity.

We do need to set some of these environment variables because some of the helper methods needs them (like enos setting up infrastructure and setting environment variables, these methods will need to do the same). I'll look to do some refactoring to see if we can reduce some of it.

enos/modules/test_e2e/main.tf
```diff
@@ -123,7 +123,7 @@ resource "enos_local_exec" "run_e2e_test" {
     E2E_AWS_HOST_SET_IPS2         = local.aws_host_set_ips2
   }
 
-  inline = ["PATH=\"${var.local_boundary_dir}:$PATH\" go test -v ${var.test_package} -count=1 -json | tparse -follow -format plain 2>&1 | tee ${path.module}/../../test-e2e-${local.package_name}.out"]
+  inline = ["set -o pipefail; PATH=\"${var.local_boundary_dir}:$PATH\" go test -v ${var.test_package} -count=1 -json | tparse -follow -format plain 2>&1 | tee ${path.module}/../../test-e2e-${local.package_name}.out"]
```
moduli, line 126: Added this fix to the PR. 
> After updating this command to save off the output from running `go test`, enos would not report when tests failed. This was due to the extra pipe hiding the exit status of the prior commands. Adding this option returns the status of the last command to exit with a non-zero status.

testing/internal/e2e/tests/database/migration_test.go
moduli: Did some refactoring. Let me know what you think

johanbrandhorst (approved):
Nice!

PR #2920, opened 2023-02-06
fix(handler): return targetId as hostId for cli edge case support

ddebko:
### Summary:

The recent feature of supporting associating a network address directly to a Target has caused a backwards compatibility issue with the older boundary binary cli. When authorizing a boundary session for a target tcp type and utilizing the ssh helper, the hostId value becomes a require value because it is used for `HostKeyAlias`

### Fix:

When authorizing a session, check if the target has a direct network address association and mock the hostId value with the targetId value. Using the targetId value as the mock value for hostId will guarantee a consistent HostKeyAlias value, which will prevent the ssh error: `Warning: Remote host identification has changed`.

### DIFF
--- a/internal/daemon/controller/handlers/targets/target_service.go
+++ b/internal/daemon/controller/handlers/targets/target_service.go
@@ -40,6 +40,7 @@ import (
 	"github.com/hashicorp/boundary/internal/types/subtypes"
 	"github.com/hashicorp/boundary/sdk/pbs/controller/api/resources/scopes"
 	pb "github.com/hashicorp/boundary/sdk/pbs/controller/api/resources/targets"
+	fm "github.com/hashicorp/boundary/version"
 	"github.com/hashicorp/go-bexpr"
 	"github.com/hashicorp/go-secure-stdlib/strutil"
 	"github.com/mr-tron/base58"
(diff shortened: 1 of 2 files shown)


internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -40,6 +40,7 @@ import (
 	"github.com/hashicorp/boundary/internal/types/subtypes"
 	"github.com/hashicorp/boundary/sdk/pbs/controller/api/resources/scopes"
 	pb "github.com/hashicorp/boundary/sdk/pbs/controller/api/resources/targets"
+	fm "github.com/hashicorp/boundary/version"
```
johanbrandhorst, line 43: Does this need to be aliased?

ddebko: I wanted to avoid an any confusion between the package name and when we declare the variable name `version` in this file. I used fm for feature manager

johanbrandhorst: Sounds good to me

version/feature_manager.go
```diff
@@ -28,6 +28,7 @@ const (
 	MultiHopSessionFeature
 	IncludeStatusInCli
 	CredentialLibraryVaultSubtype
+	ShhIntoTcpTargetAddress
```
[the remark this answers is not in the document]
ddebko: I like this idea a lot, because this is indeed an interesting edge case to understand with when reading the code.

johanbrandhorst (approved):
Pending Todd's review

version/feature_manager.go
```diff
@@ -65,6 +66,10 @@ func init() {
 		MetaInfo:    []Metadata{OSS, HCP},
 		Constraints: mustNewConstraints("< 0.14.0"),
 	}
+	featureMap[ShhIntoTcpTargetAddress] = MetadataConstraint{
```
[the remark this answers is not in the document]
ddebko: Good catch.

louisruch (approved):
LGTM

version/feature_manager.go
```diff
@@ -28,6 +28,7 @@ const (
 	MultiHopSessionFeature
 	IncludeStatusInCli
 	CredentialLibraryVaultSubtype
+	SshIntoTcpTargetAddress
```
tmessi, line 31: I find this feature flag name confusing. At first glance it makes it seem like support for sshing into tcp targets with address will be dropped. I think this should be something more like `UseTargetIdForHostId`.

jefferai: 👍

jefferai (approved):

PR #2921, opened 2023-02-06
add changelog update for vendor label for linux packaging

claire-labry:
This PR is just an update to the changelog to reflect a bug that was fixed in our [actions-packaging-linux](https://github.com/hashicorp/actions-packaging-linux) action. We received a report from a user stating that vendor labels were missing when downloading linux packages (for products that were released via CRT) We fixed this in [this PR](https://github.com/hashicorp/actions-packaging-linux/pull/14) and we added a vendor label and set the default to `HashiCorp` This will only affect future releases that go through our linux promotion pipeline -- historical releases will not have this vendor label.


johanbrandhorst (approved):
LGTM

johanbrandhorst:
😬 I'm not sure if we've copied the changelog somewhere, but it probably won't hurt to add this here anyway. May want to update https://github.com/hashicorp/boundary/releases/tag/v0.12.0 too?

PR #2925, opened 2023-02-07
[QT-400] Update ci boundary-enterprise hc-service-uri to include the schedule workflow trigger

mikebaum:
Since the `test_ci_cleanup` workflow will be running on a nightly basis via a schedule workflow trigger, the `hc-service-uri` needs to be updated to add the `event_type` `schedule` in order to ensure the service user is permitted to assume the Doormat role for scheduled workflow runs.


ryancragun (approved):
I don't recall whether or not the `schedule` event type requires further attributes or not. Given that, I think this looks good!

mikebaum:
> I don't recall whether or not the `schedule` event type requires further attributes or not. Given that, I think this looks good!

I checked with releng and the schedule trigger can only apply to the default branch. It is actually documented as such in the GitHub Actions docs.

PR #2944, opened 2023-02-09
[QT-400] Fix doormat hc-service-uri for ci bootstrap ci doormat role

mikebaum:
When using the doormat GHA it is required to set a tag named `hc-service-uri` on the role the doormat user is to assume. When the `ci-test-cleanup` scheduled workflow was added, we needed to update the tag to include the new event type. Unfortunately there is a 256 character limit on the size of an AWS tag, therefore we needed to create a new role to support the new workflow. With this PR, I modify the doormat role and policy to extract a second role and policy that is tailored to be used with the scheduled cleanup workflow.


jaymalasinha (approved):
lgtm

PR #2946, opened 2023-02-09
chore: Add slack notifications when e2e or bats tests fail

moduli:
This PR adds slack notifications when e2e or bats tests fail. This will be used for me to track failures and triage any issues. It is currently set to post in `#proj-boundary-tests`.

Screenshot 2023-02-08 at 9 44 51 PM

2 secrets were added to the repository to facilitate this. 
- `SLACK_BOUNDARY_TEST_BOT_CHANNEL_ID`
- `SLACK_BOUNDARY_TEST_BOT_TOKEN`

Here is an example run where tests failed and the slack jobs ran: https://github.com/hashicorp/boundary/actions/runs/4130485934

### DIFF
--- a/.github/workflows/enos-run.yml
+++ b/.github/workflows/enos-run.yml
@@ -32,6 +32,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       GITHUB_TOKEN: ${{ secrets.SERVICE_USER_GITHUB_TOKEN }}
+    outputs:
+      status: ${{ steps.enos-status.outputs.status }}
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -135,8 +137,6 @@ jobs:
           enos scenario exec --chdir ./enos ${{ matrix.filter }} --cmd "version"
       - name: Run Enos scenario
         id: run
-        # Continue once and retry
-        continue-on-error: true
         env:
           ENOS_VAR_aws_region: us-east-1
           ENOS_VAR_aws_ssh_keypair_name: enos-ci-ssh-key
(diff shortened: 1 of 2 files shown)


.github/workflows/enos-run.yml
```diff
@@ -155,20 +155,10 @@ jobs:
...
-          ENOS_VAR_aws_region: us-east-1
-          ENOS_VAR_aws_ssh_keypair_name: enos-ci-ssh-key
-          ENOS_VAR_aws_ssh_private_key_path: ./support/private_key.pem
-          ENOS_VAR_local_boundary_dir: ./support/
-          ENOS_VAR_crt_bundle_path: ./support/boundary.zip
-          ENOS_VAR_tfc_api_token: ${{ secrets.TF_API_TOKEN }}
-          ENOS_VAR_test_email: ${{ secrets.SERVICE_USER_EMAIL }}
-        run: |
-          export ENOS_VAR_enos_user=$GITHUB_ACTOR && \
-          enos scenario run --timeout 60m0s --chdir ./enos ${{ matrix.filter }}
```
tmessi, line 171: Did you mean to include this removal of the retry?

tmessi: > Most enos-side failures I've seen don't get resolved unless you do a destroy first

In cases where the retry is working, a failure won't be reported. So you might not see it. You could add another slack notification to notify when a retry is going to happen. Then you would have visability into how often a retry is happening, and how much it is "working".

moduli: I'll put the retry back in. I'll set it so the notifications send anytime the first attempt fails and can tell from there if a retry fixes it or not.

johanbrandhorst (approved):

tmessi (approved):

PR #2948, opened 2023-02-10
test(e2e/migration): Add additional resources to database before migrating

moduli:
This PR updates the e2e migration test to add some new resources to the database before performing the migration.
- Vault credential (type: username/password)
- User/Role/Group


johanbrandhorst (approved):
v nice

PR #2949, opened 2023-02-10
Downstream worker connections are tracked by worker id

talanknight:
When possible downstream workers will include their worker id information in
the connectios they establish with the upstream using the nodeenrollment
library.  The tracking listener now tracks connections that are connected in
this same way instead of just connections of type *tls.Conn.


jefferai (approved):
Approved assuming tests pass

PR #2960, opened 2023-02-14
feat(db): Introduce migrations for ssh targets

hugoamvieira:


internal/db/schema/migrations/oss/postgres/64/01_ssh_targets.up.sql
```diff
@@ -0,0 +1,120 @@
...
+
+  insert into oplog_ticket (name, version)
+  values ('target_ssh', 1)
+  on conflict do nothing;
+
+  -- These views depend on target_all_subtypes, so to drop target_all_subtypes,
+  -- we need to drop these first. They will be recreated later on this migration
+  -- version (see 02_wh_sessions and 03_wh_ssh_cert_library).
+  drop view if exists whx_host_dimension_source;
+  drop view if exists whx_credential_dimension_source;
```
tmessi, line 82: These are not being created again within this migration file. Can we group things together so the drop and create happens in a single migration? That might mean combining this and the next two migrations, but that means we will only drop these once, and it is easier to confirm that it gets created again.

tmessi (changes requested):

internal/db/schema/migrations/oss/postgres/64/01_ssh_targets.up.sql
hugoamvieira: That's fair, my only concern there is that the file would get too big but at least it offers that visibility.

internal/target/target.go
```diff
@@ -117,7 +122,12 @@
 
 	alloc, ok := subtypeRegistry.allocFunc(t.Subtype())
 	if !ok {
-		return nil, errors.New(ctx, errors.InvalidParameter, op, fmt.Sprintf("%s is an unknown target subtype of %s", t.PublicId, t.Type))
+		return nil, errors.Wrap(ctx,
+			errTargetSubtypeNotFound,
+			op,
+			errors.WithCode(errors.InvalidParameter),
+			errors.WithMsg(fmt.Sprintf("%s is an unknown target subtype of %s", t.PublicId, t.Type)),
```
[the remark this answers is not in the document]
hugoamvieira: Ignored as false positive

internal/db/schema/migrations/oss/postgres/64/01_ssh_targets.up.sql
hugoamvieira: Addressed with [65815a6](https://github.com/hashicorp/boundary/pull/2960/commits/65815a6bfde660c3aa395cd25140e07d59f99693)

internal/db/schema/migrations/oss/postgres/64/01_ssh_targets.up.sql
```diff
@@ -0,0 +1,363 @@
...
+  create trigger insert_ssh_target_filter_validate before insert on target_ssh
+    for each row execute procedure validate_filter_values_on_insert();
+
+  insert into oplog_ticket (name, version)
+  values ('target_ssh', 1)
+  on conflict do nothing;
+
+  -- Replaces view from 60/03_wh_sessions.up.sql
+  drop view if exists whx_host_dimension_source;
+  create view whx_host_dimension_source as
```
tmessi, line 80: I don't think you can create these views again until you have replaced the `target_all_subtypes`, otherwise the attempt to drop `target_all_subtypes` will fail.

hugoamvieira: Yeah, I just noticed I screwed that up, sorry 😂

tmessi: I think you should, drop all 3 views. Then create `target_all_subtypes` then create these `whx` views.

hugoamvieira: There we go: https://github.com/hashicorp/boundary/pull/2960/commits/0fda616a328cdb0c82d190366e77c25f8a63dd70

internal/db/schema/migrations/oss/postgres/64/01_ssh_targets.up.sql
```diff
@@ -0,0 +1,366 @@
...
+  create trigger update_ssh_target_filter_validate before update on target_ssh
+    for each row execute procedure validate_filter_values_on_update();
+
+  drop trigger if exists insert_ssh_target_filter_validate on target_ssh;
+  create trigger insert_ssh_target_filter_validate before insert on target_ssh
+    for each row execute procedure validate_filter_values_on_insert();
+
+  insert into oplog_ticket (name, version)
+  values ('target_ssh', 1)
+  on conflict do nothing;
```
tmessi, line 76:
```suggestion
  insert into oplog_ticker
    (name,         version)
  values
    ('target_ssh', 1)
  on conflict do nothing;
```

tmessi (changes requested):
some minor formatting things, otherwise I think this looks good

internal/db/schema/migrations/oss/postgres/64/01_ssh_targets.up.sql
hugoamvieira: All addressed with https://github.com/hashicorp/boundary/pull/2960/commits/b47627d41c5295717bcef0202d828d58cbc8a2e2 :)

tmessi (approved):

PR #2987, opened 2023-02-17
Get accountid from configure-aws-credentials action for OSS

joshbrand:
We were previously using the logic/method for doormat-based repos, which doesn't work with the `configure-aws-credentials` action (it masks the accountid...and also just provides it as an output out of the box.)


jaymalasinha (approved):
🚀

PR #2992, opened 2023-02-21
refact(e2e): Set up variables to prepare for docker implementation

moduli:
This PR modifies some of the e2e test suite variables in order to set it up for using docker as an infrastructure platform. These changes also make some things more explicit. Specifically...
- To maintain consistency for how we define ports, we are changing it to be explicitly defined in the enos scenario as opposed to relying on a "default" value. This change is in support of work that attempts to spin up test infrastructure using docker. The openssh-server docker image (that we use as an example host) uses Port 2222, which needs to be passed to the test suite.
- There are some instances where we need to have two different Vault addresses, one used by the Vault CLI and one by the Boundary server instance. In the case when using a Docker container, the Boundary server
needs the Vault address "internal" to the Docker network, whereas the Vault CLI running outside of the
network needs a different address.

### DIFF
--- a/testing/internal/e2e/tests/static_with_vault/env_test.go
+++ b/testing/internal/e2e/tests/static_with_vault/env_test.go
@@ -5,14 +5,14 @@ package static_with_vault_test
 
 import "github.com/kelseyhightower/envconfig"
 
-// `VaultAddr` is the address that the Boundary server uses to interact with the running Vault instance
 type config struct {
 	TargetIp         string `envconfig:"E2E_TARGET_IP" required:"true"`    // e.g. 192.168.0.1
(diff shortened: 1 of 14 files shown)


testing/internal/e2e/tests/static_with_vault/env_test.go
```diff
@@ -5,11 +5,13 @@ package static_with_vault_test
 
 import "github.com/kelseyhightower/envconfig"
 
+// `VaultAddr` is the address that the Boundary server uses to interact with the running Vault instance
```
johanbrandhorst, line 8: Put this over the field, not the struct. And no need for the backticks in code comments.

johanbrandhorst (approved):

PR #2993, opened 2023-02-22
chore(enos): Update test suite slack bot with different messages

moduli:
This PR updates the slack bot introduced in https://github.com/hashicorp/boundary/pull/2946. 

A new message will be sent if an enos job fails, but the retry passes (the retry that is built-in to the GitHub Actions workflow).
Screenshot 2023-02-21 at 7 04 47 PM

Additionally, the previous behavior would still report a slack message when a user manually kicked off "Re-run failed jobs" even if the re-run passed. There was no easy fix for that, so the workflow was modified to report on each scenario. 
Screenshot 2023-02-21 at 7 07 36 PM


johanbrandhorst (approved):
👏🏻

PR #2997, opened 2023-02-22
ci: Run go tests in github action

tmessi:
This runs the make targets:

- make test
- make test-api
- make test-sdk

The go tests in the main module are split by package and run in parallel
in a matrix build with GOMAXPROCS=1. This is done primarily to work
around an issue that is only seen on the github runners.  When running
at a higher GOMAXPROCS tests that use vault instances in docker
containers start to intermitently fail with errors from vault, usually
either a 400, 403, or TLS error. It is not clear what the cause of this
is, since the same tests run fine on self hosted runners and on
CircleCI.

Co-authored-by: Sarah Thompson <sthompson@hashicorp.com>
Co-authored-by: Daniel Kimsey <daniel.kimsey@hashicorp.com>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,190 @@
+name: test
+
+on:
+  - workflow_dispatch
+  - push
+  - workflow_call
+
+permissions:
+  contents: read
+
+jobs:
+  setup:
+    outputs:
+      go-version: ${{ steps.get-go-version.outputs.go-version }}
+      cache-go-build: ${{ steps.go-cache-paths.outputs.go-build }}
+      cache-go-mod: ${{ steps.go-cache-paths.outputs.go-mod }}
+      cache-go-bin: ${{ steps.go-cache-paths.outputs.go-bin }}
+      go-cache-key: ${{ steps.go-cache-key.outputs.key }}
+      plugin-cache-path: ${{ steps.plugin-cache-paths.outputs.path }}
+      plugin-cache-key: ${{ steps.plugin-cache-key.outputs.key }}
+    runs-on: ${{ fromJSON(vars.RUNNER) }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: '0'
+      - name: Determine Go version
+        id: get-go-version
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,156 @@
...
+
+  test:
+    needs:
+      - setup
+      - set-test-package-matrix
+    runs-on: ${{ fromJSON(vars.RUNNER) }}
+    strategy:
+      fail-fast: false
+      matrix:
+        package: ${{ fromJson(needs.set-test-package-matrix.outputs.package-matrix) }}
```
johanbrandhorst, line 119: Very nice. Does this run one job per package? If so, that sounds like we'll do a lot of setup for what could be a small amount of tests?

johanbrandhorst: Looks like this test run spent 4 minutes building boundary and 0.1s running the tests: https://github.com/hashicorp/boundary/actions/runs/4245687164/jobs/7382157690.

tmessi: Yea, specifically that is building the plugins, not all of boundary. I think we could move that plugin build step up prior to the matrix and cache it like we do the go mod cache. I would note that we have several packages that take >10m to run though. So as long as jobs are not waiting for runners, we are always limited by the slowest package.

johanbrandhorst: I'm generally ecstatic about this new run strategy, I think it'll fix a lot of issues around test timeouts and CI execution times, don't get me wrong! I wonder if it'd be worth amending the script in a way that we can group together known fast-running tests in a way that any new packages automatically get their own matrix coordinate. That way, we can amortize the cost of things like the postgres startup for tests that we know run fast.

tmessi: I think those can be follow up tasks if need be. The primary goal here is to get this running as a github action, we can do additional optimizations later.

johanbrandhorst: Totally fair 😄

johanbrandhorst (approved):
🎉 I'm very excited about this, great job!

PR #3000, opened 2023-02-22
chore(enos): Always destroy enos infrastructure

moduli:
This PR updates the enos GitHub action workflow to always run the `Destroy` command. 

Currently, we do the following
- `enos scenario run` (with `continue-on-error`)
- if failure, `enos scenario run` (retry)
- `enos scenario destroy`

If the Retry fails, it marks the workflow as failed (it does not have `continue-on-error`) and skips the destroy step. When testing locally, `enos scenario run` does not cleanup infrastructure if there is a failure. This leads to leaving some infrastructure hanging around.


joshbrand (approved):
👍

PR #3016, opened 2023-02-23
Update ui build scripts to check edition

ZedLi:
Updated the UI build scripts to check the edition to know which boundary-ui repo to pull artifacts from. OSS will pull from `boundary-ui` while hcp and ent will pull from `boundary-ui-enterprise`.

I'll followup with a PR in the enterprise repo to actually commit the new `VERSION_ent` file.

### DIFF
--- a/scripts/build-ui.sh
+++ b/scripts/build-ui.sh
@@ -5,30 +5,45 @@
 
 set -e
 
-if [ -z "$UI_COMMITISH" ]; then
-	echo "Must set UI_COMMITISH"; exit 1
-fi
-
 if [ -z "$UI_CLONE_DIR" ]; then
 	echo "Must set UI_CLONE_DIR"; exit 1
 fi
 
 if [ -z "$UI_VERSION_FILE" ]; then
-	echo "Must set UI_CLONE_DIR"; exit 1
+	echo "Must set UI_VERSION_FILE"; exit 1
 fi
+
+UI_EDITION=$(make --no-print-directory edition)
+
+if [ "$UI_EDITION" == "oss" ]; then
+  UI_REPO=https://github.com/hashicorp/boundary-ui
+  REPO_NAME=boundary-ui
+else
+  UI_VERSION_FILE="${UI_VERSION_FILE}_ent"
+  UI_REPO=https://github.com/hashicorp/boundary-ui-enterprise
+  REPO_NAME=boundary-ui-enterprise
+fi
+
+UI_CURRENT_COMMIT=$(head -n1 < "${UI_VERSION_FILE}" | cut -d' ' -f1)
+
+if [ -z "$UI_COMMITISH" ]; then
+  echo "==> Building default UI version from $UI_VERSION_FILE: $UI_CURRENT_COMMIT"
+  export UI_COMMITISH="$UI_CURRENT_COMMIT"
+else
+  echo "==> Building custom UI version $UI_COMMITISH"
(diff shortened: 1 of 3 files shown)


scripts/build-ui.sh
```diff
@@ -5,30 +5,45 @@
...
+else
+  echo "==> Building custom UI version $UI_COMMITISH"
+fi;
+
 if which gh &> /dev/null;  then
     echo "Found gh cli, attempting to download ui assets"
 
-    UI_EDITION=$(make --no-print-directory edition)
-    artifact_id=$(gh api repos/hashicorp/boundary-ui/actions/artifacts | \
+    artifact_id=$(gh api "repos/hashicorp/${REPO_NAME}/actions/artifacts" --paginate | \
```
ZedLi, line 39: We were only pulling the most recent ~30 artifacts. I added `--paginate` to actually look for all of them in case we're looking for an artifact that's a bit older but this could be a bit slower. Let me know if you think this is unnecessary as github artifacts are only stored for ~90 days anyways.

scripts/build-ui.sh
```diff
@@ -66,5 +81,5 @@ git pull --ff-only origin "${UI_COMMITISH}"
 git reset --hard "${UI_COMMITISH}"
 
 yarn install
-yarn build
+EDITION=${UI_EDITION} yarn build
```
ZedLi, line 84: This is to fix an issue where if you build it locally from cloning the repository, we weren't actually setting the edition of the UI correctly.

tmessi (approved):

scripts/build-ui.sh
calcaide: I think you just kill 2 birds with one stone!! this will solve [this jira ticket we had](https://hashicorp.atlassian.net/browse/ICU-6689) 😃

scripts/build-ui.sh
calcaide: Good catch!

scripts/uiupdate.sh
```diff
@@ -8,12 +8,22 @@ set -e
...
 	echo "Must set UI_VERSION_FILE"; exit 1
 fi
+
 if [ -z "$UI_COMMITISH" ]; then
 	echo "Must set UI_COMMITISH"; exit 1
 fi
 
+UI_EDITION=$(make --no-print-directory edition)
+
+if [ "$UI_EDITION" == "oss" ]; then
```
calcaide, line 18: None blocking question and just wondering about it:
- The logic of finding out the version is within the main make file. Which is good.
- Then I see `uiupdate.sh` and `build-ui` are using (and repeating) same logic to determinate: `UI_EDITION`, `UI_REPO` and `UI_VERSION`.

Do you think in the future we could move that logic to the make file? so we apply this logic once, set those ENV variables and when using `build-ui` or `uiupdate` those are already set.

Do you think that could be douable & beneficial for the future?

calcaide (approved):
LGTM!! Thank you for this work, killing it!

cameronperera (approved):

scripts/uiupdate.sh
ZedLi: I don't know makefiles that well but because the `edition` comes from the target, I'm not exactly sure how I would set it like we do for our static variables. I could do it in each target before the script but that's basically the same thing.

lisbet-alvarez (approved):

PR #3032, opened 2023-02-27
ci: Fix permissions in labeler workflow

moduli:
This PR is a follow-up to https://github.com/hashicorp/boundary/pull/3011.

It addresses an issue where the `labeler` workflow was failing to the following reason.

Screenshot 2023-02-27 at 9 49 25 AM

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -3,7 +3,7 @@ on:
 - pull_request_target
 
 permissions:
-  content: read
+  contents: read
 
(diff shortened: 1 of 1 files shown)


moduli:
Note: I think it's still failing on this branch because it runs against `pull_request_target`, which I think means it's using the `main` version still?

tmessi (approved):

tmessi:
This might need `pull-requests: write` as well

PR #3036, opened 2023-02-28
Create a solo worker module for testing

manthonygfp:
This module adds support for provisioning a worker node separately from the Boundary cluster in Enos scenarios.

For the mean time, I have left the dummy scenario that I use to exercise the module. I will remove that before merging this pull request.

I'm wondering what outputs may be useful for a user of the module. I was thinking that the subnet and availability zone may be the details we want to provide as an output so that targets can be put in the same AZ and under the same subnet.

### DIFF
--- a/enos/enos-modules.hcl
+++ b/enos/enos-modules.hcl
@@ -11,7 +11,7 @@ module "bats_deps" {
 
 module "boundary" {
   source  = "app.terraform.io/hashicorp-qti/aws-boundary/enos"
-  version = ">= 0.2.6"
+  version = ">= 0.4.0"
 
   project_name = "qti-enos-boundary"
   environment  = var.environment
@@ -24,6 +24,21 @@ module "boundary" {
 
   ssh_aws_keypair       = var.aws_ssh_keypair_name
   alb_listener_api_port = var.alb_listener_api_port
+}
+
+module "solo_worker" {
+  source = "./modules/solo_worker"
+
+  project_name = "qti-enos-boundary"
+  environment  = var.environment
+  common_tags = {
+    "Project" : "Enos",
+    "Project Name" : "qti-enos-boundary",
+    "Enos User" : var.enos_user,
+    "Environment" : var.environment
+  }
+
+  ssh_aws_keypair = var.aws_ssh_keypair_name
 }
 
 module "build_crt" {
--- a/enos/modules/solo_worker/main.tf
+++ b/enos/modules/solo_worker/main.tf
@@ -1,0 +1,173 @@
+terraform {
+  required_providers {
+    enos = {
+      source = "app.terraform.io/hashicorp-qti/enos"
+    }
+  }
+}
+
+locals {
+  selected_az = data.aws_availability_zones.available.names[random_integer.az.result]
+}
+
+resource "random_string" "cluster_id" {
+  length  = 8
+  upper   = false
+  numeric = false
+  special = false
+}
+
+resource "random_pet" "worker" {
+  separator = "_"
+}
+
+data "aws_availability_zones" "available" {
+  state = "available"
+  filter {
+    name   = "zone-name"
+    values = var.availability_zones
+  }
+}
+
+data "aws_availability_zone" "worker_az" {
+  name = local.selected_az
+}
+
+data "aws_kms_key" "kms_key" {
+  key_id = var.kms_key_arn
+}
+
+resource "random_integer" "az" {
+  min = 0
+  max = length(data.aws_availability_zones.available.names) - 1
+  keepers = {
+    # Generate a new integer each time the list of aws_availability_zones changes
+    # keepers have to be strings, sort the list in case order changes but zones don't
+    listener_arn = join("", sort(data.aws_availability_zones.available.names))
+  }
+}
+
+# Create a subnet so that the worker doesn't share one with a controller
+resource "aws_subnet" "default" {
+  vpc_id                  = var.vpc_name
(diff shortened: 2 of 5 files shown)


manthonygfp:
My current holdup is that I'm looking to ensure that this worker will be able to work with a given target for testing.

From [here](https://github.com/hashicorp/boundary/blob/main/enos/enos-scenario-e2e-aws.hcl#L90), I can see that the tags are made and then attached to the target _after_ the Boundary cluster and its workers have been built.

Since this module creates a worker in its own subnet, I have questions about whether or not my current approach is already adequate to work with targets like in the above-linked scenario. I think a scenario built around it would need to ensure that the target is made with the solo worker in mind and, as such, the worker's module doesn't need to be modified to accommodate a target in any special way.

Any pointers about what might be missing from the worker's config or helpful outputs would be greatly appreciated!

manthonygfp:
Made https://github.com/hashicorp/terraform-enos-aws-boundary/pull/27 in relation to enabling this module to function properly coupled with the base Boundary cluster module.

moduli:
Suggestion to pull in the latest from main. This branch is a bit behind.

enos/enos-modules.hcl
```diff
@@ -26,6 +26,18 @@ module "boundary" {
   alb_listener_api_port = var.alb_listener_api_port
 }
 
+module "solo_worker" {
```
moduli, line 29: Do we need to add `project_name` here? I see it in the `boundary` module and it's defined in `variables.tf`, but I don't see it used anywhere?

enos/modules/solo_worker/main.tf
```diff
@@ -0,0 +1,178 @@
...
+  min = 0
+  max = length(data.aws_availability_zones.available.names) - 1
+  keepers = {
+    # Generate a new integer each time the list of aws_availability_zones changes
+    # keepers have to be strings, sort the list in case order changes but zones don't
+    listener_arn = join("", sort(data.aws_availability_zones.available.names))
+  }
+}
+
+resource "aws_subnet" "default" {
```
joshbrand, line 56: Maybe a comment here to clarify that this is a separate subnet from the created ones in infra

enos/modules/solo_worker/templates/worker.hcl
```diff
@@ -0,0 +1,28 @@
...
+  tags {
+    region = "["${region}"]"
+    type = ${type}
+  }
+}
+
+# must be same key as used on controller config
+kms "awskms" {
+  purpose    = "worker-auth"
+  region     = "${region}"
```
joshbrand, line 26: nit: does this need quotes still?

enos/enos-modules.hcl
```diff
@@ -26,6 +26,18 @@ module "boundary" {
...
+  source = "./modules/solo_worker"
+
+  environment = var.environment
+  common_tags = {
+    "Project" : "Enos",
+    "randomtagid" : "hippocar",
```
joshbrand, line 35: Should this be configurable or is it just a proof of concept?

joshbrand (commented):
couple questions, but looks great!

enos/modules/solo_worker/main.tf
```diff
@@ -0,0 +1,178 @@
...
+      source = "app.terraform.io/hashicorp-qti/enos"
+    }
+  }
+}
+
+locals {
+  tags = merge(
+    var.common_tags,
+    {
+      "PetName" = random_pet.worker.id
```
moduli, line 13: What is this used for? Does `pet` stand for something?

enos/modules/solo_worker/templates/worker.hcl
manthonygfp: I believe so, yes.

String templates don't interpolate strings with their quote sigils https://developer.hashicorp.com/terraform/language/expressions/strings#string-templates

And the HCL blocks don't use quotes for the keys, but they do for the values https://developer.hashicorp.com/boundary/docs/configuration/worker/kms-worker

Does that read the same to you? Or maybe there's something else that I'm missing.

enos/enos-modules.hcl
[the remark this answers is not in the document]
manthonygfp: Altered this to match the boundary module as appropriate

enos/enos-modules.hcl
manthonygfp: I've added this to be piped in to the module, but I don't really have a use for it. Still, it might be useful to have as this module gets used, but it doesn't serve a purpose right now other than mirroring the setup it came from.

enos/modules/solo_worker/templates/worker.hcl
```diff
@@ -0,0 +1,28 @@
...
+  address = "0.0.0.0"
+}
+
+worker {
+  # Name attr must be unique across workers
+  name = "worker-${id}"
+  description = "Enos Boundary worker ${id}"
+
+  # Workers must be able to reach controllers on :9201
+  controllers = ${controller_ips}
```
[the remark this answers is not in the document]
manthonygfp: https://github.com/hashicorp/boundary/pull/2151

Looks like the verbiage was changed in this PR, interesting, I'll correct that and amend the commit

enos/modules/solo_worker/main.tf
manthonygfp: I'll remove it, I use this to more easily identify which pieces of virtual hardware belong to a given instance of this module

manthonygfp:
Closing this. I'll open another PR once I have the security group issues sorted.

PR #3037, opened 2023-02-28
chore(enos): Add debug variable

moduli:
If you're running an enos scenario and there's a failure in the test suite for any reason (bug, test issue), enos will throw an error and you would not be able to access the environment variables needed to test/reproduce it locally. Enabling this `e2e_debug` variable will ensure that the enos scenario passes so that you can extract the needed environment variables.

This PR does the following
- adds an enos variable `e2e_debug` to stop scenarios from running the test suite so that you can always get the environment variables of the generated infrastructure
- adds a script `test_e2e_env.sh` to more easily extract the environment variables

### DIFF
--- a/enos/enos-variables.hcl
+++ b/enos/enos-variables.hcl
@@ -125,7 +125,7 @@ variable "local_build_target" {
   default     = "build-ui build"
 }
 
-variable "e2e_debug" {
+variable "e2e_debug_no_run" {
   description = "If set, this will prevent test suites from running"
   type        = bool
   default     = false
--- a/enos/enos-variables.hcl
+++ b/enos/enos-variables.hcl
@@ -124,3 +124,9 @@ variable "local_build_target" {
   type        = string
   default     = "build-ui build"
 }
+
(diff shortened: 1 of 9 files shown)


enos/enos-variables.hcl
```diff
@@ -124,3 +124,9 @@ variable "local_build_target" {
   type        = string
   default     = "build-ui build"
 }
+
+variable "e2e_debug" {
```
johanbrandhorst, line 128: The name of this variable doesn't clearly communicate the effect IMO, what do you think about 
```suggestion
variable "e2e_debug_no_run" {
```
Or something?

enos/modules/test_e2e/main.tf
```diff
@@ -9,6 +9,11 @@ terraform {
   }
 }
 
+variable "debug" {
```
johanbrandhorst, line 12:
```suggestion
variable "debug_no_run" {
```

johanbrandhorst (approved):

PR #3044, opened 2023-03-02
ci: Add actionlint workflow

tmessi:


shore (approved):
LGTM!

.github/workflows/test-cli-ui.yml
```diff
@@ -0,0 +1,140 @@
+name: test-cli-ui
+
+on:
+  workflow_dispatch:
+  workflow_run:
+    workflows: [ build ]
+    types:
+      - completed
```
tmessi, line 8: @moduli flagging you for visibility on this. It might be something to do with the enos workflows as well, so that the `build` workflow just does the building and these workflows that perform tests can just after the build is done.

PR #3045, opened 2023-03-02
Fix hcp billing test

irenarindos:


tmessi (changes requested):
Requested some style changes. And I think this probably ok for now, but we should put in a TODO to revisit this. It would be nice if we could test both test cases here all of the time, instead of the test asserting different things depending on the time that it is run.

irenarindos:
Created https://hashicorp.atlassian.net/browse/ICU-8338 on the backlog

tmessi (approved):

PR #3057, opened 2023-03-06
chore(enos): Install tparse directly to save time

moduli:
This PR updates the enos GitHub Action workflow to install `tparse` from GitHub releases directly as opposed to using `make tools`. 

Using `make tools` adds an extra 1-2 minutes of runtime when the only thing that is needed from it is `tparse`. While it would be nice to get `tparse` from the same place that the `go tests` get it from, I'm not sure if that benefit outweighs the extra runtime it adds.


moduli:
There are some failures from `test.yml` where it's complaining that `tparse` cannot be found... I'm presuming that it's unrelated to my changes. Looking at some of the logs, it had trouble finding the right cache?

```
Cache not found for input keys: Linux-go-6777691c3bab52208acf95cedf15a65b6b462bcdfb05c3c0ac2072660ab277a9, Linux-go
```
https://github.com/hashicorp/boundary/actions/runs/4328069979/jobs/7588858974

tmessi:
> There are some failures from `test.yml` where it's complaining that `tparse` cannot be found... I'm presuming that it's unrelated to my changes. Looking at some of the logs, it had trouble finding the right cache?
> 
> ```
> Cache not found for input keys: Linux-go-6777691c3bab52208acf95cedf15a65b6b462bcdfb05c3c0ac2072660ab277a9, Linux-go
> ```
> 
> https://github.com/hashicorp/boundary/actions/runs/4328069979/jobs/7588858974

It does seem unrelated, but it is odd since other test packages in that matrix are able to pull the cache just fine. Might just be a transient issue. But we should probably mark that cache restore step to `fail-on-cache-miss`.

As for this PR, I think it would be better to still install via `make tools`, but to cache it like the `test.yml` one is. In that workflow there is a [setup job](https://github.com/hashicorp/boundary/blob/953cc3dae324f20d40175ed0eb712f4349da4e76/.github/workflows/test.yml#L15-L63) that will cache the go mod cache, and the binaries installed in `$GOPATH/bin`. When there is a cache hit, it will save time on the run, and by including the `Makefile and `go.mod` in the hash for the cache key, it keeps the specific version codified the `Makefile` (or `go.mod` depending on the tool) so this workflow does not need to be updated when that tool version is updated.

moduli:
I'll give that a shot

.github/workflows/enos-run.yml
```diff
@@ -17,8 +17,49 @@ env:
...
+        with:
+          fetch-depth: '0'
+      - name: Set up Go
+        uses: actions/setup-go@d0a58c1c4d2b25278816e339b944508c875f3613 # v3.4.0
+        with:
+          go-version: ${{ inputs.go-version }}
+      - name: Determine go cache key
+        id: go-cache-key
+        run: |
+          echo "key=${{ runner.os }}-go-${{ hashFiles('**/go.sum', './Makefile', './tools/tools.go') }}" >> "$GITHUB_OUTPUT"
```
moduli, line 38: Even though this only cares about the Makefile (the version of `tparse`), does it make sense to still codify the cache based on the same files as its used in `test.yml`? That way, it could share those same caches and save on space.

tmessi: Yes, I would make the cache correspond to the things that actually result in a change. So since you are calling `make tools`, any changes to those files would result in a different result when calling `make tools`, even though in this case only tparse is used.

tmessi (approved):

PR #3059, opened 2023-03-06
Add extra IAM permissions for demo user creation during e2e tests

joshbrand:
These statements got mixed up in the account migration, this policy should be applied to OSS (non-doormat) projects that need IAM access for testing


moduli:
What problem is this addressing? Aren't things running fine right now?

moduli (approved):
Discussed in Slack. This is to codifying a fix that was done manually to address some permissions when attempting to create users (specifically when creating an AWS Dynamic Host Catalog for end-to-end tests).

PR #3064, opened 2023-03-08
Use tofu token from controller

### LINKED ISSUE #3056: Upgrading to `0.12.0` brakes connection to workers behind AWS NLB
**Describe the bug**
After upgrading our Boundary cluster from version `0.11.2` to version `0.12.0`, connections erroring out with ` server closed the connection unexpectedly` in pgAdmin4 and `no tofu token but not in correct session state` errors in Boundary workers.

Relevant parts of Boundary Worker logs:
```

1. pod: "boundary-worker-1" : {"connection_id":"sc_D8zk6zHJSr","msg":"connection successfully authorized","session_id":"s_hwOqxxOyO3"}
2. pod: "boundary-worker-1" : {"msg":"session successfully activated","session_id":"s_hwOqxxOyO3"}
3. pod: "boundary-worker-1" : {"connection_id":"sc_D8zk6zHJSr","msg":"connection closed","session_id":"s_hwOqxxOyO3"}
4. pod: "boundary-worker-0" : {"error":"no tofu token but not in correct session state","error_fields":{},"id":"e_9ln52LRbZM","version":"v0.1","op":"worker.(Worker).handleProxy","info":{"session_id":"s_hwOqxxOyO3"}}
5. pod: "boundary-worker-2" : {"error":"no tofu token but not in correct session state","error_fields":{},"id":"e_v9Bj5hHuTH","version":"v0.1","op":"worker.(Worker).handleProxy","info":{"session_id":"s_hwOqxxOyO3"}}
6. pod: "boundary-worker-1" : {"connection_id":"sc_cAX7tHZJGK","msg":"connection successfully authorized","session_id":"s_hwOqxxOyO3"}

```

As you can see the same session(`s_hwOqxxOyO3`), errors out when it gets routed to a different worker in the cluster.

#### About the setup
Boundary cluster with 3 controllers and 3 workers is deployed on AWS EKS. Workers are publicly exposed behind an AWS NLB. TCP port `9202` is open on an `ingress-nginx` and routes to a `ClusterIP` Kubernetes Service with three Boundary Worker Pods. Same for controllers, except the port is `9201`. This setup was working fine before the upgrade.

**To Reproduce**
1. Upgrade to version `0.12.0`
2. Connect via Boundary UI
3. Login via pgAdmin4
4. See error: `server closed the connection unexpectedly`
5. Login multiple times
6. Login successful
7. Open database
8. See error: `server closed the connection unexpectedly`

**Expected behavior**
If we downgrade back to version `0.11.2`, connections work fine without any issues.

**Additional context**
- Controller and Worker communicate via public TCP ingress connection, switching to an internal cluster network, did not solve the issue.
- We are not using the new `address` field in the Boundary Target.
- We are using the Worker Tag filters

irenarindos:
Fixes #3056

### DIFF
--- a/internal/daemon/worker/session/session.go
+++ b/internal/daemon/worker/session/session.go
@@ -221,7 +221,10 @@ func (s *sess) GetTofuToken() string {
 	if len(s.tofuToken) > 0 {
 		return s.tofuToken
 	}
-	return s.resp.GetTofuToken()
+	if s.resp != nil {
+		return s.resp.GetTofuToken()
+	}
+	return ""
 }
 
 func (s *sess) GetConnectionLimit() int32 {
--- a/internal/daemon/worker/session/session.go
+++ b/internal/daemon/worker/session/session.go
@@ -217,7 +217,7 @@ func (s *sess) GetLocalConnections() map[string]ConnInfo {
 func (s *sess) GetTofuToken() string {
 	s.lock.RLock()
 	defer s.lock.RUnlock()
-	return s.tofuToken
+	return s.resp.GetTofuToken()
 }
 
 func (s *sess) GetConnectionLimit() int32 {
--- a/internal/daemon/worker/session/session.go
+++ b/internal/daemon/worker/session/session.go
@@ -214,9 +214,13 @@ func (s *sess) GetLocalConnections() map[string]ConnInfo {
 	return res
 }
 
+// Return local tofu token, if availble; otherwise return the tofu token from the resp
 func (s *sess) GetTofuToken() string {
 	s.lock.RLock()
 	defer s.lock.RUnlock()
+	if len(s.tofuToken) > 0 {
+		return s.tofuToken
+	}
 	return s.resp.GetTofuToken()
 }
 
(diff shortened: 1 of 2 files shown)


internal/daemon/worker/session/session.go
```diff
@@ -214,9 +214,13 @@ func (s *sess) GetLocalConnections() map[string]ConnInfo {
 	return res
 }
 
+// Return local tofu token, if availble; otherwise return the tofu token from the resp
 func (s *sess) GetTofuToken() string {
 	s.lock.RLock()
 	defer s.lock.RUnlock()
+	if len(s.tofuToken) > 0 {
+		return s.tofuToken
+	}
```
jefferai, line 223: Can you also add a nil check on Resp, in case it's nil? I"m not sure why it would be, but seems like a good idea...

internal/daemon/worker/session/session.go
```diff
@@ -214,10 +214,14 @@ func (s *sess) GetLocalConnections() map[string]ConnInfo {
...
 
+// Return local tofu token, if availble; otherwise return the tofu token from the resp
 func (s *sess) GetTofuToken() string {
 	s.lock.RLock()
 	defer s.lock.RUnlock()
-	return s.tofuToken
+	if len(s.tofuToken) > 0 {
+		return s.tofuToken
+	}
+	return s.resp.GetTofuToken()
```
jefferai, line 224: Can you also add a nil check on Resp, in case it's nil? I"m not sure why it would be, but seems like a good idea...

jefferai (approved):

PR #3065, opened 2023-03-08
ci: Remove CircleCI configuration

tmessi:
CI is now performed via Github Actions. Some notable Github Action
workflows that replace the CircleCI jobs are:

- test: replaces CCI build, test-api, test-sdk
- test-sql: replaces CCI test-sql
- make-gen-delta: replaces CCI make-gen-delta

See:
    https://github.com/hashicorp/boundary/actions/workflows/test.yml
    https://github.com/hashicorp/boundary/actions/workflows/test-sql.yml
    https://github.com/hashicorp/boundary/actions/workflows/make-gen-delta.yml
    https://docs.github.com/en/actions
    https://cli.github.com/manual/gh_run


moduli (commented):
1. The PR status check still shows a failure due to "CircleCI Pipeline". Not sure where that is defined
2. It's probably not a big deal, but I see some references to circleci in `.gitattributes` and `.copywrite.hcl`

tmessi:
> 1. The PR status check still shows a failure due to "CircleCI Pipeline". Not sure where that is defined

That will go away once things are turned off in CircleCI.

> 
>     2. It's probably not a big deal, but I see some references to circleci in `.gitattributes` and `.copywrite.hcl`

Good catch, I removed these.

.gitattributes
```diff
@@ -1,2 +0,0 @@
-/.circleci/config/@build-release.yml
-/packagespec.mk linguist-generated
```
moduli, line 2: Is this related to CircleCI or is this just something we don't use anymore?

tmessi: Something we havn't used in a while.

moduli (approved):

PR #3073, opened 2023-03-10
chore(enos): Add scenario that uses docker containers for infrastructure

moduli:
This PR adds a new enos scenario (that does NOT run in CI yet) that uses docker containers for infrastructure as opposed to enos-created infrastructure in AWS. This was done to reduce run times and mitigate flakiness we've seen when using AWS infrastructure.

Comparison of run times:
```
static_with_vault scenario (current): 12m 33s
static_with_vault scenario (docker): 3m 24s
```
Current scenario: https://github.com/hashicorp/boundary/actions/runs/4386980918/jobs/7681933933
Docker scenario: https://github.com/hashicorp/boundary/actions/runs/4386883281/jobs/7681797986

This PR contains the following...
- a new enos scenario that spins up docker containers for infrastructure and runs one of the end-to-end test packages
- enos modules that are used by the enos scenario to create docker containers
- updated CI workflow to make the docker scenario work. This scenario is NOT enabled to run in CI right now.

Follow-up work includes...
 - Currently, this scenario uses a Boundary docker image of the latest public release. We will need to update it so that in CI, it will use the docker image generated by the build workflow and if run locally, will build a docker image using your local code.
 - Refactor the existing enos scenarios to have most of the tests use docker infrastructure


enos/enos-scenario-e2e-docker.hcl
```diff
@@ -0,0 +1,104 @@
...
+    }
+  }
+
+  step "create_vault" {
+    module = module.docker_vault
+    depends_on = [
+      step.create_docker_network
+    ]
+    variables {
+      image_name   = "docker.io/hashicorp/vault:1.12.2"
```
johanbrandhorst, line 58: Did we not version Vault somewhere centrally already? Or is this maybe a reason to introduce such a central place?

enos/modules/docker_boundary/auditlogs/.keep
johanbrandhorst, line 1: Is this file necessary?

enos/modules/docker_boundary/boundary-config.hcl
```diff
@@ -0,0 +1,105 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+disable_mlock = true
+
+controller {
+  name        = "docker-controller"
+  description = "A controller for a docker demo!"
```
johanbrandhorst, line 8: Maybe change this line

enos/modules/docker_boundary/init.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
johanbrandhorst, line 1: Please use
```suggestion
#!/usr/bin/env bash
```
To make this file more portable

enos/modules/docker_boundary/init.sh
```diff
@@ -0,0 +1,30 @@
...
+    --cap-add IPC_LOCK \
+    --mount type=bind,src=$SOURCE,dst=/boundary/ \
+    --network e2e_network \
+    $TEST_BOUNDARY_IMAGE \
+    boundary database init -config /boundary/boundary-config.hcl -format json \
+    >/dev/null 2>&1 # suppresses output
+
+docker wait $TEST_CONTAINER_NAME >/dev/null 2>&1
+echo $(docker logs $TEST_CONTAINER_NAME)
+docker rm -v $TEST_CONTAINER_NAME >/dev/null 2>&1 || true
```
johanbrandhorst, line 30: Instead of starting a container in the background, waiting for it to finish, printing its logs and then remove it, why don't we let it run to completion in the foreground with `--rm`?
```suggestion
docker run \
    --rm \
    --name $TEST_CONTAINER_NAME \
    -e "BOUNDARY_POSTGRES_URL=$TEST_DATABASE_ADDRESS" \
    -e "SKIP_CHOWN=true" \
    --cap-add IPC_LOCK \
    --mount type=bind,src=$SOURCE,dst=/boundary/ \
    --network e2e_network \
    $TEST_BOUNDARY_IMAGE \
    boundary database init -config /boundary/boundary-config.hcl -format json \
    >/dev/null 2>&1 # suppresses output
```

enos/modules/docker_openssh_server/main.tf
```diff
@@ -0,0 +1,79 @@
...
+  type        = string
+  default     = "ubuntu"
+}
+variable "private_key_file_path" {
+  description = "Local Path to key used to SSH onto created hosts"
+  type        = string
+}
+
+locals {
+  is_openssh = length(regexall(".*OPENSSH.*", file(var.private_key_file_path))) > 0
```
johanbrandhorst, line 38: What is this used for?

johanbrandhorst (commented):
Very excited for this, great job!

enos/modules/docker_boundary/auditlogs/.keep
moduli: I saw this in `boundary-local-env`, which I think was a way to keep that directory around so that it can be used as a mount point for the boundary container. Is there an alternative that I'm not thinking about?

enos/modules/docker_openssh_server/main.tf
moduli: 😅 Yup. No longer used. Originally had some logic to parse the key differently if it didn't say OPENSSH (i.e. RSA), but it wasn't necessary (mostly due to my lack of understanding of key encryption terminology).

enos/enos-scenario-e2e-docker.hcl
moduli: Yup we did 😅

enos/modules/docker_boundary/auditlogs/.keep
tmessi: An alternative would be to have a `mkdir` call somewhere prior to using that mount point, although I prefer this approach if this is needed. Are the audit logs actually needed though?

moduli: Not necessarily. It was an idea to potentially upload the audit log from the container if it would be helpful for debugging purposes. I'll take that part out for now.

enos/modules/docker_boundary/init.sh
```diff
@@ -0,0 +1,25 @@
...
+docker run \
+    --rm \
+    --name $TEST_CONTAINER_NAME \
+    -e "BOUNDARY_POSTGRES_URL=$TEST_DATABASE_ADDRESS" \
+    -e "SKIP_CHOWN=true" \
+    --cap-add IPC_LOCK \
+    --mount type=bind,src=$SOURCE,dst=/boundary/ \
+    --network e2e_network \
+    $TEST_BOUNDARY_IMAGE \
+    boundary database init -config /boundary/boundary-config.hcl -format json \
```
johanbrandhorst, line 25:
```suggestion
    boundary database init -config /boundary/boundary-config.hcl -format json
```

johanbrandhorst (changes requested):

johanbrandhorst (approved):

PR #3080, opened 2023-03-14
[RELENG-483] lower test run frequency per hour

jeanneryan:
Lowering the run frequency until GitHub increases our concurrency limit


claire-labry (approved):
🥳

PR #3082, opened 2023-03-14
ci(schema-diff): Generate one file per constraint

tmessi:
This yeilds a better diff when constraints are changed.


hugoamvieira (approved):
Looks good, thanks Tim! 😄

PR #3086, opened 2023-03-15
chore(enos): Use docker mirror

moduli:
This PR updates the enos GitHub Action workflow to use our internal docker mirror in order to insulate ourselves from issues with GitHub. We were experiencing test failures when trying to pull the `linuxserver/openssh-server` docker image due to an issue with GitHub Packages.

Note: There is an enos test reporting a failure in the status checks due to a bug in enos with the latest Terraform. That work is ongoing.

### DIFF
--- a/testing/internal/e2e/infra/docker.go
+++ b/testing/internal/e2e/infra/docker.go
@@ -239,7 +239,7 @@ func StartOpenSshServer(t testing.TB, pool *dockertest.Pool, network *dockertest
 
 	networkAlias := "target"
 	resource, err := pool.RunWithOptions(&dockertest.RunOptions{
-		Repository: "lscr.io/linuxserver/openssh-server",
+		Repository: "docker.mirror.hashicorp.services/linuxserver/openssh-server",
 		Env: []string{
 			"PUID=1000",
 			"PGID=1000",


.github/workflows/enos-run.yml
```diff
@@ -15,6 +15,7 @@ on:
 
 env:
   PKG_NAME: boundary
+  DOCKER_MIRROR: docker.mirror.hashicorp.services
```
tmessi, line 18: Is this env var used by enos anywhere? It gets used by the unit tests [here](https://github.com/hashicorp/boundary/blob/28756dce6717ccbac3bfc0c11ba77d1bbac18047/testing/dbtest/docker/Makefile#L9) and [here](https://github.com/hashicorp/boundary/blob/28756dce6717ccbac3bfc0c11ba77d1bbac18047/internal/credential/vault/supported.go#L40-L44).

moduli: Oh 😂 I thought it was some environment variable that got magically consumed. I'll take a look.

testing/internal/e2e/infra/docker.go
```diff
@@ -239,7 +239,7 @@ func StartOpenSshServer(t testing.TB, pool *dockertest.Pool, network *dockertest
 
 	networkAlias := "target"
 	resource, err := pool.RunWithOptions(&dockertest.RunOptions{
-		Repository: "lscr.io/linuxserver/openssh-server",
+		Repository: "docker.mirror.hashicorp.services/linuxserver/openssh-server",
```
moduli, line 242: Updating it here in the test for now. Will look at potentially refactoring this with the other docker work I have ongoing (using docker containers to run more of the e2e tests).

johanbrandhorst (approved):

PR #3092, opened 2023-03-16
Refactor Target Repository

elimt:
## Change
Refactor the Target domain repository functions so that they no longer return 4 different types.

`Target` struct now contains `HostSource` & `CredentialSources` with Getters & Setters

## Rationale
Extending `Target` to contain `HostSource` & `CredentialSources` enables `funcs` to just return `Target` instead of multiple return types. This also allows easy future refactors and extensions of `Target`

## Breaking changes
There should not be any breaking changes from this change

### DIFF
--- a/internal/cmd/base/initial_resources.go
+++ b/internal/cmd/base/initial_resources.go
@@ -461,7 +461,7 @@ func (b *Server) CreateInitialTargetWithAddress(ctx context.Context) (target.Tar
 	if err != nil {
 		return nil, fmt.Errorf("failed to create target object: %w", err)
 	}
-	tt, _, _, err := targetRepo.CreateTarget(ctx, t, opts...)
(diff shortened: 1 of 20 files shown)


mgaffney:
Why is this a "draft" PR? Is there more work anticipated?

mgaffney (commented):
Why is this a draft PR? Is there more work anticipated?

elimt:
> Why is this a draft PR? Is there more work anticipated?

@mgaffney I left it as a draft pending the outcome of the RFC. In case we decide to do something else.

Otherwise, there isn't more work left for the changes with the current proposal

mgaffney (approved):

tmessi (approved):
nice!

ddebko (approved):
LGTM

PR #3100, opened 2023-03-17
chore(enos): Use docker image generated from code

moduli:
This PR continues to extend on the docker work for e2e tests introduced here: https://github.com/hashicorp/boundary/pull/3073. It adds support for the following...
- when run in CI, it will use the docker image built in the Actions workflow 
- when run locally, it will build a boundary docker image using your local checkout and use that to start a docker container

Next step would be to refactor the existing enos scenarios to make use of the docker containers.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -357,6 +357,8 @@ jobs:
     strategy:
       matrix:
         arch: ["arm", "arm64", "386", "amd64"]
+    outputs:
+      tag: docker.io/hashicorppreview/${{ env.repo }}:${{ env.minor-version }}-dev-${{ github.sha }}
     env:
       repo: ${{ github.event.repository.name }}
       version: ${{ needs.set-product-version.outputs.product-version }}
@@ -394,6 +396,8 @@ jobs:
     with:
       artifact-name: "boundary_${{ needs.set-product-version.outputs.product-version }}_linux_amd64.zip"
       go-version: ${{ needs.product-metadata.outputs.go-version }}
+      docker-image-name: ${{ needs.build-docker.outputs.tag }}
+      docker-image-file: "boundary_default_linux_amd64_${{ needs.set-product-version.outputs.product-version }}_${{ github.sha }}.docker.dev.tar"
     secrets: inherit
 
   bats:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -358,7 +358,7 @@ jobs:
       matrix:
         arch: ["arm", "arm64", "386", "amd64"]
     outputs:
-      tag: docker.io/hashicorppreview/${{ env.repo }}:${{ env.minor-version }}-dev-${{ github.sha }}
+      name: docker.io/hashicorppreview/${{ env.repo }}:${{ env.minor-version }}-dev-${{ github.sha }}
     env:
       repo: ${{ github.event.repository.name }}
       version: ${{ needs.set-product-version.outputs.product-version }}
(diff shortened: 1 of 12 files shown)


enos/modules/build_boundary_docker_local/build.sh
```diff
@@ -0,0 +1,24 @@
...
+docker buildx inspect multiarch || docker buildx create --use --driver docker-container --name multiarch
+docker buildx build \
+    --load \
+    --platform ${GOOS}/${GOARCH} \
+    --tag ${IMAGE_NAME} \
+    --target=dev \
+    --build-arg=boundary \
+    --output="type=docker,dest=${ARTIFACT_PATH}" \
+    .
+docker buildx rm multiarch
```
moduli, line 22: Should this be added to the Makefile? cc: @tmessi

.github/workflows/enos-run.yml
```diff
@@ -66,10 +72,11 @@ jobs:
...
         include:
-          - filter: 'e2e_aws builder:crt'
-          - filter: 'e2e_database'
-          - filter: 'e2e_static builder:crt'
-          - filter: 'e2e_static_with_vault builder:crt'
+          # - filter: 'e2e_aws builder:crt'
+          # - filter: 'e2e_database'
+          # - filter: 'e2e_static builder:crt'
+          # - filter: 'e2e_static_with_vault builder:crt'
+          - filter: 'e2e_docker builder:crt'
```
moduli, line 79: - [ ] Revert before merge

.github/workflows/enos-run.yml
```diff
@@ -272,28 +296,28 @@ jobs:
...
+      #   uses: slackapi/slack-github-action@007b2c3c751a190b6f0f040e47ed024deaa72844 # v1.23.0
+      #   if: ${{ steps.run.outcome == 'failure' && steps.run_retry.outcome != 'failure' }}
+      #   with:
+      #     channel-id: ${{ secrets.SLACK_BOUNDARY_TEST_BOT_CHANNEL_ID }}
+      #     payload: |
+      #       {
+      #         "text": ":warning: e2e tests passed, but needed retry (${{ matrix.filter }}): ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}\n*Branch:* ${{ github.event.ref }}\n*SHA:* <${{ github.event.head_commit.url }}|${{ github.event.after }}>"
+      #       }
+      #   env:
+      #     SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOUNDARY_TEST_BOT_TOKEN }}
```
moduli, line 323: - [ ] Revert before merge

.github/workflows/build.yml
```diff
@@ -389,11 +391,13 @@ jobs:
     needs:
       - set-product-version
       - product-metadata
-      - build-linux
+      - build-docker
```
moduli, line 394: - [ ] Revert before merge (will change this once we have the enos scenarios in place)

enos/modules/build_boundary_docker_local/build.sh
tmessi: I know we have the [docker-build-dev](https://github.com/hashicorp/boundary/blob/main/Makefile#L358-L368) target already. Might be able to update that to support passing through additional flags to `docker buildx`

moduli: 👍 I think I see some refactoring I can do with that makefile target. Giving it a shot

enos/modules/build_boundary_docker_local/build.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+set -eux -o pipefail
+
+env
```
johanbrandhorst, line 7: Is this necessary?

johanbrandhorst (approved):

PR #3104, opened 2023-03-20
chore: Remove extraneous fail-on-cache-miss attribute

moduli:
This PR removes an extraneous `fail-on-cache-miss` attribute. This particular job is the setup step, so there should be no expectation that the cache is present.


claire-labry (approved):

dekimsey (approved):
LGTM

PR #3105, opened 2023-03-20
chore: update go version to 1.20

jimlambrt:


johanbrandhorst (commented):
I'm not sure about this change. It means that the lowest version you can use to build any of these modules is 1.20. Since `github.com/hashicorp/boundary/api` is our official API, it means any Go API SDK users have to use at least Go 1.20, which has only been released for 2 months. I think we could afford to leave this at 1.19 until 1.21 is released in August.

jimlambrt:
@johanbrandhorst Are you concerned about bumping any of the modules or just the sdk/api?

jimlambrt:
I changed the api/sdk go version back to 1.19

johanbrandhorst (approved):

PR #3107, opened 2023-03-21
Bats Tests: Target Host Sources

elimt:
- Add bats test coverage for adding a Host Source to a Target
- Create `target_host_sources.bash` file with helper functions for bats test file

## Testing Scenarios

1. Add a newly created `host-set` as a `host-source` to a target
   i. Verify the newly created `host-source` is present in both JSON & Table format
1. Verify a target can have more than 1 `host-source` by adding an additional `host-source`
1. Verify `host-source` can be disassociated from a target 

## Considerations
Created new resources like a target & host catalog instead of using already created resources because the changes could affect existing tests and it made it easy to change the newly created resources

### DIFF
--- a/internal/tests/cli/boundary/_target_host_sources.bash
+++ b/internal/tests/cli/boundary/_target_host_sources.bash
@@ -28,7 +28,6 @@ function set_target_host_sources_sources() {
 }
 
 function validate_host_sources() {
-    echo $3
     targetData=$(boundary targets read -id $1 -format $3)
     if [[ "$3" == "json" ]]; then
         for i in "${@:2}"; do
--- a/internal/tests/cli/boundary/_target_host_sources.bash
+++ b/internal/tests/cli/boundary/_target_host_sources.bash
@@ -1,0 +1,51 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+load _authorized_actions
+
+function add_target_host_sources_sources() {
+    for i in "${@:2}"; do
+        hostSource+="-host-source $i "
+    done
+
+    boundary targets add-host-sources -id $1 $hostSource
+}
+
+function remove_target_host_sources_sources() {
+    for i in "${@:2}"; do
+        hostSource+="-host-source $i "
+    done
+
+    boundary targets remove-host-sources -id $1 $hostSource
+}
+
+function set_target_host_sources_sources() {
+    for i in "${@:2}"; do
+        hostSource+="-host-source $i "
+    done
+
+    boundary targets set-host-sources -id $1 $hostSource
+}
+
+function validate_host_sources() {
+    echo $3
+    targetData=$(boundary targets read -id $1 -format $3)
+    if [[ "$3" == "json" ]]; then
+        for i in "${@:2}"; do
+            if ! echo "$targetData" | jq ".item.host_sources[]"; then
+                echo "Host source id '$i' not found on target"
+                echo "$targetData"
+                exit 1
+            fi
+        done
+    elif [[ "$3" == "table" ]]; then
+        for i in "${@:2}"; do
+            pattern="Host Sources:.*ID:.*$2*"
+            OUTPUT=$(echo $targetData)
+            if ! [[ $OUTPUT =~ $pattern ]]; then
+                echo "Host source id '$i' not found on target"
+                exit 1
+            fi
+        done
+    fi
+}
--- a/internal/tests/cli/boundary/_target_host_sources.bash
+++ b/internal/tests/cli/boundary/_target_host_sources.bash
@@ -52,7 +52,7 @@ function target_has_host_source_id() {
     elif [[ "$format" == "table" ]]; then
         for i in "${hostSources}"; do
             pattern="Host Sources:.*ID:.*$i*"
-            OUTPUT=$(boundary targets read -id $1 -format table)
+            OUTPUT=$(boundary targets read -id $tid -format table)
             if ! [[ $OUTPUT =~ $pattern ]]; then
                 echo "Host source id '$i' not found on target"
                 return 1
--- a/internal/tests/cli/boundary/_target_host_sources.bash
+++ b/internal/tests/cli/boundary/_target_host_sources.bash
@@ -3,23 +3,7 @@
 
 load _authorized_actions
 
-function add_target_host_sources_sources() {
-    for i in "${@:2}"; do
-        hostSource+="-host-source $i "
-    done
-
-    boundary targets add-host-sources -id $1 $hostSource
-}
-
-function remove_target_host_sources_sources() {
-    for i in "${@:2}"; do
-        hostSource+="-host-source $i "
-    done
-
-    boundary targets remove-host-sources -id $1 $hostSource
-}
-
-function set_target_host_sources_sources() {
+function set_target_host_sources() {
     for i in "${@:2}"; do
         hostSource+="-host-source $i "
     done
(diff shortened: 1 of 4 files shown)


internal/tests/cli/boundary/_target_host_sources.bash
```diff
@@ -0,0 +1,50 @@
...
+
+function set_target_host_sources_sources() {
+    for i in "${@:2}"; do
+        hostSource+="-host-source $i "
+    done
+
+    boundary targets set-host-sources -id $1 $hostSource
+}
+
+function validate_host_sources() {
```
moduli, line 30: It was a little difficult to quickly understand what this method is trying to do

A lot of the methods in this test suite start off by defining what the input variables are, like...
```
local tid=$1
local hsid=$2
local format=$3
```
It might make sense to do that here?

internal/tests/cli/boundary/target_host_sources.bats
```diff
@@ -0,0 +1,214 @@
...
+    [ "$status" -eq 0 ]
+}
+
+@test "boundary/host-catalogs: can create $NEW_HOST_CATALOG host catalog in default project scope" {
+    run create_host_catalog $NEW_HOST_CATALOG $DEFAULT_P_ID
+    echo "$output"
+    [ "$status" -eq 0 ]
+}
+
+@test "boundary/hosts: can multiple hosts in created host catalog" {
```
moduli, line 36:
```suggestion
@test "boundary/hosts: can create multiple hosts in created host catalog" {
```

internal/tests/cli/boundary/target_host_sources.bats
```diff
@@ -0,0 +1,214 @@
...
+    run create_host $NEW_HOST1 $hcid
+    echo "$output"
+    [ "$status" -eq 0 ]
+
+    run create_host $NEW_HOST2 $hcid
+    echo "$output"
+    [ "$status" -eq 0 ]
+}
+
+@test "boundary/host-sets: can add multiple hosts to a created host catalog" {
```
moduli, line 47:
```suggestion
@test "boundary/host-sets: can add multiple host sets to a created host catalog" {
```

internal/tests/cli/boundary/target_host_sources.bats
```diff
@@ -0,0 +1,214 @@
...
+@test "boundary/target: can add created host set to created target" {
+    local hcid=$(host_catalog_id $NEW_HOST_CATALOG $DEFAULT_P_ID)
+    local hsid1=$(host_set_id $NEW_HOST_SET1 $hcid)
+    local tid=$(target_id_from_name $DEFAULT_P_ID $TGT_NAME)
+    run assoc_host_sources $tid $hsid1
+    echo "$output"
+    [ "$status" -eq 0 ]
+}
+
+@test "boundary/target: validate only $NEW_HOST1 host source present - JSON" {
```
moduli, line 99: is `only` true here? Is that method just verifying that the specified host source is attached the target, not that it's the ONLY host source? (The same question applies to the other tests that say "only")

internal/tests/cli/boundary/target_host_sources.bats
```diff
@@ -0,0 +1,214 @@
+#!/usr/bin/env bats
```
moduli, line 1: Is there some overlap with this test file and some the existing bats tests? I'm struggling to identify the differentiator here. Is it the validating of the host sources on the target? Could that just be added to `targets.bats`?

moduli (commented):
Appreciate the effort in trying to add test coverage. Happy to discuss any of my comments.

internal/tests/cli/boundary/_target_host_sources.bash
```diff
@@ -0,0 +1,50 @@
...
+    boundary targets remove-host-sources -id $1 $hostSource
+}
+
+function set_target_host_sources_sources() {
+    for i in "${@:2}"; do
+        hostSource+="-host-source $i "
+    done
+
+    boundary targets set-host-sources -id $1 $hostSource
+}
```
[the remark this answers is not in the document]
elimt: In the first iteration of my test I used those methods but I don't need them anymore. I'll take them out.

I did not realize we had the `set-host-sources` command. I'll add coverage for that

internal/tests/cli/boundary/_target_host_sources.bash
elimt: yeah, that's better. I'll set the format as a variable and group all the input variables at the top

internal/tests/cli/boundary/target_host_sources.bats
elimt: Looking at some of the tests that mention `only`, that isn't always true. I'll fix the wording

internal/tests/cli/boundary/target_host_sources.bats
elimt: I also realized some overlap, but this test file is similar to `target_credential_sources.bats` where the test focuses on host sources. It could be expanded to cover even more use cases. `targets.bats` may not be focused on all the possible use cases for `host-source`.

internal/tests/cli/boundary/_target_host_sources.bash
elimt: yeah, I like that. More descriptive

internal/tests/cli/boundary/_target_host_sources.bash
```diff
@@ -0,0 +1,54 @@
...
+    boundary targets add-host-sources -id $1 $hostSource
+}
+
+function remove_target_host_sources() {
+    for i in "${@:2}"; do
+        hostSource+="-host-source $i "
+    done
+
+    boundary targets remove-host-sources -id $1 $hostSource
+}
```
moduli, line 28: We may want to add variable for these arguments for clarity
```
local tid=$1
local hostSources...
```

internal/tests/cli/boundary/_target_host_sources.bash
```diff
@@ -0,0 +1,54 @@
...
+
+function target_has_host_source_id() {
+    local tid=$1
```
moduli, line 31: For clarity, I might suggest to also add variables for the input arguments here
```
local tid=$1
local hostSources...  # I'm not sure how to handle the array
local format=$3  # I think I preferred format being the last option
```

internal/tests/cli/boundary/_target_host_sources.bash
```diff
@@ -0,0 +1,54 @@
...
+    ids=$(boundary targets read -id $tid -format json | jq '.item.host_sources[].id')
```
moduli, line 32: This line is only used in the `json` block, right? Should that be moved into there?

internal/tests/cli/boundary/target.bats
```diff
@@ -69,7 +70,7 @@ load _helpers
 
 @test "boundary/target: admin user can add default host set to created target" {
   local id=$(target_id_from_name $DEFAULT_P_ID $TGT_NAME)
-  run assoc_host_sources $id $DEFAULT_HOST_SET
+  run add_target_host_sources $id $DEFAULT_HOST_SET
   echo "$output"
   [ "$status" -eq 0 ]
 }
```
moduli, line 76: The test below this one uses `target_has_host_source_id`. I'm a little surprised that it's working given the argument changes made to `target_has_host_source_id`. What's going on there?

internal/tests/cli/boundary/_target_host_sources.bash
```diff
@@ -0,0 +1,54 @@
...
+            done
+        done
+        return 1
+    elif [[ "$2" == "table" ]]; then
+        for i in "${@:3}"; do
+            pattern="Host Sources:.*ID:.*$i*"
+            OUTPUT=$(boundary targets read -id $1 -format table)
+            if ! [[ $OUTPUT =~ $pattern ]]; then
+                echo "Host source id '$i' not found on target"
+                exit 1
```
moduli, line 50: Should this be `return 1`?

internal/tests/cli/boundary/_target_host_sources.bash
elimt: It's trickier to have `hostSources` as the 2nd argument because it supports multiple host-sources. I've seen a pattern in other test suites to have the array at the end.
I can set `format` as a variable at the top

internal/tests/cli/boundary/target.bats
elimt: `target_has_host_source_id ` does not return 1 by default if it does not match any of the conditions so somehow it was just passing. I'll update it to return 1 and add a format here as well

internal/tests/cli/boundary/target.bats
```diff
@@ -4,6 +4,7 @@ load _auth
 load _connect
 load _targets
 load _helpers
+load _target_host_sources
```
moduli, line 7: Not sure if there is a convention for the order of these

internal/tests/cli/boundary/target_host_sources.bats
```diff
@@ -0,0 +1,239 @@
...
+    run create_host $NEW_HOST1 $hcid
+    echo "$output"
+    [ "$status" -eq 0 ]
+
+    run create_host $NEW_HOST2 $hcid
+    echo "$output"
+    [ "$status" -eq 0 ]
+}
+
+@test "boundary/host-sets: can create add multiple host sets to a created host catalog" {
```
moduli, line 47:
```suggestion
@test "boundary/host-sets: can create multiple host sets in a host catalog" {
```

internal/tests/cli/boundary/_target_host_sources.bash
```diff
@@ -0,0 +1,64 @@
...
+                if [ $(strip "$id") == "$hsid" ]; then
+                    return 0
+                fi
+            done
+        done
+        return 1
+    elif [[ "$format" == "table" ]]; then
+        for i in "${hostSources}"; do
+            pattern="Host Sources:.*ID:.*$i*"
+            OUTPUT=$(boundary targets read -id $1 -format table)
```
moduli, line 55:
```suggestion
            OUTPUT=$(boundary targets read -id $tid -format table)
```

moduli (approved):
Some minor comments

internal/tests/cli/boundary/target.bats
elimt: Just scanning through other `bats` files I have not noticed a convention nor any implications from switching the order for `targets.bats`

ddebko (approved):
LGTM

PR #3109, opened 2023-03-22
fix(handler): return human readable error for json cred w/ empty object

ddebko:
## Summary

When creating a JSON credential with an empty object, the return error is not a human readable message. 

Request:
```
cred, err := credClient.Create(tc.Context(), credential.JsonSubtype.String(), cs.Item.Id, credentials.WithName("foo"), credentials.WithJsonCredentialObject(map[string]any{}))
```

### Current Error Message:
```
{
  "kind":"Internal",
  "message":"credentials.(Service).createInRepo: unable to create credential: static.(Repository).CreateJsonCredential: missing object: parameter violation: error #100"
}
```

### Fixed Error Message:
```
{
  "kind":"InvalidArgument",
  "message":"Error in provided request.",
  "details":  {
      "request_fields":[{
        "name":"attributes.object",
        "description":"This is a required field and cannot be set to empty."
      }]
  }
}
```

### DIFF
--- a/internal/daemon/controller/handlers/credentials/credential_service.go
+++ b/internal/daemon/controller/handlers/credentials/credential_service.go
@@ -783,13 +783,9 @@ func validateCreateRequest(req *pbs.CreateCredentialRequest) error {
 		case credential.JsonSubtype.String():
 			object := req.GetItem().GetJsonAttributes().GetObject()
 			if object == nil || len(object.AsMap()) <= 0 {
-				badFields[objectField] = "Field required for creating a json credential."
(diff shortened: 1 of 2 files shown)


jefferai (changes requested):
This change looks good but if you look at the logic between update and create there are some important differences... can you sync them?

ddebko:
@jefferai I synced the logic between create & update. The only difference in the update function is the check on the maskField for the objects field.

jefferai (approved):

PR #3114, opened 2023-03-24
fix(schema-diff): Include foreign key constraints in diff

tmessi:
With the change to report changes to constraints instead of all of the 
post-data section, foreign key constraints where no longer captured in the
diff. This now extracts the foreign key constraints and generates a diff of
them between versions.

Also includes some additional documentation about this tool.

### DIFF
--- a/scripts/schema-diff.sh
+++ b/scripts/schema-diff.sh
@@ -2,7 +2,6 @@
 # Copyright (c) HashiCorp, Inc.
 # SPDX-License-Identifier: MPL-2.0
 
-
 die() {
     echo "$@"
     exit 255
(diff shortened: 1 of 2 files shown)


scripts/schema-diff.sh
```diff
@@ -1,7 +1,40 @@
 #!/usr/bin/env bash
 # Copyright (c) HashiCorp, Inc.
 # SPDX-License-Identifier: MPL-2.0
-
+#
+# This script can aide in comparing the database schema between two commit.
```
mgaffney, line 5:
```suggestion
# This script can aid in comparing the database schema between two commits.
```

mgaffney (approved):
Thanks!

scripts/schema-diff.sh
```diff
@@ -1,7 +1,40 @@
...
+#
+#   rm -r .schema-diff
+#
+# Limitations:
+#
+# Since this script is extracting specifc parts of the schema from
+# a database dump, there may be some differences that are not extracted and therefore
+# not reported. Some notable limitations follow:
+#
+# - domain defintiions: These are not easily extracted from a pg_dump. Therefore
```
hugoamvieira, line 36:
```suggestion
# - domain definitions: These are not easily extracted from a pg_dump. Therefore
```

hugoamvieira (approved):

PR #3115, opened 2023-03-24
test(e2e): Refactor WaitForSessionStatus methods

moduli:
This PR refactors some methods in the end-to-end test suite to reduce some code duplication, specifically the `WaitForSessionToBeActiveCli`, to be able to accept any desired state.

The PR does the following...
- refactors the `WaitForSessionToBeActiveCli` method to `WaitForSessionStatusCli`, which allows it to specify a sessionId and the desired state to wait for.
- moves away from string matching for statuses and uses the constants defined in code
- includes a protection against a panic in the bytes up and down tests to check if there are any connections in a session before trying to access it.


johanbrandhorst (approved):
Nice!

PR #3117, opened 2023-03-24
internal/db: add set_once_columns function

johanbrandhorst:
Also replace existing before_soft_delete_credential_vault_store function and trigger with the new function.

### DIFF
--- a/internal/db/schema/migrations/oss/postgres/66/02_use_set_once_columns.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/66/02_use_set_once_columns.up.sql
@@ -1,0 +1,12 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: MPL-2.0
+
+begin;
+
+  drop trigger before_soft_delete_credential_vault_store on credential_vault_store;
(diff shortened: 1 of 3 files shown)


internal/db/schema/migrations/oss/postgres/66/02_use_set_once_columns.up.sql
```diff
@@ -0,0 +1,12 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: MPL-2.0
+
+begin;
+
+  drop trigger before_soft_delete_credential_vault_store on credential_vault_store;
+  drop function before_soft_delete_credential_vault_store;
```
mgaffney, line 7: Can you add a comment in [10/04_vault_credential_up.sql](https://github.com/hashicorp/boundary/blob/034698318b760cdcf1648bd6b14f0319e501cb24/internal/db/schema/migrations/oss/postgres/10/04_vault_credential.up.sql#L70) that the function was replaced with a reference to this file?

mgaffney (approved):
Looks great! Just one small request. Thanks!

PR #3120, opened 2023-03-28
feat(ui): serve controller metadata from ui handler

A-440Hz:
this PR adds a noop function to the UI handler to allow controller metadata to be served for licensed Boundary editions

### DIFF
--- a/internal/daemon/controller/handler_ui.go
+++ b/internal/daemon/controller/handler_ui.go
@@ -17,6 +17,11 @@ func init() {
 	handleUi = handleUiWithAssets
 }
 
+// serveMetadata provides controller metadata to the UI for licensed versions of Boundary.
+var serveMetadata = noopServeMetadata
+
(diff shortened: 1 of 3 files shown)


internal/daemon/controller/handler_ui.go
```diff
@@ -17,6 +17,11 @@ func init() {
 	handleUi = handleUiWithAssets
 }
 
+// serveMetadata provides controller metadata to the UI for licensed versions of Boundary.
+var serveMetadata = noopServeMetadata
+
+func noopServeMetadata(w http.ResponseWriter) {}
```
jefferai, line 23: No particular reason to make this a separate func -- just inline it in the var

internal/db/sqltest/Makefile
hugoamvieira, line 1: Is this needed, or just a commit error?

A-440Hz: Thanks for pointing this out!

PR #3126, opened 2023-03-29
fix(cliui): Use correct git sha reference for artifacts

tmessi:
The git reference in github.sha is the latest commit on the default
branch (ie `main`). This instead used the commit hash for the workflow
that ran and triggered this workflow.

Blame: e76b3b95114ad696a89fe36f6b1bced4b090b9cf

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -404,3 +404,12 @@ jobs:
       docker-image-name: ${{ needs.build-docker.outputs.name }}
       docker-image-file: "boundary_default_linux_amd64_${{ needs.set-product-version.outputs.product-version }}_${{ github.sha }}.docker.dev.tar"
     secrets: inherit
+  bats:
+    uses: ./.github/workflows/test-cli-ui.yml
+    if: "! github.event.pull_request.head.repo.fork"
+    needs:
+      - set-product-version
+      - build-linux
+    with:
+      product-version: ${{ needs.set-product-version.outputs.product-version }}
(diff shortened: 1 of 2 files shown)


moduli (commented):
This should return a status to the PR checks, right? I'm not seeing it. I'm also trying to find the associated run in github actions. Is it this run? https://github.com/hashicorp/boundary/actions/runs/4557828644. Does it look like it's attached to the right commit?

tmessi:
> This should return a status to the PR checks, right? I'm not seeing it. I'm also trying to find the associated run in github actions. Is it this run? https://github.com/hashicorp/boundary/actions/runs/4557828644. Does it look like it's attached to the right commit?

I don't think it has run yet. It shouldn't run until the `build` workflow is done, which currently means waiting for enos to finish.

moduli:
Ooh right. build.yml includes enos 😅

.github/workflows/build.yml
```diff
@@ -404,3 +404,9 @@ jobs:
       docker-image-name: ${{ needs.build-docker.outputs.name }}
       docker-image-file: "boundary_default_linux_amd64_${{ needs.set-product-version.outputs.product-version }}_${{ github.sha }}.docker.dev.tar"
     secrets: inherit
+  bats:
+    uses: ./.github/workflows/test-cli-ui.yml
+    needs:
+      - set-product-version
```
moduli, line 410: Does this also need `build-linux`?

PR #3127, opened 2023-03-30
Backport of cleanup of github actions from CCI migration

tmessi:


tmessi:
@moduli can you double check the changes to the enos/testing workflows? 

@alvin-huang or really any one from rel-eng, can you double check the build.yml workflow?

.github/workflows/test-cli-ui.yml
```diff
@@ -0,0 +1,125 @@
...
+      - name: Install Vault
+        run: |
+          unzip /tmp/bats-cli-ui-deps/vault.zip -d /usr/local/bin
+      - name: Download Linux AMD64 Boundary bundle
+        uses: dawidd6/action-download-artifact@5e780fc7bbd0cac69fc73271ed86edf5dcb72d67 # v2.26.0
+        with:
+          workflow: build.yml
+          commit: ${{ github.event.workflow_run.head_sha }}
+          name: boundary_${{ inputs.product-version }}_linux_amd64.zip
+          path: /tmp
```
moduli, line 81: It looks like this is failing in the PR status checks? It can't find an artifact to download (it's looking for `boundary_0.12.2_linux_amd64.zip` but it found `boundary_0.12.1_linux_amd64.zip`). Not sure what's going on there. Is it looking for artifacts in the wrong branch or something?

This is what's used in `enos-run.yml`. Should we use that instead?
```
      - name: Download Linux AMD64 Boundary bundle
        id: download
        uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
        with:
          name: ${{ inputs.artifact-name }}
          path: ./enos/support/downloads
```

tmessi: Hmm, yea. I think that is wrong on `main`, let me fix it there and I'll cherry-pick it to this.

PR #3137, opened 2023-03-30
[QT-516] Add a step to the enos run workflow to automatically upload application logs when a scenario fails

mikebaum:
Recently the enos-provider was updated to add automatic application log file collection for all boundary and remote exec resources. This PR updates the enos run workflow to turn on this feature and to add an upload artifact step to store the logs in the build artifacts. The logs will be zipped up in a file that has the same name as the scenario that failed, with spaces being replaced by underscores.

### DIFF
--- a/.github/workflows/enos-run.yml
+++ b/.github/workflows/enos-run.yml
@@ -129,11 +129,13 @@ jobs:
         uses: hashicorp/action-setup-enos@v1    # TSCCR: loading action configs: failed to query HEAD reference: failed to get advertised references: authorization failed
         with:
           github-token: ${{ secrets.SERVICE_USER_GITHUB_TOKEN }}
-      - name: Set up AWS SSH private key
+      - name: Prepare scenario dependencies
+        id: prepare_scenario
         run: |
           mkdir -p ./enos/support
           echo "${{ secrets.SSH_KEY_PRIVATE_CI }}" > ./enos/support/private_key.pem
           chmod 600 ./enos/support/private_key.pem
+          echo "debug_data_artifact_name=enos-debug-data_$(echo ${{ matrix.scenario }} | sed -e 's/ /_/g' | sed -e 's/:/=/g')" >> $GITHUB_OUTPUT
       - name: Set up dependency cache
         id: dep-cache
         uses: actions/cache@69d9d449aced6a2ede0bc19182fadc3a0a42d2b0 # v3.2.6
@@ -222,7 +224,6 @@ jobs:
           ENOS_VAR_test_email: ${{ secrets.SERVICE_USER_EMAIL }}
           ENOS_VAR_boundary_docker_image_name: ${{ inputs.docker-image-name }}
           ENOS_VAR_boundary_docker_image_file: ./support/boundary_docker_image.tar
-          ENOS_DEBUG_DATA_ROOT_DIR: ./support/debug-data
         run: |
           mkdir -p ./enos/terraform-plugin-cache
           export ENOS_VAR_enos_user=$GITHUB_ACTOR && \
@@ -248,14 +249,6 @@ jobs:
           mkdir -p ./enos/terraform-plugin-cache
           export ENOS_VAR_enos_user=$GITHUB_ACTOR && \
           enos scenario run --timeout 60m0s --chdir ./enos ${{ matrix.filter }}
-      - name: Upload Debug Data
-        if: steps.run.outcome == 'failure'
-        uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
-        with:
-          # The name of the artifact is the same as the matrix scenario name with the spaces replaced with underscores.
-          name: ${${{ matrix.filter }}// /_}_run.zip
-          path: $ENOS_DEBUG_DATA_ROOT_DIR
-          retention-days: 30
       - name: Upload e2e tests output
         uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
         with:
@@ -283,9 +276,18 @@ jobs:
           ENOS_VAR_test_email: ${{ secrets.SERVICE_USER_EMAIL }}
           ENOS_VAR_boundary_docker_image_name: ${{ inputs.docker-image-name }}
           ENOS_VAR_boundary_docker_image_file: ./support/boundary_docker_image.tar
+          ENOS_DEBUG_DATA_ROOT_DIR: ./support/debug-data
         run: |
           export ENOS_VAR_enos_user=$GITHUB_ACTOR && \
           enos scenario run --timeout 60m0s --chdir ./enos ${{ matrix.filter }}
+      - name: Upload Debug Data
+        if: failure()
+        uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
+        with:
+          # The name of the artifact is the same as the matrix scenario name with the spaces replaced with underscores and colons replaced by equals.
+          name: ${{ steps.prepare_scenario.outputs.debug_data_artifact_name }}
+          path: ./enos/support/debug-data
+          retention-days: 30
       - name: Destroy Enos scenario
         if: ${{ always() && steps.run_retry.outcome == 'failure' }}
         env:
--- a/.github/workflows/enos-run.yml
+++ b/.github/workflows/enos-run.yml
@@ -80,6 +80,7 @@ jobs:
     runs-on: ${{ fromJSON(vars.RUNNER) }}
     env:
       GITHUB_TOKEN: ${{ secrets.SERVICE_USER_GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/enos-run.yml
```diff
@@ -222,6 +222,7 @@ jobs:
           ENOS_VAR_test_email: ${{ secrets.SERVICE_USER_EMAIL }}
           ENOS_VAR_boundary_docker_image_name: ${{ inputs.docker-image-name }}
           ENOS_VAR_boundary_docker_image_file: ./support/boundary_docker_image.tar
+          ENOS_DEBUG_DATA_ROOT_DIR: ./support/debug-data
```
moduli, line 225: How does enos create the data in this directory. Will this data get overwritten by the debug data in the following step?

.github/workflows/enos-run.yml
```diff
@@ -242,10 +243,19 @@ jobs:
...
         run: |
           mkdir -p ./enos/terraform-plugin-cache
           export ENOS_VAR_enos_user=$GITHUB_ACTOR && \
           enos scenario run --timeout 60m0s --chdir ./enos ${{ matrix.filter }}
+      - name: Upload Debug Data
+        if: steps.run.outcome == 'failure'
+        uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
+        with:
+          # The name of the artifact is the same as the matrix scenario name with the spaces replaced with underscores.
+          name: ${${{ matrix.filter }}// /_}_run.zip
```
moduli, line 256: Suggesting a new name to keep all of the enos/e2e test files grouped together. Either...
```
name: test-e2e-enos-${${{ matrix.filter }}// /_}_run.zip
name: enos-${${{ matrix.filter }}// /_}_run.zip
```

.github/workflows/enos-run.yml
mikebaum: The provider automatically will download the log files on failure. The log files will follow the pattern: `boundary_<host ip>.log`. If the scenario is run again the file will have a different host ip address so it will not be overwritten.

.github/workflows/enos-run.yml
mikebaum: Sounds good, I think I'll go with your second suggestion. I will need to re-work this PR a bit since we cannot assign a value in a workflow using a shell expression. We're working out the approach in the Vault repo first, then we'll just apply that strategy here.

.github/workflows/enos-run.yml
moduli: Right now, we're adding this variable **for each scenario** in these steps
- Output Terraform version info
- Run Enos scenario

Are there any conflicts there?

mikebaum: Not sure I quite understand you're comment. Clarifying a bit more the mechanism of how this works, may answer your question...

1. Application logs are only downloaded during a terraform plan or apply.
2. The feature is part of the enos Terraform provider, so logs are only downloaded if an enos resource fails, specifically (in the case of boundary), `enos_boundary_start` and any `enos_remote_exec`.
3. If during re-run, the same resource fails (or any other that accesses the same host),  the log files will be downloaded once again and will overwrite the existing downloaded log. In this case it is okay since presumably the new log file will fully contain the contents of the previously downloaded log file and any new information since the last time the logs were captured.

moduli: Ultimately, my question is why this new environment variable is being added to two spots and if it's necessary, how it works. The PR is adding this environment variable to two steps:
- `Output Terraform version info` which does an `enos scenario check`
- `Run Enos scenario` which does an `enos scenario run`
- Then, if `Run Enos Scenario` fails, upload the logs

So, do the logs from the `Output Terraform version info` provide anything? If that step fails, the logs don't get uploaded (logs only get uploaded if `run` fails). If that step passes but `run` fails, does the information from `output terraform version info` provide any additional information?

Happy to jump on a call to clarify further.

mikebaum: Ah, okay, now I understand your comment. Yeah there's no need to add the env var to the output step. It should only be added to the two steps where you are actually running the scenario. I will update. Good catch.

.github/workflows/enos-run.yml
```diff
@@ -283,9 +276,18 @@ jobs:
           ENOS_VAR_test_email: ${{ secrets.SERVICE_USER_EMAIL }}
           ENOS_VAR_boundary_docker_image_name: ${{ inputs.docker-image-name }}
           ENOS_VAR_boundary_docker_image_file: ./support/boundary_docker_image.tar
+          ENOS_DEBUG_DATA_ROOT_DIR: ./support/debug-data
```
jaymalasinha, line 279: `ENOS_DEBUG_DATA_ROOT_DIR: ./enos/support/debug-data`

.github/workflows/enos-run.yml
```diff
@@ -283,9 +276,18 @@ jobs:
...
         run: |
           export ENOS_VAR_enos_user=$GITHUB_ACTOR && \
           enos scenario run --timeout 60m0s --chdir ./enos ${{ matrix.filter }}
+      - name: Upload Debug Data
+        if: failure()
+        uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
+        with:
+          # The name of the artifact is the same as the matrix scenario name with the spaces replaced with underscores and colons replaced by equals.
+          name: ${{ steps.prepare_scenario.outputs.debug_data_artifact_name }}
+          path: ./enos/support/debug-data
```
jaymalasinha, line 289: we can use `${{ env. ENOS_DEBUG_DATA_ROOT_DIR }}` ?

jaymalasinha (commented):
Just couple comments, otherwise looks good

.github/workflows/enos-run.yml
mikebaum: That env var is only defined in the steps that actually run enos.

.github/workflows/enos-run.yml
```diff
@@ -273,9 +276,18 @@ jobs:
           ENOS_VAR_test_email: ${{ secrets.SERVICE_USER_EMAIL }}
           ENOS_VAR_boundary_docker_image_name: ${{ inputs.docker-image-name }}
           ENOS_VAR_boundary_docker_image_file: ./support/boundary_docker_image.tar
+          ENOS_DEBUG_DATA_ROOT_DIR: ./support/debug-data
         run: |
           export ENOS_VAR_enos_user=$GITHUB_ACTOR && \
           enos scenario run --timeout 60m0s --chdir ./enos ${{ matrix.filter }}
+      - name: Upload Debug Data
+        if: failure()
```
moduli, line 284: Sometimes, other parts of the workflow fail, which would unnecessarily trigger this. In other similar cases, I had to use something like
```
if: ${{ always() && steps.run_retry.outcome == 'failure' }}
```

.github/workflows/enos-run.yml
```diff
@@ -129,11 +129,13 @@ jobs:
...
         with:
           github-token: ${{ secrets.SERVICE_USER_GITHUB_TOKEN }}
-      - name: Set up AWS SSH private key
+      - name: Prepare scenario dependencies
+        id: prepare_scenario
         run: |
           mkdir -p ./enos/support
           echo "${{ secrets.SSH_KEY_PRIVATE_CI }}" > ./enos/support/private_key.pem
           chmod 600 ./enos/support/private_key.pem
+          echo "debug_data_artifact_name=enos-debug-data_$(echo ${{ matrix.scenario }} | sed -e 's/ /_/g' | sed -e 's/:/=/g')" >> $GITHUB_OUTPUT
```
moduli, line 138: There are some linter errors here
```
.github/workflows/enos-run.yml:134:9: shellcheck reported issue in this script: SC2086:info:4:121: Double quote to prevent globbing and word splitting [shellcheck]
    |
134 |         run: |
    |         ^~~~
.github/workflows/enos-run.yml:134:211: property "scenario" is not defined in object type {filter: string} [expression]
```

.github/workflows/enos-run.yml
jaymalasinha: since our variable is constant we can just define it at job level and use it.

mikebaum: > since our variable is constant we can just define it at job level and use it.

good call... I'll add it as a job level env var and then use your suggestion.

.github/workflows/enos-run.yml
mikebaum: good suggestion, I like this.

.github/workflows/enos-run.yml
mikebaum: Fixed. The linter actually caught one real issue... as I had copied this from the vault repo, where they use `matrix.scenario` rather than `matrix.filter` for the enos filter.

jaymalasinha (approved):
🥇 🚀

moduli (approved):

PR #3139, opened 2023-03-31
chore: Fix issues detected by GitHub Actions linter

moduli:
After installing the GitHub Actions VSCode extension, it reported a few linter errors on some of our workflow files. This PR attempts to address the majority of them. After these changes, the linter still has a couple of warnings about it being unsure if some variable access is correct, but I think those are working correctly.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -98,9 +98,9 @@ jobs:
         id: generate-metadata-file
         uses: hashicorp/actions-generate-metadata@v1  # TSCCR: loading action configs: failed to query HEAD reference: failed to get advertised references: authorization failed
         with:
+          repository: boundary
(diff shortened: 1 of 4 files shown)


moduli:
> What linter does the extension run? It would be nice if it could run in CI instead of just locally. We have [actionlint](https://github.com/hashicorp/boundary/blob/main/.github/workflows/actionlint.yml) set to run already, but since these issues were not found either the extension is running a different linter, or running it with different options.

Good question. I'll dig further.

johanbrandhorst (approved):

moduli:
@tmessi I'm having trouble finding out which linter is being used here...
- https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-github-actions&ssr=false
- https://github.com/github/vscode-github-actions

PR #3141, opened 2023-03-31
docs: Update Resource table IA

Dan-Heath:
The [Resource Table topic](https://developer.hashicorp.com/boundary/docs/concepts/security/permissions/resource-table) is difficult to read. There is too much information in the table, and the last column requires scrolling to see the whole thing.

As an MVP redesign of this topic, I broke each of the resource types out into its own table and gave it an H2 heading. I moved the applicable scopes information to the introduction of each table to reduce the number of columns. I also added a TOC at the top of the topic for internal linking.

I think these changes should improve readability and findability while we figure out what additional updates we might want to make to this topic.

[View the updated page in the preview deployment here](https://boundary-b42zqb14n-hashicorp.vercel.app/boundary/docs/concepts/security/permissions/resource-table).


randallmorey (approved):
Looks great!  Thank you.

stellarsquall (approved):

moduli:
There is a script that runs in CI to check if the existing Resources table needs to be updated. Because this PR is removing the table, the script is detecting that the table is missing and is reporting the `make-gen-delta` failure you see in the PR. 

If the intention is to remove the table, then I think we’ll have to do a few more changes:
- Remove perms-table from this line https://github.com/hashicorp/boundary/blob/main/Makefile#L136
- Cleanup
  - Remove https://github.com/hashicorp/boundary/blob/main/Makefile#L131-L133
  - Remove https://github.com/hashicorp/boundary/blob/main/internal/website/permstable/permstable.go

Dan-Heath:
Robin and I spoke to Jeff about the automation that creates this table. When Jeff has time, Robin and I will work with him to update the format of this topic, while preserving the automation and checks. I am closing this PR for now in favor of that work.

PR #3146, opened 2023-04-04
fix (workflows): use github ctx for repo name in legacy link checker

jimlambrt:


tmessi:
Can you add `".github/workflows/check-legacy-links-format.yml"` to the `paths` stanza?

ashleemboyer:
@jimlambrt that makes sense! Can you give me an example of one of those downstream repos so I can double-check that these changes will work with the `dev-portal` workflow being called? 😊

ashleemboyer:
@jimlambrt going to surface my comment from Slack here for posterity -- my recommendation is to disable this workflow for downstream repos, as it should only run on source content files in the main product repos. It can be disabled manually in the GitHub Actions tab, or this workflow can be ignored from being copied into downstream repos.

tmessi:
> @jimlambrt going to surface my comment from Slack here for posterity -- my recommendation is to disable this workflow for downstream repos, as it should only run on source content files in the main product repos. It can be disabled manually in the GitHub Actions tab, or this workflow can be ignored from being copied into downstream repos.

In that case I would recommend that we just add a conditional to the workflow, like we have for some others that should only run on the upstream, ie:

```yaml
jobs:
  check-links:
    if: github.repository == 'hashicorp/boundary'
    // rest of workflow
```

jimlambrt:
@ashleemboyer @tmessi mind a re-review?

tmessi (approved):

PR #3148, opened 2023-04-04
[QT-405] Create solo worker module

manthonygfp:
QT-405 is a request to add support for workers outside of the ones included with the default Boundary cluster module `terraform-enos-aws-boundary`.

This change adds a new module to this repo for provisioning discrete Boundary KMS workers. I can provide a simple scenario that exercises this at a reviewer's request, but did not include it in the PR because it doesn't test a quality requirement but instead simply turns on a cluster and adds a worker to it.

An updated `enos` from homebrew is required to run this. I use 0.0.18.
An updated `terraform` is required to support this. I use 1.4.4. The default homebrew install from Terraform is currently broken (if you have it fully updated. Make sure to install Terraform per the instructions in the TF documentation which includes using the HashiCorp tap.
An updated `enos-provider` is required to support this. At the time of this writing, this change is pending PR approval. You can download that branch and change the provider configuration in this repository to use a local reference to it or simply wait for the PR to be merged.


manthonygfp:
My changes the core Boundary cluster module in support of this PR has been released. I've updated the version required by Enos scenarios and am moving this PR out of draft status 🎉

enos/modules/solo_worker/templates/worker.hcl
```diff
@@ -0,0 +1,28 @@
+listener "tcp" {
+  purpose = "proxy"
+  tls_disable = true
+  address = "0.0.0.0"
+}
+
+worker {
+  # Name attr must be unique across workers
+  name = "worker-${id}"
```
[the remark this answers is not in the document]
manthonygfp: I might be misunderstanding, but this should always be unique to this worker because `id` comes from the `random_pet` generated in the solo_worker module.

```
    id                   = random_pet.worker.id
```

enos/modules/solo_worker/variables.tf
```diff
@@ -0,0 +1,131 @@
...
+variable "cluster_tag" {
+  description = "The cluster_tag from the Boundary cluster module"
+  type        = string
+}
+
+variable "worker_type_tags" {
+  description = "A list of tags to add in the worker's configuration file"
+  type        = list(string)
+  default     = ["prod", "webservers"]
+}
```
moduli, line 121: I'm trying to use this variable to insert my own tags here, but I'm not seeing it reflect when I try to examine the workers from boundary
```
boundary workers list
boundary workers read -id {WORKER_ID}
```

moduli: False alarm. Did not see the right worker in the list.

enos/modules/solo_worker/main.tf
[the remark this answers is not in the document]
manthonygfp: Sure :) With the planned change to the core Boundary module, it would save us from changing it later anyways.

enos/modules/solo_worker/templates/worker.hcl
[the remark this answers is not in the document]
moduli: 👍 Got it. Misread it. This is fine.

enos/modules/worker/templates/worker.hcl
```diff
@@ -0,0 +1,31 @@
...
+    type = ${type}
+  }
+}
+
+# must be same key as used on controller config
+kms "awskms" {
+  purpose    = "worker-auth"
+  region     = "${region}"
+  kms_key_id = "${kms_key_id}"
+}
```
moduli, line 31: I'm still a little confused about the pki worker type we're seeing in boundary. I did notice that this configuration file looks different than what's in the docs?

https://developer.hashicorp.com/boundary/docs/configuration/worker/kms-worker

enos/modules/worker/main.tf
moduli, line 1: Should we add `depends_on` attributes throughout to prevent any potential race conditions?

enos/modules/worker/templates/worker.hcl
moduli, line 1: I'm still trying to understand why these workers are appearing as pki workers...

moduli (commented):
Tested out the functionality. Looks like it's working pretty well!

enos/modules/worker/main.tf
manthonygfp: I'm pretty comfortable with the implicit deps and the explicit ones are already set for serial operations that don't have inter-related variables.

In a review of this, I did realize that `aws_vpc_security_group_ingress_rule.worker_to_controller` did need to exist _before_ `enos_boundary_start.worker_start`. It's extremely unlikely that they would ever occur out of order, but I've pushed up a change that makes the relationship explicit.

enos/modules/worker/templates/worker.hcl
manthonygfp: It is strange, but as the behavior occurs in the other core module, I'm not considering addressing it a part of this change.

moduli (approved):
Functionally, this appears to be working. Will rely on @joshbrand review on the AWS specific stuff.

enos/enos-scenario-e2e-database.hcl
```diff
@@ -46,6 +46,13 @@ scenario "e2e_database" {
     module = module.random_stringifier
   }
 
+  step "map2list" {
```
joshbrand, line 49: we can probably rename this to something more clear, "subnet_list" maybe?

enos/modules/map2list/main.tf
```diff
@@ -0,0 +1,5 @@
+variable "map" {}
```
joshbrand, line 1: context: we can't currently use built in functions (like `keys()`) in enos scenarios themselves, this is on the roadmap though

enos/modules/worker/main.tf
```diff
@@ -0,0 +1,204 @@
...
+resource "aws_vpc_security_group_ingress_rule" "worker_to_controller" {
+  description       = "This rule allows traffic from a worker to controllers over WAN"
+  security_group_id = var.controller_sg_id
+  cidr_ipv4         = "${aws_instance.worker.public_ip}/32"
+  from_port         = 9201
+  to_port           = 9201
+  ip_protocol       = "tcp"
+}
+
+data "aws_route_table" "default" {
```
joshbrand, line 117: Maybe a comment here, we're fetching the existing route tables from the `infra` module, so we can add a new route for our isolated subnet

enos/modules/target/main.tf
```diff
@@ -50,7 +44,7 @@ resource "aws_instance" "target" {
   ami                    = var.ami_id
   instance_type          = var.instance_type
   vpc_security_group_ids = [aws_security_group.boundary_target.id]
-  subnet_id              = tolist(data.aws_subnets.infra.ids)[count.index % length(data.aws_subnets.infra.ids)]
+  subnet_id              = var.subnet_ids[count.index % length(var.subnet_ids)]
```
moduli, line 47: Clarification question: I don't totally follow this. So, from the boundary cluster side, how come there are multiple subnet_ids? Why does it change based on the count index?

moduli (commented):
Scenarios pass in CI!

enos/modules/target/main.tf
joshbrand: This is to allow multiple targets, what we get from the infra module is a list of every subnet available to create in, and normally, we would want to distribute instances across them to test networking boundaries. So if you spun up 4 instances in an AZ that had 4 subnets available, we'd put once instance in each, instead of all 4 in one

PR #3149, opened 2023-04-04
test(e2e): Verify different credentials are brokered to target

stasryzhov:
[ICU-8561](https://hashicorp.atlassian.net/browse/ICU-8561)

This PR updates the e2e test `TestCliStaticCredentialStore` to do the following:

- Verify JSON-credentials are brokered to target correctly
- Verify private-key credentials are brokered to target correctly
- Use files and constants to store test data instead of hardcoding it

[ICU-8561]: https://hashicorp.atlassian.net/browse/ICU-8561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/testing/internal/e2e/tests/static/credential_store_test.go
+++ b/testing/internal/e2e/tests/static/credential_store_test.go
@@ -21,7 +21,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-const testCredentialsFile = "testdata/credential.json"
+const (
+	testCredentialsFile = "testdata/credential.json"
+	testPassword        = "password"
+)
 
 // TestCliStaticCredentialStore validates various credential-store operations using the cli
 func TestCliStaticCredentialStore(t *testing.T) {
@@ -49,7 +52,7 @@ func TestCliStaticCredentialStore(t *testing.T) {
 	// Create static credentials
 	newCredentialStoreId := boundary.CreateNewCredentialStoreStaticCli(t, ctx, newProjectId)
 	boundary.CreateNewStaticCredentialPrivateKeyCli(t, ctx, newCredentialStoreId, c.TargetSshUser, c.TargetSshKeyPath)
-	pwCredentialsId := boundary.CreateNewStaticCredentialPasswordCli(t, ctx, newCredentialStoreId, c.TargetSshUser, "password")
+	pwCredentialsId := boundary.CreateNewStaticCredentialPasswordCli(t, ctx, newCredentialStoreId, c.TargetSshUser, testPassword)
 	jsonCredentialsId := boundary.CreateNewStaticCredentialJsonCli(t, ctx, newCredentialStoreId, testCredentialsFile)
 
 	// Get credentials for target (expect empty)
@@ -86,7 +89,7 @@ func TestCliStaticCredentialStore(t *testing.T) {
 	require.NoError(t, err)
 
 	expectedCredentials := []map[string]any{
-		{"username": c.TargetSshUser, "password": "password"},
+		{"username": c.TargetSshUser, "password": testPassword},
 		expectedJsonCredentials,
 	}
 
--- a/testing/internal/e2e/tests/static/credential_store_test.go
+++ b/testing/internal/e2e/tests/static/credential_store_test.go
@@ -156,10 +156,7 @@ func createPrivateKeyPemFile(fileName string) error {
 	if err != nil {
 		return err
 	}
-	return savePrivateKeyToFile(fileName, key)
-}
 
-func savePrivateKeyToFile(fileName string, key *rsa.PrivateKey) error {
 	pemFile, err := os.Create(fileName)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


testing/internal/e2e/tests/static/credential_store_test.go
```diff
@@ -57,6 +62,8 @@ func TestCliStaticCredentialStore(t *testing.T) {
 	require.NoError(t, err)
 	require.True(t, newSessionAuthorizationResult.Item.Credentials == nil)
 
+	// Add credentials to target
```
moduli, line 65: Should we also add the private key credentials from line 51 to the target?

testing/internal/e2e/tests/static/credential_store_test.go
```diff
@@ -67,13 +74,17 @@ func TestCliStaticCredentialStore(t *testing.T) {
...
-	assert.Equal(t, c.TargetSshUser, retrievedUser)
-	assert.Equal(t, "password", retrievedPassword)
+	brokeredCredentials := make([]map[string]any, 0, 2)
+	for _, credential := range newSessionAuthorizationResult.Item.Credentials {
+		brokeredCredentials = append(brokeredCredentials, credential.Credential)
+	}
+
+	expectedCredentials := []map[string]any{
+		{"username": c.TargetSshUser, "password": "password"},
+		{"username": "name-json", "password": "password-json"},
```
moduli, line 84: Would it make sense to get this line from the JSON file as opposed to duplicating its contents?

testing/internal/e2e/tests/static/credential_store_test.go
```diff
@@ -67,13 +74,17 @@ func TestCliStaticCredentialStore(t *testing.T) {
...
-	require.True(t, ok)
-	assert.Equal(t, c.TargetSshUser, retrievedUser)
-	assert.Equal(t, "password", retrievedPassword)
+	brokeredCredentials := make([]map[string]any, 0, 2)
+	for _, credential := range newSessionAuthorizationResult.Item.Credentials {
+		brokeredCredentials = append(brokeredCredentials, credential.Credential)
+	}
+
+	expectedCredentials := []map[string]any{
+		{"username": c.TargetSshUser, "password": "password"},
```
moduli, line 83: Maybe we should set password as a variable

testing/internal/e2e/tests/static/credential_store_test.go
```diff
@@ -67,13 +79,29 @@ func TestCliStaticCredentialStore(t *testing.T) {
...
+	// Prepare expected credentials
+	testCredentialsJson, err := os.ReadFile(testCredentialsFile)
+	require.NoError(t, err)
+	var expectedJsonCredentials map[string]any
+	err = json.Unmarshal(testCredentialsJson, &expectedJsonCredentials)
+	require.NoError(t, err)
+
+	sshPrivateKeyFileContent, err := os.ReadFile(c.TargetSshKeyPath)
+	require.NoError(t, err)
+	sshPrivateKey := strings.TrimSuffix(string(sshPrivateKeyFileContent), "\n")
```
stasryzhov, line 96: Note: trimming the content here as read file adds a new-line symbol after the last line

stasryzhov:
Thank you for the comments @moduli 👍 The suggestions are implemented

testing/internal/e2e/tests/static/credential_store_test.go
moduli: Could we use `TrimSpace` instead? That way you don't need to specify the `\n`

testing/internal/e2e/tests/static/credential_store_test.go
```diff
@@ -67,13 +79,29 @@ func TestCliStaticCredentialStore(t *testing.T) {
...
+
+	expectedCredentials := []map[string]any{
+		{"username": c.TargetSshUser, "password": testPassword},
+		{"username": c.TargetSshUser, "private_key": sshPrivateKey},
+		expectedJsonCredentials,
+	}
+
+	assert.ElementsMatch(t, expectedCredentials, brokeredCredentials)
```
moduli, line 104: One thing I just thought about. If this assertion fails, this would print out the contents of the private key, would I don't think we want to do. Should we do something like generate a new key here as opposed to using the key in `c.TargetSshKeyPath`?

stasryzhov: That makes perfect sense. I've added a test .pem file generation and clean up for every test execution

testing/internal/e2e/tests/static/credential_store_test.go
```diff
@@ -111,6 +151,29 @@ func TestCliStaticCredentialStore(t *testing.T) {
...
+	}
+	defer pemFile.Close()
+
+	privateKey := &pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(key),
+	}
+
+	return pem.Encode(pemFile, privateKey)
+}
```
stasryzhov, line 175: Don't want to put it to a new package for now as it's used within this file only

moduli: I was a little confused about the organization of these methods. Does it make sense to just make it one method?

moduli (approved):
Overall, looks good. One more minor comment to consider.

PR #3151, opened 2023-04-04
(chore): Upgrade go version

A-440Hz:
Go version 1.19.8 contains a security fix


CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 Canonical reference for changes, improvements, and bugfixes for Boundary.
 
-## Next
+## 0.12.2 (2023/04/04)
```
louisruch, line 5: Fly by comment: I believe this should actually be `0.12.1` and we should add a new item for `0.12.2` and include a similar changelog entry as https://github.com/hashicorp/boundary/blob/main/CHANGELOG.md#security

johanbrandhorst (changes requested):
Please make the changelog updates Louis detailed, seems like we missed it last time.

CHANGELOG.md
```diff
@@ -2,7 +2,15 @@
...
+
+### Security
+
+* Boundary now uses Go 1.19.8 to address security vulnerability (CVE-2023-24536) See the
```
johanbrandhorst, line 9:
```suggestion
* Boundary now uses Go 1.19.8 to address CVE-2023-24536. See the
```

CHANGELOG.md
```diff
@@ -2,7 +2,15 @@
 
 Canonical reference for changes, improvements, and bugfixes for Boundary.
 
-## Next
+## 0.12.2 (2023/04/04)
```
johanbrandhorst, line 5: I think we probably want to add a ## Next section here as well, for next time.

johanbrandhorst (changes requested):

johanbrandhorst (approved):
LGTM

PR #3160, opened 2023-04-12
test(e2e): add test to verify user is logged out on its auth-token deletion

stasryzhov:
[ICU-8671](https://hashicorp.atlassian.net/browse/ICU-8671)

This PR creates new e2e test `TestUserIsLoggedOutWhenAuthTokenIsDeletedCli`. The test does the following: 
- Creates new account and assigns a new user to it
- Authenticates new user and assigns a name to its auth-token. The name is needed to be able to do operations via the user without having to authenticate it again (i.e. that would create a new auth-token)
- Verifies that the user is logged in (i.e. it can perform actions that require to be logged in)
- Deletes the user's auth-token by admin
- Verifies that the user is logged out (i.e. it cannot perform actions that require to be logged in)

Also, the PR refactors the `AuthenticateCli` function to take Options as a parameter. This allows to add additional args to the CLI command (e.g. `-token-name`)

[ICU-8671]: https://hashicorp.atlassian.net/browse/ICU-8671?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/testing/internal/e2e/boundary/authenticate.go
+++ b/testing/internal/e2e/boundary/authenticate.go
@@ -53,11 +53,12 @@ func AuthenticateAdminCli(t testing.TB, ctx context.Context) {
 }
 
 // AuthenticateCli uses the cli to authenticate the specified Boundary instance
-func AuthenticateCli(t testing.TB, ctx context.Context, authMethodId string, loginName string, password string) {
+func AuthenticateCli(t testing.TB, ctx context.Context, authMethodId string, loginName string, password string, opt ...e2e.Option) {
 	c, err := LoadConfig()
 	require.NoError(t, err)
 
-	output := e2e.RunCommand(ctx, "boundary",
+	var options []e2e.Option
+	options = append(options,
 		e2e.WithArgs(
 			"authenticate", "password",
 			"-addr", c.Address,
@@ -67,7 +68,11 @@ func AuthenticateCli(t testing.TB, ctx context.Context, authMethodId string, log
 		),
 		e2e.WithEnv("E2E_TEST_BOUNDARY_PASSWORD", password),
 	)
+	options = append(options, opt...)
+
+	output := e2e.RunCommand(ctx, "boundary", options...)
 	require.NoError(t, output.Err, string(output.Stderr))
+	t.Logf("Logged in as: %s", loginName)
 }
 
 // GetAuthenticationTokenCli uses the cli to get an auth token that can be used in subsequent
@@ -94,24 +99,4 @@ func GetAuthenticationTokenCli(t testing.TB, ctx context.Context, loginName stri
 	require.NoError(t, err)
 
 	return fmt.Sprint(authenticationResult.Item.Attributes["token"])
-}
-
-func GetAuthenticationTokenIDCli(t testing.TB, ctx context.Context, loginName string, password string) string {
-	//c, err := LoadConfig()
-	//require.NoError(t, err)
-
-	output := e2e.RunCommand(ctx, "boundary",
-		e2e.WithArgs(
-			"auth-tokens", "list",
-			"-token-name", loginName,
-			"-format", "json",
-		),
-	)
-	require.NoError(t, output.Err, string(output.Stderr))
-
-	var authenticationResult ListAuthTokensCliOutput
-	err := json.Unmarshal(output.Stdout, &authenticationResult)
-	require.NoError(t, err)
-
-	return fmt.Sprint(authenticationResult.Items[0].ID)
 }
--- a/testing/internal/e2e/boundary/authenticate.go
+++ b/testing/internal/e2e/boundary/authenticate.go
@@ -95,3 +95,23 @@ func GetAuthenticationTokenCli(t testing.TB, ctx context.Context, loginName stri
 
 	return fmt.Sprint(authenticationResult.Item.Attributes["token"])
 }
+
+func GetAuthenticationTokenIDCli(t testing.TB, ctx context.Context, loginName string, password string) string {
+	//c, err := LoadConfig()
+	//require.NoError(t, err)
+
+	output := e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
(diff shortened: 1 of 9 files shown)


testing/internal/e2e/boundary/authenticate.go
```diff
@@ -53,11 +53,12 @@ func AuthenticateAdminCli(t testing.TB, ctx context.Context) {
 }
 
 // AuthenticateCli uses the cli to authenticate the specified Boundary instance
-func AuthenticateCli(t testing.TB, ctx context.Context, authMethodId string, loginName string, password string) {
+func AuthenticateCli(t testing.TB, ctx context.Context, authMethodId string, loginName string, password string, opt ...e2e.Option) {
```
stasryzhov, line 56: Example of usage:
```go
// Authenticate user and assign a name to its auth token
boundary.AuthenticateCli(t, context.Background(), bc.AuthMethodId, testAccountName, acctPassword,
	e2e.WithArgs("-token-name", testAccountName),
)
```

testing/internal/e2e/tests/static/delete_auth_token_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// Check if user is logged out
+	output = e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"auth-tokens", "list",
+			"-token-name", testAccountName,
+		),
+	)
+	require.Error(t, output.Err, fmt.Sprintf("User '%s' is still logged in", testAccountName))
+	t.Logf("User is logged out: %s", testAccountName)
```
moduli, line 80: I wonder if there is a better way to validate this.

My original thought was to do something like this: https://github.com/hashicorp/boundary/blob/12aa7a0f5342d3569c9f189d73f4d7e54e5f9833/testing/internal/e2e/tests/static/session_end_delete_user_test.go#L61-L81 (use `GetAuthenticationTokenCli` to get you a token - I assume this creates an auth token that appears in `boundary auth-tokens list`). The admin would then delete that auth token, and then the test would try to use the token again with some other command, and that should fail.

moduli (changes requested):

testing/internal/e2e/tests/static/delete_auth_token_test.go
moduli: After discussion with Stan, the current is fine.

testing/internal/e2e/boundary/authtoken.go
```diff
@@ -0,0 +1,35 @@
...
+	"encoding/json"
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/boundary/api/authtokens"
+	"github.com/hashicorp/boundary/testing/internal/e2e"
+	"github.com/stretchr/testify/require"
+)
+
+// GetAuthenticationTokenIdCli uses the CLI to get an auth-token ID by its name
```
moduli, line 17: I think I was confused by this comment. To me, this reads as if you're looking up an auth-token ID using its name, but based on our earlier conversation, this is using the token name to authenticate the command.

Two things:
- How do we know we're retrieving the desired auth token here? The test is assuming this is the user's auth token? Is that the only one that is in the list?
- To keep things methods in this package consistent, should we not make `tokenName` required?

stasryzhov: After discussion with @moduli, we decided to keep current approach and rename the method for more clarity.
Also, I've updated the comment to provide more detailed description of what the method does

testing/internal/e2e/tests/static/auth_token_delete_test.go
```diff
@@ -0,0 +1,89 @@
...
+	ctx := context.Background()
+	boundary.AuthenticateAdminCli(t, ctx)
+	newAccountId, acctPassword := boundary.CreateNewAccountCli(t, ctx, bc.AuthMethodId, testAccountName)
+	t.Cleanup(func() {
+		boundary.AuthenticateAdminCli(t, context.Background())
+		output := e2e.RunCommand(ctx, "boundary",
+			e2e.WithArgs("accounts", "delete", "-id", newAccountId),
+		)
+		require.NoError(t, output.Err, string(output.Stderr))
+		t.Logf("Successfully deleted account '%s'", newAccountId)
```
moduli, line 38: Suggest to remove this log as it's in the cleanup and not the actual test

testing/internal/e2e/tests/static/auth_token_delete_test.go
```diff
@@ -0,0 +1,89 @@
...
+	})
+	newUserId := boundary.CreateNewUserCli(t, ctx, "global")
+	t.Cleanup(func() {
+		boundary.AuthenticateAdminCli(t, context.Background())
+		output := e2e.RunCommand(ctx, "boundary",
+			e2e.WithArgs("users", "delete", "-id", newUserId),
+		)
+		require.NoError(t, output.Err, string(output.Stderr))
+		t.Logf("Successfully deleted user '%s'", newUserId)
```
moduli, line 47: Suggest to remove this log as it's in the cleanup and not the actual test

testing/internal/e2e/tests/static/auth_token_delete_test.go
```diff
@@ -0,0 +1,89 @@
...
+	output = e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"auth-tokens", "list",
+			"-token-name", testAccountName,
+		),
+	)
+
+	// Expect error as the user got logged out
+	require.Error(t, output.Err, fmt.Sprintf("User '%s' is still logged in", testAccountName))
+	t.Logf("User is logged out: %s", testAccountName)
```
moduli, line 88:
```suggestion
	t.Logf("Successfully verified that token name is invalid")
```

moduli (approved):
Some additional minor comments about logging for consistency purposes.

PR #3163, opened 2023-04-14
refactor (events): QOL fix: turn off events for tests.

jimlambrt:
Quality of life fix: turn off events for tests and of course provide the ability to turn them back on if needed for specific tests.

### DIFF
--- a/internal/observability/event/context.go
+++ b/internal/observability/event/context.go
@@ -9,9 +9,14 @@ import (
 	"os"
 	"runtime"
 	"time"
+
+	"github.com/hashicorp/boundary/internal/test"
 )
 
-var ci bool
+var (
+	ci     bool
+	isTest bool
+)
 
 type key int
 
@@ -24,6 +29,7 @@ const (
 
 func init() {
 	ci = os.Getenv("CI") != ""
+	isTest = test.IsTestRun()
 }
 
 // NewEventerContext will return a context containing a value of the provided Eventer
@@ -136,7 +142,10 @@ func WriteError(ctx context.Context, caller Op, e error, opt ...Option) {
 	if !ok {
 		eventer = SysEventer()
 		if eventer == nil {
-			if !ci {
+			switch {
+			case !ci:
+			case !isTest:
+			default:
 				fallbackLogger().Error(fmt.Sprintf("%s: no eventer available to write error: %v", op, e))
 			}
 			return
--- a/internal/observability/event/eventer_config.go
+++ b/internal/observability/event/eventer_config.go
@@ -13,6 +13,7 @@ type EventerConfig struct {
 	ObservationsEnabled bool          `hcl:"observations_enabled"` // ObservationsEnabled specifies if observation events should be emitted.
 	SysEventsEnabled    bool          `hcl:"sysevents_enabled"`    // SysEventsEnabled specifies if sysevents should be emitted.
 	Sinks               []*SinkConfig `hcl:"-"`                    // Sinks are all the configured sinks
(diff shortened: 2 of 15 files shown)


internal/observability/event/context.go
```diff
@@ -136,7 +142,10 @@ func WriteError(ctx context.Context, caller Op, e error, opt ...Option) {
 	if !ok {
 		eventer = SysEventer()
 		if eventer == nil {
-			if !ci {
+			switch {
+			case !ci:
+			case !isTest:
```
johanbrandhorst, line 147: Nit: what do you think about
```suggestion
			case !ci, !isTest:
```

internal/observability/event/eventer_config.go
```diff
@@ -13,6 +13,7 @@ type EventerConfig struct {
 	ObservationsEnabled bool          `hcl:"observations_enabled"` // ObservationsEnabled specifies if observation events should be emitted.
 	SysEventsEnabled    bool          `hcl:"sysevents_enabled"`    // SysEventsEnabled specifies if sysevents should be emitted.
 	Sinks               []*SinkConfig `hcl:"-"`                    // Sinks are all the configured sinks
+	ErrorEventsDisabled bool          `hcl:"-"`                    // ErrorEventsDisabled will disable error events from being emitted.  This should only be used to turn off error events in tests.
```
johanbrandhorst, line 16: Why is this the inverse of the other options? I assume it's because we want a different default behavior, but I'd prefer if this was the same as the others and we just initialized it explicitly to the default. I worry with this inversion that someone could accidentally configure this or one of the others to the opposite of what they want. What do you think?

internal/test/test.go
```diff
@@ -0,0 +1,31 @@
...
+	"testing"
+)
+
+func init() {
+	testing.Init()
+	flag.Parse()
+}
+
+// IsTestRun will return whether or not we're operating within the context of a
+// test.
```
johanbrandhorst, line 17: It looks to me like this function actually checks whether some verbose test flag is set. Is that correct? I like the functionality, but I think the function is confusingly named and documented. Maybe `IsVerboseTestRun`?

Also, is the intent to override the behavior of tests configured to disable event logging if a verbose test run is requested? If so, would it make it impossible to run a verbose test run without all the events?

johanbrandhorst (commented):
This is such a welcome feature 🙏🏻. Thank you!

internal/observability/event/eventer_config.go
jimlambrt: Of course, I did this intentionally.  We don't want anyone configuring them and I wanted to increase the load on anyone trying to fool with setting error events.

internal/observability/event/context.go
jimlambrt: Of course, I did this intentionally as well since I find it easier to read.

internal/test/test.go
jimlambrt: pushed an amended commit.. is that better?

johanbrandhorst: This seems reasonable for now, but I noticed that Go 1.21 will include a function for this: https://github.com/golang/go/commit/7f38067acb738c43d870400dd648662d31456f5f. Maybe we could add a TODO to this comment to replace it with this new function when available?

jimlambrt: happy to.  ty!

jimlambrt:
results running locally. 

local-run-output

johanbrandhorst (approved):

jefferai:
Help I can't see events from tests and it's impossible to debug

PR #3169, opened 2023-04-20
fix (test): update how we determine when/if a test is running.

jimlambrt:
Previous solution caused some issues when using: go run <file> -h


johanbrandhorst (approved):
YOLO

tmessi (approved):
Thanks, I tested this out and it seems to fix the issue for me.

PR #3173, opened 2023-04-20
chore(e2e): Add support for variable binary names for end-to-end tests

joshbrand:
This PR updates enos workflows to add support for variable binary names in order to test various Boundary builds.

### DIFF
--- a/enos/enos-modules.hcl
+++ b/enos/enos-modules.hcl
@@ -11,7 +11,7 @@ module "bats_deps" {
 
 module "boundary" {
   source  = "app.terraform.io/hashicorp-qti/aws-boundary/enos"
-  version = ">= 0.5.0"
+  version = ">= 0.6.1"
 
   project_name = "qti-enos-boundary"
   environment  = var.environment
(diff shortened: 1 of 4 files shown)


enos/enos-variables.hcl
```diff
@@ -154,3 +154,15 @@ variable "docker_mirror" {
   type        = string
   default     = "docker.mirror.hashicorp.services"
 }
+
+variable "boundary_license" {
```
moduli, line 158: I assume this is a file path to a license file? Could we specify that somewhere so it's specific (whether it's the description of the variable name)

enos/enos-modules.hcl
```diff
@@ -22,8 +22,10 @@ module "boundary" {
     "Environment" : var.environment
   }
 
-  ssh_aws_keypair       = var.aws_ssh_keypair_name
```
moduli, line 25: Was this moved down to be sorted alphabetically? If so, then `boundary_binary_name` and `boundary_license` should be swapped?

moduli (commented):
Some minor comments

enos/enos-modules.hcl
joshbrand: spelling is hard!

enos/enos-variables.hcl
joshbrand: Actually it's the license string, I'll put in code to read it from a file for ease though!

moduli:
Closing for now. Rethinking approach.

PR #3177, opened 2023-04-21
test(e2e): Add test that uses new worker resource

moduli:
This PR adds a new e2e test that utilizes the new enos worker resource. The scenario sets up a worker and a target in a separate subnet from the controller. The test will then do the following...
- Set an incorrect egress worker filter and confirm that it cannot connect to the target
- Set the correct egress worker filer and confirm that it can connect to the target


testing/internal/e2e/tests/static_with_worker/worker_test.go
```diff
@@ -0,0 +1,99 @@
...
+	err = json.Unmarshal([]byte(c.WorkerTags), &workerTags)
+	require.NoError(t, err)
+	require.NotEmpty(t, workerTags)
+
+	t.Logf("Setting correct worker filter...")
+	output = e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"targets", "update", "tcp",
+			"-id", newTargetId,
+			"-egress-worker-filter", "\""+workerTags[0]+"\" in \"/tags/type\"",
```
stasryzhov, line 75: Optionally, we may get rid of the quotes shielding:
```suggestion
			"-egress-worker-filter", fmt.Sprintf(`"%s" in "/tags/type"`, workerTags[0]),
```

testing/internal/e2e/tests/static_with_worker/worker_test.go
```diff
@@ -0,0 +1,99 @@
...
+	newProjectId := boundary.CreateNewProjectCli(t, ctx, newOrgId)
+	newTargetId := boundary.CreateNewAddressTargetCli(t, ctx, newProjectId, c.TargetPort, c.TargetIp)
+
+	// Set incorrect worker filter, expect connection failure
+	t.Logf("Setting incorrect worker filter...")
+	output := e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"targets", "update", "tcp",
+			"-id", newTargetId,
+			"-egress-worker-filter", "\"prod\" in \"/tags/type\"",
```
stasryzhov, line 39: Same as above, optionally we may get rid of the shielding:
```suggestion
			"-egress-worker-filter", `"prod" in "/tags/type"`,
```

testing/internal/e2e/tests/static_with_worker/env_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+import "github.com/kelseyhightower/envconfig"
+
+type config struct {
+	TargetIp         string `envconfig:"E2E_TARGET_IP" required:"true"`    // e.g. 192.168.0.1
+	TargetSshKeyPath string `envconfig:"E2E_SSH_KEY_PATH" required:"true"` // e.g. /Users/username/key.pem
+	TargetSshUser    string `envconfig:"E2E_SSH_USER" required:"true"`     // e.g. ubuntu
+	TargetPort       string `envconfig:"E2E_SSH_PORT" required:"true"`     // e.g. 22
+	WorkerIp         string `envconfig:"E2E_WORKER_IP" required:"true"`    // e.g. 192.168.0.2
+	WorkerTags       string `envconfig:"E2E_WORKER_TAG" required:"true"`   // e.g.
```
stasryzhov, line 14: May we provide an example of the E2E_WORKER_TAG value or remove the comment?

testing/internal/e2e/tests/static_with_worker/worker_test.go
```diff
@@ -0,0 +1,99 @@
...
+			"-o", "UserKnownHostsFile=/dev/null",
+			"-o", "StrictHostKeyChecking=no",
+			"-o", "IdentitiesOnly=yes", // forces the use of the provided key
+			"-o", "ConnectTimeout=3",
+			"-p", "{{boundary.port}}", // this is provided by boundary
+			"{{boundary.ip}}",
+			"hostname", "-i",
+		),
+	)
+	require.Error(t, output.Err, string(output.Stderr))
```
johanbrandhorst, line 61: Do we want to check something more specific about this error? What does it look like?

johanbrandhorst (approved):
I agree with Stan's points, otherwise this looks good!

johanbrandhorst (approved):

stasryzhov (approved):
👍

moduli:
I'm thinking about refactoring the enos scenarios a little bit. I may move this to the "AWS" scenario so that...
- it will get run in CI
- trying to limit the amount of AWS resources we spin up if we add this particular scenario to CI

PR #3179, opened 2023-04-24
fix(perms): Allow for other grants' ids to change overall permissions

hugoamvieira:


internal/perms/acl_test.go
```diff
@@ -777,6 +777,56 @@ func TestACL_ListPermissions(t *testing.T) {
...
 			},
 		},
+		{
+			name: "List type + id read grants",
+			aclGrants: []scopeGrant{
+				{
+					scope: "o_1",
+					grants: []string{
+						"type=session;actions=list",
+						"id=s_2;type=session;actions=read",
```
tmessi, line 787: Can you maybe try a few more test cases?

A couple that come to mind, but maybe see if there are some others you can try:

```
type=session;actions=list
id=*;type=session;actions=read:self
id=s_1;type=session;actions=read
```


```
type=session;actions=list
id=s_1;type=session;actions=read
id=*;type=session;actions=cancel
```


```
type=target;actions=list
id=s_1;type=session;actions=read
```

hugoamvieira: Added these in with https://github.com/hashicorp/boundary/pull/3179/commits/5b1327f963cf2ec96dcec1fe1f6b49f6674ffde3. I also changed the code to hopefully read a little easier and explain what's going on better.

Also changed the tests to have the correct action set!

hugoamvieira:
Closing as problem is fixed with https://github.com/hashicorp/boundary/pull/3183

PR #3180, opened 2023-04-24
test(e2e): Clean up user if test fails

moduli:
When investigating a separate issue, noticed some logs showing that some test data was not cleaned up. Turns out, if the test `TestCliSessionEndWhenUserIsDeleted` fails prior to it deleting the user, then subsequent tests will get impacted.

This PR updates that test to have an additional cleanup step if the test isn't able to get to it.

### DIFF
--- a/testing/internal/e2e/tests/static/session_end_delete_user_test.go
+++ b/testing/internal/e2e/tests/static/session_end_delete_user_test.go
@@ -18,7 +18,7 @@ import (
 // TestCliSessionEndWhenUserIsDeleted tests that an active session is canceled when the respective
 // user who started the session is deleted.
 func TestCliSessionEndWhenUserIsDeleted(t *testing.T) {
-	deleted := false
+	userIsDeleted := false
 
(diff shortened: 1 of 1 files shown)


testing/internal/e2e/tests/static/session_end_delete_user_test.go
```diff
@@ -18,6 +18,8 @@ import (
 // TestCliSessionEndWhenUserIsDeleted tests that an active session is canceled when the respective
 // user who started the session is deleted.
 func TestCliSessionEndWhenUserIsDeleted(t *testing.T) {
+	deleted := false
```
stasryzhov, line 21: Although in this context the meaning of this variable is quite clear, we may want to be more explicit anyway, e.g.:
```suggestion
	isUserDeleted := false
```
or

```suggestion
	userIsDeleted := false
```

stasryzhov (approved):
Good catch! Left a minor suggestion that's up to you 👍

PR #3183, opened 2023-04-26
Fix ListPermissions when separate id/type grants are used

jefferai:
This makes two changes:

1. ListPermissions assumed every grant string had a type field, which does not match all grant formats. It now infers type from ID for formats which do not specify type.

2. Allow ID-only grant strings to be valid so long as the type specified matches the type of the specified ID.

### DIFF
--- a/internal/perms/acl_test.go
+++ b/internal/perms/acl_test.go
@@ -382,13 +382,14 @@ func Test_ACLAllowed(t *testing.T) {
 
 func TestACL_ListPermissions(t *testing.T) {
 	tests := []struct {
-		name           string
-		userId         string
-		aclGrants      []scopeGrant
-		scopes         map[string]*scopes.ScopeInfo // *scopes.ScopeInfo isn't used at the moment.
-		resourceType   resource.Type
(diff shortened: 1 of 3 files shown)


internal/perms/acl_test.go
```diff
@@ -777,6 +779,31 @@ func TestACL_ListPermissions(t *testing.T) {
...
+			name:         "separate_type_id_resource_grants",
+			scopes:       map[string]*scopes.ScopeInfo{"p_1": nil},
+			resourceType: resource.Target,
+			actionSet:    action.ActionSet{action.Read, action.Cancel},
+			aclGrants: []scopeGrant{
+				{
+					scope: "p_1",
+					grants: []string{
+						"type=target;actions=list",
+						"id=ttcp_1234567890;actions=read",
```
johanbrandhorst, line 792: I see the action part of this grant didn't make it into the permissions, is this an invalid grant? Should there not be a "read" action too?

johanbrandhorst (approved):

hugoamvieira (approved):

PR #3185, opened 2023-04-26
Update resource-table.mdx

xingluw:
Typo on table, "worker" does not have "s" in the grant string format.

### DIFF
--- a/internal/website/permstable/permstable.go
+++ b/internal/website/permstable/permstable.go
@@ -866,7 +866,7 @@ var worker = &Resource{
 		{
 			Path: "/workers",
 			Params: map[string]string{
-				"Type": "workers",
+				"Type": "worker",
 			},
 			Actions: append(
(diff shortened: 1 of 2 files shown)


jefferai:
This will be overridden on generation -- you need to update it in `permstable.go` and then run `make perms-table` to regenerate.

xingluw:
I'm not sure how to do that, but if it will be regenerated for the next release then we can close this PR and just wait

moduli:
@xingluw The table in `resource-table.mdx` is generated from some other source files, so your change here will get reverted after you merge this. To make the change permanent, you will need to update `permstable.go` on line 869 and 895 (changing `workers` to `worker`), and then run `make perms-table` to regenerate the table in `resource-table.mdx`. 

I have those changes locally and I can push them here if you would like.

moduli (approved):

PR #3189, opened 2023-04-27
chore(e2e): Update test attributes for new builds

joshbrand:
This is mostly scenario support for future pipeline work and some housekeeping cleanup. All new attributes have safe defaults won't break current tests. Also, updates the worker module to use new provider attribute for ipv4 public addressing and simplifes variables in the workflow.

### DIFF
--- a/.github/workflows/enos-run.yml
+++ b/.github/workflows/enos-run.yml
@@ -9,6 +9,9 @@ on:
       artifact-name:
         required: true
         type: string
+      edition:
+        required: true
+        type: string
       go-version:
         required: true
         type: string
@@ -223,6 +226,7 @@ jobs:
           ENOS_VAR_crt_bundle_path: ./support/boundary.zip
           ENOS_VAR_tfc_api_token: ${{ secrets.TF_API_TOKEN }}
           ENOS_VAR_test_email: ${{ secrets.SERVICE_USER_EMAIL }}
+          ENOS_VAR_boundary_edition: ${{ inputs.edition }}
           ENOS_VAR_boundary_docker_image_name: ${{ inputs.docker-image-name }}
           ENOS_VAR_boundary_docker_image_file: ./support/boundary_docker_image.tar
         run: |
@@ -243,6 +247,7 @@ jobs:
           ENOS_VAR_crt_bundle_path: ./support/boundary.zip
           ENOS_VAR_tfc_api_token: ${{ secrets.TF_API_TOKEN }}
           ENOS_VAR_test_email: ${{ secrets.SERVICE_USER_EMAIL }}
+          ENOS_VAR_boundary_edition: ${{ inputs.edition }}
           ENOS_VAR_boundary_docker_image_name: ${{ inputs.docker-image-name }}
           ENOS_VAR_boundary_docker_image_file: ./support/boundary_docker_image.tar
         run: |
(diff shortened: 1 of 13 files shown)


.github/workflows/enos-run.yml
```diff
@@ -80,7 +83,19 @@ jobs:
...
+      ENOS_VAR_aws_ssh_keypair_name: ${{ github.event.repository.name }}-ci-ssh-key
+      ENOS_VAR_aws_ssh_private_key_path: ./support/private_key.pem
+      ENOS_VAR_local_boundary_dir: ./support/boundary
+      ENOS_VAR_local_boundary_ui_dir: ./support/boundary-ui
+      ENOS_VAR_crt_bundle_path: ./support/boundary.zip
+      ENOS_VAR_tfc_api_token: ${{ secrets.TF_API_TOKEN }}
+      ENOS_VAR_test_email: ${{ secrets.SERVICE_USER_EMAIL }}
+      ENOS_VAR_boundary_edition: ${{ inputs.edition }}
+      ENOS_VAR_boundary_docker_image_name: ${{ inputs.docker-image-name }}
+      ENOS_VAR_boundary_docker_image_file: ./support/boundary_docker_image.tar
```
moduli, line 98: Nice

enos/modules/read_license/main.tf
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+variable "file_name" {}
+
+output "license" {
+  value = file(var.file_name)
```
moduli, line 7: How does this behave if the file_name is invalid/non-existent? It seems like the runs on this PR worked fine, but I was a little surprised by that.

enos/enos-scenario-e2e-static-with-vault.hcl
```diff
@@ -16,6 +16,7 @@ scenario "e2e_static_with_vault" {
   locals {
     aws_ssh_private_key_path = abspath(var.aws_ssh_private_key_path)
     boundary_install_dir     = abspath(var.boundary_install_dir)
+    boundary_license_path    = abspath(var.boundary_license_path != null ? var.boundary_license_path : joinpath(path.root, "./support/boundary.hclic"))
```
moduli, line 19: I was thinking about the `joinpath(path.root, "./support/boundary.hclic")` part here and was wondering if that was the appropriate thing to set here. I haven't thought through alternatives, though...

joshbrand: it's a default we use in vault for CI, but up for changes!

moduli: We can just leave it like this for now.

enos/modules/read_license/main.tf
joshbrand: it bails out quickly when planning the scenario:
```Error: Invalid function argument

  on ../../modules/read_license/main.tf line 4, in output "license":
   4:   value = file(var.file_name)
    ├────────────────
    │ while calling file(path)
    │ var.file_name is "/tmp/fakenotafile.hclic"
```

moduli: So this needs to be set in oss? EDIT: Just kidding. I missed the `skip_step` in the read license step.

moduli (approved):

PR #3199, opened 2023-05-03
.github: fix test matrix script

johanbrandhorst:
The test matrix script was excluding directories that only contained external package tests (i.e. tests packages with _test suffix). The following tests were not being run:

```
./internal/daemon/controller/common/scopeids
./internal/daemon/controller/handlers/authtokens
./internal/daemon/controller/handlers/groups
./internal/daemon/controller/handlers/host_catalogs
./internal/daemon/controller/handlers/host_sets
./internal/daemon/controller/handlers/hosts
./internal/daemon/controller/handlers/roles
./internal/daemon/controller/handlers/scopes
./internal/daemon/controller/handlers/sessions
./internal/daemon/controller/handlers/targets/tcp
./internal/daemon/controller/handlers/users
./internal/db/schema/internal/edition
./internal/db/schema/internal/provider
./internal/db/schema/migrations/oss
./internal/scheduler/cleaner
./internal/tests/api/accounts
./internal/tests/api/authmethods
./internal/tests/api/authtokens
./internal/tests/api/credentiallibraries
./internal/tests/api/credentials
./internal/tests/api/credentialstores
./internal/tests/api/groups
./internal/tests/api/hostcatalogs
./internal/tests/api/hosts
./internal/tests/api/hostsets
./internal/tests/api/managedgroups
./internal/tests/api/roles
./internal/tests/api/scopes
./internal/tests/api/sessions
./internal/tests/api/targets
./internal/tests/api/users
./internal/tests/api/workers
./internal/util
./testing/dbtest
./testing/internal/e2e/tests/aws
./testing/internal/e2e/tests/database
./testing/internal/e2e/tests/static
./testing/internal/e2e/tests/static_with_vault
```

PR #3200, opened 2023-05-03
ci(test): Remove matrix for running tests per package

tmessi:
The original motivation for the matrix was to mitigate an issue some
tests failing that used a vault docker container. When the tests ran in
a single step, these tests would almost always fail. By splitting the
tests per package, the failure rate of these tests was around 20%.

GitHub was able to identify the cause. There was an issue with localhost
resolution that causes requests to the vault containers to sometimes
fail, resulting in the kinds of failures that were seen. Note that the
same issue was not seen when accessing the test postgres database that
was running in a container, since the test code connects via 127.0.0.1,
while the vault container used localhost. This commit includes the
recommended fix so that localhost name resolution works as expected on
the GitHub hosted runners.

It also removes the test matrix since it is no longer needed as a work
around. In addition it seems advisable to remove the matrix given the
recent discovery that some tests were not running with the matrix. While
this was fixed, removing the matrix reduces the risk of any future
mistakes from the test splitting.

This also seems to result in a faster pipeline execution time. The
overhead to determine the matrix values, wait for a runner to be
available, and restore from the cache seems to add up to a longer
overall execution time.

Eliminating the matrix can also help with another type of failure seen.
Sometimes restoring from the cache fails. By eliminating the test
matrix, it will greatly reduce the number of cache restoring done by the
workflow, which will reduce the chances of an issue restoring from
cache.

The only remaining benefit from the matrix, was that if there was a
transient test failure, a re-run of only the failed jobs would be
relatively quick (depending on which package had the failure), while
without the matrix the entire test suite will need to be re-run.

See: https://github.com/hashicorp/boundary/pull/3199
Refs:
    86214a85e1c5553618b8f83592b3548e6b0aeb1f
    e04438b7f87f9811bc43038a3f31c3ab4c76f854

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -112,42 +112,10 @@ jobs:
         run: |
           make test-${{ matrix.module }}
 
-  set-test-package-matrix:
-    runs-on: ${{ fromJSON(vars.RUNNER) }}
-    needs:
-      - setup
-    outputs:
-      package-matrix: ${{ steps.set-matrix.outputs.matrix }}
-    steps:
-      - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
-      - name: Set up go
-        uses: actions/setup-go@4d34df0c2316fe8122ab82dc22947d607c0c91f9 # v4.0.0
-        with:
-          go-version: "${{ needs.setup.outputs.go-version }}"
-          cache: false
-      - name: Set up Go modules cache
-        uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
-        with:
-          path: |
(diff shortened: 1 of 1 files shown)


sarahethompson (approved):

moduli (commented):
Just checking if we still wanted to keep or get rid of `set-test-package-matrix.sh`

Glad we got a fix for the flaky test!

moduli:
> > Just checking if we still wanted to keep or get rid of `set-test-package-matrix.sh`
> > Glad we got a fix for the flaky test!
> 
> We could probably remove it, I left it around so people don't need to dig through the git history to find the script if we wanted to experiment with splitting in the future.

👍 No problem. I don't have a strong opinion about it. Just wanted to check.

johanbrandhorst (approved):

PR #3213, opened 2023-05-09
feat(e2e): Add configurable timeout for go test execution within scenarios

joshbrand:
The default `go test` timeout is 10 minutes, and we're coming awfully close to that limit in some test runs. This adds a tunable for that timeout we can customize per-execution if needed.

### DIFF
--- a/enos/enos-variables.hcl
+++ b/enos/enos-variables.hcl
@@ -172,3 +172,9 @@ variable "boundary_license_path" {
   type        = string
   default     = null
 }
+
+variable "test_timeout" {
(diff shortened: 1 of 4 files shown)


enos/enos-variables.hcl
```diff
@@ -172,3 +172,9 @@ variable "boundary_license_path" {
   type        = string
   default     = null
 }
+
+variable "test_timeout" {
```
moduli, line 176: Maybe consider renaming this to `go_test_timeout` to make it more specific

moduli (approved):

PR #3217, opened 2023-05-10
Add daily sessions pending view and accompanying snapshot table

mikemountain:


tmessi (changes requested):
Looking pretty good, but some test are failing due to some naming inconsistencies.

internal/db/schema/migrations/oss/postgres/70/01_hcp_billing_daily.up.sql
```diff
@@ -0,0 +1,64 @@
...
+      select date_trunc('day', session_pending_time), count(*)
+        from wh_session_accumulating_fact
+       where session_pending_time < date_trunc('day', timestamp 'today')
+    group by date_trunc('day', session_pending_time)
+  ),
+  daily_range (day) as (
+      select date_trunc('day', time)
+        from generate_series(
+                   (select min(session_pending_time) from wh_session_accumulating_fact),
+                   now(),
```
tmessi, line 48: I think this still results in `hcp_billing_daily_sessions_all` returning a row for today with a value of 0. I think if we want this view to be "the sum of pending sessions for yesterday and all previous days" like the comment says, then this needs to be:

```suggestion
                   timestamp 'yesterday',
```

tmessi (changes requested):

internal/db/schema/migrations/oss/postgres/70/01_hcp_billing_daily.up.sql
mikemountain: When replacing `now()` with `timestamp 'yesterday'`, the hcp_billing_daily_sessions_all view doesn't actually return up to yesterday's data; this is using timestamp 'yesterday':
```
select * from hcp_billing_daily_sessions_all;
         day         | sessions_pending_count 
---------------------+------------------------
 2023-05-15 00:00:00 |                      1
(1 row)
```

And this is using `now()`, which returns what's expected:
```
boundary=#   select * from hcp_billing_daily_sessions_all;
         day         | sessions_pending_count 
---------------------+------------------------
 2023-05-16 00:00:00 |                   1440
 2023-05-15 00:00:00 |                      1
 ```

Comment (tmessi) [reply internal/db/schema/migrations/oss/postgres/70/01_hcp_billing_daily.up.sql:48]: Ah ok, so when generating the series for the `daily_range` CTE, there are some subtle differences between `now()`, `timestamp 'today'` and `timestamp 'yesterday'` which depend on the starting point of the series `select min(session_pending_time) from wh_session_accumulating_fact`:

With `now()`:
```sql
select *
  from generate_series(
             (select min(session_pending_time) from wh_session_accumulating_fact),
             now(),
             '1 day'::interval
       ) as time;
             time
-------------------------------
 2023-05-11 11:35:32.191343+00
 2023-05-12 11:35:32.191343+00
 2023-05-13 11:35:32.191343+00
 2023-05-14 11:35:32.191343+00
 2023-05-15 11:35:32.191343+00
 2023-05-16 11:35:32.191343+00
 2023-05-17 11:35:32.191343+00
 2023-05-18 11:35:32.191343+00
(8 rows)
```

With `timestamp 'today'`:

```sql
select *
  from generate_series(
             (select min(session_pending_time) from wh_session_accumulating_fact),
             timestamp 'today',
             '1 day'::interval
       ) as time;
             time
-------------------------------
 2023-05-11 11:35:32.191343+00
 2023-05-12 11:35:32.191343+00
 2023-05-13 11:35:32.191343+00
 2023-05-14 11:35:32.191343+00
 2023-05-15 11:35:32.191343+00
 2023-05-16 11:35:32.191343+00
 2023-05-17 11:35:32.191343+00
(7 rows)
```

With `timestamp 'yesterday'`:
```sql
select *
  from generate_series(
             (select min(session_pending_time) from wh_session_accumulating_fact),
             timestamp 'yesterday',
             '1 day'::interval
       ) as time;
             time
-------------------------------
 2023-05-11 11:35:32.191343+00
 2023-05-12 11:35:32.191343+00
 2023-05-13 11:35:32.191343+00
 2023-05-14 11:35:32.191343+00
 2023-05-15 11:35:32.191343+00
 2023-05-16 11:35:32.191343+00
(6 rows)
```

```sql
select now() as now,
       timestamp 'today' as today,
       timestamp 'yesterday' as yesterday,
       (select min(session_pending_time) from wh_session_accumulating_fact) as start;
              now              |        today        |      yesterday      |             start
-------------------------------+---------------------+---------------------+-------------------------------
 2023-05-18 14:50:08.591431+00 | 2023-05-18 00:00:00 | 2023-05-17 00:00:00 | 2023-05-11 11:35:32.191343+00
(1 row)
```

So, since the time portion of start of the series is not at the start of a day (ie 00:00:00), depending on the time portion for the start date, and the current time, `now` and `timestamp 'today'` can yield different results, with `now` allowing for an additional value in the series if the current time is after the time for the start date. So using the examples above, if `now()` was > `2023-05-18 11:35.32.191343+00` it would yield the same as `timestamp 'today'`.

I think we might want to adjust how this series is generated so that the start of the series is truncated to the start of the day, so the `daily_range` CTE would become:

```sql
daily_range (day) as (
    select *
      from generate_series(
                 date_trunc('day', (select min(session_pending_time) from wh_session_accumulating_fact)),
                 timestamp 'yesterday',
                 '1 day'::interval
           ) as time
),
```

tmessi: I think this also impacts the hourly and possibly monthly _all views. cc @mgaffney

internal/db/schema/migrations/oss/postgres/70/01_hcp_billing_daily.up.sql
```diff
@@ -0,0 +1,64 @@
...
+  with
+  daily_counts (day, sessions_pending_count) as (
+      select date_trunc('day', session_pending_time), count(*)
+        from wh_session_accumulating_fact
+       where session_pending_time < date_trunc('day', timestamp 'today')
+    group by date_trunc('day', session_pending_time)
+  ),
+  daily_range (day) as (
+      select bucket
+      from generate_series(
```
tmessi, line 46:
```suggestion
      select bucket
        from generate_series(
```

internal/db/sqltest/tests/hcp/billing/daily_sessions_yesterday.sql
```diff
@@ -0,0 +1,75 @@
...
+    )
+insert into wh_session_accumulating_fact
+      (session_id, auth_token_id,
+       host_key, user_key, credential_group_key,
+       session_pending_date_key, session_pending_time_key, session_pending_time
+      )
+select session_id, auth_token_id,
+       host_key, user_key, credential_group_key,
+       session_pending_date_key, session_pending_time_key, session_pending_time
+from fake_sessions;
```
tmessi, line 51:
```suggestion
select session_id, auth_token_id,
       host_key, user_key, credential_group_key,
       session_pending_date_key, session_pending_time_key, session_pending_time
  from fake_sessions;
```

tmessi (approved):
Couple formatting nits, otherwise looks good

internal/db/schema/migrations/oss/postgres/70/01_hcp_billing_daily.up.sql
```diff
@@ -0,0 +1,64 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: MPL-2.0
+
+begin;
+  create table sessions_pending_daily_snapshot (
+    date date primary key,
```
mgaffney, line 6:
```suggestion
    snapshot_date date primary key,
```

mgaffney (changes requested):

mgaffney (approved):

PR #3218, opened 2023-05-10
fix(cli): correctly print secrets in authorize-session for current vault credential source subtypes

A-440Hz:
The current `authorize-session` print function has a case statement which checks whether the session credential type matches "vault" or "static" before accessing and printing the credential's SessionSecret field. The current credential libraries are generated with types "vault-generic" or "vault-ssh-certificate".

In my understanding, this case statement should be updated so we also print read and print credential secrets for all current vault credential source types.


hugoamvieira (commented):
I'd just like to hear others' thoughts on this to ensure that the way it was before wasn't by design.

I don't think it is, but probably worth checking just to make sure.

A-440Hz:
@hugoamvieira I asked @tmessi and he helped confirm this is a change we want to make

hugoamvieira (approved):

PR #3221, opened 2023-05-12
test(e2e): Add test that checks that session ends after deleting project

moduli:
This PR adds an e2e test that checks that a session is cancelled after deleting the corresponding project it belongs on. This continues the series of tests that test a session ends after other kinds of resources are deleted.

This particular test is slightly different in that it does not read the session id to check that the session status itself is cancelled and solely relying on the goroutine that initiates a user's session to end. This is due to how the data is structured (details here: https://hashicorp.atlassian.net/browse/ICU-7296)

### DIFF
--- a/testing/internal/e2e/tests/static/session_end_delete_project_test.go
+++ b/testing/internal/e2e/tests/static/session_end_delete_project_test.go
@@ -19,7 +19,7 @@ import (
 // project for the session is deleted.
 func TestCliSessionEndWhenProjectIsDeleted(t *testing.T) {
 	e2e.MaybeSkipTest(t)
-	c, err := loadConfig()
+	c, err := loadTestConfig()
 	require.NoError(t, err)
 	bc, err := boundary.LoadConfig()
 	require.NoError(t, err)
--- a/testing/internal/e2e/tests/static/session_end_delete_project_test.go
+++ b/testing/internal/e2e/tests/static/session_end_delete_project_test.go
@@ -34,12 +34,7 @@ func TestCliSessionEndWhenProjectIsDeleted(t *testing.T) {
 		require.NoError(t, output.Err, string(output.Stderr))
 	})
 	newProjectId := boundary.CreateNewProjectCli(t, ctx, newOrgId)
-	newHostCatalogId := boundary.CreateNewHostCatalogCli(t, ctx, newProjectId)
-	newHostSetId := boundary.CreateNewHostSetCli(t, ctx, newHostCatalogId)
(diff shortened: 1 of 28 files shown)


stasryzhov (commented):
Well done! I really like the approach with a separate goroutine 👍
Left only a little question about potential simplification

testing/internal/e2e/tests/static/session_end_delete_project_test.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+// TestCliSessionEndWhenProjectIsDeleted tests that an active session is canceled when the respective
+// project for the session is deleted.
+func TestCliSessionEndWhenProjectIsDeleted(t *testing.T) {
+	e2e.MaybeSkipTest(t)
+	c, err := loadConfig()
+	require.NoError(t, err)
+	bc, err := boundary.LoadConfig()
+	require.NoError(t, err)
```
johanbrandhorst, line 25: Forgive my ignorance, but what is the difference between these two? Maybe we should call the first one `loadTestConfig` in light of the existence of the second one?

testing/internal/e2e/tests/static/session_end_delete_project_test.go
```diff
@@ -0,0 +1,111 @@
...
+	output := e2e.RunCommand(ctx, "boundary", e2e.WithArgs("scopes", "delete", "-id", newProjectId))
+	require.NoError(t, output.Err, string(output.Stderr))
+
+	// Check if session has terminated
+	t.Log("Waiting for session to be canceling/terminated...")
+	select {
+	case output := <-errChan:
+		// `boundary connect` returns a 255 when cancelled
+		require.Equal(t, 255, output.ExitCode, string(output.Stdout), string(output.Stderr))
+	case <-time.After(time.Second * 5):
```
johanbrandhorst, line 106: Is this timeout long enough?

johanbrandhorst (approved):
Thanks for this!!

testing/internal/e2e/tests/static/session_end_delete_project_test.go
moduli: This is the same timeout as the other tests that do something similar and we haven't seen any flakiness yet.

testing/internal/e2e/tests/static/session_end_delete_project_test.go
moduli: Renamed for clarification

PR #3224, opened 2023-05-13
ci: adjust condition

chipitsine:
those jobs should not run in forks


modrake:
This logic is designed for separating out repos that share this workflow; our protection from fork-based runs is implemented in other layers of our workflows.

PR #3225, opened 2023-05-15
Add s3 bucket module for Enos scenarios

manthonygfp:
This Terraform module can be used with Enos to supply a bucket in S3 along with a policy that allows a given user to write and delete objects in said bucket. I've done some simple testing to confirm that the above behavior is possible. A simple way to test this for yourself is to use a scenario like the one below, grab the outputs from the scenario to know the bucket name and AWS user ID (in order to assume this user's access). You'll need to get the key for this user from the state file itself as it's a sensitive value that can't be shown in Terraform output.

```
module "bucket" {
  source = "./modules/bucket"
}

scenario "bucket_test" {
  terraform_cli = terraform_cli.default
  terraform     = terraform.default
  providers     = [provider.aws.default]

  step "create_iam_user" {
    module = module.iam_setup

    variables {
      test_id    = "<TODO PUT AN ID FOR YOURSELF HERE>"
      test_email = "<TODO PUT AN EMAIL FOR YOURSELF HERE>"
    }
  }

  step "create_bucket" {
    module = module.bucket

    variables {
      cluster_tag = "TODO MAKE AN ARBITRARY CLUSTER TAG HERE"
      user        = step.create_iam_user.user_name
    }
  }

  output "bucket_name" {
    value = step.create_bucket.bucket_name
  }

  output "access_key_id" {
    value = step.create_iam_user.access_key_id
  }
}
```

If you've pulled down the branch to help me with testing, please note that I've force pushed after rebasing the module to bring the base up to date with main.

### DIFF
--- a/enos/ci/service-user-iam/main.tf
+++ b/enos/ci/service-user-iam/main.tf
@@ -216,7 +216,10 @@ data "aws_iam_policy_document" "enos_policy_document" {
       "rds:ListTagsForResource",
       "rds:ModifyDBInstance",
       "rds:ModifyDBSubnetGroup",
-      "rds:RemoveTagsFromResource"
+      "rds:RemoveTagsFromResource",
+      "s3:ListAllMyBuckets",
+      "s3:CreateBucket",
+      "s3:DeleteBucket",
     ]
     resources = ["*"]
   }
@@ -243,7 +246,9 @@ data "aws_iam_policy_document" "aws_nuke_policy_document" {
       "iam:ListUserTags",
       "iam:ListUsers",
       "iam:UntagUser",
-      "servicequotas:ListServiceQuotas"
+      "servicequotas:ListServiceQuotas",
+      "s3:ListAllMyBuckets",
+      "s3:DeleteBucket",
     ]
     resources = ["*"]
   }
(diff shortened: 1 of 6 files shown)


enos/ci/service-user-iam/main.tf
moduli, line 1: Question for my understanding: How are these two policies used? Why were these specific permissions granted in these different policies?

moduli (approved):
Looks good! Just one comment to help my understanding.

enos/ci/service-user-iam/main.tf
joshbrand: These are user policies (as opposed to the bucket policy below) The first policy is what is used to grant permissions to the static CI user that runs enos, so everything required to bring up, use, and tear down resources. The second is purely for `aws-nuke` our cleanup process, it requires a few more accounting resources to be happy. In the future, if you wanted to add, say, Aurora, or Elasticache for some reason, you'd need to update these again.

enos/modules/bucket/main.tf
```diff
@@ -0,0 +1,33 @@
...
+    ]
+
+    resources = [
+      "${aws_s3_bucket.default.arn}/*",
+    ]
+  }
+}
+
+resource "aws_iam_user_policy" "default" {
+  name   = "${aws_s3_bucket.default.id}_${var.user}_access"
```
joshbrand, line 30: NICE

joshbrand (approved):
looks good!

enos/ci/service-user-iam/main.tf
moduli: Thanks for the explanation. That makes sense.

Just checking, is `ListAllMyBuckets` required?

PR #3231, opened 2023-05-17
Update nodeenrollment to fix TLS issue; add debug

jefferai:
Add more logging to worker auth rotation and test flags. This
makes it much easier to figure out when and why rotation is
happening and control it for tests/debugging.

This also updates nodeenrollment to the `v0.1.18+tlsfix` tag.


johanbrandhorst (approved):

jimlambrt (approved):
perhaps check the failing test?

PR #3232, opened 2023-05-18
style(cli): boundary connect output returns pasteable X.509 certificates

A-440Hz:
When a target accesses a session credential from a credential library with a vault path pointing to a .crt certificate, this credential is treated as an unspecified credential (which isn't username-password, ssh-private-key, or json type). Instead of being unmarshalled into a struct, the raw secret is decoded into a []byte by `base64.StdEncoding.DecodeString`, and then sent to a `bytes.Buffer` by `json.Indent` before being converted back into a string. This results in the certificate being returned in the CLI output as a json value containing `\n` characters like so (and in my environment there's some `\r`s as well):

image

I had thought that the newline characters were a [valid format](https://serverfault.com/questions/466683/can-an-ssl-certificate-be-on-a-single-line-in-a-file-no-line-breaks) for a crt file , but I was told by the reporter that this format rendered the certificate useless for their [kubernetes workflow](https://www.hashicorp.com/blog/how-to-connect-to-kubernetes-clusters-using-boundary). Furthermore, the newline between "BEGIN" and "CERTIFICATE" and the tabs before each "CERTIFICATE" makes the output bad to copy/paste.

I understand that setting -format=json fixes the latter issue. 

This PR attempts to unmarshal and print all unspecified credential values without indentation when `-format=table`, so the certificates can be displayed like so in a nice-looking block.

image

### DIFF
--- a/internal/cmd/commands/connect/funcs.go
+++ b/internal/cmd/commands/connect/funcs.go
@@ -120,7 +120,7 @@ func fmtSecretForTable(indent int, sc *targets.SessionCredential) []string {
 			certs := make(map[string]any, len(iface))
 			for kk, vv := range iface {
 				if p, _ := pem.Decode([]byte(fmt.Sprintf("%v", vv))); p != nil {
-					// iface is a certificate. Print without indents.
+					// If the value is PEM formatted, add it to a map to be formatted without indents.
 					certs[kk] = vv
 				}
 			}
--- a/internal/cmd/commands/connect/funcs.go
+++ b/internal/cmd/commands/connect/funcs.go
@@ -4,7 +4,6 @@
 package connect
 
 import (
-	"bytes"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
@@ -13,6 +12,7 @@ import (
 
 	"github.com/hashicorp/boundary/api/targets"
 	"github.com/hashicorp/boundary/internal/cmd/base"
+	"github.com/mitchellh/mapstructure"
 )
 
 func generateSessionInfoTableOutput(in SessionInfo) string {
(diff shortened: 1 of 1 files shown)


A-440Hz:
image

The switch cases currently assume the PEM certificate will be located in the first json layer of a vault kv store i.e., `data: <key_name>: <.crt here>`. If it is possible for the certificate to be located in a different location i.e. `data: <key_name>: <another_key_name>: <.crt here>`, then the output will not display properly.

If it is possible for an 'unspecified' Boundary credential to come from a different vault path with a different expected format, then it might not be displayed properly either. Do you know what possible 'unspecified' Boundary credential formats I could expect to find, or who I can ask about this topic?

internal/cmd/commands/connect/funcs.go
```diff
@@ -99,15 +99,46 @@ func fmtSecretForTable(indent int, sc *targets.SessionCredential) []string {
...
 		return origSecret
 	}
-	secretStr := strings.Split(dst.String(), "\n")
-	if len(secretStr) > 0 {
-		secretStr[0] = fmt.Sprintf("%s    %s", prefixStr, secretStr[0])
+	for k, iface := range baseCredMap {
+		switch iface := iface.(type) {
+		case nil:
+			out = append(out, fmt.Sprintf("%s    %s:", prefixStr, k))
+		case map[string]any:
```
A-440Hz, line 116: https://github.com/hashicorp/boundary/pull/3232#issuecomment-1552248076

internal/cmd/commands/connect/funcs.go
```diff
@@ -99,15 +100,60 @@ func fmtSecretForTable(indent int, sc *targets.SessionCredential) []string {
...
+			// TODO: check if this is the only possible case for an 'unspecified' type credential
+			// and get rid of the other cases if this is so.
+			out = append(out, fmt.Sprintf("%s    %s:", prefixStr, k))
+			certs := make(map[string]any, len(iface))
```
johanbrandhorst, line 121: Nit: is there a good reason to bother setting the size of the map here? It seems to me we don't really know how big it's going to be.

internal/cmd/commands/connect/funcs.go
```diff
@@ -99,15 +100,60 @@ func fmtSecretForTable(indent int, sc *targets.SessionCredential) []string {
...
+			for kk, vv := range iface {
+				if p, _ := pem.Decode([]byte(fmt.Sprintf("%v", vv))); p != nil {
```
johanbrandhorst, line 123: I'd sooner we have another type assertion than this string formatting, to make it a bit more predictable
```go
switch iface := iface.(type) {
case []byte:
	if p, _ := pem.Decode(iface); p != nil {
```
etc. Maybe one for `string` too. What do you think?

johanbrandhorst (commented):
This will be a great UX feature. I wonder, can we add an e2e test that ensures the behavior? Maybe a test that checks that we can successfully parse a certificate from the output?

internal/cmd/commands/connect/funcs.go
A-440Hz: would you expect a case where the unmarshalled interface is anything other than a []byte or a string? What would you recommend I do in that scenario? Does it sound reasonable to return an error in that case?

jefferai:
Why 0.14 and not 0.13?

PR #3238, opened 2023-05-22
doc: Add missing copyright headers

moduli:
This PR adds copyright headers to some new files that were introduced.


jefferai (approved):
Thank you!!!

PR #3241, opened 2023-05-24
Fix test timing

jefferai:


internal/daemon/worker/testing.go
```diff
@@ -445,6 +445,11 @@ func NewTestMultihopWorkers(t testing.TB,
 		WorkerAuthDebuggingEnabled: enableAuthDebugging,
 		DownstreamWorkerAuthKms:    childDownstreamWrapper,
 	})
+	t.Cleanup(kmsWorker.Shutdown)
+
+	// Give time for it to be inserted into the database
+	time.Sleep(2 * time.Second)
+
```
moduli, line 452: I think make-gen-delta is complaining about this extra line here

jefferai: that's what I get for porting the patch over in vim

internal/daemon/worker/testing.go
```diff
@@ -532,9 +537,10 @@ func NewTestMultihopWorkers(t testing.TB,
 		WorkerAuthKms:              childDownstreamWrapper2,
 		WorkerAuthDebuggingEnabled: enableAuthDebugging,
 	})
+	t.Cleanup(childKmsWorker.Shutdown)
 
 	// Sleep so that workers can startup and connect.
-	time.Sleep(10 * time.Second)
+	time.Sleep(12 * time.Second)
```
moduli, line 543: Out of curiosity, I was wondering if there are other ways to check if things are connected rather than using sleeps. Is there a way to check the worker's status and verify that it's connected?

moduli (approved):

internal/daemon/worker/testing.go
jefferai: I had that thought too but (a) don't want to spend the time now and (b) it's not clear to me for all dependent tests what metric would be correct. On some ent tests for instance "connected" can mean something very different. So probably it's best up to the caller of this function to do that verification.

PR #3243, opened 2023-05-25
Fix worker auth rotation test

jefferai:
Most of the changes here are changes or enhancements that are useful for tests. The crux of the issue was that I had pulled up one var to the package level so it could be read from tests, but that var was not being reset between tests. It's not even used in tests anymore, so I put it back to function-local.


jefferai:
Note: GHA has been super flaky all night and this test has not failed on previous runs, nor does it fail locally. It's a context deadline exceeded so I'm chalking it up to the same load issues causing all sorts of problems tonight, and merging.

PR #3248, opened 2023-05-26
feat: license util config

mikemountain:


johanbrandhorst (approved):

jefferai:
For Boundary it makes more sense for this to be in the Controller block...is it mandated or standardized that it lives at the top level?

jefferai (approved):

internal/daemon/controller/controller.go
```diff
@@ -11,6 +11,8 @@ import (
 	"sync"
 	"sync/atomic"
 
+	"github.com/hashicorp/boundary/internal/census"
+
 	"github.com/hashicorp/boundary/internal/auth/ldap"
```
johanbrandhorst, line 16: Nit: import groups

johanbrandhorst (changes requested):

johanbrandhorst (approved):

PR #3249, opened 2023-05-30
Allow slashes in authorize-session when using target name

jefferai:
This requires specifying `**` as the selector in the proto, which will match any segments and simply requires that the verb is the last part of the URL, which it already is.


internal/proto/controller/api/services/v1/target_service.proto
```diff
@@ -77,7 +77,7 @@ service TargetService {
   // AuthorizeSession creates authorization information from a given Target.
   rpc AuthorizeSession(AuthorizeSessionRequest) returns (AuthorizeSessionResponse) {
     option (google.api.http) = {
-      post: "/v1/targets/{id}:authorize-session"
+      post: "/v1/targets/{id=**}:authorize-session"
```
jimlambrt, line 80: Mind adding a comment here about the use of `**`

jimlambrt (approved):
One small suggestion.

internal/proto/controller/api/services/v1/target_service.proto
jefferai: Done.

PR #3250, opened 2023-05-30
Update changelog

jefferai:


jefferai:
Going to merge as it fixes the delta change too, can always update the text after the fact.

PR #3257, opened 2023-06-01
Upgrade grpc-gateway dependency

johanbrandhorst:
## [proto: correct classification of authz token](https://github.com/hashicorp/boundary/commit/15cbe83966a045a0f07c4fe51293c96d964075f9)

The target authorization token had the invalid classification
"private". Fortunately, classification is a fail-closed system,
so this won't have exposed users authorization tokens, but
we should still use the correct classification. Use "secret".

## [scripts: add Go script to remove gotags comments](https://github.com/hashicorp/boundary/commit/e5edc4ad9d6f8f37ad680ab7a32ad1965286f234)

The new version of grpc-gateway includes our gotags
comments in the generate swagger documentation. This
script removes all mentions of these.

## [Upgrade grpc-gateway dependency](https://github.com/hashicorp/boundary/commit/55000438d5ed40785edd315d172e6944530666be)

Mainly we need this to fix generation of our docs side API docs,
but it also includes a nice fix for the new Download RPC, i.e. it
now contains the correct response schema (a binary stream).


internal/gen/controller.swagger.json
```diff
@@ -89,12 +89,14 @@
         "parameters": [
           {
             "name": "auth_method_id",
+            "description": "",
```
johanbrandhorst, line 92: These empty descriptions should be functionally equivalent to what we had before

internal/gen/controller.swagger.json
```diff
@@ -202,12 +207,6 @@
             "schema": {
               "$ref": "#/definitions/controller.api.resources.accounts.v1.Account"
             }
-          },
-          {
-            "name": "update_mask",
-            "in": "query",
-            "required": false,
-            "type": "string"
```
johanbrandhorst, line 210: Note that this parameter was never used or expected before - this is a bug fix!

internal/gen/controller.swagger.json
```diff
@@ -2817,16 +2847,10 @@
...
-                },
-                "error": {
-                  "$ref": "#/definitions/google.rpc.Status"
-                }
-              },
-              "title": "Stream result of google.api.HttpBody"
+              "type": "string",
+              "format": "binary",
+              "properties": {},
+              "title": "Free form byte stream"
```
johanbrandhorst, line 2853: This now has the correct representation for the Download RPC return

johanbrandhorst:
I believe the CI failures are not due to this change, so we should be good to merge this

jimlambrt (approved):
Ty!

tmessi (approved):
lgtm, just some ideas about the new script.

scripts/remove-gotags-comments/main.go
```diff
@@ -0,0 +1,52 @@
...
+			// The two cases we're covering are:
+			//   - When gotags appears on its own.
+			//   - When gotags appears at the end of another comment (preceded by \n\n).
+			for _, prefix := range []string{"\\n\\n", ""} {
+				swaggerBytes = bytes.ReplaceAll(swaggerBytes, []byte(fmt.Sprintf("%s@gotags: %sclass:\\\"%s\\\"%s", prefix, wrapper, classification, wrapper)), nil)
+			}
+		}
+	}
+	// Some fields have a comment explaining that their classification is manually managed
+	swaggerBytes = bytes.ReplaceAll(swaggerBytes, []byte("\\n\\nclassified as public via taggable implementation"), nil)
```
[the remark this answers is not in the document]
johanbrandhorst: I tried writing this script with sed and jq before deciding on Go and I'm particularly happy that it lets us use the constants from the classification package, but thanks for those tips!

PR #3263, opened 2023-06-02
Support multiple IDs in grants

jefferai:
This allows specifying multiple IDs in grants via an `ids` field, which will eventually become the normal way to specify IDs. Internally, each grant with multiple IDs turns into multiple ACL grants each with a single ID; to make this distinction clearer and allow for better separation of concerns an AclGrant type has been introduced, so really multiple IDs in a Grant turn into multiple AclGrants.

The nice thing about doing it this way is that it becomes purely a parsing and validation concern; no actual authorization checking logic has changed.

For backwards compatibility with grants already stored in roles, we do not error out if `id` is specified. However, eventually we will simply not support grants with `id` coming via the API; they will fail API validation, such that new grants will require the use of the updated `ids` field.

A follow-on PR will plumb this through the service handlers/API/CLI, but I decided to do it in separate chunks to make reviewing easier, given that they are easily separable into the internal and external aspects of the change.


internal/proto/controller/api/resources/roles/v1/role.proto
```diff
@@ -25,7 +25,11 @@ message Principal {
 
 message GrantJson {
   // Output only. The ID, if set.
-  string id = 1; // @gotags: `class:"public"`
+  // Deprecated: use "ids" instead.
+  string id = 1 [deprecated = true]; // @gotags: `class:"public"`
+
+  // Output only. The IDs, if set.
+  repeated string ids = 4; // @gotags: `class:"public"`
```
irenarindos, line 32: I had assumed we listed fields in protos in order of their field number but is that not the case?

internal/perms/grants.go
```diff
@@ -343,154 +393,219 @@ func Parse(scopeId, grantString string, opt ...Option) (Grant, error) {
...
-	if err := grant.validateType(); err != nil {
-		return Grant{}, errors.WrapDeprecated(err, op)
+	if len(grant.ids) > 1 && slices.Contains(grant.ids, "*") {
+		return Grant{}, errors.NewDeprecated(errors.InvalidParameter, op, fmt.Sprintf("input grant string %q contains both wildcard and non-wildcard values in %q field", grantString, "ids"))
 	}
 
-	if err := grant.parseAndValidateActions(); err != nil {
-		return Grant{}, errors.WrapDeprecated(err, op)
-	}
+	opts := getOpts(opt...)
```
irenarindos, line 403: Should this also be changed to return an error? I notice you updated another options file for that

internal/perms/grants.go
```diff
@@ -343,154 +393,219 @@ func Parse(scopeId, grantString string, opt ...Option) (Grant, error) {
...
-			// Create a dummy resource and pass it through Allowed and ensure that
-			// we get allowed.
-			acl := NewACL(grant)
-			r := Resource{
-				ScopeId: scopeId,
-				Id:      grant.id,
-				Type:    grant.typ,
+
+		// We don't need to do these twice as they don't depend on IDs; they
+		// also clear state such as actionsBeingParsed
```
irenarindos, line 472:
```suggestion
		// also clearly state such as actionsBeingParsed
```

internal/perms/grants.go
jefferai: No, I did that package specifically to solve some test issues I found locally when those env vars are set since they didn't act as true/false. At other points we can potentially update other options but that was for a specific reason at the moment.

internal/perms/grants.go
jefferai: Nope, this was correct. They clear state in the grant struct :-)

internal/proto/controller/api/resources/roles/v1/role.proto
jefferai: Generally we like to do that, which is why I also like to have buffer between numbers, e.g. 10, 20. But there's no buffer here and I thought it'd be more useful to have the field next to the deprecated one it's replacing.

irenarindos (approved):

internal/perms/grants.go
```diff
@@ -343,154 +393,219 @@ func Parse(scopeId, grantString string, opt ...Option) (Grant, error) {
...
-			if opts.withAccountId != "" {
-				grant.id = strings.ToValidUTF8(opts.withAccountId, string(unicode.ReplacementChar))
-			}
-			origId = grant.id
-			grant.id = "acctoidc_dummy"
-		default:
-			return Grant{}, errors.NewDeprecated(errors.InvalidParameter, op, fmt.Sprintf("unknown template %q in grant %q value", grant.id, "id"))
-		}
+	if grant.id != "" && len(grant.ids) > 0 {
+		return Grant{}, errors.NewDeprecated(errors.InvalidParameter, op, fmt.Sprintf("input grant string %q contains both %q and %q fields", grantString, "id", "ids"))
```
jimlambrt, line 397: Any appetite for adding a ctx parameter to Parse(...), so we can stop using
errors.NewDeprecated in this func?

jimlambrt: also, can we add consts for `id` and `ids`?

jimlambrt (approved):
Looks great and ty!  Just a couple minor suggestions you might consider before merging.

internal/perms/grants.go
jefferai: I do want to add `ctx` but I don't want to muddle this PR with those changes -- I can do that as a follow-on.

internal/perms/grants_test.go
```diff
@@ -836,10 +1115,15 @@ func FuzzParse(f *testing.F) {
...
+	f.Add("ids=*;type=*;actions=*")
 	f.Add("id=*;type=*;actions=read,list")
+	f.Add("ids=*;type=*;actions=read,list")
 	f.Add("id=foobar;actions=read;output_fields=version,id,name")
+	f.Add("ids=foobar,foobaz;actions=read;output_fields=version,id,name")
 	f.Add("id={{account.id}};actions=update,read")
-	f.Add(`{id:"foobar","type":"host-catalog","actions":["create"]}`)
+	f.Add("ids={{account.id}},{{user.id}};actions=update,read")
+	f.Add(`{"id":"foobar","type":"host-catalog","actions":["create"]}`)
+	f.Add(`{"ids":["foobar"],"type":"host-catalog","actions":["create"]}`)
```
johanbrandhorst, line 1126: 👌🏻

globals/prefixes.go
```diff
@@ -163,10 +163,24 @@ var prefixToResourceType = map[string]resource.Type{
 	SessionRecordingPrefix:                     resource.SessionRecording,
 }
 
+var resourceTypeToPrefixes map[resource.Type][]string = func() map[resource.Type][]string {
+	ret := make(map[resource.Type][]string)
+	for k, v := range prefixToResourceType {
+		ret[v] = append(ret[v], k)
+	}
+	return ret
+}()
```
johanbrandhorst, line 172: If this is the only use of this new global, maybe put in as a local variable in `ResourcePrefixesFromType` and use a global sync.Once to populate it? That'd make it immutable at runtime.

johanbrandhorst (approved):

globals/prefixes.go
jefferai: If it's local how would you not have to populate it each time you call the function? You leave the function, it falls out of scope. Or do you mean keep the variable at the package level, but simply do the population in `ResourcePrefixesFromType`?

That does feel like it adds some complexity and I'm not at all worried about it being modified at runtime. 🤷

internal/perms/grants.go
jefferai: Done, sent you a new PR link.

globals/prefixes.go
johanbrandhorst: You're right, it'd still have to be a global, nevermind me!

PR #3265, opened 2023-06-02
Update dev worker auth flag names

jefferai:
This brings it in line with the updated names that have been settled on in light of deprecating the old KMS method.


louisruch (approved):
lgtm

PR #3272, opened 2023-06-03
Fix billing views

mgaffney:
Below are the commits included in this PR (I've included them here for
convience): 

## test(db): Conditionally skip monthly billing view tests

Certain tests fail when run on the first day of the month because of
edge cases in the setup of test data. This adds conditional logic to
skip those tests when run on the first day of a month.

## fix(db): Fix generated series to yield consistent results

The hcp billing `_all` views built a range of buckets using the
`generate_series` function using the minimum value of the
`session_pending_time` column from the `wh_session_accumulating_fact`
table as the start time and `now()` as the end time. It then truncated
values in the series using the `date_trunc` function. This allowed the
generated series of buckets returned to vary and could result in one
additional bucket being included in the series depending on the current
time and the start time.

As an illustrative example, the `hourly_all` query used:

    select date_trunc('hour',time)
      from generate_series(
             (select min(session_pending_time) -- start time
                from wh_session_accumulating_fact),
              now(),
              '1 hour'::interval
           ) as time;

Substituting the start time with a hard coded value to show the issue:

    -- Show current time, where the minutes portion is 55
    select now();
                  now
    -------------------------------
     2023-05-22 14:55:40.567605+00
    (1 row)

    -- generate series with start time's minutes portion at 57
    select date_trunc('hour',time)
      from generate_series(
              '2023-05-22 01:57:00'::timestamptz,
              now(),
              '1 hour'::interval
           ) as time;

           date_trunc
    ------------------------
     2023-05-22 01:00:00+00
     2023-05-22 02:00:00+00
     2023-05-22 03:00:00+00
     2023-05-22 04:00:00+00
     2023-05-22 05:00:00+00
     2023-05-22 06:00:00+00
     2023-05-22 07:00:00+00
     2023-05-22 08:00:00+00
     2023-05-22 09:00:00+00
     2023-05-22 10:00:00+00
     2023-05-22 11:00:00+00
     2023-05-22 12:00:00+00
     2023-05-22 13:00:00+00
    (13 rows)

    -- Wait a few minutes
    select now();
                 now
    ------------------------------
     2023-05-22 14:58:48.75287+00
    (1 row)

    -- Now run again and get an extra result, even though still
    -- within the same hour
    select date_trunc('hour',time)
      from generate_series(
              '2023-05-22 01:57:00'::timestamptz,
              now(),
              '1 hour'::interval
           ) as time;

           date_trunc
    ------------------------
     2023-05-22 01:00:00+00
     2023-05-22 02:00:00+00
     2023-05-22 03:00:00+00
     2023-05-22 04:00:00+00
     2023-05-22 05:00:00+00
     2023-05-22 06:00:00+00
     2023-05-22 07:00:00+00
     2023-05-22 08:00:00+00
     2023-05-22 09:00:00+00
     2023-05-22 10:00:00+00
     2023-05-22 11:00:00+00
     2023-05-22 12:00:00+00
     2023-05-22 13:00:00+00
     2023-05-22 14:00:00+00
    (14 rows)

To fix this, we call the `generate_series` function truncating the start
time argument to the bucket size:

    -- Hourly
    select bucket
      from generate_series(
              date_trunc('hour', select min(session_pending_time)
                                   from wh_session_accumulating_fact)),
              now(),
              '1 hour'::interval
           ) as bucket;

    -- Monthly
    select bucket
      from generate_series(
              date_trunc('month', select min(session_pending_time)
                                    from wh_session_accumulating_fact)),
              now(),
              '1 month'::interval
           ) as bucket;

Blame: 41b1b5b4b74b38141ddf934ff274dfb5c5b53d0e

## fix(db): Fix daily billing views

The hcp daily billing views used the PostgreSQL special input values
`today` and `yesterday` in the `create view` statements. PostgreSQL
evaluates these strings and converts them to there actual values when
the `create view` statement is executed effectively turning them into
constants. The PostgreSQL documentation specifically cautions against
their use when creating views:

  Caution: While the input strings `now`, `today`, `tomorrow`, and
  `yesterday` are fine to use in interactive SQL commands, they can have
  surprising behavior when the command is saved to be executed later,
  for example in prepared statements, views, and function definitions.
  The string can be converted to a specific time value that continues to
  be used long after it becomes stale. Use one of the SQL functions
  instead in such contexts. For example, `CURRENT_DATE + 1` is safer
  than `'tomorrow'::date`.

  https://www.postgresql.org/docs/12/datatype-datetime.html

The results returned from the hcp daily billing views can vary based on
the time zone setting of the PostgreSQL server. This adds 2 table
functions that call their respective views with the time zone set to
UTC. Calling these functions instead of their underlying views directly
prevents the returned values from varying when the server's time zone is
not set to UTC.

Blame: ace2def49dce2971ce37db44d94faac0e2289bb7

## feat(db): Add function to update billing snapshot table

This adds the `update_sessions_pending_daily_snapshot` SQL function
which updates the `sessions_pending_daily_snapshot` table by inserting
values for any days missing from the table.


mgaffney:
I'm closing this PR to make additional changes.

I noticed last night that the `daily_sessions_all.sql` tests fail when run between approximately 8pm and 12am EDT. I'm going to remove the `hcp_billing_daily_sessions_yesterday` and `hcp_billing_daily_sessions_all` views along with their tests since the views are no longer needed. I will open a new PR once those changes are done.

PR #2160, opened 2022-06-09
perf(session): Improve list response time

tmessi:
Notable changes include:

- Change to not include terminated sessions by default. Use the new
  `include_terminated=true` query parameter to return terminated
  sessions.
- Remove connections from list response. Connection information is still
  returned for a read of an individual session.
- Creates a periodic job that will delete terminated sessions.

PR #3280, opened 2023-06-05
fix(session): List performance with large number of session connections

tmessi:
As part of the performance improvements for listing sessions for v0.9.0,
the view used to retrieve sessions for listing was updated to no longer
include a join on session_connection. However, a later migration that
was adding additional columns to this view inadvertently re-added the
join and session_connection columns. This resulted in requests to list
sessions performing this join, and returning additional rows for each
connection in a session. The application would ignore this additional
data, since it is no longer returned via the API. This would be
particularly impactful in cases where there are a large number of
connections per session, as this would greatly increase the amount of
data returned by the database.

This fix removes the session_connection join from this view, which
should restore the original performance improvements even in cases where
there is a large number of session connections per session.

See: https://github.com/hashicorp/boundary/pull/2160
Ref: 32070678dc712812a42a511bb68f668658bd327a
Fixes: 1e3c941be11086e4fa3cbba3e90cb160377103b3

This should also help with #3281.

### DIFF
--- a/internal/db/schema/migrations/oss/postgres/69/02_session_worker_protocol.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/69/02_session_worker_protocol.up.sql
@@ -17,7 +17,7 @@ begin;
   );
 
   drop view session_list;
-  -- Replaces view from 60/02_sessions.up.sql to add swp.worker_id
+  -- Replaces view from 64/04_session_list.up.sql to add swp.worker_id
   -- Replaced in 72/03_session_list_perf_fix.up.sql
   create view session_list as
   select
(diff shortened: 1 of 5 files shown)


internal/db/schema/migrations/oss/postgres/69/02_session_worker_protocol.up.sql
```diff
@@ -18,6 +18,7 @@ begin;
 
   drop view session_list;
   -- Replaces view from 60/02_sessions.up.sql to add swp.worker_id
+  -- Replaced in 72/02_sessiojn_list_perf_fix.up.sql
```
johanbrandhorst, line 21:
```suggestion
  -- Replaced in 72/02_session_list_perf_fix.up.sql
```

johanbrandhorst (approved):

PR #3287, opened 2023-06-06
feat(bsr/convert): Remove asciicast minimum window size

tmessi:
This removes the minimum on width and height, instead it will just used
the largest width, height seen as the initial size for the asciicast.
This also provides the ability to still set a minimum via options to
convert.ToAsciicast if the caller has a known minimum size. Also note
that in cases where there is no size information, ie if there was no pty
request, this will still pick a default size of 80x24.


irenarindos (commented):
Looks great, one small ask!

PR #3288, opened 2023-06-06
fix(session): List performance with large number of session connections

tmessi:
As part of the performance improvements for listing sessions for v0.9.0,
the view used to retrieve sessions for listing was updated to no longer
include a join on session_connection. However, a later migration that
was adding additional columns to this view inadvertently re-added the
join and session_connection columns. This resulted in requests to list
sessions performing this join, and returning additional rows for each
connection in a session. The application would ignore this additional
data, since it is no longer returned via the API. This would be
particularly impactful in cases where there are a large number of
connections per session, as this would greatly increase the amount of
data returned by the database.

This fix removes the session_connection join from this view, which
should restore the original performance improvements even in cases where
there is a large number of session connections per session.

See: https://github.com/hashicorp/boundary/pull/2160
Ref: 32070678dc712812a42a511bb68f668658bd327a
Fixes: 1e3c941be11086e4fa3cbba3e90cb160377103b3

### DIFF
--- a/internal/db/schema/migrations/oss/postgres/60/02_sessions.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/60/02_sessions.up.sql
@@ -173,6 +173,7 @@ begin;
     for each row execute procedure insert_session();
 
   -- Replaces view from 59/01_target_ingress_egress_worker_filters.up.sql
+  -- Replaced in 64/04_session_list.up.sql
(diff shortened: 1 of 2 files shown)


tmessi:
The `build` CI jobs seem to be failing due to not finding the corresponding UI assets.

jefferai (approved):

PR #3298, opened 2023-06-08
api: remove connection ID

johanbrandhorst:
Removing this until we figure out why it's sometimes missing. May add it back again later.


johanbrandhorst:
> Does this compile? I think the session_recordings service and tests would need to be updated as well.

I ran `go build ./...`, I think most of the logic is in the ent repo?

PR #3302, opened 2023-06-09
Modify some tests to expose error cases that weren't being exposed

jefferai:


jefferai:
Note: this has previously been reviewed and gone through CI, merging.

PR #3321, opened 2023-06-13
chore: Add check for `go mod tidy`

moduli:
This PR adds another check to the `make-gen-delta` workflow to ensure that `go mod tidy` changes have been committed to the branch. There was a recent case where `go.mod` and `go.sum` were out of sync with the code base. Adding this check would prevent that from happening.

An alternate implementation is to update `make gen` to run `go mod tidy`. Curious to hear if there are any opinions here.

### DIFF
--- a/.github/workflows/make-gen-delta.yml
+++ b/.github/workflows/make-gen-delta.yml
@@ -9,7 +9,7 @@ permissions:
 
 jobs:
   make-gen-delta:
-    name: "Check for uncommited changes from make gen"
+    name: "Check for uncommitted changes from make gen"
     runs-on: ${{ fromJSON(vars.RUNNER) }}
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/make-gen-delta.yml
```diff
@@ -32,6 +32,9 @@ jobs:
       - name: Running make gen
         run: |
           make gen
+      - name: Running go mod tidy
+        run: |
+          go mod tidy
```
tmessi, line 37: This should probably happen right before `make tools`. Which is actually a reason to do it here instead of as a part of the `gen` target.

johanbrandhorst (approved):
I think this looks fine, running go mod tidy every generation shouldn't be necessary

tmessi (approved):

PR #3334, opened 2023-06-14
test(e2e): Rename test packages

moduli:
This PR renames some of the e2e test packages for clarity. When I originally used `static`, I was referring to a "static target", but I'm not sure if it really makes sense. I opted to use "base" here to refer to the base infrastructure needed to run boundary (i.e. basic setup). Curious to hear other opinions.

This is in preparation for switching some of the enos scenarios to primarily use docker.


johanbrandhorst (approved):
Alternatively, `basic`?

stasryzhov (approved):

PR #3337, opened 2023-06-15
chore: Ignore tags in workflow

moduli:
This PR updates the `build` workflow to ignore running on tags. This was done to prevent extra workflows from triggering that aren't necessary.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,9 +1,13 @@
 name: build
 
 on:
-  - workflow_dispatch
-  - push
-  - workflow_call
+  push:
+    branches:
+      - '**'
+    tags-ignore:
+      - '**'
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -1,9 +1,11 @@
 name: build
 
 on:
-  - workflow_dispatch
-  - push
-  - workflow_call
+  push:
+    branches:
+      - '**'
```
tmessi, line 6: Does this need an explicit `tags-ignore`?

tmessi: Seems like it wouldn't hurt to explicitly exclude all tags, and it seems safer to me to do so, in-case we change the branches to something other than "all branches".

tmessi (approved):

PR #2376, opened 2022-08-18
Add mechanism for plugin to normalize attributes

jefferai:
This introduces a new normalization call to the plugin RPC service. If
implemented, the plugin will be given the attributes prior to saving
them in a create/update call, and can perform any normalization needed,
e.g. ensuring that a parameter passed in is represented in a canonical
way. As an example, the CLI allows a user of the AWS plugin to pass in a
string, but canonically it is a string slice, and should be returned
that way as well. Not by chance, this is the exact behavior of the
testing implementation of this function.

This does not update any individual plugin, but rather sets the
controller-side logic up for plugins to support this functionality.

PR #3338, opened 2023-06-15
fix: add plugin normalization for onCreate methods

elimt:
onUpdate for host catalog & host set make a plugin call to normalize attributes passed in. This behavior was introduced in [PR-2376](https://github.com/hashicorp/boundary/pull/2376) but was only applied to onUpdate calls. This created different data structures for onCreate calls and onUpdate calls.


elimt:
> Would you mind, as a follow up to this PR, updating the documentation for the Normalize proto service definition (or at least creating a Jira task to do so) to communicate what You, Jeff, and I talked about regarding when Normalize would be appropriate and not, and the non-obvious consequences of using it?

@talanknight You mean this definition right here: https://github.com/hashicorp/boundary/blob/main/internal/proto/plugin/v1/host_plugin_service.proto#L44 ?

PR #3339, opened 2023-06-15
test(e2e): Update env variables and target creation

stasryzhov:
Ticket: [ICU-9194](https://hashicorp.atlassian.net/browse/ICU-9194)

This PR prepares OSS for session recording tests:
- Updates environment variables for aws e2e tests
- Refactors the `CreateNewTargetCli` method for targets creation via CLI to support additional options

[ICU-9194]: https://hashicorp.atlassian.net/browse/ICU-9194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/enos/enos-variables.hcl
+++ b/enos/enos-variables.hcl
@@ -182,5 +182,4 @@ variable "go_test_timeout" {
 variable "aws_region" {
   description = "AWS region where the resources will be created"
   type        = string
-  default     = ""
 }
--- a/enos/enos-variables.hcl
+++ b/enos/enos-variables.hcl
@@ -178,3 +178,9 @@ variable "go_test_timeout" {
   type        = string
   default     = "10m"
 }
+
+variable "aws_region" {
+  description = "AWS region where the resources will be created"
+  type        = string
+  default     = ""
+}
--- a/testing/internal/e2e/boundary/target.go
+++ b/testing/internal/e2e/boundary/target.go
@@ -48,53 +48,50 @@ func AddHostSourceToTargetApi(t testing.TB, ctx context.Context, client *api.Cli
 // Returns the id of the new target.
 func CreateNewTargetCli(t testing.TB, ctx context.Context, projectId string, defaultPort string, opt ...target.Option) string {
 	opts := target.GetOpts(opt...)
+	var args []string
 
-	args := []string{
-		"targets", "create", "tcp",
+	// Set target type. Default to tcp if not specified
+	if opts.WithType != "" {
+		args = append(args, string(opts.WithType))
+	} else {
+		args = append(args, "tcp")
+	}
+
+	args = append(args,
 		"-scope-id", projectId,
 		"-default-port", defaultPort,
 		"-name", "e2e Target",
 		"-format", "json",
-	}
+	)
 
+	if opts.WithAddress != "" {
+		args = append(args, "-address", opts.WithAddress)
+	}
 	if opts.WithDefaultClientPort != 0 {
 		args = append(args, "-default-client-port", fmt.Sprintf("%d", opts.WithDefaultClientPort))
 	}
+	if opts.WithEnableSessionRecording != false {
+		args = append(args, "-enable-session-recording", fmt.Sprintf("%v", opts.WithEnableSessionRecording))
+	}
+	if opts.WithStorageBucketId != "" {
+		args = append(args, "-storage-bucket-id", opts.WithStorageBucketId)
+	}
+	if opts.WithIngressWorkerFilter != "" {
+		args = append(args, "-ingress-worker-filter", opts.WithIngressWorkerFilter)
+	}
 
 	output := e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs("targets", "create"),
 		e2e.WithArgs(args...),
 	)
 	require.NoError(t, output.Err, string(output.Stderr))
+
 	var newTargetResult targets.TargetCreateResult
 	err := json.Unmarshal(output.Stdout, &newTargetResult)
 	require.NoError(t, err)
+
 	newTargetId := newTargetResult.Item.Id
 	t.Logf("Created Target: %s", newTargetId)
-
-	return newTargetId
-}
-
-// CreateNewAddressTargetCli uses the cli to create a new target using an
-// address in boundary.
-// Returns the id of the new target.
-func CreateNewAddressTargetCli(t testing.TB, ctx context.Context, projectId string, defaultPort string, address string) string {
-	output := e2e.RunCommand(ctx, "boundary",
-		e2e.WithArgs(
-			"targets", "create", "tcp",
-			"-scope-id", projectId,
-			"-default-port", defaultPort,
-			"-name", "e2e Target",
-			"-address", address,
-			"-format", "json",
-		),
-	)
-	require.NoError(t, output.Err, string(output.Stderr))
-	var newTargetResult targets.TargetCreateResult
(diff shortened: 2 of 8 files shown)


testing/internal/e2e/boundary/bucket.go
```diff
@@ -0,0 +1,44 @@
...
+	"testing"
+	"time"
+
+	"github.com/hashicorp/boundary/api/storagebuckets"
+	"github.com/hashicorp/boundary/testing/internal/e2e"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/net/context"
+)
+
+// CreateNewStorageBucket uses boundary CLI to create new Storage Bucket in the us-east-1 region
```
johanbrandhorst, line 18: Could the region be a parameter, even if we end up setting an environment variable before making the call?

testing/internal/e2e/boundary/scope.go
```diff
@@ -45,7 +47,7 @@ func CreateNewOrgCli(t testing.TB, ctx context.Context) string {
 	output := e2e.RunCommand(ctx, "boundary",
 		e2e.WithArgs(
 			"scopes", "create",
-			"-name", "e2e Org",
+			"-name", fmt.Sprintf("e2e Org %s", time.Now().String()),
```
johanbrandhorst, line 50: Was this added to avoid conflicts? When do we create 2 orgs at the same time? Adding this would make it harder to detect when a test doesn't clean up properly (since it won't conflict anymore).

testing/internal/e2e/boundary/session_recording.go
```diff
@@ -0,0 +1,116 @@
...
+			"-format", "json",
+		),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+
+	var sessionRecordingListResult sessionrecordings.SessionRecordingListResult
+	err := json.Unmarshal(output.Stdout, &sessionRecordingListResult)
+	require.NoError(t, err)
+
+	return sessionRecordingListResult.Items[0].Id
```
johanbrandhorst, line 37: Maybe add a `require.GreaterOrEq(t, len(sessionRecordingListResult.Items), 1)` to prevent an unseemly panic here?

testing/internal/e2e/boundary/session_recording.go
```diff
@@ -0,0 +1,116 @@
...
+					"-format", "json",
+				))
+			return output.Err
+		},
+		backoff.WithMaxRetries(backoff.NewConstantBackOff(3*time.Second), 5),
+		func(err error, td time.Duration) {
+			t.Logf("%s. Retrying...", err.Error())
+		},
+	)
+	require.NoError(t, err)
```
johanbrandhorst, line 109: Instead of retrying this until it passes, what do you think about retrying `read` on the session recording and checking the `state`? When it's `available` the download should work first time.

testing/internal/e2e/boundary/scope.go
stasryzhov, line 50: **Note:**
This is adjusted to make it easier to execute the test against local environment multiple times.
Storage Buckets cannot be deleted once there has been a session recorded and the Orgs cannot be deleted if they have a used Storage Bucket.
The test creates new organization. As we cannot delete it after the test, the organization has to have a unique name.

testing/internal/e2e/boundary/session_recording.go
```diff
@@ -0,0 +1,116 @@
...
+	t.Logf("Found Channel Recording: %s", channel.Id)
+	return channel.Id
+}
+
+// WaitAndDownloadSessionRecordingFile uses boundary CLI to periodically attempt to download
+// a session recording file from AWS S3 Bucket.
+// The method will fail if the file is not available on AWS yet or downloading failed.
+// It will return downloaded file name otherwise. The file will be saved in the current directory.
+func WaitAndDownloadSessionRecordingFile(t testing.TB, ctx context.Context, channelId string) string {
+	t.Log("Waiting for session recording file to download it...")
```
moduli, line 93:
```suggestion
	t.Log("Waiting for session recording file to download...")
```

testing/internal/e2e/boundary/bucket.go
```diff
@@ -0,0 +1,44 @@
...
+func CreateNewStorageBucket(t testing.TB, ctx context.Context, orgId, bucketName string) string {
+	output := e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"storage-buckets", "create",
```
moduli, line 22: If we're reusing the bucket for multiple tests, we may want to specify a unique recordings directory when creating the bucket. I think it's the `bucket-prefix` option

testing/internal/e2e/boundary/target.go
```diff
@@ -98,6 +98,34 @@ func CreateNewAddressTargetCli(t testing.TB, ctx context.Context, projectId stri
 	return newTargetId
 }
 
+// CreateNewSshTargetWithStorageBucketCli uses the cli to create a new SSH target using an
+// address in boundary. It enables session recording and sets a storage bucket.
+// Returns the id of the new target.
+func CreateNewSshTargetWithStorageBucketCli(t testing.TB, ctx context.Context, projectId, defaultPort, address, bucketId string) string {
```
moduli, line 104: Trying to think about how to structure this method. We will, for a separate test, need to create an ssh target (with or without session recording). Is there a clean way to have a single method for this? Maybe we use the `opt` pattern as seen in `CreateNewTargetCli`?

moduli (changes requested):
I think these changes should go into `boundary-enterprise`

testing/internal/e2e/boundary/scope.go
moduli: What if, for now, we set the `-scope-id` to `global`?

testing/internal/e2e/tests/aws/session_recordings_test.go
```diff
@@ -0,0 +1,74 @@
...
+	output := e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"connect",
+			"-target-id", newTargetId,
+			"-exec", "/usr/bin/ssh", "--",
+			"-o", "UserKnownHostsFile=/dev/null",
+			"-o", "StrictHostKeyChecking=no",
+			"-p", "{{boundary.port}}",
+			"{{boundary.ip}}",
+			fmt.Sprintf("echo %s", testValue),
```
moduli, line 56: we'll probably want to do more than just echo a single value

testing/internal/e2e/tests/aws/session_recordings_test.go
```diff
@@ -0,0 +1,74 @@
...
+		),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+
+	// Download session recording file from AWS S3 once it is available
```
moduli, line 61: we should validate that the session recording state is set to finished/completed.

johanbrandhorst (approved):

moduli (approved):

enos/enos-variables.hcl
```diff
@@ -178,3 +178,9 @@ variable "go_test_timeout" {
   type        = string
   default     = "10m"
 }
+
+variable "aws_region" {
+  description = "AWS region where the resources will be created"
+  type        = string
+  default     = ""
```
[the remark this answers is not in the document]
stasryzhov: Note: @moduli and I decided to make the variable required

moduli (approved):

PR #3340, opened 2023-06-16
test(e2e): Update worker module to create recordings directory if defined

moduli:
This PR updates the enos worker module to create, if defined, the recordings directory on the worker instance prior to starting the worker binary. This was done to mitigate permissions issues; the worker was unable to create the directory on its own.


johanbrandhorst (approved):
I see we're only updating the worker here, but have we also added a "bsr" purpose key to the server side configuration? It's required for session recording.

moduli:
> I see we're only updating the worker here, but have we also added a "bsr" purpose key to the server side configuration? It's required for session recording.

Yup. Working on that here: https://github.com/hashicorp/terraform-enos-aws-boundary/pull/39

enos/modules/worker/main.tf
```diff
@@ -203,3 +221,19 @@ resource "enos_boundary_start" "worker_start" {
...
   }
 }
+
+resource "enos_remote_exec" "update_directory" {
+  depends_on = [enos_boundary_start.worker_start]
+  count      = var.recording_storage_path == "" ? 0 : 1
+
+  inline = [
+    "sudo chmod -R 755 ${var.recording_storage_path}",
+    "sudo chown boundary:boundary ${var.recording_storage_path}"
```
tmessi, line 231: Can this not be done as part of the `create_directory`?

enos/modules/worker/main.tf
```diff
@@ -203,3 +221,19 @@ resource "enos_boundary_start" "worker_start" {
     }
   }
 }
+
+resource "enos_remote_exec" "update_directory" {
+  depends_on = [enos_boundary_start.worker_start]
```
tmessi, line 226: This seems odd to me, wouldn't the worker fail to start if the permissions are not correct? Shouldn't the dependency be the other way around?

enos/modules/worker/main.tf
moduli: When I tried it, the `boundary` user didn't exist until after the worker started.

enos/modules/worker/main.tf
moduli: It seemed to start ok if the permissions were wide open. This update step has a dependency on `worker_start` to ensure that the boundary user is available.

enos/modules/worker/main.tf
tmessi: What is the resource that creates the `boundary` user? Can that either:
- be separated from starting the worker so this can depend just on the existence of the user, and then starting the worker can depend on this.
- or update that resource to be the thing that can optionally create this directory after it has created the user?

moduli: Exploring an option to create the directory after it creates the user: https://github.com/hashicorp/enos-provider/pull/189. Need to work with the enos team on validating the implementation.

moduli: Got that added to `enos-provider`

PR #3341, opened 2023-06-16
api: remove create from auth-tokens help

johanbrandhorst:
The auth-tokens help command included a reference to a non-existent auth-tokens create command. Replace the default help and also add a helpful pointer to the authenticate command.

Before:
```
$ boundary auth-tokens -h
Usage: boundary auth-tokens [sub command] [options] [args]

  This command allows operations on Boundary auth token resources. Example:

    Create an auth token:

      $ boundary auth-tokens create -name prodops -description "For ProdOps usage"

  Please see the auth-tokens subcommand help for detailed usage information.

Subcommands:
    delete    Delete an auth token
    list      List an auth token
    read      Read an auth token
```

After:
```
$ boundary auth-tokens -h
Usage: boundary auth-tokens [sub command] [options] [args]

  This command allows operations on Boundary auth token resources. Example:

    List all auth tokens:

      $ boundary auth-tokens list -recursive

  Please see the auth-tokens subcommand help for detailed usage information.
  Note: To create an auth token, see the authenticate subcommand.

Subcommands:
    delete    Delete an auth token
    list      List an auth token
    read      Read an auth token
```

### DIFF
--- a/api/authtokens/authtokens.gen.go
+++ b/api/authtokens/authtokens.gen.go
@@ -44,7 +44,6 @@ func (n AuthTokenReadResult) GetResponse() *api.Response {
 	return n.response
 }
 
-type AuthTokenCreateResult = AuthTokenReadResult
 type AuthTokenUpdateResult = AuthTokenReadResult
 
 type AuthTokenDeleteResult struct {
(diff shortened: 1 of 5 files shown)


irenarindos (approved):

jefferai (commented):
As someone who hasn't used it via systemd I don't have a strong opinion here; however, I do think that @omkensey is likely right that in any given deployment you're likely to be running one or the other, and if you really want to run both on one machine there's not much reason not to do it in the same process.

PR #3342, opened 2023-06-16
systemd: make controller config example default

### LINKED ISSUE #2358: systemd unit name doesn't allow instances ('@controller' or '@worker')
**Describe the bug**
The boundary.service file uses the %i variable, so it's clearly meant to be run as 'systemd@controller.service', 'systemd@worker.service', and so on.

But systemctl/systemd doesn't understand how to create instances on a unit whose name doesn't end in an '@' sign. So in order to actually use the unit file, I have to copy /usr/lib/systemd/system/boundary.service to /etc/systemd/system/boundary@.service.

Then I am able to `systemctl enable boundary@controller` successfully.

Please rename the unit file to `boundary@.service`.

**To Reproduce**
1. See error on running `enable` command:
`# systemctl enable boundary@controller`
Failed to enable unit: Unit file boundary@controller.service does not exist.

**Expected behavior**
systemctl should be able to create the instance links.

**Additional context**
I'm running on Amazon Linux 2.

johanbrandhorst:
Our systemd setup included both a controller and worker configuration example, and a %i templated config file meant to be used to determine which to use. For ease of use, merge the two config files into a predictably named "boundary.hcl".

Fixes #2358


jefferai (commented):
As someone who hasn't used it via systemd I don't have a strong opinion here; however, I do think that @omkensey is likely right that in any given deployment you're likely to be running one or the other, and if you really want to run both on one machine there's not much reason not to do it in the same process.

.release/linux/package/etc/boundary.d/boundary.hcl
```diff
@@ -1,3 +1,112 @@
+# # Note that this is an example systemd file and is not intended to be functional as-is.
```
omkensey, line 1: This isn't really a systemd-managed file; it's just a config file mentioned by a systemd unit.

.release/linux/package/etc/boundary.d/boundary.hcl
```diff
@@ -1,3 +1,112 @@
...
 # # Full configuration options can be found at https://www.boundaryproject.io/docs/configuration
 
+# # What follows is an example set up for a Boundary worker. Further down in the file you will
```
omkensey, line 4: Rather than put two configs in one file, a more common/usable approach is to have separate example configs in the same directory, that the user can copy, edit or symlink.  So here you could either:
* Keep the controller config here as the default in `boundary.hcl`, with a separate worker config in a different file in the `boundary.d` directory that could be swapped in or referred to in-place; or
* Have both configs in separate files in the `boundary.d` directory and make `boundary.hcl` a symlink to one of them.

omkensey (changes requested):

.release/linux/package/etc/boundary.d/boundary.hcl
johanbrandhorst: Does that make sense given that the user could theoretically want to run both the controller and the worker in the same process (using the same config file, as is done here)?

johanbrandhorst: I separated this into two files, with controller as boundary.hcl.

omkensey (approved):
LGTM

PR #3343, opened 2023-06-19
docs: update NormalizeSetData comments

elimt:
Update comments for `NormalizeSetData` to include when `NormalizeSetData` could be useful and the non-obvious consequences of using it.

Follow-up PR for: #3338


elimt:
@talanknight In addition to the changes in this PR, I could also update the changelog to highlight the deprecation

PR #3346, opened 2023-06-20
Update census to run less frequently

mikemountain:
Census is currently running about once a minute, this will reduce that run-time to once per hour. Also includes a table to track the most recent log time.

### DIFF
--- a/internal/census/census_job.go
+++ b/internal/census/census_job.go
@@ -51,7 +51,6 @@ func (c *censusJob) Status() scheduler.JobStatus {
 // Run performs the required work depending on the implementation.
 // The context is used to notify the job that it should exit early.
 func (c *censusJob) Run(ctx context.Context) error {
-	const op = "census.(censusJob).Run"
 	err := RunFn(ctx, c)
 	return err
 }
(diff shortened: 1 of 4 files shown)


internal/census/census_job.go
```diff
@@ -61,10 +60,9 @@ func runInternal(ctx context.Context, c *censusJob) error {
 }
 
 // NextRunIn returns the duration until the next job run should be scheduled.
-// We report as ready immediately after a successful run. This doesn't mean that
-// this job will run immediately, only about as often as the configured scheduler interval.
+// Census will run every hour to ensure any interrupted jobs will be re-attempted
 func (c *censusJob) NextRunIn(_ context.Context) (time.Duration, error) {
-	return 0, nil
+	return 3600, nil
```
louisruch, line 65: `NextRunIn` returns a `duration` not number of seconds, I believe this will effectively still be `0` since we round to seconds and convert to an int when inserting into the database `int(nextRunIn.Round(time.Second).Seconds()`. 

Rather use `time.Hour` here.

mgaffney, line 65:
```suggestion
	return 1*time.Hour, nil
```
Would doing something like the above be the more idiomatic way in Go?

internal/db/schema/migrations/oss/postgres/73/01_census_logging.up.sql
```diff
@@ -0,0 +1,20 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: MPL-2.0
+
+begin;
+
+  create table census_last_logged (
```
mgaffney, line 6: Where is this table used?

mgaffney (changes requested):

johanbrandhorst (approved):

mgaffney (approved):

jefferai (approved):

PR #3347, opened 2023-06-20
Add warnings to the customer facing Boundary API.

talanknight:
Warnings are helpful for deprecation notices or when a request configures Boundary in a way that is valid but might not work as intended.  A recent example of where warnings might be helpful is when a user updates a target's "filter" field which is deprecated in favor of "ingress_worker_filter" or "egress_worker_filter".

This PR is a draft for now while the RFC is being reviewed and discussions are ongoing.

### DIFF
--- a/internal/cmd/base/format.go
+++ b/internal/cmd/base/format.go
@@ -280,6 +280,25 @@ func (c *Command) PrintApiError(in *api.Error, contextStr string, opt ...Option)
 	}
 }
 
+func (c *Command) PrintWarning(w *api.Warning) {
+	if w == nil {
+		return
+	}
+	switch Format(c.UI) {
+	case "table":
+		var descStr string
+		switch {
+		case w.Action != nil:
+			descStr = fmt.Sprintf("%s: %s", w.Action.Name, w.Action.Warning)
+		case w.Behavior != nil:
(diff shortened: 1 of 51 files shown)


internal/warning/warning.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+type warningKey int
+
+var (
+	warnerContextkey warningKey
+	warningHeader    = "x-boundary-warning"
+)
+
+type Warner struct {
```
johanbrandhorst, line 26: image

internal/warning/warning_test.go
```diff
@@ -0,0 +1,144 @@
...
+	resp, err := http.Get(fmt.Sprintf("http://%s/health", lis.Addr().String()))
+	require.NoError(t, err)
+	got := resp.Header.Get(warningHeader)
+
+	require.NoError(t, err)
+	assert.Equal(t, string(want), got)
+
+	assert.NoError(t, httpSrv.Shutdown(ctx))
+	grpcSrv.GracefulStop()
+	wg.Wait()
```
johanbrandhorst, line 132: These steps could (should?) probably be in separate `t.Cleanup` invocations to ensure they always run

PR #3349, opened 2023-06-21
chore: update copyright_year in copywrite config

cameronperera:
Updating the `copyright_year` to 2020 after reviewing copywrite [README](https://github.com/hashicorp/copywrite#config-structure) that suggests using the year the project started.

### DIFF
--- a/.copywrite.hcl
+++ b/.copywrite.hcl
@@ -2,7 +2,7 @@ schema_version = 1
 
 project {
   license        = "MPL-2.0"
-  copyright_year = 2023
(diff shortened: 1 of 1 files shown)


johanbrandhorst (approved):
Huh, I wonder what this setting is actually for seeing as our license headers don't contain the year.

cameronperera:
> Huh, I wonder what this setting is actually for seeing as our license headers don't contain the year.

Yeah not sure. It is optional as well which I find odd.

PR #3350, opened 2023-06-21
chore: update gldap dep to fix intermittent shutdown/stop error

jimlambrt:
Occasionally, we see an error from the gldap test directory reported
during the shutdown/close of the directories listener.   This update
should fix that issue.


johanbrandhorst (approved):
:yolo:

PR #3353, opened 2023-06-22
refactor: make secrets optional for storage bucket

ddebko:
## Summary:

This PR makes the field `secrets` optional for the storage bucket resource. The reason for removing this constraint is because we are providing users the option to utilize dynamic credentials rather than static credentials.

### Changes:

- dropped the `not null` constraint from the `secrets_hmac` column in the `storage_plugin_storage_bucket` table
- remove nil checks on the secrets field in the loopback plugin
- add nil checks on the attributes field in the loopback plugin
- update the gorm tag for the field `secrets_hmac` in the `StorageBucket` proto definition
- removed `missing secrets` test case from unit tests
- added `missing attributes` test case to unit tests
- added unit tests for validating storage bucket plugin methods work without secrets for the loopback plugin

This PR is dependent on the [aws plugin PR](https://github.com/hashicorp/boundary-plugin-aws/pull/28).


hugoamvieira (approved):
As we discussed, I'd like others to also review these PRs as I feel I lack the full context, but LGTM.

PR #3361, opened 2023-06-26
fix(ldap managed grp): allow ldap managed grps to be added as principals

jimlambrt:


internal/iam/repository_principal_role.go
```diff
@@ -610,6 +610,11 @@ func splitPrincipals(ctx context.Context, principals []string) (users, groups, m
 				managedGroups = make([]string, 0, len(principals))
 			}
 			managedGroups = append(managedGroups, principal)
+		case strings.HasPrefix(principal, globals.LdapManagedGroupPrefix):
```
mgaffney, line 613: Can we add a test for this?

jimlambrt: It's not currently test for either OIDC or LDAP and doing so will require refactoring the tests a bit.  I'd prefer to do that in a separate PR, since I'm currently blocked for a demo.

johanbrandhorst (approved):

PR #3376, opened 2023-06-29
chore: Remove check-legacy-links-format workflow

moduli:
This PR deletes the `check-legacy-links-format.yml` workflow. We observed that it was failing with the following error
```
Loaded env from /home/runner/work/boundary/boundary/.env
Error: Cannot find module './generate-collection-sidebar-nav-data'
```
https://github.com/hashicorp/boundary/actions/runs/5413302725/jobs/9838661652?pr=3340

Per similar PRs like https://github.com/hashicorp/consul/pull/17647, the workflow has been deprecated in favor of similar checks in other places.


BRKalow (approved):
Thanks!

PR #3380, opened 2023-07-05
Update SSH target attributes url typo

edorion:
https://developer.hashicorp.com/boundary/docs/concepts/docmain-model/targets#ssh-target-attributes-hcp-ent should be https://developer.hashicorp.com/boundary/docs/concepts/domain-model/targets#ssh-target-attributes-hcp-ent

(domain not docmain)


Dan-Heath (approved):
Thank you!

PR #3381, opened 2023-07-05
Update create-storage-bucket.mdx PKI worker to session recording config

edorion:
link PKI worker link to session recoding config


Dan-Heath (approved):
Thank you!

PR #3382, opened 2023-07-05
Update enable-session-recording.mdx to link bsr controler config to bsr kms key doc

edorion:
Update enable-session-recording.mdx to link bsr controler config to bsr kms key doc


Dan-Heath (approved):
Thank you!

PR #3383, opened 2023-07-05
docs: Update HCP only references

Dan-Heath:
The **HCP only** tag should be updated to **HCP/ENT** in places where a feature is supported by both versions. This PR fixes some topics that I missed in the 0.13 release.


xingluw (approved):
LGTM

stellarsquall (approved):
Thanks!

PR #3387, opened 2023-07-05
test(e2e): Add test to connect to target via its name and scope

stasryzhov:
This PR adds a test to verify connection to targets via a combination of the `-target-name` and `-target-scope-name` options.

The PR also:
- adds an optional parameter to the `boundary.CreateNewProjectCli` to be able to optionally set custom project name
- adds support of the `target.WithName` option for the `boundary.CreateNewTargetCli`

### DIFF
--- a/testing/internal/e2e/boundary/scope.go
+++ b/testing/internal/e2e/boundary/scope.go
@@ -64,14 +64,24 @@ func CreateNewOrgCli(t testing.TB, ctx context.Context) string {
 // CreateNewProjectCli creates a new project in boundary using the cli. The project will be created
 // under the provided org id.
 // Returns the id of the new project.
-func CreateNewProjectCli(t testing.TB, ctx context.Context, orgId string) string {
+func CreateNewProjectCli(t testing.TB, ctx context.Context, orgId string, opt ...ScopeOption) string {
+	opts := getScopeOpts(opt...)
+	var args []string
+
+	args = append(args,
+		"scopes", "create",
+		"-scope-id", orgId,
+		"-format", "json",
+	)
+
+	if opts.WithName != "" {
+		args = append(args, "-name", opts.WithName)
+	} else {
+		args = append(args, "-name", "e2e Project")
+	}
+
 	output := e2e.RunCommand(ctx, "boundary",
-		e2e.WithArgs(
-			"scopes", "create",
-			"-name", "e2e Project",
-			"-scope-id", orgId,
-			"-format", "json",
-		),
+		e2e.WithArgs(args...),
 	)
 	require.NoError(t, output.Err, string(output.Stderr))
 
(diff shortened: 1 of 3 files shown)


testing/internal/e2e/tests/base/target_tcp_connect_test.go
```diff
@@ -65,6 +65,59 @@ func TestCliTcpTargetConnectTargetBasic(t *testing.T) {
...
+	boundary.AddHostToHostSetCli(t, ctx, newHostSetId, newHostId)
+	newTargetId := boundary.CreateNewTargetCli(t, ctx, newProjectId, c.TargetPort, target.WithName(testTargetName))
+	boundary.AddHostSourceToTargetCli(t, ctx, newTargetId, newHostSetId)
+
+	// Connect to target via target and scope names, and print host's IP address
+	output := e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"connect",
+			"-target-name", testTargetName,
+			"-target-scope-name", testProjectName,
```
stasryzhov, line 101: Connecting to target via its name and scope

moduli: How do these options work? Do you always have to provide both options?

testing/internal/e2e/boundary/scope.go
```diff
@@ -83,3 +93,29 @@ func CreateNewProjectCli(t testing.TB, ctx context.Context, orgId string) string
 	t.Logf("Created Project Id: %s", newProjectId)
 	return newProjectId
 }
+
+// getScopeOpts iterates the inbound ScopeOptions and returns a struct
+func getScopeOpts(opt ...ScopeOption) scopeOptions {
```
moduli, line 98: For targets, we use boundary's target options already defined in the application code. Is there not a corresponding one for scopes?

testing/internal/e2e/tests/base/target_tcp_connect_test.go
stasryzhov: Yes. The `-target-name` should be used in conjunction with either `-target-scope-name` or `-target-scope-id`

Documentation: [Selecting Targets](https://developer.hashicorp.com/boundary/docs/hcp/get-started/connect-to-target#selecting-targets)

So we may want to add a target connect attempt with `-target-name` and `-target-scope-id` as well

moduli: Sounds good. Should we also modify the names to include some other special characters?

johanbrandhorst (approved):
Approved pending Michael's comments

testing/internal/e2e/boundary/scope.go
stasryzhov: I haven't found it in the code base. If I missed that, let me know, please 🙏

moduli: I'm not super familiar with this. I did find this:https://github.com/hashicorp/boundary-enterprise/blob/main/api/scopes/option.gen.go?

moduli: Hm. I might be confusing them with the API options (https://github.com/hashicorp/boundary-enterprise/blob/main/api/scopes/option.gen.go).

moduli (approved):

PR #3389, opened 2023-07-06
bug(workerAuth): allow duplicate workerAuth inserts if records match

irenarindos:
Fixes https://hashicorp.atlassian.net/browse/ICU-9986

### DIFF
--- a/internal/server/repository_workerauth.go
+++ b/internal/server/repository_workerauth.go
@@ -113,7 +113,7 @@ func (r *WorkerAuthRepositoryStorage) Store(ctx context.Context, msg nodee.Messa
 // * the workerAuth record is stored with a reference to a worker
 // * certificate bundles are stored with a reference to the workerAuth record and issuing root certificate
 func StoreNodeInformationTx(ctx context.Context, reader db.Reader, writer db.Writer, kmsCache *kms.Kms, scopeId string, node *types.NodeInformation, _ ...Option) error {
-	const op = "server.(WorkerAuthRepositoryStorage).storeNodeInformation"
+	const op = "server.(WorkerAuthRepositoryStorage).StoreNodeInformationTx"
 	if isNil(reader) {
 		return errors.New(ctx, errors.InvalidParameter, op, "missing reader")
 	}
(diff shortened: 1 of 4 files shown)


internal/server/repository_workerauth.go
```diff
@@ -233,7 +233,23 @@ func StoreNodeInformationTx(ctx context.Context, reader db.Reader, writer db.Wri
 		return errors.Wrap(ctx, err, op)
 	}
 
-	// Store WorkerAuth
+	// Check if we already have a workerAuth record for this key id
+	nodeAuthLookup := allocWorkerAuth()
+	nodeAuthLookup.WorkerKeyIdentifier = node.Id
+	if err := reader.LookupById(ctx, nodeAuthLookup); err != nil {
+		switch {
+		case errors.Convert(err) != nil && errors.Convert(err).Code == errors.RecordNotFound:
```
johanbrandhorst, line 241: Can we not just use
```suggestion
		case errors.IsNotFoundError(err):
```
?

internal/server/worker_auth.go
```diff
@@ -50,6 +51,21 @@ func (w *WorkerAuth) decrypt(ctx context.Context, cipher wrapping.Wrapper) error
 	return nil
 }
 
+func (w *WorkerAuth) compare(other *WorkerAuth) bool {
+	switch {
+	case w.WorkerId != other.WorkerId:
+		return false
+	case !bytes.EqualFold(w.WorkerEncryptionPubKey, other.WorkerEncryptionPubKey):
```
johanbrandhorst, line 58: Is using `EqualFold` actually correct here? I don't think we want to be doing any normalization as it could affect the underlying data, shouldn't we just use `Equal`?

johanbrandhorst (commented):
Thanks for this!

internal/server/worker_auth.go
irenarindos: Ah thanks! GoLand autocomplete bit me on this one

jefferai (approved):
Looks good -- can you add a comment into the tests that explain why the test fails? That is, the comment says to try to store the "same" node information twice, but it's not clear why that would pass or fail (when really it depends on the fact that we only care about some data being the same).

internal/server/repository_workerauth.go
irenarindos: ty!

PR #3396, opened 2023-07-06
docs: Update headings for new style rule

Dan-Heath:
The new HashiCorp style for headings is sentence case (**C**apitalize only the first letter). Currently our docs use initial caps (**C**apitalize **E**ach **I**nitial **L**etter).

This pull request changes all levels of headings to use sentence case capitalization. It also updates the table of contents.

I corrected some instances in which a 2nd level heading was at the top of a topic, and it was causing the file name to display instead of the title. I also added simple intro sentences in some places where titles were nested without content in between.


stellarsquall (approved):
wow 🎉

PR #3397, opened 2023-07-06
fix(cmd): pass in PublicClusterAddr from the config

ddebko:
## Summary:

Users should be able to include a public_cluster_addr field in the controller stanza of their config file.  The value entered there should be used when reporting down to workers which addresses they should use when attempting to connect to the set of controller.  This no longer works.

### Fix:

Pass in the public_cluster_addr from the config into the `SetupControllerPublicClusterAddress` function. When public_cluster_addr is not set by the user, it should fall back to the cluster listener value.


irenarindos:
Should we close this PR?

PR #3412, opened 2023-07-12
docs: Update enable session recording on a target

DhariniJeeva:
We don't need to inject an application credential source to enable session recording on a target. We only need an injected application credential source before making a target connection.


Dan-Heath (approved):
Great catch @DhariniJeeva , thank you! I pushed one small update to make the sentence preceding your change singular instead of plural. Otherwise it LGTM!

AdamBouhmad (approved):
Looks great!

PR #3415, opened 2023-07-12
chore: Prevent build and test workflows on docs-only changes

moduli:
This PR prevents the `build` and `test` GitHub Actions workflows from running if the only changes are in the `website/` directory. This allows the docs changes to more rapidly make changes without having to wait for the longer workflows, especially since docs changes should have no impact on the application.

The script and implementation was heavily borrowed from Vault: https://github.com/hashicorp/vault/blob/main/.github/scripts/verify_changes.sh

This was tested by this test commit: https://github.com/hashicorp/boundary/commit/47b2dd898d34128817f25d49e4d53e122e4114ac. I forced `is_docs_change=true` to validate that it will prevent `build` and `test` workflows from executing. I also confirmed that making a change in the `website/` directory will get detected by the script.

### DIFF
--- a/.github/scripts/verify_changes.sh
+++ b/.github/scripts/verify_changes.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+#  This script validates if the git diff contains only docs changes
+
+event_type=$1 # GH event type (pull_request)
+ref_name=$2 # branch reference that triggered the workflow
+base_ref=$3 # PR branch base ref
+
+contains() {
+    target=$1; shift
+    for i; do
+        if [[ "$i" == "$target" ]]; then
+            return 0
+        fi
+    done
+    return 1
+}
+
+if [[ "$event_type" == "pull_request" ]]; then
+  git fetch --no-tags --prune origin $base_ref
+  head_commit="HEAD"
+  base_commit="origin/$base_ref"
+else
+  git fetch --no-tags --prune origin $ref_name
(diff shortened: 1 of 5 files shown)


tmessi:
Could we just use a [paths-ignore](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths) for the workflows instead of creating a new script?

moduli:
> Could we just use a [paths-ignore](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths) for the workflows instead of creating a new script?

That seems like a much nicer to do it.

johanbrandhorst (approved):

tmessi (changes requested):
You could also do this for the [go linter](https://github.com/hashicorp/boundary/blob/main/.github/workflows/linting.yml) and [gen delta](https://github.com/hashicorp/boundary/blob/main/.github/workflows/make-gen-delta.yml) workflows.

moduli:
> You could also do this for the [go linter](https://github.com/hashicorp/boundary/blob/main/.github/workflows/linting.yml) and [gen delta](https://github.com/hashicorp/boundary/blob/main/.github/workflows/make-gen-delta.yml) workflows.

@tmessi I'll add it to the go linter workflow. Don't we want to keep it on the `gen delta` workflow since it does the permissions table check, which is on the website directory? https://github.com/hashicorp/boundary/blob/main/Makefile#L131-L136

PR #3418, opened 2023-07-12
chore: prefer gosec for semgrep rule and turn on codeql

mcollao-hc:
Enabling CodeQL scanning since we can use beefier runners. Also moving to [gosec](https://github.com/securego/gosec) Semgrep rules for a more targeted ruleset.

### DIFF
--- a/.github/workflows/security-scan.yml
+++ b/.github/workflows/security-scan.yml
@@ -6,7 +6,9 @@ on:
   pull_request:
     branches:
       - 'main'
-
+    paths-ignore:
+      - 'website/'
+      
 jobs:
   scan:
     runs-on: ${{ fromJSON(vars.RUNNER_LARGE) }}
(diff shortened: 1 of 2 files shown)


tmessi (approved):

.github/workflows/security-scan.yml
```diff
@@ -6,7 +6,9 @@ on:
   pull_request:
     branches:
       - 'main'
-
+    paths-ignore:
+      - 'website/'
```
moduli, line 10: Not sure if this needs the `**` here: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths
```suggestion
      - 'website/**'
```

mcollao-hc: Ah, yes. Let me get a new PR up since I was quick on the trigger here.

mcollao-hc: @moduli https://github.com/hashicorp/boundary/pull/3435

PR #3424, opened 2023-07-12
Connect option docs

johanbrandhorst:
Adds separate sections for all the different connect helpers we have. There are quite a lot of videos on this page now, so maybe we need to break out the different helpers into their own sections, but that can follow later.


Dan-Heath (approved):
Great work on this Johan! I really like how the videos turned out. I think this will be a fantastic addition to the docs.  

I agree this content could be broken out a bit to make it easier to find. We have some other work in progress to document the CLI (#3411 ), and Robin and I think the connect helpers might fit nicely in that section once it is merged. So I’m going to merge this one as is, and we’ll revisit moving those sections when the other PR gets merged.

  I made a few minor copy edits. Please let me know if you have any questions or concerns. Thank you very much!

PR #3440, opened 2023-07-13
docs: BSR KMS Configuration Requirements

mikemountain:
Add configuration requirements to docs for session recording that uses KMS encryption

### DIFF
--- a/website/content/docs/configuration/session-recording/create-storage-bucket.mdx
+++ b/website/content/docs/configuration/session-recording/create-storage-bucket.mdx
@@ -52,6 +52,19 @@ For more information about using session recording to audit user sessions, refer
          }
       ]
    }
+
+   ```
+- If KMS encryption is applied to the storage bucket, you will need to add these additional permissions for the Storage Bucket IAM User to the role policy:
+   ```json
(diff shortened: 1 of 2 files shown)


website/content/docs/configuration/session-recording/create-storage-bucket.mdx
```diff
@@ -52,6 +52,19 @@ For more information about using session recording to audit user sessions, refer
          }
       ]
    }
+
+   ```
+- If KMS encryption is applied to the storage bucket, you will need to add these additional permissions for the Storage Bucket IAM User to the role policy:
```
Dan-Heath, line 57:
```suggestion
- If you apply KMS encryption to the storage bucket, you must add these additional permissions to the role policy for the storage bucket's IAM user:
```

Dan-Heath (changes requested):
Thank you for this update! I added a suggestion that I think makes the new bullet a little easier to understand, but please double check that I didn't change the meaning.

Dan-Heath (approved):
Looks good. Thank you for this @mikemountain !

PR #3447, opened 2023-07-13
chore: Change summary errors field type

elimt:
decoding summary `Errors` field currently fails with type `error`.  Using a string type to handle `Errors` field

**Previous commit approach**
Using a custom `SummaryError` type that implements the `error` interface to fix this issue. This caused the `Errors` field to be nested. eg. `Errors: { Message: "err"}` which failed other tests

### DIFF
--- a/internal/bsr/types.go
+++ b/internal/bsr/types.go
@@ -8,6 +8,14 @@ import (
 	"time"
 )
 
+type SummaryError struct {
+	Message string
+}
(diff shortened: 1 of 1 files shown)


moduli (approved):

elimt:
Merging to unblock dependant repos. Will do a follow-up PR to address PR feedback from this PR

PR #3448, opened 2023-07-13
Update iam.mdx

xingluw:
Editing for clarity

### DIFF
--- a/website/content/docs/concepts/iam.mdx
+++ b/website/content/docs/concepts/iam.mdx
@@ -54,8 +54,8 @@ There are 3 steps to adding new users to Boundary using the password auth method
 3. In the same scope, create a new **user**, and then attach the new account to the new user. The user is now able to authenticate to Boundary.
 
 ### Configure users - OIDC/LDAP
-With the [OIDC](/boundary/tutorials/identity-management/oidc-auth) and LDAP auth methods, Boundary automatically generates accounts and users in the same scope
-as the auth method, using data provided by the identity provider. The accounts and users are only created once the user authenticates to Boundary for the first time. 
+With the [OIDC](/boundary/tutorials/identity-management/oidc-auth) and LDAP auth methods, Boundary uses data from the identity provider to automatically generate accounts and users in the same scope
+as the auth method. The accounts and users are only created once the user authenticates to Boundary for the first time. 
 The same applies to OIDC/LDAP [managed groups](/boundary/tutorials/identity-management/oidc-idp-groups).
 
 ### Grant permissions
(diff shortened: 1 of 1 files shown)


website/content/docs/concepts/iam.mdx
```diff
@@ -41,6 +41,10 @@ method's identity provider to provide up-to-date information.
 [permissions](/boundary/docs/concepts/security/permissions) which are granted to any principal (user or group) from any scope. Roles
 can be contained by any scope, and the permissions can be applied to the same scope or any child scope.
 
+<Tip>
```
xingluw, line 44: @Dan-Heath would it be appropriate to change this into some kind of note saying that we want to change this to support multiple roles in the future? This is a minor change we publicly comment on, not sure if it is acceptable in documentation

website/content/docs/concepts/iam.mdx
```diff
@@ -50,9 +54,9 @@ There are 3 steps to adding new users to Boundary using the password auth method
 3. In the same scope, create a new **user**, and then attach the new account to the new user. The user is now able to authenticate to Boundary.
 
 ### Configure users - OIDC/LDAP
-[OIDC](/boundary/tutorials/identity-management/oidc-auth) and LDAP accounts and users are auto-vivified (automatically generated) in Boundary in the same scope
-as the auth method. The accounts and users are only created once the user authenticates to Boundary for the first time. The same applies to OIDC/LDAP
-[managed groups](/boundary/tutorials/identity-management/oidc-idp-groups).
+With the [OIDC](/boundary/tutorials/identity-management/oidc-auth) and LDAP auth methods, Boundary automatically generates accounts and users in the same scope
+as the auth method, using data provided by the identity provider. The accounts and users are only created once the user authenticates to Boundary for the first time.
```
Dan-Heath, line 58:
```suggestion
With the [OIDC](/boundary/tutorials/identity-management/oidc-auth) and LDAP auth methods, Boundary uses data from the identity provider to automatically generate accounts and users in the same scope
as the auth method. The accounts and users are only created once the user authenticates to Boundary for the first time. 
```

website/content/docs/concepts/iam.mdx
Dan-Heath: I'd recommend against describing future plans like this in the docs. It could potentially be confusing to describe the way we want a feature to work in the future. Readers might not remember the future caveat later and mistakenly think it is supported now. Also, the text itself can become dated quickly if the update doesn't happen according to our anticipated schedule.

Dan-Heath (approved):
This is a nice clarification, thank you! I suggested one minor rewrite that I think simplifies the sentence, but please let me know if you have any questions.

website/content/docs/concepts/iam.mdx
xingluw: Makes sense, won't put that in

xingluw:
Thanks @Dan-Heath, you can merge if ready (not sure if you are planning it alongside other docs day PRs)

PR #3462, opened 2023-07-17
test(e2e): Use vault-generic credential library type

moduli:
The `vault` credential library type will be deprecated at some point in favor of the `vault-generic` credential library type. This PR updates the e2e tests to use the `vault-generic` credential library type (and keeps one test that uses the `vault` credential library type. It will be removed once the option is deprecated).


stasryzhov (approved):
👍

PR #3475, opened 2023-07-18
test(e2e): Update grants to use new ids field

moduli:
Ticket: [ICU-10291](https://hashicorp.atlassian.net/browse/ICU-10291)

This PR updates e2e tests to use the new `ids` field in permission grants, which was introduced in `0.13.1`. 

One thing to note: originally, the migration test was failing with this change (before `0.13.1` was released) because it would try to upgrade from `0.13.0 (latest release) -> latest changes`) and  `0.13.0` did not recognize the `ids` field. Once `0.13.1` was released, that became the latest released version, which does support this field. We may look to address this in future work if we decide to update the migration test to try to update from a variety of different versions (i.e. `0.12.X -> latest`, 0.11.X -> latest`).

[ICU-10291]: https://hashicorp.atlassian.net/browse/ICU-10291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/testing/internal/e2e/tests/base/session_cancel_group_test.go
+++ b/testing/internal/e2e/tests/base/session_cancel_group_test.go
@@ -99,7 +99,7 @@ func TestCliSessionCancelGroup(t *testing.T) {
 
 	// Create a role for a group
 	newRoleId := boundary.CreateNewRoleCli(t, ctx, newProjectId)
-	boundary.AddGrantToRoleCli(t, ctx, newRoleId, "id=*;type=target;actions=authorize-session")
+	boundary.AddGrantToRoleCli(t, ctx, newRoleId, "ids=*;type=target;actions=authorize-session")
 	boundary.AddPrincipalToRoleCli(t, ctx, newRoleId, newGroupId)
 
 	// Connect to target to create a session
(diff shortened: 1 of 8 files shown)


stasryzhov:
> Ticket: [ICU-10291](https://hashicorp.atlassian.net/browse/ICU-10291)
> 
> This PR updates e2e tests to use the new `ids` field in permission grants, which was introduced in `0.13.1`.
> 
> One thing to note: originally, the migration test was failing with this change (before `0.13.1` was released) because it would try to upgrade from `0.13.0 (latest release) -> latest changes`) and `0.13.0` did not recognize the `ids` field. Once `0.13.1` was released, that became the latest released version, which does support this field. We may look to address this in future work if we decide to update the migration test to try to update from a variety of different versions (i.e. `0.12.X -> latest`, 0.11.X -> latest`).

Do we want to put a ticket for that to the backlog?

stasryzhov (approved):

moduli:
> Do we want to put a ticket for that to the backlog?

@stasryzhov Yup. We have one: https://hashicorp.atlassian.net/browse/ICU-9292. Already updated that ticket with the information about the `ids` field change.

PR #3481, opened 2023-07-19
feat: Perform asciicast conversion sanity checks

elimt:
Check that the conversion to asciicast is done on `exec` or `shell` session program. 
Unsupported session programs will throw an error.

### DIFF
--- a/internal/bsr/convert/convert.go
+++ b/internal/bsr/convert/convert.go
@@ -55,20 +55,28 @@ func ToAsciicast(ctx context.Context, session *bsr.Session, tmp storage.TempFile
 			return nil, fmt.Errorf("%s: %w", op, err)
 		}
 
-		// TODO sanity checks before getting the data files:
-		// - check connection summary to see if there was an exec or shell request
+		chs := ch.Summary.(*ssh.ChannelSummary)
 
-		reqScanner, err := ch.OpenRequestScanner(ctx, bsr.Inbound)
-		if err != nil {
-			return nil, fmt.Errorf("%s: %w", op, err)
+		switch chs.SessionProgram {
+		case ssh.Shell, ssh.Exec:
+			reqScanner, err := ch.OpenRequestScanner(ctx, bsr.Inbound)
+			if err != nil {
+				return nil, fmt.Errorf("%s: %w", op, err)
+			}
+
+			msgScanner, err := ch.OpenMessageScanner(ctx, bsr.Outbound)
+			if err != nil {
+				return nil, fmt.Errorf("%s: %w", op, err)
+			}
+
+			return sshChannelToAsciicast(ctx, reqScanner, msgScanner, tmp, options...)
+		default:
+			if chs.SessionProgram == "" {
+				return nil, fmt.Errorf("%s: session program not set for asciicast conversion", op)
+			}
+			return nil, fmt.Errorf("%s: unsupported %q session program for asciicast conversion", op, chs.SessionProgram)
 		}
 
-		msgScanner, err := ch.OpenMessageScanner(ctx, bsr.Outbound)
-		if err != nil {
-			return nil, fmt.Errorf("%s: %w", op, err)
-		}
-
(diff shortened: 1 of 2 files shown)


internal/bsr/convert/convert.go
```diff
@@ -55,20 +55,28 @@ func ToAsciicast(ctx context.Context, session *bsr.Session, tmp storage.TempFile
...
-		if err != nil {
-			return nil, fmt.Errorf("%s: %w", op, err)
+			msgScanner, err := ch.OpenMessageScanner(ctx, bsr.Outbound)
+			if err != nil {
+				return nil, fmt.Errorf("%s: %w", op, err)
+			}
+
+			return sshChannelToAsciicast(ctx, reqScanner, msgScanner, tmp, options...)
+		default:
+			if chs.SessionProgram == "" {
```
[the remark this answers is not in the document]
elimt: Yeah, I agree. That looks cleaner. Moved it as case statement

internal/bsr/convert/convert.go
```diff
@@ -55,20 +55,27 @@ func ToAsciicast(ctx context.Context, session *bsr.Session, tmp storage.TempFile
 			return nil, fmt.Errorf("%s: %w", op, err)
 		}
 
-		// TODO sanity checks before getting the data files:
-		// - check connection summary to see if there was an exec or shell request
+		chs := ch.Summary.(*ssh.ChannelSummary)
```
tmessi, line 58: This should either check the type assertion, i.e:

```golang
chs, ok := ch.Summary.(*ssh.ChannelSummary)
if !ok {
  // return error, probably indicating some like an unexpected or malformed bsr.
}
```

```go
switch chs := ch.Summary.(type) {
case *ssh.ChannelSummary:
  switch chs.SessionProgram {
  // rest of switch
default:
  // return an error, probably indicating some like an unexpected or malformed bsr.
```

tmessi: Also, can you add a test case of a BSR that is malformed in this way?

internal/bsr/convert/convert_test.go
```diff
@@ -0,0 +1,332 @@
...
+		enc, err := bsr.NewChunkEncoder(ctx, w, bsr.NoCompression, bsr.NoEncryption)
+		require.NoError(t, err)
+
+		for _, c := range chunks {
+			_, err := enc.Encode(ctx, c)
+			require.NoError(t, err)
+		}
+	}
+
+	cases := []struct {
```
tmessi, line 44: Can you include test cases for all of the [valid ssh.SessionPrograms](https://github.com/hashicorp/boundary/blob/456c5dbb4d6c64886abcacb425b7faf5a39de673/internal/bsr/ssh/types.go#L39-L48). It looks like this already has some for `ssh.Exec`, `ssh.Shell` and `ssh.Sumsystem`, so can you add ones for `ssh.None` and `ssh.NotApplicable`?

internal/bsr/convert/convert.go
```diff
@@ -55,20 +55,27 @@ func ToAsciicast(ctx context.Context, session *bsr.Session, tmp storage.TempFile
...
+				return nil, fmt.Errorf("%s: %w", op, err)
+			}
 
-		msgScanner, err := ch.OpenMessageScanner(ctx, bsr.Outbound)
-		if err != nil {
-			return nil, fmt.Errorf("%s: %w", op, err)
+			msgScanner, err := ch.OpenMessageScanner(ctx, bsr.Outbound)
+			if err != nil {
+				return nil, fmt.Errorf("%s: %w", op, err)
+			}
```
tmessi, line 70: Can you add some tests that cover the error cases for these?

tmessi (changes requested):
Looking pretty good.

ddebko (approved):
This looks good to me.

tmessi (approved):

irenarindos (approved):

PR #3483, opened 2023-07-19
docs: Reorg session recording operations

Dan-Heath:
During Boundary Docs Day, we added some new content to the session recording topics for Validating recordings and Validating the data store. We added them to the Operations topic, since this is ongoing work a user may be required to perform after initial setup. But I think they actually warrant their own topics, and would be easier to find if we put them on the TOC. 

This PR performs a minor reorganization of the Operations topics for session recording. It creates a new Overview sub-topic, and breaks out the two validation topics to be standalone (I did not rewrite anything, this is just moving pieces of content around):

image

I changed the title of the section for validating data in the external store from "Validate the data integrity in the external object store" to "How Boundary validates data integrity in the external object store," since the user doesn't perform any action in that topic.

There were also some unrelated Docs Day updates to the [Manage session recordings](https://developer.hashicorp.com/boundary/docs/operations/manage-recorded-sessions) topic to add additional instructions for viewing and downloading recordings. Since this PR moves the validation sections out of that topic, and the remaining content focuses on finding and viewing recordings, I'd also like to change the topic title to **Find and view recorded sessions**, which I think is more descriptive than **Manage session recordings**.

These changes should make it easier to find this session recording content via the TOC and search (both from within the help system and external searches thanks to improved SEO).

[View the update in the preview deployment.](https://boundary-296lk1em3-hashicorp.vercel.app/boundary/docs/operations/session-recordings)


stellarsquall (approved):
I like this approach, and this it improves discoverability and usability 👍

AdamBouhmad (approved):

AdamBouhmad:
Looks great -- super clear re-org. Thank you @Dan-Heath!!

PR #3498, opened 2023-07-25
fix(bsr): allow reading twice from the same file

ddebko:
# Summary:

This adjusts the behavior of a memFile so that we can open the same file multiple times and re-read the file multiple times.

## Note:

This issue was resolved, a deep copy was implemented to support concurrent reads:
```
I wasn't sure if I should return a deepCopy of MemFile when calling OpenFile. With the current implementation, calling
OpenFile twice will reset the first reference. This fix assumes that calling OpenFile will not happen concurrently.
```

### DIFF
--- a/internal/bsr/internal/fstest/fs.go
+++ b/internal/bsr/internal/fstest/fs.go
@@ -142,28 +142,44 @@ func (m *MemContainer) OpenFile(_ context.Context, n string, option ...storage.O
 		return nil, fmt.Errorf("cannot create file in read-only mode: %w", ErrReadOnly)
 	}
 
-	var f *MemFile
+	var src, dst *MemFile
 
 	if opts.WithCreateFile {
 		// create or truncate just like os.Create
-		f = &MemFile{
+		src = &MemFile{
 			Buf: bytes.NewBuffer([]byte{}),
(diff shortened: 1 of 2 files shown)


tmessi:
> I wasn't sure if I should return a deepCopy of MemFile when calling OpenFile. With the current implementation, calling OpenFile twice will reset the first reference. This fix assumes that calling OpenFile will not happen concurrently.

I think a deep cop of the MemFile when calling OpenFile would be a good idea. We should also do some locking so it can safely happen concurrently.

internal/bsr/internal/fstest/fs.go
```diff
@@ -142,28 +142,44 @@ func (m *MemContainer) OpenFile(_ context.Context, n string, option ...storage.O
 		return nil, fmt.Errorf("cannot create file in read-only mode: %w", ErrReadOnly)
```
irenarindos, line 142: Can you update the function doc for OpenFile to explain how opts.WithCreateFile works with the src and dst files?
(I think the supported ops is also out of date)

irenarindos (approved):

tmessi (approved):

PR #3500, opened 2023-07-26
fix(bsr): limit chunk length

irenarindos:
Sets a max limit on chunk length to prevent unbounded memory allocation during decoding.

### DIFF
--- a/internal/bsr/chunk.go
+++ b/internal/bsr/chunk.go
@@ -20,7 +20,9 @@ const (
 	crcSize       = 4
 
 	// MaxChunkLength sets an upper bound on BSR chunk lengths
-	MaxChunkLength = 256 * 1024
+	// Default to 64MB, as this is the limit for protobufs, which back our chunks
+	// TODO: Should this be configurable as an option?
+	MaxChunkLength = 64 * 100 * 1000
 )
 
 // Chunk Types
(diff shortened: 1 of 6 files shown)


elimt (approved):

tmessi (changes requested):

ddebko (approved):

irenarindos:
@tmessi changed the limit to 64MB and only enforced on decode, as we talked about

internal/bsr/chunk.go
```diff
@@ -18,6 +18,11 @@ const (
 	crcDataSize   = protocolSize + chunkTypeSize + directionSize + timestampSize
 	chunkBaseSize = lengthSize + crcDataSize
 	crcSize       = 4
+
+	// MaxChunkLength sets an upper bound on BSR chunk lengths
+	// Default to 64MB, as this is the limit for protobufs, which back our chunks
+	// TODO: Should this be configurable as an option?
+	MaxChunkLength = 64 * 100 * 1000
```
tmessi, line 25: I still think this should probably be named something like `MaxChunkDataLength`.

tmessi (commented):
One comment, otherwise looks good to me.

tmessi (approved):

PR #3504, opened 2023-07-26
feat(bsr): support container validation

ddebko:
# Summary:

This PR adds a new method for the bsr *container called `ValidateChecksums`. This method calculates the checksums of all files that are listed in the SHA256SUM file and compares the expected value. Here is an example output from a session container:

```
ChecksumValidation{
  "bsrKey.pub": &FileChecksumValidation{
	  Filename: "bsrKey.pub",
	  Passed:   false,
	  Error:  error("checksum mismatch"),
  },
  "pubKeyBsrSignature.sign": &FileChecksumValidation{
	  Filename: "pubKeyBsrSignature.sign",
	  Passed:   true,
  },
  "pubKeySelfSignature.sign": &FileChecksumValidation{
	  Filename: "pubKeySelfSignature.sign",
	  Passed:   true,
  },
  "wrappedBsrKey": &FileChecksumValidation{
	  Filename: "wrappedBsrKey",
	  Passed:   true,
  },
  "wrappedPrivKey": &FileChecksumValidation{
	  Filename: "wrappedPrivKey",
	  Passed:   true,
  },
  "session-meta.json": &FileChecksumValidation{
	  Filename: "session-meta.json",
	  Passed:   true,
  },
  "session-recording.meta": &FileChecksumValidation{
	  Filename: "session-recording.meta",
	  Passed:   true,
  },
  "session-recording-summary.json": &FileChecksumValidation{
	  Filename: "session-recording-summary.json",
	  Passed:   true,
  },
}
```

### DIFF
--- a/internal/bsr/container.go
+++ b/internal/bsr/container.go
@@ -33,6 +33,30 @@ const (
 // ContainerType defines the type of container.
 type ContainerType string
 
+// FileChecksumValidation is a validation report on a file's checksum value
+type FileChecksumValidation struct {
+	Filename string
+	Passed   bool
+	Error    error
+}
+
+// ChecksumValidation is a map where the key is a file name
+// and the value contains a validation report on whether or
+// not the file matches its expected checksum
+type ChecksumValidation map[string]*FileChecksumValidation
+
+// GetFailedItems returns a filtered map of FileChecksumValidation that have failed
+func (cv ChecksumValidation) GetFailedItems() ChecksumValidation {
+	failedValidations := ChecksumValidation{}
+	for fileName, validation := range cv {
+		if validation.Passed {
+			continue
+		}
+		failedValidations[fileName] = validation
+	}
+	return failedValidations
+}
+
 // Valid container types.
 const (
 	SessionContainer    ContainerType = "session"
@@ -47,9 +71,9 @@ type container struct {
 
 	// Fields primarily used for writing
 	journal    *journal.Journal
-	sumName    string
 	meta       *checksum.File
 	sum        *checksum.File
+	sumName    string
 	sumEncoder *json.Encoder
 	checksums  *sign.File
 	sigs       storage.File
@@ -513,4 +537,43 @@ func (c *container) close(_ context.Context) error {
 	}
 
 	return closeError
+}
+
+// ValidateChecksums iterates over all files in the SHA256SUM
+// file and verifies that each file checksum is as expected.
+//
+// This function expects that the container's kms keys
+// are loaded into memory and the signature files are
+// verified.
+func (c *container) ValidateChecksums(ctx context.Context) (ChecksumValidation, error) {
+	const op = "bsr.(container).Validate"
+	if len(c.shaSums) == 0 {
+		return nil, fmt.Errorf("%s: missing checksums", op)
+	}
+	if c.keys == nil {
+		return nil, fmt.Errorf("%s: missing keys", op)
+	}
+	checksumValidation := make(ChecksumValidation, len(c.shaSums))
+	for fileName, expectedChecksum := range c.shaSums {
+		report := &FileChecksumValidation{
+			Filename: fileName,
+		}
+		checksumValidation[fileName] = report
+		f, err := c.container.OpenFile(ctx, fileName)
+		if err != nil {
+			report.Error = err
+			continue
+		}
+		actualChecksum, err := crypto.Sha256Sum(ctx, f, crypto.WithHexEncoding(true))
+		if err != nil {
+			report.Error = err
+			continue
+		}
+		if !bytes.Equal(expectedChecksum, actualChecksum) {
+			report.Error = fmt.Errorf("checksum mismatch")
+			continue
+		}
+		report.Passed = true
+	}
+	return checksumValidation, nil
 }
--- a/internal/bsr/container.go
+++ b/internal/bsr/container.go
@@ -559,12 +559,7 @@ func (c *container) ValidateChecksums(ctx context.Context) (ContainerChecksumVal
 			Filename: fileName,
 		}
 		checksumValidation[fileName] = report
-		f, err := c.container.OpenFile(ctx, fileName)
-		if err != nil {
-			report.Error = err
-			continue
-		}
-		actualChecksum, err := crypto.Sha256Sum(ctx, f, crypto.WithHexEncoding(true))
+		actualChecksum, err := c.computeFileChecksum(ctx, fileName, crypto.WithHexEncoding(true))
 		if err != nil {
 			report.Error = err
 			continue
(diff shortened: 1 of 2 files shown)


internal/bsr/container.go
```diff
@@ -514,3 +545,40 @@ func (c *container) close(_ context.Context) error {
...
+		actualChecksum, err := crypto.Sha256Sum(ctx, f, crypto.WithHexEncoding(true))
+		if err != nil {
+			checksumValidations[bsrFileName] = BSRChecksumStatus(fmt.Sprintf(bsrFailedCheckTemplate, err.Error()))
+			continue
+		}
+		if bytes.Equal(expectedChecksum, actualChecksum) {
+			checksumValidations[bsrFileName] = BSRCheckPass
+			continue
+		}
+		checksumValidations[bsrFileName] = BSRChecksumStatus(fmt.Sprintf(bsrFailedCheckTemplate, "checksum mismatch"))
```
irenarindos, line 581: I'm wondering if this should be !bytes.Equal with the checksum mismatch error status, and if the fallthrough should be the BSRCheckPass case for readability

internal/bsr/container.go
```diff
@@ -28,11 +28,39 @@ const (
...
 )
 
 // ContainerType defines the type of container.
 type ContainerType string
 
+// BSRFileName defines the name of a bsr file.
+type BSRFileName string
+
+// BSRChecksumStatus defines the status of the checksum validation
+type BSRChecksumStatus string
```
irenarindos, line 46: I think this should be a struct that has uses status enum value and an error field; if the status enum is FAILED, then GetFailedItems can get the error field for the specifics on why it failed

internal/bsr/container.go
```diff
@@ -28,11 +28,39 @@ const (
 	checksumFileName        = "SHA256SUM"
 	sigFileName             = "SHA256SUM.sig"
 	journalFileName         = ".journal"
+	bsrFailedCheckTemplate  = "FAILED: %s"
+)
+
+// Valid BSR checksum status values.
+const (
+	BSRCheckPass BSRChecksumStatus = "OK"
```
irenarindos, line 36: I feel like this should be an enum and FAILED should be one of the BSRChecksumStatus values.

internal/bsr/container.go
ddebko: I like this idea. 

@elimt Would you like GetFailedItems to return a map[string]string or would you prefer a map[string]struct, incase you need to handle any logic dealing with the status output.

internal/bsr/testing.go
```diff
@@ -3,13 +3,67 @@
...
+}
+
+// TestWriter should be used for testing a RequestWriter & MessageWriter
+func TestWriter(t *testing.T, w io.Writer) {
+	n, err := w.Write([]byte("hello world"))
+	require.NoError(t, err)
+	require.Equal(t, len("hello world"), n)
+	closer := w.(io.Closer)
+	require.NoError(t, closer.Close())
+}
```
tmessi, line 65: I am not sure this is a useful test helper. This is probably sufficient for the tests you are adding here, since the contents of the files are not important, but any external package looking to use this test helper would not get something that is usable as a request writer or message writer.

internal/bsr/container.go
```diff
@@ -28,11 +28,39 @@ const (
...
 
+// BSRFileName defines the name of a bsr file.
+type BSRFileName string
+
+// BSRChecksumStatus defines the status of the checksum validation
+type BSRChecksumStatus string
+
+// BSRChecksumValidation defines the checksum validation status for a set of files.
+// The key is a file name and the value is the checksum status.
+type BSRChecksumValidation map[BSRFileName]BSRChecksumStatus
```
tmessi, line 50: I think we might want the value in this map to be more than just a status. I think there are a few ways this could be done, but in my mind the goal would be to separate the string output from the internal structure. So for example this could become something like:

```go
type FileChecksumValidation struct {
  Filename string
  Passed bool
  Error error
}

type ChecksumValidation map[string]FileChecksumValidation
```

Then the actual output where things like "OK" or "FAILED: <reason>" is displayed could be handled by the cli.

internal/bsr/testing.go
```diff
@@ -3,13 +3,67 @@
...
+		},
+	}
+}
+
+func TestChannelRecordingMeta(s string) *ChannelRecordingMeta {
+	return &ChannelRecordingMeta{
+		Id:   s,
+		Type: "chan",
+	}
+}
```
tmessi, line 56: Similarly I am also not sure if these are helpful test helpers. In particular I am not sure if these are good defaults for external tests to use, and the helpers don't allow for easily providing alternatives, instead the code would need to set fields on the returned struct, which in some cases can not be set since the fields are not exported.

internal/bsr/container.go
elimt: We wanted to support verbose mode and non-verbose mode. Using `GetFailedItems` to get the error field/specifics will work for the non-verbose mode.

For the verbose mode, if we want to return the status of every file do we expect the status to be strings like "failed" or will we return the status enum?

tmessi (changes requested):

internal/bsr/container.go
ddebko: I don't think we want to return the status enum because that will only look like a integer to the user, unless we add a String() method to the enum. Maybe we need a proto definition to share between the validation functions.

internal/bsr/testing.go
ddebko: I can change the helper function so that the parameter includes a `[]byte` type that gets passed into the Write function. This can help make it more useful. I can also add a comment explaining that the bytes needs to be a bsr chunk type.

tmessi: I think I would prefer just removing the test helper until there is obvious need/use for a helper that can encapsulate some complex and common test setup.

internal/bsr/container.go
tmessi: I think the distinction between verbose and non-verbose mode is something for the cli layer to consider, and the bsr package should just return a structure that contains the full results in a way that is easily used programmatically (see my [comment here](https://github.com/hashicorp/boundary/pull/3504#discussion_r1275267614)). So the functionality of `GetFailedItems` to select just the files that failed, along with the logic of combining the status with the error would be something done in the cli.

internal/bsr/container.go
```diff
@@ -33,6 +33,45 @@ const (
...
+			status += fmt.Sprintf("%s: %v", status, f.Error)
+		}
+	}
+	return fmt.Sprintf("%s: %s", f.Filename, status)
+}
+
+// ChecksumValidation is a map where the key is a file name
+// and the value contains a validation report on whether or
+// not the file matches it's expected checksum
+type ChecksumValidation map[string]*FileChecksumValidation
```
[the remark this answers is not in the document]
elimt: Do we need a map for this if the key will always be the filename? 
If we have the filename in `FileChecksumValidation`, couldn't this just be a slice of `FileChecksumValidation`

internal/bsr/container.go
```diff
@@ -514,3 +538,42 @@ func (c *container) close(_ context.Context) error {
...
+		f, err := c.container.OpenFile(ctx, fileName)
+		if err != nil {
+			report.Error = err
+			continue
+		}
+		actualChecksum, err := crypto.Sha256Sum(ctx, f, crypto.WithHexEncoding(true))
+		if err != nil {
+			report.Error = err
+			continue
+		}
```
tmessi, line 571: We should call `f.Close()` after reading the contents of the file. It might be helpful to move this logic of opening, reading, and calculating the checksum for a single file into a function, so that you can use a `defer` to ensure the file is closed before moving on to the next file.

internal/bsr/container.go
```diff
@@ -514,3 +538,53 @@ func (c *container) close(_ context.Context) error {
...
+
+// computeFileChecksum will open a file, read its contents, and computes SHA256 message digest
+func (c *container) computeFileChecksum(ctx context.Context, fileName string, opt ...wrapping.Option) (checksum []byte, err error) {
+	var f storage.File
+	f, err = c.container.OpenFile(ctx, fileName)
+	if err != nil {
+		return
+	}
+	defer func() {
+		closeErr := f.Close()
```
ddebko, line 584: wrap this instead

tmessi (approved):

PR #3507, opened 2023-07-27
feat: Validate BSR files

elimt:
Add support for validating all files in a Boundary Session Recording (BSR). All files and sub container files will be verified by comparing it against checksums for each file in `SHA256SUM` file

### Considerations
- Validation will continue even if there's an error encountered during validation. The validation error will be added to the Validation struct `Error` field

### DIFF
--- a/internal/bsr/bsr.go
+++ b/internal/bsr/bsr.go
@@ -284,7 +284,14 @@ func Validate(ctx context.Context, sessionRecordingId string, f storage.FS, keyU
 		return nil, fmt.Errorf("%s: missing key unwrap function: %w", op, ErrInvalidParameter)
 	}
 
-	session, sessionContainerValidation, err := validation.ValidateSession(ctx, sessionRecordingId, f, keyUnwrapFn)
+	session, err := OpenSession(ctx, sessionRecordingId, f, keyUnwrapFn)
+	if err != nil {
+		validationError := fmt.Errorf("%s: failed to retrieve session for %s: %w", op, sessionRecordingId, err)
+		return nil, validationError
+	}
+
+	sessionContainerValidation, err := validation.ValidateSession(ctx, session)
+	validation.SessionRecordingValidation = sessionContainerValidation
 	if err != nil {
 		validation.Valid = false
 		return validation, fmt.Errorf("%s: failed to valid session for %s: %w", op, sessionRecordingId, err)
@@ -292,7 +299,23 @@ func Validate(ctx context.Context, sessionRecordingId string, f storage.FS, keyU
 
 	// Valid all connections under session
 	for connId := range session.Meta.connections {
-		connection, connectionContainerValidation, err := validation.ValidateConnection(ctx, connId, session)
+		// Get connection id
+		connKey := connId
+		lastDotIndex := strings.LastIndex(connId, ".connection")
+		if lastDotIndex == -1 {
+			validation.Valid = false
+			return validation, fmt.Errorf("%s: malformed BSR for: %s", op, connId)
+		}
+		connKey = connId[:lastDotIndex]
+
+		// Open current connection
+		connection, err := session.OpenConnection(ctx, connKey)
+		if err != nil {
+			validation.Valid = false
+			return validation, fmt.Errorf("%s: failed to retrieve connection for %s: %w", op, connId, err)
+		}
+
+		connectionContainerValidation, err := validation.ValidateConnection(ctx, connection)
 		sessionContainerValidation.SubContainers = append(sessionContainerValidation.SubContainers, connectionContainerValidation)
 		if err != nil {
 			validation.Valid = false
@@ -301,7 +324,23 @@ func Validate(ctx context.Context, sessionRecordingId string, f storage.FS, keyU
 
 		// Valid all channels under current connection
 		for chId := range connection.Meta.channels {
-			_, channelContainerValidation, err := validation.ValidateChannel(ctx, chId, connection)
+			// Get channel id
+			chKey := chId
+			lastDotIndex := strings.LastIndex(chId, ".channel")
+			if lastDotIndex == -1 {
+				validation.Valid = false
+				return validation, fmt.Errorf("%s: malformed BSR for: %s", op, chId)
+			}
+			chKey = chId[:lastDotIndex]
+
+			// Open current connection
+			channel, err := connection.OpenChannel(ctx, chKey)
+			if err != nil {
+				validation.Valid = false
+				return validation, fmt.Errorf("%s: failed to retrieve channel for %s: %w", op, chId, err)
+			}
+
+			channelContainerValidation, err := validation.ValidateChannel(ctx, channel)
 			connectionContainerValidation.SubContainers = append(connectionContainerValidation.SubContainers, channelContainerValidation)
 			if err != nil {
 				validation.Valid = false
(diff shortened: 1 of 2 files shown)


internal/bsr/bsr.go
```diff
@@ -237,6 +238,173 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
...
+
+		validation.SubContainer = append(validation.SubContainer, connectionValidation)
+	}
+
+	return validation, validationSummary, nil
+}
+
+// updateValidationSummary updates the ValidationSummary based on the failed checksums.
+// The updated ValidationSummary is then returned
+func updateValidationSummary(ctx context.Context, c ContainerType, id string, cv ChecksumValidation, vs ValidationSummary) ValidationSummary {
```
irenarindos, line 385: What if this was a pointer receiver that updated the underlying ValidationSummary directly?

internal/bsr/bsr.go
```diff
@@ -237,6 +238,173 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
 	return session, nil
 }
 
+type Validation struct {
+	Name               string
+	Type               ContainerType
+	Error              error
```
irenarindos, line 244: I'm wondering if we should use a go-multierror here instead

internal/bsr/bsr.go
```diff
@@ -237,6 +238,173 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
...
+	ChecksumValidation ChecksumValidation
+	SubContainer       []Validation `json:"sub_container,omitempty"`
+}
+
+type ValidationSummary struct {
```
irenarindos, line 249:
```suggestion
type SessionValidationSummary struct {
```

internal/bsr/bsr.go
```diff
@@ -237,6 +238,173 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
...
+	validationSummary = updateValidationSummary(ctx, SessionContainer, sessionRecordingId, sessionChecksumValidation, validationSummary)
+
+	validation.ChecksumValidation = sessionChecksumValidation
+
+	conns := session.Meta.connections
+
+	for connId := range conns {
+		// Get connection id
+		connKey := connId
+		lastDotIndex := strings.LastIndex(connId, ".")
```
irenarindos, line 310:
```suggestion
		lastDotIndex := strings.LastIndex(connId, ".connection")
```
Might make this clearer to others in the future

internal/bsr/bsr.go
```diff
@@ -237,6 +238,173 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
...
+		validationSummary = updateValidationSummary(ctx, ConnectionContainer, connId, connectionChecksumValidation, validationSummary)
+
+		connectionValidation.ChecksumValidation = connectionChecksumValidation
+
+		chs := conn.Meta.channels
+
+		for chId := range chs {
+			// Get channel id
+			chsKey := chId
+			lastDotIndex := strings.LastIndex(chId, ".")
```
irenarindos, line 345:
```suggestion
			lastDotIndex := strings.LastIndex(chId, ".channel")
```
Might make this clearer to others in the future

internal/bsr/bsr.go
```diff
@@ -237,6 +238,173 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
...
+
+	validation.ChecksumValidation = sessionChecksumValidation
+
+	conns := session.Meta.connections
+
+	for connId := range conns {
+		// Get connection id
+		connKey := connId
+		lastDotIndex := strings.LastIndex(connId, ".")
+		if lastDotIndex != -1 {
```
irenarindos, line 311: Shouldn't we error if this is -1? If the connection name in the session meta doesn't end in .connection, then the BSR is malformed

internal/bsr/bsr.go
```diff
@@ -237,6 +238,173 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
...
+
+		connectionValidation.ChecksumValidation = connectionChecksumValidation
+
+		chs := conn.Meta.channels
+
+		for chId := range chs {
+			// Get channel id
+			chsKey := chId
+			lastDotIndex := strings.LastIndex(chId, ".")
+			if lastDotIndex != -1 {
```
irenarindos, line 346: Shouldn't we error if this is -1? If the channel name in the connection meta doesn't end in .channel, then the BSR is malformed

internal/bsr/bsr.go
```diff
@@ -237,6 +238,173 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
...
+
+	for connId := range conns {
+		// Get connection id
+		connKey := connId
+		lastDotIndex := strings.LastIndex(connId, ".")
+		if lastDotIndex != -1 {
+			connKey = connId[:lastDotIndex]
+		}
+
+		connectionValidation := Validation{
```
irenarindos, line 315: Can we break this out into a ValidateConnection function that is called per connection in the session?

internal/bsr/bsr.go
```diff
@@ -237,6 +238,173 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
...
+
+		for chId := range chs {
+			// Get channel id
+			chsKey := chId
+			lastDotIndex := strings.LastIndex(chId, ".")
+			if lastDotIndex != -1 {
+				chsKey = chId[:lastDotIndex]
+			}
+
+			channelValidation := Validation{
```
irenarindos, line 350: Can we break this out in a ValidateChannel function that is called per channel in the connection?

internal/bsr/bsr.go
```diff
@@ -237,6 +238,173 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
...
+	FailedChecksums    map[ContainerType]map[string]ChecksumValidation
+}
+
+// Validate retrieves a BSR from storage using the sessionRecordingId and validates the BSR.
+// All files and sub container files will be verified by comparing it against checksums for
+// each file in SHA256SUM file.
+//
+// Validation will continue even if there's an error encountered during validation.
+// The validation error will be added to the Validation struct Error field.
+func Validate(ctx context.Context, sessionRecordingId string, f storage.FS, keyUnwrapFn kms.KeyUnwrapCallbackFunc) (Validation, ValidationSummary, error) {
```
tmessi, line 261: I don't think we need a separate struct for the ValidationSummary, at least not one that is completely separate from the Validation. Essentially I think one should contain the other, and also some details currently in the summary are not needed, at least as part of the bsr package.

I think instead we want something like:

```go
// Validation provides the results from validating a bsr.
type Validation struct {
  SessionRecordingId         string
  Valid                      bool
  SessionRecordingValidation *ContainerValidation
}

// ContainerValidation contains the results from validating a container in a bsr.
type ContainerValidation struct {
  Name                    string
  ContainerType           ContainerType
  Error                   error
  FileChecksumValidations ContainerChecksumValidation
  Subcontainers           []*ContainerValidation
}

func Validate(ctx context.Context, sessionRecordingId string, f storage.FS, keyUnwrapFn kms.KeyUnwrapCallbackFunc) (*Validation, error) {
  // implementation of validate
}
```

Note that this eliminates the `FailedChecksums` map. I think this is not necessary for the bsr package to track directly. We could provide a method on `Validation` to extract just the failed checksums, but I think we can also leave that as something for the caller of the bsr package to do.

tmessi (changes requested):

internal/bsr/bsr.go
```diff
@@ -237,6 +238,145 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
...
+
+// updateValidationSummary updates the value of "Valid" field in Validation struct based on if there are failed checksums
+func (v *Validation) updateValidationStatus(ctx context.Context, cv *ContainerValidation) {
+	failedChecksums := cv.FileChecksumValidations.GetFailedItems()
+
+	// Update validation filed only if there are failed checksums
+	if len(failedChecksums) > 0 {
+		v.Valid = false
+	}
+}
```
tmessi, line 378: I don't think these should be methods on the Validation struct. In particular the ValidateContainer method being an exported method makes it seems like it could be called from outside this package, but given how it depends on the unexported type `container` it would probably be difficult to do so, and should probably only ever be called by `Validate` anyway.

I think instead, these could be unexported package functions. And the `updateValidationStatus` function could probably just be included in the other function, since it is only called by that function and is rather trivial.

tmessi (changes requested):

internal/bsr/bsr.go
```diff
@@ -237,6 +238,140 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
...
+	}
+
+	// Validate all connections under session
+	for connId := range session.Meta.connections {
+		// Get connection id
+		connKey := connId
+		lastDotIndex := strings.LastIndex(connId, ".connection")
+		if lastDotIndex == -1 {
+			validation.Valid = false
+			return validation, fmt.Errorf("%s: malformed BSR for: %s", op, connId)
```
irenarindos, line 299: I'm wondering if this should be a return or if this error should be set as the error for this connection container- maybe there are other connections that we could return info for?

internal/bsr/bsr.go
```diff
@@ -237,6 +238,140 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
...
+		}
+		connKey = connId[:lastDotIndex]
+
+		// Open current connection
+		connection, err := session.OpenConnection(ctx, connKey)
+		if err != nil {
+			validation.Valid = false
+			return validation, fmt.Errorf("%s: failed to retrieve connection for %s: %w", op, connId, err)
```
irenarindos, line 307: Same question here about this being an error for this connection container

internal/bsr/bsr.go
```diff
@@ -237,6 +238,140 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
...
+		}
+
+		// Validate all channels under current connection
+		for chId := range connection.Meta.channels {
+			// Get channel id
+			chKey := chId
+			lastDotIndex := strings.LastIndex(chId, ".channel")
+			if lastDotIndex == -1 {
+				validation.Valid = false
+				return validation, fmt.Errorf("%s: malformed BSR for: %s", op, chId)
```
irenarindos, line 324: And same thing here for the channel container

internal/bsr/bsr.go
```diff
@@ -237,6 +238,140 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
...
+			}
+			chKey = chId[:lastDotIndex]
+
+			// Open current connection
+			channel, err := connection.OpenChannel(ctx, chKey)
+			if err != nil {
+				validation.Valid = false
+				return validation, fmt.Errorf("%s: failed to retrieve channel for %s: %w", op, chId, err)
```
irenarindos, line 332: And here

irenarindos (commented):
Looks great! Just one comment

internal/bsr/bsr.go
elimt: So basically to continue validating other connections even if 1 connection is malformed or has some issue? My thinking was that if it's malformed then we don't want to continue but I see your point about validating other connections and just saving the error.

internal/bsr/bsr.go
```diff
@@ -237,6 +238,140 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
...
+	case f == nil:
+		return nil, fmt.Errorf("%s: missing storage: %w", op, ErrInvalidParameter)
+	case keyUnwrapFn == nil:
+		return nil, fmt.Errorf("%s: missing key unwrap function: %w", op, ErrInvalidParameter)
+	}
+
+	session, err := OpenSession(ctx, sessionRecordingId, f, keyUnwrapFn)
+	if err != nil {
+		return nil, fmt.Errorf("%s: failed to retrieve session for %s: %w", op, sessionRecordingId, err)
+	}
```
tmessi, line 278: We should close the session at the end of this function, ie:

```suggestion
	session, err := OpenSession(ctx, sessionRecordingId, f, keyUnwrapFn)
	if err != nil {
		return nil, fmt.Errorf("%s: failed to retrieve session for %s: %w", op, sessionRecordingId, err)
	}
	defer session.Close(ctx)
```

internal/bsr/bsr.go
```diff
@@ -237,6 +238,140 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
...
+		connKey := connId
+		lastDotIndex := strings.LastIndex(connId, ".connection")
+		if lastDotIndex == -1 {
+			validation.Valid = false
+			return validation, fmt.Errorf("%s: malformed BSR for: %s", op, connId)
+		}
+		connKey = connId[:lastDotIndex]
+
+		// Open current connection
+		connection, err := session.OpenConnection(ctx, connKey)
```
tmessi, line 304: Likewise we should be calling `connection.Close(ctx)` when we are done with the connection. You might want to refactor the contents of this for loop into a function to make it easier to use `defer` so the connection is closed before iterating to the next connection.

internal/bsr/bsr.go
```diff
@@ -237,6 +238,140 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
...
+			chKey := chId
+			lastDotIndex := strings.LastIndex(chId, ".channel")
+			if lastDotIndex == -1 {
+				validation.Valid = false
+				return validation, fmt.Errorf("%s: malformed BSR for: %s", op, chId)
+			}
+			chKey = chId[:lastDotIndex]
+
+			// Open current connection
+			channel, err := connection.OpenChannel(ctx, chKey)
```
tmessi, line 329: Similarly here, we should call `channel.Close(ctx)`

tmessi (changes requested):

internal/bsr/bsr.go
elimt: When we open a session, [required fields needed](https://github.com/hashicorp/boundary/blob/ef3954c561ec693ce476550fab3e79e8c3068cc3/internal/bsr/container.go#L510) to close the session are `nil` so closing a session/connection/channel results in panics.
Files like `container.journal` or  `container.sigs` are all `nil` when we open a container. Only `container.container` is not nil.

Instead of calling defer session.Close(ctx), perhaps we could call `session.container.container.Close()` to directly close the opened container?

Also, that is one of the things I realized in some parts of the code. We don't close opened sessions, connections or channels.

ddebko (approved):

irenarindos (approved):

tmessi (approved):

PR #3511, opened 2023-07-27
fix (ldap): support only updating the bind-dn and/or bind-password

jimlambrt:


johanbrandhorst (approved):
Something something overeager validation 😂

PR #3513, opened 2023-07-27
SSH and TCP definition for Targets

dannyjknights:
If a user isn't using injected credentials for SSH but sets the target type to SSH, they will not be able to connect to the target. The SSH `type` is reserved for injected credentials setup, which isn't clear in the documentation, but needs to be, as the SSH type doesn't really denote an injected credentials deployment.

### DIFF
--- a/website/content/docs/concepts/domain-model/targets.mdx
+++ b/website/content/docs/concepts/domain-model/targets.mdx
@@ -36,6 +36,9 @@ A target has the following configurable attributes:
 
 - `description` - (optional)
 
+- `type` - (required)
+  Defines the type of connection to a target. Options are `static`, `tcp` and `ssh`. Even if the type of connection to the targer is `ssh`, the `tcp` subcommand should be selected. Using the `ssh` command is reserved for [Credential injection](https://developer.hashicorp.com/boundary/tutorials/hcp-administration/hcp-ssh-cred-injection)
+
 - `address` - (optional)
   This value represents a network resource address and is used when establishing a session.
   It does not accept a port, only an IP address or DNS name.
@@ -97,6 +100,8 @@ SSH targets have the following additional attributes:
 - `storage_bucket_id` - (optional)
   Designates the storage bucket to be used for session recording.
   This attribute is required if you set `enable_session_recording` to `true`.
+
+  If using injected SSH credentials into a Boundary session, the `type` must be set to `ssh` and not `tcp`.
 
 ## Referenced by
 


website/content/docs/concepts/domain-model/targets.mdx
```diff
@@ -36,6 +36,9 @@ A target has the following configurable attributes:
 
 - `description` - (optional)
 
+- `type` - (required)
+  Defines the type of connection to a target. Options are `static`, `tcp` and `ssh`. Even if the type of connection to the targer is `ssh`, the `tcp` subcommand should be selected. Using the `ssh` command is reserved for [Credential injection](https://developer.hashicorp.com/boundary/tutorials/hcp-administration/hcp-ssh-cred-injection)
```
Dan-Heath, line 40:
```suggestion
  Defines the type of connection to a target. Options are `static`, `tcp`, and `ssh`. Even if the type of connection to the target is `ssh`, you should select the `tcp` subcommand. Using the `ssh` command is reserved for [credential injection](https://developer.hashicorp.com/boundary/tutorials/hcp-administration/hcp-ssh-cred-injection)
```

website/content/docs/concepts/domain-model/targets.mdx
```diff
@@ -98,6 +101,8 @@ SSH targets have the following additional attributes:
   Designates the storage bucket to be used for session recording.
   This attribute is required if you set `enable_session_recording` to `true`.
 
+  If using injected SSH credentials into a Boundary session, the `type` must be set to `ssh` and not `tcp`.
```
Dan-Heath, line 104:
```suggestion
  If you inject SSH credentials into a Boundary session, you must set the `type` to `ssh` and not `tcp`.
```

Dan-Heath (changes requested):
Thank you for creating this Danny! I suggested a couple minor copy edits, but overall I think it makes sense based on the conversation in Slack (https://hashicorp.slack.com/archives/C01AQDJF3SA/p1690462004937979).

Dan-Heath (approved):

website/content/docs/concepts/domain-model/targets.mdx
```diff
@@ -36,6 +36,9 @@ A target has the following configurable attributes:
 
 - `description` - (optional)
 
+- `type` - (required)
+  Defines the type of connection to a target. Options are `static`, `tcp`, and `ssh`. Even if the type of connection to the target is `ssh`, you should select the `tcp` subcommand. Using the `ssh` command is reserved for [credential injection](https://developer.hashicorp.com/boundary/tutorials/hcp-administration/hcp-ssh-cred-injection)
```
tmessi, line 40: I believe the only valid target types are `tcp` and `ssh`, not `static`.

Also looking at the other domain model docs pages, we don't tend to include a `type` attribute in the list here. Instead the subtypes are listed below under sub-headings. Perhaps that would be a better place to call out that SSH Targets are required for credential injection, that you can only use SSH Targets if you want to use credential injection?

dannyjknights: Hey @tmessi , thanks for the feedback. Apologies on the 'static' reference. I could have sworn that I read that somewhere but probably got confused. I agree, perhaps under the SSH target attributes HCP/ENT section we could change the wording to:

`"SSH targets can source username/password or SSH private key credentials from Vault [credential libraries](https://developer.hashicorp.com/boundary/docs/concepts/domain-model/credential-libraries) or static [credentials](https://developer.hashicorp.com/boundary/docs/concepts/domain-model/credentials). If using injected crednetials, Boundary then injects the credentials into the SSH session between a client and end host. This allows users to securely connect to remote hosts using SSH, while never being in possession of a valid credential for that target host.

If the target type is set to SSH then brokered or injected credentials has to be used. You cannot reference a static .pem file to use`

Or something like that @Dan-Heath

tmessi: It might help to state clearly the requirements around SSH Targets and injected application credentials:

- An SSH Target **must** have at least one injected application credential.
- A TCP Target **can not** have any injected application credentials.

Therefore:
- If you want to use brokered credentials when connecting to a target running SSH, you must use a TCP Target.
- If you want to use injected application credentials when connecting to a target running SSH, you must use an SSH Target.
- Also if you want to enable session recording for a target running SSH, you must use injected application credentials and an SSH Target.

website/content/docs/concepts/domain-model/targets.mdx
```diff
@@ -81,9 +95,16 @@ TCP targets have the following additional attribute:
 ### SSH target attributes <sup>HCP/ENT</sup>
 
 SSH targets can source username/password or SSH private key credentials from Vault [credential libraries][] or static
-[credentials][]. Boundary then injects credentials into the SSH session between a client and end host. This allows users to
+[credentials][]. If you use injected credentials, Boundary injects credentials into the SSH session between a client and end host. Injected credentials allow users to
```
tmessi, line 98: Both the new wording and the previous wording seems a little confusing to me. We might want to make it clear again here that injected application credentials are required for an SSH target. Perhaps something like:

```
SSH targets use injected application credentials to authenticate
an SSH session between the client and end host.
This allows users to securely connect to remote hosts using SSH,
while never being in the possession of a valid credential for that target host.
The injected credentials can be
a username/password or username/private key credential
from Vault [credential libraries][] or static [credentials][],
or a SSH certificate from Vault [SSH certificate credential libraries][].
```

website/content/docs/concepts/domain-model/targets.mdx
```diff
@@ -81,9 +95,16 @@ TCP targets have the following additional attribute:
...
 securely connect to remote hosts using SSH, while never being in possession of a valid credential for that target host.
 
+<Note>
+
+If you set the target `type` to `ssh`, you must use brokered or injected credentials.
+You cannot reference a static .pem file.
+
+</Note>
+
```
tmessi, line 107: This note is still confusing to me. Again an SSH target requires an injected application credential. You can also have brokered credentials in addition to the injected ones, but it is not a "borkered or injected".

dannyjknights: Yes this isn't right based on our conversation. Something like Tim put previously with a slight edit:

<Note>

If you want to use brokered credentials or a static .pem file when connecting to a target running SSH, you must set the target type to TCP

If you want to use injected application credentials when connecting to a target running SSH, you must set the target type to SSH

 </Note>

Dan-Heath:
Thank you both for the feedback! Apologies for my confusion. I revised the paragraph under **SSH target attributes** using Tim’s suggested text. I deleted the Note altogether, since it just rephrased the text under **Target types** (but please let me know if you disagree, I can add it back).

Is the text under **Target types** OK? I am not sure if there's a difference between "a target running SSH" and "an SSH target" in that first bullet. I think the second is easier to read, but I'm not sure if I changed the meaning from the original suggestion.

tmessi:
> Is the text under **Target types** OK? I am not sure if there's a difference between "a target running SSH" and "an SSH target" in that first bullet. I think the second is easier to read, but I'm not sure if I changed the meaning from the original suggestion.

There might be a better way to phrase it, but I was purposefully looking to avoid the use of "SSH target" in that sentence. The idea was to note that a user is trying to connect to an SSH service running on a target (or perhaps more accurately a host), and they can so that with either a TCP Target or an SSH Target, depending on the other details like how the credentials are issued, i.e. brokered or injected or wanting to have session recording enabled.

Dan-Heath:
> There might be a better way to phrase it, but I was purposefully looking to avoid the use of "SSH target" in that sentence. The idea was to note that a user is trying to connect to an SSH service running on a target (or perhaps more accurately a host), and they can so that with either a TCP Target or an SSH Target, depending on the other details like how the credentials are issued, i.e. brokered or injected or wanting to have session recording enabled.

Got it, thanks Tim! That makes sense. I mainly rewrote that part b/c non-native English readers can struggle with _-ing_ words. I changed those bullets to read "a target that runs SSH" in the latest commit. I am OK with changing it back to the ing form though, if that's not accurate.

tmessi (approved):

PR #3514, opened 2023-07-28
docs: Add note regarding HCP workers

Dan-Heath:
The `initial_upstreams` parameter is not used for HCP workers, but this has caused some confusion with users. ([Via Slack](https://hashicorp.slack.com/archives/C01AQDJF3SA/p1690549357466249)) This PR adds clarification that HCP users should configure workers with `hcp_boundary_cluster_id` instead.

[See the update in the preview deployment.](https://boundary-chlj6cxfl-hashicorp.vercel.app/boundary/docs/configuration/worker#initial_upstreams)

### DIFF
--- a/website/content/docs/configuration/worker/index.mdx
+++ b/website/content/docs/configuration/worker/index.mdx
@@ -74,6 +74,9 @@ worker {
   using env or file, their contents must formatted as a JSON array:
   `["127.0.0.1", "192.168.0.1", "10.0.0.1"]`
 
+  HCP Boundary workers require the `hcp_boundary_cluster_id` parameter instead of `initial upstreams`.
+  If you configure an HCP worker with `initial_upstreams`, the worker configuration fails.
+
 - `tags` - A map of key-value pairs where values are an array of strings. Most
   commonly used for [filtering](/boundary/docs/concepts/filtering) targets a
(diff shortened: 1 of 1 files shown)


website/content/docs/configuration/worker/index.mdx
```diff
@@ -74,6 +74,9 @@ worker {
   using env or file, their contents must formatted as a JSON array:
   `["127.0.0.1", "192.168.0.1", "10.0.0.1"]`
 
+  HCP Boundary workers require the [`hcp_boundary_cluster_id`](/boundary/docs/configuration/worker/#hcp_boundary_cluster_id) parameter instead of `initial upstreams`.
+  If you configure an HCP worker with `initial_upstreams`, the worker configuration fails.
+
```
irenarindos, line 79: Looks great!

I'm wondering if we should move the section below on hcp_boundary_cluster_id to come before tags, and if we should change the wording in that section to remove the "can" so it doesn't sound optional

Dan-Heath: Yes! I like both ideas. Good catch on the wording for `hcp_boundary_cluster_id`. I just added those updates.

irenarindos (approved):

PR #3518, opened 2023-07-31
 client daemon: caching POC (do not merge)

jimlambrt:
Just a draft to easily see diffs from main


jimlambrt:
Moved all of this to a LLB.

PR #3519, opened 2023-07-31
fix: Close opened BSR files

elimt:
- Add checks to close opened files in a container if not nil
- Close files opened during reads for `OpenSession`, `OpenConnection` & `OpenChannel`
- Close opened scanner readers

### Considerations
Update `fs.MemFS` to support returning the original files instead of copies when the `withOriginalOption` is passed. Had to do this to get the current `Closed` state of each file. Because all changes were just editing a copy, `fs.MemFS` was not getting updated with the true closed state of the fil

### DIFF
--- a/internal/bsr/bsr.go
+++ b/internal/bsr/bsr.go
@@ -200,8 +200,11 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
 	// Load and verify recording metadata
 	sha256Reader, err := crypto.NewSha256SumReader(ctx, cc.metaFile)
 	if err != nil {
+		cc.metaFile.Close()
 		return nil, fmt.Errorf("%s: %w", op, err)
 	}
+	defer sha256Reader.Close()
+
 	meta, err := decodeSessionRecordingMeta(ctx, sha256Reader)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


irenarindos (approved):
Thanks for this!

internal/bsr/bsr.go
```diff
@@ -200,8 +200,11 @@ func OpenSession(ctx context.Context, sessionRecordingId string, f storage.FS, k
 	// Load and verify recording metadata
 	sha256Reader, err := crypto.NewSha256SumReader(ctx, cc.metaFile)
 	if err != nil {
+		cc.metaFile.Close()
 		return nil, fmt.Errorf("%s: %w", op, err)
 	}
+	defer sha256Reader.Close()
```
ddebko, line 206: I think this Close() function will return an error and we will lose that context with the defer. I think that we should create an anonymous function for this defer, like so:
```
defer func() {
if closeErr := sha256Reader.Close(); closeErr != nil {
    // wrap or join closeErr with err
}
}()
```
In order to keep the context of returning the closeErr from the anonymous function, we will need to utilize named return values.

ddebko (changes requested):

tmessi (approved):

ddebko (approved):

PR #3521, opened 2023-07-31
Backport of server: allow release version to be set in test into release/0.13.x

hc-github-team-secure-boundary:
## Backport

This PR is auto-generated from #3517 to be assessed for backporting due to the inclusion of the label backport/0.13.x.

**WARNING** automatic cherry-pick of commits failed. Commits will require human attention.

The below text is copied from the body of the original PR.

---

The old logic made it impossible for the caller to set the release version with WithReleaseVersion. Only
set the release version when it's not explicitly set already.


johanbrandhorst:
Oh boy nope

PR #3525, opened 2023-07-31
telemetry: accounts proto

sepehrfrgh:
This PR adds observation gotags to proper fields for telemetry purposes and modifies remove-gotags-comment to remove observation gotags from comments 
cc: @jimlambrt @ajayreshc

### DIFF
--- a/scripts/remove-gotags-comments/main.go
+++ b/scripts/remove-gotags-comments/main.go
@@ -39,6 +39,10 @@ func run(swaggerPath string) error {
 			//   - When gotags appears on its own.
 			//   - When gotags appears at the end of another comment (preceded by \n\n).
 			for _, prefix := range []string{"\\n\\n", ""} {
+				// The two cases we're replacing gotags with empty message:
+				//   - When class gotags is set with eventstream gotags.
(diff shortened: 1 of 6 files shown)


scripts/remove-gotags-comments/main.go
```diff
@@ -39,6 +39,7 @@ func run(swaggerPath string) error {
 			//   - When gotags appears on its own.
 			//   - When gotags appears at the end of another comment (preceded by \n\n).
 			for _, prefix := range []string{"\\n\\n", ""} {
+				swaggerBytes = bytes.ReplaceAll(swaggerBytes, []byte(fmt.Sprintf("%s@gotags: %sclass:\\\"%s\\\" eventstream:\\\"observation\\\"%s", prefix, wrapper, classification, wrapper)), nil)
```
johanbrandhorst, line 42: Maybe add a comment inside the loop to say that we need to cover both the case where the class is set on its own, and where it's set with the eventstream?

jimlambrt (approved):
Add so it begins!  Looks good.

ajayreshc (approved):
LGTM

PR #3526, opened 2023-07-31
Move files to new directories and add persona table

talanknight:
The goal of this PR is to get agreement on the package placement and the persona + 1 example resource tables.

Future work with the personas includes using them in the refresh ticker and in the request handler.

### DIFF
--- a/internal/daemon/cache/repository_test.go
+++ b/internal/daemon/cache/repository_test.go
@@ -81,7 +81,6 @@ func TestRepository_LookupPersona(t *testing.T) {
 		assert.NoError(t, err)
 		assert.NotNil(t, p)
 	})
-
 }
 
 func TestRepository_RefreshTargets(t *testing.T) {


internal/daemon/cache/repository_test.go
```diff
@@ -0,0 +1,148 @@
...
+	assert.ErrorContains(t, err, "not found")
+	assert.Nil(t, gotP)
+
+	p, err = r.LookupPersona(ctx, addr+"0", p.TokenName)
+	assert.NoError(t, err)
+	assert.NotNil(t, p)
+	t.Logf("got %#v", p)
+}
+
+func TestRepository_RefreshTargets(t *testing.T) {
```
jimlambrt, line 49: Perhaps add a test that "saves" an error?

jimlambrt (changes requested):
Looks great.  Just a couple really minor bits.

jimlambrt (approved):

PR #3529, opened 2023-08-01
internal/daemon/controller/handlers/sessions: fix dropped error

alrs:
This fixes a dropped `err` variable in `internal/daemon/controller/handlers/sessions`.


johanbrandhorst (approved):
Good spot, thank you for this!

PR #3533, opened 2023-08-02
Telemetry: auth_methods proto

sepehrfrgh:
This PR adds observation gotags to auth_methods for telemetry purposes
cc: @jimlambrt @ajayreshc

### DIFF
--- a/internal/proto/controller/api/resources/authmethods/v1/auth_method.proto
+++ b/internal/proto/controller/api/resources/authmethods/v1/auth_method.proto
@@ -18,10 +18,10 @@ option (custom_options.v1.domain) = "auth";
 // AuthMethod contains all fields related to an Auth Method resource
 message AuthMethod {
   // Output only. The ID of the Auth Method.
-  string id = 10; // @gotags: `class:"public"`
+  string id = 10; // @gotags: `class:"public" eventstream:"observation"`
 
   // The ID of the Scope of which this Auth Method is a part.
-  string scope_id = 20 [json_name = "scope_id"]; // @gotags: `class:"public"`
+  string scope_id = 20 [json_name = "scope_id"]; // @gotags: `class:"public" eventstream:"observation"`
 
   // Output only. Scope information for this Auth method.
   resources.scopes.v1.ScopeInfo scope = 30;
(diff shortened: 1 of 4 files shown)


internal/proto/controller/api/services/v1/auth_method_service.proto
```diff
@@ -194,7 +194,7 @@ message LdapLoginAttributes {
 
 message AuthenticateRequest {
   // The ID of the Auth Method in the system that should be used for authentication.
-  string auth_method_id = 1 [json_name = "auth_method_id"]; // @gotags: `class:"public"`
+  string auth_method_id = 1 [json_name = "auth_method_id"]; // @gotags: `class:"public" eventstream:"observation"`
```
sepehrfrgh, line 197: @jimlambrt I was unsure if we should include these `AuthenticateRequest` & `AuthenticateResponse` to this PR. What do you think?

internal/proto/controller/api/resources/authmethods/v1/auth_method.proto
```diff
@@ -101,7 +101,7 @@ message PasswordAuthMethodAttributes {
       this: "attributes.min_login_name_length"
       that: "MinLoginNameLength"
     }
-  ]; // @gotags: `class:"public"`
+  ]; // @gotags: `class:"public" eventstream:"observation"`
```
sepehrfrgh, line 104: @jimlambrt I was unsure about these 2 fields to see if we should include them or not. `min_login_name_length` & `min_password_length`

jimlambrt: I would say: nope.  It's customer input.

internal/proto/controller/api/resources/authmethods/v1/auth_method.proto
```diff
@@ -111,7 +111,7 @@ message PasswordAuthMethodAttributes {
       this: "attributes.min_password_length"
       that: "MinPasswordLength"
     }
-  ]; // @gotags: `class:"public"`
+  ]; // @gotags: `class:"public" eventstream:"observation"`
```
jimlambrt, line 114: probably not:  it's customer input.

internal/proto/controller/api/services/v1/auth_method_service.proto
```diff
@@ -96,17 +96,17 @@ service AuthMethodService {
...
   resources.authmethods.v1.AuthMethod item = 1;
 }
 
 message ListAuthMethodsRequest {
-  string scope_id = 1 [json_name = "scope_id"]; // @gotags: `class:"public"`
-  bool recursive = 20 [json_name = "recursive"]; // @gotags: `class:"public"`
-  string filter = 30 [json_name = "filter"]; // @gotags: `class:"public"`
+  string scope_id = 1 [json_name = "scope_id"]; // @gotags: `class:"public" eventstream:"observation"`
+  bool recursive = 20 [json_name = "recursive"]; // @gotags: `class:"public" eventstream:"observation"`
+  string filter = 30 [json_name = "filter"]; // @gotags: `class:"public" eventstream:"observation"`
```
jimlambrt, line 109: I'm not sure about this one, given that it could contain anything the user types.

internal/proto/controller/api/services/v1/auth_method_service.proto
jimlambrt: I think, we could be okay as long as we're careful about which fields are included.

jimlambrt (changes requested):

internal/proto/controller/api/services/v1/auth_method_service.proto
jimlambrt: I think, we can include `recursive` if we want, since it's not something users type in.... but whether or not we need it: I'll leave that for you and PM to decide.

jimlambrt (approved):

ddebko (approved):

johanbrandhorst (approved):

PR #3534, opened 2023-08-02
Telemetry: groups proto

sepehrfrgh:
This PR adds observation tags to group proto for telemetry purposes.
cc: @jimlambrt @ajayreshc

### DIFF
--- a/internal/proto/controller/api/resources/groups/v1/group.proto
+++ b/internal/proto/controller/api/resources/groups/v1/group.proto
@@ -14,19 +14,19 @@ option go_package = "github.com/hashicorp/boundary/sdk/pbs/controller/api/resour
 
 message Member {
   // Output only. The ID of the member.
-  string id = 10; // @gotags: `class:"public"`
+  string id = 10; // @gotags: `class:"public" eventstream:"observation"`
 
   // Output only. The Scope ID of the member.
-  string scope_id = 20 [json_name = "scope_id"]; // @gotags: `class:"public"`
+  string scope_id = 20 [json_name = "scope_id"]; // @gotags: `class:"public" eventstream:"observation"`
 }
 
 // Group contains all fields related to a Group resource
 message Group {
   // Output only. The ID of the Group.
-  string id = 10; // @gotags: `class:"public"`
+  string id = 10; // @gotags: `class:"public" eventstream:"observation"`
 
   // The ID of the scope of which this Group is a part.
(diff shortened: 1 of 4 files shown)


internal/proto/controller/api/resources/groups/v1/group.proto
```diff
@@ -14,19 +14,19 @@ option go_package = "github.com/hashicorp/boundary/sdk/pbs/controller/api/resour
...
   // Output only. The ID of the Group.
-  string id = 10; // @gotags: `class:"public"`
+  string id = 10; // @gotags: `class:"public" eventstream:"observation"`
 
   // The ID of the scope of which this Group is a part.
-  string scope_id = 20 [json_name = "scope_id"]; // @gotags: `class:"public"`
+  string scope_id = 20 [json_name = "scope_id"]; // @gotags: `class:"public" eventstream:"observation"`
 
   // Output only. Scope information for this Group.
   resources.scopes.v1.ScopeInfo scope = 30;
```
johanbrandhorst, line 32: What does it mean to add this tag to only some fields in a message?

sepehrfrgh: These fields are specifically chosen for the telemetry pipeline and for the product team where they are not customer-specified values

johanbrandhorst: Maybe I'm missing something - what happens to the fields that are tagged? What values can the `eventstream` tag take?

sepehrfrgh: we read and detect them inside observations_event and we will add them to the observation pipeline. for now just an observation value but we might have another type later down the line.

johanbrandhorst: So when reviewing this you just want me to check that the right fields are tagged?

sepehrfrgh: Exactly, Thank you 🙏

johanbrandhorst (approved):

internal/proto/controller/api/resources/groups/v1/group.proto
sepehrfrgh: and for sure If I messed up in any other parts 😄

jimlambrt (approved):

ddebko (approved):

PR #3535, opened 2023-08-02
Client daemon now uses unix sockets and update pathing

talanknight:
The pathing now matches the RFC.

The unix sockets are now used for the daemon listener and client.  Some changes to the client had to be made to support windows style file pathing.

Left as a TODO is figuring out socket file permissions on windows.

### DIFF
--- a/internal/cmd/commands/daemon/listener.go
+++ b/internal/cmd/commands/daemon/listener.go
@@ -35,8 +35,14 @@ func listener(ctx context.Context) (net.Listener, error) {
 	if err != nil {
 		return nil, errors.Wrap(ctx, err, op)
 	}
-
 	socketName := filepath.Join(homeDir, sockAddr)
+	if err := os.Remove(socketName); err != nil {
+		// If the socket existed before and wasn't cleaned up delete it now.
+		if !os.IsNotExist(err) {
+			return nil, errors.Wrap(ctx, err, op)
+		}
+	}
+
 	socketPath := filepath.Dir(socketName)
 	if err := os.MkdirAll(socketPath, socketDirPerms); err != nil {
 		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("failed to create boundary directory"))


johanbrandhorst (approved):

internal/cmd/commands/daemon/listener.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		return nil, errors.Wrap(ctx, err, op)
+	}
+	if err := os.Chmod(socketName, socketPerms); err != nil {
+		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("changing socket permissiosn"))
+	}
+	return l, nil
+}
+
+func daemonAddress() string {
```
jimlambrt, line 52: how does this work given the implementation of `listen(...)` above?  It would
appear `daemonAddress()` and the listener would not refer to the same unix socket.

jimlambrt: Also, I'm not sure the perms are correct here... but I'm not sure that's
important if `daemonAddress()` isn't correct overall.

internal/cmd/commands/daemon/search.go
```diff
@@ -212,13 +203,15 @@ func (c *SearchTargetsCommand) printListTable(items []*targets.Target) string {
...
 
-func SearchClient(ctx context.Context, addr string, flagOutputCurl bool) (*api.Client, error) {
+func SearchClient(ctx context.Context, flagOutputCurl bool) (*api.Client, error) {
 	const op = "daemon.SearchClient"
 	client, err := api.NewClient(nil)
 	if err != nil {
 		return nil, errors.Wrap(ctx, err, op)
 	}
-	client.SetAddr(addr)
+	if err := client.SetAddr(daemonAddress()); err != nil {
```
jimlambrt, line 212: See my questions about `daemonAddress()`: does this produce the right addr?

internal/cmd/commands/daemon/server.go
```diff
@@ -213,8 +211,8 @@ func (s *server) start(ctx context.Context, port uint) error {
 	// background bits
 	var tic *refreshTicker
 	{
-		s.info["Listening port"] = strconv.FormatUint(uint64(port), 10)
-		s.infoKeys = append(s.infoKeys, "Listening port")
+		s.info["Listening address"] = daemonAddress()
```
jimlambrt, line 214: See my questions about daemonAddress(): does this produce the right addr?

jimlambrt (changes requested):

mikemountain (approved):

jimlambrt (approved):
ty!

PR #3537, opened 2023-08-03
tests: add timeout to key version destruction test

johanbrandhorst:


tmessi (approved):
Is to address the test that sometimes hangs that was identified in https://github.com/hashicorp/boundary/pull/3536?

This will at least cause it to fail faster, but do we have any idea why it is failing sometimes?

johanbrandhorst:
> Is to address the test that sometimes hangs that was identified in #3536?
> 
> This will at least cause it to fail faster, but do we have any idea why it is failing sometimes?

Yeah this will address the hanging, but we still don't know why it would be failing sometimes. Hopefully next time this fails it'll give us some clue.

johanbrandhorst:
I added a log for the jobs found in the latest poll, I don't know what else we'd log here. The logs from the rest of the system should tell us more already, I think.

tmessi (approved):

PR #3539, opened 2023-08-04
SQL Tables, functions, initial job

mikemountain:


mgaffney (changes requested):
Why not follow the pattern use for history tables?

internal/cleanup/cleanup.go
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package cleanup
```
johanbrandhorst, line 4: As discussed, lets rebase this on https://github.com/hashicorp/boundary/pull/3544 and move this job into that package (or a subpackage, more likely)

internal/cleanup/cleanup_job.go
```diff
@@ -0,0 +1,74 @@
...
+	return scheduler.JobStatus{}
+}
+
+// Run performs the required work depending on the implementation.
+// The context is used to notify the job that it should exit early.
+func (c *cleanupJob) Run(ctx context.Context) error {
+	const op = "cleanup.(cleanupJob).Run"
+	if _, err := c.w.DoTx(ctx, db.StdRetryCnt, db.ExpBackoff{},
+		func(_ db.Reader, w db.Writer) error {
+			_, err := w.Exec(ctx, `select * from cleanup_deleted_tables()`, nil)
```
johanbrandhorst, line 46: According to https://www.postgresql.org/docs/current/sql-syntax-calling-funcs.html#SQL-SYNTAX-CALLING-FUNCS-POSITIONAL, we should be able to make this just
```suggestion
			_, err := w.Exec(ctx, `select cleanup_deleted_tables()`, nil)
```

internal/cleanup/cleanup_job.go
```diff
@@ -0,0 +1,74 @@
...
+	"context"
+	"time"
+
+	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/scheduler"
+)
+
+type cleanupJob struct {
+	r db.Reader
```
johanbrandhorst, line 16: It doesn't look like this is used in the job, can we just remove it?

internal/cleanup/cleanup_job.go
```diff
@@ -0,0 +1,74 @@
...
+// Status reports the job’s current status.
+func (c *cleanupJob) Status() scheduler.JobStatus {
+	return scheduler.JobStatus{}
+}
+
+// Run performs the required work depending on the implementation.
+// The context is used to notify the job that it should exit early.
+func (c *cleanupJob) Run(ctx context.Context) error {
+	const op = "cleanup.(cleanupJob).Run"
+	if _, err := c.w.DoTx(ctx, db.StdRetryCnt, db.ExpBackoff{},
```
johanbrandhorst, line 44: Wrapping the exec in a transaction gives us automatic retries, but since this job just reruns every now and then on its own, maybe we can just skip the transaction?

internal/cleanup/cleanup_job.go
```diff
@@ -0,0 +1,74 @@
...
+		return err
+	}
+	return nil
+}
+
+// NextRunIn returns the duration until the next job run should be scheduled.
+// We report as ready immediately after a successful run. This doesn't mean that
+// this job will run immediately, only about as often as the configured scheduler interval.
+func (c *cleanupJob) NextRunIn(_ context.Context) (time.Duration, error) {
+	return 0, nil
```
johanbrandhorst, line 63: This seems more often than necessary. I'd say we could run this job once a day and it'd be okay. What do you think?
```suggestion
	return 24*time.Hour, nil
```

internal/cleanup/cleanup_job.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// Name is the unique name of the job.
+func (c *cleanupJob) Name() string {
+	return "job_run_cleanup"
```
johanbrandhorst, line 68: How about
```suggestion
	return "deleted_items_table_cleanup"
```

internal/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/iam"
+	"github.com/hashicorp/boundary/internal/scheduler"
+	"github.com/hashicorp/boundary/internal/session"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPrototypeInsertDeleteAndClear(t *testing.T) {
```
johanbrandhorst, line 20: Is this still a prototype test? I think we can probably rename this
```suggestion
func TestCleanupJob(t *testing.T) {
```

internal/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,183 @@
...
+		t.Errorf("error getting db connection %s", err)
+	}
+
+	_, err = db.Exec(fmt.Sprintf(`delete from target where public_id = '%s'`, session.TargetId))
+	if err != nil {
+		t.Errorf("error deleting from target %s", err)
+	}
+
+	// make sure that the trigger works
+	rows, err := db.Query("select count(public_id) from target_deleted")
```
johanbrandhorst, line 40: Can we use `QueryRowContext`?
```suggestion
	var count int
	err := db.QueryRowContext(ctx, "select count(public_id) from target_deleted").Scan(&count)
```

internal/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,183 @@
...
+	if err != nil {
+		t.Errorf("error checking target_deleted table %s", err)
+	}
+	defer rows.Close()
+	var count int
+	for rows.Next() {
+		err = rw.ScanRows(ctx, rows, &count)
+		if err != nil {
+			t.Error("unable to scan rows for census sessions pending count")
```
johanbrandhorst, line 49: 😬

internal/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,183 @@
...
+	count = 0
+	for rows.Next() {
+		err = rw.ScanRows(ctx, rows, &count)
+		if err != nil {
+			t.Error("unable to scan rows for census sessions pending count")
+		}
+	}
+	require.Equal(t, 1, count)
+
+	_, err = db.Exec(`update target_deleted set delete_time = '2023-05-18';`)
```
johanbrandhorst, line 77: This will work, but we could just set it to `time.Now().AddDate(0, -2, 0)` to make it a little less weird.
```suggestion
	_, err = db.Exec("update target_deleted set delete_time = $1", time.Now().AddDate(0, -2, 0))
```

internal/db/schema/migrations/oss/postgres/76/01_deleted_tables_and_triggers.up.sql
```diff
@@ -0,0 +1,247 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: MPL-2.0
+
+begin;
+  create table target_deleted (
+    public_id wt_public_id primary key,
+    delete_time wt_timestamp
```
johanbrandhorst, line 7: I think this should be `not null`?
```suggestion
    delete_time wt_timestamp not null
```

internal/db/schema/migrations/oss/postgres/76/01_deleted_tables_and_triggers.up.sql
```diff
@@ -0,0 +1,247 @@
...
+  );
+
+  create index target_deleted_time_idx on target_deleted (delete_time);
```
johanbrandhorst, line 10: I think this should be
```suggestion
  create index target_deleted_delete_time_idx on target_deleted (delete_time);
```
to be consistent with the recommendation for unique constraints in [the SQL style guide](https://github.com/hashicorp/boundary/blob/main/sql-style-guide.md).

internal/db/schema/migrations/oss/postgres/76/01_deleted_tables_and_triggers.up.sql
```diff
@@ -0,0 +1,247 @@
...
+    ('iam_user_deleted'),
+    ('recording_session_deleted'),
+    ('auth_managed_group_deleted'),
+    ('server_worker_deleted'),
+    ('auth_account_deleted');
+
+  create or replace function insert_deleted_id() returns trigger
+  as $$
+  begin
+    execute format('INSERT INTO %I (public_id, delete_time) values ($1, now())', tg_argv[0]) using old.public_id;
```
johanbrandhorst, line 193:
```suggestion
    execute format('insert into %I (public_id, delete_time) values ($1, now())', tg_argv[0]) using old.public_id;
```

internal/db/sqltest/tests/cleanup/cleanup_old_records.sql
```diff
@@ -0,0 +1,19 @@
...
+  select has_function('cleanup_deleted_tables');
+  select volatility_is('cleanup_deleted_tables', 'volatile');
+  select isnt_strict('cleanup_deleted_tables');
+
+  insert into session_deleted (public_id, delete_time) select 'p1234567890', now()::date;
+  insert into session_deleted (public_id, delete_time) select 'p9876543210', now()::date - '30 days'::interval;
+
+  select * from cleanup_deleted_tables();
+
+  select is(count(*), 1::bigint) from session_deleted;
```
johanbrandhorst, line 16: Nit: can you add a count before the cleanup as well to confirm that it's 2?

mikemountain:
closing in favour of https://github.com/hashicorp/boundary/pull/3568

PR #3541, opened 2023-08-04
Telemetry: Add observation tag to host proto

sepehrfrgh:
This PR adds observation gotags into `host.proto` and `host_service.proto` for telemetry purposes
cc: @jimlambrt @ajayreshc


internal/proto/controller/api/resources/hosts/v1/host.proto
```diff
@@ -51,20 +51,20 @@ message Host {
   ]; // @gotags: `class:"public"`
```
johanbrandhorst, line 51: Are we adding PluginInfo separately, or is that not included?

sepehrfrgh: Yes, we are adding them inside `plugin.proto`

jimlambrt (approved):

PR #3543, opened 2023-08-04
Telemetry: session proto

sepehrfrgh:
This PR adds observation gotags into `session.proto` and `session_service.proto` for telemetry purposes
** please review all the fields as this proto has more fields

cc: @jimlambrt @ajayreshc

### DIFF
--- a/internal/proto/controller/api/services/v1/session_service.proto
+++ b/internal/proto/controller/api/services/v1/session_service.proto
@@ -47,7 +47,7 @@ service SessionService {
 }
 
 message GetSessionRequest {
-  string id = 1; // @gotags: `class:"public"`
+  string id = 1; // @gotags: `class:"public" eventstream:"observation"`
 }
 
 message GetSessionResponse {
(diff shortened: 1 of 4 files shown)


internal/proto/controller/api/services/v1/session_service.proto
```diff
@@ -47,28 +47,28 @@ service SessionService {
...
 
 message ListSessionsRequest {
-  string scope_id = 1; // @gotags: `class:"public"`
+  string scope_id = 1; // @gotags: `class:"public" eventstream:"observation"`
   bool recursive = 20 [json_name = "recursive"]; // @gotags: `class:"public"`
   string filter = 30 [json_name = "filter"]; // @gotags: `class:"public"`
   // Experimental. By default only non-terminated (i.e. pending, active, canceling) are returned.
   // Set this option to include terminated sessions as well.
-  bool include_terminated = 40 [json_name = "include_terminated"]; // @gotags: `class:"public"`
+  bool include_terminated = 40 [json_name = "include_terminated"]; // @gotags: `class:"public" eventstream:"observation"`
```
johanbrandhorst, line 63: If we're including this, we should include `recursive` too, right?

jimlambrt (approved):

PR #3545, opened 2023-08-07
Cache refreshing now uses all stored personas.

talanknight:
Personas have been expanded to include the keyring type as part of the primary key.
When the daemon starts up, it creates a persona if the current keyring type and token name have an auth token.
When the cache refreshes, it removes any personas that have been stale longer than the default time.
All persona information is moved from headers to be query parameters.

Future work will add a daemon server path that adds a persona to the cache.


jimlambrt (approved):
Looks great.  We should spend some time chatting 'bout tokens.

PR #3566, opened 2023-08-10
Decouple daemon and cache server

talanknight:
Right now I left erroring stubs for the windows version of the daemon commands.

### DIFF
--- a/internal/cmd/commands.go
+++ b/internal/cmd/commands.go
@@ -63,212 +63,180 @@ func initCommands(ui, serverCmdUi cli.Ui, runOpts *RunOptions) {
 			}, nil
 		},
 
-		"authenticate": func() (cli.Command, error) {
-			return &authenticate.Command{
+		"authenticate": daemon.Wrap(ui,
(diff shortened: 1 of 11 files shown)


internal/cmd/commands/daemon/start_windows.go
```diff
@@ -0,0 +1,42 @@
...
+func DefaultDotDirectory(ctx context.Context) (string, error) {
+	const op = "daemon.DefaultDotDirectory"
+	homeDir, err := homedir.Dir()
+	if err != nil {
+		return "", errors.Wrap(ctx, err, op)
+	}
+	return fmt.Sprintf(dotDirectoryNameTemplate, homeDir), nil
+}
+
+// start will ensure this is the only daemon running and spin off a seperate process.
```
jimlambrt, line 34:
```suggestion
// start will ensure this is the only daemon running and spin off a separate process.
```

jimlambrt (approved):
Just a couple minor things to clean up before merging, but looks great!

PR #3569, opened 2023-08-11
Re-enable copywrite with hack

johanbrandhorst:
## [Re-enable copywrite with hack](https://github.com/hashicorp/boundary/commit/d54a2f7998ff57a4d70e289f6b28c24d00160452)

Adds a hack that removes existing copyright headers
from files before rerunning copywrite with directory
specific configuration.

Also remove the massive copyright change commit from blame


jimlambrt (approved):
🙈

PR #3578, opened 2023-08-14
test(e2e): Refactor migration test docker repo definition

moduli:
This PR updates the e2e migration test based on a comment here: https://github.com/hashicorp/boundary-enterprise/pull/670#discussion_r1291668121

It refactors where `docker pull` is executed in order to reduce the number of places where a docker repo needs to be specified (In the comment above, I had modified the image that `docker run` would use, but did not modify the corresponding `docker pull`). Additionally, this makes it simpler to modify which version of Boundary to use in the event we want to run the migration test on a different starting version.


johanbrandhorst (approved):
Thanks!

PR #3579, opened 2023-08-14
Add persona handler to daemon server

talanknight:
Persona information now get sent to the daemon for every CLI command.  If the auth token has changed or a new persona was added then cache refresh is automatically triggered.

This PR also explicitly wraps each command the daemon startup/persona sending happens and starts the daemon through a call to `exec.Command`.

### DIFF
--- a/internal/cmd/commands/daemon/ticker.go
+++ b/internal/cmd/commands/daemon/ticker.go
@@ -62,8 +62,10 @@ func newRefreshTicker(ctx context.Context, refreshIntervalSeconds int64, store *
 	}, nil
 }
 
-// refresh starts the refresh logic if it is currently waiting. If refresh is currently happening
-// then this is a noop.
+// refresh starts the refresh logic if it is currently waiting. If refresh is
+// currently happening then this queues a refresh for when the current refresh
+// iteration is done, unless there is already a refresh queued in which case
+// this is a noop.
 func (rt *refreshTicker) refresh() {
 	select {
 	case rt.refreshChan <- struct{}{}:


internal/cmd/commands/daemon/command_wrapper.go
```diff
@@ -0,0 +1,87 @@
...
+	args[0] = binary
+
+	var stdErr bytes.Buffer
+	cmd := exec.Command(args[0], "daemon", "start")
+	cmd.Stderr = &stdErr
+	cmd.Env = os.Environ()
+	// We use Run here instead of Start because the command spawns off a subprocess and returns.
+	// We do not want to send the request to add a persona to the cache until we know the daemon
+	// has starte dup.
+	err = cmd.Run()
```
jimlambrt, line 75:
```suggestion
	// has started up.
	err = cmd.Run()
```

internal/cmd/commands/daemon/persona_handler.go
```diff
@@ -0,0 +1,120 @@
...
+	"net/http"
+
+	"github.com/hashicorp/boundary/api/authtokens"
+	"github.com/hashicorp/boundary/internal/cmd/base"
+	"github.com/hashicorp/boundary/internal/daemon/cache"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/util"
+)
+
+type readTokenFromKeyringer interface {
```
jimlambrt, line 20:
```suggestion
type tokenReader interface {
```

jimlambrt: Just a suggestion... I'm not a fan of the current name.

internal/cmd/commands/daemon/persona_handler.go
```diff
@@ -0,0 +1,120 @@
...
+		TokenName:    p.TokenName,
+		BoundaryAddr: p.BoundaryAddr,
+		AuthTokenId:  p.AuthTokenId,
+	}
+}
+
+func newPersonaHandlerFunc(ctx context.Context, store *cache.Store, atReader readTokenFromKeyringer, refresher refresher) (http.HandlerFunc, error) {
+	const op = "daemon.newPersonaHandlerFunc"
+	switch {
+	case util.IsNil(store):
```
jimlambrt, line 50: perhaps check the other parameters?

internal/cmd/commands/daemon/persona_handler.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+
+		repo, err := cache.NewRepository(ctx, store)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		foundP, err := repo.LookupPersona(ctx, perReq.BoundaryAddr, perReq.KeyringType, perReq.TokenName)
+		if err != nil {
+			http.Error(w, "error performign persona lookup", http.StatusInternalServerError)
```
jimlambrt, line 104:
```suggestion
			http.Error(w, "error performing persona lookup", http.StatusInternalServerError)
```

jimlambrt (changes requested):
Looks great.  Just a few minor bits and you'll need some tests.

jimlambrt (approved):

PR #3581, opened 2023-08-14
Telemetry: servers proto

sepehrfrgh:
This PR adds observation gotags into `servers.proto` and `session_service.proto` in the servers package for telemetry purposes.
cc: @jimlambrt @ajayreshc

### DIFF
--- a/internal/proto/controller/servers/services/v1/session_service.proto
+++ b/internal/proto/controller/servers/services/v1/session_service.proto
@@ -36,10 +36,10 @@ service SessionService {
 
 message LookupSessionRequest {
   // The session ID from the client
-  string session_id = 10; // @gotags: `class:"public"`
+  string session_id = 10; // @gotags: `class:"public" eventstream:"observation"`
   // The id of the requesting worker, used for filtering to ensure this worker
   // can handle this session
-  string worker_id = 20; // @gotags: `class:"public"`
+  string worker_id = 20; // @gotags: `class:"public" eventstream:"observation"`
 }
 
 // LookupSessionResponse contains information necessary for a client to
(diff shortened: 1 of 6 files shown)


internal/proto/controller/servers/services/v1/session_service.proto
```diff
@@ -49,14 +49,14 @@ message LookupSessionResponse {
   string tofu_token = 20; // @gotags: `class:"secret"`
   uint32 version = 30; // @gotags: `class:"public"`
   string endpoint = 40; // @gotags: `class:"public"`
-  google.protobuf.Timestamp expiration = 50; // @gotags: `class:"public"`
-  controller.servers.services.v1.SESSIONSTATUS status = 60; // @gotags: `class:"public"`
+  google.protobuf.Timestamp expiration = 50; // @gotags: `class:"public" eventstream:"observation"`
+  controller.servers.services.v1.SESSIONSTATUS status = 60; // @gotags: `class:"public" eventstream:"observation"`
```
johanbrandhorst, line 53: This status is not recorded in `SessionJobInfo` and `MonitorSessionJobInfo`, but it is here and elsewhere. Should it be recorded everywhere?

internal/proto/controller/servers/services/v1/session_service.proto
```diff
@@ -69,32 +69,32 @@ message ActivateSessionRequest {
...
-  string worker_id = 20; // @gotags: `class:"public"`
+  string session_id = 10; // @gotags: `class:"public" eventstream:"observation"`
+  string worker_id = 20; // @gotags: `class:"public" eventstream:"observation"`
 }
 
 message AuthorizeConnectionResponse {
-  string connection_id = 10; // @gotags: `class:"public"`
-  controller.servers.services.v1.CONNECTIONSTATUS status = 20; // @gotags: `class:"public"`
+  string connection_id = 10; // @gotags: `class:"public" eventstream:"observation"`
+  controller.servers.services.v1.CONNECTIONSTATUS status = 20; // @gotags: `class:"public" eventstream:"observation"`
```
johanbrandhorst, line 97: Ditto status here and in `Connection` above

internal/proto/controller/servers/services/v1/session_service.proto
sepehrfrgh: I think these two protos is mostly for Job proto message and they are not part of the telemetry and I focus more on `ServerWorkerStatus` which will be used as a telemetry. I added the ids for those messages just in case but it wont hurt to have it since it is not customer input

johanbrandhorst (approved):

jimlambrt (approved):

PR #3600, opened 2023-08-17
config: add max page size

johanbrandhorst:
MaxPageSize overrides the default and max page size. The default page size is what is used when the page size is not explicitly provided by the user. The max page size is the greatest number the page size can be set to before it is rejected by the controller.


mikemountain (approved):
LGTM!

PR #3613, opened 2023-08-21
docs: Add `storage-buckets` commands

Dan-Heath:
Adds the `storage-buckets` commands on top of the existing assembly branch for CLI commands #3411.


Dan-Heath:
Closed in favor of #3615

PR #3618, opened 2023-08-22
Windows support for client side daemon

talanknight:
This PR attempts to improve the daemon behavior of the boundary client daemon by adding a "-background" flag which can be passed in to `boundary deamon start` to run it in the background.  It also spins off a different process using exec.Command instead of relying completely on the library we were using before.  This allows that code to be shared across all platforms and reserves the platform specific code for just pid file management and the `boundary daemon stop` behavior.   All commands wrapped in the daemon.CommandWraper attempt to create a new daemon by passing in the `-background` flag.

Future work will involve having `boundary daemon start -background` return an error when it detects the spun off process failed to start.

The log now always writes to the log file at ~/.boundary/cache.log and writes to stderr when the daemon is run in the foreground.

Regarding the daemon on windows:
- Since windows doesn't have POSIX signals `boundary daemon stop` force kills the daemon process instead of using the built in signal handling that happens in non-windows binaries.  Since almost everything is stored in memory this shouldn't be a big deal, but as we start writing to places for persistence (maybe personas being reloaded at start up) this might cause problems.
- The pid files do not use locks and relies on file permissions to ensure that no other processes attempt to open it for writing.  Future work can add locks to avoid race conditions here if needed.

### DIFF
--- a/internal/cmd/commands/daemon/stop_nonwindows.go
+++ b/internal/cmd/commands/daemon/stop_nonwindows.go
@@ -32,6 +32,9 @@ func (s *StopCommand) stop(ctx context.Context) error {
 	if err != nil {
 		return errors.Wrap(ctx, err, op)
 	}
+	if p == nil {
+		return errors.New(ctx, errors.NotFound, op, "daemon not running")
+	}
 	if err := p.Signal(syscall.SIGTERM); err != nil {
 		return errors.Wrap(ctx, err, op, errors.WithMsg("sending sigterm to process"))
 	}


internal/cmd/commands/daemon/server.go
```diff
@@ -191,8 +184,16 @@ func (s *cacheServer) serve(ctx context.Context, cmd commander, l net.Listener)
 	}
 	mux.HandleFunc("/v1/personas", personaFn)
 
+	logger, err := event.SysEventer().StandardLogger(ctx, "daemon.serve: ", event.ErrorType)
```
jimlambrt, line 187: what if the eventer is nil?

jimlambrt (changes requested):
Looks good.  Just a few questions about the sys eventer.

internal/cmd/commands/daemon/server.go
jimlambrt: Isn't setupLogging(...) called in Run(...)?  I'd be happy if it was called in newServer(...).

jimlambrt: ty!

jimlambrt (approved):

PR #3622, opened 2023-08-23
Add telemetry observation tags to plugin.proto

ajayreshc:
Add eventstream:"observation" tags to plugin.proto field.

Closed a previous PR (3591) and created it fresh, as there was an issue with a rebase in the my branch.


sepehrfrgh (approved):
Thank you

PR #3627, opened 2023-08-23
Add generic pagination function

johanbrandhorst:


mikemountain (approved):

mikemountain (approved):

internal/pagination/pagination.go
```diff
@@ -10,14 +10,231 @@ import (
...
+	if req.GetRefreshToken() != "" {
+		// Note that refresh token parsing and validation happens after authorization,
+		// since validation requires access to the grants hash.
+		refreshToken, err = parseRefreshToken(ctx, req.GetRefreshToken())
+		if err != nil {
+			return nil, errors.Wrap(ctx, err, op)
+		}
+		if err := validateRefreshToken(ctx, refreshToken, grantsHash, resourceType); err != nil {
+			return nil, errors.Wrap(ctx, err, op)
+		}
```
johanbrandhorst, line 114: Note that parsing and validation of the refresh token now happens _after_ authorization. If we wanted to parse first, we'd have to do authorization in this function, which I wanted to avoid.

mikemountain (approved):

internal/pagination/pagination.go
mikemountain: Is there any reason we wouldn't want to authorise before validating?  Correct me if I'm wrong, but to me it makes more sense to check if the user can even make the request, before checking if the request is correct

johanbrandhorst: We generally do authentication -> request validation -> authorization AFAICT. We'll now do authentication -> some request validation -> authorization -> some more request validation.

johanbrandhorst:
This should be good to go :)

mikemountain (approved):

Issue #3631, opened 2023-08-24
Non-Interactive SSH sessions do not behave as expected


steven-whaley:
**Describe the bug**
When attempting to use "boundary connect ssh" to run commands non-interactively on the target host the boundary client does not behave as expected and instead parses the commands as a hostname and throws an error.  

`swhaley@swhaley-M43L9KG44N ~ % boundary connect ssh -target-id tssh_fLlnqxhIsD -- whoami
ssh: Could not resolve hostname whoami: nodename nor servname provided, or not known`

**To Reproduce**
1. Attempt to run non-interactive commands on the target via "boundary connect ssh" as shown above
2. See error

**Expected behavior**
I would expect that this would behave identically to SSH without boundary where the output of the commands is returned to the terminal and then the SSH session is closed.  

**Additional context**
You can work around this with -exec but ideally the behavior would conform to the ssh client behavior when boundary is not involved.  

Workaround: 
`boundary connect -exec ssh -target-id tssh_fLlnqxhIsD -- -o "NoHostAuthenticationForLocalhost=yes" -p {{boundary.port}} {{boundary.ip}} -l ec2-user "whoami; pwd" `

PR #3633, opened 2023-08-24
WIP: bug(cmd): Order passthrough args after generated

louisruch:


johanbrandhorst:
This is for https://github.com/hashicorp/boundary/issues/3631, right?

louisruch:
@tmessi @jefferai PTAL
You might notice I also changed when we add the destination because I noticed we where adding a number of arguments after the destination instead of what the man page for ssh actually called for

jefferai (approved):
Looks great, thank you!

PR #3638, opened 2023-08-25
tests (grants): add basic fuzz test for grants.Parse

jimlambrt:


jefferai (approved):

internal/perms/grants_test.go
```diff
@@ -1072,6 +1072,31 @@ func Test_Parse(t *testing.T) {
...
+		`id={{    user.id}};actions=read,update`,
+	}
+	for _, tc := range tc {
+		f.Add(tc)
+	}
+	f.Fuzz(func(t *testing.T, s string) {
+		_, _ = Parse(context.Background(), "global", s)
+	})
+}
+
```
johanbrandhorst, line 1099: There's already a parse fuzz test at the bottom of this file, why don't we add these cases to that?

jimlambrt: I'll add these tests to the other one and run that one in CI

johanbrandhorst (approved):

PR #3639, opened 2023-08-28
handlers/targets/tcp: run tests in parallel

johanbrandhorst:
This reduces the runtime from 79s to 24s on my machine


johanbrandhorst:
> If I'm interpreting these tests correctly, it looks like the data they use is independent of each other (each test/subtest creates the data it needs and won't step on other tests)?

All the tests use `conn, _ := db.TestSetup(t, "postgres")` to create a new DB in the postgres, so they should all be independent, which is why this should be a safe change.

PR #3640, opened 2023-08-28
Requests with version info must match the daemon's version

talanknight:


internal/cmd/commands/daemon/version_interceptor.go
```diff
@@ -0,0 +1,32 @@
...
+func versionEnforcement(h http.Handler) http.Handler {
+	if util.IsNil(h) {
+		return nil
+	}
+
+	needVer := version.Get().VersionNumber()
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		gotVer := r.Header.Get(VersionHeaderKey)
+		if gotVer != "" && needVer != gotVer {
+			writeError(w, fmt.Sprintf("version mismatch between requester (%q) and server (%q)", gotVer, needVer), http.StatusBadRequest)
```
johanbrandhorst, line 27: This will presumably be seen on a users terminal, could we add a call to action to the error? Something like "You may need to restart your daemon" (presumably this happens immediately after a user upgrades their CLI).

jimlambrt (approved):

PR #3641, opened 2023-08-29
fix (migrations): drop constraint iam_scope_fkey on oplog_entry

jimlambrt:
The existing FK from iam_scope(public_id) means that oplog entries are deleted when a scope is deleted.  We should not be deleting these oplog entries.  Dropping the FK will stop this cascading delete.  Of course, we still need the scope_id on in the oplog_entry table and the existing on insert trigger will make sure the scope id is valid.


tmessi (approved):

jimlambrt:
This won't really work as is.  We need a more comprehensive solution that deals with oplog scopes more completely.  I'll write something up.

PR #3643, opened 2023-08-29
chore: Update changelog

louisruch:


louisruch:
@johanbrandhorst thanks for catching that I missed the changelog entry

johanbrandhorst (approved):
Thanks! I don't think it needs to be part of this PR, but it would be nice to add a small section about this new flag to the connect helpers part of the docs. It's in two places:

Dev mode: https://developer.hashicorp.com/boundary/docs/getting-started/dev-mode/connect-to-dev-target#use-connect-helpers
HCP getting started: https://developer.hashicorp.com/boundary/docs/hcp/get-started/connect-to-target#use-connect-helpers

louisruch:
> Thanks! I don't think it needs to be part of this PR, but it would be nice to add a small section about this new flag to the connect helpers part of the docs. It's in two places:
> 
> Dev mode: https://developer.hashicorp.com/boundary/docs/getting-started/dev-mode/connect-to-dev-target#use-connect-helpers HCP getting started: https://developer.hashicorp.com/boundary/docs/hcp/get-started/connect-to-target#use-connect-helpers

Ill make a new PR

PR #3644, opened 2023-08-29
Fix address not returning on 10001+ list value

jefferai:
Fixes ICU-10786


johanbrandhorst:
Hm, that fuzz failure is curious, we saw the same thing when I first added the fuzz target and it should be fixed already. https://github.com/hashicorp/boundary/pull/2534#issuecomment-1279235997

johanbrandhorst (approved):

jimlambrt (approved):
Ty

PR #3645, opened 2023-08-29
genapi: add pagination support

johanbrandhorst:
Adds the `WithRefreshToken` and `WithPageSize` options to specify list pagination parameters. If a refresh token is not set, list functions will automatically page through the entire result set.

This new functionality is exercised by https://github.com/hashicorp/boundary/blob/main/internal/daemon/controller/gateway_test.go.

### DIFF
--- a/internal/api/genapi/templates.go
+++ b/internal/api/genapi/templates.go
@@ -266,7 +266,49 @@ func (c *Client) List(ctx context.Context, {{ .CollectionFunctionArg }} string, 
 		return nil, apiErr
 	}
 	target.response = resp
-	return target, nil
+	if opts.withRefreshToken != "" || target.ResponseType == "complete" || target.ResponseType == "" {
+		return target, nil
+	}
+	// if refresh token is not set explicitly and there are more results,
+	// automatically fetch the rest of the results.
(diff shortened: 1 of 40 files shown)


internal/api/genapi/templates.go
```diff
@@ -266,7 +266,49 @@ func (c *Client) List(ctx context.Context, {{ .CollectionFunctionArg }} string,
...
 	}
 	target.response = resp
-	return target, nil
+	if opts.withRefreshToken != "" || target.ResponseType == "complete" {
+		return target, nil
+	}
+	// if refresh token is not set explicitly and there are more results,
+	// automatically fetch the rest of the results.
+	for {
+		req, err := c.client.NewRequest(ctx, "GET", "targets", nil, apiOpts...)
```
tmessi, line 275: Should this use a template variable?

```suggestion
		req, err := c.client.NewRequest(ctx, "GET", "{{ .CollectionPath }}", nil, apiOpts...)
```

johanbrandhorst: Good spot 😁

mikemountain (approved):

PR #3647, opened 2023-08-30
Fix fuzz corpus upload

johanbrandhorst:


johanbrandhorst:
I think the actionlint warnings are bogus. This should be good to go.

jimlambrt (approved):
Looks good.  There's a lint error

PR #3653, opened 2023-08-31
Finish Uploading BSR During Forced Worker Shutdown

kheina:
give the storage struct an empty context so that it can close cleanly during a forced shutdown. fixes an issue where if a worker is forced to shut down via double ctrl-c, the session recording is missing all recording data. (bsr file only contained keys on aws)

tested via command line worker and killed with double ctrl-c


louisruch (approved):
LGTM

PR #3659, opened 2023-09-01
ci: attempt to silence action linter

johanbrandhorst:
The lint warnings are bogus but they keep failing so lets try to work around the suggested issues.


moduli (approved):
Cool. Looks like it's passing now. Thanks!

PR #3660, opened 2023-09-01
internal/perms: Fix grant parser bugs

johanbrandhorst:
This fixes two bugs in the grant parser:

1. The parser would accept "," as an output field when parsing from JSON. This would then make it into the canonical string, which would make it look like we had a trailing comma. Disallow the use of a comma as an output field.
2. The parser would accept empty IDs when parsing from JSON. Make this behavior consistent with the text behavior.

### DIFF
--- a/internal/perms/grants.go
+++ b/internal/perms/grants.go
@@ -236,7 +236,7 @@ func (g *Grant) unmarshalJSON(ctx context.Context, data []byte) error {
 				return errors.New(ctx, errors.InvalidParameter, op, fmt.Sprintf("unable to interpret %q element %q as string", "ids", id))
 			}
 			if idStr == "" {
-				return errors.New(ctx, errors.InvalidParameter, op, `segment "ids=" not formatted correctly, missing value`)
+				return errors.New(ctx, errors.InvalidParameter, op, "empty ID provided")
 			}
 			g.ids[i] = idStr
 		}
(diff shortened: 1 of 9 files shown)


internal/perms/grants.go
```diff
@@ -235,6 +235,9 @@ func (g *Grant) unmarshalJSON(ctx context.Context, data []byte) error {
 			if !ok {
 				return errors.New(ctx, errors.InvalidParameter, op, fmt.Sprintf("unable to interpret %q element %q as string", "ids", id))
 			}
+			if idStr == "" {
+				return errors.New(ctx, errors.InvalidParameter, op, `segment "ids=" not formatted correctly, missing value`)
```
jefferai, line 239: Can this be more specific, as in the error on line 236? "ids" isn't missing a value, some id within it is...

jefferai (approved):

jimlambrt (approved):
Ty!

jimlambrt (approved):
Ty!

internal/perms/grants.go
```diff
@@ -315,9 +329,17 @@ func (g *Grant) unmarshalText(ctx context.Context, grantString string) error {
 		switch kv[0] {
 		case "id":
 			g.id = kv[1]
+			if strings.Contains(g.id, ",") {
+				return errors.New(ctx, errors.InvalidParameter, op, "ID cannot contain a comma, semicolon or equals sign")
```
jefferai, line 333: This doesn't match the actual check. Not sure if you intended to check for semicolon/equal or not here.

johanbrandhorst: Good spot, I will update this error message. The parser makes it impossible for an ID to contain a semicolon or equals sign by this point.

PR #3670, opened 2023-09-05
feat (auth/ldap): add Repo support for max page size and deref aliases

jimlambrt:
Note: we're merging to a LLB.

This PR adds two new configuration fields for LDAP auth methods: `MaximumPageSize` and `DereferenceAliases` (see: [proto definitions](https://github.com/hashicorp/boundary/blob/899436b6bcfc0650d1cb31d4c7f5b6a0d20f7370/internal/auth/ldap/store/ldap.pb.go#L185))

After this PR, we'll add the new fields to the API.


internal/db/schema/migrations/oss/postgres/76/01_ldap.up.sql
```diff
@@ -0,0 +1,126 @@
...
+        references auth_ldap_method (public_id)
+        on delete cascade
+        on update cascade,
+    dereference_aliases text not null
+      constraint auth_ldap_deref_aliases_enm_fkey
+        references auth_ldap_deref_aliases_enm (name)
+        on delete cascade
+        on update cascade
+  );
+  comment on table auth_ldap_bind_credential is
```
tmessi, line 50: Is this supposed to be `comment on table auth_ldap_deref_aliases`?

tmessi (changes requested):

internal/db/schema/migrations/oss/postgres/76/01_ldap.up.sql
jimlambrt: why yes it is. ty!

tmessi (approved):

PR #3673, opened 2023-09-05
Targets delete when a persona changes address or user id

talanknight:
This PR refines the primary keys for personas so only 1 persona can exist per keystore/token name combination, and when the boundary address+user id values get updated the targets associated with the old address/user id gets deleted through a trigger.

This PR also breaks up the repository into a few files beyond the repository.go file.  One that handles the persona logic and another for the refresh logic.

### DIFF
--- a/internal/daemon/cache/repository_persona.go
+++ b/internal/daemon/cache/repository_persona.go
@@ -115,14 +115,13 @@ func (r *Repository) LookupPersona(ctx context.Context, tokenName, keyringType s
 		return nil, errors.Wrap(ctx, err, op)
 	}
 
-	if opts.withBoundaryAddress != "" && opts.withBoundaryAddress != p.BoundaryAddr {
+	switch {
+	case opts.withBoundaryAddress != "" && opts.withBoundaryAddress != p.BoundaryAddr:
 		// If we found a persona that doesn't have the provided address it
 		// is not the correct one, so the return should indicate the looked up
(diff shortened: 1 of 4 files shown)


internal/daemon/cache/repository_refresh.go
```diff
@@ -0,0 +1,81 @@
...
+	"context"
+	stderrors "errors"
+	"fmt"
+
+	"github.com/hashicorp/boundary/api"
+	"github.com/hashicorp/boundary/api/targets"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/observability/event"
+)
+
```
jimlambrt, line 16: should we add unit tests for the bits here?

internal/daemon/cache/repository_persona.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	if opts.withAuthTokenId != "" && opts.withAuthTokenId != p.AuthTokenId {
+		// If we found a persona that doesn't have the provided auth token id
+		// is not the correct one, so the return should indicate the looked up
+		// persona could not be found.
+		return nil, nil
+	}
+
+	if opts.withUpdateLastAccessedTime {
```
jimlambrt, line 132: consider a unit test for this?

jimlambrt (changes requested):

PR #3679, opened 2023-09-06
feat: add API support for additional LDAP auth method fields.

jimlambrt:
New Fields:

- **maximum_page_size** optionally specifies a maximum ldap search result size to use when retrieving the authenticated user's group memberships. This can be used to avoid reaching the LDAP server's max result size.
- **dereference_aliases** will control how aliases are dereferenced when performing the search. Possible values are: never, finding, searching, and always. If unset, a default of "never" is used. When set to "finding", it will only dereference aliases during name resolution of the base. When set to "searching", it will dereference aliases after name resolution.

There will be fast-follower PRs to add support for these new fields to the CLI and TF provider.


johanbrandhorst (commented):
The description says that it's adding config support but I don't see that change here?

johanbrandhorst (approved):

PR #3682, opened 2023-09-06
Add sessions to the client side cache

talanknight:


internal/daemon/cache/repository_refresh.go
```diff
@@ -35,6 +36,27 @@ func defaultTargetFunc(ctx context.Context, addr, token string) ([]*targets.Targ
 	return l.Items, nil
 }
```
jimlambrt, line 38: are there unit tests for these new bits?

jimlambrt (changes requested):
Looks good.  we just need to align on the FK bits.

PR #3685, opened 2023-09-07
feat (auth/ldap): add support for new fields to CLI

jimlambrt:
Added support for new flags for: max-page-size and deref-aliases

### DIFF
--- a/internal/cmd/commands/authmethodscmd/ldap_funcs.go
+++ b/internal/cmd/commands/authmethodscmd/ldap_funcs.go
@@ -9,8 +9,10 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"strings"
 
 	"github.com/hashicorp/boundary/api/authmethods"
+	"github.com/hashicorp/boundary/internal/auth/ldap"
 	"github.com/hashicorp/boundary/internal/cmd/base"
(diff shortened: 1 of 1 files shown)


internal/cmd/commands/authmethodscmd/ldap_funcs.go
```diff
@@ -369,6 +389,36 @@ func extraLdapFlagHandlingFuncImpl(c *LdapCommand, _ *base.FlagSets, opts *[]aut
 		*opts = append(*opts, authmethods.WithLdapAuthMethodGroupFilter(c.flagGroupFilter))
 	}
 
+	switch c.flagDerefAliases {
+	case "":
+	case "null":
```
johanbrandhorst, line 394: Maybe 
```suggestion
	case "", "null":
```
?

johanbrandhorst (approved):

internal/cmd/commands/authmethodscmd/ldap_funcs.go
jimlambrt: Hum... your suggestion isn't consistent with the pattern used throughout the "commands".  It works correctly, so for now I'll be consistent.

PR #3689, opened 2023-09-07
fix: Flaky upstream connection test

elimt:
The connectivity state changes for `upstream_connection_state` health field is causing the test to be flaky. Commenting out the field during assertions. `upstream_connection_state` health field is tested in other tests


johanbrandhorst (approved):
`cmp` 👌🏻

moduli (approved):

PR #3690, opened 2023-09-07
test(e2e): Add a test for a single-hop worker

moduli:
This PR adds a new e2e test that does the following...
- create a target with an egress worker filter and confirm that you can connect to it
- update the target to use a worker filter that can't reach the host and confirm that you cannot connect to it
- attempt to create a target with an ingress worker filter and confirm that the operation fails

A new enos scenario was also added to CI to support this test. This scenario creates the following...
```
docker network: e2e_db
- postgres database
- boundary controller

docker network: e2e_cluster
- boundary controller
- worker
- vault

docker network: e2e_host
- worker
- host
```

## One thing I ran into...
There were some differences when running this directly from your machine vs using the docker test runner. 

Without any changes to the worker config, tests were passing when running directly, but failing when using the docker test runner.
```
Messages:   	Unable to connect to worker at 0.0.0.0:9402
    kex_exchange_identification: read: Connection reset by peer
    Connection reset by 127.0.0.1 port 36119
        error fetching connection to send session teardown request to worker: Unable to connect to worker at 0.0.0.0:9402
```
Modifying the worker config file to include the container name resulted in it passing when using the docker test runner, but for running it locally, you do need to add an extra entry in `/etc/hosts` to include `127.0.0.1  localhost  worker`
```
public_addr = "${worker_name}:${port}"   # "worker:9402"
```

### DIFF
--- a/testing/internal/e2e/tests/base_with_worker/target_tcp_worker_connect_ssh_test.go
+++ b/testing/internal/e2e/tests/base_with_worker/target_tcp_worker_connect_ssh_test.go
@@ -18,6 +18,15 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// TestCliTcpTargetWorkerConnectTarget uses the boundary cli to do the
+// following...
+// - create a target with an egress worker filter and confirm that you can
+// connect to it
+// - update the target to use a worker filter that can't reach the host and
+// confirm that you cannot connect to it
+// - attempt to create a target with an ingress worker filter and confirm that
+// the operation fails
+// Note: This test is specific to the community version
 func TestCliTcpTargetWorkerConnectTarget(t *testing.T) {
 	e2e.MaybeSkipTest(t)
 	c, err := loadTestConfig()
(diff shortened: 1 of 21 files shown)


johanbrandhorst (approved):
Could you add a comment/message wherever the failure is when the `/etc/host` change is _not_ made when running this locally, to make the fix easily discoverable by the user?

testing/internal/e2e/tests/base_with_worker/target_tcp_worker_connect_ssh_test.go
```diff
@@ -0,0 +1,196 @@
...
+			"-remote-command", "hostname -i",
+			"--",
+			"-o", "UserKnownHostsFile=/dev/null",
+			"-o", "StrictHostKeyChecking=no",
+			"-o", "IdentitiesOnly=yes", // forces the use of the provided key
+		),
+	)
+	// Note: If this test fails due to: "Unable to connect to worker at
+	// worker:9402", modify your /etc/hosts file to contain...
+	// `127.0.0.1  localhost  worker``
```
stasryzhov, line 143: 👍

stasryzhov (approved):

PR #3695, opened 2023-09-08
Enable graceful shutdown in windows

talanknight:


internal/cmd/commands/daemon/stop_handler.go
```diff
@@ -0,0 +1,33 @@
...
+	case util.IsNil(cancelFn):
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "server cancelFn is missing")
+	}
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodPost {
+			writeError(w, "only method POST allowed", http.StatusMethodNotAllowed)
+		}
+
+		w.WriteHeader(http.StatusNoContent)
+		defer func() {
```
jefferai, line 29: Why not do the cancel inline here? The cancel function should be fast and async from any actual actions that happen as a result of it.

internal/cmd/commands/daemon/start.go
```diff
@@ -130,7 +130,8 @@ func (c *StartCommand) AutocompleteFlags() complete.Flags {
 
 func (c *StartCommand) Run(args []string) int {
 	const op = "daemon.(StartCommand).Run"
-	ctx := c.Context
+	ctx, cancel := context.WithCancel(c.Context)
```
jefferai, line 133: Rather than create another context here, can the handler just take `c.ContextCancel` and call it?

internal/cmd/commands/daemon/stop_handler.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/util"
+)
+
+// newStopHandlerFunc creates a handler that cancelles the server context.
```
jimlambrt, line 14:
```suggestion
// newStopHandlerFunc creates a handler that cancels the server context.
```

jimlambrt (approved):
One small change.

PR #3706, opened 2023-09-12
chore: update azurevaultkey wrapper dependency.

jimlambrt:
This update brings in [PR #155](https://github.com/hashicorp/go-kms-wrapping/pull/155) which adds support for azure workload identity authentication


moduli (commented):
How is Boundary impacted by the changes here (I'm not sure what parts of Boundary use this)? Should we validate anything before merging?

jimlambrt:
I don't think there's a direct impact to boundary.  There's just an additional option for azure key vault.

johanbrandhorst (approved):

moduli (approved):

PR #3708, opened 2023-09-12
test(e2e): Add worker health check when setting up infra

moduli:
This PR updates the e2e test suite to add a worker health check when setting up a docker worker container. This adds testing coverage on the `health` endpoint as well as ensures that the container is ready to use before running tests.

### DIFF
--- a/enos/modules/docker_postgres/main.tf
+++ b/enos/modules/docker_postgres/main.tf
@@ -63,6 +63,8 @@ resource "docker_container" "postgres" {
     "POSTGRES_USER=${var.user}",
     "POSTGRES_PASSWORD=${var.password}",
   ]
+  cpu_set = "1-2"
+  memory  = 2000
   volumes {
     host_path      = abspath(path.module)
     container_path = "/etc/postgresql/"
--- a/enos/modules/docker_worker/main.tf
+++ b/enos/modules/docker_worker/main.tf
@@ -63,6 +63,7 @@ resource "docker_image" "boundary" {
 
 locals {
   recording_storage_path = "/recordings"
+  port_ops               = var.port + 1
 }
 
 resource "docker_container" "worker" {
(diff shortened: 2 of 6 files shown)


enos/modules/docker_postgres/main.tf
```diff
@@ -63,6 +63,8 @@ resource "docker_container" "postgres" {
     "POSTGRES_USER=${var.user}",
     "POSTGRES_PASSWORD=${var.password}",
   ]
+  cpu_set = "1-2"
+  memory  = 2000
```
moduli, line 67: During testing in CI, observed one instance where the worker test flaked due to our favorite `unexpected rsv bits` error (the test did pass on the retry). We previously addressed this issue by having the tests run on RUNNER_LARGE.

I haven't been able to reproduce the flake yet (re-running it ~10 times). This change here is one idea to try to limit the resources used by other containers.

enos/modules/docker_worker/main.tf
```diff
@@ -115,6 +121,14 @@ resource "docker_container" "worker" {
   }
 }
 
+resource "enos_local_exec" "check_address" {
+  depends_on = [
+    docker_container.worker
+  ]
+
+  inline = ["timeout 10s bash -c 'until echo $(curl -s -i \"http://0.0.0.0:${local.port_ops}/health?worker_info=1\") | grep -i \\\"upstream_connection_state\\\":\\\"READY\\\"; do sleep 2; done'"]
```
johanbrandhorst, line 129: 🤢

johanbrandhorst (approved):

enos/modules/docker_worker/main.tf
moduli: Is there a better way to do this?

johanbrandhorst: Haha no it's fine, it's just the escapes within escapes 😂

moduli: Yea... I couldn't figure out another option

stasryzhov (approved):

PR #3715, opened 2023-09-12
test(e2e): Add missing proxy listener to config file

moduli:
While working on a multihop test, I discovered that I made an error in a boundary config file when I added a collocated worker back to the file. This PR adds the necessary proxy listener to the config file so that the worker can properly load.


stasryzhov (approved):

johanbrandhorst (approved):
:yolo_stamp:

PR #3716, opened 2023-09-12
event(observation): Add request / response observation and add telemetry filter

ajayreshc:
Changes:
* Define option: `WithTelemetry()` (to specify if telemetry events are enabled)
* Create new `filterProtoMessage` function to sanitise proto messages based on a filter.
* Add `Request` and `Response` to `observation` type.
* Add `Request` / `Response` handling to `ComposeFrom` to included filtered telemetry data.
* Unit tests for `filterProtoMessage` and `ComposeFrom`.

### DIFF
--- a/internal/observability/event/event_observation.go
+++ b/internal/observability/event/event_observation.go
@@ -4,7 +4,9 @@
 package event
 
 import (
+	"encoding/json"
 	"fmt"
+	"reflect"
 
 	"github.com/hashicorp/eventlogger"
 	"github.com/hashicorp/eventlogger/filters/gated"
@@ -115,6 +117,15 @@ func (o *observation) ComposeFrom(events []*eventlogger.Event) (eventlogger.Even
 				}
 			}
 			if g.Request.Details != nil {
+				res := recurseStructureWithTagFilter(
+					g.Request.Details,
+					map[string]string{
+						"eventstream": "observation",
+					},
+					false,
+				)
+				data, _ := json.Marshal(res)
+				fmt.Printf("RESULT request = %s", string(data))
 				msgReq.Details = g.Request.Details
 			}
 			if g.Request.Operation != "" {
@@ -140,6 +151,16 @@ func (o *observation) ComposeFrom(events []*eventlogger.Event) (eventlogger.Even
 				msgRes.StatusCode = g.Response.StatusCode
 			}
 			if g.Response.Details != nil {
+				res := recurseStructureWithTagFilter(
+					g.Response.Details,
+					map[string]string{
+						"eventstream": "observation",
+					},
+					false,
+				)
+				data, _ := json.Marshal(res)
+				fmt.Printf("RESULT response = %s \n", string(data))
+
 				msgRes.Details = g.Response.Details
 			}
 			if g.Response.DetailsUpstreamMessage != nil {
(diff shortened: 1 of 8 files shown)


internal/observability/event/telemetry_filter.go
```diff
@@ -0,0 +1,118 @@
...
+	// check to see if it's an exported struct field
+	if !fv.CanSet() {
+		return nil
+	}
+
+	fv.SetZero()
+
+	return nil
+}
+
```
sepehrfrgh, line 49: nitpick: Comment on here `recurseStructureWithProtoFilter` would be really useful

ajayreshc: I decided to remove the error return instead, because it had become unnecessary in filterValue()

internal/observability/event/telemetry_filter.go
```diff
@@ -0,0 +1,119 @@
...
+	"google.golang.org/protobuf/proto"
+)
+
+// protoFilter is a signature for a struct field validation test
+type protoFilter func(field reflect.StructField) bool
+
+// telemetryFilter checks a struct field should be included in observation telemetry data
+func telemetryFilter(field reflect.StructField) bool {
+	if field.Tag.Get("eventstream") == "observation" {
+		// log.Printf("%s is allowed by Tags (%s:\"%s\")", name, tag, tagValue)
```
johanbrandhorst, line 19: I assume this was left in here by mistake?

internal/observability/event/telemetry_filter.go
```diff
@@ -0,0 +1,119 @@
...
+
+func filterProtoMessage(msg proto.Message, filterFunc protoFilter) (proto.Message, error) {
+	if msg == nil {
+		return nil, errors.New("nil message")
+	}
+
+	cloneMsg := proto.Clone(msg)
+	err := recurseStructureWithProtoFilter(reflect.ValueOf(cloneMsg), filterFunc, false)
+	return cloneMsg, err
+}
```
johanbrandhorst, line 119: I'm late to the design of this feature, but it seems to me that we could do this better with a proper protobuf option and protobuf reflection rather than using a struct tag and Go reflection. All our protobuf generation uses the Go APIv2, which supports rich reflection of protobuf options on generated structs. See https://go.dev/blog/protobuf-apiv2 for an introduction and a simple example of doing almost exactly what we're doing here, but using a protobuf option instead of a struct tag. Perhaps this was considered already and rejected, but personally I think I'd prefer this to the struct tags that we're using now, which is a bit of a hack (it's not portable to other protobuf languages and requires post-processing of our generated files). There is a precedent for using protobuf options in our protobuf files for custom logic such as this. What do you think?

sepehrfrgh: Thank you for the comment, Johan. This is an interesting idea and I was not aware of this before. Tomorrow is our telemetry poodle, I would be really happy to invite you over if you are available and we talk about it. I will start the thread on Slack so we can talk more about it. Quick note that we are pretty close to the deadline to deliver this feature and I am not sure about switching to protobuf options at this time but we can definitely talk about it tomorrow

johanbrandhorst (approved):

internal/observability/event/telemetry_filter.go
jimlambrt: I agree with @johanbrandhorst: we should remove this commented out code.

internal/observability/event/event_observation.go
```diff
@@ -102,6 +110,55 @@ func (o *observation) ComposeFrom(events []*eventlogger.Event) (eventlogger.Even
 				Payload:   g.Detail,
 			})
 		}
+		if g.Request != nil {
+			msgReq := &Request{}
+			if v, ok := payload["request"]; ok {
```
jimlambrt, line 115: perhaps make `request` a const? (see DetailsField)

internal/observability/event/event_observation.go
```diff
@@ -102,6 +110,55 @@ func (o *observation) ComposeFrom(events []*eventlogger.Event) (eventlogger.Even
...
+			if g.Request.Operation != "" {
+				msgReq.Operation = g.Request.Operation
+			}
+			if g.Request.Endpoint != "" {
+				msgReq.Endpoint = g.Request.Endpoint
+			}
+			if g.Request.DetailsUpstreamMessage != nil {
+				msgReq.DetailsUpstreamMessage = g.Request.DetailsUpstreamMessage
+			}
+			payload["request"] = msgReq
```
jimlambrt, line 137: convert to const?

internal/observability/event/event_observation.go
```diff
@@ -102,6 +110,55 @@ func (o *observation) ComposeFrom(events []*eventlogger.Event) (eventlogger.Even
...
+		}
+		if g.Response != nil {
+			msgRes := &Response{}
+			if v, ok := payload["response"]; ok {
```
jimlambrt, line 141: perhaps make `response` a const? (see DetailsField)

internal/observability/event/event_observation.go
```diff
@@ -102,6 +110,55 @@ func (o *observation) ComposeFrom(events []*eventlogger.Event) (eventlogger.Even
...
+				filteredResponse, err := filterProtoMessage(g.Response.Details, telemetryFilter)
+				if err != nil {
+					continue
+				}
+				msgRes.Details = filteredResponse
+			}
+			if g.Response.DetailsUpstreamMessage != nil {
+				msgRes.DetailsUpstreamMessage = g.Response.DetailsUpstreamMessage
+			}
+			payload["response"] = msgRes
```
jimlambrt, line 160: convert to const?

internal/observability/event/telemetry_filter.go
jimlambrt: the error seems to still be here...

jimlambrt: I'd love a comment on this recurseStructureWithProtoFilter as well.

internal/observability/event/telemetry_filter.go
```diff
@@ -0,0 +1,145 @@
...
+			}
+			// if slice is empty after processing, we can zero its length
+			if zeroCount == value.Len() && kind == reflect.Slice {
+				value.SetLen(0)
+			}
+		}
+	case reflect.Struct:
+		fields := value.Type()
+		num := fields.NumField()
+		for i := 0; i < num; i++ {
```
jimlambrt, line 106: consider removing `num := fields.NumField()` and then you can just:

```suggestion
		for i := 0; i < fields.NumField(); i++ {
```

internal/observability/event/telemetry_filter.go
```diff
@@ -0,0 +1,145 @@
...
+	if fv == reflect.ValueOf(nil) {
+		return
+	}
+
+	if fv.Kind() == reflect.Ptr {
+		fv = fv.Elem()
+	}
+
+	// check to see if it's an exported struct field
+	if !fv.CanSet() {
```
jimlambrt, line 56: ty!

internal/observability/event/telemetry_filter.go
```diff
@@ -0,0 +1,145 @@
...
+func recurseStructureWithProtoFilter(value reflect.Value, filterFunc protoFilter, isObservable bool) error {
+	kind := value.Kind()
+
+	switch kind {
+	case reflect.Interface, reflect.Ptr:
+		value = value.Elem()
+		return recurseStructureWithProtoFilter(value, filterFunc, isObservable)
+	case reflect.Map:
+		for _, k := range value.MapKeys() {
+			mVal := value.MapIndex(k)
```
jimlambrt, line 80: Should we have unit tests for this code path?

jimlambrt (changes requested):
Looks amazing.  I left some minor comments, but I did have one bit that I think needs an additional unit test.

internal/observability/event/event_observation.go
ajayreshc: Done

internal/observability/event/event_observation.go
ajayreshc: Done.

internal/observability/event/event_observation.go
ajayreshc: Done.

internal/observability/event/event_observation.go
ajayreshc: Done.

internal/observability/event/telemetry_filter.go
ajayreshc: Yes it was. It's been removed.

internal/observability/event/telemetry_filter.go
ajayreshc: I've added a comment. Apologies for the confusion too, I was referring to the error return in filterValue()

internal/observability/event/telemetry_filter.go
ajayreshc: I added map tests to telemetry_filter_test.go, and also changed to MapIter() in the latest version.

internal/observability/event/telemetry_filter.go
ajayreshc: Done.

jimlambrt (approved):
Looks great!  There's a minor `make gen` issue with sorting of includes which should be resolved before merging.

sepehrfrgh (approved):
LGTM Thank you April

PR #3719, opened 2023-09-13
feat(event): new storage event type

ddebko:
### Summary

This PR adds a new event type for storage, adding the ability to introduce new events for session recording.

### DIFF
--- a/internal/event/event_type.go
+++ b/internal/event/event_type.go
@@ -16,12 +16,13 @@ const (
 	AuditType       Type = "audit"       // AuditType represents audit events
 	ErrorType       Type = "error"       // ErrorType represents error events
 	SystemType      Type = "system"      // SysType represents system events
+	StorageType     Type = "storage"     // StorageType represents storage events
 )
 
 func (et Type) Validate() error {
 	const op = "event.(Type).Validate"
 	switch et {
-	case EveryType, ObservationType, AuditType, ErrorType, SystemType:
+	case EveryType, ObservationType, AuditType, ErrorType, SystemType, StorageType:
 		return nil
 	default:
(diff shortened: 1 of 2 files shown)


internal/event/event_type.go
```diff
@@ -16,12 +16,13 @@ const (
 	AuditType       Type = "audit"       // AuditType represents audit events
 	ErrorType       Type = "error"       // ErrorType represents error events
 	SystemType      Type = "system"      // SysType represents system events
+	StorageType     Type = "storage"     // StorageType represents session recording events
```
irenarindos, line 19: Should this maybe say session recording storage events? Or just storage events?

internal/event/event_type_test.go
```diff
@@ -0,0 +1,66 @@
...
+package event
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestType_Validate(t *testing.T) {
```
irenarindos, line 13: Thanks for adding these!

internal/event/event_type.go
ddebko: I was wondering this too. The reason why I chose to keep it vague is because storage could potentially be used for more than just session recordings. I'm not sure if that's a good enough reason to keep it vague though. I'm happy to change it to `SessionRecordingStorageType Type = "session_recording_storage_type"`

irenarindos: Ah sorry, I meant the comment indicating that it's a session recording event; should the comment just say that it represents a storage event?

ddebko: Good catch, sorry for misreading your original feedback. I changed it to say `StorageType represents storage events`

irenarindos (approved):

PR #3722, opened 2023-09-13
Telemetry: Adding telemetry flag and and telemetry sink event type

sepehrfrgh:
In this PR, we are adding a new event type called telemetry to both flag and sink types.
Telemetry event is associated with observation events and the following logic must be applied 

- Case 1: `Observation` flag is true, `Telemetry` flag is false -> This is a current and default behavior and we should just get the observation events without telemetry events

- Case 2: `Observation` flag is false, `Telemetry` flag is true -> In this situation, we throw an error saying that the Observation flag should be set to true if you want to receive telemetry events

     - From CLI if the flag is set with `--observation-events false --telemetry-events true` user will receive the following error message: ` event.(EventerConfig).Validate: telemetry events require observation event to be enabled: invalid parameter`
     - From sink event config if the user `event_types` specifies `telemetry` without specifying the `observation`
It will receive the following error message: 
`event.(SinkConfig).Validate: telemetry event type requires observation event type to be specified: invalid parameter`

- Case 3: `Observation` flag is true, `Telemetry` flag is true -> In this situation, the user will receive an observation event including telemetry events

### DIFF
--- a/internal/observability/event/sink_config.go
+++ b/internal/observability/event/sink_config.go
@@ -6,6 +6,7 @@ package event
 import (
 	"fmt"
 	"io"
+	"slices"
 	"time"
 )
 
@@ -77,6 +78,11 @@ func (sc *SinkConfig) Validate() error {
 	}
 	if len(sc.EventTypes) == 0 {
 		return fmt.Errorf("%s: missing event types: %w", op, ErrInvalidParameter)
+	}
+
+	// This checks if the telemetry event is specified in the sink but the observation event is not specified.
+	if !slices.Contains(sc.EventTypes, ObservationType) && slices.Contains(sc.EventTypes, TelemetryType) {
+		return fmt.Errorf("%s: telemetry event type requires observation event type to be specified: %w", op, ErrInvalidParameter)
 	}
 
 	for _, et := range sc.EventTypes {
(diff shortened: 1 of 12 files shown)


internal/observability/event/sink_config.go
```diff
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"io"
 	"time"
+
+	"golang.org/x/exp/slices"
```
johanbrandhorst, line 11: We use Go 1.21 so you can use the standard library `slices` package

internal/observability/event/sink_config.go
```diff
@@ -79,6 +81,11 @@ func (sc *SinkConfig) Validate() error {
 		return fmt.Errorf("%s: missing event types: %w", op, ErrInvalidParameter)
 	}
 
+	// this checks if the telemetry event is specified in sink but observation event is not specified
```
johanbrandhorst, line 84:
```suggestion
	// This checks if the telemetry event is specified in the sink but the observation event is not specified.
```

johanbrandhorst (changes requested):
Are we adding some documentation about this new event type?

johanbrandhorst (approved):

johanbrandhorst:
Did you see my comment about docs?

sepehrfrgh:
Yes, if you referring to boundary documentation, I already created the ticket and assigned it to Dan to update https://developer.hashicorp.com/boundary/docs/configuration/events.
Also, I think after we merge this to`llb-telemetry` we need to write these changes to changelog.
Please let me know I need to update doc in any other places Thanks

jimlambrt (approved):

PR #3724, opened 2023-09-14
test(e2e): Refactor vault setup method

moduli:
This PR makes an adjustment to the test suite to support tests for multihop.

The primary change is to the `vault.Setup` method, allowing it to pass in the desired `boundary-controller-policy` file. This enables test packages in `boundary-enterprise` to pass in their own file.


stasryzhov (approved):
👍

PR #3728, opened 2023-09-14
remove deprecated option application-credential-source

kheina:


internal/proto/controller/api/resources/targets/v1/target.proto
```diff
@@ -170,18 +170,13 @@ message Target {
...
   ]; // @gotags: `class:"public"`
 
-  // Output only. The IDs of the application credential source ids associated with this Target.
   // Deprecated use "brokered_credential_source_ids" instead.
-  repeated string application_credential_source_ids = 400 [
-    json_name = "application_credential_source_ids",
-    deprecated = true
-  ]; // @gotags: `class:"public"`
-  // Output only. The application credential sources associated with this Target.
+  reserved "application_credential_source_ids";
```
louisruch, line 174: I believe the standard is to include all these reserved fields at the bottom of this Message under the `// Deprecated fields` comment

internal/proto/controller/api/services/v1/target_service.proto
```diff
@@ -272,10 +272,8 @@ message AddTargetCredentialSourcesRequest {
   uint32 version = 2; // @gotags: `class:"public"`
 
   // Deprecated use brokered_credential_source_ids instead
-  repeated string application_credential_source_ids = 3 [
-    json_name = "application_credential_source_ids",
-    deprecated = true
-  ]; // @gotags: `class:"public"`
+  reserved "application_credential_source_ids";
```
louisruch, line 275: Similar re adding below, will not comment on the others here

kheina:
> Some minor feedback on moving the reserved deprecated fields, other than that:
> 
>     * You should just confirm there are no enterprise `ssh` specific tests that will fail when `Application credentials are removed`
> 
>     * Add a changelog that this was removed

I didn't check ent, so I'll look there to make sure. I did find a scenario in local env and a tf doc that had it. iirc everywhere else already had it listed as deprecated. I'll double check to make sure the docs say it's been removed if it only says deprecated

louisruch (approved):
This is great thanks!

PR #3730, opened 2023-09-14
Revert "Backport of SSH and TCP definition for Targets into release/0.13.x"

Dan-Heath:
Reverts hashicorp/boundary#3726


Dan-Heath:
We saw some unusual behavior on #3726 , and I thought it needed to be reverted. But I confirmed with Johan that it is not necessary, so I am closing this PR.

PR #3737, opened 2023-09-15
Change cache schema from personas to tokens+users

talanknight:
This PR breaks apart users and tokens.  The cli and the cache daemon's api only operate on tokens and boundary addresses.  The repo keeps the concept of users internal to it and maintains them based on the state of the tokens provided by the caller.  This allows resource refreshes to be performed on a per user basis while allowing the cli to call the api using only the information already present or passed in through env variables or flags (boundary address, keyring type/token names, etc)

In an upcoming PR, when an auth token is first seen by the repo, a call to boundary is made to validate that the user id reported in the auth token is correct before persisting it.  For now we trust the user id reported in the auth token resource stored in the keyring.

### DIFF
--- a/internal/daemon/cache/repository_token.go
+++ b/internal/daemon/cache/repository_token.go
@@ -99,8 +99,6 @@ func (r *Repository) AddKeyringToken(ctx context.Context, bAddr string, token Ke
 			}
 		}
 
-		// TODO: Tokens are probably not the thing to limit here.  Do we want to limit users?
-		// Also, we can perform this check as an after create trigger on the user table
 		var users []*user
 		if err := reader.SearchWhere(ctx, &users, "true", []any{}, db.WithLimit(-1)); err != nil {
 			return errors.Wrap(ctx, err, op)


internal/cmd/commands/daemon/addtoken.go
```diff
@@ -77,7 +77,7 @@ func (c *AddPersonaCommand) Run(args []string) int {
 	return base.CommandSuccess
 }
 
-func (c *AddPersonaCommand) AddPersona(ctx context.Context) (*api.Error, error) {
+func (c *AddTokenCommand) AddPersona(ctx context.Context) (*api.Error, error) {
```
jimlambrt, line 80: Perhaps just `Add(...)`?

jimlambrt (approved):
One small ref to persona.  Looks great.

internal/daemon/cache/schema.sql
```diff
@@ -2,84 +2,72 @@
...
-  keyring_type text not null,
-  token_name text not null,
-  boundary_addr text not null,
-  auth_token_id text not null,
-  user_id text not null,
+-- user contains the boundary user information for the boundary user that owns
+-- the information in the cache.
+create table if not exists user (
+  id text not null primary key
+    check (length(id) > 0),
```
mgaffney, line 9: Where does the `id` come from?

internal/daemon/cache/schema.sql
```diff
@@ -2,84 +2,72 @@
...
+
+-- token contains the token information for a user
+create table if not exists token (
+  keyring_type text not null
+    check (length(keyring_type) > 0),
+  token_name text not null
+    check (length(token_name) > 0),
+  auth_token_id text not null
+    check (length(auth_token_id) > 0),
+  user_id text not null
```
mgaffney, line 22: Can the same `auth_token_id` belong to more than one `user_id`?

internal/daemon/cache/schema.sql
```diff
@@ -2,84 +2,72 @@
...
+end;
+
+create trigger token_delete_delete_orphaned_users after delete on token
+begin
+delete from user
+where
+    id not in (select user_id from token);
+end;
+
+create table if not exists target (
```
mgaffney, line 44: Can you add a comment about what this table contains?

internal/daemon/cache/schema.sql
```diff
@@ -2,84 +2,72 @@
...
-    boundary_user_id = old.user_id;
-end;
-
-create table if not exists cache_session (
-  keyring_type text not null,
-  token_name text not null,
-  boundary_addr text not null,
-  boundary_user_id text not null,
-  id text not null,
+create table if not exists session (
```
mgaffney, line 57: Can you add a comment about what this table contains?

internal/daemon/cache/schema.sql
```diff
@@ -2,84 +2,72 @@
...
+delete from user
+where
+    id not in (select user_id from token);
+end;
+
+create table if not exists target (
+  user_id text not null
+    references user(id)
+    on delete cascade,
+  id text not null
```
mgaffney, line 48: Where does this `id` come from?

internal/daemon/cache/schema.sql
```diff
@@ -2,84 +2,72 @@
...
-  keyring_type text not null,
-  token_name text not null,
-  boundary_addr text not null,
-  boundary_user_id text not null,
-  id text not null,
+create table if not exists session (
+  user_id text not null
+    references user(id)
+    on delete cascade,
+  id text not null
```
mgaffney, line 61: Where does this `id` come from?

mgaffney (changes requested):

mgaffney (approved):
Nice work!

jimlambrt (approved):
Looks great. Ty for making these changes.

PR #3740, opened 2023-09-15
all: refactor credential library listing

johanbrandhorst:
The new structure separates the methods that operate on a generic credential library entity out of the vault package and into the credential package. This isn't strictly necessary for credential libraries since they all live in the vault repo at the moment, but this structure is necessary for credential stores and it makes more sense from a structure point to do it for both resources.


internal/daemon/controller/common/common.go
```diff
@@ -19,20 +20,21 @@ import (
...
-	PasswordAuthRepoFactory        func() (*password.Repository, error)
-	ServersRepoFactory             func() (*server.Repository, error)
-	StaticRepoFactory              func() (*static.Repository, error)
-	PluginHostRepoFactory          func() (*pluginhost.Repository, error)
-	PluginRepoFactory              func() (*plugin.Repository, error)
-	ConnectionRepoFactory          func() (*session.ConnectionRepository, error)
-	WorkerAuthRepoStorageFactory   func() (*server.WorkerAuthRepositoryStorage, error)
-	PluginStorageBucketRepoFactory func() (*pluginstorage.Repository, error)
+	AuthTokenRepoFactory                = oidc.AuthTokenRepoFactory
+	GenericCredentialLibraryRepoFactory func() (*credential.CredentialLibraryRepository, error)
```
johanbrandhorst, line 24: Leaning towards calling this something else to avoid the use of the word "generic". `Root`? `Shared`? It's the repository that operates on all credential libraries.

mikemountain: `Base`?

mikemountain (approved):

internal/credential/repository_credentiallibrary.go
```diff
@@ -0,0 +1,88 @@
...
+	"time"
+
+	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/db/timestamp"
+	"github.com/hashicorp/boundary/internal/errors"
+)
+
+// A CredentialLibraryRepository provides read only information
+// generic to all credential store types.
+type CredentialLibraryRepository struct {
```
mgaffney, line 17: Why `CredentialLibraryRepository` and not just `Repository`?

internal/credential/repository_credentiallibrary.go
```diff
@@ -0,0 +1,88 @@
...
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "db.Reader")
+	}
+
+	return &CredentialLibraryRepository{
+		reader: r,
+	}, nil
+}
+
+// Now returns the current timestamp in the DB.
+func (clr *CredentialLibraryRepository) Now(ctx context.Context) (time.Time, error) {
```
mgaffney, line 35: Will we need to repeat this code in other repositories?

internal/credential/repository_credentiallibrary.go
johanbrandhorst: Because there will be a `CredentialStoreRepository` and `CredentialRepository` alongside this soon. They all have to answer the question "How many of X are there?" independently, so I thought naming the repository was better than having `GetTotalCredentialStores()`, `GetTotalCredentialLibraries()` on a single meta-repository. What do you think?

internal/credential/repository_credentiallibrary.go
johanbrandhorst: Yes, we repeat this in a lot of different repositories. Do you think we should make a repository for just this part that we could use in the handlers?

internal/credential/repository_credentiallibrary.go
johanbrandhorst: It's more consistent with other repositories, which generally deal with a single resource type (e.g. ssh, tcp, oidc, password).

johanbrandhorst: Here's the credential store implementation: https://github.com/hashicorp/boundary/pull/3742

johanbrandhorst: And credentials: https://github.com/hashicorp/boundary/pull/3747

internal/daemon/controller/handlers/credentiallibraries/credentiallibrary_service.go
```diff
@@ -152,6 +167,10 @@ func (s Service) ListCredentialLibraries(ctx context.Context, req *pbs.ListCrede
 	if err != nil {
 		return nil, errors.Wrap(ctx, err, op)
 	}
+	genericRepo, err := s.baseCredentialLibraryRepoFn()
```
mikemountain, line 170: Nit/question: should we change `genericRepo` to `baseRepo` to keep it consistent?

internal/credential/vault/credential_library.go
```diff
@@ -134,13 +133,3 @@ type sortCredentialLibrary struct {
 	publicId   string
 	updateTime time.Time
 }
-
-type deletedCredentialLibrary struct {
```
mikemountain, line 138: Do you think it would make sense to also move `sortCredentialLibrary` into the new credential package, and call it something like `sortCredentialItem`? Since the only values are `publicId` and `updateTime` I think it should be okay to share across

internal/daemon/controller/handlers/credentiallibraries/credentiallibrary_service.go
johanbrandhorst: doh, thanks

internal/credential/vault/credential_library.go
johanbrandhorst: Honestly down that path lies a more generic sort item used across all repositories. I think we should consider that independently and leave this as is for now.

johanbrandhorst:
Merging this for now, we can make more changes before merging the LLB

PR #3756, opened 2023-09-20
Telemetry: Add request and response interceptors with deny filter update

sepehrfrgh:
This PR will add 

- request and response events into WriteObservation 
- gRPC `request_info` into the observation pipeline
- default `deny` filter for when `observation_events` is enabled and `worker_status` events emitting from worker to controller 
   `"/type" contains "observation" and "/data/request_info/method" contains "ServerCoordinationService/Status"`
- if the user specifies the `deny` filter, it will overwrite the default `deny` will be available

### DIFF
--- a/internal/observability/event/cloudevents_formatter_node.go
+++ b/internal/observability/event/cloudevents_formatter_node.go
@@ -69,12 +69,13 @@ func newCloudEventsFormatterFilter(source *url.URL, format cloudevents.Format, o
 			}
 			n.deny = append(n.deny, f)
 		}
+	} else {
+		defaultDenyFilters, err := defaultCloudEventsDenyFilters()
+		if err != nil {
+			return nil, err
+		}
+		n.deny = append(n.deny, defaultDenyFilters...)
 	}
-	defaultDenyFilters, err := defaultCloudEventsDenyFilters()
-	if err != nil {
-		return nil, err
-	}
-	n.deny = append(n.deny, defaultDenyFilters...)
 	n.Predicate = newPredicate(n.allow, n.deny)
 	return &n, nil
 }
@@ -83,7 +84,7 @@ func defaultCloudEventsDenyFilters() ([]*filter, error) {
 	const (
 		op = "event.defaultCloudEventsDenyFilters"
 		// denyWorkStatusEvents is a default filter for worker to controller API status requests
-		denyWorkStatusEvents = `"/Data/RequestInfo/Method" contains "ServerCoordinationService/Status"`
+		denyWorkStatusEvents = `"/type" contains "observation" and "/data/request_info/method" contains "ServerCoordinationService/Status"`
 	)
 	f, err := newFilter(denyWorkStatusEvents)
 	if err != nil {
--- a/internal/observability/event/context.go
+++ b/internal/observability/event/context.go
@@ -109,11 +109,11 @@ func WriteObservation(ctx context.Context, caller Op, opt ...Option) error {
 		return fmt.Errorf("%s: specify either header or details options or request or response for an event payload: "+
 			"%w", op, ErrInvalidParameter)
 	}
-	// For the case that telemetry is not enabled, and we have events coming form interceptors.
+	// For the case that the telemetry is not enabled, and we have events coming from interceptors.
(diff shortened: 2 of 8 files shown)


internal/observability/event/cloudevents_formatter_node.go
```diff
@@ -69,12 +69,13 @@ func newCloudEventsFormatterFilter(source *url.URL, format cloudevents.Format, o
 			}
 			n.deny = append(n.deny, f)
 		}
+	} else {
```
sepehrfrgh, line 72: @jimlambrt This is the place where I am checking if there are no user-specified `deny` filters, we add the default `deny` filter. since it is in `else` for the case that the user specifies the `deny` filter. We no longer have this default filter. I wanted to double-check the logic with you.
Another scenario could be the user specifying the `deny` filter when the observation event is enabled which with this logic, we will override the default again and we emit the `ServerCoordinationService/Status`

internal/observability/event/context.go
```diff
@@ -105,8 +105,17 @@ func WriteObservation(ctx context.Context, caller Op, opt ...Option) error {
 		}
 	}
 	opts := getOpts(opt...)
-	if opts.withDetails == nil && opts.withHeader == nil && !opts.withFlush {
-		return fmt.Errorf("%s: specify either header or details options for an event payload: %w", op, ErrInvalidParameter)
+	if opts.withDetails == nil && opts.withHeader == nil && !opts.withFlush && opts.withRequest == nil && opts.withResponse == nil {
+		return fmt.Errorf("%s: specify either header or details options or request or response for an event payload: "+
+			"%w", op, ErrInvalidParameter)
+	}
+	// For the case that telemetry is not enabled, and we have events coming form interceptors.
```
sepehrfrgh, line 112: @jimlambrt In here I am checking if the telemetry is not enabled but we are having requests and responses. I am returning `nil`. Is it necessary to add a log here?

internal/observability/event/cloudevents_formatter_node.go
jimlambrt: I wasn't aware that we were already filtering ServerCooridnationService msgs out
of audit events.  

Given that, I'm not sure our change is quite right.  Perhaps
we should add `"/type" contains "observation"` to the original default filter,
so it only applies to audit events and we should always add that filter because
that's the current behavior.  

Then we can add the conditional addition of the new observation filter.  Make sense?

internal/observability/event/context.go
```diff
@@ -105,8 +105,17 @@ func WriteObservation(ctx context.Context, caller Op, opt ...Option) error {
...
+	if !eventer.conf.TelemetryEnabled && (opts.withRequest != nil || opts.withResponse != nil) {
+		return nil
+	}
+	// If telemetry in enabled we add it to the optiones
```
jimlambrt, line 116:
```suggestion
	// If telemetry in enabled we add it to the options
```

internal/observability/event/context.go
jimlambrt: No need to log anything.

jimlambrt (changes requested):
Looks pretty good overall.  I had a few suggestions.

internal/observability/event/cloudevents_formatter_node.go
ajayreshc: Would that be `"/type" contains "audit"` on the first one?

ajayreshc: So, in short... The existing default filter path was incorrectly defined, and didn't work? Which would mean we haven't been auto-filtering previously... 

Technically then, default adding the observation filter if no other deny filters exist would be correct I think (deferring to @jimlambrt here), but also raises the question of whether or not a ticket needs to be raised within backend to fix the behaviour for audit events.

jimlambrt: If the exiting path was broken, then you can just ignore my comment/concern.  Your changes are perfectly fine.

ajayreshc: I just confirmed against `main` that this is indeed the case. Running `boundary dev -audit-events true` and seeing the  status events coming through.

ajayreshc (approved):
Thanks Sepehr

PR #3757, opened 2023-09-20
all: refactor accounts to use a base repository

mikemountain:
Similar case to #3740, the new structure separates the methods that operate on a generic account entity out of the ldap/pw/oidc packages and into the auth package


johanbrandhorst:
I think we're abandoning this change, right?

PR #3758, opened 2023-09-20
Fix flaky TestServer_ReloadInitialUpstreams

jefferai:
With the KMS changes this should never have worked at all (because the different databases means the credentials conveyed after the KMS-based authentication would never be valid). However, it was sometimes passing because of a race condition: the worker could be found in the database before the new credentials were returned to the worker, causing the worker to start at the KMS step with the second controller.

This changes the test to use the same database for two controllers, but to only ever have one running at a time, and uses a sequence of two changed (or unchanged, depending) status updates from the worker to ensure that the reload was successful.

### DIFF
--- a/internal/cmd/commands/server/worker_initial_upstreams_reload_test.go
+++ b/internal/cmd/commands/server/worker_initial_upstreams_reload_test.go
@@ -11,8 +11,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/hashicorp/boundary/internal/daemon/controller"
 	"github.com/hashicorp/boundary/internal/server"
-	"github.com/hashicorp/boundary/testing/controller"
+	tc "github.com/hashicorp/boundary/testing/controller"
 	"github.com/mitchellh/cli"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/atomic"
@@ -51,25 +52,18 @@ func TestServer_ReloadInitialUpstreams(t *testing.T) {
 	recoveryWrapper, _ := wrapperWithKey(t)
 	workerAuthWrapper, key := wrapperWithKey(t)
 
-	// Create two controllers, each with their own database. In practice it
-	// would be odd to have separate databases, but it makes it easy for the
-	// test to assert that the worker has connected to the second controller
-	// after a reload signal is sent.
-	testController := controller.NewTestController(
+	testController := tc.NewTestController(
 		t,
-		controller.WithWorkerAuthKms(workerAuthWrapper),
-		controller.WithRootKms(rootWrapper),
-		controller.WithRecoveryKms(recoveryWrapper),
+		tc.WithWorkerAuthKms(workerAuthWrapper),
+		tc.WithRootKms(rootWrapper),
+		tc.WithRecoveryKms(recoveryWrapper),
+		tc.DisableDatabaseDestruction(),
 	)
-	defer testController.Shutdown()
-	testController2 := controller.NewTestController(
-		t,
-		controller.WithWorkerAuthKms(workerAuthWrapper),
-		controller.WithRootKms(rootWrapper),
-		controller.WithRecoveryKms(recoveryWrapper),
-	)
-	defer testController2.Shutdown()
-	require.NotEqual(testController.Config().DatabaseUrl, testController2.Config().DatabaseUrl)
+	t.Cleanup(testController.Shutdown)
+
+	testController2 := testController.AddClusterControllerMember(t, &controller.TestControllerOpts{
+		DisableAutoStart: true,
+	})
 
 	wg := &sync.WaitGroup{}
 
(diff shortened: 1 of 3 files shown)


internal/daemon/controller/listeners.go
```diff
@@ -276,13 +276,13 @@ func (c *Controller) configureForCluster(ln *base.ServerListener) (func(), error
 		}
 		go func() {
 			err := splitListener.Start()
-			if err != nil {
+			if !errors.Is(err, net.ErrClosed) {
```
tmessi, line 279: Should this still check for `nil` as well? I don't know if `Start` ever returns a `nil` error, but it might in the future?

internal/daemon/controller/listeners.go
```diff
@@ -276,13 +276,13 @@ func (c *Controller) configureForCluster(ln *base.ServerListener) (func(), error
...
 				event.WriteError(c.baseContext, op, err, event.WithInfoMsg("splitListener.Start() error"))
 			}
 		}()
 		go func() {
 			err := ln.GrpcServer.Serve(metric.InstrumentClusterTrackingListener(multiplexingAuthedListener, grpcListenerPurpose))
-			if err != nil {
+			if !errors.Is(err, net.ErrClosed) {
```
tmessi, line 285: Same here, should this also check for `nil`? The docs for `grpc.(Server).Serve` state:
> Serve will return a non-nil error unless Stop or GracefulStop is called.

So it seems it will return nil if stopped.

tmessi (commented):
Just a question about the changes to the error handling, otherwise looks good to me.

internal/daemon/controller/listeners.go
jefferai: `errors.Is` won't match if it's nil, so I think it still serves that function...I might be missing something here.

There's nothing explicitly wrong with this message, but it throws an error in the log on normal shutdown...I noticed this while looking at the tests, and it seems nice to not error when expected. There is a slight change we could actually hide a real issue with the network listener closing but it seems unlikely that if there was an issue that we'd start the listener in the first place.

internal/daemon/controller/listeners.go
jefferai: It does seem like that should be the case, but existing behavior seems to not agree:
```
{"id":"i9CIGvYZ3D","source":"https://hashicorp.com/boundary/arch/worker","specversion":"1.0","type":"error","data":{"error":"use of closed network connection","error_fields":{},"id":"e_8Jo7RdIvuk","version":"v0.1","op":"controller.configureForCluster","info":{"msg":"multiplexingAuthedListener error"}},"datacontentype":"application/cloudevents","time":"2023-09-20T13:40:26.300154787-04:00"}
```
I didn't chase this down too much other than it's the grpc server set on the cluster listener and we seem to be calling GracefulStop on it.

internal/daemon/controller/listeners.go
tmessi: > `errors.Is` won't match if it's nil

Right so, `errors.Is(nil, net.ErrClosed)` will be `false`, but we are doing `!errors.Is` so this `if` will be `true` when the error is `nil`. I don't think we want to write an error event in that case.

internal/cmd/commands/server/worker_initial_upstreams_reload_test.go
```diff
@@ -91,25 +85,73 @@ func TestServer_ReloadInitialUpstreams(t *testing.T) {
...
-				timeout.Stop()
-				break pollFirstController
+				switch {
+				case lastStatusTime.IsZero():
+					lastStatusTime = w.GetLastStatusTime().AsTime().Round(time.Second)
+					t.Log("recording initial last status time", lastStatusTime.Format(time.RFC3339))
+				default:
+					if !lastStatusTime.Equal(w.GetLastStatusTime().AsTime().Round(time.Second)) {
+						lastStatusTime = w.GetLastStatusTime().AsTime().Round(time.Second)
+						t.Log("recording updated last status time", lastStatusTime.Format(time.RFC3339Nano))
```
moduli, line 112: Any reason with this one is using `RFC3339Nano` whereas the first log is using `RFC3339`?

moduli (approved):
Ran the job in CI a few more times and it looks like they are all resulting in a pass.

internal/daemon/controller/listeners.go
jefferai: Oh I see what you're saying. Gotcha. Yes, will change!

jefferai: Fixed

internal/cmd/commands/server/worker_initial_upstreams_reload_test.go
jefferai: Ah, nope. I'll just remove them actually.

tmessi (approved):

PR #3759, opened 2023-09-20
docs: Update worker auth storage path

Dan-Heath:
From a Slack conversation: https://hashicorp.slack.com/archives/C01AQDJF3SA/p1694709178068199

The example we use for the `auth_storage_path` is `/boundary/demo-worker-1`. However, this assumes the Boundary user has access to create a directory in the root file system. This seems unlikely.

A better example would be to use `/var/lib/boundary`. 

This PR updates the `auth_storage_path` to reflect the best practice.

- [PKI worker configuration](https://boundary-8cdbn42ol-hashicorp.vercel.app/boundary/docs/configuration/worker/pki-worker)
- [Configure workers](https://boundary-8cdbn42ol-hashicorp.vercel.app/boundary/docs/install-boundary/configure-workers)

Jira: https://hashicorp.atlassian.net/browse/SPE-432

### DIFF
--- a/website/content/docs/configuration/worker/pki-worker.mdx
+++ b/website/content/docs/configuration/worker/pki-worker.mdx
@@ -15,7 +15,7 @@ Example (not safe for production!):
 
 ```hcl
 worker {
-  auth_storage_path="/boundary/demo-worker-1"
+  auth_storage_path="/var/lib/boundary"
   initial_upstreams = ["10.0.0.1"]
 }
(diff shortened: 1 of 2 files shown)


stellarsquall (approved):
I think this is great 👍 

after the docs are updated we can fix the deployment guide, too

sdoyle88:
LGTM!  Thanks!

PR #3761, opened 2023-09-21
Refactor counting and deleted ID listing

johanbrandhorst:


internal/credential/credential_libraries.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/hashicorp/boundary/internal/errors"
+)
+
+// CredentialLibraryRepository defines the interface expected
+// to get the total number of credential libraries.
+type CredentialLibraryRepository interface {
+	GetTotalCredentialLibraries(context.Context) (int, error)
+	GetTotalSSHCertificateCredentialLibraries(context.Context) (int, error)
+	ListDeletedCredentialLibraryIds(context.Context, time.Time, ...Option) ([]string, error)
+	ListDeletedSSHCertificateCredentialLibraryIds(context.Context, time.Time, ...Option) ([]string, error)
```
johanbrandhorst, line 20: Note that the option types here must _not_ be the subtype repository option (e.g. Vault) as it leads to an import cycle. We could use a separate `internal/credentials/options` package, but this felt OK to me.

internal/credential/credential_libraries.go
```diff
@@ -0,0 +1,57 @@
...
+	numGenericLibs, err := repo.GetTotalCredentialLibraries(ctx)
+	if err != nil {
+		return 0, nil
+	}
+	numSSHCertLibs, err := repo.GetTotalSSHCertificateCredentialLibraries(ctx)
+	if err != nil {
+		return 0, nil
+	}
+	return numGenericLibs + numSSHCertLibs, nil
+}
```
johanbrandhorst, line 34: This was straightforward

internal/credential/credential_libraries.go
```diff
@@ -0,0 +1,57 @@
...
+		deletedLibIds, err := repo.ListDeletedCredentialLibraryIds(ctx, since, WithReaderWriter(r, w))
+		if err != nil {
+			return err
+		}
+		deletedSSHCertLibIds, err := repo.ListDeletedSSHCertificateCredentialLibraryIds(ctx, since, WithReaderWriter(r, w))
+		if err != nil {
+			return err
+		}
+		deletedIds = append(deletedLibIds, deletedSSHCertLibIds...)
+		// TODO: Get transaction timestamp too
```
johanbrandhorst, line 50: In a future PR I will add the transaction timestamp to the db.Writer and it'll be available here and returned from this function.

internal/daemon/controller/handlers/credentiallibraries/credentiallibrary_service.go
```diff
@@ -138,20 +138,33 @@ func NewService(
...
+}
+
+func (clr *credentialLibraryRepo) ListDeletedIds(ctx context.Context, since time.Time) ([]string, error) {
+	return credential.ListDeletedCredentialLibraryIds(ctx, nil /* What do here? */, clr.repo, since)
+}
+
+// Temporary - will be replaced once generic function is refactored
+func (clr *credentialLibraryRepo) Now(ctx context.Context) (time.Time, error) {
+	return time.Now(), nil
+}
```
johanbrandhorst, line 166: This little wrapper type is only necessary for using the generic function. It could also just be functions passed directly. I think this little wrapper type is fine for the few places we'll need it, but welcome opinions.

internal/daemon/controller/handlers/credentiallibraries/credentiallibrary_service.go
```diff
@@ -138,20 +138,33 @@ func NewService(
...
+type credentialLibraryRepo struct {
+	repo *vault.Repository
+}
+
+func (clr *credentialLibraryRepo) GetTotalItems(ctx context.Context) (int, error) {
+	return credential.GetTotalCredentialLibraries(ctx, clr.repo)
+}
+
+func (clr *credentialLibraryRepo) ListDeletedIds(ctx context.Context, since time.Time) ([]string, error) {
+	return credential.ListDeletedCredentialLibraryIds(ctx, nil /* What do here? */, clr.repo, since)
```
johanbrandhorst, line 160: This function takes a db.Writer, which isn't something the handler generally has access to. How should we pass it in here?

internal/db/schema/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql
```diff
@@ -2,97 +2,176 @@
...
+  -- Target repository list logic is implemented
+  -- against views, so use a view for the deleted
+  -- ids too.
+  -- Note: does not end in _deleted so as
+  -- not to be attached a trigger in the below function.
+  create view target_all_subtypes_deleted_view
+  as
+    select * from target_tcp_deleted
+    union
+    select * from target_ssh_deleted;
```
johanbrandhorst, line 24: The target repo is different from the other repo in this way. Any objections to this?

internal/db/schema/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql
```diff
@@ -2,97 +2,176 @@
...
 
-  create table host_catalog_deleted (
+  -- Credentials
+  create table credential_static_json_credential_deleted (
     public_id wt_public_id primary key,
     delete_time wt_timestamp not null
   );
 
-  create table host_set_deleted (
+  create table credential_static_username_password_credential_deleted (
```
mgaffney, line 98: I get the following 2 warnings:
```
psql:/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql:207: NOTICE:  identifier "credential_static_username_password_credential_deleted_delete_time_idx" will be truncated to "credential_static_username_password_credential_deleted_delete_t"
psql:/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql:207: NOTICE:  identifier "credential_static_ssh_private_key_credential_deleted_delete_time_idx" will be truncated to "credential_static_ssh_private_key_credential_deleted_delete_tim"
```

Can we use index names for these two tables that are under the postgresql naming limit?

mikemountain (approved):

internal/db/schema/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql
```diff
@@ -2,97 +2,176 @@
 -- SPDX-License-Identifier: BUSL-1.1
 
 begin;
-  create table target_deleted (
+  create table target_tcp_deleted (
```
mgaffney, line 5: Can we please add a `comment on table` statement for all the tables?

internal/db/schema/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql
mgaffney, line 24:
```suggestion
    select public_id, delete_time from target_tcp_deleted
    union
    select public_id, delete_time from target_ssh_deleted;
```
Using a `select *` in a view can have side effects when an underlying table is updated.

internal/db/schema/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql
johanbrandhorst: I'll truncate this table name so we can keep using the same loop below for the index and trigger creation, thanks

internal/db/schema/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql
```diff
@@ -2,97 +2,176 @@
...
+    public_id wt_public_id primary key,
+    delete_time wt_timestamp not null
+  );
+
+  -- Target repository list logic is implemented
+  -- against views, so use a view for the deleted
+  -- ids too.
+  -- Note: does not end in _deleted so as
+  -- not to be attached a trigger in the below function.
+  create view target_all_subtypes_deleted_view
```
mgaffney, line 20: Can we please add a `comment on view` statement for the view?

internal/db/schema/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql
```diff
@@ -2,97 +2,176 @@
...
+  create table recording_session_deleted (
+    public_id wt_public_id primary key,
+    delete_time wt_timestamp not null
+  );
+
+  -- Workers
+  create table server_worker_deleted (
     public_id wt_public_id primary key,
     delete_time wt_timestamp not null
   );
```
mgaffney, line 177: The functions in this file also need `comment on` statements added

internal/db/schema/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql
johanbrandhorst: Ah haha that's not going to work 😂. OK I'll think of something here.

internal/credential/credential_libraries.go
```diff
@@ -0,0 +1,86 @@
...
+	"time"
+
+	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/util"
+)
+
+// CredentialLibraryRepository defines the interface expected
+// to get the total number of credential libraries and deleted ids.
+type CredentialLibraryRepository interface {
```
mgaffney, line 17:
```suggestion
type LibraryRepository interface {
```
Wouldn't `credential.LibraryRepository` be more like the existing interfaces in this package?

internal/credential/credential_stores.go
```diff
@@ -0,0 +1,87 @@
...
+	ListDeletedCredentialStoreIds(context.Context, time.Time, ...Option) ([]string, error)
+}
+
+// NewCredentialStoreService returns a new credential store service.
+func NewCredentialStoreService(ctx context.Context, writer db.Writer, repos ...CredentialStoreRepository) (*CredentialStoreService, error) {
+	const op = "credential.NewCredentialStoreService"
+	switch {
+	case util.IsNil(writer):
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing DB writer")
+	case len(repos) == 0:
```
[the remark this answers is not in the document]
johanbrandhorst: Yeah that makes sense, will do!

internal/credential/credential_stores.go
```diff
@@ -0,0 +1,87 @@
...
+	"time"
+
+	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/util"
+)
+
+// CredentialStoreRepository defines the interface expected
+// to get the total number of credential stores and deleted ids.
+type CredentialStoreRepository interface {
```
mgaffney, line 17: Same comment as above

mgaffney (changes requested):
These are comments from an initial review. I will add more later today.

internal/credential/credential_libraries.go
johanbrandhorst: Sure, that sounds good

internal/db/schema/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql
johanbrandhorst: I added alter index statements at the end, but that doesn't stop the NOTICEs from appearing. If you would prefer we don't have the notices appear, I'll have to change these tables to not be part of the loop, which I can do of course. Let me know what you prefer 😄.

internal/db/schema/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql
```diff
@@ -113,7 +292,11 @@ begin;
        and c.relname operator(pg_catalog.~) '^(.+_deleted)$' collate pg_catalog.default
        and pg_catalog.pg_table_is_visible(c.oid);
   $$ language sql;
+  comment on function get_deletion_tables is
+    'get_deletion_tables returns a set containing all the deleted table names by looking for '
+    'all tables that end in _deleted.';
 
+  -- Assign a "before delete" trigger on all deleted tables to run insert_deleted_id
```
mgaffney, line 299: We usually use the same name for the trigger and the trigger function. Any reason for not doing that here?

internal/db/schema/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql
```diff
@@ -2,100 +2,274 @@
...
+  -- Workers
+  create table server_worker_deleted (
+    public_id wt_public_id primary key,
+    delete_time wt_timestamp not null
+  );
+  comment on table server_worker_deleted is
+    'server_worker_deleted holds the ID and delete_time of every worker. '
+    'It is automatically trimmed of records older than 30 days by a job.';
 
   create or replace function insert_deleted_id() returns trigger
```
mgaffney, line 274:
```suggestion
  create function insert_deleted_id() returns trigger
```

internal/db/schema/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql
```diff
@@ -104,6 +278,11 @@ begin;
     return old;
   end;
   $$ language plpgsql;
+  comment on function insert_deleted_id is
+    'insert_deleted_id is a function that inserts a record into the table '
+    'specified by the first trigger argument. It takes the public ID from the row '
+    'affected by the trigger and the current timestamp. It is used to populate rows '
+    'of the deleted tables.';
 
   create function get_deletion_tables() returns setof name
```
mgaffney, line 287: Is this function needed outside of this migration? If not can we drop it at the end of the migration?

internal/db/schema/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql
```diff
@@ -113,7 +292,11 @@ begin;
        and c.relname operator(pg_catalog.~) '^(.+_deleted)$' collate pg_catalog.default
        and pg_catalog.pg_table_is_visible(c.oid);
   $$ language sql;
+  comment on function get_deletion_tables is
+    'get_deletion_tables returns a set containing all the deleted table names by looking for '
+    'all tables that end in _deleted.';
 
+  -- Assign a "before delete" trigger on all deleted tables to run insert_deleted_id
   do $$
```
mgaffney, line 300: Why is this a `before delete` trigger and not an `after delete` trigger?

internal/db/schema/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql
mgaffney: This is fine

mgaffney (changes requested):
I noticed many of the exported names in the `credential` package contain "Credential". Can we do something to eliminate these repetitive names?

internal/db/schema/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql
mikemountain: It gets used [here](https://github.com/hashicorp/boundary/blob/bb0536a0cbdc0c5f8afb2e03cf53f18f76d0e2c2/internal/pagination/purge/purge.go#L29) as well, where it goes through all the deletion tables and deletes records older than 30 days

mikemountain (approved):

internal/credential/credential_libraries.go
```diff
@@ -0,0 +1,86 @@
...
+// LibraryService coordinates calls to across different subtype repositories
+// to gather information about all credential libraries.
+type LibraryService struct {
+	repo   LibraryRepository
+	writer db.Writer
+}
+
+// GetTotalItems gets an estimate of the total number of credential libraries across all types
+func (s *LibraryService) GetTotalItems(ctx context.Context) (int, error) {
+	const op = "credential.(*CredentialLibraryRepository).GetTotalItems"
```
mgaffney, line 48: Can we fix the `op` string to match the method name and struct type?

internal/credential/credential_libraries.go
```diff
@@ -0,0 +1,86 @@
...
+	numSSHCertLibs, err := s.repo.GetTotalSSHCertificateLibraries(ctx)
+	if err != nil {
+		return 0, errors.Wrap(ctx, err, op)
+	}
+	return numGenericLibs + numSSHCertLibs, nil
+}
+
+// ListDeletedIds lists all deleted credential library IDs across all types
+func (s *LibraryService) ListDeletedIds(ctx context.Context, since time.Time) ([]string, error) {
+	const op = "credential.(*CredentialLibraryRepository).ListDeletedIds"
```
mgaffney, line 62: Same as above
> Can we fix the `op` string to match the method name and struct type?

internal/credential/credential_stores.go
```diff
@@ -0,0 +1,84 @@
...
+// StoreService coordinates calls to across different subtype repositories
+// to gather information about all credential stores.
+type StoreService struct {
+	repos  []StoreRepository
+	writer db.Writer
+}
+
+// GetTotalItems gets an estimate of the total number of credential stores across all types
+func (s *StoreService) GetTotalItems(ctx context.Context) (int, error) {
+	const op = "credential.(*CredentialStoreRepository).GetTotalItems"
```
mgaffney, line 48: > Can we fix the `op` string to match the method name and struct type?

internal/credential/credential_stores.go
```diff
@@ -0,0 +1,84 @@
...
+			return 0, errors.Wrap(ctx, err, op)
+		}
+		totalNumStores += numStores
+	}
+	return totalNumStores, nil
+}
+
+// ListDeletedIds lists all deleted credential store IDs across all types
+func (s *StoreService) ListDeletedIds(ctx context.Context, since time.Time) ([]string, error) {
+	const op = "credential.(*CredentialStoreRepository).ListDeletedIds"
```
mgaffney, line 62: Looks like all the `op` strings may need to be fixed

internal/credential/static/repository_credential.go
```diff
@@ -880,3 +882,87 @@ func (r *Repository) DeleteCredential(ctx context.Context, projectId, id string,
 
 	return rowsDeleted, nil
 }
+
+// GetTotalCredentials returns an estimate of the number of Vault credential stores
+func (r *Repository) GetTotalCredentials(ctx context.Context) (int, error) {
+	const op = "vault.(Repository).GetTotalCredentials"
```
mgaffney, line 888: `op` string?

internal/credential/static/repository_credential.go
```diff
@@ -880,3 +882,87 @@ func (r *Repository) DeleteCredential(ctx context.Context, projectId, id string,
 
 	return rowsDeleted, nil
 }
+
+// GetTotalCredentials returns an estimate of the number of Vault credential stores
```
mgaffney, line 886: There are multiple lines with `Vault` in them. Copy/paste error?

internal/db/schema/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql
mgaffney: > No particular reason outside of postgres suggesting `before` is more consistent, can switch it to an `after`

Where was this suggestion? Can you share a link?

mikemountain: [At the bottom in compatibility](https://www.postgresql.org/docs/current/sql-createtrigger.html) - I may have misunderstood this though, after reading it again

internal/credential/credential_libraries.go
```diff
@@ -0,0 +1,86 @@
...
+// LibraryRepository defines the interface expected
+// to get the total number of credential libraries and deleted ids.
+type LibraryRepository interface {
+	GetTotalLibraries(context.Context) (int, error)
+	GetTotalSSHCertificateLibraries(context.Context) (int, error)
+	ListDeletedLibraryIds(context.Context, time.Time, ...Option) ([]string, error)
+	ListDeletedSSHCertificateLibraryIds(context.Context, time.Time, ...Option) ([]string, error)
+}
+
+// NewLibraryService returns a new credential library service.
```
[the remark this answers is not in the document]
tmessi: This is also a departure from what has been done previously in that this is creating a struct for the service, while previously services where implemented as a function. I think it is fine to have a struct, at least if we really need multiple methods, but we might want to pick a name that is more descriptive of the particular service that is being implemented, rather than claiming the "LibraryService" name.

internal/credential/credential_libraries.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+// LibraryRepository defines the interface expected
+// to get the total number of credential libraries and deleted ids.
+type LibraryRepository interface {
+	GetTotalLibraries(context.Context) (int, error)
+	GetTotalSSHCertificateLibraries(context.Context) (int, error)
+	ListDeletedLibraryIds(context.Context, time.Time, ...Option) ([]string, error)
+	ListDeletedSSHCertificateLibraryIds(context.Context, time.Time, ...Option) ([]string, error)
+}
```
tmessi, line 22: This interface seems a bit odd. For one, it is not clear to me that this collection of methods constitutes a "repository". It might be the case that the vault.Repository implements this interface, but that doesn't mean this combination of methods *is* a repository.

Also, does `GetTotalLibraries` include the count of `GetTotalSSHCertificateLibraries`? The names seem to imply that it would, but then why have both?

And lastly, while we currently only have credential libraries from the vault package. In the future we might have additional ones. I would imagine that other credential libraries would have types that do not overlap with vault, so it seems odd that they would need to implement a "GetTotalSSHCertificateLibraries" method if only vault provides SSHCertificateLibraries. Perhaps it would help to create a service in the vault package that can aggregate all vault credential libraries. Then the service in the credential package can aggregate all of the libraries from those services (which would currently just be the vault one).

tmessi (changes requested):

internal/credential/credential_libraries.go
```diff
@@ -0,0 +1,86 @@
...
+
+// LibraryService coordinates calls to across different subtype repositories
+// to gather information about all credential libraries.
+type LibraryService struct {
+	repo   LibraryRepository
+	writer db.Writer
+}
+
+// GetTotalItems gets an estimate of the total number of credential libraries across all types
+func (s *LibraryService) GetTotalItems(ctx context.Context) (int, error) {
```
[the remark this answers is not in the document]
johanbrandhorst: This is fair feedback, I've kinda disliked the name from the start, but it'll need to be a change across all the repositories as it's used in the pagination "Repository" interface, so we should do it in a separate PR and rebase this on top of it.

internal/credential/credential_libraries.go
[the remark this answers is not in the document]
johanbrandhorst: This was originally named `CredentialLibraryService` and Mike suggested dropping the stuttering since it was `credential.CredentialLibraryService`. Now it's `credential.LibraryService`. Do you have another suggestion?

internal/credential/credential_libraries.go
johanbrandhorst: That's fair, I think we can more accurately name this `VaultLibraryRepository`, which at least helps make it clear that this repository is not necessarily expected to be implemented by other credential library repositories. If we add other credential library types in the future they would likely get their own interface definition. That way we wouldn't need to add a service wrapping all Vault library interactions.

The naming situation comes from following the naming conventions. In the Vault repository, the separation between "libraries" (sometimes called "generic libraries") and "ssh certificate libraries" is by file and by name. In implementing methods that operate on the "generic library" type, we're following the convention of just calling it "library", in the context of the Vault repository. Same thing with calling it "ssh certificate library" for SSH certificate libraries. I agree that this repository definition looks strange as a result, but the alternative I see is to rename the use of the "library" term to be "generic library" in the Vault repository. While `GetTotalLibraries` exists, it does not include the count for all library types in the Vault repository. I'd rather not make a rename from "Library" to "GenericLibrary" in the Vault repository in this PR, but I _do_ agree that such a change would be beneficial.

internal/credential/credential_libraries.go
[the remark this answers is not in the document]
johanbrandhorst: I've renamed the files these services are defined in, thanks

internal/credential/credential_libraries.go
johanbrandhorst: I've renamed this interface, do you think that is enough?

internal/credential/credential_libraries.go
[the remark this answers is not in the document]
johanbrandhorst: https://github.com/hashicorp/boundary/pull/3785

internal/db/schema/migrations/oss/postgres/78/01_deleted_tables_and_triggers.up.sql
johanbrandhorst: I changed this to `after delete`.

johanbrandhorst:
I've taken a look at the outstanding comments and resolved some, made some changed suggestions. Please take another look. Note that I am away next week so @mikemountain will be merging this once it is ready.

johanbrandhorst:

PR #3763, opened 2023-09-21
Filtering events: fix default deny filter behavior

sepehrfrgh:
This PR fixes the default behavior of the deny filter, if the user specifies any filter (allow/deny), that will override the default deny filter but if the user does not specify any filter, the default deny filter will apply for observation events.


jimlambrt (approved):

ajayreshc (approved):
LGTM

PR #3764, opened 2023-09-22
Add TransactionTimestamp, use for listing deleted IDs

johanbrandhorst:
## [db: add TransactionTimestamp](https://github.com/hashicorp/boundary/commit/17c52eda89a1be17b5d04a3f92d87eca59a5399a)

The TransactionTimestamp method returns the timestamp
of the current transaction.

## [all: source deleted ID timestamp from transaction](https://github.com/hashicorp/boundary/commit/ed8471ea127a2eebabad0d7a13368f4319a11021)

Removes the Now() function and replaces it with a timestamp
sourced from the same transaction within which the list
call is made. This should ensure we always get the correct
list of deleted IDs and simplifies the repositories.

### DIFF
--- a/internal/auth/ldap/repository_account.go
+++ b/internal/auth/ldap/repository_account.go
@@ -148,17 +148,19 @@ func (r *Repository) ListAccounts(ctx context.Context, withAuthMethodId string, 
 	return accts, nil
 }
 
-// ListDeletedIds lists the public IDs of any accounts deleted since the timestamp provided.
+// ListDeletedIds lists the public IDs of any accounts deleted since the timestamp provided,
+// and returns the timestamp of the transaction, to be used in other ListDeletedIds transactions.
+// This should ensure the correct list of deleted IDs is always returned.
 func (r *Repository) ListDeletedIds(ctx context.Context, since time.Time) ([]string, time.Time, error) {
 	const op = "account.(Repository).ListDeletedIds"
 	var deletedAccounts []*deletedAccount
-	var transactionTimestamp time.Time
+	var now time.Time
 	if _, err := r.writer.DoTx(ctx, db.StdRetryCnt, db.ExpBackoff{}, func(r db.Reader, w db.Writer) error {
 		if err := r.SearchWhere(ctx, &deletedAccounts, "delete_time >= ?", []any{since}); err != nil {
 			return errors.Wrap(ctx, err, op, errors.WithMsg("failed to query deleted accounts"))
 		}
 		var err error
-		transactionTimestamp, err = r.TransactionTimestamp(ctx)
+		now, err = r.Now(ctx)
 		if err != nil {
 			return errors.Wrap(ctx, err, op, errors.WithMsg("failed to query transaction timestamp"))
 		}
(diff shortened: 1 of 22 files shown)


internal/pagination/pagination.go
```diff
@@ -154,7 +147,7 @@ func PaginateRequest[T any, PbT ResponseItem](
 	// something to populate LastItemId and LastItemUpdatedTime with.
 	if len(resp.Items) > 0 || refreshToken != nil {
 		newRefreshToken := &pbs.ListRefreshToken{
-			CreatedTime:         timestamppb.New(newRefreshTokenCreateTime),
+			CreatedTime:         timestamppb.New(newTokenCreateTime),
```
johanbrandhorst, line 150: This is where the logic happens for listing deleted IDs without overlapping timestamps. The first request uses the current time from the system for the timestamp. Subsequent requests uses the time from the refresh token, getting the new timestamp from the DB transaction.

mikemountain (approved):

internal/db/read_writer_test.go
```diff
@@ -3193,3 +3193,16 @@ func TestDb_oplogMsgsForItems(t *testing.T) {
 // 		})
 // 	}
 // }
+
+func TestDb_TransactionTimestamp(t *testing.T) {
```
mgaffney, line 3197: Can we add tests for calling `TransactionTimestamp` multiple times within the same transaction?

mgaffney (changes requested):

internal/db/read_writer_test.go
mgaffney: 1. Multiple calls to `TransactionTimestamp` within the same transaction should return equal `time.Time` values (might want to sleep for a second between calls).
2. Multiple calls to `TransactionTimestamp` not in a transaction should return values of `time.Time` that are not equal (again, might want to sleep for a second between calls).

internal/db/read_writer.go
```diff
@@ -525,6 +528,22 @@ func (rw *Db) SearchWhere(ctx context.Context, resources any, where string, args
 	return nil
 }
 
+// TransactionTimestamp returns the current transaction timestamp.
+func (rw *Db) TransactionTimestamp(ctx context.Context) (time.Time, error) {
```
mgaffney, line 532:
```suggestion
func (rw *Db) Now(ctx context.Context) (time.Time, error) {
```
I think `Now` seems more intuitive.

internal/auth/ldap/repository_account.go
```diff
@@ -149,17 +149,28 @@ func (r *Repository) ListAccounts(ctx context.Context, withAuthMethodId string,
 }
 
 // ListDeletedIds lists the public IDs of any accounts deleted since the timestamp provided.
-func (r *Repository) ListDeletedIds(ctx context.Context, since time.Time) ([]string, error) {
+func (r *Repository) ListDeletedIds(ctx context.Context, since time.Time) ([]string, time.Time, error) {
```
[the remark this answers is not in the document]
johanbrandhorst: Renaming of this method will happen in https://github.com/hashicorp/boundary/pull/3777, which applies the patterns in https://github.com/hashicorp/boundary/pull/3761. I agree that the struct is a nice idea!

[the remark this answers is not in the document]
johanbrandhorst: Unfortunately, the use of a struct for the return type here will complicate the use of this method across many services as it's not clear where we would define the type such that it could be used in the `pagination` package interface using this method (through an interface). I'm not sure the added complexity needed to allow the struct is worth it. Given the conversation today, we should only need one timestamp return - the transaction timestamp minus some duration based on the "default request timeout". I'll leave this PR up for that buffer to be added.

johanbrandhorst:

PR #3766, opened 2023-09-22
Ignore non-public tagged struct fields from telemetry

ajayreshc:
Changes: 
* `telemetryFilter()` requires both `eventstream:"observation"` and `class:"public"` explicitly specified on fields.
* Added test case. 
* Updated existing telemetry tests.


jimlambrt (approved):
ty!

PR #3768, opened 2023-09-22
Remove expired auth tokens from cache

talanknight:


internal/daemon/cache/repository_token.go
```diff
@@ -318,6 +319,12 @@ func cleanOrphanedAuthTokens(ctx context.Context, writer db.Writer, idToKeyringl
 	delete from auth_token
 	where
 		last_accessed_time < @last_accessed_time
+	or
+		-- sqlite stores expiration_time as a string in a format that might not match
+		-- what is being used by current_timestamp.  Using datetime() makes the
+		-- formats match which allow the string comparison perfoemed here to work
```
jimlambrt, line 325:
```suggestion
		-- formats match which allow the string comparison performed here to work
```

jimlambrt (approved):
One minor suggestion.  Looks good!

PR #3769, opened 2023-09-22
Collect resource info in repository_<resource>s.go

talanknight:
This PR moves the session and target specific info into their respective repository_ files.  Adding a new resource should then be a matter of creating a new respository_newresource.go file and test that match sessions and targets, and then adding a single method call to refresh a struct to the search handler.

### DIFF
--- a/internal/cmd/commands/daemon/search_handler.go
+++ b/internal/cmd/commands/daemon/search_handler.go
@@ -12,7 +12,6 @@ import (
 	"github.com/hashicorp/boundary/api/sessions"
 	"github.com/hashicorp/boundary/api/targets"
 	"github.com/hashicorp/boundary/internal/daemon/cache"
-	"github.com/hashicorp/boundary/internal/daemon/controller/handlers"
 	"github.com/hashicorp/boundary/internal/errors"
 	"github.com/hashicorp/boundary/internal/util"
 )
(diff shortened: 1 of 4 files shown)


internal/daemon/cache/repository_token.go
```diff
@@ -318,6 +319,12 @@ func cleanOrphanedAuthTokens(ctx context.Context, writer db.Writer, idToKeyringl
 	delete from auth_token
 	where
 		last_accessed_time < @last_accessed_time
+	or
+		-- sqlite stores expiration_time as a string in a format that might not match
+		-- what is being used by current_timestamp.  Using datetime() makes the
+		-- formats match which allow the string comparison perfoemed here to work
```
johanbrandhorst, line 325:
```suggestion
		-- formats match which allow the string comparison performed here to work
```

internal/cmd/commands/daemon/search_handler.go
```diff
@@ -99,50 +112,41 @@ func newSearchTargetsHandlerFunc(ctx context.Context, repo *cache.Repository) (h
...
+	list func(context.Context, string) ([]T, error)
+	// query takes a context, an auth token, and a query string and returns all
+	// resources for that auth token that matches the provided query parameter
+	query        func(context.Context, string, string) ([]T, error)
+	searchResult func([]T) *SearchResult
 }
 
-func searchSessions(ctx context.Context, repo *cache.Repository, authTokenId, query string, filter *handlers.Filter) (*SearchResult, error) {
-	var found []*sessions.Session
+func (l *searchFns[T]) search(ctx context.Context, authTokenId, query string, filter *handlers.Filter) (*SearchResult, error) {
```
jimlambrt, line 131: perhaps a comment that this implements the `resourceSearcher` interface and what
it does?

internal/cmd/commands/daemon/search_handler.go
```diff
@@ -38,6 +38,24 @@ func newSearchTargetsHandlerFunc(ctx context.Context, repo *cache.Repository) (h
 	case util.IsNil(repo):
 		return nil, errors.New(ctx, errors.InvalidParameter, op, "repository is missing")
 	}
+
```
jimlambrt, line 41: Any thoughts on implementing unit tests for the bits in this file?

internal/daemon/cache/repository_refresh_test.go
```diff
@@ -29,6 +25,14 @@ func noopRetrievalFn[T any](context.Context, string, string) ([]T, error) {
 	return nil, nil
 }
 
+// staticRetrievalFn returns a function that satisfies the With*RetrievalFn
+// and returns the provided slice and a nil error always
+func staticRetrievalFn[T any](ret []T) func(context.Context, string, string) ([]T, error) {
```
jimlambrt, line 30: static retrieval of what?  while this is clever, I'm left scratching my head.
perhaps a better name might help?  Hum... `testStaticResourceRetrievalFunc`???

internal/daemon/cache/repository_refresh_test.go
```diff
@@ -29,6 +25,14 @@ func noopRetrievalFn[T any](context.Context, string, string) ([]T, error) {
 	return nil, nil
 }
 
+// staticRetrievalFn returns a function that satisfies the With*RetrievalFn
```
jimlambrt, line 28: It's not clear what the With*RetrievalFn might be since it's not searchable term
in the code base.  Perhaps: returns a function that satisifies a "retrieval
function" like TargetRetrievalFunc or SessionRetrievalFunc.

jimlambrt (changes requested):
Some suggestions on readablitity and perhaps some additional unit tests.  ty.

internal/daemon/cache/search.go
```diff
@@ -0,0 +1,196 @@
...
+
+// SearchService is a domain service that can search across all resources in the
+// cache.
+type SearchService struct {
+	searchableResources map[SearchableResource]resourceSearcher
+}
+
+// queryAndLister defines the methods needed for listing an querying the
+// resources supported by the boundary frontend.
+type queryAndLister interface {
```
jimlambrt, line 70: If it's used in an exported func, can we export this as well?

jimlambrt (approved):
IMO, this refactor is a huge win. Ty for taking it on.  I had just a minor suggestion.

PR #3772, opened 2023-09-25
Bump UI Commit

dvonanderson:
Update the UI commit to include: 
- post connection workflow feature
- LDAP changes
-  assume role changes


internal/ui/VERSION_ent
```diff
@@ -0,0 +1,4 @@
+cf5332d41f9ccf116c9f3132daa398c635605712
```
ZedLi, line 1: This shouldn't be included

dvonanderson: ok will remove

PR #3775, opened 2023-09-25
Update .copywrite.hcl

mdeggies:
Ignores the directories licensed under MPL


johanbrandhorst (approved):
Does this mean we can rewrite the copywrite target in the Makefile? I think we're just invoking it multiple times.

mdeggies:
Not yet! We are planning some core updates to the copyright tool in Q4 that will remove the need for your workaround (cd'ing into the different subdirs and running copyright). For now, though, it doesn't yet support overwriting headers or supporting multiple license files.

PR #3776, opened 2023-09-25
test: Update tests to account for deprecated features in 0.14.0

moduli:
This PR updates tests to account for deprecated features in `0.14.0`.
- Tests must use `boundary credential-stores vault-generic` as opposed to `boundary credential-stores vault`
- On CLI errors, tests must use the `status_code` field instead of the `status` field (https://github.com/hashicorp/boundary/commit/e66ea15fef8e0e84282d0292012e90581839a5a8)


stasryzhov (approved):

irenarindos (approved):

internal/daemon/controller/handlers/credentiallibraries/credentiallibrary_service_test.go
```diff
@@ -809,42 +809,6 @@ func TestCreate(t *testing.T) {
 				},
 			},
 		},
-		{
-			name: "Create a valid vault CredentialLibrary with no type set subtype",
```
moduli, line 813: Removed this test since it seems like the current CLI behavior is to force the user to provide a subtype?

tmessi (commented):
This looks good to me. I think we should add something to the CHANGELOG to note this change in behavior. The original changelog that noted the deprecation can be seen [here](https://github.com/hashicorp/boundary/blob/060e5856b9742b784b8d181e26657db12548d378/CHANGELOG.md?plain=1#L216-L231).

CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 
 ## Next
 
+### Deprecations/Changes
+* Per the note in Boundary 0.14.0, the `vault` credential library subtype has now been
```
tmessi, line 8: I think this should be `0.12.0`

moduli: 😅

tmessi (approved):

PR #3785, opened 2023-09-29
all: rename GetTotalItems to EstimatedCount

johanbrandhorst:
This now more closely aligns with the role of this function, which is to get an estimated count of the number of items of a specific resource in a repository.


johanbrandhorst:
Note that I am away next week so @mikemountain will merge this if accepted.

johanbrandhorst:

PR #3791, opened 2023-10-03
Expose Valid Principals for Vault SSH Signed Certs

kheina:
This PR allows users to add additional valid principals when creating a vault ssh signed cert credential library.

by default, only the username is passed to vault to sign as a valid principal, but others may be needed in certain use cases. this pr adds `additional_valid_principals` api option and `-additional-valid-principal` cli option to add principals to the `"valid_principals"` vault api request in addition to the required `username` field which is already passed under `"valid_principals"`.


internal/credential/vault/private_library.go
```diff
@@ -905,8 +911,13 @@ func (lib *sshCertIssuingCredentialLibrary) retrieveCredential(ctx context.Conte
 		}
 	}
 
+	validPrincipals := username
+	if len(lib.AdditionalValidPrincipals) > 0 {
```
kheina, line 915: this should be a check against `""`

carry over from when it was an array

internal/credential/vault/private_library_test.go
```diff
@@ -1370,21 +1381,34 @@ func TestRepository_sshCertIssuingCredentialLibrary_retrieveCredential(t *testin
...
+			require.NoError(err)
+			cert, ok := pub.(*ssh.Certificate)
+			require.True(ok)
+
+			// ensure both keys are the same type
+			require.Equal(priv.PublicKey().Type(), cert.Key.Type())
+
+			if vp, ok := tt.expected["valid_principals"]; ok {
+				require.Equal(vp, cert.ValidPrincipals)
+			}
```
kheina, line 1411: it actually parses the cert to confirm the valid principals were added! 🎉

tmessi (changes requested):
I think this is looking good. Can you add some test cases to the [cli tests](https://github.com/hashicorp/boundary/blob/main/internal/tests/cli/boundary/credential_libraries.bats) for setting the new field?

louisruch (commented):
LGTM, agree that we should try add some CLI tests and Tim called out

tmessi (approved):

jesus-christ666:
Ok

PR #3794, opened 2023-10-03
Move search logic into the domain

talanknight:


internal/daemon/cache/search.go
```diff
@@ -0,0 +1,155 @@
...
+	"fmt"
+
+	"github.com/hashicorp/boundary/api/sessions"
+	"github.com/hashicorp/boundary/api/targets"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/util"
+	"github.com/hashicorp/go-bexpr"
+)
+
+// SearchParams contains the paramaters for searching in the cache
```
jimlambrt, line 14:
```suggestion
// SearchParams contains the parameters for searching in the cache
```

internal/daemon/cache/search.go
```diff
@@ -0,0 +1,155 @@
...
+	// returned SearchResult will have it's "Targets" field populated so the
+	// searchResult should take the passed in paramater and assign it to the
+	// appropriate field in the SearchResult.
+	searchResult func([]T) *SearchResult
+}
+
+// searcher is an interface that only resourceSearchFns[T] is expected to satisfy.
+// Specifying this interface allows the code to have a map with searchFns values
+// which have different bound generic types.
+type searcher interface {
```
jimlambrt, line 111:
```suggestion
type resourceSearcher interface {
```

internal/daemon/cache/search.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/hashicorp/boundary/api/targets"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/util"
+	"github.com/hashicorp/go-bexpr"
+)
+
+// SearchParams contains the paramaters for searching in the cache
+type SearchParams struct {
+	// the name of the resource. eg. "targets" or "sessions"
+	Resource string
```
jimlambrt, line 17: if you defined a `type SearchableResource string` and defined the valid enums
for it, you coiuld have type safety here and in the `searchableResources` map

internal/daemon/cache/search.go
```diff
@@ -0,0 +1,155 @@
...
+// SearchResult returns the results from searching the cache.
+type SearchResult struct {
+	Targets  []*targets.Target
+	Sessions []*sessions.Session
+}
+
+// SearchService is a domain service that can search across all resources in the
+// cache.
+type SearchService struct {
+	repo                *Repository
```
jimlambrt, line 35: is this used anywhere besides `NewSearchService`?

internal/daemon/cache/search.go
```diff
@@ -0,0 +1,155 @@
...
+	searchableResources map[string]searcher
+}
+
+func NewSearchService(ctx context.Context, repo *Repository) (*SearchService, error) {
```
jimlambrt, line 39: could repo be an interface with a defined set of receiver funcs?  I'm not sure
it needs to depend on the entire concrete type.   I didn't look to see what's
not used from the concrete type.

jimlambrt (changes requested):
overall, I think this is much better.

PR #3801, opened 2023-10-05
Daemon and search restricted to supported platforms

talanknight:
I created an "unsupported" command which is assigned by default to the "daemon" and "search" commands. This is overwritten by the platforms that do support these commands.


jimlambrt (approved):
just some spelling bits.  looks great.,

PR #3808, opened 2023-10-06
fix(cli): Remove websocket message size limit

tmessi:
When the boundary client establishes a connection to the websocket used
to communicate with a worker, it was configured with a max websocket
message size of 32KiB. This was the default behavior for v1.8.7 of
nhooyr/websocket. However, there are cases where the worker can send
messages to the client that are larger than 32KiB. When this happens the
client returns an error and terminates the connection.

Notably this seems to only currently impact sessions made to SSH
Targets, although it could theoretically impact any target type
depending on the details of how the worker performs the writing of the
messages. The issue can be reproduced by attempting to scp a large
enough file from an SSH Target to the client. In this case the
connection gets terminated and results in a partial transfer of the
file.

This message size limit appears to be unnecessary, and the default
behavior is changing in v1.8.8 of nhooyr/websocket to make the message
size unlimited. Since this version is not yet released, this commit
switches the dependency to a fork with a version that is v1.8.7 plus the
patch to disable the limit. Once v1.8.8 is release this dependency can
be set back.

See:
    https://github.com/nhooyr/websocket/pull/254
    https://github.com/nhooyr/websocket/pull/256

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,10 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
   submitted ([PR](https://github.com/hashicorp/boundary/pull/3800))
 * worker: Fix an issue that could cause intermittent startup issues on slow
   systems ([PR](https://github.com/hashicorp/boundary/pull/3803))
+* cli: Remove websocket max message size. This fixes issues where large message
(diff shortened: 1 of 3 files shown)


louisruch (approved):
LGTM

PR #3810, opened 2023-10-06
test(server): Fix flakiness in TestServer_ReloadWorkerTags

moduli:
This PR attempts to address flakiness in `TestServer_ReloadWorkerTags` by replacing hard-coded sleeps with some smarter polling similarly used in `TestServer_ReloadInitialUpstreams`. This test would occasionally fail for the following reason...
```
--- FAIL: TestServer_ReloadWorkerTags (121.34s)
                                               
    worker_tags_reload_test.go:126: 
        	Error Trace:	/home/runner/actions-runner/_work/boundary-enterprise/boundary-enterprise/internal/cmd/commands/server/worker_tags_reload_test.go:108
        	            				/home/runner/actions-runner/_work/boundary-enterprise/boundary-enterprise/internal/cmd/commands/server/worker_tags_reload_test.go:126
        	Error:      	Should be true
        	Test:       	TestServer_ReloadWorkerTags
```

The test is expecting the controller to get updated worker tags, but sometimes, it's not getting the expected values. The initial theory is that the hard-coded sleeps didn't wait long enough for the changes to propagate to the controller.

### DIFF
--- a/internal/cmd/commands/server/worker_tags_reload_test.go
+++ b/internal/cmd/commands/server/worker_tags_reload_test.go
@@ -80,6 +80,7 @@ func TestServer_ReloadWorkerTags(t *testing.T) {
 		controller.WithWorkerAuthKms(workerAuthWrapper),
 		controller.WithRootKms(rootWrapper),
 		controller.WithRecoveryKms(recoveryWrapper),
+		controller.WithEnableEventing(),
 	)
 	t.Cleanup(testController.Shutdown)
 
@@ -154,6 +155,7 @@ pollController:
 	// Set new tags on worker
 	t.Log("Restart worker with new tags...")
 	cmd.presetConfig.Store(fmt.Sprintf(workerBaseConfig+tag2Config, key, testController.ClusterAddrs()[0]))
+	t.Logf("%s", workerBaseConfig+tag2Config)
 	cmd.SighupCh <- struct{}{}
 	select {
 	case <-cmd.reloadedCh:
--- a/internal/cmd/commands/server/worker_tags_reload_test.go
+++ b/internal/cmd/commands/server/worker_tags_reload_test.go
@@ -176,13 +176,11 @@ pollControllerAgain:
 			w, err = serversRepo.LookupWorkerByName(testController.Context(), "test")
 			require.NoError(err)
 			if w != nil {
+				t.Logf("%s, %+v", w.GetLastStatusTime().AsTime().Round(time.Second), w.GetConfigTags())
 				switch {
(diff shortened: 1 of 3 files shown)


internal/cmd/commands/server/worker_tags_reload_test.go
```diff
@@ -104,29 +110,88 @@ func TestServer_ReloadWorkerTags(t *testing.T) {
 		w, err := serversRepo.LookupWorkerByName(testController.Context(), name)
 		require.NoError(err)
 		require.NotNil(w)
-		v, ok := w.CanonicalTags()[key]
-		require.True(ok)
+		tags := w.CanonicalTags()
+		v, ok := tags[key]
+		require.True(ok, fmt.Sprintf("EXPECTED: map[%s:%s], ACTUAL: %+v", key, values, tags))
```
moduli, line 115: Updated logging to get more information

internal/cmd/commands/server/worker_tags_reload_test.go
```diff
@@ -104,29 +110,88 @@ func TestServer_ReloadWorkerTags(t *testing.T) {
...
+				default:
+					if !lastStatusTime.Equal(w.GetLastStatusTime().AsTime().Round(time.Second)) {
+						timeout.Stop()
+						break pollController
+					}
+				}
+			}
+			poll.Reset(time.Second)
+		}
+	}
```
moduli, line 149: This is from `TestServer_ReloadInitialUpstreams`

internal/cmd/commands/server/worker_tags_reload_test.go
```diff
@@ -104,29 +110,88 @@ func TestServer_ReloadWorkerTags(t *testing.T) {
...
 	case <-time.After(15 * time.Second):
-		t.Fatalf("timeout")
+		t.Fatalf("timeout waiting for worker start")
+	}
+
+	// Give time to populate up to the controller
+	timeout.Reset(15 * time.Second)
+	poll.Reset(time.Second)
+	lastStatusTime = time.Time{}
+	startTime := time.Now()
```
moduli, line 166: This is a slightly modified version of the loop. `TestServer_ReloadInitialUpstreams` is different in that it uses a second controller and can set `lastStatusTime` to zero. In this case, the idea is to wait for 2 status updates starting from this point in time.

moduli:
Just got a failure when testing this locally... back to the drawing board...

moduli:
Looks like there might be actually something in the boundary application causing this. More investigation is needed. Closing this PR.

PR #3817, opened 2023-10-10
prototype: refactor targets into domain layer

johanbrandhorst:
This PR contains a prototype of a new design for the pagination logic for the target endpoint. It moves the bulk of the logic into a service level function in the target domain package, and creates a new domain package for handling refresh tokens.

As a next step, we will prototype using these patterns for the accounts endpoint, which should both help expose problems with the design and opportunities for unifying logic.

The purpose of this PR is to get alignment on the design - it doesn't need to be merged in its current state and anything is up for discussion.


internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -234,87 +234,115 @@ func (s Service) ListTargets(ctx context.Context, req *pbs.ListTargetsRequest) (
...
-	filter, err := handlers.NewFilter(ctx, req.GetFilter())
+
+	repo, err := s.repoFn(target.WithPermissions(userPerms))
 	if err != nil {
 		return nil, err
 	}
-	listTargetsFn := func(prevPageLast target.Target, refreshToken *pbs.ListRefreshToken, limit int) ([]target.Target, error) {
-		opts := []target.Option{
-			target.WithLimit(limit),
+	filterItemFn := func(item target.Target) (bool, error) {
```
johanbrandhorst, line 252: This function is used in the domain to allow filtering of items based on permissions and the user supplied filter. The alternative we considered is to write some logic for converting the application level filtering to a domain level filter (that would work across all domains), but this would certainly mean we would miss the November release. We could come back to refactor this at a later point if we miss the November release.

internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -234,87 +234,115 @@ func (s Service) ListTargets(ctx context.Context, req *pbs.ListTargetsRequest) (
...
+			PermissionsHash:     rt.PermissionsHash,
+			LastItemId:          rt.LastItemId,
+			LastItemUpdatedTime: rt.LastItemUpdatedTime.AsTime(),
+		}
+		err = refreshtoken.ValidateRefreshToken(
+			ctx,
+			domainRefreshToken,
+			refreshtoken.ResourceTypeTarget,
+			&authResults,
+		)
```
johanbrandhorst, line 299: The refresh token is validated in the application layer as it is part of request validation, but this could be moved to the domain if preferred.

internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -234,87 +234,115 @@ func (s Service) ListTargets(ctx context.Context, req *pbs.ListTargetsRequest) (
...
+		// We're doing the conversion from the protobuf types to the
+		// domain types here rather than in the domain so that the domain
+		// doesn't need to know about the protobuf types.
+		domainRefreshToken := &refreshtoken.RefreshToken{
+			CreatedTime:         rt.CreatedTime.AsTime(),
+			ResourceType:        domainResourceType,
+			PermissionsHash:     rt.PermissionsHash,
+			LastItemId:          rt.LastItemId,
+			LastItemUpdatedTime: rt.LastItemUpdatedTime.AsTime(),
+		}
```
johanbrandhorst, line 293: We opted not to have a `NewRefreshToken` method that performs validation as we sometimes construct refresh tokens from trusted information.

internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -234,87 +234,115 @@ func (s Service) ListTargets(ctx context.Context, req *pbs.ListTargetsRequest) (
...
-
-		filterable, err := subtypes.Filterable(pbItem)
+		var domainResourceType refreshtoken.ResourceType
+		switch rt.ResourceType {
+		case pbs.ResourceType_RESOURCE_TYPE_TARGET:
+			domainResourceType = refreshtoken.ResourceTypeTarget
+		// TODO: Add more when generalizing this
+		default:
+			domainResourceType = refreshtoken.ResourceTypeUnknown
+		}
```
johanbrandhorst, line 283: This function would likely be refactored out into a separate package to unify logic across all list handlers in a future PR.

internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -1868,6 +1896,51 @@ func validateListRequest(ctx context.Context, req *pbs.ListTargetsRequest) error
...
+	const op = "list.marshalRefreshToken"
+	if token == nil {
+		return "", errors.New(ctx, errors.InvalidParameter, op, "token is required")
+	}
+	marshaled, err := proto.Marshal(token)
+	if err != nil {
+		return "", errors.Wrap(ctx, err, op)
+	}
+	return base58.Encode(marshaled), nil
+}
```
johanbrandhorst, line 1942: These functions will likely be moved to a separate package to unify logic across list handlers

internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -234,87 +234,115 @@ func (s Service) ListTargets(ctx context.Context, req *pbs.ListTargetsRequest) (
...
+	var listResp *target.ListResponse
+	if req.GetRefreshToken() == "" {
+		listResp, err = target.List(ctx, repo, &authResults, pageSize, filterItemFn)
+		if err != nil {
+			return nil, err
 		}
-
-		pbItem, err := toProto(ctx, item, outputOpts...)
+	} else {
+		rt, err := parseRefreshToken(ctx, req.GetRefreshToken())
```
johanbrandhorst, line 272: The parsing of the refresh token from a string into a API level protobuf type remains the responsibility of the application layer.

internal/db/read_writer.go
```diff
@@ -70,6 +70,9 @@ type Reader interface {
 
 	// ScanRows will scan sql rows into the interface provided
 	ScanRows(ctx context.Context, rows *sql.Rows, result any) error
+
+	// Now returns the current transaction timestamp.
+	Now(ctx context.Context) (time.Time, error)
```
johanbrandhorst, line 75: This logic is imported from #3764 to support setting the timestamp based on the timestamp of the transaction used to list deleted IDs.

internal/refreshtoken/refreshtoken.go
```diff
@@ -0,0 +1,73 @@
...
+type RefreshToken struct {
+	CreatedTime         time.Time
+	ResourceType        ResourceType
+	PermissionsHash     []byte
+	LastItemId          string
+	LastItemUpdatedTime time.Time
+}
+
+// ValidateRefreshToken validates a refresh token.
+func ValidateRefreshToken(
```
johanbrandhorst, line 38: This function is only used in API handlers to validate incoming refresh tokens. Domains can create their own refresh tokens from trusted data using the struct directly.

internal/target/service_list.go
```diff
@@ -0,0 +1,99 @@
...
+			CreatedTime:         time.Now(),
+			ResourceType:        refreshtoken.ResourceTypeTarget,
+			PermissionsHash:     grantsHash,
+			LastItemId:          targets[len(targets)-1].GetPublicId(),
+			LastItemUpdatedTime: targets[len(targets)-1].GetUpdateTime().AsTime(),
+		}
+	}
+
+	return resp, nil
+}
```
johanbrandhorst, line 99: Much of the logic in this function may be possible to delegate to a generic function, but we'll consider that refactor in a future PR after implementing pagination for another domain.

internal/target/service_list_more.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/refreshtoken"
+	rt "github.com/hashicorp/boundary/internal/refreshtoken"
+)
+
+func ListMore(
```
johanbrandhorst, line 15: This name isn't final, but we'd want something to illustrate that this is different from the default list because it takes a refresh token, and as such needs to get the deleted IDs.

internal/target/service_list_more.go
```diff
@@ -0,0 +1,111 @@
...
+		},
+	}
+
+	if len(targets) > 0 {
+		resp.RefreshToken.LastItemId = targets[len(targets)-1].GetPublicId()
+		resp.RefreshToken.LastItemUpdatedTime = targets[len(targets)-1].GetUpdateTime().AsTime()
+	}
+
+	return resp, nil
+}
```
johanbrandhorst, line 111: We might unify some of this logic with List in a future PR.

internal/target/service_list_response.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+// will likely eventually become a generic response and moved from target
+type ListResponse struct {
+	Items               []Target
+	DeletedIds          []string
+	EstimatedTotalItems int
+	CompleteListing     bool
+	RefreshToken        *refreshtoken.RefreshToken
+}
```
johanbrandhorst, line 17: Should this be in its own file?

internal/refreshtoken/refreshtoken.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// A RefreshToken is returned in list endpoints for the purposes of pagination
+type RefreshToken struct {
+	CreatedTime         time.Time
+	ResourceType        ResourceType
+	PermissionsHash     []byte
+	LastItemId          string
+	LastItemUpdatedTime time.Time
+}
```
tmessi, line 35: I think it might make sense to have a constructor function for this, to ensure it is valid and to help with constructing tokens, ie:

```go
func New(res boundary.Resource, permissionsHash []byte) *RefreshToken {
  return &RefreshToken{
    CreatedTime: time.Now(),
    ResourceType: res.GetType(), // Note this would need to be created, but would probably be useful to be able to get a resource.Type from a boundary.Resource
    PermissionsHash: permissionsHash,
    LastItemId: res.GetPublicId(),
    LastItemUpdatedTime: res.GetUpdateTime(),
  }
}
```

internal/target/repository.go
```diff
@@ -351,23 +351,33 @@ func (r *Repository) listPermissionWhereClauses() ([]string, []any) {
 	return where, args
 }
 
-// ListDeletedIds lists the public IDs of any targets deleted since the timestamp provided.
-func (r *Repository) ListDeletedIds(ctx context.Context, since time.Time) ([]string, error) {
+// ListDeletedIds lists the public IDs of any targets deleted since the timestamp provided,
+// and the timestamp of the transcation within which the targets were listed.
+func (r *Repository) ListDeletedIds(ctx context.Context, since time.Time) ([]string, time.Time, error) {
```
tmessi, line 356: I think this can be un-exported

internal/target/repository.go
```diff
@@ -351,23 +351,33 @@ func (r *Repository) listPermissionWhereClauses() ([]string, []any) {
...
 	}
-	return targetIds, nil
+	return targetIds, transactionTimestamp, nil
 }
 
-// GetTotalItems returns an estimate of the total number of items across all targets.
-func (r *Repository) GetTotalItems(ctx context.Context) (int, error) {
-	const op = "target.(Repository).GetTotalItems"
+// EstimatedCount returns an estimate of the total number of items across all targets.
+func (r *Repository) EstimatedCount(ctx context.Context) (int, error) {
```
tmessi, line 379: I think this can be un-exported.

internal/target/service_list_more.go
```diff
@@ -0,0 +1,111 @@
...
+		Items:               targets,
+		DeletedIds:          deletedIds,
+		EstimatedTotalItems: totalItems,
+		CompleteListing:     completeListing,
+		RefreshToken: &refreshtoken.RefreshToken{
+			// Use the timestamp of the deleted IDs transaction with a
+			// buffer to account for overlapping transactions. It is okay
+			// to return a deleted ID more than once. The buffer corresponds
+			// to Postgres' default transaction timeout.
+			CreatedTime:         transactionTimestamp.Add(-30 * time.Second),
```
tmessi, line 97: This advances the CreateTime of the refresh token, which means that, as long as refresh requests keep happening, the token will never expire. Is this the intended behavior? I thought refresh tokens are meant to eventual expire to force clients to make a fresh request.

For example, the only way [this condition would be true](https://github.com/hashicorp/boundary/pull/3817/files#diff-1811705c88d7ca2ba4ba4ac4bde5dce13d821453ce236c4bb834d3e11bfc1e85R59) is if a client takes more than 30 days to make a follow-up request.

internal/target/service_list_more.go
```diff
@@ -0,0 +1,111 @@
...
+			PermissionsHash:     grantsHash,
+			LastItemId:          tok.LastItemId,
+			LastItemUpdatedTime: tok.LastItemUpdatedTime,
+		},
+	}
+
+	if len(targets) > 0 {
+		resp.RefreshToken.LastItemId = targets[len(targets)-1].GetPublicId()
+		resp.RefreshToken.LastItemUpdatedTime = targets[len(targets)-1].GetUpdateTime().AsTime()
+	}
```
tmessi, line 108: I think this could be moved into a "Refresh" method on a RefreshToken. I think the only thing that should change between the refresh token that was passed in is the LastItemId and LastItemUpdatedTime. So similarly to the `New` function that was mentioned, you could have something like:

```go
func (r *RefreshToken) Refresh(res boundary.Resource) error {
  // validation like that resource type matches,
  // and that update time is newer than current update time

  r.LastItemId = res.GetPublicId()
  r.LastItemUpdatedTime = res.GetUpdateTime()
}
```

tmessi (changes requested):
Nice, I think this helps a lot with seeing the new domain functionality.

internal/target/service_list_more.go
johanbrandhorst: Yes, this is my understanding of the RFC, that tokens should only expire if not used for 30 days.

internal/target/service_list_more.go
johanbrandhorst: I view refresh tokens as immutable, even though they are sometimes based on values from a previous refresh token, so I don't think this is the right function. Conceptually I think the grants hash should come from the auth result every time, not the previous token. Create time should come from the database, and the resource type is specific to the resource. Let me think about how to change this though.

internal/refreshtoken/refreshtoken.go
```diff
@@ -0,0 +1,73 @@
...
+	ResourceTypeUnknown ResourceType = iota
+	ResourceTypeSession
+	ResourceTypeTarget
+)
+
+// GrantsHasher defines the interface used
+// to retrieve a hash of a users grants.
+type GrantsHasher interface {
+	GrantsHash(ctx context.Context) ([]byte, error)
+}
```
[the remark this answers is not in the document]
johanbrandhorst: Good call, I've removed this and called it grants hash throughout

internal/refreshtoken/refreshtoken.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"bytes"
+	"context"
+	"time"
+
+	"github.com/hashicorp/boundary/internal/errors"
+)
+
+type ResourceType int
```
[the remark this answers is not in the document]
johanbrandhorst: :notbad:

internal/refreshtoken/refreshtoken.go
johanbrandhorst: There are two primary users of this type:
1. The application layer when converting from the protobuf type to the domain type.
2. The domain layer when constructing a new one from trusted information.
I don't think either of the use cases are helped by a constructor like this, as it doesn't do validation of the values, and if it's just a matter of creating the struct we can do that inline. We also wouldn't want a constructor to perform validation as it isn't always necessary.

internal/target/repository.go
johanbrandhorst: Good idea!

internal/target/repository.go
johanbrandhorst: Good idea

internal/refreshtoken/refreshtoken.go
tmessi: > I don't think either of the use cases are helped by a constructor like this, as it doesn't do validation of the values, and if it's just a matter of creating the struct we can do that inline. 

The `New` function above ensures that 3 related values, ResoureType, LastItemId, and LastItemUpdatedTime, all come from the same source. Thus the caller does not need to know or enforce the relationship between these fields. I think it also does perform some validation that is helpful, even if this is happening from a trusted source. For example, it ensures that you don't get a RefreshToken with a ResourceType of a Target, with a LastItemId for a Session.

Note: This would only be used when constructing *new* refresh tokens, not by the application layer for transformation.

internal/target/service_list_more.go
jimlambrt: My intention was to force a full refresh every 30 days.

internal/target/service_list.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/hashicorp/boundary/internal/refreshtoken"
+	"github.com/hashicorp/boundary/internal/types/resource"
+)
+
+func List(
+	ctx context.Context,
+	repo *Repository,
+	grantsHash []byte,
+	pageSize int,
+	filterItemFn func(Target) (bool, error),
```
tmessi, line 20: It might be a good idea to define this function as a type in the target package, ie:

```go
type ListFilterFunc func(Target) (bool, error)
```

Then the type can be documented to explain how it should behave.

internal/target/service_list_more.go
tmessi: Maybe something like `RefreshList`, `ListRefresh`, or `ListUpdatedSince`?

internal/target/service_list_response.go
tmessi: If it is in its own file, I don't think the file should start with `service_`. I think we should reserve that for files that contain service implementations, not just types.

internal/daemon/controller/handlers/targets/target_service.go
tmessi: You could move these into the [handlers](https://github.com/hashicorp/boundary/tree/mikemountain-domain-rewrite/internal/daemon/controller/handlers) package. That is where we have some things used by all the handler implementations.

internal/target/service_list_more.go
johanbrandhorst: Seems I misunderstood this part of the RFC, thanks for the clarification. We'll need to track two timestamps then, the create time and "update" time. We'll make the changes necessary!

internal/refreshtoken/refreshtoken.go
johanbrandhorst: Fair point, we created a `New`, `Refresh` and `RefreshLastItem` set of functions for mutating the refresh token, please take a look!

internal/target/service_list_more.go
johanbrandhorst: We've added a `New`, `Refresh` and `RefreshLastItem` set of functions, please take a look.

internal/target/service_list_more.go
johanbrandhorst: Done, thanks!

internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -234,87 +234,105 @@ func (s Service) ListTargets(ctx context.Context, req *pbs.ListTargetsRequest) (
...
+		rt, err := handlers.ParseRefreshToken(ctx, req.GetRefreshToken())
 		if err != nil {
 			return nil, err
 		}
-		if filter.Match(filterable) {
-			return pbItem, nil
+		// We're doing the conversion from the protobuf types to the
+		// domain types here rather than in the domain so that the domain
+		// doesn't need to know about the protobuf types.
+		domainRefreshToken := &refreshtoken.Token{
```
jimlambrt, line 283: You have a New(...) factory, but it's not used here?

jimlambrt (commented):
Overall, I think it's a big win.  Nothing but minor nits from me. I think, we should be careful when refactoring some of the bits into generic functions that readability doesn't suffer too much.

internal/daemon/controller/handlers/targets/target_service.go
johanbrandhorst: They are used for different purposes, the New factory is only used when we have a resource. I personally dislike constructors but I think we could make a constructor for this purpose and call it `New`, and in the existing use case for `New` we'll instead rename that function to `FromResource`.

internal/daemon/controller/handlers/refresh_token.go
```diff
@@ -0,0 +1,60 @@
...
+		pbs.ResourceType_RESOURCE_TYPE_SESSION_RECORDING,
+		pbs.ResourceType_RESOURCE_TYPE_AUTH_TOKEN,
+		pbs.ResourceType_RESOURCE_TYPE_ACCOUNT,
+		pbs.ResourceType_RESOURCE_TYPE_CREDENTIAL_LIBRARY,
+		pbs.ResourceType_RESOURCE_TYPE_CREDENTIAL_STORE:
+		return resource.Unknown
+	default:
+		return resource.Unknown
+	}
+}
```
johanbrandhorst, line 60: @mikemountain: we need to add a comment and tests for this before merge

internal/target/repository_test.go
```diff
@@ -312,3 +316,108 @@ func TestRepositoryListPermissionWhereClauses(t *testing.T) {
 		})
 	}
 }
+
+func Test_listDeletedIds(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+	conn, mock := db.TestSetupWithMock(t)
+	wrapper := db.TestWrapper(t)
+	mock.ExpectQuery(
```
johanbrandhorst, line 325: I don't love having to resort to sql mocks for these tests, unfortunately I can't think of any other way to test this since the functions are unexported, and there are no concrete implementations of targets in this package.

internal/target/service_list_refresh.go
```diff
@@ -0,0 +1,109 @@
...
+	// Use the timestamp of the deleted IDs transaction with a
+	// buffer to account for overlapping transactions. It is okay
+	// to return a deleted ID more than once. The buffer corresponds
+	// to Postgres' default transaction timeout.
+	updatedTime := transactionTimestamp.Add(-30 * time.Second)
+	if updatedTime.Before(tok.CreatedTime) {
+		// Ensure updated time isn't before created time due
+		// to the buffer.
+		updatedTime = tok.CreatedTime
+	}
```
johanbrandhorst, line 101: @jimlambrt @tmessi note this truncation of the updated time to at least the timestamp of the created time. It means much of our planned buffer disappears when a user performs the initial paginating-of-all-results, but hopefully it shouldn't matter since the number of deleted IDs since their tokens were created should also be small.

internal/target/list_response.go
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package target
+
+import (
+	"github.com/hashicorp/boundary/internal/refreshtoken"
+)
+
+// will likely eventually become a generic response and moved from target
```
mgaffney, line 10: Can we turn this into a TODO comment and then add proper godoc for this since it is exported?

internal/refreshtoken/refresh_token.go
```diff
@@ -0,0 +1,129 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package refreshtoken
```
mgaffney, line 4: Can we add godoc for the package?

jimlambrt (approved):

internal/target/service_list_refresh.go
tmessi: I know this is already merged, but maybe as a follow up, I think this logic should be in a method on the RefreshToken, i.e. the RefreshToken should ensure that it's updated time is never before it's creation time. I also wonder if the buffer time should be something defined in the refreshtoken package. I am thinking something like:

```go
package refreshtoken

const DefaultUpdateTimeBuffer = -30 * time.Second


func (r *RefreshToken) SetUpdateTime(u time.Time, opts ...Option) {
  // opts stuff

  uu = u.Add(opts.withUpdateTimeBuffer)
  if uu.Before(r.CreatedTime) {
     r.UpdatedTime = r.CreatedTime
  } else {
     r.UpdatedTime = uu
  }
}
```

internal/target/service_list.go
```diff
@@ -0,0 +1,94 @@
...
+	limit := pageSize + 1
+	opts := []Option{
+		WithLimit(limit),
+	}
+
+	targets := make([]Target, 0, limit)
+dbLoop:
+	for {
+		// Request another page from the DB until we fill the final items
+		page, err := repo.ListTargets(ctx, opts...)
```
tmessi, line 39: With this change, I think we should make the `ListTargets` method on the repository un-exported.

internal/target/service_list_refresh.go
johanbrandhorst: Fair point, I'll make a separate PR for this.

internal/target/service_list.go
johanbrandhorst: Good idea

johanbrandhorst: I had a quick look at this and unfortunately it means that it's very difficult to perform database testing of this method. Since all concrete implementations of targets exist in subpackages that import this package, it is impossible to use these types in a package testing this function (if it is unexported). I'm thus tempted to leave this exported just for ease of testing. We had the same problem with ListDeletedIds and EstimatedCount but those were easier to mock. I'd like to use the real DB for the list targets test since it relies heavily on the query for ordering and so on.

tmessi: You could add something to [exports_test](https://github.com/hashicorp/boundary/blob/main/internal/target/exports_test.go) to expose the method for tests, while leaving it un-exported outside of tests.

johanbrandhorst: 🤯 I didn't know we had a pattern like this, thank you!

PR #3825, opened 2023-10-11
Rate limit thin line

tmessi:
This adds the remaining pieces to complete a thin line for applying rate
limits to the controller api:

1. Adds an http middleware handler function to check and apply the
  limits. This middleware is added to the controller's API handler
  functions to check if a request is allowed by a rate.Limiter prior to
  processing the request.
2. Adds a rate.Limiter to the controller. This is initialized by
   generating a slice of rate.Limits from the ratelimit.Config provided
   on the Controller's config.

I am just opening this as a draft since there are several outstanding
things:

1. I still don't like some of the naming for the access to the action
   registrar. So any suggestions there are welcome.
2. In registering actions for each resource, I discovered a variety of
   ways the actions get defined and used in the app services. See
   b35da8fc2 for more details.  I think this can be cleaned up, but that
   might come later after more of the rate limiting functionality is
   done.
3. Additional tests need to be added in a number of places.


mgaffney (commented):
Great work!

internal/ratelimit/config.go
```diff
@@ -28,3 +41,124 @@ type Config struct {
...
+					Period:      DefaultPeriod,
+				}
+			}
+		}
+	}
+
+	for _, cc := range c {
+		var resourceSet []resource.Type
+		switch {
+		case len(cc.Resources) == 1 && cc.Resources[0] == "*":
```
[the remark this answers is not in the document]
tmessi: Yea, and actually we already have some, `resource.All` and `action.All`

internal/ratelimit/handler.go
```diff
@@ -0,0 +1,125 @@
...
+			return
+		}
+
+		allowed, quota, err := l.Allow(res, a)
+		if err != nil && !errors.Is(err, rate.ErrLimiterFull) {
+			// TODO: what to do here?
+			// The only other error here should be rate.ErrLimitNotFound, which
+			// shouldn't be possible given how we initialize the limiter,
+			// so perhaps a 500 makes sense?
+			rw.WriteHeader(http.StatusInternalServerError)
```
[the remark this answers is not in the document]
tmessi: I could see a 503 when the error is `rate.ErrLimiterFull`. I was going to do a 429, but the 503 might be better to distinguish between being rate limited vs the server being overloaded.

However, in the case of `rate.ErrLimitNotFound` I don't think it would be temporary, since it would indicate that the limiter was not properly initialized.

internal/types/action/registrar.go
```diff
@@ -0,0 +1,70 @@
...
+	collection ActionSet
+	valid      ActionSet
+}
+
+type byResource struct {
+	mu sync.RWMutex
+	m  map[resource.Type]*resourceActionSets
+}
+
+func (b *byResource) add(r resource.Type, individual ActionSet, collection ActionSet) error {
```
jimlambrt, line 24: should this just panic if there's an error since it's only used during
registration which should only happen during init?

internal/types/action/registrar.go
```diff
@@ -0,0 +1,70 @@
...
+var byResourceRegistrar = &byResource{}
+
+// RegisterResource registers the valid ActionSets for a resource.
+func RegisterResource(r resource.Type, individual ActionSet, collection ActionSet) error {
+	return byResourceRegistrar.add(r, individual, collection)
+}
+
+// SetForResource returns the valid ActionSet for the given resource.
+// If the provided resource was not registered, and error is returned.
+func SetForResource(r resource.Type) (ActionSet, error) {
```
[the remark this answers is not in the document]
jimlambrt: I like the original `action.SetForResource(...)` over
`action.ResourceActionSet(...)`.  With that said, what about `action.RegisteredActionSet(...)`?

internal/ratelimit/handler.go
```diff
@@ -0,0 +1,125 @@
...
+	"fmt"
+	"net/http"
+	"regexp"
+
+	"github.com/hashicorp/boundary/internal/types/action"
+	"github.com/hashicorp/boundary/internal/types/resource"
+	rate "github.com/hashicorp/go-rate"
+)
+
+var pathRegex = regexp.MustCompile(`/v1/(?P<resource>[\w-]+)((/(?P<id>[\w/]+))?(:(?P<action>[\w-:]+)?)?)?`)
```
jimlambrt, line 17: giddy up!

internal/ratelimit/handler.go
```diff
@@ -0,0 +1,125 @@
...
+
+func extractResourceAction(path, method string) (res, act string, err error) {
```
jimlambrt, line 19: I sort of like @mgaffney 's suggestion to return defined error codes which can
be used by the caller (the handler) to return the appropriate error/status.

internal/ratelimit/handler.go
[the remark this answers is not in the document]
jimlambrt: Hum... thinking a 429 if it's full and a 503 if it's not properly initialized.
Returning a 503 when the service is available but overloaded seems odd to me
since the client should likely behave the same way when experiencing a rate
limit or an overloaded service because it's out of quota.

internal/ratelimit/config.go
```diff
@@ -28,3 +41,124 @@ type Config struct {
...
+			resourceSet = allResources
+		default:
+			for _, r := range cc.Resources {
+				rr, ok := resource.Map[r]
+				if !ok {
+					return nil, fmt.Errorf("unknown resource %s", r)
```
jimlambrt, line 102: should we define a new error code for this?

internal/ratelimit/config.go
```diff
@@ -28,3 +41,124 @@ type Config struct {
...
+				}
+				validActionMap := make(map[string]action.Type, len(validActions))
+				for _, a := range validActions {
+					validActionMap[a.String()] = a
+				}
+
+				for _, aStr := range cc.Actions {
+					a, ok := validActionMap[aStr]
+					if !ok {
+						return nil, fmt.Errorf("action %s not valid for resource %s", aStr, res.String())
```
jimlambrt, line 142: should we define a new error code for this?

jimlambrt (changes requested):

internal/ratelimit/handler.go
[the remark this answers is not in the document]
tmessi: I am not sure I like a `503` for when it is not properly initialized, since it is not something that would be fixed by a configuration change, this would be a bug in the particular release and the only way the service would become available again is with an upgrade/downgrade. This is akin to a panic so I think a `500` would be better.

For when the limiter is full, I do kind of like `503` better. A `429` states:
> The 429 status code indicates that the user has sent too many requests in a given amount of time ("rate limiting").

But when the limiter is full, it is not that the user made too many requests, it is that the service is overwhelmed, which seems more in line with a `503` (my emphasis):
> The 503 (Service Unavailable) status code indicates that the server is currently unable to handle the request **due to a temporary overload** or scheduled maintenance, which will likely be alleviated after some delay.

In addition, with the 503 we can still set a `Retry-After` based on the time until quotas will be deleted, but we would not include the `RateLimit` header, since no quota was created.

As for clients, they could:
1. Minimally check for the `Retry-After` header and retry after that time.
2. Be explicit and check for either a 429 or 503 and the `Retry-After` header.

This would also help operators, since, with our existing monitoring, they could differentiate between requests that consumed too much quota (429), and requests that failed because of the limiter running out of space (503).

internal/ratelimit/handler.go
```diff
@@ -0,0 +1,125 @@
...
+// set to the number of seconds the client should wait to make it's next request.
+func Handler(l *rate.Limiter, next http.Handler) http.Handler {
+	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		path := req.URL.Path
+		method := req.Method
+
+		res, a, err := extractResourceAction(path, method)
+		if err != nil {
+			// TODO: what to do here?
+			// We could:
```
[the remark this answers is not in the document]
tmessi: For this and the status codes mentioned in comments above, see 1e1c7f9b6

internal/ratelimit/handler.go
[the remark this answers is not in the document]
tmessi: 44d49eb0d switches to a 503 on ErrLimiterFull, on other errors it is still a 500. @jimlambrt let me know if you disagree with my reasoning [above](https://github.com/hashicorp/boundary/pull/3825#discussion_r1376299682), I'm happy to continue discussing it.

internal/ratelimit/handler.go
tmessi: See: https://github.com/hashicorp/boundary/pull/3825#discussion_r1376685709

internal/ratelimit/config.go
tmessi: That seems like a good idea. In looking through the existing codes, I am a little surprised that we don't have something for a configuration error, which is what I was thinking this would be. It seems that for most of the config parsing we don't return errors with error codes.

Do you think I should make an error code specific to specifying a resource that is not known? Or should I look to add a more generic "configuration error"?

internal/daemon/controller/handlers/accounts/account_service.go
```diff
@@ -108,6 +108,23 @@ func init() {
 	if oidcMaskManager, err = handlers.NewMaskManager(context.Background(), handlers.MaskDestination{&oidcstore.Account{}}, handlers.MaskSource{&pb.Account{}, &pb.OidcAccountAttributes{}}); err != nil {
 		panic(err)
 	}
+
+	// TODO: make action.ActionSet a Set so this isnt necessary?
```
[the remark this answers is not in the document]
tmessi: I refactored `action.ActionSet` to be implemented as a set here: 66e44b75f
Then these TODOs are removed here: b5bfde605

internal/types/action/registrar.go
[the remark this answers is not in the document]
tmessi: To address some TODOs and the comment [here](https://github.com/hashicorp/boundary/pull/3825#discussion_r1373200746), I refactored `action.ActionSet` so it is implemented as a set, see 66e44b75f.

In that commit, I introduce a couple additional package functions that operate on ActionSets. For them I opted to use "ActionSet" in the function name, ie:

- `action.ActionSetUnion`
- `action.ActionSetDifference`

These stutter, which I would generally want to avoid, but it is consistent with `action.ActionSet`. I think I would prefer to either:
1. Be consistent, in which case I think function in question here should be `action.ActionSetForResource`
3. Or fix all of the stutters and refactor `action.ActionSet` to `action.Set`. This would allow this function to be `action.SetForResource` which could still be confusing if you see "Set" and think of a setter. However, since this is a package function rather than a method on a struct, and if there is a type called Set, I don't think it should be that confusing.

internal/types/action/registrar.go
tmessi: I updated RegisterResource to panic if called twice with the same resource, see: 21f7ac863. And then I refactored where this is called with the fixup here: 0cb15e6ec.

internal/daemon/controller/handlers/authmethods/z_init.go
```diff
@@ -0,0 +1,20 @@
...
+
+package authmethods
+
+import (
+	"github.com/hashicorp/boundary/internal/types/action"
+	"github.com/hashicorp/boundary/internal/types/resource"
+	"golang.org/x/exp/maps"
+)
+
+// init functions process in alphabetical order of filenames within a package.
```
jimlambrt, line 12: ty for the comment!

internal/ratelimit/config.go
jimlambrt: I'd start with `configuration error` in this case.

internal/ratelimit/handler.go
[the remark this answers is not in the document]
jimlambrt: I'm fine with it.  My thinking was: few clients handle 503s as a retry-able error but most will deal with 429s properly.  In other words, a 503 will like result in the client returning an error to the caller and a 429 will likely be retried before returning an error to the caller.

jimlambrt (commented):
Looks great.  Moving fwd!

tmessi:
@mgaffney @jimlambrt I moved this out of draft since it should be ready for a final review. All changes since the last review are done as fixup commits. They are mostly additional tests, some minor fixes due to test failures, and some fixes to handler and api test that has assertions that relied on the order of action sets.

Also, note that `make-gen-detla` is failing, I believe this is unrelated and should be resolved when I rebase the llb on main.

internal/daemon/controller/controller.go
```diff
@@ -119,6 +120,8 @@ type Controller struct {
 	apiGrpcServerListener grpcServerListener
 	apiGrpcGatewayTicket  string
 
+	ratelimiter *rate.Limiter
```
jimlambrt, line 123:
```suggestion
	rateLimiter *rate.Limiter
```

tmessi: Thanks, fixed in [03ef8dd](https://github.com/hashicorp/boundary/pull/3825/commits/03ef8ddb690b60ba9a01bf9e05021c321e373b53)

internal/types/action/registrar.go
```diff
@@ -0,0 +1,73 @@
...
+		return nil, fmt.Errorf("resource not found: %s", r)
+	}
+	return a, nil
+}
+
+var byResourceRegistrar = &byResource{}
+
+// RegisterResource registers ActionSets for r. If RegisterResource is called
+// twice with the same resource, it panics.
+func RegisterResource(r resource.Type, individual ActionSet, collection ActionSet) {
```
jimlambrt, line 59: do we want unit tests for these two exported functions.

jimlambrt (changes requested):

internal/daemon/controller/controller.go
```diff
@@ -245,6 +248,16 @@ func New(ctx context.Context, conf *Config) (*Controller, error) {
 	}
 	c.clusterListener = clusterListeners[0]
 
+	ratelimits, err := conf.RawConfig.Controller.ApiRateLimits.Limits(c.baseContext)
```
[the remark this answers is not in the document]
tmessi: Ty, fixed in [badaf09](https://github.com/hashicorp/boundary/pull/3825/commits/badaf096826502bc289108f8690f6effad122919)

internal/daemon/controller/handler.go
```diff
@@ -93,7 +94,8 @@ func (c *Controller) apiHandler(props HandlerProperties) (http.Handler, error) {
 		return nil, err
 	}
 
-	corsWrappedHandler := wrapHandlerWithCors(mux, props)
+	ratelimitedHandler := ratelimit.Handler(c.ratelimiter, mux)
```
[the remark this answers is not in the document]
tmessi: Ty, fixed in [badaf09](https://github.com/hashicorp/boundary/pull/3825/commits/badaf096826502bc289108f8690f6effad122919)

internal/types/action/registrar.go
tmessi: Added a test in [670d49b](https://github.com/hashicorp/boundary/pull/3825/commits/670d49b35466294b14617cd446a6ec05d44ed3e2) and then expanded on the test some in [cda1217](https://github.com/hashicorp/boundary/pull/3825/commits/cda1217130627c6280d87fdd810cf594f7804401). I looked at adding some additional test cases, like attempting to register and invalid `resource.Type`, but it seems we don't have an existing way to checking if a resource.Type is valid currently. I can look into adding some validation logic for it, but I think that should be a separate PR.

internal/ratelimit/config.go
```diff
@@ -4,15 +4,90 @@
...
 import (
+	"context"
+	"fmt"
+	"sync"
 	"time"
+
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/types/action"
+	"github.com/hashicorp/boundary/internal/types/resource"
+	rate "github.com/hashicorp/go-rate"
```
mgaffney, line 15:
```suggestion
	"github.com/hashicorp/go-rate"
```

internal/ratelimit/handler.go
```diff
@@ -0,0 +1,153 @@
...
+
+import (
+	"fmt"
+	"math"
+	"net/http"
+	"regexp"
+
+	"github.com/hashicorp/boundary/internal/types/action"
+	"github.com/hashicorp/boundary/internal/types/resource"
+	rate "github.com/hashicorp/go-rate"
```
mgaffney, line 14:
```suggestion
	"github.com/hashicorp/go-rate"
```

internal/types/action/action.go
```diff
@@ -215,29 +215,61 @@ func (act Type) IsActionOrParent(suspect Type) bool {
...
+
+// Add adds the provided Types t to a, any duplicates are ignored.
+func (a ActionSet) Add(t ...Type) {
+	for _, tt := range t {
+		a[tt] = struct{}{}
+	}
+}
+
+// ActionSetUnion returns a new ActionSet that is the union sets.
+func ActionSetUnion(sets ...ActionSet) ActionSet {
```
mgaffney, line 236:
```suggestion
func Union(sets ...ActionSet) ActionSet {
```
Why not just `Union`?

internal/types/action/action.go
```diff
@@ -215,29 +215,61 @@ func (act Type) IsActionOrParent(suspect Type) bool {
...
+	for _, s := range sets {
+		for k := range s {
+			a.Add(k)
+		}
+	}
+	return a
+}
+
+// ActionSetDifference returns a new ActionSet that is the difference of a-b.
+func ActionSetDifference(a ActionSet, b ActionSet) ActionSet {
```
mgaffney, line 247:
```suggestion
func Difference(a ActionSet, b ActionSet) ActionSet {
```
Why not just `Difference`?

internal/types/action/action.go
```diff
@@ -215,29 +215,61 @@ func (act Type) IsActionOrParent(suspect Type) bool {
...
+}
+
+// Add adds the provided Types t to a, any duplicates are ignored.
+func (a ActionSet) Add(t ...Type) {
+	for _, tt := range t {
+		a[tt] = struct{}{}
+	}
+}
+
+// ActionSetUnion returns a new ActionSet that is the union sets.
```
mgaffney, line 235:
```suggestion
// ActionSetUnion returns a new ActionSet that is the union of the sets.
```

internal/daemon/controller/controller.go
```diff
@@ -50,6 +50,7 @@ import (
 	external_plugins "github.com/hashicorp/boundary/sdk/plugins"
 	"github.com/hashicorp/boundary/version"
 	"github.com/hashicorp/go-hclog"
+	rate "github.com/hashicorp/go-rate"
```
mgaffney, line 53:
```suggestion
	"github.com/hashicorp/go-rate"
```

mgaffney (approved):

internal/types/action/action.go
tmessi: For the reasons I outlined [here](https://github.com/hashicorp/boundary/pull/3825#discussion_r1377660801), but happy to reconsider it.

mgaffney (approved):

jimlambrt (approved):
ty!

PR #3839, opened 2023-10-12
Add SQL changes to deleted tables, refactor deleted ID and count methods

johanbrandhorst:
This takes the changes introduced in https://github.com/hashicorp/boundary/pull/3761 and applies it to the new pattern. Most of this has already been reviewed.


johanbrandhorst:
This will be done in https://github.com/hashicorp/boundary/pull/3840 instead

PR #3840, opened 2023-10-12
Deleted tables, credential domain refactor

johanbrandhorst:
This brings in the DB changes from #3839 and refactors the credential domain to match the new pattern, moving pagination logic from the application layer to the domain layer.


internal/credential/credential.go
```diff
@@ -154,3 +155,11 @@ type SshPrivateKey interface {
...
 	PrivateKeyPassphrase() []byte
 }
+
+// SortResource defines a struct used to hold information
+// relating to a resource that is being used for sorting and
+// paginating.
+type SortResource struct {
+	PublicId   string
+	UpdateTime time.Time
+}
```
johanbrandhorst, line 165: This is used in the new credential domain-wide `WithStartPageAfterItem` option and can be used for a library, store or credential.

internal/credential/list_response.go
```diff
@@ -0,0 +1,38 @@
...
+
+// ListStoresResponse holds the information returned from ListStores and ListStoresRefresh
+// TODO(johanbrandhorst): Move this into a shared struct when refactoring.
+type ListStoresResponse struct {
+	Items               []Store
+	DeletedIds          []string
+	EstimatedTotalItems int
+	CompleteListing     bool
+	RefreshToken        *refreshtoken.Token
+}
```
johanbrandhorst, line 38: I have a prototype already that removes the need for these, but will wait to update this PR pending review of https://github.com/hashicorp/boundary/pull/3844

johanbrandhorst: #3856

internal/credential/credential.go
tmessi: A [boundary.Resource](https://github.com/hashicorp/boundary/blob/main/internal/boundary/boundary.go#L28-L32) has a `GetPublicId()` and a `GetUpdateTime()` method. Could that interface be used instead of introducing a new type?

tmessi (commented):
I started reviewing this, but I am wondering if I should wait for #3844 to get merged in first?

johanbrandhorst:
> I started reviewing this, but I am wondering if I should wait for #3844 to get merged in first?

The only change after #3844 is merged is in the service_list files, so everything else can be reviewed, but you can also wait :).

internal/credential/credential.go
johanbrandhorst: That's a great idea! Let me try that

johanbrandhorst: I really liked this idea but it wasn't quite possible to use `boundary.Resource`, so I defined a new interface in the options file. I also added a function to the refresh token to convert to a "partial resource" which can be used for this. Let me know what you think!

mikemountain (approved):

internal/refreshtoken/partial_resource.go
```diff
@@ -0,0 +1,29 @@
...
+
+	"github.com/hashicorp/boundary/internal/db/timestamp"
+	"github.com/hashicorp/boundary/internal/types/resource"
+)
+
+type partialResource struct {
+	publicId     string
+	updateTime   time.Time
+	resourceType resource.Type
+}
```
tmessi, line 17: This stands out to me as something odd. Partly due to the name "partial resource". While this is not exported, the same name is used elsewhere as an interface that this happens to implement. In fact there are several interfaces, one per domain package that are all the same, and this happens to implement them all. 

I think this might be cleaned up an simplified by a type in the `pagination` package.

```go
package pagination

type Item struct {
  PublicId   string
  UpdateTime time.Time
  Type       resource.Type
}
```

```go
package refreshtoken

func (rt *Token) LastItem() *pagination.Item {
  // alternatively, just change Token
  // to have a LastItem field of type pagination.Item
  return &pagination.Item{
    PublicId: rt.LastItemId,
    UpdateTime: rt.LastItemUpdatedTime,
    Type: rt.ResourceType,
  }
}
```

```go
package target

func WithStartPageAfterItem(item *pagination.Item) Option {
  return func(o *options) {
    o.WithStartPageAfterItem = item
  }
}
```

tmessi: Although looking at https://github.com/hashicorp/boundary/pull/3844 that is updating the `pagination` package to import `refreshtoken`, so the suggestion above would create an import cycle. So maybe `Item` should be in `refreshtoken`.

tmessi (changes requested):

internal/refreshtoken/partial_resource.go
johanbrandhorst: This doesn't change the need for an interface is the domain packages, since we need to support both using a `boundary.Resource` and an item from a refresh token. See the changes to `service_list_refresh.go` here: https://github.com/hashicorp/boundary/pull/3840/commits/8a20fb8f0c9221d64a91be13490cb3a5ca247785#diff-c8bedb31d2abcb15d2ea66a841d4a2608aa7bc879a089ad6a8c077ea661d9545L68. I'm happy to move the type around, but I still think we need the interface in each domain.

tmessi: Maybe have the interface in `pagination` then? ie:

```go
package pagination

type Item interface {
  GetPublicId() string
  GetUpdateTime() time.Time
  GetResourceType() resource.Type
}
```

```go
package refreshtoken

type Item struct {} // implements pagination.Item

func (rt *Token) LastItem() *Item {
  // alternatively, just change Token
  // to have a LastItem field of type Item
  return &Item{
    PublicId: rt.LastItemId,
    UpdateTime: rt.LastItemUpdatedTime,
    Type: rt.ResourceType,
  }
}
```

```go
package target

func WithStartPageAfterItem(item pagination.Item) Option {
  return func(o *options) {
    o.WithStartPageAfterItem = item
  }
}
```

johanbrandhorst: That works for me!

johanbrandhorst:
This is now at the head of the review queue, please take a look!

internal/credential/service_library.go
```diff
@@ -0,0 +1,45 @@
...
+// VaultLibraryRepository defines the interface expected
+// to get the total number of credential libraries and deleted ids.
+type VaultLibraryRepository interface {
+	EstimatedLibraryCount(context.Context) (int, error)
+	EstimatedSSHCertificateLibraryCount(context.Context) (int, error)
+	ListDeletedLibraryIds(context.Context, time.Time, ...Option) ([]string, error)
+	ListDeletedSSHCertificateLibraryIds(context.Context, time.Time, ...Option) ([]string, error)
+	ListCredentialLibraries(context.Context, string, ...Option) ([]Library, error)
+	ListSSHCertificateCredentialLibraries(context.Context, string, ...Option) ([]Library, error)
+}
```
tmessi, line 24: This doesn't seem like a "repository" to me, even though these are all methods that exist on the vault.Repository, this combination of methods by themselves does not seem like a repository.

In addition, this exposes details of `vault` into `credential`. I am thinking it might make sense to instead create an interface like:

```go
type SubtypeLibraryService interface {
    EstimatedCount(context.Context) (int, error)
    ListDeletedIds(context.Context, time.Time, ...Option) ([]string, error)
    List(context.Context, string, ...Option) ([]Library, error)
}
```

Then in `vault` add a service that implements this interface. This allows for vault to manage combining SSHCertificateLibraries with GenericLibraries. Then the `credential.LibraryService` can receive a `[]SubtypeLibraryService` that can be expanded if/when there are other non-vault credential libraries.

johanbrandhorst: That's a nice idea, let me try it!

johanbrandhorst: I liked this, and I think it's also a pattern that we could in theory apply to more resources, but I don't think it's worth the extra indirection. Please take a look!

internal/credential/vault/service_list.go
```diff
@@ -0,0 +1,91 @@
...
+		if err != nil {
+			return err
+		}
+		transactionTimestamp, err = r.Now(ctx)
+		if err != nil {
+			return err
+		}
+		deletedIds = append(deletedGenericIds, deletedSSHCertIds...)
+		return nil
+	}); err != nil {
```
johanbrandhorst, line 87: I went back and forth on whether to implement the transaction and timestamp here or in the credential service layer. I think from a design point of view it is more "correct" for this to be in the credential service layer, since in theory any future credential library implementation would need to have its ListDeletedIds called from that layer, and so that is where the transaction and timestamp would go. However, doing it this way lets us forgo the credential.LibraryService struct altogether, reducing clutter in the controller and the credential library application layer, so I went this route. It's not a big change to make if we add another credential library type in the future.

I'm happy to change it if this is a concern, of course 😄.

internal/credential/service_list_credentials.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/hashicorp/boundary/internal/pagination"
+)
+
+// Repository defines the interface expected
+// to get the total number of credentials and deleted ids.
+type Repository interface {
+	EstimatedCredentialCount(context.Context) (int, error)
+	ListDeletedCredentialIds(context.Context, time.Time) ([]string, time.Time, error)
+	ListCredentials(context.Context, string, ...Option) ([]Static, error)
+}
```
tmessi, line 19: Similarly to the credential libraries, this combination of methods does not seem like a Repository. Perhaps name this `CredentialService` to follow what you did for `LibraryService`?

internal/credential/service_store.go
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) HashiCorp, Inc.
```
tmessi, line 1: Can this file be combined into `service_list_stores.go`? For libraries and credentials that seems to be the pattern. This is the only one that has a separate `service_<type>.go` file.

internal/credential/service_store.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/hashicorp/boundary/internal/util"
+)
+
+// StoreRepository defines the interface expected
+// to gather information about credential stores.
+type StoreRepository interface {
+	EstimatedStoreCount(context.Context) (int, error)
+	ListDeletedStoreIds(context.Context, time.Time, ...Option) ([]string, error)
+	ListCredentialStores(context.Context, []string, ...Option) ([]Store, error)
+}
```
tmessi, line 21: Similarly to credentials, this combination methods does not seems like a repository. Can this be made into something like `SubtypeStoreService`?

tmessi (changes requested):

internal/credential/service_store.go
johanbrandhorst: I will merge it, but the reason it was separate was because it is now the only resource which has a service struct at this level. Prior to the libraries refactor it too had a `service_library.go` file.

johanbrandhorst:
Converting to draft pending reworking of PR structure.

johanbrandhorst:
Superceded by #3898, #3901 and #3902

PR #3841, opened 2023-10-12
draft: refactor auth methods domain layer

mikemountain:
This PR is a POC of a new pattern in the auth domain. It is not thoroughly tested but we've pushed it up so that we can get early feedback on the design. The biggest difference between this design and the one in the just-merged #3817 is the use of a service struct instantiated in the controller. This is required because when listing deleted IDs it needs to happen in a single transaction across all repositories.


internal/auth/service_list_auth_methods_refresh.go
```diff
@@ -0,0 +1,159 @@
...
+		if err != nil {
+			return errors.Wrap(ctx, err, op)
+		}
+		transactionTimestamp, err = r.Now(ctx)
+		if err != nil {
+			return errors.Wrap(ctx, err, op)
+		}
+		deletedIds = append(deletedLdapIds, append(deletedOidcIds, deletedPwIds...)...)
+		return nil
+	}); err != nil {
```
johanbrandhorst, line 49: This logic is what requires us to create a service struct. We need the deleted auth methods from all 3 domains to be collated in a single transaction with a single transaction timestamp. This requires a DB reader, which isn't available in the handlers, so we need to create the service struct in the controller. If we didn't need to do this all in a single transaction, we could forgo the whole struct.

internal/auth/service_list_auth_methods_refresh.go
```diff
@@ -0,0 +1,159 @@
...
+		if err != nil {
+			return nil, errors.Wrap(ctx, err, op)
+		}
+		finalPage := append(ldapPage, append(oidcPage, pwPage...)...)
+		slices.SortFunc(finalPage, func(i, j AuthMethod) int {
+			return i.GetUpdateTime().AsTime().Compare(j.GetUpdateTime().AsTime())
+		})
+		if len(finalPage) > limit {
+			finalPage = finalPage[:limit]
+		}
```
johanbrandhorst, line 85: This is a little annoying as we're requesting 3x the size of each page, but I'm not sure there's another way to do it. This logic guarantees that we'll get the correct order of auth methods, even if all of them are in one particular auth method.

internal/auth/ldap/repository_auth_method_read.go
```diff
@@ -32,19 +34,36 @@ func (r *Repository) LookupAuthMethod(ctx context.Context, publicId string, opt
...
 }
 
-// ListAuthMethods returns a slice of AuthMethods for the scopeId. The
-// WithUnauthenticatedUser, WithLimit and WithOrder options are supported and
-// all other options are ignored.
-func (r *Repository) ListAuthMethods(ctx context.Context, scopeIds []string, opt ...Option) ([]*AuthMethod, error) {
+// ListAuthMethods returns a slice of AuthMethods for the scopeId. Supported options:
+//   - auth.WithLimit
+//   - auth.WithStartPageAfterItem
+func (r *Repository) ListAuthMethods(ctx context.Context, scopeIds []string, opt ...auth.Option) ([]auth.AuthMethod, error) {
```
johanbrandhorst, line 40: The only way to allow us to call into the repositories from the `auth` package is to avoid any types defined in the subpackages (or we'd have import cycles). That is why we're using `auth.Option` and `auth.AuthMethod` here.

internal/auth/ldap/repository_auth_method_read.go
```diff
@@ -32,19 +34,36 @@ func (r *Repository) LookupAuthMethod(ctx context.Context, publicId string, opt
...
+	}
+	newOpts := []Option{
+		WithUnauthenticatedUser(ctx, opts.WithUnauthenticatedUser),
+	}
+	if opts.WithLimit != 0 {
+		newOpts = append(newOpts, WithLimit(ctx, opts.WithLimit))
+	}
+	if opts.WithStartPageAfterItem != nil {
+		newOpts = append(newOpts, WithStartPageAfterItem(ctx, opts.WithStartPageAfterItem.PublicId, opts.WithStartPageAfterItem.UpdateTime))
+	}
```
johanbrandhorst, line 57: We have to convert the auth options to the repository specific options here before calling `getAuthMethod`, since that function is shared with `LookupAuthMethod`.

internal/auth/ldap/repository_auth_method_read.go
```diff
@@ -32,19 +34,36 @@ func (r *Repository) LookupAuthMethod(ctx context.Context, publicId string, opt
...
+	authMethods, err := r.getAuthMethods(ctx, "", scopeIds, newOpts...)
 	if err != nil {
 		return nil, errors.Wrap(ctx, err, op)
 	}
-	return authMethods, nil
+	var ams []auth.AuthMethod
+	for _, am := range authMethods {
+		ams = append(ams, am)
+	}
```
johanbrandhorst, line 65: We're forced to make this simple conversion before returning, for the same reason as above.

internal/auth/ldap/repository_auth_method_read.go
```diff
@@ -206,6 +225,44 @@ func (r *Repository) getAuthMethods(ctx context.Context, authMethodId string, sc
...
+	var count int
+	for rows.Next() {
+		if err := r.reader.ScanRows(ctx, rows, &count); err != nil {
+			return 0, errors.Wrap(ctx, err, op, errors.WithMsg("failed to query total auth methods"))
+		}
+	}
+	return count, nil
+}
+
```
johanbrandhorst, line 265: Note to @mikemountain; we need to actually implement the logic for the new `StartPageAfterItem` in `getAuthMethod` below.

internal/auth/oidc/auth_method.go
```diff
@@ -464,3 +466,20 @@ func ParseAccountClaimMaps(ctx context.Context, m ...string) ([]ClaimMap, error)
 	}
 	return claimMap, nil
 }
+
+// sortAuthMethod is used to encapsulate the information
+// needed for sorting and filtering for pagination.
+type sortAuthMethod struct {
```
mikemountain, line 472: remove

PR #3844, opened 2023-10-14
New generic pagination

johanbrandhorst:
## [internal/pagination: add new pagination logic](https://github.com/hashicorp/boundary/commit/0f56a6a7ee1a8b9797a0bb663f92f39c35d6c2b9)

This unifies the core of the pagination logic in
the new pattern, significantly simplifying the
domain logic and removing repetition. This also
makes it easier to test the generic logic in
isolation.

I've established that this pattern can be applied
both to the simpler pagination in the target
domain and the more complicated logic in
the auth domain.

## [internal/target: migrate to generic pagination](https://github.com/hashicorp/boundary/commit/1f0e86fa497d0020c76a8ce68562543d880918d5)

Jim, I hope this doesn't add too much indirection to the domain logic. I'm open to suggestion to improve it in any way, but I found that implementing the pagination logic through copy-paste-edit was error prone and this significantly reduces the burden when implementing a new resource.

### DIFF
--- a/internal/pagination/pagination2.go
+++ b/internal/pagination/pagination2.go
@@ -127,13 +127,13 @@ func List[T boundary.Resource](
 // resources that have been deleted since the refresh token was last used.
 func ListRefresh[T boundary.Resource](
 	ctx context.Context,
-	tok *refreshtoken.Token,
 	grantsHash []byte,
 	pageSize int,
 	filterItemFn ListFilterFunc[T],
 	listRefreshItemsFn ListRefreshItemsFunc[T],
 	estimatedCountFn EstimatedCountFunc,
 	listDeletedIDsFn ListDeletedIDsFunc,
+	tok *refreshtoken.Token,
 ) (*ListResponse2[T], error) {
 	const op = "pagination.ListRefresh"
 
--- a/internal/pagination/pagination2.go
+++ b/internal/pagination/pagination2.go
@@ -1,0 +1,223 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package pagination
+
+import (
+	"context"
+	"time"
+
+	"github.com/hashicorp/boundary/internal/boundary"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/refreshtoken"
+)
+
+// ListResponse2 represents the response from the paginated list operation.
+type ListResponse2[T boundary.Resource] struct {
+	// Items contains the page of items. They have
+	// been filtered according to the behaviour of
+	// the filter item func, and will always have as
+	// many items as requested in the page size, unless
+	// there were not enough elements available, in
+	// which case it will contain all elements that
+	// were available.
+	Items []T
+	// CompleteListing signifies whether this page contains
+	// the final item currently available. This indicates
+	// that it may be appropriate to wait some time before
+	// requesting additional pages.
+	CompleteListing bool
+	// RefreshToken is the token that the caller can use
+	// to request a new page of items. The items in the
+	// new page will have been updated more recently
+	// than all the items in the previous page. This
+	// field may be empty if there were no results for a
+	// List call.
+	RefreshToken *refreshtoken.Token
+	// DeletedIds contains a list of item IDs that have been
+	// deleted since the last request for items. This can happen
+	// both during the initial pagination or when requesting a
+	// refresh. This is always empty for the initial List call.
+	DeletedIds []string
+	// EstimatedItemCount is an estimate on exactly how many
+	// items matching the filter function are available. If
+	// a List call is complete, this number is equal to
+	// the number of items returned. Otherwise, the
+	// estimated count function is consulted for an estimate.
+	EstimatedItemCount int
+}
+
+// ListFilterFunc is a callback used to filter out resources that don't match
+// some criteria. The function must return true for items that should be included in the final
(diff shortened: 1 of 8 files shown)


internal/pagination/pagination2.go
```diff
@@ -0,0 +1,182 @@
...
+	"context"
+	"time"
+
+	"github.com/hashicorp/boundary/internal/boundary"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/refreshtoken"
+)
+
+// ListResponse2 represents the response from the paginated list operation.
+type ListResponse2[T boundary.Resource] struct {
```
johanbrandhorst, line 16: This type will be renamed to `ListResponse` when we can remove the old pagination logic.

internal/pagination/pagination2.go
```diff
@@ -0,0 +1,182 @@
...
+	RefreshToken       *refreshtoken.Token
+	DeletedIds         []string
+	EstimatedItemCount int
+}
+
+// This function is a callback passed down from the application service layer
+// used to filter out resources that don't match any user-supplied filter.
+type ListFilterFunc[T boundary.Resource] func(context.Context, T) (bool, error)
+
+type ListItemsFunc2[T boundary.Resource] func(context.Context, T, int) ([]T, error)
```
johanbrandhorst, line 28: DItto

internal/pagination/pagination2.go
johanbrandhorst, line 1: This file will be renamed to `pagination.go` when the old one can be removed.

Makefile
```diff
@@ -300,7 +300,7 @@ test-sql:
 
 .PHONY: test
 test:
-	go test "$(TEST_PACKAGE)" -tags="$(BUILD_TAGS)" $(TESTARGS) -json -cover -timeout $(TEST_TIMEOUT) | tparse -follow
+	go test "$(TEST_PACKAGE)" -tags="$(BUILD_TAGS)" $(TESTARGS) -json -cover -timeout $(TEST_TIMEOUT)
```
johanbrandhorst, line 303: I temporarily removed this to help me debug the test failure

internal/pagination/pagination2.go
```diff
@@ -0,0 +1,189 @@
...
+
+// ListDeletedIDsFunc is used to list the IDs of the resources deleted since
+// the given timestamp. It returns a slice of IDs and the timestamp of the
+// instant in which the slice was created.
+type ListDeletedIDsFunc func(ctx context.Context, since time.Time) ([]string, time.Time, error)
+
+// List lists items according to the page size,
+// filtering out entries that do not
+// pass the filter item fn. It returns a new refresh token
+// based on the grants hash and the returned items.
```
tmessi, line 44: I think this could include additional details to explain more of what this function does, for example:

```go
// List returns a ListResponse. The response will contain at most a
// number of items equal to the pageSize. Items are fetched using the
// listItemsFn and then items are checked using the filterItemFn
// to determine if they should be included in the response.
// The response includes a new refresh token based on the grants and items.
// The estimatedCountFn is used to provide an estimated total number of
// items that can be returned by making additional requests using the provided
// refresh token.

File: internal/pagination/pagination2.go [hunk 5]
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	return resp, nil
+}
+
+// ListRefresh lists items according to the page size
+// and refresh token, filtering out entries that do not
+// pass the filter item fn. It returns a new refresh token
+// based on the old one, the grants hash, and the returned
+// items.
```
tmessi, line 87: Similarly I think this godoc could be expanded upon so any caller understands what each of the inputs are for and what to expect in the response.

internal/pagination/pagination2.go
```diff
@@ -0,0 +1,189 @@
...
+func ListRefresh[T boundary.Resource](
+	ctx context.Context,
+	tok *refreshtoken.Token,
+	grantsHash []byte,
+	pageSize int,
+	filterItemFn ListFilterFunc[T],
+	listItemsFn ListItemsFunc2[T],
+	estimatedCountFn EstimatedCountFunc,
+	listDeletedIDsFn ListDeletedIDsFunc,
```
tmessi, line 96: I am wondering if there is a way to ensure that these functions only operate on the resource `T`. As it stands, there is nothing at compile time that would prevent using `target.Target` for `T` and then passing an `estimatedCountFn` that provides the count of `hosts.Host`.

tmessi (changes requested):

internal/pagination/pagination2.go
johanbrandhorst: Good point. As you recall, originally this functionality was behind a single repository interface rather than 3 separate callbacks. I wonder if it would make sense to add that single repository back, so that the caller is forced to have
```
ListItems(..)([]T, error)
EstiamtedCount(...)(int, error)
ListDeletedIds(...)([]string, time.Time, error)
```
On a single value? It would add a bit of toil on the caller side because you'd have to have some type that just curries the functions but it would at least tie the `T` to the other two functions, perhaps?

Alternatively, having them take a `T` parameter that's completely unused, but that seems like it could get confusing for future readers. What do you think?

tmessi: I am not sure I like either enough to request a specific change.

tmessi (approved):

tmessi (approved):

jimlambrt (approved):
Ty for your patience.

PR #3846, opened 2023-10-16
test(e2e): Add test for creating an auth method

moduli:
This PR adds a test to the e2e test suite that does the following
- creates a new org
- creates a new password auth method in that org
- sets the new auth method as the primary auth method for the new org
- creates an account in the new auth method
- logs in to the new org using the new auth method (a user will be auto-created)
- create another account in the new auth method
- create another user and attach account
- log in as the new user

### DIFF
--- a/testing/internal/e2e/tests/base/auth_method_password_test.go
+++ b/testing/internal/e2e/tests/base/auth_method_password_test.go
@@ -1,0 +1,99 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package base_test
+
+import (
+	"context"
+	"encoding/json"
+	"testing"
+
+	"github.com/hashicorp/boundary/api/authmethods"
+	"github.com/hashicorp/boundary/api/scopes"
+	"github.com/hashicorp/boundary/testing/internal/e2e"
+	"github.com/hashicorp/boundary/testing/internal/e2e/boundary"
+	"github.com/stretchr/testify/require"
+)
+
+// TestCliAuthMethodPassword uses the boundary cli to create a new password auth
+// method, set it as the primary auth method for a scope, and log in to boundary
+// using a new account in that auth method.
+func TestCliAuthMethodPassword(t *testing.T) {
+	e2e.MaybeSkipTest(t)
+
+	ctx := context.Background()
+	boundary.AuthenticateAdminCli(t, ctx)
+	newOrgId := boundary.CreateNewOrgCli(t, ctx)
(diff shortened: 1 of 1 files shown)


testing/internal/e2e/tests/base/auth_method_password_test.go
```diff
@@ -0,0 +1,99 @@
...
+			"-name", "e2e Auth Method",
+			"-format", "json",
+		),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+	var newAuthMethodResult authmethods.AuthMethodCreateResult
+	err := json.Unmarshal(output.Stdout, &newAuthMethodResult)
+	require.NoError(t, err)
+	newAuthMethodId := newAuthMethodResult.Item.Id
+	// Need to manually clean up auth method until ICU-7833 is addressed
```
stasryzhov, line 48: 👍

testing/internal/e2e/tests/base/auth_method_password_test.go
```diff
@@ -0,0 +1,99 @@
...
+			"-id", newOrgId,
+			"-primary-auth-method-id", newAuthMethodId,
+			"-format", "json",
+		),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+	var updateScopeResult scopes.ScopeUpdateResult
+	err = json.Unmarshal(output.Stdout, &updateScopeResult)
+	require.NoError(t, err)
+	require.Equal(t, newAuthMethodId, updateScopeResult.Item.PrimaryAuthMethodId)
```
stasryzhov, line 85: Apart from this scenario, do we want to also check that a new user is created based on a primary auth method when it's used for the first time?

doc: https://developer.hashicorp.com/boundary/docs/commands/scopes/update#primary-auth-method-id

moduli: I'm not sure I follow what the expected behavior is there.

moduli: Oh I see now...

stasryzhov: We could do this:
1. Create a login/password auth method and its account
2. Make the auth method primary
3. Login with that login/password
4. Login as admin
5. Check that a new user was created and the primary auth method is attached to it

moduli: Added.

johanbrandhorst (approved):

stasryzhov (approved):
🚀

PR #3848, opened 2023-10-16
Target listing review comments

johanbrandhorst:
### [internal/target: unexport ListTargets](https://github.com/hashicorp/boundary/commit/d4b922d2285f582397392b3a1357c0489b11401a)

This also adds back the testing of ListDeletedIds and EstimatedCount using the pattern of exporting internal methods to test files only.


internal/refreshtoken/refresh_token.go
```diff
@@ -76,17 +81,26 @@ func FromResource(res boundary.Resource, grantsHash []byte) *Token {
...
-// RefreshLastItem refreshes a token's updated time and last item
+// RefreshLastItem refreshes a token's updated time and last item.
+// It accounts for database transaction inaccuracies by subtracting
+// UpdatedTimeBuffer from the provided timestamp while ensuring that
+// the updated time is never before the created time of the token.
 func (rt *Token) RefreshLastItem(res boundary.Resource, updatedTime time.Time) *Token {
-	rt.UpdatedTime = updatedTime
 	rt.LastItemId = res.GetPublicId()
 	rt.LastItemUpdatedTime = res.GetUpdateTime().AsTime()
+	rt = rt.Refresh(updatedTime)
```
johanbrandhorst, line 103: Now that Refresh does more than a trivial amount of work, we call it from here instead of copying the logic

internal/target/exports_test.go
```diff
@@ -7,4 +7,7 @@ package target
 var (
 	AllocTargetView         = allocTargetView
 	TargetsViewDefaultTable = targetsViewDefaultTable
+	ListDeletedIds          = (*Repository).listDeletedIds
+	EstimatedCount          = (*Repository).estimatedCount
+	ListTargets             = (*Repository).listTargets
```
johanbrandhorst, line 12: 👨🏻‍🍳 💋

internal/target/repository_credential_source_test.go
```diff
@@ -44,9 +44,12 @@ func (h hooks) VetCredentialSources(ctx context.Context, cls []*target.Credentia
 	return targettest.VetCredentialSources(ctx, cls, creds)
 }
 
+func init() {
+	target.Register(targettest.Subtype, hooks{}, globals.TcpTargetPrefix)
+}
+
 func TestRepository_SetTargetCredentialSources(t *testing.T) {
 	ctx := context.Background()
-	target.Register(targettest.Subtype, hooks{}, globals.TcpTargetPrefix)
```
johanbrandhorst, line 53: I had to move this to make these hooks available to all external tests. Without this, ListTarget fails to read the test target type from the DB.

internal/target/repository_test.go
```diff
@@ -316,108 +312,3 @@ func TestRepositoryListPermissionWhereClauses(t *testing.T) {
...
-	})
-
-	t.Run("errors when query errors", func(t *testing.T) {
-		mock.ExpectQuery(
-			`select`,
-		).WillReturnError(fmt.Errorf("query error"))
-		_, err = repo.estimatedCount(ctx)
-		require.ErrorContains(t, err, "query error")
-	})
-}
```
johanbrandhorst, line 423: So happy to replace these with real tests again, thanks @tmessi!

internal/target/targettest/target.go
```diff
@@ -368,5 +369,10 @@ func TestNewTestTarget(ctx context.Context, t *testing.T, conn *db.DB, projectId
 		err := rw.CreateItems(context.Background(), newCredLibs)
 		require.NoError(err)
 	}
+	if len(opts.WithAddress) != 0 {
+		addr := target.TestNewTargetAddress(id, opts.WithAddress)
+		err := rw.Create(ctx, addr)
+		require.NoError(err)
+	}
```
johanbrandhorst, line 376: Had to add this to support the `WithAddress` option to targettest.

internal/target/tcp/repository_test.go
```diff
@@ -357,42 +221,3 @@ func TestRepository_DeleteTarget(t *testing.T) {
...
-	require.NoError(t, err)
-
-	got, err := repo.ListTargets(ctx, target.WithLimit(-1))
-	require.NoError(t, err)
-	assert.Equal(t, total, len(got))
-
-	for _, tar := range got {
-		assert.Equal(t, "1.2.3.4", tar.GetAddress())
-	}
-}
```
johanbrandhorst, line 398: These tests were all moved to the target package mostly unchanged

mikemountain (approved):

internal/target/repository_credential_source_test.go
tmessi: It might be good to move this `init` into a `ext_test.go` file. And we should probably rename any test files in this package that use `package target_test` to have `_ext_test.go` as a suffix. I think @jimlambrt started that pattern of using `_ext_test` and it would probably help make it easier to identify which files contain tests that will be in the external package.

internal/refreshtoken/refresh_token.go
```diff
@@ -17,6 +17,11 @@ import (
 	"github.com/hashicorp/boundary/internal/types/resource"
 )
 
+// UpdatedTimeBuffer is used to automatically adjust the updated
+// time of a refresh token to account for delays between database
+// transactions. The buffer corresponds to Postgres' default transaction timeout.
```
tmessi, line 22: > The buffer corresponds to Postgres' default transaction timeout.

I don't think this is correct. As far as I know, postgres does not have a transaction timeout. There is an `idle_in_transaction_session_timeout` that will terminate a session that has been idle for too long. And there is `statement_timeout` but the default is `0`, meaning no timeout.

internal/target/repository_credential_source_test.go
johanbrandhorst: Good idea, I'll just move this test into the `_ext_test.go` file.

internal/refreshtoken/refresh_token.go
johanbrandhorst: @mikemountain where did we source this from? I can't remember 😅

mikemountain: @johanbrandhorst I can't seem to find where we did end up sourcing this from... but that being said, I think 30 seconds is still a reasonable buffer

johanbrandhorst: I'll replace the Postgres timeouts with a different blurb 👍🏻

internal/refreshtoken/refresh_token.go
```diff
@@ -76,17 +81,26 @@ func FromResource(res boundary.Resource, grantsHash []byte) *Token {
 	}
 }
 
-// Refresh refreshes a token's updated time
+// Refresh refreshes a token's updated time. It accounts for database
+// transaction inaccuracies by subtracting UpdatedTimeBuffer from the
```
tmessi, line 85: I don't think "inaccuracies" is the right word here. It is the nature of a [MVCC](https://www.postgresql.org/docs/current/mvcc-intro.html). There may have been items that have an older update time than `updatedTime` that where not visible at the time `updatedTime` was returned.

johanbrandhorst: Reworded this slightly, please take a look.

tmessi (approved):

PR #3851, opened 2023-10-17
chore: Bump gRPC to v1.58.3

hugoghx:
gRPC v1.58.3 fixes a vulnerability in the HTTP stack where a malicious HTTP/2 client which rapidly creates requests and immediately resets them can cause excessive server resource consumption.

See https://github.com/grpc/grpc-go/pull/6708 for more details.

Also updates the CHANGELOG and Boundary version (backport from `release/0.14.1`)


CHANGELOG.md
```diff
@@ -2,7 +2,17 @@
 
 Canonical reference for changes, improvements, and bugfixes for Boundary.
 
- ## Next
+## Next
+
+## 0.14.1 (2023/10/17)
+
+### Bug Fixes
```
tmessi, line 9: I think this should be `Security`. That seems to be what we have used in the past for things like this.

version/VERSION
```diff
@@ -1 +1 @@
-0.14.0
+0.14.1
```
tmessi, line 1: For main, I think this should be `0.15.0` now?

CHANGELOG.md
hugoghx: Oh, TIL! Addressed with https://github.com/hashicorp/boundary/pull/3851/commits/ea53302945207c0f8aa91207fd545c717d388a18

version/VERSION
tmessi: I believe this is used to determine the version of any artifact that is built on commits to a branch. So for example the artifacts from [this](https://github.com/hashicorp/boundary/commit/7d57a99e348965ec0b5403190329f997c7a3dfbb) commit [on main](https://github.com/hashicorp/boundary/actions/runs/6501744981) show as `0.14.0`.

I think this is supposed to be set to the *next* version so that the artifacts are already built with the correct version when we go to do a release. And since any 0.14.x version should come from the release/0.14.x branch, main should be set to 0.15.0.

hugoghx: Interesting! I've made this change now: 

In the release branch, however, it wouldn't be set to `0.15.0` right?

tmessi: In the release branch it should be 0.14.1. And after 0.14.1 is released we should bump the release branch to 0.14.2, although that might happen automatically, I can't remember if that is part of the release automation now or not.

moduli: Agreed with Tim. It also helps to ensure we're developing on `main` with the current deprecations. When we did `0.14.0`, we didn't bump the version number until release time and a bunch of tests started failing because of some deprecated features.

tmessi (approved):

PR #3857, opened 2023-10-18
Store and use refresh tokens in the client cache

talanknight:
This PR stores refresh tokens and uses them for subsequent requests.

If the refresh token gets old enough or we detect an invalid refresh token error the refresh token is deleted from the cache.

### DIFF
--- a/internal/clientcache/internal/cache/repository_refresh_token.go
+++ b/internal/clientcache/internal/cache/repository_refresh_token.go
@@ -5,7 +5,6 @@ package cache
 
 import (
 	"context"
-	"time"
 
 	"github.com/hashicorp/boundary/internal/db"
 	"github.com/hashicorp/boundary/internal/errors"
@@ -96,8 +95,6 @@ func upsertRefreshToken(ctx context.Context, writer db.Writer, u *user, rt resou
 		UserId:       u.Id,
 		ResourceType: rt,
 		RefreshToken: tok,
-		UpdateTime:   time.Now(),
-		CreateTime:   time.Now(),
 	}
 
 	switch tok {
@@ -106,7 +103,7 @@ func upsertRefreshToken(ctx context.Context, writer db.Writer, u *user, rt resou
 	default:
 		onConflict := &db.OnConflict{
 			Target: db.Columns{"user_id", "resource_type"},
-			Action: db.SetColumns([]string{"refresh_token", "update_time"}),
+			Action: db.SetColumns([]string{"refresh_token"}),
 		}
 		if err := writer.Create(ctx, refTok, db.WithOnConflict(onConflict)); err != nil {
 			return errors.Wrap(ctx, err, op)
@@ -136,8 +133,6 @@ type refreshToken struct {
 	UserId       string       `gorm:"primaryKey"`
 	ResourceType resourceType `gorm:"primaryKey"`
 	RefreshToken RefreshTokenValue
-	UpdateTime   time.Time `gorm:"default:(strftime('%Y-%m-%d %H:%M:%f','now'))"`
-	CreateTime   time.Time `gorm:"default:(strftime('%Y-%m-%d %H:%M:%f','now'))"`
 }
 
 func (*refreshToken) TableName() string {
--- a/internal/clientcache/internal/db/schema.sql
+++ b/internal/clientcache/internal/db/schema.sql
@@ -37,8 +37,6 @@ create table if not exists refresh_token(
     constraint only_known_resource_types_allowed,
   refresh_token text not null
     check (length(refresh_token) > 0),
-  update_time timestamp not null default (strftime('%Y-%m-%d %H:%M:%f','now')),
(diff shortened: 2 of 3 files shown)


internal/clientcache/internal/cache/options.go
```diff
@@ -42,6 +45,14 @@ func WithUpdateLastAccessedTime(b bool) Option {
 	}
 }
 
+// WithDuration provides an option for specifying a duration
```
johanbrandhorst, line 48: A duration of what? Without any context I don't know what this function does. Could you expand this doc string?

internal/clientcache/internal/cache/refresh_test.go
```diff
@@ -24,10 +25,42 @@ import (
 // testStaticResourceRetrievalFunc returns a function that always returns the
 // provided slice and a nil error. The returned function can be passed into the
 // options that provide a resource retrieval func such as
-// WithTargetRetrievalFunc and WithSessionRetrievalFunc.
-func testStaticResourceRetrievalFunc[T any](ret []T) func(context.Context, string, string) ([]T, error) {
-	return func(ctx context.Context, s1, s2 string) ([]T, error) {
-		return ret, nil
+// WithTargetRetrievalFunc and WithSessionRetrievalFunc.  The provided refresh
+// token determines the returned value and is an string representation of an
```
johanbrandhorst, line 29:
```suggestion
// token determines the returned value and is a string representation of an
```

internal/clientcache/internal/cache/repository_refresh_token.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/hashicorp/boundary/internal/util"
+	"github.com/hashicorp/go-dbw"
+)
+
+// RefreshTokenValue is the the type for the actual refresh token value handled
+// by the client cache.
+type RefreshTokenValue string
+
+// lookupRefreshToken returns the refresh token stored in the db unless it is
+// to old in which case it does not return a refresh token.
```
johanbrandhorst, line 21:
```suggestion
// too old in which case it does not return a refresh token.
```

johanbrandhorst: Does that mean that it returns an empty string? Or an error? Please expand this comment.

internal/clientcache/internal/cache/repository_targets.go
```diff
@@ -57,24 +60,37 @@ func (r *Repository) refreshTargets(ctx context.Context, u *user, tokens map[Aut
...
 
 	// Find and use a token for retrieving targets
 	var gotResponse bool
 	var resp []*targets.Target
+	var removedIds []string
+	var newRefreshToken RefreshTokenValue
 	var retErr error
 	for at, t := range tokens {
-		resp, err = opts.withTargetRetrievalFunc(ctx, u.Address, t)
+
```
johanbrandhorst, line 75: Rogue newline
```suggestion
```

internal/refreshtoken/refresh_token.go
```diff
@@ -101,32 +101,32 @@ func (rt *Token) Validate(
...
 	if rt.LastItemId == "" {
-		return errors.New(ctx, errors.InvalidParameter, op, "refresh token missing last item ID")
+		return errors.New(ctx, errors.InvalidRefreshToken, op, "refresh token missing last item ID")
 	}
 	if rt.LastItemUpdatedTime.After(time.Now()) {
-		return errors.New(ctx, errors.InvalidParameter, op, "refresh token last item was updated in the future")
+		return errors.New(ctx, errors.InvalidRefreshToken, op, "refresh token last item was updated in the future")
 	}
 	if rt.ResourceType != expectedResourceType {
-		return errors.New(ctx, errors.InvalidParameter, op, "refresh token resource type does not match expected resource type")
```
johanbrandhorst, line 129: Could you update the refresh token tests to check for this error code?

johanbrandhorst: Also note that we probably should add this code to the handlers.ParseRefreshToken function.

johanbrandhorst (changes requested):

johanbrandhorst (approved):

internal/clientcache/internal/db/schema.sql
```diff
@@ -13,6 +13,35 @@ create table if not exists user (
...
+create table if not exists refresh_token(
+  user_id text not null
+    references user(id)
+    on delete cascade,
+  resource_type text not null
+    references resource_type_enm(string)
+    constraint only_known_resource_types_allowed,
+  refresh_token text not null
+    check (length(refresh_token) > 0),
+  update_time timestamp not null default (strftime('%Y-%m-%d %H:%M:%f','now')),
```
jimlambrt, line 40: Do we want a trigger for these columns?  See https://github.com/hashicorp/go-kms-wrapping/blob/main/extras/kms/migrations/sqlite/02_version.up.sql

jimlambrt (changes requested):
Just one question.  ty!

jimlambrt (approved):
ty!

PR #3864, opened 2023-10-18
Migrate some credential bits around

jefferai:
This moves some credential definitions to globals and adds a few helpers to the authenticate command.


mgaffney (approved):

api/authmethods/authenticate.go
```diff
@@ -51,15 +53,33 @@ func (a *AuthenticateResult) UnmarshalJSON(inBytes []byte) error {
...
+	}
+	return token, nil
+}
+
+// Authenticate is a generic authenticate API call that returns a generic
+// result. See the documentation for the attributes required for any given auth
+// method.
+//
+// Only some auth methods support multiple commands. If the documentation does
+// not specify what command to use when and with which attriubutes, use
```
johanbrandhorst, line 81:
```suggestion
// not specify what command to use when and with which attributes, use
```

jefferai: Fixed in https://github.com/hashicorp/boundary/pull/3865 !

PR #3865, opened 2023-10-19
Add some useful dev flags

jefferai:


internal/cmd/commands/dev/dev.go
```diff
@@ -373,6 +375,18 @@ func (c *Command) Flags() *base.FlagSets {
 		EnvVar: "BOUNDARY_DEV_PLUGIN_EXECUTION_DIR",
 		Usage:  "Specifies where Boundary should write plugins that it is executing; if not set defaults to system temp directory.",
 	})
+	f.BoolVar(&base.BoolVar{
+		Name:   "skip-plugins",
+		Target: &c.flagSkipPlugins,
+		Usage:  "Skip loading compiled-in plugins. This does not prevent loopback plugins from being loaded if enabled.",
+		Hidden: true,
```
jefferai, line 382: Note that these are hidden. These are not intended for end users, but are useful when trying to compile and test on windows/arm64.

jefferai:
No idea what's up with the linter.

internal/daemon/worker/proxy/options.go
```diff
@@ -51,3 +52,11 @@ func WithPostConnectionHook(fn func(net.Conn)) Option {
 		o.WithPostConnectionHook = fn
 	}
 }
+
+// WithDnsServer allows specifying lookup of the endpoint to happen via an
+// alternate DNS server. Must be in host:port form.
+func WithDnsServer(with string) Option {
```
johanbrandhorst, line 58: To me a `host:port` combination is an address, mind naming this as such to maybe help a bit?
```suggestion
func WithDnsServerAddress(with string) Option {
```

internal/daemon/worker/proxy/proxydialer.go
```diff
@@ -19,19 +22,70 @@ var GetEndpointDialer = directDialer
...
+			host, port, err := net.SplitHostPort(endpoint)
+			if err != nil {
+				return nil, errors.Wrap(ctx, err, op, errors.WithMsg("error splitting host/port"))
+			}
+			_, err = netip.ParseAddr(host)
+			if err == nil || host == "localhost" {
+				// It's an IP address or something we definitely can't resolve,
+				// continue
+				goto dial
+			}
```
johanbrandhorst, line 48: Could we invert this if statement and remove the goto?
```suggestion
			// It's an IP address or something we definitely can't resolve,
			// continue
			if err != nil && host != "localhost" {
				dnsClient := new(dns.Client)
				...
			}
```

internal/daemon/worker/proxy/proxydialer.go
```diff
@@ -19,19 +22,70 @@ var GetEndpointDialer = directDialer
...
+			dnsClient := new(dns.Client)
+			dnsClient.Net = parsedDnsServer.Scheme
```
johanbrandhorst, line 50: In the `WithDnsServer` comment it says it expects the server in `host:port` form, but here we seem to use the scheme. Should it be `scheme://host:port`?

internal/daemon/worker/proxy/proxydialer.go
```diff
@@ -19,19 +22,70 @@ var GetEndpointDialer = directDialer
...
 	}
-	d, err := NewProxyDialer(ctx, func(opt ...Option) (net.Conn, error) {
+	opts := GetOpts(opt...)
+	dnsServer := opts.WithDnsServer
+	parsedDnsServer, err := url.Parse(dnsServer)
+	if err != nil {
+		return nil, errors.Wrap(ctx, err, op)
+	}
+
+	d, err := NewProxyDialer(ctx, func(dialerOpt ...Option) (net.Conn, error) {
```
johanbrandhorst, line 37: Does this function benefit from being a closure? Wonder if it'd be slightly easier to read if it was a named function

internal/daemon/worker/proxy/proxydialer.go
```diff
@@ -19,19 +22,70 @@ var GetEndpointDialer = directDialer
...
+			m.RecursionDesired = true
+			m.Question = make([]dns.Question, 1)
+			m.Question[0] = dns.Question{
+				Name:   fmt.Sprintf("%s.", host),
+				Qtype:  dns.TypeA,
+				Qclass: dns.ClassINET,
+			}
+			resp, _, err := dnsClient.Exchange(m, parsedDnsServer.Host)
+			if err != nil {
+				return nil, errors.Wrap(ctx, err, op, errors.WithMsg("error splitting host/port"))
```
johanbrandhorst, line 62: I think this error message is wrong

internal/host/plugin/job_set_sync.go
```diff
@@ -54,7 +54,7 @@ func newSetSyncJob(ctx context.Context, r db.Reader, w db.Writer, kms *kms.Kms,
 		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing db.Writer")
 	case kms == nil:
 		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing kms")
-	case len(plgm) == 0:
+	case plgm == nil:
```
johanbrandhorst, line 57: Is this change correct? Is an empty map ok?

johanbrandhorst (changes requested):

johanbrandhorst:
> No idea what's up with the linter.

We've had some caching issues recently, might work on a rerun? I triggered it again.

internal/daemon/worker/proxy/options.go
jefferai: Will make that change but do it in a separate commit so I update the comment too.

internal/daemon/worker/proxy/proxydialer.go
jefferai: Good catch, fixed

internal/daemon/worker/proxy/proxydialer.go
jefferai: Fixed

internal/host/plugin/job_set_sync.go
jefferai: Yeah -- I checked this with Todd. The intent of the check is to ensure we don't access a nil pointer; if there are no plugins then it shouldn't fail, it should have nothing to do.

internal/daemon/worker/proxy/proxydialer.go
jefferai: Maybe, but it's not a change from this PR and I want to keep it scoped to the new stuff.

internal/daemon/worker/proxy/proxydialer.go
jefferai: Sure, I can

johanbrandhorst:
> > No idea what's up with the linter.
> 
> We've had some caching issues recently, might work on a rerun? I triggered it again.

Nah the linter is broken, no idea what's up

johanbrandhorst (approved):

PR #3871, opened 2023-10-19
Add status command to the boundary client cache

talanknight:
Sample output:

```
> boundary daemon status

Status:
  Domain Socket:       /Users/myname/.boundary/socket/daemon.sock
  Uptime:              8m9s
  User Count:          2

  User:
    Address:               http://127.0.0.1:9200
    AuthToken Count:       1
    Id:                    u_1234567890
    AuthToken:
      Id:                      at_YdVyvcFO0h
      In Keyring:              true
    Target:
      Count:                    2
      Since Full Refresh:       8m9s
      Since Last Refresh:       10s
    Session:
      Count:                   0
  User:
    Address:               http://127.0.0.1:9200
    AuthToken Count:       1
    Id:                    u_0987654321
    AuthToken:
      Id:                      at_l6vMxulkKj
    Target:
      Count:                    2
      Since Full Refresh:       55s
      Since Last Refresh:       10s
    Session:
      Count:                   0
```

### DIFF
--- a/internal/clientcache/cmd/daemon/status.go
+++ b/internal/clientcache/cmd/daemon/status.go
@@ -14,7 +14,6 @@ import (
 	"github.com/hashicorp/boundary/api"
 	"github.com/hashicorp/boundary/internal/clientcache/internal/daemon"
 	"github.com/hashicorp/boundary/internal/cmd/base"
-	"github.com/hashicorp/boundary/internal/errors"
 	"github.com/hashicorp/boundary/version"
 	"github.com/mitchellh/cli"
 	"github.com/posener/complete"
@@ -97,7 +96,6 @@ func (c *StatusCommand) Run(args []string) int {
 }
 
 func (c *StatusCommand) Status(ctx context.Context) (*api.Response, error) {
-	const op = "daemon.(StatusCommand).Status"
 	dotPath, err := DefaultDotDirectory(ctx)
 	if err != nil {
 		return nil, err
@@ -107,10 +105,9 @@ func (c *StatusCommand) Status(ctx context.Context) (*api.Response, error) {
 }
 
 func status(ctx context.Context, daemonPath string) (*api.Response, error) {
-	const op = "daemon.status"
 	client, err := api.NewClient(nil)
 	if err != nil {
-		return nil, errors.Wrap(ctx, err, op)
+		return nil, fmt.Errorf("Error creating a new API client: %w", err)
 	}
 	addr := daemon.SocketAddress(daemonPath)
 	_, err = os.Stat(strings.TrimPrefix(addr, "unix://"))
@@ -118,7 +115,7 @@ func status(ctx context.Context, daemonPath string) (*api.Response, error) {
 		return nil, errDaemonNotRunning
 	}
 	if err := client.SetAddr(addr); err != nil {
-		return nil, errors.Wrap(ctx, err, op)
+		return nil, fmt.Errorf("Error when setting the client's address: %w", err)
 	}
 	// Because this is using the real lib it can pick up from stored locations
 	// like the system keychain. Explicitly clear the token for now
--- a/internal/clientcache/internal/daemon/status_handler.go
+++ b/internal/clientcache/internal/daemon/status_handler.go
@@ -16,19 +16,19 @@ import (
 
 // RefreshTokenStatus is the status of a resource token
 type RefreshTokenStatus struct {
-	Age      time.Duration `json:"age"`
-	LastUsed time.Duration `json:"last_used"`
+	Age      time.Duration `json:"age,omitempty"`
+	LastUsed time.Duration `json:"last_used,omitempty"`
 }
 
 type ErrorStatus struct {
-	Error        string        `json:"error"`
-	LastReturned time.Duration `json:"last_returned"`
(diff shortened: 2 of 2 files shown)


jimlambrt (approved):
ty!

internal/clientcache/internal/daemon/status_handler.go
```diff
@@ -0,0 +1,155 @@
...
+
+type ErrorStatus struct {
+	Error        string
+	LastReturned time.Duration
+}
+
+// ResourceStatus contains the status of a specific resource type contained in
+// the cache for a specific user.
+type ResourceStatus struct {
+	Name         string
```
jefferai, line 31: Please add `json` tags to the various structs here and anywhere else we are marshaling as the default JSON output does not match our standard casing.

internal/clientcache/internal/daemon/status_handler.go
```diff
@@ -0,0 +1,155 @@
...
+	// The boundary address for this user
+	Address string
+	// The auth tokens used by this user to authenticate with the boundary instance
+	AuthTokens []AuthTokenStatus
+	// The resources tracked by the cache for this user
+	Resources []ResourceStatus
+}
+
+// StatusResult is the struct returned to status requests.
+type StatusResult struct {
```
jefferai, line 58: Given that there will never be anything marshaled as JSON do you even need the tags here at all?

jefferai: Oops, sorry, `,omitempty` not `-,omitempty`. But please fix the names :-)

internal/clientcache/cmd/daemon/status.go
```diff
@@ -0,0 +1,214 @@
...
+	case "json":
+		c.UI.Output(resp.Body.String())
+	default:
+		c.UI.Output(printStatusTable(res))
+	}
+	return base.CommandSuccess
+}
+
+func (c *StatusCommand) Status(ctx context.Context) (*api.Response, error) {
+	const op = "daemon.(StatusCommand).Status"
```
jefferai, line 100: This const is unused

internal/clientcache/cmd/daemon/status.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	return status(ctx, dotPath)
+}
+
+func status(ctx context.Context, daemonPath string) (*api.Response, error) {
+	const op = "daemon.status"
+	client, err := api.NewClient(nil)
+	if err != nil {
+		return nil, errors.Wrap(ctx, err, op)
```
jefferai, line 113: These errors will bubble up to printing directly to CLI. We don't return raw errors for CLI commands; they are human readable, capitalized with punctuation (but can also include raw error strings). For example: `c.PrintCliError(fmt.Errorf("Error creating API client: %w", err))`

jefferai: Also we don't include the op anywhere in the CLI at this point. Maybe we should but I don't think we should introduce that pattern here.

internal/clientcache/cmd/daemon/status.go
```diff
@@ -0,0 +1,214 @@
...
+	addr := daemon.SocketAddress(daemonPath)
+	_, err = os.Stat(strings.TrimPrefix(addr, "unix://"))
+	if strings.HasPrefix(addr, "unix://") && err != nil {
+		return nil, errDaemonNotRunning
+	}
+	if err := client.SetAddr(addr); err != nil {
+		return nil, errors.Wrap(ctx, err, op)
+	}
+	// Because this is using the real lib it can pick up from stored locations
+	// like the system keychain. Explicitly clear the token for now
```
jefferai, line 124: I'm a bit confused by this. Why are we using the Boundary API client? Does it actually make sense for making simple calls across a unix socket?

internal/clientcache/internal/daemon/status_handler.go
```diff
@@ -0,0 +1,155 @@
...
+	KeyringlessReferences int    `json:"keyringless_references,omitempty"`
+}
+
+// UserStatus contains the status of a specific user tracked by the cache
+type UserStatus struct {
+	// The Id of the user this status is for
+	Id string `json:"id"`
+	// The boundary address for this user
+	Address string `json:"address"`
+	// The auth tokens used by this user to authenticate with the boundary instance
```
jefferai, line 51: Curious in this file generally why some things are omitempty and some aren't. Is it that the ones not marked omitempty are never expected to be optional?

jefferai (commented):
Looking pretty good, just a few more comments

internal/clientcache/internal/daemon/status_handler.go
```diff
@@ -0,0 +1,155 @@
...
+type ErrorStatus struct {
+	Error        string        `json:"error,omitempty"`
+	LastReturned time.Duration `json:"last_returned,omitempty"`
+}
+
+// ResourceStatus contains the status of a specific resource type contained in
+// the cache for a specific user.
+type ResourceStatus struct {
+	Name         string              `json:"name,omitempty"`
+	Count        int                 `json:"count"`
```
jefferai, line 32: Missing `omitempty`?

jefferai (approved):
Left one more comment!

PR #3875, opened 2023-10-20
feat(worker): Worker Local Storage State DB Column

elimt:
- Add new `local_storage_state` column to `server_worker` table which keeps track of the state of the local disk space.

- The valid values for `local_storage_state` are: 
-- `available`: The worker local storage state is at an acceptable state
-- `lowStorage`: The worker is below the minimum threshold for local storage
-- `outOfStorage`: The worker is out of local disk space
-- `notConfigured`: The worker does not have a local storage path configured
-- `unknown`: The default local storage state of a worker. Used when the local storage state of a worker is not yet known

- Worker proto definition updates needed for new `local_storage_state` field which represents the local disk space of the worker.


internal/db/schema/migrations/oss/postgres/78/01_server_worker_local_storage_state.up.sql
```diff
@@ -0,0 +1,86 @@
...
+insert into server_worker_local_storage_state_state_enm (state) values
+  ('available'),
+  ('lowStorage'),
+  ('criticallyLow'),
+  ('outOfStorage'),
+  ('notConfigured'),
+  ('unknown');
+
+alter table server_worker
+  add column local_storage_state text not null default 'available'
```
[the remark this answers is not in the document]
elimt: I was trying to decide on the best possible default value for this when we upgrade a Boundary setup and then migration happens.

Since we decided on only picking workers for session recording that have an "available" local storage state, workers running on 0.13 will start off as "not configured" or "unknown" and will never be updated since nothing will be reporting the local storage state. 

In order not to break older workers/setups I wasn't too sure what the default should be. I left it as "available" so as not to break setups after migration.

internal/db/schema/migrations/oss/postgres/34/04_views.up.sql
```diff
@@ -5,7 +5,7 @@ begin;
 
 -- worker_aggregate view allows the worker and configuration to be read at the
 -- same time.
--- Updated in 51/01_server_worker_release_version.up.sql
```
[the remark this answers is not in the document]
elimt: I saw a pattern where when a table/view is updated, the older references are also updated. Do I only have to update the most recent references and leave the older references?

internal/db/schema/migrations/oss/postgres_78_01_test.go
```diff
@@ -0,0 +1,104 @@
...
+				Name:                  "oss",
+				BinarySchemaVersion:   currentMigration,
+				DatabaseSchemaVersion: currentMigration,
+				DatabaseSchemaState:   schema.Equal,
+			},
+		},
+	}
+	require.Equal(want, state)
+
+	// Assert worker has been migrated and received default state of 'available'
```
tmessi, line 96: This comment seems to be incorrect? Or the assertion below is incorrect?

tmessi (changes requested):
Should this PR be against an llb instead of main?

internal/db/schema/migrations/oss/postgres_78_01_test.go
elimt: The comment is incorrect. Forgot to update that after changing the default value. Updated now. Thanks!

elimt: Also updated the target branch to `llb-worker-local-storage-state`

internal/db/sqltest/Makefile
```diff
@@ -32,7 +32,8 @@ TESTS ?= tests/setup/*.sql \
 		 tests/storage/*.sql \
 		 tests/domain/*.sql \
 		 tests/history/*.sql \
-		 tests/recording/*.sql
+		 tests/recording/*.sql \
+		 tests/worker/*.sql \
```
ddebko, line 36: Remove the last `\` on last line

internal/server/options.go
```diff
@@ -267,3 +269,10 @@ func WithWorkerPool(workerIds []string) Option {
 		o.withWorkerPool = workerIds
 	}
 }
+
+// WithLocalStorageState provides an optional local storage state.
+func WithLocalStorageState(state string) Option {
```
ddebko, line 274: We should add a unit test for this option in the internal/server/options_test.go file.

internal/server/repository_worker.go
```diff
@@ -597,6 +597,7 @@ func (r *Repository) CreateWorker(ctx context.Context, worker *Worker, opt ...Op
 	}
 
 	worker.OperationalState = UnknownOperationalState.String()
+	worker.LocalStorageState = UnknownLocalStorageState.String()
```
ddebko, line 600: We should add a unit test in the internal/server/repository_worker_test.go file for the `TestRepository_CreateWorker` method. Something similar to the `success-with-state` test case.

internal/server/worker.go
```diff
@@ -108,12 +123,13 @@ func NewWorker(scopeId string, opt ...Option) *Worker {
...
-			Address:          opts.withAddress,
-			ReleaseVersion:   opts.withReleaseVersion,
-			OperationalState: opts.withOperationalState,
+			ScopeId:           scopeId,
+			Name:              opts.withName,
+			Description:       opts.withDescription,
+			Address:           opts.withAddress,
+			ReleaseVersion:    opts.withReleaseVersion,
+			OperationalState:  opts.withOperationalState,
+			LocalStorageState: opts.withLocalStorageState,
```
ddebko, line 132: We should add a new unit test for the internal/server/worker_test.go file to validate the LocalStorageState. We can add a test case to the `TestWorker_New` method.

ddebko (changes requested):

ddebko (approved):

internal/proto/controller/storage/servers/store/v1/worker.proto
```diff
@@ -65,6 +65,16 @@ message Worker {
   // The state of the worker, to indicate if the worker is active or in shutdown.
   // @inject_tag: `gorm:"not_null"`
   string operational_state = 150;
+
+  // The local_storage_state indicates the state of the local disk space of the worker.
+  // Possible values are:
```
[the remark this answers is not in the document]
elimt: Yes. I just updated it. Thanks!

internal/server/repository_worker.go
```diff
@@ -597,6 +597,15 @@ func (r *Repository) CreateWorker(ctx context.Context, worker *Worker, opt ...Op
...
+
+	if opts.withLocalStorageState != "" {
```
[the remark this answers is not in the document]
elimt: Okay I just left a comment about that. Thanks for the review!

PR #3883, opened 2023-10-23
Add skip daemon flag and send recently authed tokens to daemon

talanknight:
Added the ability to skip any daemon specific logic in the client cache command wrapper by using a flag and adding auth tokens retrieved but not stored to the client daemon.

### DIFF
--- a/internal/clientcache/cmd/daemon/stop_windows.go
+++ b/internal/clientcache/cmd/daemon/stop_windows.go
@@ -69,12 +69,15 @@ func (s *StopCommand) stop(ctx context.Context) error {
 func stopThroughHandler(ctx context.Context, dotPath string) (*api.Error, error) {
 	const op = "daemon.stopThroughHandler"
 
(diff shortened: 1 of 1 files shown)


internal/clientcache/cmd/daemon/command_wrapper.go
```diff
@@ -64,10 +77,56 @@ func (w *CommandWrapper) startDaemon(ctx context.Context) bool {
...
+// provided context is done. It returns an error if the unix socket is not found
+// before the context is done.
+func waitForDaemon(ctx context.Context) error {
+	dotPath, err := DefaultDotDirectory(ctx)
+	if err != nil {
+		return err
+	}
+	timer := time.NewTimer(0)
+
+	addr := daemon.SocketAddress(dotPath)
```
jefferai, line 119: Just a note that it may be nicer to pass around the address as a `*url.URL` and reference e.g. `Path`, depending on how often something like this has to be done.

jefferai (approved):
Looks good!

Issue #3881, opened 2023-10-22
CPU burn / slow >= v0.14.0


guyguy333:
**Describe the bug**

Running Hashicorp Boundary in a container (at least) burns CPU and each API request from UI is slow (~1s).

![Screenshot 2023-10-22 at 11 42 22](https://github.com/hashicorp/boundary/assets/4040959/eb6117a1-fdc0-4013-9679-6ac2671bd66b)

**To Reproduce**

Steps to reproduce the behavior in dev mode:
1. Order a fresh Linux VM with Ubuntu 22.04 LTS (AMD EPYC 4 cores, 16Gb RAM)
2. Install Docker
3. Fix permissions for the test (horrible fix, but it's just to reproduce) : sudo chmod 777 /var/run/docker.sock
4. Run boundary dev mode : `docker run --net=host -v /var/run/docker.sock:/var/run/docker.sock --rm hashicorp/boundary:latest dev`
5. Boundary starts and then CPU burns

Steps to reproduce the behavior in dev mode with external Postgres 15 (another machine):
1. Order a fresh Linux VM with Ubuntu 22.04 LTS (AMD EPYC 4 cores, 16Gb RAM)
2. Install Docker
4. Run boundary dev mode : `docker run --rm hashicorp/boundary:latest dev -database-url=XXXXX`
5. Boundary starts (but is really slow, it's about few minutes) and then CPU burns

Steps to reproduce the behavior in production mode with external Postgres 15 (another machine):
1. Order a fresh Linux VM with Ubuntu 22.04 LTS (AMD EPYC 4 cores, 16Gb RAM)
2. Install Docker
4. Run boundary mode with production config enabling worker
5. Boundary starts (but is really slow) and then CPU burn. Each API request from UI is about 1s

However, in production mode, I found that disabling worker, stops CPU burning but API requests are still really slow from UI.

If I run in dev mode on my laptop (Apple M1 PRO) without Docker, I don't have CPU burn issue and everything is fast. If I run on VMs without container, I also have the issue (Linux Ubuntu 22.04 LTS related issue ?)

It has been tested on two different cloud providers for VMs, both running Ubuntu 22.04 LTS 

I don't have the CPU burn issue with v0.13.1 but I've the issue with v0.14.0 and v0.14.1. However, I've really slow API requests and so slow UI with v0.13.1.

**Expected behavior**

Hashicorp boundary is responsive and no longer burn CPU.

**Additional context**

elimt:
[...]

PR #3884, opened 2023-10-23
fix: monitorUpstreamConnectionState CPU consumption

elimt:
Bug found in https://github.com/hashicorp/boundary/issues/3881

`monitorUpstreamConnectionState()` is a goroutine that listens for GRPC client connection changes using GRPC's `WaitForStateChange` method. 

`monitorUpstreamConnectionState()` is currently consuming the a lot of CPU. It seems to be continuously running and not actually waiting for state changes.

Reference for GRPC Connection State Transitions: https://grpc.github.io/grpc/core/md_doc_connectivity-semantics-and-api.html

`state` was never getting updated so the default state was never getting updated which was causing the loop not to wait for new state chanegs

**Before Fix**

image

**After Fix**

image

### DIFF
--- a/internal/daemon/worker/controller_connection.go
+++ b/internal/daemon/worker/controller_connection.go
@@ -434,7 +434,7 @@ func monitorUpstreamConnectionState(ctx context.Context, cc *grpc.ClientConn, co
 	}
 
 	for cc.WaitForStateChange(ctx, state) {
-		state := cc.GetState()
+		state = cc.GetState()
 
 		// if the client is shutdown, exit function
 		if state == connectivity.Shutdown {
@@ -442,7 +442,5 @@ func monitorUpstreamConnectionState(ctx context.Context, cc *grpc.ClientConn, co
 		}
 
 		connectionState.Store(state)
-
-		time.Sleep(10 * time.Millisecond)
 	}
 }
--- a/internal/daemon/worker/controller_connection.go
+++ b/internal/daemon/worker/controller_connection.go
@@ -442,5 +442,7 @@ func monitorUpstreamConnectionState(ctx context.Context, cc *grpc.ClientConn, co
 		}
 
 		connectionState.Store(newState)
+
+		time.Sleep(10 * time.Millisecond)
 	}
 }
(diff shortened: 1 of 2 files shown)


internal/daemon/worker/controller_connection.go
```diff
@@ -442,5 +442,7 @@ func monitorUpstreamConnectionState(ctx context.Context, cc *grpc.ClientConn, co
 		}
 
 		connectionState.Store(newState)
+
+		time.Sleep(10 * time.Millisecond)
```
johanbrandhorst, line 446: Is this intended to be a long term solution or just a hotfix that we can get out quickly to users? It feels like we need to spend more time digging into the source of the issue. Why is the state changing so often? It could indicate something about the application that's worrying.

elimt: Pushed another fix which should address the issue. Took out sleep change

internal/daemon/worker/controller_connection.go
```diff
@@ -434,13 +434,13 @@ func monitorUpstreamConnectionState(ctx context.Context, cc *grpc.ClientConn, co
 	}
 
 	for cc.WaitForStateChange(ctx, state) {
-		newState := cc.GetState()
+		state := cc.GetState()
```
johanbrandhorst, line 437: I'm confused, how does this fix the issue? This code looks functionally identical to the old code. Is `state` supposed to overwrite the variable in the outer state? If so, I think you want
```suggestion
		state = cc.GetState()
```

johanbrandhorst (changes requested):

internal/daemon/worker/controller_connection.go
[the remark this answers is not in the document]
elimt: Updated the `newState` to `state`. That fixed the issue

internal/daemon/worker/controller_connection.go
elimt: Yes my bad. I pushed an older code change. 

The new update should be `state = cc.GetState()` now

johanbrandhorst (approved):
Thanks! I expect it will be hard to test this, but if you find some way, please do!

moduli (approved):
Change makes sense to me. Confirmed that running `boundary dev` against this branch does not show boundary running at 100% CPU.

ajayreshc (approved):
LGTM, as `state` never became `newState` even when it changed. If the zero value of `state` (Idle I believe) didn't match getState() then the result is a tight loop that never resolves itself, because it would always fall out in `WaitForStateChange()` without polling the channel for a new value.

PR #3885, opened 2023-10-23
List pagination schema changes

johanbrandhorst:
## [internal/db: add deleted tables for all resources](https://github.com/hashicorp/boundary/commit/3ea48bcc4c448c70b5b9f8594762ee496ce6f46d)

Creates a new foo_deleted table for each subtype of each resource that needs to support pagination. The deleted
tables are used to track deletions of resources and return them to users of the new pagination API. We track only the ID and delete_time of the resource, so there's no way to filter out resources that the user could not see before it was deleted. This is intentional to keep the design simple and is not considered a security issue.

### DIFF
--- a/internal/db/sqltest/tests/purge/deleted_table_tests.sql
+++ b/internal/db/sqltest/tests/purge/deleted_table_tests.sql
@@ -1,0 +1,86 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+  create function op_table(deletion_table_name name) returns text
+  as $$
+    select split_part(deletion_table_name, '_deleted', 1);
+  $$ language sql;
+
+  create function has_operational_table(deletion_table_name name) returns text
+  as $$
+    select has_table(op_table(deletion_table_name));
+  $$ language sql;
+
+  -- tests that the deletion table has the insert trigger
+  create function has_insert_trigger(deletion_table_name name) returns text
+  as $$
+    select * from collect_tap(
+      has_trigger(op_table(deletion_table_name), 'insert_deleted_id'),
+      trigger_is(op_table(deletion_table_name), 'insert_deleted_id', 'insert_deleted_id')
+    );
+  $$ language sql;
+
+  -- tests the public_id column
+  create function has_public_id(deletion_table_name name) returns text
+  as $$
+    select * from collect_tap(
+      has_column(deletion_table_name, 'public_id'),
+      col_not_null(deletion_table_name, 'public_id'),
+      col_hasnt_default(deletion_table_name, 'public_id')
+    );
+  $$ language sql;
+
+  -- tests the delete_time column
+  create function has_delete_time(deletion_table_name name) returns text
+  as $$
+    select * from collect_tap(
+      has_column(deletion_table_name, 'delete_time'),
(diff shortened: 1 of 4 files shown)


johanbrandhorst:
Note that the schema diff job seems to break because this diff is too large 😬. Hopefully it's still easy enough to read without the diff.

tmessi:
> Note that the schema diff job seems to break because this diff is too large 😬. Hopefully it's still easy enough to read without the diff.

See: https://github.com/hashicorp/boundary/pull/3909

tmessi:
> Note that the schema diff job seems to break because this diff is too large 😬. Hopefully it's still easy enough to read without the diff.

I would also note that since this is all new tables and functions, the diff isn't as useful. Although it would help catch if something was accidentally being altered.

jimlambrt (approved):

tmessi (changes requested):
Can you update `internal/db/sqltests/tests/Makefile` so the new tests run in CI? They are failing for me when I run them manually.

You need to add the new directory to the list of tests here: 

https://github.com/hashicorp/boundary/blob/752f20fa276f029d1f7d5a36820e9929abc0446f/internal/db/sqltest/Makefile#L20-L35

internal/db/sqltest/tests/purge/trigger_insert_on_delete.sql
```diff
@@ -0,0 +1,41 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+
+  select plan(13);
+
```
mgaffney, line 7:
```suggestion

```
Delete trailing whitespace

internal/db/sqltest/tests/purge/trigger_insert_on_delete.sql
```diff
@@ -0,0 +1,41 @@
...
+  -- To test the trigger that moves deleted rows into their appropriate tables, we'll use auth_token
+  -- Ensure session state table is populated, and session_deleted and auth_token_deleted tables are empty
+  select is(count(*), 1::bigint) from session_state where session_id = 's1_____clare';
+  select is(count(*), 0::bigint) from session_deleted;
+  select is(count(*), 0::bigint) from auth_token_deleted;
+
+  -- Check that we have a session for a particular auth token
+  select is(count(*), 2::bigint) from session where auth_token_id = 'tok____clare';
+
```
mgaffney, line 22:
```suggestion

```
Delete trailing whitespace

mgaffney (approved):

johanbrandhorst:
@tmessi thanks for spotting that, looks like we missed the renaming of the trigger.

internal/db/sqltest/tests/purge/deleted_table_tests.sql
```diff
@@ -0,0 +1,86 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+  create function op_table(deletion_table_name name) returns text
+  as $$
+    select split_part(deletion_table_name, '_deleted', 1);
+  $$ language sql;
```
tmessi, line 8:
```suggestion
  $$ language sql;
```

tmessi (approved):

PR #3887, opened 2023-10-23
Add Now method to DB handle

johanbrandhorst:
The Now() method returns the current transaction timestamp.

### DIFF
--- a/internal/db/read_writer.go
+++ b/internal/db/read_writer.go
@@ -70,6 +70,11 @@ type Reader interface {
 
 	// ScanRows will scan sql rows into the interface provided
 	ScanRows(ctx context.Context, rows *sql.Rows, result any) error
+
+	// Now returns the current transaction timestamp. Now will return the same
+	// timestamp whenever it is called within a transaction. In other words, calling
+	// Now at the start and at the end of a transaction will return the same value.
+	Now(ctx context.Context) (time.Time, error)
 }
 
 // Writer interface defines create, update and retryable transaction handlers
@@ -531,6 +536,27 @@ func (rw *Db) SearchWhere(ctx context.Context, resources any, where string, args
 		return wrapError(ctx, err, op)
 	}
 	return nil
+}
+
+// Now returns the current transaction timestamp. Now will return the same
+// timestamp whenever it is called within a transaction. In other words, calling
+// Now at the start and at the end of a transaction will return the same value.
+func (rw *Db) Now(ctx context.Context) (time.Time, error) {
+	const op = "db.(*Db).Now"
+	// The Postgres docs define the different pre-defined time variables available:
+	// https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT.
+	// The value produced by this function is equivalent to current_timestamp.
+	rows, err := rw.Query(ctx, "select now()", nil)
+	if err != nil {
+		return time.Time{}, errors.Wrap(ctx, err, op, errors.WithMsg("failed to query current timestamp"))
+	}
+	var now time.Time
+	for rows.Next() {
+		if err := rw.ScanRows(ctx, rows, &now); err != nil {
+			return time.Time{}, errors.Wrap(ctx, err, op, errors.WithMsg("failed to query current timestamp"))
+		}
+	}
+	return now, nil
 }
 
 func isNil(i any) bool {
--- a/internal/db/read_writer_test.go
+++ b/internal/db/read_writer_test.go
@@ -3193,3 +3193,37 @@ func TestDb_oplogMsgsForItems(t *testing.T) {
 // 		})
 // 	}
 // }
+
+func TestDb_Now(t *testing.T) {
+	t.Parallel()
+	assert, require := assert.New(t), require.New(t)
+	conn, _ := TestSetup(t, "postgres")
+	rw := New(conn)
+	ctx := context.Background()
+	now, err := rw.Now(ctx)
+	require.NoError(err)
+
+	now2, err := rw.Now(ctx)
+	require.NoError(err)
+
+	// Calling Now() outside a transaction should result in different timestamps.
+	// Using time.Equal instead of assert.Equal since it's more consistently
+	// accurate for time.Time structs.
+	assert.False(now.Equal(now2))
+
+	_, err = rw.DoTx(ctx, StdRetryCnt, ExpBackoff{}, func(r Reader, _ Writer) error {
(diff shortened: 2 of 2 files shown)


internal/db/read_writer_test.go
```diff
@@ -3193,3 +3193,16 @@ func TestDb_oplogMsgsForItems(t *testing.T) {
...
+	t.Parallel()
+	conn, _ := TestSetup(t, "postgres")
+	rw := New(conn)
+	ctx := context.Background()
+	now, err := rw.Now(ctx)
+	require.NoError(t, err)
+	// Check that it's within 1 second of now according to the system
+	// If this is flaky... just increase the limit 😬.
+	assert.True(t, now.Before(time.Now().Add(time.Second)))
+	assert.True(t, now.After(time.Now().Add(-time.Second)))
```
mgaffney, line 3207: Can we test calling `Now` multiple times within a single transaction and then multiple calls not within a transaction?
* When in the same transaction, `Now` should return identical `time.Time` values.
* When not in a transaction, `Now` should return different `time.Time` values.

mgaffney (changes requested):

internal/db/read_writer_test.go
johanbrandhorst: Thanks, done!

internal/db/read_writer.go
```diff
@@ -533,6 +536,22 @@ func (rw *Db) SearchWhere(ctx context.Context, resources any, where string, args
 	return nil
 }
 
+// Now returns the current transaction timestamp.
```
jimlambrt, line 539: Mind adding a short explanation to the comment: `Now(...) will return the same time whenever it's called during the same transaction.  In other words, Now(...) at the start and Now(...) at the end of a transaction will always be equal.

jimlambrt (approved):

internal/db/read_writer.go
```diff
@@ -533,6 +536,22 @@ func (rw *Db) SearchWhere(ctx context.Context, resources any, where string, args
...
+func (rw *Db) Now(ctx context.Context) (time.Time, error) {
+	const op = "db.(*Db).Now"
+	rows, err := rw.Query(ctx, "select current_timestamp", nil)
```
tmessi, line 542: It might be good to provide a link to the [postgres docs](https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT) since they provide a good explanation of this and the alternatives. 

I would also note that we have tended to use `now()` elsewhere in the code so it might be good to do that here as well.

tmessi (approved):

internal/db/read_writer_test.go
```diff
@@ -3193,3 +3193,35 @@ func TestDb_oplogMsgsForItems(t *testing.T) {
...
+
+	now2, err := rw.Now(ctx)
+	require.NoError(t, err)
+
+	// Calling Now() outside a transaction should result in different timestamps
+	assert.False(t, now.Equal(now2))
+
+	_, err = rw.DoTx(ctx, StdRetryCnt, ExpBackoff{}, func(r Reader, w Writer) error {
```
mgaffney, line 3215:
```suggestion
	_, err = rw.DoTx(ctx, StdRetryCnt, ExpBackoff{}, func(r Reader, _ Writer) error {
```
`w` is not used

internal/db/read_writer_test.go
```diff
@@ -3193,3 +3193,35 @@ func TestDb_oplogMsgsForItems(t *testing.T) {
...
 // 	}
 // }
+
+func TestDb_Now(t *testing.T) {
+	t.Parallel()
+	conn, _ := TestSetup(t, "postgres")
+	rw := New(conn)
+	ctx := context.Background()
+	now, err := rw.Now(ctx)
+	require.NoError(t, err)
```
mgaffney, line 3203:
```suggestion
	require.NoError(err)
```

mgaffney (changes requested):
Some minor change requests to the test to make it more inline with our other tests.

internal/db/read_writer_test.go
```diff
@@ -3193,3 +3193,35 @@ func TestDb_oplogMsgsForItems(t *testing.T) {
...
+	// Check that it's within 1 second of now according to the system
+	// If this is flaky... just increase the limit 😬.
+	assert.True(t, now.Before(time.Now().Add(time.Second)))
+	assert.True(t, now.After(time.Now().Add(-time.Second)))
+
+	now2, err := rw.Now(ctx)
+	require.NoError(t, err)
+
+	// Calling Now() outside a transaction should result in different timestamps
+	assert.False(t, now.Equal(now2))
```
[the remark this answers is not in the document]
johanbrandhorst: `testify` [falls](https://github.com/stretchr/testify/blob/f97607b89807936ac4ff96748d766cf4b9711f78/assert/assertions.go#L801) [back to](https://github.com/stretchr/testify/blob/f97607b89807936ac4ff96748d766cf4b9711f78/assert/assertions.go#L65) [`reflect.DeepEqual`](https://pkg.go.dev/reflect#DeepEqual) for equality comparisons, and this is not a reliable way to compare timestamps for equality since it performs field by field comparisons. See https://stackoverflow.com/questions/75824869/reflect-deepequal-not-working-on-timestamp-comparison, https://wcsiu.com/2018/07/06/go-reflect-deep-equal-pitfall.html, etc. I make a practice of using `time.Equal` whenever I need to do timestamp comparisons, even in cases where it doesn't _really_ matter.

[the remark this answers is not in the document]
mgaffney: I agree that using `reflect.DeepEqual` to compare instances of `time.Time` that have been created using different constructors from the `time` package can cause errors. However, in this case, all instances of `time.Time` are coming from the same lines of code which should produce instances that are equal using `reflect.DeepEqual`.

Either way, can you add a comment in the test code to capture the reasoning?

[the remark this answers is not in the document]
johanbrandhorst: Sure!

johanbrandhorst:
I'll rebase this before merging, don't worry about the merge conflicts.

mgaffney (changes requested):
The sql tests in `sqltest/tests/purge` are not included in the Makefile

tmessi:
> The sql tests in `sqltest/tests/purge` are not included in the Makefile

That was fixed in #3885, once this PR is rebased it shouldn't include any changes to `internal/db/sqltest/tests/purge/*`

mgaffney (approved):

PR #3888, opened 2023-10-23
Add GetResourceType to boundary.Resource

johanbrandhorst:
## [internal/boundary: add GetResourceType to Resource](https://github.com/hashicorp/boundary/pull/3888/commits/38ca39f0a59ba24ef355e1204b48c7fb61a1be75)

The resource type method is used to automatically create refresh tokens from a resource when that resource is the last item returned in a page.


johanbrandhorst:
Superceded by #3897

PR #3889, opened 2023-10-23
Add refreshtoken package

johanbrandhorst:
## [internal/refreshtoken: add refreshtoken package](https://github.com/hashicorp/boundary/pull/3889/commits/cf4c9a567dca3987882c946377a9b6cd5e5591bd)

The refreshtoken package encapsulates domain logic surrounding the refresh token. A refresh token is used both for paginating through a collection and for requesting any updates to that collection.

This PR depends on #3888 and all preceding PRs.

PR #3890, opened 2023-10-23
Add generic pagination package

johanbrandhorst:
## [internal/pagination: add new pagination package](https://github.com/hashicorp/boundary/commit/d1912b0cd46fe243ee9c34dee30ae294c3a99178)

The pagination package provides generic functions for iterating over a collection of items such that each incoming request is filled even after filtering is applied.

## [internal/pagination: add new purge package](https://github.com/hashicorp/boundary/commit/229c590a014eaf473ce9da74573ff0ab98a49906)

The purge package contains a job that is used to purge records from the foo_deleted tables when they become older than 30 days.

This PR depends on #3889 and all preceding PRs.


johanbrandhorst:
Superceded by #3897

PR #3892, opened 2023-10-23
config: add max_page_size variable

johanbrandhorst:
The max_page_size variable lets operators control the size of the pagination pages. This value configures both the maximum permitted page size and the default page size when none is explicitly provided.

Operators can tweak this option up or down to increase or reduce the load on the database during pagination.


johanbrandhorst:
Superceded by #3906

PR #3893, opened 2023-10-23
api: add automatic list pagination

johanbrandhorst:
Updates all List endpoints to automatically paginate through results to preserve backwards compatibility. Adding support for user controllable pagination could be considered at a later time.


johanbrandhorst:
Superceded by #3906

PR #3908, opened 2023-10-24
Update websocket library

tmessi:
The nhooyr/websocket library now has a release with the new default
behavior that removes the message size limit. This allows us to switch
off of the fork and back to the upstream.

In the same release that includes the fix we need for the boundary
client, the wspb helper package was removed from websocket. This package
contained some functions to read and write protobuf messages. However,
it was using the deprecated github.com/golang/protobuf/proto library. To
avoid this dependancy in the library they opted to just remove wspb.

Thus this provides a replacement implementation that uses
google.golang.org/protobuf/proto. Notably the newer proto library seems
to lack an equivalent of proto.NewBuffer so writes to the websocket and
not able to leverage a sync.Pool of byte.Buffers to reduce allocations.

Refs: 88bb253554a8d805b9d333be8e4f1b3557d0ba60
See: https://github.com/nhooyr/websocket/commit/ac385120c6e34fa6584f3856d5db949a21bbb65e


johanbrandhorst (approved):

internal/wspb/wspb.go
```diff
@@ -0,0 +1,72 @@
...
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/hashicorp/boundary/internal/errors"
+	"google.golang.org/protobuf/proto"
+	"nhooyr.io/websocket"
+)
+
+var pool sync.Pool
```
jefferai, line 19: You don't define a `New` function for the Pool, so I believe it's possible that the pool will only ever store a single buffer -- that is, a Get followed by a Put, which is then returned via Get, and so on.

Maybe this is what you intend but I sort of doubt it...

jefferai (changes requested):
Hitting Request Changes to make sure my comment is seen/addressed

internal/wspb/wspb.go
tmessi: I based this off the old implementation from the websocket library which did not provide a New function. I don't think defining a `New` function would change this. The doc for `New` states:
> 	// New optionally specifies a function to generate
>	// a value when Get would otherwise return nil.
>	// It may not be changed concurrently with calls to Get.

And for Get:
> Get selects an arbitrary item from the Pool, removes it from the Pool, and returns it to the caller. Get may choose to ignore the pool and treat it as empty. Callers should not assume any relation between values passed to Put and the values returned by Get.
> 
> If Get would otherwise return nil and p.New is non-nil, Get returns the result of calling p.New. 

So `Get` only calls `New` if the pool is empty and would otherwise return `nil`. The current implementation in `getBuffer` checks for `nil` after calling `Get` and if it is nil, allocates a new buffer.

If `Get` and `Put` are always called in serial without any concurrency then the pool should only have at most one byte.Buffer in it. However, if there are multiple calls to `Get` before a `Put`, this code will result in new allocations of bytes.Buffers via `getBuffer`. Once `Put` is called for each buffer, the pool size would grow.

I could include a `New` function here and change the logic in `getBuffer` so that the pool is doing the allocation when necessary. That is probably more intuitive, so I am happy to make that change, but I don't think it would change the behavior.

jefferai: Yeah that's fair. I guess in the serialized pattern that's totally fine anyways. All good!

jefferai (approved):

PR #3913, opened 2023-10-25
feat (apptoken): added proto storage for apptoken

jimlambrt:
This includes addition to Makefile for injecting tags


cameronperera (approved):
LGTM

uruemu (approved):

PR #3933, opened 2023-10-26
Cameronperera/app token grant type

cameronperera:


internal/apptoken/app_token_grant_test.go
```diff
@@ -0,0 +1,139 @@
...
+			wantErrContains: "missing app token id",
+			wantErrMatch:    errors.T(errors.InvalidParameter),
+		},
+		{
+			name:            "invalid grant",
+			appTokenId:      testAppTokenId,
+			grant:           "id=*;type=*;actions=",
+			wantErrContains: "parsing grant string",
+			wantErrMatch:    errors.T(errors.InvalidParameter),
+		},
```
cameronperera, line 63: What should be the correct error to match this against? It is currently passing but I don't think it should be.

cameronperera: Fixed.

uruemu (approved):

internal/apptoken/app_token_grant.go
```diff
@@ -0,0 +1,75 @@
...
+	case appTokenId == "":
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing app token id")
+	case grant == "":
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing grant")
+	}
+
+	// Validate that the grant parses successfully. Note that we fake the scope
+	// here to avoid a lookup as the scope is only relevant at actual ACL
+	// checking time and we just care that it parses correctly.
+	perm, err := perms.Parse(ctx, "o_abcd1234", grant)
```
jimlambrt, line 36: perhaps define a const for fakeScopeId to be clear what it is and why

jimlambrt (approved):
One minor nit, but looks great!

internal/apptoken/app_token_grant.go
cameronperera: Good idea. I will update it now.

PR #3941, opened 2023-10-26
SQLite allows having 1 set to MaxOpenConnections

talanknight:
When MaxOpenConnections > 1 sqlite errors out when a read happens concurrently with a write tx.


jimlambrt (approved):
ty!

PR #3945, opened 2023-10-27
Telemetry(observations): add oidc and ldap observation events

sepehrfrgh:
This PR adds an observation event for OIDC and LDAP authentication methods. This observation event will include the user ID and auth token start and end time. This will be used for telemetry purposes
cc: @jimlambrt


johanbrandhorst (approved):

jimlambrt (changes requested):
These look great.  Do we want to assert anything about the contents of the observations in the unit tests?

jimlambrt (approved):
Looks great. Ty!

PR #3947, opened 2023-10-27
Add FullFetch to resource updating ticker

talanknight:
FullFetch is a method that fetches all resources for a user and stores them in the database.  This is for users of boundary deployments that do not yet support refresh tokens.

FullFetch runs on users which:
1. Has at least 1 resource in the cache,  AND
2. Has non resources in the cache.

Refresh is still used for instances where no resources have been synced yet, even if there are no refresh tokens for that user in the cache.  The reasoning is that a query that results in 0 resources being synced is at least as cheap as a query that uses refresh tokens and we don't know, with the current data we store, if that user has had their resources synced and seen nothing, or if the user has not had any resources synced yet.

### DIFF
--- a/internal/clientcache/internal/daemon/options.go
+++ b/internal/clientcache/internal/daemon/options.go
@@ -12,12 +12,12 @@ import (
 )
 
 type options struct {
-	withDebug                          bool
-	withRefreshInterval                time.Duration
-	withFullFetchInterval              time.Duration
-	withIntervalRandomizationFactor    float64
-	withIntervalRandomizationFactorSet bool
-	withBoundaryTokenReaderFunc        cache.BoundaryTokenReaderFn
+	withDebug                              bool
+	withRefreshInterval                    time.Duration
+	withFullFetchInterval                  time.Duration
+	testWithIntervalRandomizationFactor    float64
+	testWithIntervalRandomizationFactorSet bool
+	withBoundaryTokenReaderFunc            cache.BoundaryTokenReaderFn
 }
 
 // Option - how options are passed as args
@@ -68,14 +68,15 @@ func withFullFetchInterval(_ context.Context, d time.Duration) Option {
 	}
 }
 
-// withIntervalRandomizationFactor provides an optional interval randomization factor.
-func withIntervalRandomizationFactor(_ context.Context, f float64) Option {
+// testWithIntervalRandomizationFactor provides an optional test interval
+// randomization factor.
+func testWithIntervalRandomizationFactor(_ context.Context, f float64) Option {
 	return func(o *options) error {
 		if f < 0 {
-			return fmt.Errorf("withIntervalRandomizationFactor must be non negative")
+			return fmt.Errorf("testWithIntervalRandomizationFactor must be non negative")
 		}
-		o.withIntervalRandomizationFactor = f
-		o.withIntervalRandomizationFactorSet = true
+		o.testWithIntervalRandomizationFactor = f
+		o.testWithIntervalRandomizationFactorSet = true
 		return nil
 	}
 }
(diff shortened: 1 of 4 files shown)


internal/clientcache/internal/cache/repository_targets.go
```diff
@@ -147,13 +198,49 @@ func (r *Repository) refreshTargets(ctx context.Context, u *user, tokens map[Aut
 	return nil
 }
 
+// upsertTargets upserts the provided targets to be stored for the provided user.
+func upsertTargets(ctx context.Context, w db.Writer, u *user, in []*targets.Target) error {
+	const op = "cache.upserTargets"
```
jimlambrt, line 203:
```suggestion
	const op = "cache.upsertTargets"
```

internal/clientcache/internal/cache/repository_targets_test.go
```diff
@@ -432,5 +432,6 @@ func TestDefaultTargetRetrievalFunc(t *testing.T) {
 	assert.NoError(t, err)
 	assert.NotEmpty(t, refTok2)
 	assert.Empty(t, removed2)
+	// Note: This will sometimes fail until PR 3897 is merged
```
jimlambrt, line 435: ty!

internal/clientcache/internal/cache/refresh.go
```diff
@@ -124,3 +169,56 @@ func (r *RefreshService) Refresh(ctx context.Context, opt ...Option) error {
 	}
 	return retErr
 }
+
+// FullFetch is used for users of boundary instances which do not support
+// refresh tokens. For these users a call to FullFetch retrieves all resources
+// for that user and updates the database with the result. Users are identified
+// as being in this group if they have at least one resource in the cache and
+// do not have any refresh tokens. Users who do not have any resources nor
+// refresh tokens are updated using Refresh since a query for no resources
```
jimlambrt, line 178: This is a bit confusing.  "a query for no resources should be just as quick as q
query using refresh tokens".  It sounds like your comparing a query that returns
no resources but I think you're asserting that a full refresh costs the same as
a query using refresh tokens when you're starting with no resources in the cache.

internal/clientcache/internal/daemon/options.go
```diff
@@ -40,6 +46,40 @@ func WithDebug(_ context.Context, debug bool) Option {
...
+	return func(o *options) error {
+		if d <= 0 {
+			return fmt.Errorf("provided full fetch interval %q must be positive", d)
+		}
+		o.withFullFetchInterval = d
+		return nil
+	}
+}
+
+// withIntervalRandomizationFactor provides an optional interval randomziation factor.
```
jimlambrt, line 71:
```suggestion
// withIntervalRandomizationFactor provides an optional interval randomization factor.
```

internal/clientcache/internal/daemon/options.go
```diff
@@ -40,6 +46,40 @@ func WithDebug(_ context.Context, debug bool) Option {
...
+func withIntervalRandomizationFactor(_ context.Context, f float64) Option {
```
jimlambrt, line 72: Is this currently used?

jimlambrt: I see it's supported in newRefreshTicker(...) but I didn't see any way for the
user to change/configure it.

jimlambrt (approved):
Just a few minor bits, but looks great!

internal/clientcache/internal/daemon/options.go
jimlambrt: should we name it testWithIntervalRandomizationFactor(...) then?

PR #3948, opened 2023-10-27
Handle log rotation

talanknight:
In addition to adding log rotation and management, this also fixes a bug where the dot directory doesn't exist yet.

### DIFF
--- a/internal/clientcache/cmd/daemon/logfile_test.go
+++ b/internal/clientcache/cmd/daemon/logfile_test.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/json"
+	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


internal/clientcache/cmd/daemon/start.go
```diff
@@ -162,16 +167,24 @@ func (c *StartCommand) Run(args []string) int {
 	// report if the daemon started or not.
 
 	logFilePath := filepath.Join(dotDir, logFileName)
```
jimlambrt, line 169: perhaps a comment that explains how this will work with the eventing.  it's a
bit of state which is easy to explain now... and future devs might appreciate
it.

Also, can you add a unit test for this (a new log file and it's rotation)?

jimlambrt (changes requested):

jimlambrt (approved):
ty!

PR #3952, opened 2023-10-30
internal/tests: fix vault token with short TTL

johanbrandhorst:
This fails locally for me without this change. The error message is
```
{"kind":"Internal", "message":"credentialstores.(Service).createInRepo: unable to create credential store: vault.(Repository).CreateCredentialStore: unknown, unknown: error #0: scheduler interval must be greater than token ttl. scheduler jobs interval value: 1m0s"}
```
Not sure if it's something about my local setup (vault version??) but it might affect others so I thought I'd submit this fix.

### DIFF
--- a/internal/tests/api/credentialstores/credentialstore_test.go
+++ b/internal/tests/api/credentialstores/credentialstore_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/http"
 	"testing"
+	"time"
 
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
Turns out this was due to the vault token being dependent on the Go test timeout. @mgaffney is submitting a proper fix for this that will fix it for all uses of the test server.

PR #3959, opened 2023-10-31
Workers list pagination

johanbrandhorst:
Add pagination support for workers


internal/daemon/cluster/handlers/worker_service.go
```diff
@@ -176,7 +176,7 @@ func (ws *workerServiceServer) Status(ctx context.Context, req *pbs.StatusReques
 
 	authorizedDownstreams := &pbs.AuthorizedDownstreamWorkerList{}
 	if len(req.GetConnectedWorkerPublicIds()) > 0 {
-		knownConnectedWorkers, err := serverRepo.ListWorkers(ctx, []string{scope.Global.String()}, server.WithWorkerPool(req.GetConnectedWorkerPublicIds()), server.WithLiveness(-1))
+		knownConnectedWorkers, err := serverRepo.ListWorkersUnpaginated(ctx, []string{scope.Global.String()}, server.WithWorkerPool(req.GetConnectedWorkerPublicIds()), server.WithLiveness(-1))
```
johanbrandhorst, line 179: This rename is the only question mark in this PR - the `servers` domain already has both `ListWorkers` as a service level function and `(*Repository).ListWorkers`. This renames both to `ListWorkersUnpaginated`, but I did consider whether there was some way to unify the logic into the new `ListWorkers`. Unfortunately, I think the use cases are distinct enough that it won't make sense to support only one - after all, the new `ListWorkers` uses the logic from this lower-level `ListWorkers` method, and the parameters are significantly different.

johanbrandhorst:
We're not paginating workers

PR #3960, opened 2023-10-31
telemetry(observations): fix the details flag check

sepehrfrgh:
This PR fix the telemetry checks, with this PR we only emit authenticate events if and only if the telemetry flag is turned on

### DIFF
--- a/internal/event/context_test.go
+++ b/internal/event/context_test.go
@@ -440,14 +440,14 @@ func Test_WriteObservation(t *testing.T) {
 			cleanup: func() { event.TestResetSystEventer(t) },
 		},
 		{
-			name:               "simple",
+			name:               "simple-header",
 			ctx:                testCtx,
 			telemetryFlag:      true,
 			observationPayload: testPayloads,
 			header:             testWantHeader,
 		},
 		{
-			name:               "simple",
+			name:               "simple-details",
 			ctx:                testCtx,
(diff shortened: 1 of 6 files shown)


internal/event/context_test.go
```diff
@@ -440,13 +440,18 @@ func Test_WriteObservation(t *testing.T) {
...
 		},
 		{
-			name:                    "simple",
-			ctx:                     testCtx,
-			telemetryFlag:           true,
-			observationPayload:      testPayloads,
-			header:                  testWantHeader,
-			details:                 testWantDetails,
-			observationSinkFileName: c.AllEvents.Name(),
+			name:               "simple",
```
jimlambrt, line 443:
```suggestion
			name:               "simple-header",
```

internal/event/context_test.go
```diff
@@ -440,13 +440,18 @@ func Test_WriteObservation(t *testing.T) {
...
+			ctx:                testCtx,
+			telemetryFlag:      true,
+			observationPayload: testPayloads,
+			header:             testWantHeader,
+		},
+		{
+			name:               "simple",
```
jimlambrt, line 450:
```suggestion
			name:               "simple-details",
```

jimlambrt (approved):
Minor suggestions.  Looks good!

PR #3967, opened 2023-10-31
Add generic read/update/delete commands

jefferai:
This adds three generic commands that use the resource prefix to determine the subcommand to run and pass through args over to that subcommand. Since this occurs very early on and to avoid loading essentially every internal resource package, I moved subtype definitions to globals.


internal/cmd/commands.go
```diff
@@ -57,1268 +62,1283 @@ func initCommands(ui, serverCmdUi cli.Ui, runOpts *RunOptions) {
...
 				Func:    "create",
 			}, nil
 		},
 		"credential-libraries update": func() (cli.Command, error) {
 			return &credentiallibrariescmd.Command{
-				Command: base.NewCommand(ui),
-				Func:    "update",
-			}, nil
-		},
-		"credential-libraries update vault": func() (cli.Command, error) {
```
johanbrandhorst, line 419: This looks like a breaking change, have we waited long enough to remove this?

internal/cmd/main.go
```diff
@@ -85,10 +87,55 @@ func setupEnv(args []string) (retArgs []string, format string, outputCurlString
...
+	// Make sure we override the ID in commands
+	runOpts.ImplicitId = args[1]
+
+	// Now add the rest of the args
+	newArgs = append(newArgs,
+		args[2:]...,
+	)
+
+	return newArgs
+}
```
johanbrandhorst, line 132: 😱 I love it

internal/cmd/main.go
```diff
@@ -222,15 +292,38 @@ func groupedHelpFunc(f cli.HelpFunc) cli.HelpFunc {
 
 		fmt.Fprintf(tw, "Usage: boundary <command> [args]\n")
 
-		otherCommands := make([]string, 0, len(commands))
+		genericCommands := make([]string, 0, 3)
+		clientCommands := make([]string, 0, 6)
+		typeSpecificCommands := make([]string, 0, len(commands)-3-6)
```
johanbrandhorst, line 297: How about
```suggestion
		typeSpecificCommands := make([]string, 0, len(commands)-cap(genericCommands)-cap(clientCommands))
```

jefferai: Oh shoot I wanted something like that and totally forgot that cap exists now!

internal/cmd/commands.go
jefferai: I will double check but I'm pretty sure I checked this and the answer is yes

jefferai: See updated changelog

johanbrandhorst (approved):

PR #3969, opened 2023-11-01
Update sdk deps

jefferai:


jimlambrt (approved):
the go-eventlogger versions are fine.  Feel free to take use the @main version over v0.2.5 as well, but it's not critical.

jefferai:
I've updated to use the two go-evenlogger packages from `main`

PR #3970, opened 2023-11-01
Update api deps

jefferai:


johanbrandhorst (approved):

moduli:
@jefferai I think we might have needed to run a `go mod tidy` here? Some status checks are failing

PR #3974, opened 2023-11-01
Add indexes on update_time

talanknight:


internal/db/schema/migrations/oss/postgres/79/02_update_time_indexes.up.sql
```diff
@@ -0,0 +1,39 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+
+create index if not exists auth_ldap_account_update_time_ix on auth_ldap_account (update_time);
```
johanbrandhorst, line 6: We use the  `_idx` suffix for the indices created in the other migration file. I see there is precedent for either in our SQL migrations, but what do you think about usinx `_idx` here for consistency within this migration group?

internal/db/sqltest/tests/index/update_time.sql
```diff
@@ -0,0 +1,10 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+  select plan(1);
+
+  select is(count(*), 32::bigint) from pg_indexes where schemaname = 'public' and indexname like tablename || '_update_time_ix';
```
johanbrandhorst, line 7: Crazy idea; could we build this test based on the return of `get_deletion_tables()`? It returns all the tables ending in `_deleted`, which should be 1:1 to all tables we want to have an update_time index. Essentially, loop over all the deleted tables, remove the `_deleted` suffix and then call [`has_index`](https://pgtap.org/documentation.html#has_index) on the resulting table name. What do you think? That way, if a user adds a `_deleted` table, we ensure that the original table has an index on the update_time.

johanbrandhorst (commented):
Thanks!

internal/db/sqltest/tests/index/update_time.sql
```diff
@@ -0,0 +1,14 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+  select plan(1);
+  
+  select results_eq(
+    'select count(*)::bigint from get_deletion_tables()',
+    'select count(*)::bigint from pg_indexes where schemaname = ''public'' and indexname like tablename || ''_update_time_idx''',
```
johanbrandhorst, line 9: I don't want to be pedantic but this isn't _really_ testing that each deletion table "parent" has an update_time index. Can't we make a loop like we do for the deletion tables here: https://github.com/hashicorp/boundary/blob/c6b26fceaa7f62ca028bc4b05487f99f4ec6a2fd/internal/db/sqltest/tests/purge/deleted_table_tests.sql#L63-L83?

mgaffney: > I don't want to be pedantic but this isn't _really_ testing that each deletion table "parent" has an update_time index. Can't we make a loop like we do for the deletion tables here:

I agree. This is not testing if the indexes exist. Any reason not to use pgTap's [`is_indexed`](https://pgtap.org/documentation.html#is_indexed) function?

mgaffney (changes requested):

johanbrandhorst (approved):
LGTM

internal/db/sqltest/tests/index/update_time.sql
```diff
@@ -0,0 +1,28 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+  create function op_table(deletion_table_name name) returns text
+  as $$
+    select split_part(deletion_table_name, '_deleted', 1);
+  $$ language sql;
+
```
mgaffney, line 9:
```suggestion

```
The empty line has spaces in it

mgaffney (approved):

Issue #695, opened 2020-10-15
Recording SSH sessions


mickeypash:
**Is your feature request related to a problem? Please describe.**
For enterprises in healthcare and finance it's sometimes a requirement to have the SSH session recorded.
Also for ease of use, metadata is just not sufficient.

**Describe the solution you'd like**
Would Boundary support recording SSH sessions so that they can be shared.

**Describe alternatives you've considered**
The current alternative is [Gravitational Teleport's Recording](https://gravitational.com/teleport/features/record-ssh-sessions/)

**Explain any additional use-cases**
If there are any use-cases that would help us understand the use/need/value please share them as they can help us decide on acceptance and prioritization.
- Meeting compliance requirements
- Knowledge-sharing with the team
- Better visibility, as you don't have to sift through logs.

**Additional context**

PPacent:
@mickeypash, thanks for the suggestion! Session recording for ssh and other protocols is definitely in our vision for Boundary but we don't yet have a timeline for the delivery of this capability. For now, we have a few big-ticket items for post-launch outlined in [our roadmap](https://www.boundaryproject.io/docs/roadmap). That said, for the next set of investments we'll be listening to community feedback (like this post) to see what comes next.

darkedges:
I have added https://github.com/hashicorp/boundary/issues/707 which address using Apache Guacamole to help with this. SSH Session recording is a functionality they offer. So not sure if Boundary or Guacamole should be the right place for adding in this support.

anoncam:
Something like [tlog](https://github.com/Scribery/tlog) would be interesting.

macgahe:
@mickeypash I face the same challenge with recording SSH sessions which I see no point of doing this when you have solutions which you can easily integrate it to give more observability and threat detection.. In our scenario, the video recording was mostly to capture what the user was doing during the ssh session.. 

So the approach I worked on was simply enable Auditlogs + sudo rules + Sudo Logs and integrate all of that with a SIEM solution of your choice, in our case we setup rules to triggers alerts and was even more productive and easy than going through 4 hours of video on a specific change plus the extra storage we consume keeping those videos.. 

I would recommend you, specially if you work for a Financial Institution which needs to be PCI-DSS standard compliance, to look for a more effective approach IMHO. 



> **Is your feature request related to a problem? Please describe.**
> For enterprises in healthcare and finance it's sometimes a requirement to have the SSH session recorded.
> Also for ease of use, metadata is just not sufficient.
> 
[...]

PR #3981, opened 2023-11-02
docs: Add enterprise badge

Dan-Heath:
This pull request adds enterprise badges/alerts to Boundary docs pages where it is appropriate to call out that a feature is supported only in Enterprise or HCP Boundary. The enterprise badge and enterprise alert are existing devdot web components.

I added an **HCP/ENT** badge to the table of contents for any topics that only apply to one of the Enterprise versions. If there are nested topics, I only applied the badge to the top level:

image

For other examples in the TOC, view:

- Concepts > Domain model > Session recordings & Storage buckets
- Configuration > Session recordings
- Operations > Session recordings
- Commands (CLI) > session-recordings & storage-buckets

At the topic level, I added an alert to any topics or sections that only applied to one of the Enterprise versions. I added the alert directly under the level 1 heading, if the whole topic was targeted to the Ent versions:

image

I can't change the Boundary logo in the alert, but I did add the text "This feature requires HCP Boundary or Boundary Enterprise" and included a link to the product page (similar to Vault).

If only one section of a topic applies to the Enterprise versions, I added the alert directly underneath the relevant heading. See the example here for Credential injection in the Credential management topic:

image

I did a find/replace for any places where we were still using the HCP/ENT "sup" tag and replaced them with the alert. There are a couple exceptions where I left an inline "sup" tag, if we used it in a sentence in a field definition. It seemed disruptive to include the alert in the middle of a paragraph. But there aren't many instances like that.

This is a manual process, so we would have to remember to add the tag/alert for new features going forward.

[See the changes in the preview deployment](https://boundary-4epzqpxnx-hashicorp.vercel.app/boundary/docs)

### DIFF
--- a/website/content/docs/commands/session-recordings/download.mdx
+++ b/website/content/docs/commands/session-recordings/download.mdx
@@ -7,6 +7,8 @@ description: |-
 
 # session-recordings download
 
+<EnterpriseAlert product="boundary">This feature requires <a href="https://www.hashicorp.com/products/boundary">HCP Boundary or Boundary Enterprise</a></EnterpriseAlert>
+
 Command: `boundary session-recordings download`
 
 The `boundary session-recordings download` command lets you download a Boundary session recording.
--- a/website/content/docs/commands/session-recordings/index.mdx
+++ b/website/content/docs/commands/session-recordings/index.mdx
@@ -7,6 +7,8 @@ description: |-
 
 # session-recordings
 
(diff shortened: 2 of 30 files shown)


AdamBouhmad (approved):
*huge* improvement based on feedback in https://github.com/hashicorp/boundary/issues/695. Great work @Dan-Heath

PR #3992, opened 2023-11-03
Add generic read/update/delete commands

jefferai:
This adds three generic commands that use the resource prefix to determine the subcommand to run and pass through args over to that subcommand.

This replaces #3967 with a different approach that maintains the different packages/types of the various subtypes (important for proto reflection).


internal/daemon/worker/auth_rotation_test.go
```diff
@@ -126,6 +131,11 @@ func TestRotationTicking(t *testing.T) {
 			nodeenrollment.WithStorageWrapper(w.Config().WorkerAuthStorageKms),
 		)
 		require.NoError(err)
+		if bytes.Equal(currKey, currNodeCreds.CertificatePublicKeyPkix) {
+			// Load due to parallel tests may mean that we didn't hit rotation
+			// yet, loop back to evaluate again
+			continue
+		}
```
johanbrandhorst, line 138: Is this a rogue test fix for this flake in the middle of this PR? Mind putting it in a separate PR?

internal/target/tcp/register.go
```diff
@@ -18,6 +19,7 @@ type targetHooks struct{}
 
 func init() {
 	target.Register(Subtype, targetHooks{}, TargetPrefix)
+	globals.RegisterPrefixSubtype(globals.TcpTargetPrefix, Subtype)
```
johanbrandhorst, line 22: Will you follow up with an enterprise PR after merge?

internal/tests/api/credentials/credentials_test.go
```diff
@@ -500,7 +500,7 @@ func TestUpdateAfterKeyRotation(t *testing.T) {
 	for {
 		jobs, err := scopesClient.ListKeyVersionDestructionJobs(ctx, proj.PublicId)
 		require.NoError(err)
-		if len(jobs.Items) == 1 {
+		if len(jobs.Items) >= 1 {
```
johanbrandhorst, line 503: image

Is this the correct fix? I think I wrote this flaky test 😛.

jefferai: We want to ensure that rotation happens right? If the test is slow and rotation happens more than once, the test fails, but the actual behavior is success.

internal/target/tcp/register.go
jefferai: Yes

johanbrandhorst (approved):

PR #4002, opened 2023-11-04
fix(plugin): remove unit test race condition

ddebko:
## Summary

The following flaky tests were found in the plugin loopback package:

```
go test github.com/hashicorp/boundary/internal/plugin/loopback -count=100
--- FAIL: Test_GetObjectStream_Server (0.00s)
    testing_grpc_stream_test.go:208: 
                Error Trace:    /Users/mycow/Developer/boundary/internal/plugin/loopback/testing_grpc_stream_test.go:208
                Error:          Should be true
                Test:           Test_GetObjectStream_Server
--- FAIL: Test_GetObjectStream_Server (0.00s)
    testing_grpc_stream_test.go:171: 
                Error Trace:    /home/johan/src/hashicorp/boundary/internal/plugin/loopback/testing_grpc_stream_test.go:171
                Error:          An error is expected but got nil.
                Test:           Test_GetObjectStream_Server

go test github.com/hashicorp/boundary/internal/plugin/loopback -count=100 -run Test_GetObjectStream_Server
--- FAIL: Test_GetObjectStream_Server (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x2 addr=0x18 pc=0x100a318a8]

goroutine 33 [running]:
```

### Fix:

This PR addresses the flaky tests by refactoring the tests to be isolated from each other. Please run the following command several times and if the error persists let me know:

```
go test github.com/hashicorp/boundary/internal/plugin/loopback -count=100 -run Test_GetObjectStream_Server
```

### DIFF
--- a/internal/plugin/loopback/testing_grpc_stream_test.go
+++ b/internal/plugin/loopback/testing_grpc_stream_test.go
@@ -6,8 +6,10 @@ package loopback
 import (
 	"fmt"
 	"io"
+	"strings"
 	"sync"
 	"testing"
+	"time"
 
 	plgpb "github.com/hashicorp/boundary/sdk/pbs/plugin"
 	tr "github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


moduli (commented):
Confirmed that running tests a bunch do not show the flake.

Do you happen to know how the previous code resulted in a flake? Weren't the tests already running serially and also presumably "resetting the stream"?

ddebko:
> Do you happen to know how the previous code resulted in a flake? Weren't the tests already running serially and also presumably "resetting the stream"?

I actually don't understand how line 171 could fail its check because it should always return an invalid type error. This specific check is testing out that the default case for the switch in the SendMsg method is working as expected. At this point in the unit test we aren't trying to interact with the channels.

My best guess is that a different go routine is mutating the stream reference at the same time and causing this weird issue.

internal/plugin/loopback/testing_grpc_stream_test.go
```diff
@@ -16,195 +18,339 @@ import (
...
-	ctx := stream.server.Context()
-	require.NotNil(ctx)
-
-	// Validate RecvMsg does not read from the channel messages
-	streamSize := len(stream.messages)
-	var msg *getObjectStreamResponse
-	err = stream.server.RecvMsg(msg)
-	require.NoError(err)
-	require.Nil(msg)
-	require.Equal(streamSize, len(stream.messages))
```
moduli, line 131: I'm not seeing a corresponding test for this in the new `Test_GetObjectStream_Server`? Is it needed? (Disclaimer: I don't have a lot of context on what this is testing 😅 )

moduli (approved):

internal/plugin/loopback/testing_grpc_stream_test.go
ddebko: The corresponding test for the server side is `recieve empty message`

PR #4004, opened 2023-11-06
Minor test fixes

jefferai:
The auth rotation test fix allows us to keep checking if we did not yet see the expected bytes in case the call to lookup succeeds but we are late in rotating due to parallelism.

The api credentials test fix allows us to check if at _least_ one rotation has happened. That way if two rotations happen before we check again due to parallelism, this is deemed a successful result.

### DIFF
--- a/internal/daemon/worker/auth_rotation_test.go
+++ b/internal/daemon/worker/auth_rotation_test.go
@@ -4,6 +4,7 @@
 package worker_test
 
 import (
+	"bytes"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):

moduli:
@jefferai Looks like `TestUpdateAfterKeyRotation` failed on the merge to `main`

```
--- FAIL: TestUpdateAfterKeyRotation (200.66s)

    credentials_test.go:502: 
        	Error Trace:	/home/runner/work/boundary/boundary/internal/tests/api/credentials/credentials_test.go:502
        	            	error performing client request during ListKeyVersionDestructionJobs call: context deadline exceeded
        	Test:       	TestUpdateAfterKeyRotation`
```
https://github.com/hashicorp/boundary/actions/runs/6776501533/job/18418159172

PR #4009, opened 2023-11-06
internal/genapi: correct list response body and map

johanbrandhorst:
When using the new auto-pagination logic, we would previously return the Body and Map of the last response, instead of one corresponding to the returned result. Correct this by overwriting the Body and Map after pagination has finished.

### DIFF
--- a/internal/api/genapi/templates.go
+++ b/internal/api/genapi/templates.go
@@ -276,6 +276,9 @@ func (c *Client) List(ctx context.Context, {{ .CollectionFunctionArg }} string, 
 	for i, item := range target.Items {
 		idToIndex[item.Id] = i
 	}
+	// Removed IDs in the response may contain duplicates,
+	// maintain a set to avoid returning duplicates to the user.
+	removedIds := map[string]struct{}{}
 	for {
 		req, err := c.client.NewRequest(ctx, "GET", "{{ .CollectionPath }}", nil, apiOpts...)
 		if err != nil {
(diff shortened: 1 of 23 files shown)


internal/api/genapi/templates.go
```diff
@@ -671,8 +689,13 @@ var structTemplate = template.Must(template.New("").Funcs(
 package {{ .Package }}
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
+	"errors"
 	"fmt"
+	"io"
+	"net/url"
```
johanbrandhorst, line 698: I added some imports that were already used but added but post-generation formatting steps.

internal/listtoken/list_token.go
```diff
@@ -387,8 +387,6 @@ func (tk *Token) Validate(
 			return errors.New(ctx, errors.InvalidParameter, op, "list token's refresh component missing last item ID")
 		case st.LastItemUpdateTime.After(time.Now()):
 			return errors.New(ctx, errors.InvalidParameter, op, "list token's refresh component's last item was updated in the future")
-		case st.LastItemUpdateTime.Before(tk.CreateTime):
```
johanbrandhorst, line 390: Validating that the update time of the last item is after the create time of the token cannot account for items whose update time are earlier than the original query because of the added database transaction timeout buffer.

johanbrandhorst: Eh this check was a bit overzealous anyway, I don't think it adds that much confidence. We do have a variable for the offset but I'd still prefer we don't do this check.

PR #4010, opened 2023-11-06
Target pagination e2e test

johanbrandhorst:
Exercises target pagination both over the CLI and the API. The max page size of the controller is configured to make the tests run quicker while still exercising the desired behavior.

### DIFF
--- a/enos/modules/test_e2e_docker/main.tf
+++ b/enos/modules/test_e2e_docker/main.tf
@@ -175,6 +175,10 @@ variable "worker_tag_collocated" {
   type        = string
   default     = ""
 }
+variable "max_page_size" {
+  description = "Max allowed page size for pagination requests"
+  type        = number
+}
 variable "postgres_user" {
   description = "Username for accessing the postgres database"
   type        = string
@@ -294,6 +298,7 @@ resource "enos_local_exec" "run_e2e_test" {
     E2E_LDAP_USER_NAME            = var.ldap_user_name
     E2E_LDAP_USER_PASSWORD        = var.ldap_user_password
     E2E_LDAP_GROUP_NAME           = var.ldap_group_name
+    E2E_MAX_PAGE_SIZE             = var.max_page_size
     BOUNDARY_DIR                  = abspath(var.local_boundary_src_dir)
     BOUNDARY_CLI_DIR              = abspath(var.local_boundary_dir)
     MODULE_DIR                    = abspath(path.module)
--- a/enos/enos-scenario-e2e-docker-base-plus.hcl
+++ b/enos/enos-scenario-e2e-docker-base-plus.hcl
@@ -130,6 +130,7 @@ scenario "e2e_docker_base_plus" {
       ldap_user_name           = step.create_ldap_server.user_name
       ldap_user_password       = step.create_ldap_server.user_password
       ldap_group_name          = step.create_ldap_server.group_name
+      max_page_size            = step.create_boundary.max_page_size
     }
   }
 }
--- a/enos/enos-scenario-e2e-docker-base-with-vault.hcl
+++ b/enos/enos-scenario-e2e-docker-base-with-vault.hcl
@@ -138,6 +138,7 @@ scenario "e2e_docker_base_with_vault" {
       vault_addr_internal      = step.create_vault.address_internal
       vault_root_token         = step.create_vault.token
       vault_port               = step.create_vault.port
+      max_page_size            = step.create_boundary.max_page_size
     }
   }
 }
--- a/enos/enos-scenario-e2e-docker-base-with-worker.hcl
+++ b/enos/enos-scenario-e2e-docker-base-with-worker.hcl
@@ -183,6 +183,7 @@ scenario "e2e_docker_base_with_worker" {
       vault_port               = step.create_vault.port
       worker_tag_egress        = local.egress_tag
       worker_tag_collocated    = step.create_boundary.worker_tag
+      max_page_size            = step.create_boundary.max_page_size
     }
   }
 }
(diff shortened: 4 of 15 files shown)


testing/internal/e2e/tests/base/target_pagination_test.go
moduli, line 1: Per the comments in the Slack thread, we may want to adjust the number of targets we create for timing reasons. It looks like in the CI run, each test took about 40s (locally, for me, they take about 30s).

If I understand it right, this is a parameter in the config file? We can adjust `enos/modules/docker_boundary/boundary-config.hcl` accordingly.

testing/internal/e2e/tests/base/target_pagination_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+	var newTargets targets.TargetListResult
+	err = json.Unmarshal(output.Stdout, &newTargets)
+	require.NoError(t, err)
+
+	require.Len(t, initialTargets.Items, globals.DefaultMaxPageSize+1)
+	assert.Equal(t, newTargetId, newTargets.Items[len(newTargets.Items)-1].Id)
+	assert.Empty(t, initialTargets.ResponseType)
+	assert.Empty(t, initialTargets.RemovedIds)
+	assert.Empty(t, initialTargets.RefreshToken)
```
moduli, line 104: Was this meant to validate `newTargets`?

```
require.Len(t, newTargets.Items, globals.DefaultMaxPageSize)
assert.Equal(t, newTargetId, newTargets.Items[len(newTargets.Items)-1].Id)
assert.Empty(t, newTargets.ResponseType)
assert.Empty(t, newTargets.RemovedIds)
assert.Empty(t, newTargets.RefreshToken)
```

testing/internal/e2e/tests/base/target_pagination_test.go
```diff
@@ -0,0 +1,181 @@
...
+			"-format=json",
+		),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+
+	var initialTargets targets.TargetListResult
+	err = json.Unmarshal(output.Stdout, &initialTargets)
+	require.NoError(t, err)
+
+	require.Len(t, initialTargets.Items, globals.DefaultMaxPageSize+1)
```
moduli, line 70: I might be missing something conceptually here...

I thought that when using the list endpoint, we're only going to get `defaultMaxPageSize` items back from the endpoint and that we'd have to invoke another `list` to get the next page of items?

testing/internal/e2e/tests/base/target_pagination_test.go
```diff
@@ -0,0 +1,181 @@
...
+	require.NoError(t, err)
+	_, err = tClient.Delete(ctx, initialTargets.Items[0].Id)
+	require.NoError(t, err)
+
+	// List again, should have the new and deleted target
+	newTargets, err := tClient.List(ctx, newProjectId, targets.WithRefreshToken(initialTargets.RefreshToken))
+	require.NoError(t, err)
+
+	require.Len(t, newTargets.Items, 1)
+	assert.Equal(t, newTargetResult.Item.Id, newTargets.Items[0].Id)
```
moduli, line 171: Is this always going to be the case? The new target id is going to the first element? Also, is that different from the CLI? It looks like we're checking that the new target is the last in the list?

testing/internal/e2e/tests/base/target_pagination_test.go
johanbrandhorst: This is the case over the API directly, yes, but the SDK will automatically paginate to the end, which means both the "Api" and "Cli" tests will get all results after the first call.

testing/internal/e2e/tests/base/target_pagination_test.go
johanbrandhorst: Doh, good shout

testing/internal/e2e/tests/base/target_pagination_test.go
johanbrandhorst: Yes and yes. It should always be the first (and only) element returned, since we're using a refresh token created by the previous call. Only new targets and updated targets since that List call will show up in this call. Since this is the only new or updated target, it's the only one to show up. I could add a target update as well to ensure that shows up here 👍🏻 .

johanbrandhorst: It is also different from the CLI because the CLI doesn't expose any of the pagination logic to the user (at the moment). There's no flag to provide a refresh token, so there's little point in returning it in the result.

testing/internal/e2e/tests/base/target_pagination_test.go
johanbrandhorst: I templated this and propagated it into the test so the test knows how many iterations to make

enos/modules/test_e2e_docker/main.tf
```diff
@@ -172,6 +172,10 @@ variable "worker_tag_collocated" {
   type    = string
   default = ""
 }
+variable "max_page_size" {
```
moduli, line 175: We will need to set a default value here since not all scenarios will pass this value into this module.

moduli (changes requested):

enos/modules/test_e2e_docker/main.tf
johanbrandhorst: Hm, I really wanted that to not have a default value so that it had to be set explicitly and we couldn't accidentally miss it. I'll just add it to the places that are erroring if that's ok?

moduli: Yea that works, too.

moduli: We'll just need to update scenarios in `boundary-enterprise`, too, when this gets merged.

moduli (approved):

PR #4011, opened 2023-11-07
Create Client and update Errors for client daemon commands

talanknight:
A follow up PR for https://github.com/hashicorp/boundary/pull/3871 where @jefferai pointed out we don't need to use the api.Client, as that is a bit overkill for just sending a request over a domain socket.  He also pointed out that the Errors returned should be CLI focused, which means that they should be Capitalized, end in a period, and following existing convention of not including the op.

### DIFF
--- a/internal/clientcache/internal/cache/repository_targets_test.go
+++ b/internal/clientcache/internal/cache/repository_targets_test.go
@@ -350,7 +350,7 @@ func TestRepository_QueryTargets(t *testing.T) {
 	require.NoError(t, r.AddKeyringToken(ctx, addr, kt1))
 	require.NoError(t, r.AddKeyringToken(ctx, addr, kt2))
 
-	query := `(name % name1 or name % name2) and scope_id = "p_123"`
+	query := `(name % 'name1' or name % 'name2') and scope_id = "p_123"`
 
 	errorCases := []struct {
 		name        string


internal/clientcache/internal/client/client.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// Body returns the pre-read response body.
+func (r Response) Body() []byte {
+	return r.contents
+}
+
+// Get sends a GET http request to the provided path.  The vals provided are
+// encoded and attached to the request if present.
+func (c *Client) Get(ctx context.Context, path string, vals *url.Values) (*Response, *api.Error, error) {
```
jefferai, line 72: If you wanted you could probably combine these, or share an internal function between GET/POST. Either way works.

api/client.go
```diff
@@ -240,16 +234,6 @@ func (c *Config) ConfigureTLS() error {
 // This also removes any trailing "/v1"; we'll use that in our commands so we
 // don't require it from users.
 func (c *Config) setAddr(addr string) error {
-	if strings.HasPrefix(addr, "unix://") {
```
jefferai, line 243: This seems to ignore the comment on the field you removed

internal/clientcache/internal/daemon/404_handler.go
```diff
@@ -0,0 +1,18 @@
...
+// SPDX-License-Identifier: BUSL-1.1
+
+package daemon
+
+import (
+	"context"
+	"net/http"
+)
+
+// newStopHandlerFunc creates a handler that cancels the server context.
```
jefferai, line 11: It doesn't seem like it cancels the server context...

It's also not clear from the comment here what this is supposed to do. Why does a 404 handler cancel anything?

api/client.go
jefferai: Oh, the comment says that the field takes a filename approach instead of a URL approach since parsing file names as a URL may error out on some platforms -- and then it removes the code that keeps it from being parsed as a URL. :-)

jefferai: Confirmed this will be tracked separately, resolving

internal/clientcache/cmd/daemon/addtoken.go
```diff
@@ -76,38 +77,48 @@ func (c *AddTokenCommand) Run(args []string) int {
...
+		token := apiClient.Token()
 		if parts := strings.SplitN(token, "_", 4); len(parts) == 3 {
 			pa.AuthTokenId = strings.Join(parts[:2], "_")
 		} else {
-			return nil, errors.New(ctx, errors.InvalidParameter, op, "found auth token is not in the proper format")
+			return nil, nil, errors.New("The found auth token is not in the proper format.")
+		}
+		if c.FlagOutputCurlString {
+			// do not output the actual auth token if we are printing out the curl string
+			pa.AuthToken = "$BOUNDARY_TOKEN"
```
jefferai, line 114: Please look at how this is done in the normal client via `boundary config get-token`

internal/clientcache/internal/daemon/404_handler.go
jefferai: This doesn't seem updated?

jefferai (approved):

PR #4022, opened 2023-11-08
Fix three minor issues:

jefferai:
* If passing a private key without specifying it in URL syntax, give a proper error message (previously we were only making sure it was a scheme we knew how to handle, but we need to also be more informative if it's not a URL at all)

* If the handshake from the proxy fails, don't hang, just exit

* Update some error messages to use capitalization in line with our standard CLI error format


louisruch (approved):
LGTM

PR #4023, opened 2023-11-08
Copy wspb into sdk in preparation for moving it out of main

jefferai:


tmessi (approved):

sdk/wspb/wspb.go
```diff
@@ -0,0 +1,72 @@
...
+// messages via a websocket.
+package wspb
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/hashicorp/boundary/internal/errors"
```
tmessi, line 14: Probably need to switch to just using the stdlib `errors` package instead.

jefferai: yes, thanks

jefferai:
I can confirm that the failing test is a bit flaky locally as well so ignoring...

jefferai:
At least, ignoring in this PR :-)

PR #4032, opened 2023-11-10
feat(oidc-prompt): schema changes for `auth_oidc_prompt`

elimt:
Boundary OIDC method today does not support passing in prompts during authentication. Prompts are optional OIDC parameters that determine the behaviour of authentication server: https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

## Changes
Schema changes for supporting `prompts` for OIDC auth method.

- New `auth_oidc_prompt` table which contains all the prompts for OIDC auth method
- New `auth_oidc_prompt_enm` table which contains possible enum values for a prompt
  - Currently supported:
    - `none`: 
The Authorization Server MUST NOT display any authentication or consent user interface pages
    - `login`: The Authorization Server SHOULD prompt the End-User for reauthentication
    - `consent`: The Authorization Server SHOULD prompt the End-User for consent before returning information to the Client
    - `select_account`: The Authorization Server SHOULD prompt the End-User to select a user account
- `oidc_auth_method_with_value_obj` view has been updated to return `prompt` value


internal/db/schema/migrations/oss/postgres/79/01_auth_oidc_prompt.up.sql
```diff
@@ -0,0 +1,105 @@
...
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+
+-- auth_oidc_prompt_enm entries define the supported oidc auth method prompts.
+create table auth_oidc_prompt_enm (
+  name text primary key
+    constraint only_predefined_auth_oidc_prompts_allowed
+    check (
+      name in (
```
elimt, line 11: The constraint needs to support validation for  comma separated values. Maybe we could use regex for this

internal/proto/controller/storage/auth/oidc/store/v1/oidc.proto
```diff
@@ -163,6 +163,15 @@ message AuthMethod {
     this: "AccountClaimMaps"
     that: "attributes.account_claim_maps"
   }];
+
+  // prompts are the optional prompts allowed for an oidc auth method.
+  // Thee value specifies whether the authorization server prompts
```
jimlambrt, line 168:
```suggestion
  // These value objects specify whether the authorization server prompts
```

internal/proto/controller/storage/auth/oidc/store/v1/oidc.proto
```diff
@@ -163,6 +163,15 @@ message AuthMethod {
...
+  // the end-user for reauthentication, account selection and consent
```
jimlambrt, line 169: Perhaps add: `These are Value Objects that will be stored as Prompt messages,
and are operatated on as a complete set.`

internal/proto/controller/storage/auth/oidc/store/v1/oidc.proto
```diff
@@ -163,6 +163,15 @@ message AuthMethod {
...
+  // @inject_tag: `gorm:"-"`
+  repeated string prompts = 220 [(custom_options.v1.mask_mapping) = {
+    this: "prompts"
+    that: "attributes.prompts"
+  }];
```
jimlambrt, line 174: I think, we need a `message Prompt` as well.  See one of the other value objects
like signing algs as an example.

internal/proto/controller/storage/auth/oidc/store/v1/oidc.proto
```diff
@@ -163,6 +163,15 @@ message AuthMethod {
     this: "AccountClaimMaps"
     that: "attributes.account_claim_maps"
   }];
+
+  // prompts are the optional prompts allowed for an oidc auth method.
+  // Thee value specifies whether the authorization server prompts
+  // the end-user for reauthentication, account selection and consent
+  // @inject_tag: `gorm:"-"`
+  repeated string prompts = 220 [(custom_options.v1.mask_mapping) = {
+    this: "prompts"
```
jimlambrt, line 172:
```suggestion
    this: "Prompts"
```

internal/db/schema/migrations/oss/postgres/79/01_auth_oidc_prompt.up.sql
jimlambrt: Why does it need to support a comma delimited list.  Isn't there a one to many
relationship between the auth method and it's associated prompts?

internal/db/schema/migrations/oss/postgres/79/01_auth_oidc_prompt.up.sql
```diff
@@ -0,0 +1,105 @@
...
+    on update cascade,
+  prompt text
+    constraint auth_oidc_prompt_enm_fkey
+    references auth_oidc_prompt_enm(name)
+    on delete restrict
+    on update cascade,
+  primary key(oidc_method_id, prompt)
+);
+comment on table auth_oidc_prompt is
+  'auth_oidc_prompt entries are the prompts allowed for an oidc auth method.  An oidc auth method may have 0 or 1 comma separated prompts value';
```
jimlambrt, line 42: I don't think the "comma separated" part of this comment is correct.

Aren't these the prompts "specified" for the auth method vs what's "allowed".

internal/db/schema/migrations/oss/postgres/79/01_auth_oidc_prompt.up.sql
```diff
@@ -0,0 +1,105 @@
...
+);
+
+insert into auth_oidc_prompt_enm (name) values
+  ('none'),
+  ('login'),
+  ('consent'),
+  ('select_account');
+
+-- auth_oidc_prompt entries are the prompts allowed for an oidc auth method.
+create table auth_oidc_prompt (
```
jimlambrt, line 27: should there be some sort of immutable trigger for this table that defines
what's mutable if anything?

internal/db/schema/migrations/oss/postgres/79/01_auth_oidc_prompt.up.sql
```diff
@@ -0,0 +1,105 @@
...
+
+
+create trigger
+  default_create_time_column
+before
+insert on auth_oidc_prompt
+  for each row execute procedure default_create_time();
+
+-- we will drop the oidc_auth_method_with_value_obj view, so we can recreate it
+-- and add the oidc claim's scopes to the returned set.
```
jimlambrt, line 52:
```suggestion
-- and add the oidc prompts to the returned set.
```

internal/db/schema/migrations/oss/postgres/79/01_auth_oidc_prompt.up.sql
```diff
@@ -0,0 +1,105 @@
...
+drop view oidc_auth_method_with_value_obj;
+
+-- oidc_auth_method_with_value_obj is useful for reading an oidc auth method
+-- with its associated value objects (algs, auds, certs, claims scopes and
```
jimlambrt, line 56:
```suggestion
-- with its associated value objects (algs, auds, certs, claims scopes, prompts and
```

internal/db/sqltest/tests/auth/oidc/prompt.sql
```diff
@@ -0,0 +1,32 @@
...
+
+  -- Insert valid valid_auth_oidc_prompt_value value to test constraint with a valid value
+  prepare valid_auth_oidc_prompt_value as
+     insert into auth_oidc_prompt
+     (oidc_method_id,       prompt)
+     values
+        ('aom___widget',   'select_account');
+
+  select lives_ok('valid_auth_oidc_prompt_value');
+
```
jimlambrt, line 30: Should we test the recreated view here as well?

jimlambrt (changes requested):
Just a few questions here and there.  Ty!

jimlambrt (changes requested):
Just two minor bits.

jimlambrt (approved):
ty!

PR #4043, opened 2023-11-14
Observation(event): adding op instead of type details paylaod

sepehrfrgh:
This PR adds an `op` string instead of a `type` inside details payload in the observation event.
cc: @jimlambrt

### DIFF
--- a/internal/event/event_observation.go
+++ b/internal/event/event_observation.go
@@ -107,9 +107,6 @@ func (o *observation) ComposeFrom(events []*eventlogger.Event) (eventlogger.Even
 			if _, ok := payload[DetailsField]; !ok {
 				payload[DetailsField] = []gated.EventPayloadDetails{}
 			}
-			if g.Op == "" {
-				g.Op = "Unknown"
-			}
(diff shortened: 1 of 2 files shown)


internal/event/event_observation.go
```diff
@@ -107,8 +107,11 @@ func (o *observation) ComposeFrom(events []*eventlogger.Event) (eventlogger.Even
 			if _, ok := payload[DetailsField]; !ok {
 				payload[DetailsField] = []gated.EventPayloadDetails{}
 			}
+			if g.Op == "" {
```
sepehrfrgh, line 110: @jimlambrt I think this check is never going to happen because in `newObservation` there is a validation check for an empty `op` string however, since `ComposeFrom` calls from the filter, Do you  think is there any situation that could lead to an empty `op`?

jimlambrt: I agree, it should be unreachable unless newObservation changes.

jimlambrt (approved):
ty!

internal/event/event_observation.go
sepehrfrgh: Sure then I will remove it

sepehrfrgh: @jimlambrt should be this for 0.15.x release?

PR #4044, opened 2023-11-14
feat(oidc-prompt): repository changes for OIDC Prompt

elimt:
Boundary OIDC method does not currently support passing in prompts during authentication. This change adds the capability of passing OIDC prompts. Prompts are optional OIDC parameters that determine the behaviour of the authentication server: https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

### Changes

- Update create, read, write and delete  repository changes to support new `Prompt` option for OIDC auth method
- Update test helper functions to support creating a test auth method with prompts


jimlambrt (approved):
Looks great. ty!!!!

PR #4046, opened 2023-11-15
Refresh specific resources before searching when they are stale

talanknight:
When a user's resources haven't been updated within the max search staleness, running `boundary search -resource targets` will perform a best effort refresh of targets for the requesting user before performing the search and returning the result.  Both the max staleness and the search blocking refresh timeout are configurable.

While making these changes I ran across and fixed a number of bugs and inconsistencies.  This includes:
* Removing the `-seconds` suffix from the duration flags
* Passing on the flags onto the background process when -background flag is passed in
* Checking the cli version against the daemon version for more than just the `v1/stop` handler
* Using the server context instead of the request context when saving per user/resource errors in the db
* Correctly setting the SavedError creation time
* Only auto start the daemon and send the token used when the command run is successful

### DIFF
--- a/internal/clientcache/cmd/daemon/status.go
+++ b/internal/clientcache/cmd/daemon/status.go
@@ -20,7 +20,7 @@ import (
 	"github.com/posener/complete"
 )
 
-var errDaemonNotRunning = stderr.New("The deamon process is not running.")
+var errDaemonNotRunning = stderr.New("The daemon process is not running.")
 
 var (
 	_ cli.Command             = (*AddTokenCommand)(nil)
--- a/internal/clientcache/internal/cache/refresh.go
+++ b/internal/clientcache/internal/cache/refresh.go
@@ -151,11 +151,14 @@ func (r *RefreshService) refreshableUsers(ctx context.Context, in []*user) ([]*u
 // method returns without any requests to the boundary controller.
 // While the criteria of whether the user will refresh or not is almost the same
 // as the criteria used in Refresh, RefreshForSearch will not refresh any
-// data if there is not at least 1 refresh token stored for the resource. It
+// data if there is not a refresh token stored for the resource. It
 // might make sense to change this in the future, but the reasoning used is
 // that we should not be making an initial load of all resources while blocking
 // a search query, in case we have not yet even attempted to load the resources
 // for this user yet.
+// Note: Currently, if the context timesout we stop refreshing completely and
+// return to the caller.  A possible enhancement in the future would be to return
+// when the context is Done, but allow the refresh to proceed in the background.
 func (r *RefreshService) RefreshForSearch(ctx context.Context, authTokenid string, resourceType SearchableResource, opt ...Option) error {
 	const op = "cache.(RefreshService).RefreshForSearch"
 	if r.maxSearchRefreshTimeout > 0 {
@@ -218,6 +221,8 @@ func (r *RefreshService) RefreshForSearch(ctx context.Context, authTokenid strin
 				return errors.Wrap(ctx, err, op)
 			}
 		}
+	default:
+		return errors.New(ctx, errors.InvalidParameter, op, "unrecognized resource type")
 	}
 
 	return nil
--- a/internal/clientcache/internal/daemon/server.go
+++ b/internal/clientcache/internal/daemon/server.go
@@ -32,7 +32,12 @@ import (
 )
 
 const (
-	DefaultSearchStaleness      = 60 * time.Second
+	// DefaultSearchStaleness is the default amount of time that can pass before
+	// a search request causes a blocking refresh to happen for the resources
+	// being searched for.
+	DefaultSearchStaleness = 60 * time.Second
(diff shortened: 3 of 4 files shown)


internal/clientcache/cmd/daemon/status.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/posener/complete"
 )
 
-var errDaemonNotRunning = stderr.New("daemon is not running")
+var errDaemonNotRunning = stderr.New("The deamon process is not running.")
```
johanbrandhorst, line 23:
```suggestion
var errDaemonNotRunning = stderr.New("The daemon process is not running.")
```

internal/clientcache/internal/cache/refresh.go
```diff
@@ -131,6 +144,85 @@ func (r *RefreshService) refreshableUsers(ctx context.Context, in []*user) ([]*u
...
+		rtv, err := r.repo.lookupRefreshToken(ctx, u, sessionResourceType)
+		if err != nil {
+			return errors.Wrap(ctx, err, op)
+		}
+		if rtv != nil && time.Since(rtv.UpdateTime) > r.maxSearchStaleness {
+			if err := r.repo.refreshSessions(ctx, u, tokens, opt...); err != nil {
+				return errors.Wrap(ctx, err, op)
+			}
+		}
+	}
```
johanbrandhorst, line 221: Can we have a default case with an error so we know if we forget to add new resources here?

internal/clientcache/internal/cache/refresh.go
```diff
@@ -131,6 +144,85 @@ func (r *RefreshService) refreshableUsers(ctx context.Context, in []*user) ([]*u
...
+// data if there is not at least 1 refresh token stored for the resource. It
+// might make sense to change this in the future, but the reasoning used is
+// that we should not be making an initial load of all resources while blocking
+// a search query, in case we have not yet even attempted to load the resources
+// for this user yet.
+func (r *RefreshService) RefreshForSearch(ctx context.Context, authTokenid string, resourceType SearchableResource, opt ...Option) error {
+	const op = "cache.(RefreshService).RefreshForSearch"
+	if r.maxSearchRefreshTimeout > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, r.maxSearchRefreshTimeout)
```
johanbrandhorst, line 163: I thought a bit more about this, and it would be really cool if instead of canceling the refresh operation completely this would simply return early and read from the DB while letting the refresh operation continue to run in the background. It would be a little more complicated as you'd essentially need to start a goroutine that writes to the DB and the send a signal on a channel, where this goroutine could wait on that channel or this context to timeout. Anyway, that would mean that the update could still take place, but asynchronously, if the user has waited long enough already. What do you think? It could also be something we do later, not necessarily something we do now.

internal/clientcache/internal/daemon/server.go
```diff
@@ -31,6 +31,11 @@ import (
 	"github.com/hashicorp/go-secure-stdlib/parseutil"
 )
 
+const (
+	DefaultSearchStaleness      = 60 * time.Second
+	DefaultSearchRefreshTimeout = 1 * time.Second
```
johanbrandhorst, line 36: Could you add comments to explain what these are used for?

internal/clientcache/internal/daemon/ticker.go
```diff
@@ -15,11 +15,9 @@ import (
 )
 
 const (
-	DefaultRefreshIntervalSeconds = 60
-	defaultRefreshInterval        = DefaultRefreshIntervalSeconds * time.Second
+	DefaultRefreshInterval = 60 * time.Second
 
-	DefaultFullFetchIntervalSeconds = 60 * 5
-	defaultFullFetchInterval        = DefaultFullFetchIntervalSeconds * time.Second
+	DefaultFullFetchInterval = 60 * 5 * time.Second
```
johanbrandhorst, line 20: Could you add comments to explain what these are used for?

johanbrandhorst (approved):

internal/clientcache/internal/daemon/server.go
```diff
@@ -31,6 +31,16 @@ import (
...
 )
 
+const (
+	// DefaultSearchStaleness is the default amount of time that can pass before
+	// a search request causes a blocking refresh to happen for the resources
+	// being searched for.
+	DefaultSearchStaleness = 60 * time.Second
+	// DefaultSearchRefreshTimeout is the default amount of time that a refresh
+	// that is blocking a search request can run before timing out.
+	DefaultSearchRefreshTimeout = 1 * time.Second
```
jefferai, line 41: Is this too low? Not sure if we've run benchmarks around how fast we can get a list from Boundary when refreshing but it would be a shame if Boundary is actually returning fresh results and we simply cut it off. Wonder if this should be somewhere between 1-2.

jefferai (approved):
Looks good! Just one question.

PR #4052, opened 2023-11-17
Target list pagination refactor

johanbrandhorst:
## Target pagination refactor

This PR contains the proposed refactor of the target list endpoint to align with the new design detailed in ICU-110. The most significant changes are in the listtoken, pagination and target packages. The commits are split by package. Going commit by commit should give a decent overview of the changes.

## List Token state machine

The list token transitions are well understood and defined by the subtype of the token. The following diagram describes the valid state transitions:

```mermaid
stateDiagram-v2
    InitialResponse --> PaginationToken: More pages in initial phase
    InitialResponse --> StartRefreshToken: No more pages in initial phase
    PaginationToken --> PaginationToken: More results after this page
    PaginationToken --> StartRefreshToken: End of initial pagination phase
    StartRefreshToken --> StartRefreshToken: End of refresh phase
    StartRefreshToken --> RefreshToken: More results after this page
    RefreshToken --> StartRefreshToken: End of refresh phase
    RefreshToken --> RefreshToken: More results after this page
```

For more detail please consider the RFC.


internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -268,37 +268,82 @@ func (s Service) ListTargets(ctx context.Context, req *pbs.ListTargetsRequest) (
 	}
 
 	var listResp *pagination.ListResponse[target.Target]
-	if req.GetRefreshToken() == "" {
+	sortBy := "created_time"
```
johanbrandhorst, line 271: It's a little odd to have this here, but we have to know which sort order is used, and the best indicator of that is based on which type of input token we're working with. For `StartRefreshToken` and `RefreshToken`, it's `updated_time`, otherwise it's `created_time`.

internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -268,37 +268,82 @@ func (s Service) ListTargets(ctx context.Context, req *pbs.ListTargetsRequest) (
...
-		if err != nil {
-			return nil, err
-		}
-		if err := domainRefreshToken.Validate(ctx, resource.Target, grantsHash); err != nil {
-			return nil, err
-		}
-		listResp, err = target.ListRefresh(ctx, grantsHash, pageSize, filterItemFn, domainRefreshToken, repo)
-		if err != nil {
-			return nil, err
+		switch st := pt.Token.(type) {
```
johanbrandhorst, line 282: When we have a list token, we have 3 different scenarios:
1. We need to build a second page for the initial pagination phase. This is using a `ListToken_PaginationToken`.
2. We need to start a refresh phase. This is done using a `ListToken_StartRefreshToken`.
3. We need to continue a refresh phase. This is done using a `ListToken_RefreshToken`.

internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -319,22 +364,48 @@ func (s Service) ListTargets(ctx context.Context, req *pbs.ListTargetsRequest) (
...
+					PhaseEndTime:       timestamppb.New(st.PhaseEndTime),
+					LastUseTime:        timestamppb.New(st.LastUseTime),
+					LastItemId:         st.LastItemId,
+					LastItemUpdateTime: timestamppb.New(st.LastItemUpdateTime),
+				},
+			}
+		default:
+			return nil, errors.New(ctx, errors.Internal, op, "unexpected list token subtype")
+		}
+		resp.ListToken, err = handlers.MarshalListToken(ctx, lt)
```
johanbrandhorst, line 408: This logic could likely all be in a helper, but we'll know for sure when working on sessions next

internal/db/schema/migrations/oss/postgres/79/02_root_aggregate_updates.up.sql
```diff
@@ -0,0 +1,46 @@
...
+  end;
+  $$ language plpgsql;
+  comment on function update_target_table_update_time() is
+    'update_target_table_update_time is used to automatically update the update_time '
+    'of the base table whenever one of the subtype tables are updated';
+
+  create trigger update_target_table_update_time before update on target_tcp
+    for each row execute procedure update_target_table_update_time();
+  create trigger update_target_table_update_time before update on target_ssh
+    for each row execute procedure update_target_table_update_time();
```
johanbrandhorst, line 44: We'll need something like this migration for all the base tables. Should we split it into separate migration files?

internal/listtoken/item.go
```diff
@@ -1,20 +1,19 @@
...
 
-// Item represents a generic resource with a public ID, update time
-// and resource type.
+// Item represents a generic resource with a public ID,
+// create time, update time and resource type.
 type Item struct {
 	publicId     string
-	updateTime   time.Time
+	createTime   *timestamp.Timestamp
+	updateTime   *timestamp.Timestamp
```
johanbrandhorst, line 16: Create Time and Update Time are never set together on this type. When we have a `PaginationToken` component we set the create time, and when we have a `RefreshToken` component we set the update time. I don't _love_ this, but it _is_ really nice to be able to use this alongside one of the `boundary.Resource` types. In practice the service list method validation will ensure we use the right list token component in the right place.

internal/listtoken/list_token.go
```diff
@@ -0,0 +1,281 @@
...
+
+func (*StartRefreshToken) isTokenSubtype() {}
+
+type RefreshToken struct {
+	PhaseStartTime     time.Time
+	PhaseEndTime       time.Time
+	LastUseTime        time.Time
+	LastItemId         string
+	LastItemUpdateTime time.Time
+}
```
johanbrandhorst, line 52: The StartRefreshToken and RefreshToken types are very similar - indeed, StartRefreshToken is just a subset of the RefreshToken. We made these two different types because the alternative would have been to add many optional fields to the RefreshToken. This way, we can use strict validation of the types at construction time and carry those guarantees through the domain logic.

internal/pagination/example_test.go
```diff
@@ -113,12 +113,11 @@ func ExampleList() {
 }
 
 func ExampleListRefresh() {
```
johanbrandhorst, line 115: These examples will be updated before this is taken out of draft

internal/pagination/pagination.go
```diff
@@ -56,19 +56,14 @@ type ListFilterFunc[T boundary.Resource] func(ctx context.Context, item T) (bool
 // If prevPageLastItem is empty, it returns a slice of T starting with the least recently updated.
 type ListItemsFunc[T boundary.Resource] func(ctx context.Context, prevPageLastItem T, limit int) ([]T, error)
 
-// ListRefreshItemsFunc returns a slice of T that have been updated since prevPageLastItem.
-// If prevPageLastItem is empty, it returns a slice of T that have been updated since the
-// item in the refresh token.
-type ListRefreshItemsFunc[T boundary.Resource] func(ctx context.Context, tok *refreshtoken.Token, prevPageLastItem T, limit int) ([]T, error)
```
johanbrandhorst, line 62: This function was no longer necessary, and it simplifies both the service functions and the pagination logic not to use it.

internal/pagination/pagination.go
```diff
@@ -56,19 +56,14 @@ type ListFilterFunc[T boundary.Resource] func(ctx context.Context, item T) (bool
...
-
 // EstimatedCountFunc is used to estimate the total number of items
 // available for the resource that is being listed.
 type EstimatedCountFunc func(ctx context.Context) (int, error)
 
 // ListDeletedIDsFunc is used to list the IDs of the resources deleted since
 // the given timestamp. It returns a slice of IDs and the timestamp of the
 // instant in which the slice was created.
-type ListDeletedIDsFunc func(ctx context.Context, since time.Time) ([]string, time.Time, error)
```
johanbrandhorst, line 71: We made the decision at this point to stop using the transaction timestamp in the return from the ListDeletedIDs callback. It was introduced _before_ we decided to add a database read timeout buffer to the queries. We could add this back, but we think that a `time.Now()` prior to making the query should provide sufficient accuracy (together with the buffer).

internal/target/query.go
```diff
@@ -61,5 +61,57 @@ select public_id, project_id from target
...
+    select public_id
+      from target
+      where update_time > @updated_after_time
+      and (update_time, public_id) < (@last_item_update_time, @last_item_id) -- last item update time and id
+      order by update_time desc, public_id asc
+      fetch first %d rows only
+    )
+  select * from target_all_subtypes
+    where public_id in (select public_id from targets)
+    order by update_time desc, public_id asc;
```
johanbrandhorst, line 115: Please review these queries, we'd love feedback on improving them.

internal/target/service_list_refresh.go
```diff
@@ -38,25 +41,32 @@ func ListRefresh(
...
 		}
-		return repo.listTargets(ctx, opts...)
+		// Add the database read timeout to account for any creations missed due to concurrent
+		// transactions in the original list pagination phase.
+		return repo.listTargetsRefresh(ctx, rt.PhaseEndTime.Add(-globals.DatabaseReadTimeout), opts...)
+	}
+	listDeletedIdsFn := func(ctx context.Context, since time.Time) ([]string, error) {
+		// Add the database read timeout to account for any deletions missed due to concurrent
+		// transactions in previous requests.
+		return repo.listDeletedIds(ctx, since.Add(-globals.DatabaseReadTimeout))
```
johanbrandhorst, line 68: This seemed like the right place to do the database read timeout buffer compensation, rather than in the pagination or listtoken packages.

internal/target/repository.go
```diff
@@ -237,52 +237,69 @@ func (r *Repository) FetchAuthzProtectedEntitiesByScope(ctx context.Context, pro
...
 
-	if len(r.permissions) == 0 {
-		return []Target{}, nil
+	opts := GetOpts(opt...)
+	limit := r.defaultLimit
+	if opts.WithLimit != 0 {
+		limit = opts.WithLimit
 	}
 
-	where, args := r.listPermissionWhereClauses()
```
johanbrandhorst, line 244: This PR removes the use of the list permission where clause in favor of simplifying the querying and performing this filtering later. This was added as a stop gap to avoid requesting results that were never going to be shown to the user at a time when we were reading a whole table at once, and shouldn't be necessary anymore. It would still allow us to avoid requesting items that the user doesn't have access to, but I'd be worried about the impact on the querying performance given that we're very carefully trying to make Postgres use the indexes we've created. Feedback appreciated? Should we keep this? How can we do so without affecting the querying performance?

internal/target/repository.go
```diff
@@ -237,52 +237,69 @@ func (r *Repository) FetchAuthzProtectedEntitiesByScope(ctx context.Context, pro
...
+func (r *Repository) queryTargets(ctx context.Context, query string, args []any) ([]Target, error) {
+	const op = "target.(Repository).queryTargets"
+
+	rows, err := r.reader.Query(ctx, query, args)
 	if err != nil {
 		return nil, errors.Wrap(ctx, err, op)
 	}
-
+	defer rows.Close()
+	var foundTargets []*targetView
```
johanbrandhorst, line 297: I'd like to draw attention to how this listing works, because it is currently unique to Targets among all our resources:
1. A view is created that uses an SQL union to append all entries for all target subtypes to a single "table", carefully crafting the union such that table rows are uniformly typed.
2. We create a type (`targetView`) that can read the structure of the view and can be used to scan the result from the query
3. We loop over each entry in the view and, depending on the "type", further unmarshal the value using a registered subtype handler. The final slice is one of the top level resource interface type.
This structure allows the top level ("base") package, in this case `internal/target` to perform the query and scan the results without having specific knowledge of how to scan the subtype rows. My question is this - do we want to apply this pattern to other resource that have a similar sort of structure? Examples include auth methods, credential stores, host catalogs, etc. It would require creating a subtype registry for all of these domain packages and using this registration model. Personally I'd love to _avoid_ this, but I'm not sure how else to solve this problem, since we have to very carefully avoid an import cycle.

internal/target/repository.go
tmessi: We should not remove this. It wasn't really added as a stop gap. It was a notable performance improvement by moving the filtering *into* the query where it would be more performant. Moving the filtering out of the query is far more likely to be less performant. We should examine some query plans to see how the proposed indexes are used. We can leverage CTEs to help guide postgres if need be, but more often there would be good reason for postgres to avoid using an index.

internal/target/repository.go
tmessi: We can probably look more closely when we get to one of those packages. Personally I think it is a good pattern, and there is no need to avoid it. But then again, I did implement it in the target package :shrug:.

internal/target/repository.go
johanbrandhorst: OK cool, we'll add it back. I'll try to weave it into the CTE we're using to pick the rows rather than the wrapping query that uses the view, but feedback would be appreciate on how to best construct the query.

internal/target/query.go
```diff
@@ -61,5 +61,61 @@ select public_id, project_id from target
...
+    select public_id
+      from target
+      where (create_time, public_id) < (@last_item_create_time, @last_item_id)
+      and %s
+      order by create_time desc, public_id asc
+      fetch first %d rows only
+    )
+  select * from target_all_subtypes
+    where public_id in (select public_id from targets)
+    order by create_time desc, public_id asc;
```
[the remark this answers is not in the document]
johanbrandhorst: Thanks!

internal/target/query.go
```diff
@@ -61,5 +61,61 @@ select public_id, project_id from target
...
 select sum(reltuples::bigint) as estimate from pg_class where oid in ('target_tcp'::regclass, 'target_ssh'::regclass)
+`
+
+	listTargetsTemplate = `
+  with targets as (
+    select public_id
+      from target
+      where %s
+      order by create_time desc, public_id asc
+      fetch first %d rows only
```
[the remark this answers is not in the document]
johanbrandhorst: Makes sense

internal/pagination/pagination.go
johanbrandhorst: OK, that is fine, we'll need to make that adjustment both to this function and the list functions since we use timestamps relative to that point as well.

internal/listtoken/item.go
```diff
@@ -0,0 +1,48 @@
...
+
+// Item defines a subset of a boundary.Resource that can
+// be used as an input to a DB operation for the purposes
+// of pagination and sorting.
+type Item interface {
+	GetPublicId() string
+	GetCreateTime() *timestamp.Timestamp
+	GetUpdateTime() *timestamp.Timestamp
+	GetResourceType() resource.Type
+}
```
johanbrandhorst, line 19: I had to move this from `pagination` to this package because we need it in this package and can't import the `pagination` package here. It doesn't seem to have caused any problems.

internal/target/query.go
```diff
@@ -61,5 +61,69 @@ select public_id, project_id from target
...
+         and (update_time, public_id) < (@last_item_update_time, @last_item_id)
+         -- search condition for applying permissions is constructed
+         and %s
+    order by update_time desc, public_id asc
+       limit %d
+  )
+    select *
+      from target_all_subtypes
+     where public_id in (select public_id from targets)
+  order by update_time desc, public_id asc;
```
johanbrandhorst, line 127: Still interested in feedback on the queries :)

tmessi (changes requested):

globals/globals.go
```diff
@@ -29,6 +29,10 @@ var (
 	// DefaultMaxPageSize is the maximum list page size allowed if not set in the config.
 	DefaultMaxPageSize = 1000
 
+	// DatabaseReadTimeout is used to account for database state mutations
+	// missed due to concurrent transactions.
+	DatabaseReadTimeout = 30 * time.Second
```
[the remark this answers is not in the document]
johanbrandhorst: I'm happy to change the name, but do we need to adjust the duration? I recall we just sorta picked this as an approximately appropriate amount of time. What should we change the name to? I'm also happy to actually use the `DefaultMaxRequestDuration` if that makes more sense. What do you think?

internal/listtoken/item.go
johanbrandhorst: The problem is that the `Transition` function needs to take an `Item` parameter. We can't take an unexported `item`, since it's used from the pagination package. We could take an `any` and then type assert to the unexported `item` but that seems worse.

tmessi: It looks like `Transition` only needs a `boundary.Resource` since it only calls `GetPublicId()` and `GetCreateTime()`.

internal/target/service_list_refresh.go
johanbrandhorst: I'm happy to change it, but it seems like the knowledge of this subtraction belongs in the specific domain package rather than in the listtoken domain. The listtoken, to me, is responsible for providing a type that maintains the state of the pagination and helps transition between the phases. It doesn't need to (shouldn't?) know about the database specific details such as this timestamp adjustment. It's also only used in the refresh phase, so if we put this in the listtoken, how do we make clear when the timestamp has been adjusted already? It seems simpler to me to handle that at query time than that token construction time.

internal/listtoken/item.go
johanbrandhorst: That works :+1:

johanbrandhorst: Done

internal/target/service_list_refresh.go
tmessi: I think that makes sense. I think the downside is that there are a lot of places where this subtraction needs to be done.

johanbrandhorst: It's true, but at least it's a simple operation, and I intend to include this comment in each place to add context to the reader.

internal/listtoken/list_token.go
```diff
@@ -0,0 +1,441 @@
...
+	grantsHash []byte,
+	previousDeletedIdsTime time.Time,
+	phaseUpperBound time.Time,
+	phaseLowerBound time.Time,
+	lastItemId string,
+	lastItemUpdateTime time.Time,
+) (*Token, error) {
+	const op = "listtoken.NewRefresh"
+
+	if len(grantsHash) == 0 {
```
jimlambrt, line 216: switch?

internal/listtoken/list_token.go
```diff
@@ -0,0 +1,441 @@
...
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing grants hash")
+	}
+	if createTime.After(time.Now()) {
```
jimlambrt, line 219: do you want one value for now or several?

internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -238,67 +238,117 @@ func (s Service) ListTargets(ctx context.Context, req *pbs.ListTargetsRequest) (
 	if req.GetPageSize() != 0 && uint(req.GetPageSize()) < s.maxPageSize {
 		pageSize = int(req.GetPageSize())
 	}
-	filter, err := handlers.NewFilter(ctx, req.GetFilter())
-	if err != nil {
-		return nil, err
+	filterItemFn := func(ctx context.Context, item target.Target) (bool, error) {
```
jimlambrt, line 241: these bits might be more simple as as switch with a default for this and a case
for `req.GetFilter() != ""`

internal/daemon/controller/handlers/targets/target_service.go
jimlambrt: might be better to set no default here, and just set `create_time` explicitly
where it's appropriate.

internal/daemon/controller/handlers/list_token.go
```diff
@@ -8,42 +8,80 @@ import (
...
 	}
 	return &tok, nil
 }
 
-// MarshalRefreshToken marshals a refresh token to its string representation.
-func MarshalRefreshToken(ctx context.Context, token *pbs.ListRefreshToken) (string, error) {
-	const op = "handlers.MarshalRefreshToken"
+// MarshalListToken assembles and marshals a list token to its string representation.
+func MarshalListToken(ctx context.Context, token *listtoken.Token, resourceType pbs.ResourceType) (string, error) {
+	const op = "handlers.MarshalListToken"
```
jimlambrt, line 35: why not check for an unspecified resourceType?  using a switch?

internal/target/repository.go
```diff
@@ -234,95 +235,129 @@ func (r *Repository) FetchAuthzProtectedEntitiesByScope(ctx context.Context, pro
...
+
+// listTargetsRefresh lists targets in a project based on the data in the WithPermissions option
+// provided to the Repository constructor. If no permissions are available, this function
+// is a no-op.
+// Supported options:
+//   - WithLimit which overrides the limit set in the Repository object
+//   - WithStartPageAfterItem which sets where to start listing from
+func (r *Repository) listTargetsRefresh(ctx context.Context, updatedAfter time.Time, opt ...Option) ([]Target, time.Time, error) {
+	const op = "target.(Repository).listTargetsRefresh"
+
```
jimlambrt, line 273: perhaps check updatedAfter for `IsZero()`?

internal/target/service_list_page.go
```diff
@@ -0,0 +1,67 @@
...
+	ctx context.Context,
+	grantsHash []byte,
+	pageSize int,
+	filterItemFn pagination.ListFilterFunc[Target],
+	tok *listtoken.Token,
+	repo *Repository,
+) (*pagination.ListResponse[Target], error) {
+	const op = "target.ListPage"
+
+	if len(grantsHash) == 0 {
```
jimlambrt, line 31: switcheroo?

internal/target/service_list_refresh_page.go
```diff
@@ -0,0 +1,80 @@
...
+	ctx context.Context,
+	grantsHash []byte,
+	pageSize int,
+	filterItemFn pagination.ListFilterFunc[Target],
+	tok *listtoken.Token,
+	repo *Repository,
+) (*pagination.ListResponse[Target], error) {
+	const op = "target.ListRefreshPage"
+
+	if len(grantsHash) == 0 {
```
jimlambrt, line 35: switcheroo?

jimlambrt (changes requested):
Looks great.  Some minor bits.

internal/daemon/controller/handlers/targets/target_service.go
johanbrandhorst: Done

internal/listtoken/item.go
johanbrandhorst: Good ideas!

johanbrandhorst: Done

internal/pagination/pagination.go
```diff
@@ -115,6 +110,35 @@ func List[T boundary.Resource](
...
+			resp.ListToken, err = listtoken.NewPagination(
+				ctx,
+				listTime, // Use list time as the create time of the token
+				lastItem.GetResourceType(),
+				grantsHash,
+				lastItem.GetPublicId(),
+				lastItem.GetCreateTime().AsTime(),
+			)
+		}
+		if err != nil {
```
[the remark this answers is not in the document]
johanbrandhorst: Done

internal/listtoken/list_token.go
```diff
@@ -0,0 +1,441 @@
...
+	grantsHash []byte,
+	lastItemId string,
+	lastItemCreateTime time.Time,
+) (*Token, error) {
+	const op = "listtoken.NewPagination"
+
+	if len(grantsHash) == 0 {
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing grants hash")
+	}
+	if createTime.After(time.Now()) {
```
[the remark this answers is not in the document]
johanbrandhorst: I don't think it matters so much here, if these timestamps are so close to now that there's a behavioral difference I'm OK with it.

internal/listtoken/list_token.go
johanbrandhorst: See other answer

internal/daemon/controller/handlers/targets/target_service.go
johanbrandhorst: Sounds good

internal/daemon/controller/handlers/list_token.go
johanbrandhorst: Good idea

jimlambrt (approved):
ty!

tmessi (approved):

internal/target/repository.go
```diff
@@ -234,95 +235,132 @@ func (r *Repository) FetchAuthzProtectedEntitiesByScope(ctx context.Context, pro
...
-		args,
-		db.WithLimit(limit),
-		db.WithOrder(withOrder),
-	)
-	if err != nil {
-		return nil, errors.Wrap(ctx, err, op)
+	return r.queryTargets(ctx, query, args, limit)
+}
+
+// listTargetsRefresh lists targets in a project based on the data in the WithPermissions option
```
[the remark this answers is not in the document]
johanbrandhorst: Good point, will update!

internal/target/query.go
```diff
@@ -61,5 +61,272 @@ select public_id, project_id from target
...
+    ingress_worker_filter,
+    default_client_port,
+    storage_bucket_id,
+    enable_session_recording,
+    'ssh' as type
+  from ssh_targets
+)
+select * 
+  from all_targets
+order by create_time desc, public_id asc;
```
tmessi, line 128:
```suggestion
with targets as (
    select public_id
      from target
     -- search condition for applying permissions is constructed
     where %s
  order by create_time desc, public_id asc
     limit %d
),
tcp_targets as (
  select *
    from target_tcp
   where public_id in (select public_id from targets)
),
ssh_targets as (
  select *
    from target_ssh
   where public_id in (select public_id from targets)
),
all_targets as (
  select public_id,
         project_id,
         name,
         description,
         default_port,
         session_max_seconds,
         session_connection_limit,
         version,
         create_time,
         update_time,
         worker_filter,
         egress_worker_filter,
         ingress_worker_filter,
         default_client_port,
         null as storage_bucket_id,
         false as enable_session_recording,
         'tcp' as type
    from tcp_targets
   union
  select public_id,
         project_id,
         name,
         description,
         default_port,
         session_max_seconds,
         session_connection_limit,
         version,
         create_time,
         update_time,
         worker_filter,
         egress_worker_filter,
         ingress_worker_filter,
         default_client_port,
         storage_bucket_id,
         enable_session_recording,
         'ssh' as type
    from ssh_targets
)
  select *
    from all_targets
order by create_time desc, public_id asc;
```

Can you update the formatting of the queries like above. This creates a "river" so that keywords all line up and helps with readability of the queries.

PR #4060, opened 2023-11-27
subtypes,handlers: add runtime trace regions

johanbrandhorst:
The subtype.Filterable and handlers.ProtoToStruct functions are costly and worth measuring whenever runtime tracing is in use. This has almost no cost when tracing is not in use, so there's little harm in having them in place. Runtime tracing like this allows us to create user regions viewable in the pprof trace format, like this:

image

### DIFF
--- a/internal/daemon/controller/handlers/accounts/account_service.go
+++ b/internal/daemon/controller/handlers/accounts/account_service.go
@@ -187,7 +187,7 @@ func (s Service) ListAccounts(ctx context.Context, req *pbs.ListAccountsRequest)
 
 		// This comes last so that we can use item fields in the filter after
 		// the allowed fields are populated above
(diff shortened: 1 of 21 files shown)


tmessi (approved):
This seems useful. Would you mind adding some details to the commit/pr and/or a new readme to describe what else is needed to run things with the tracing enabled and how to collect and visualize it?

johanbrandhorst:
> This seems useful. Would you mind adding some details to the commit/pr and/or a new readme to describe what else is needed to run things with the tracing enabled and how to collect and visualize it?

Great idea, I'll add something.

johanbrandhorst:
There it is, I added a TRACING.md, please take a look. I didn't want to put it at the top level but I recognize that this might make it hard to discover. Thoughts?

tmessi (approved):

PR #4064, opened 2023-11-28
Expand rate limiting to support limits per IP address and Auth Token

tmessi:


internal/ratelimit/handler.go
```diff
@@ -127,21 +148,23 @@ func Handler(l *rate.Limiter, next http.Handler) http.Handler {
...
-
 			// The only other error here should be rate.ErrLimitNotFound, which
 			// shouldn't be possible given how we initialize the limiter and
 			// the checks done by extractResourceAction.
 			rw.WriteHeader(http.StatusInternalServerError)
 			return
 		}
 
+		l.SetUsageHeader(quota, rw.Header())
+		l.SetPolicyHeader(res, a, rw.Header())
```
mgaffney, line 166: `SetPolicyHeader` returns an error. Should we check it?

internal/ratelimit/handler.go
```diff
@@ -114,10 +116,29 @@ func extractResourceAction(path, method string) (res, act string, err error) {
...
-		path := req.URL.Path
-		method := req.Method
+		reqInfo, ok := event.RequestInfoFromContext(req.Context())
+		if !ok || reqInfo == nil || reqInfo.ClientIp == "" {
+			// TODO: what to do here?
+			// We could:
+			// 1. Ignore and continue ahead, with the ip being empty...
+			// 2. Have fallback logic to extract the Ip
+			// 3. Return an error since it would mean the middlewares are not setup correctly.
+			rw.WriteHeader(http.StatusInternalServerError)
```
mgaffney, line 126: I agree with option 3.

internal/ratelimit/handler.go
```diff
@@ -114,10 +116,29 @@ func extractResourceAction(path, method string) (res, act string, err error) {
...
 
-		res, a, err := extractResourceAction(path, method)
+		authtoken, ok := req.Context().Value(globals.ContextAuthTokenPublicIdKey).(string)
+		if !ok || authtoken == "" {
+			// TODO: what to do here?
+			// We could:
+			// 1. Ignore and continue ahead, with the authtoken being empty...
+			// 2. Have fallback logic to extract the authtoken
+			// 3. Return an error since it would mean the middlewares are not setup correctly.
+			rw.WriteHeader(http.StatusInternalServerError)
```
mgaffney, line 137: Again, I agree with option 3.

mgaffney (approved):

internal/ratelimit/handler.go
tmessi: The only thing I think we could do with an error from this would be wrap it so there is an error event. At least currently, we should never encounter an error here, since the only possible error from this we would have encountered earlier and returned from the handler before getting here.

internal/ratelimit/handler.go
tmessi: To be clear, by "return an error" I mean what is currently implemented here with setting a 500 response header and returning from the handler.

internal/ratelimit/handler.go
```diff
@@ -114,10 +116,29 @@ func extractResourceAction(path, method string) (res, act string, err error) {
...
+			// 1. Ignore and continue ahead, with the ip being empty...
+			// 2. Have fallback logic to extract the Ip
+			// 3. Return an error since it would mean the middlewares are not setup correctly.
+			rw.WriteHeader(http.StatusInternalServerError)
+			return
+		}
 
-		res, a, err := extractResourceAction(path, method)
+		authtoken, ok := req.Context().Value(globals.ContextAuthTokenPublicIdKey).(string)
+		if !ok {
```
tmessi, line 131: I removed the `authtoken == ""` check, since this gets set to an empty string for unauthed requests like for the request to `/v1/auth-methods/{id}:authenticate`.

mgaffney (approved):

PR #4067, opened 2023-11-29
feat(ratelimit): Emit sys event detailing the rate limit configuration

tmessi:
A sys event will be emitted when the controller is constructed, and if
the rate limits are changed via a SIGHUP. The event contains the details
of the rate limits for each combination of resource and action. This
event can be used by operators to confirm the limits are set as they
would expect based on the controller config.

### DIFF
--- a/internal/daemon/controller/controller.go
+++ b/internal/daemon/controller/controller.go
@@ -264,6 +264,7 @@ func New(ctx context.Context, conf *Config) (*Controller, error) {
 		return nil, fmt.Errorf("error initializing rate limiter: %w", err)
 	}
 	c.rateLimiter.Store(rateLimiter)
+	ratelimit.WriteLimitsSysEvent(c.baseContext, rateLimits, conf.RawConfig.Controller.ApiRateLimiterMaxEntries)
 
 	var pluginLogger hclog.Logger
 	for _, enabledPlugin := range c.enabledPlugins {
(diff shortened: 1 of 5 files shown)


internal/daemon/controller/controller.go
```diff
@@ -264,6 +264,7 @@ func New(ctx context.Context, conf *Config) (*Controller, error) {
 		return nil, fmt.Errorf("error initializing rate limiter: %w", err)
 	}
 	c.rateLimiter.Store(rateLimiter)
+	ratelimit.WriteLimitsSysEvent(c.baseContext, rateLimits, conf.RawConfig.Controller.ApiRateLimiterMaxEntries)
```
mgaffney, line 267: `ratelimit.WriteLimitsSysEvent` returns an error. Can we handle the error? If not, can we `_ = ratelimit.WriteLimitsSysEvent` and add a comment to explain?

internal/daemon/controller/controller.go
```diff
@@ -666,6 +667,7 @@ func (c *Controller) ReloadRateLimiter(newLimitConfigs ratelimit.Configs, newMax
 	}
 
 	old := c.rateLimiter.Swap(limiter)
+	ratelimit.WriteLimitsSysEvent(c.baseContext, ratelimits, newMaxEntries)
```
mgaffney, line 670: Same as above

internal/ratelimit/limiter.go
```diff
@@ -14,3 +19,63 @@ func NewLimiter(limits []*rate.Limit, maxEntries int) (*rate.Limiter, error) {
...
+
+type (
+	actionLimits    []limit
+	resourceActions map[string]actionLimits
+	resources       map[string]resourceActions
+)
+
+// WriteLimitsSysEvent writes a sys event that contains all of the provided
+// rate limits.
+func WriteLimitsSysEvent(ctx context.Context, limits []*rate.Limit, maxEntries int) error {
```
mgaffney, line 42: When can this function return an error?

mgaffney (changes requested):

internal/ratelimit/limiter.go
tmessi: Hmmm. I don't think it can. I'll update it.

tmessi:
Closing in favor of #4089

PR #4075, opened 2023-11-30
Add credential pagination

johanbrandhorst:
Adds pagination to credentials. This uses a simpler pattern than that in Targets, as all resources listed are in the same subpackage.


internal/credential/static/repository_credential.go
```diff
@@ -715,48 +719,170 @@ func (r *Repository) UpdateJsonCredential(ctx context.Context,
...
+			db.WithOrder("create_time desc, public_id asc"),
+			db.WithLimit(limit),
+		); err != nil {
+			return err
+		}
+
+		ret, err = r.queryCredentials(ctx, rd, credentialIds)
+		if err != nil {
+			return err
+		}
```
johanbrandhorst, line 768: I want to draw attention to how this is currently implemented. Instead of what we're doing in the target domain, where we have a single query that assembles a final list where each item must be parsed via a subtype registry, this reuses the existing types by splitting the gathering of the IDs from the reading of the elements. I think this should be a safe way to implement this since:

1. All the operations are happening in the same transaction, so there won't be any deletions between the first and the subsequent queries.
2. The only expensive part of this is the initial ID search, and it should hit the appropriate indexes.

Thoughts appreciated!

jimlambrt (approved):
looks good. nothing surprised me in the PR.

tmessi (approved):
Looks great!

PR #4076, opened 2023-11-30
Add target type and session user_id as searchable

talanknight:
Since user_id in the session table conflicted with the existing column name rerferencing the owner of the synced data, I renamed that column to `owner_user_id`.  LMK what you thnk about that name.

Also note that a bug was uncovered in this PR which when creating with an OnConflict.   Using db.UpdateAll did not update any columns that the new resource being inserted didn't have a non-zero value for.  This would have resulted in addresses remaining in the cache if a target was updated to use host sets instead of addresses for example, or a description being removed in the controller but not getting removed from the cache.

### DIFF
--- a/internal/clientcache/internal/db/schema.sql
+++ b/internal/clientcache/internal/db/schema.sql
@@ -109,7 +109,7 @@ create table if not exists keyring_token (
 -- specific fields extracted to facilitate searching over those fields
 create table if not exists target (
   -- the boundary user id of the user who has was able to read/list this target
-  owner_user_id text not null
+  fk_user_id text not null
     references user(id)
     on delete cascade,
   -- the boundary id of the target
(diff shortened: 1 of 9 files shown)


internal/clientcache/internal/db/schema.sql
```diff
@@ -109,7 +109,7 @@ create table if not exists keyring_token (
 -- specific fields extracted to facilitate searching over those fields
 create table if not exists target (
   -- the boundary user id of the user who has was able to read/list this target
-  user_id text not null
+  owner_user_id text not null
```
jimlambrt, line 112: perhaps, `fk_user_id`?

jimlambrt (changes requested):
Just need to settle on a name.

jimlambrt (approved):
ty!

PR #4077, opened 2023-11-30
Allow forced refresh attempt when searching

talanknight:
This PR allows the requester to force a refresh attempt before performing the search.  This can be done on the CLI by passing in the flag `-force-refresh` or by making a request to the search endpoint with `force_refresh=true` query param set.

### DIFF
--- a/internal/clientcache/internal/cache/options.go
+++ b/internal/clientcache/internal/cache/options.go
@@ -77,7 +77,8 @@ func WithSessionRetrievalFunc(fn SessionRetrievalFunc) Option {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


internal/clientcache/internal/cache/options.go
```diff
@@ -75,3 +76,11 @@ func WithSessionRetrievalFunc(fn SessionRetrievalFunc) Option {
 		return nil
 	}
 }
+
+// WithIgnoreStalenessTime provides an option for ignoring the staleness time
```
jefferai, line 80: Comment name doesn't match function name

jefferai (approved):
Looks good!

PR #4079, opened 2023-11-30
Sessions list pagination

mikemountain:
Similarly to #4052, this PR continues the proposed refactor of the list endpoints to align with the new design detailed in ICU-110, with the session implementation

### DIFF
--- a/internal/daemon/controller/common/scopeids/scope_ids_test.go
+++ b/internal/daemon/controller/common/scopeids/scope_ids_test.go
@@ -51,7 +51,7 @@ func TestListingScopeIds(t *testing.T) {
 	sessionsRepoFn := func(opt ...session.Option) (*session.Repository, error) {
 		return session.NewRepository(ctx, rw, rw, kms, opt...)
 	}
-	sess, err := sessions.NewService(ctx, sessionsRepoFn, iamRepoFn)
+	sess, err := sessions.NewService(ctx, sessionsRepoFn, iamRepoFn, 1000)
 	require.NoError(t, err)
 
 	tcs := []struct {


internal/daemon/controller/handlers/sessions/session_service.go
```diff
@@ -164,51 +205,74 @@ func (s Service) ListSessions(ctx context.Context, req *pbs.ListSessionsRequest)
...
-		Type: resource.Session,
-	}
-	for _, item := range sesList {
-		res.Id = item.GetPublicId()
-		res.ScopeId = item.GetProjectId()
-		authorizedActions := authResults.FetchActionSetForId(ctx, item.GetPublicId(), IdActions, auth.WithResource(&res)).Strings()
-		if len(authorizedActions) == 0 {
-			continue
+	var listResp *pagination.ListResponse[*session.Session]
+	sortBy := "created_time"
```
johanbrandhorst, line 209: We changed this in the target PR to set in each if statement explicitly, do you mind making that change here?

internal/daemon/controller/handlers/sessions/session_service_test.go
```diff
@@ -448,6 +452,15 @@ func TestList(t *testing.T) {
 		wantIncludeTerminatedSessions = append(wantIncludeTerminatedSessions, expected)
 	}
 
+	// Run analyze to update postgres estimates
+	_, err = sqlDB.ExecContext(ctx, "analyze")
+	require.NoError(t, err)
+
+	slices.Reverse(wantSession)
```
johanbrandhorst, line 459: Mind adding a comment to explain why they are reversed?

internal/daemon/controller/handlers/sessions/session_service_test.go
```diff
@@ -590,6 +695,338 @@ func TestList(t *testing.T) {
...
+	// Compare without comparing the list token
+	assert.Empty(t,
+		cmp.Diff(
+			got,
+			&pbs.ListSessionsResponse{
+				Items:        []*pb.Session{allSessions[0]},
+				ResponseType: "delta",
+				SortBy:       "updated_time",
+				SortDir:      "desc",
+				// Should contain the deleted target
```
johanbrandhorst, line 953: ⏰ 
```suggestion
				// Should contain the deleted session
```

internal/db/schema/migrations/oss/postgres/79/03_session_base_table_updates.up.sql
```diff
@@ -0,0 +1,14 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+
+  -- Add new indexes for the update time queries.
```
johanbrandhorst, line 6:
```suggestion
  -- Add new indexes for the create time and update time queries.
```

internal/session/query.go
```diff
@@ -422,6 +422,69 @@ update session_credential
...
+	  from session
+	   -- search condition for applying permissions is constructed
+	 where %s
+  order by create_time desc, public_id asc
+	 limit %d
+)
+   select *
+     from session_list
+    where session_list.public_id in (select * from session_ids)
+ order by create_time desc, public_id asc;
```
johanbrandhorst, line 438: It took me a while to see it, but I like @tmessi's query formatting - in this CTE, we start by indenting the `order by` by 2 spaces, and then align all the other keywords (select, from, where, limit) with the end of that string, so you end up with a vertical space between key words and values. I think maybe yours was correct but with tabs instead of spaces, which I think we should avoid in string literals.
```suggestion
with session_ids as (
    select public_id
      from session
           -- search condition for applying permissions is constructed
     where %s
  order by create_time desc, public_id asc
     limit %d
)
   select *
     from session_list
    where session_list.public_id in (select * from session_ids)
 order by create_time desc, public_id asc;
```

internal/session/repository_session.go
```diff
@@ -274,71 +274,173 @@ func (r *Repository) LookupSession(ctx context.Context, sessionId string, opt ..
...
+		return nil, time.Time{}, nil
+	}
+
 	opts := getOpts(opt...)
 
+	var permissionWhereClause string
+	if len(where) > 0 {
+		permissionWhereClause = "(" + strings.Join(where, " or ") + ")"
+		if !opts.withTerminated {
+			permissionWhereClause += " and termination_reason is null "
```
johanbrandhorst, line 297: Do we actually need a space at the end of these?

internal/session/testing.go
```diff
@@ -273,3 +273,31 @@ func TestTofu(t testing.TB) []byte {
 func TestCert(jobId string) (ed25519.PrivateKey, []byte, error) {
 	return newCert(context.Background(), jobId, []string{"127.0.0.1", "localhost"}, time.Now().Add(5*time.Minute), rand.Reader)
 }
+
+// TestListSessions exposes the repo listSessions method for testing purposes.
+func TestListSessions(t testing.TB, repo *Repository, ctx context.Context, opts ...Option) ([]*Session, time.Time) {
+	targets, ttime, err := repo.listSessions(ctx, opts...)
+	require.NoError(t, err)
+	return targets, ttime
+}
```
johanbrandhorst, line 282: Do we need these functions? There's no risk of import cycles in this hierarchy. We can just have an internal test file that tests the unexported methods?

johanbrandhorst (changes requested):
This looks great, thanks, just some minor comments

johanbrandhorst (approved):
Fix the build error and 🚢

PR #4089, opened 2023-12-01
ratelimit: Emit a sys event and allow disabling the limiter

tmessi:
d7373ae03 feat(controller): Emit sys event detailing the rate limit configuration

A sys event will be emitted when the controller is constructed, and if the
rate limits are changed via a SIGHUP. The event contains the details of the
rate limits for each combination of resource and action. This event can be
used by operators to confirm the limits are set as they would expect based
on the controller config.

092ab0c53 feat(ratelimier): Support disabling rate limiting

b11847260 refact(ratelimiter): Updates due to refactor in go-rate

### DIFF
--- a/internal/daemon/controller/rate_limiter_test.go
+++ b/internal/daemon/controller/rate_limiter_test.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"sync"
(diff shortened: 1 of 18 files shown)


internal/daemon/controller/rate_limiter_test.go
```diff
@@ -0,0 +1,672 @@
...
+			err = wantDecoder.Decode(want)
+			require.NoError(t, err)
+
+			rlc, err := newRateLimiterConfig(testCtx, tc.configs, tc.maxSize, tc.disabled)
+			require.NoError(t, err)
+
+			rlc.writeSysEvent(testCtx)
+
+			defer func() { _ = os.WriteFile(tc.sinkFileName, nil, 0o666) }()
+			b, err := ioutil.ReadFile(tc.sinkFileName)
```
mgaffney, line 612: `ioutil.ReadFile` is deprecated

mgaffney (approved):
Nice work!

PR #4093, opened 2023-12-04
feat(api): Use Retry-After header when retrying a request

tmessi:
If the api.Client receives a 429 or 503 along with a Retry-After header,
the client will automatically wait the amount of time specified, and
then re-try the request.

Note that by default the boundary cli does not perform any retries. In
order for the cli to automatically retry when rate limited, users must
set the `BOUNDARY_MAX_RETRIES` environment variable.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -136,6 +136,26 @@ type TLSConfig struct {
 	Insecure bool
 }
 
+// RateLimitLinearJitterBackoff wraps the retryablehttp.LinearJitterBackoff.
+// It first checks if the response status code is http.StatusTooManyRequests
+// (HTTP Code 429) or http.StatusServiceUnavailable (HTTP Code 503). If it is
+// and the response contains a Retry-After response header, it will wait the
+// amount of time specified by the header. Otherwise, this calls
+// LinearJitterBackoff.
+// See: https://pkg.go.dev/github.com/hashicorp/go-retryablehttp#LinearJitterBackoff
+func RateLimitLinearJitterBackoff(min, max time.Duration, attemptNum int, resp *http.Response) time.Duration {
+	if resp != nil {
+		if resp.StatusCode == http.StatusTooManyRequests || resp.StatusCode == http.StatusServiceUnavailable {
+			if s, ok := resp.Header["Retry-After"]; ok {
+				if sleep, err := strconv.ParseInt(s[0], 10, 64); err == nil {
+					return time.Second * time.Duration(sleep)
+				}
+			}
+		}
+	}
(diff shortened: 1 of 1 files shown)


moduli (commented):
Question... now that we learned we need `BOUNDARY_MAX_RETRIES`, is this still needed? I tried running this against the current `llb-rate-limiting` and it looks like it's retrying on the 429
```
❯ go test -v -count=1  github.com/hashicorp/boundary/testing/internal/e2e/tests/base -run TestCliRateLimit
=== RUN   TestCliRateLimit
    scope.go:60: Created Org Id: o_3gdhcPM5SW
    scope.go:93: Created Project Id: p_AEZrW0Kcfe
    target.go:101: Created Target: ttcp_YIsSjguqGC
    rate_limit_test.go:182: 102.518791ms
    rate_limit_test.go:186: 96.437709ms
    rate_limit_test.go:190: 91.479375ms
    rate_limit_test.go:194: 89.686875ms
    rate_limit_test.go:198: 91.08225ms
    rate_limit_test.go:202: 3.749644s
--- PASS: TestCliRateLimit (6.94s)
```

tmessi:
> Question... now that we learned we need `BOUNDARY_MAX_RETRIES`, is this still needed? I tried running this against the current `llb-rate-limiting` and it looks like it's retrying on the 429
> 
> ```
> ❯ go test -v -count=1  github.com/hashicorp/boundary/testing/internal/e2e/tests/base -run TestCliRateLimit
> === RUN   TestCliRateLimit
>     scope.go:60: Created Org Id: o_3gdhcPM5SW
>     scope.go:93: Created Project Id: p_AEZrW0Kcfe
>     target.go:101: Created Target: ttcp_YIsSjguqGC
>     rate_limit_test.go:182: 102.518791ms
>     rate_limit_test.go:186: 96.437709ms
>     rate_limit_test.go:190: 91.479375ms
>     rate_limit_test.go:194: 89.686875ms
>     rate_limit_test.go:198: 91.08225ms
>     rate_limit_test.go:202: 3.749644s
> --- PASS: TestCliRateLimit (6.94s)
> ```

I believe this happens to pass because of the small limit period you picked. Without this change the client just uses the Jitter retry. It does not look at the `Retry-After` header. And the jitter picks a random time between the min and max, which based on [this](https://github.com/hashicorp/boundary/blob/da5dd383ad08d3977bdf94a40158526238b0f514/api/client.go#L807-L808) seems to be a time between 1-1.5 seconds.

moduli:
Cool. Just checking

moduli (approved):

mgaffney (approved):

PR #4095, opened 2023-12-04
Change session list queries to avoid using the session_list view

mikemountain:


internal/session/query.go
```diff
@@ -424,64 +424,260 @@ where session_id = ?
...
+          s.key_id,
+          s.termination_reason,
+          s.version,
+          s.create_time,
+          s.update_time,
+          s.endpoint,
+          s.worker_filter,
+          s.egress_worker_filter,
+          s.ingress_worker_filter,
+          swp.worker_id,
```
johanbrandhorst, line 478: As @tmessi mentioned, it seems we could drop this from the query as it's not used in constructing the session: https://github.com/hashicorp/boundary/blob/da5dd383ad08d3977bdf94a40158526238b0f514/internal/session/repository.go#L143-L164. That means we could drop the whole `session_worker_protocols` CTE. Could you do a pass over the other columns here and see if there's anything else we can drop?

johanbrandhorst (changes requested):

johanbrandhorst (approved):
🎉 nice!

tmessi (changes requested):
I think we should also update the [`sessionListView` struct](https://github.com/hashicorp/boundary/blob/e108fd32d0ca2aedc4a79caeaf7ae08d2f6561fe/internal/session/session.go#L567-L596) to:

- Remove `ProtocolWorkerId` since it is now not being populated, and does not seem to be used anywhere already.
- Rename it since it is no longer used with a view.

johanbrandhorst (changes requested):
See my comment

internal/db/schema/migrations/oss/postgres/80/10_session_base_table_updates.up.sql
```diff
@@ -0,0 +1,20 @@
...
+                        user_id,
+                        termination_reason);
+  create index session_update_time_list_idx
+            on session (update_time desc,
+                        public_id,
+                        project_id,
+                        user_id,
+                        termination_reason);
+
+commit;
```
johanbrandhorst, line 20: Do you need to add/update the session index SQL test?

johanbrandhorst (approved):
LGTM, thanks!

johanbrandhorst (changes requested):
Lets wait with merging this until we have performed all the session performance tests.

johanbrandhorst (approved):

tmessi (approved):

PR #4108, opened 2023-12-06
fix(db): make bucket_prefix column immutable for storage buckets

ddebko:
# Summary

The storage domain does not allow the storage bucket prefix value to be mutated during an update action because it is not supported in the mask. However the column is mutable when directly interacting with the database table. This PR updates the `immutable_columns` trigger on the `storage_plugin_storage_bucket` table to include `bucket_prefix`.

### DIFF
--- a/internal/db/schema/migrations/oss/postgres/80/01_storage_bucket.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/80/01_storage_bucket.up.sql
@@ -1,0 +1,11 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+
+drop trigger immutable_columns on storage_plugin_storage_bucket;
+
+create trigger immutable_columns before update on storage_plugin_storage_bucket
+  for each row execute procedure immutable_columns('public_id', 'scope_id', 'create_time', 'bucket_name', 'bucket_prefix');
+
+commit;


elimt (approved):

internal/db/schema/migrations/oss/postgres/80/01_storage_bucket.up.sql
```diff
@@ -0,0 +1,11 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+
+drop trigger if exists immutable_columns ON storage_plugin_storage_bucket;
```
mgaffney, line 6:
```suggestion
drop trigger immutable_columns on storage_plugin_storage_bucket;
```
The trigger should exist and ff it doesn't we want to error out.

internal/db/schema/migrations/oss/postgres/80/01_storage_bucket.up.sql
```diff
@@ -0,0 +1,11 @@
...
+
+create trigger immutable_columns before update on storage_plugin_storage_bucket
+    for each row execute procedure immutable_columns('public_id', 'scope_id', 'create_time', 'bucket_name', 'bucket_prefix');
```
mgaffney, line 9: Please fix the indentation to follow our style guide

mgaffney (changes requested):

mgaffney (approved):

PR #4113, opened 2023-12-06
Auth token pagination

mikemountain:
Add support for auth token list pagination


internal/authtoken/repository.go
```diff
@@ -290,29 +292,147 @@ func (r *Repository) ValidateToken(ctx context.Context, id, token string, opt ..
...
-	var atvs []*authTokenView
-	if err := r.reader.SearchWhere(ctx, &atvs, "auth_account_id in (select public_id from auth_account where scope_id in (?))", []any{withScopeIds}, db.WithLimit(opts.withLimit)); err != nil {
-		return nil, errors.Wrap(ctx, err, op)
+	limit := r.limit
+	if opts.withLimit != 0 {
+		// non-zero signals an override of the default limit for the repo.
+		limit = opts.withLimit
+	}
+
+	whereScopeIds := "('" + strings.Join(withScopeIds, "', '") + "')"
```
mikemountain, line 310: @johanbrandhorst I was between this method and looping through `withScopeIds` to turn them into `@scope_id_%d` args, do you have a preference?

internal/authtoken/authtoken.go
```diff
@@ -167,3 +169,40 @@ func EncryptToken(ctx context.Context, kmsCache *kms.Kms, scopeId, publicId, tok
...
+	return ""
+}
+
+func (at AuthToken) GetName() string {
+	return ""
+}
+
+func (at AuthToken) GetVersion() uint32 {
+	return 0
+}
```
johanbrandhorst, line 198: Could you add similar comments to these two too?

internal/authtoken/query.go
```diff
@@ -0,0 +1,138 @@
...
+           aa.iam_user_scope_id,
+           at.create_time,
+           at.update_time,
+           at.approximate_last_access_time,
+           at.expiration_time,
+           at.status 
+      from auth_token at
+      join auth_accounts aa on aa.public_id = at.auth_account_id
+  order by at.create_time desc, at.public_id asc
+     limit %d
```
johanbrandhorst, line 35: These two should be in the first CTE to make this efficient, I think. This might need to be something like 

```sql
with auth_tokens as (
  select public_id
   from auth_token
  order by at.create_time desc, at.public_id asc
  limit %d
),
auth_accounts as (
...
from auth_accounts
where %s -- Scope restriction is constructed
),
final as (
...
from auth_tokens at
join auth_accounts aa
)
select *
  from final
order by ...
```
That way, the very root of the query gets the limit applied efficiently using the create_time, public_id index.

internal/authtoken/repository.go
johanbrandhorst: Definitely using args, this here is much more vulnerable to SQL injection. I think the `scope_id in` part should also be in here somewhere.

johanbrandhorst: Credential libraries has an example of this: https://github.com/hashicorp/boundary/pull/4090/files#diff-edc9e7d38750a08b4a30397ed90cfe1ebb2398c64e9ebb2d7009356d9e2f4996

internal/authtoken/repository.go
```diff
@@ -290,29 +292,147 @@ func (r *Repository) ValidateToken(ctx context.Context, id, token string, opt ..
...
+func (r *Repository) listAuthTokensRefresh(ctx context.Context, updatedAfter time.Time, withScopeIds []string, opt ...Option) ([]*AuthToken, time.Time, error) {
+	const op = "authtoken.(Repository).listAuthTokensRefresh"
+
+	if updatedAfter.IsZero() {
+		return nil, time.Time{}, errors.New(ctx, errors.InvalidParameter, op, "missing updated after time")
+	}
+
+	if len(withScopeIds) == 0 {
+		return nil, time.Time{}, errors.New(ctx, errors.InvalidParameter, op, "missing scope id")
+	}
```
johanbrandhorst, line 336: Can we make this a switch?

johanbrandhorst (commented):
Looks great! I think we can improve the queries slightly

internal/authtoken/query.go
```diff
@@ -0,0 +1,178 @@
...
+      limit %d
+),
+auth_accounts as (
+    select public_id,
+           auth_method_id,
+           scope_id,
+           iam_user_id,
+           iam_user_scope_id
+      from auth_account
+     where %s
```
johanbrandhorst, line 30: Nice, this is better, but I think we can do one better with 
```suggestion
     where %s
       and public_id in (select auth_account_id from auth_tokens)
```
What do you think? That will only select the auth accounts that we actually need.

johanbrandhorst (approved):

PR #4114, opened 2023-12-06
Credential stores pagination

johanbrandhorst:
Add pagination of credential stores. Note the use of the registration pattern for subtypes and the querying happening at the domain package level.


internal/credential/repository_store.go
```diff
@@ -0,0 +1,190 @@
...
+type SubtypeStoreService interface {
+	ListDeletedStoreIds(context.Context, time.Time, ...Option) ([]string, error)
+}
+
+// StoreRepository coordinates calls across different subtype services
+// to gather information about all credential stores.
+type StoreRepository struct {
+	services []SubtypeStoreService
+	writer   db.Writer
+}
```
johanbrandhorst, line 32: I called this a repository now, since it accesses the DB. I had a look at the DDD definitions of Service and Repository and this feels more like a repository now. Thoughts appreciated.

internal/credential/repository_store.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/hashicorp/boundary/internal/db/timestamp"
+	"github.com/hashicorp/boundary/internal/errors"
+	"github.com/hashicorp/boundary/internal/pagination"
+	"github.com/hashicorp/boundary/internal/util"
+)
+
+// SubtypeStoreService defines the interface used to
+// list deleted store IDs of all subtypes.
+type SubtypeStoreService interface {
+	ListDeletedStoreIds(context.Context, time.Time, ...Option) ([]string, error)
```
johanbrandhorst, line 24: Deleted ID listing is still performed per-subtype as the deleted ID tables are per-subtype.

internal/credential/repository_store.go
```diff
@@ -0,0 +1,190 @@
...
+		return 0, errors.Wrap(ctx, err, op, errors.WithMsg("failed to query total credential stores"))
+	}
+	var count int
+	for rows.Next() {
+		if err := s.writer.ScanRows(ctx, rows, &count); err != nil {
+			return 0, errors.Wrap(ctx, err, op, errors.WithMsg("failed to query total credential stores"))
+		}
+	}
+	return count, nil
+}
```
johanbrandhorst, line 134: This could also be per-subtype but TBH I didn't think it was worth having this done per subtype-table when this package has to know about both the subtype tables for the list query anyway. This is also easier to do atomically in this package (no need to pass around transactions). Thoughts appreciated.

internal/credential/registry.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/hashicorp/boundary/globals"
+)
+
+// NewFunc is a function that creates a Store from the provided result.
+type NewFunc func(context.Context, *StoreListQueryResult) (Store, error)
```
tmessi, line 15: Should this be `NewStoreFunc`? And does this need to be exported? It looks like it is only used by unexported functions/methods.

internal/credential/repository_store.go
tmessi: I think we might want to get the deleted ids via a CTE that unions the deleted id tables. That would eliminate the need for this service and an implementation in each subtype, and since we don't need anything about the specific subtypes, just the ids, that seems like a simpler approach.

internal/credential/repository_store.go
tmessi: Yea I think doing this here is the better option.

internal/proto/controller/storage/credential/store/v1/credential.proto
```diff
@@ -58,3 +60,64 @@ message Dynamic {
   // @inject_tag: `gorm:"not_null"`
   string library_id = 2;
 }
+
+message StoreListQueryResult {
```
tmessi, line 64: Do we need this as a proto?

tmessi (changes requested):
Looking pretty good. I have just a few questions/comments.

internal/proto/controller/storage/credential/store/v1/credential.proto
johanbrandhorst: Not necessarily, I just tried to follow the convention, I'm happy to replace this with a struct.

mikemountain (approved):

tmessi (approved):

johanbrandhorst:
@moduli I added an e2e test to this PR

moduli (approved):
e2e test looks good

PR #4125, opened 2023-12-07
test(e2e): Use vault-generic credential library type (#3462)

moduli:
This PR fixes some test failures  in `release/0.13.x` so that it uses the `vault-generic` credential library type (instead of the deprecated `vault` type)

(cherry picked from commit 1cb6743f690d94388e0419aa6bd8c4a4085de4bb)


moduli:
Misunderstood the failures. The failure in `release/0.13.x` is because the migration test uses the latest CLI (which doesn't support the `vault` credential library type) even though `0.13.x` still supports it.

PR #4133, opened 2023-12-08
Add client side caching to enable the boundary search command

talanknight:
Adds the boundary client daemon which caches target and session data from boundary controllers.
This adds a number of new commands including:
boundary daemon start
boundary daemon stop
boundary daemon status
boundary daemon add-token
and
boundary search

The boundary daemon auto starts in the background when running a cli command that interacts with a controller. Running `boundary authenticate` automatically adds the auth token returned to the boundary client daemon unless a flag is passed in to disable that behavior.

Logs, Pid files, and unix sockets are created in a `.boundary` directory located in the user's home directory.


johanbrandhorst (approved):
Amazing!

PR #4136, opened 2023-12-08
Do not cache or allow searching for old controllers

talanknight:
When the client daemon attempts to sync data from a controller that doesn't support list tokens, it marks it as not supported. When a user attempts to search for data from a controller that is not supported, it returns an error.  The client daemon rechecks at some interval (default 1 hour) to see if the controller has been updated to support refresh tokens again.

A small note: A controller that doesn't support refresh tokens, but also doesn't have any resources to cache is not considered unsupported. Only when resources are added to it and the controller does not support refresh tokens is it considered not supported.

A controller is marked as "not supported" in the client daemon by adding a sentinel value in the refresh token table for the resource which it was querying when it saw the lack of a list token.

### DIFF
--- a/internal/clientcache/cmd/daemon/start.go
+++ b/internal/clientcache/cmd/daemon/start.go
@@ -104,26 +104,26 @@ func (c *StartCommand) Flags() *base.FlagSets {
 	f.DurationVar(&base.DurationVar{
 		Name:    "refresh-interval",
 		Target:  &c.flagRefreshInterval,
-		Usage:   `Specifies the interval between refresh token supported cache refreshes. Default: 5 minutes`,
+		Usage:   `Specifies the interval between refresh token supported cache refreshes.`,
 		Default: daemon.DefaultRefreshInterval,
 	})
 	f.DurationVar(&base.DurationVar{
 		Name:    "recheck-support-interval",
 		Target:  &c.flagRecheckSupportInterval,
-		Usage:   `Specifies the interval between checking if a boundary instances is supported when it previously was not. Default: 1 hour`,
+		Usage:   `Specifies the interval between checking if a boundary instances is supported when it previously was not.`,
 		Default: daemon.DefaultRecheckSupportInterval,
 		Hidden:  true,
 	})
 	f.DurationVar(&base.DurationVar{
 		Name:    "max-search-staleness",
 		Target:  &c.flagMaxSearchStaleness,
-		Usage:   `Specifies the duration of time that can pass since the resource was last updated before performing a search waits for the resources being refreshed first. Default: 1 minute`,
+		Usage:   `Specifies the duration of time that can pass since the resource was last updated before performing a search waits for the resources being refreshed first.`,
 		Default: daemon.DefaultSearchStaleness,
 	})
 	f.DurationVar(&base.DurationVar{
 		Name:    "max-search-refresh-timeout",
 		Target:  &c.flagMaxSearchRefreshTimeout,
-		Usage:   `If a search request triggers a best effort refresh, this specifies how long the refresh should run before timing out. Default: 1 second`,
+		Usage:   `If a search request triggers a best effort refresh, this specifies how long the refresh should run before timing out.`,
 		Default: daemon.DefaultSearchRefreshTimeout,
 	})
 	f.BoolVar(&base.BoolVar{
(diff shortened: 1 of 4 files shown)


internal/clientcache/cmd/daemon/start.go
```diff
@@ -104,14 +104,15 @@ func (c *StartCommand) Flags() *base.FlagSets {
...
 		Default: daemon.DefaultRefreshInterval,
 	})
 	f.DurationVar(&base.DurationVar{
-		Name:    "full-fetch-interval",
-		Target:  &c.flagFullFetchInterval,
-		Usage:   `Specifies the interval between full cache refresh for boundary instances which do not support refresh tokens. Default: 5 minutes`,
-		Default: daemon.DefaultFullFetchInterval,
+		Name:    "recheck-support-interval",
+		Target:  &c.flagRecheckSupportInterval,
+		Usage:   `Specifies the interval between checking if a boundary instances is supported when it previously was not. Default: 1 hour`,
```
jefferai, line 113:
```suggestion
		Usage:   `Specifies the interval between checking if a Boundary instance is supported when it previously was not.`,
```

internal/clientcache/cmd/daemon/start.go
```diff
@@ -104,14 +104,15 @@ func (c *StartCommand) Flags() *base.FlagSets {
 	f.DurationVar(&base.DurationVar{
 		Name:    "refresh-interval",
 		Target:  &c.flagRefreshInterval,
-		Usage:   `Specifies the interval between refresh token supported cache refreshes. Default: 1 minute`,
+		Usage:   `Specifies the interval between refresh token supported cache refreshes. Default: 5 minutes`,
```
jefferai, line 107:
```suggestion
		Usage:   `Specifies the interval between refresh token supported cache refreshes.`,
```

jefferai: The CLI library automatically includes the default information so you should never write it manually as it's a good way to get out of sync if the default changes.

internal/clientcache/cmd/daemon/start.go
```diff
@@ -319,8 +320,8 @@ func (c *StartCommand) makeBackground(ctx context.Context, dotDir string) (bool,
 	if c.flagRefreshInterval > 0 && c.flagRefreshInterval != daemon.DefaultRefreshInterval {
 		args = append(args, "-refresh-interval", c.flagRefreshInterval.String())
 	}
-	if c.flagFullFetchInterval > 0 && c.flagFullFetchInterval != daemon.DefaultFullFetchInterval {
-		args = append(args, "-full-fetch-interval", c.flagFullFetchInterval.String())
+	if c.flagRecheckSupportInterval > 0 && c.flagRecheckSupportInterval != daemon.DefaultRecheckSupportInterval {
+		args = append(args, "-full-fetch-interval", c.flagRecheckSupportInterval.String())
 	}
```
jefferai, line 325:
```suggestion
	args = append(args, "-full-fetch-interval", c.flagRecheckSupportInterval.String())
```

jefferai: Since there is a default this is guaranteed to always be populated, any reason to not always pass it in?

internal/clientcache/internal/daemon/search_handler.go
```diff
@@ -123,3 +136,17 @@ func toApiResult(sr *cache.SearchResult) *SearchResult {
 		Sessions: sr.Sessions,
 	}
 }
+
+var errSearchNotSupported = &api.Error{
+	Kind:    "Unsupported Search Request",
+	Message: "The requesting user is for a Boundary instance that doesn't support search. The Boundary instance must support refresh tokens for search to be supported.",
```
jefferai, line 142:
```suggestion
	Message: "The request is for a Boundary instance that doesn't support search. The Boundary instance must support refresh tokens for search to be supported.",
```

internal/clientcache/internal/cache/status.go
```diff
@@ -84,8 +90,22 @@ func (s *StatusService) Status(ctx context.Context) (*Status, error) {
 	}
 	for _, u := range users {
 		us := UserStatus{
-			Id:      u.Id,
-			Address: u.Address,
+			Id: u.Id,
+		}
+
+		{
```
jefferai, line 96: Any reason to have this be in a block? It's part of populating the user status struct above and it's already in a block (the for loop body).

internal/clientcache/internal/daemon/status_handler.go
```diff
@@ -42,12 +42,22 @@ type AuthTokenStatus struct {
 	KeyringlessReferences int    `json:"keyringless_references,omitempty"`
 }
 
+// Information about the boundary controller the data is being cached from.
+type BoundaryStatus struct {
+	// The boundary address for this user
+	Address string `json:"address,omitempty"`
+	// Whether the controller responses are supported by the cache
+	CacheSupport string `json:"cache_response,omitempty"`
```
jefferai, line 50: Mismatch between JSON name and name of the struct member

jefferai (commented):
Made some comments, mostly cleanup things

internal/clientcache/internal/cache/status.go
```diff
@@ -41,12 +41,18 @@ type AuthTokenStatus struct {
 	KeyringlessReferences int
 }
 
+type BoundaryStatus struct {
+	// The boundary address for this user
+	Address          string
```
jefferai, line 46: Just to check, this doesn't end up being serialized and sent back to the user, right?

jefferai: Ah nm, I thought it was also somewhere else and I see it below!

jefferai (approved):

PR #4145, opened 2023-12-12
test(e2e): Modify rate limit test due to timing issues

moduli:
The e2e rate limit test was running into failures in `boundary-enterprise` due to what seems like the GitHub Actions runner taking longer to execute CLI commands. The test had some hard-coded assumptions about timing that were incorrect.

This PR updates the test to not rely on those assumptions.


moduli:
Tested and passed 3 times here: https://github.com/hashicorp/boundary-enterprise/actions/runs/7184112769/job/19566786351

tmessi (approved):

PR #4150, opened 2023-12-13
Some pagination test fixes

johanbrandhorst:
It might be surprising that these were passing, but I think functionally this doesn't change anything since our pagination logic actually mutates the incoming list token directly. Better still to have this look correct to avoid confusion.


mikemountain (approved):
I noticed these too, and had justified it to myself that it should still work regardless 😆 good call on just making it clear across the board. Better to remove confusion.

PR #4157, opened 2023-12-14
internal/handlers: fix some comments

johanbrandhorst:


johanbrandhorst:
I'll do this in #4162

PR #4161, opened 2023-12-14
Add pagination for iam Roles

talanknight:


internal/db/schema/migrations/oss/postgres/80/99_iam_updates.up.sql
johanbrandhorst, line 1: I think I understand why this is 99, but we'll need to line up these migrations before merging the LLB, so I'd sooner we have duplicates (and startup panics) so we don't forget. Therefore, could we make this `08_iam_updates.up.sql`?

internal/iam/query.go
```diff
@@ -176,4 +176,41 @@ const (
...
 	select role_id as role_id, role_scope as scope_id, role_grant as grant from final;
 		`
+
+	listRolesTemplate = `
+	select *
+		from iam_role
+		where %s
+	order by create_time desc, public_id asc
+		limit %d;
+	`
```
johanbrandhorst, line 186: With this simple a query, can we not just use `db.reader.SearchWhere` and `db.WithOrder` and `db.WithLimit`? That means we can skip these queries altogether.

internal/iam/repository_role.go
```diff
@@ -184,16 +187,141 @@ func (r *Repository) DeleteRole(ctx context.Context, withPublicId string, _ ...O
...
+
+	limit := r.defaultLimit
+	if opts.withLimit != 0 {
+		// non-zero signals an override of the default limit for the repo.
+		limit = opts.withLimit
+	}
+
+	var args []any
+	whereClause := "scope_id in @scope_ids"
+	args = append(args, sql.Named("scope_ids", withScopeIds))
```
johanbrandhorst, line 206: Does this work? In my experience I'd had to manually expand slices into multiple arguments when using named parameters.

internal/iam/repository_role.go
```diff
@@ -184,16 +187,141 @@ func (r *Repository) DeleteRole(ctx context.Context, withPublicId string, _ ...O
...
+		defer rows.Close()
+
+		var inRet []*Role
+		for rows.Next() {
+			var rv Role
+			if err := rd.ScanRows(ctx, rows, &rv); err != nil {
+				return errors.Wrap(ctx, err, op, errors.WithMsg("scan row failed"))
+			}
+			inRet = append(inRet, &rv)
+		}
```
johanbrandhorst, line 278: I think we can replace all of this with `rd.SearchWhere(ctx, &ret, whereClause, args, db.WithOrder(bla), db.WithLimit(bla))`?

internal/iam/repository_role_test.go
```diff
@@ -630,21 +633,113 @@ func TestRepository_ListRoles(t *testing.T) {
...
 			assert.Equal(tt.createCnt, len(testRoles))
-			got, err := repo.ListRoles(context.Background(), []string{tt.args.withScopeId}, tt.args.opt...)
+			got, ttime, err := repo.listRoles(context.Background(), []string{tt.args.withScopeId}, tt.args.opt...)
 			if tt.wantErr {
 				require.Error(err)
 				return
 			}
 			require.NoError(err)
-			assert.Equal(tt.wantCnt, len(got))
+			assert.Equal(tt.wantCnt, len(got)) // Transaction timestamp should be within ~10 seconds of now
```
johanbrandhorst, line 650: Mind putting this on the next line?
```suggestion
			assert.Equal(tt.wantCnt, len(got))
			// Transaction timestamp should be within ~10 seconds of now
```

internal/iam/service_list_ext_test.go
```diff
@@ -0,0 +1,583 @@
...
+		// Expect no results.
+		resp6, err := iam.ListRolesRefresh(ctx, []byte("some hash"), 1, filterFunc, resp5.ListToken, repo, []string{org.GetPublicId()})
+		require.NoError(t, err)
+		require.Equal(t, resp6.ListToken.GrantsHash, []byte("some hash"))
+		require.True(t, resp6.CompleteListing)
+		require.Equal(t, resp6.EstimatedItemCount, 5)
+		require.Empty(t, resp6.DeletedIds)
+		require.Empty(t, resp6.Items)
+
+		// Create some new auth tokens
```
johanbrandhorst, line 405: Busted 😂 
```suggestion
		// Create some new roles
```

internal/iam/service_list_ext_test.go
```diff
@@ -0,0 +1,583 @@
...
+		// request a refresh, nothing should be returned
+		resp3, err := iam.ListRolesRefresh(ctx, []byte("some hash"), 1, filterFunc, resp.ListToken, repo, []string{org.GetPublicId()})
+		require.NoError(t, err)
+		require.Equal(t, resp3.ListToken.GrantsHash, []byte("some hash"))
+		require.True(t, resp3.CompleteListing)
+		require.Equal(t, resp3.EstimatedItemCount, 5)
+		require.Empty(t, resp3.DeletedIds)
+		require.Empty(t, resp3.Items)
+
+		// Create some new tokens
```
johanbrandhorst, line 485:
```suggestion
		// Create some new roles
```

johanbrandhorst (changes requested):
Looks great, just some minor suggestions!

internal/iam/repository_role.go
johanbrandhorst: 🤷🏻 I guess it must've been something else that broke for me when I did something like this

johanbrandhorst: I'm gonna submit a PR to replace the instances of the workaround I implemented for this then :P

johanbrandhorst (approved):
LGTM, thanks!

PR #4165, opened 2023-12-15
Auth Method list pagination

mikemountain:


internal/auth/options.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// WithUnauthenticatedUser provides an option for filtering results for
+// an unauthenticated users.
+func WithUnauthenticatedUser(_ context.Context, enabled bool) Option {
+	return func(o *Options) error {
+		o.WithUnauthenticatedUser = enabled
+		return nil
+	}
+}
```
johanbrandhorst, line 84: Do we need this option?

internal/auth/query.go
```diff
@@ -0,0 +1,844 @@
...
+      from auth_password_method_with_is_primary
+     where public_id in (select public_id from auth_methods)
+),
+final as (
+    select public_id,
+           scope_id,
+           case when ldap.is_primary_auth_method is true then
+             'ldap'
+           else null end
+           as primary_auth_method_subtype,
```
johanbrandhorst, line 58: Can't we just leave this as a boolean? This seems unnecessarily complicated
```suggestion
           is_primary_auth_method,
```
If the ldap registered method is called, it'll be because the subtype is ldap, and this will be true or false to indicate whether it's the primary auth method, no?

internal/auth/query.go
```diff
@@ -0,0 +1,844 @@
...
+           version,
+           state,
+           null as start_tls,                    -- Add to make union uniform
+           null as insecure_tls,                 -- Add to make union uniform
+           null as discover_dn,                  -- Add to make union uniform
+           null as anon_group_search,            -- Add to make union uniform
+           null as upn_domain,                   -- Add to make union uniform
+           null as enable_groups,                -- Add to make union uniform
+           null as use_token_groups,             -- Add to make union uniform
+           null as maximum_page_size,            -- Add to make union uniform
```
johanbrandhorst, line 130: I see you got tired of adding these. IMO either do it for every line or just do it for the first in the sequence consistently 😂

internal/auth/repository_auth_method.go
```diff
@@ -0,0 +1,230 @@
...
+
+	var inClauses []string
+	var args []any
+	for i, scopeId := range scopeIds {
+		arg := "scope_id_" + strconv.Itoa(i)
+		inClauses = append(inClauses, "@"+arg)
+		args = append(args, sql.Named(arg, scopeId))
+	}
+	inClause := strings.Join(inClauses, ", ")
+	whereClause := "scope_id in (" + inClause + ")"
```
johanbrandhorst, line 65: Apparently it _is_ possible to do this with
```suggestion
	whereClause := "scope_id in (@scope_ids)"
	args := []any{sql.Named("scope_ids", scopeIds)}
```
. I haven't yet tested it, but it's what Todd did in Roles. I intend to submit a PR to do this for the rest of the places we do this dance so you might want to try it here too.

internal/auth/repository_auth_method.go
```diff
@@ -0,0 +1,230 @@
...
+			return errors.Wrap(ctx, err, op)
+		}
+		defer rows.Close()
+		for rows.Next() {
+			var id string
+			if err := rows.Scan(&id); err != nil {
+				return errors.Wrap(ctx, err, op)
+			}
+			deletedAuthMethodIDs = append(deletedAuthMethodIDs, id)
+		}
```
johanbrandhorst, line 149: Can't we scan directly into `deletedAuthMethodIDs`?
```suggestion
		for rows.Next() {
			if err := r.ScanRows(ctx, &deletedAuthMethodIDs); err != nil {
				return errors.Wrap(ctx, err, op)
			}
		}
```

internal/auth/repository_auth_method.go
```diff
@@ -0,0 +1,230 @@
...
+				}
+			case "oidc":
+				databaseWrapper, err := amr.kms.GetWrapper(ctx, am.ScopeId, kms.KeyPurposeDatabase, kms.WithKeyId(am.KeyId))
+				if err != nil {
+					return errors.Wrap(ctx, err, op, errors.WithMsg("unable to get database wrapper"))
+				}
+				if err := structwrapping.UnwrapStruct(ctx, databaseWrapper, am); err != nil {
+					return errors.Wrap(ctx, err, op, errors.WithCode(errors.Decrypt))
+				}
+			}
```
johanbrandhorst, line 215: Why are we including secrets when listing at all? It's generally the case that we just drop the secrets from the list responses (e.g. see the credential store list lookup skipping the vault token and certificate key, only the HMACs are included: https://github.com/hashicorp/boundary/blob/f19aa9292758312e06c72ae2c6fd9a5db29b1c51/internal/credential/vault/repository_credential_store.go#L271-L289), so I would think we should just remove these fields from the query and struct.

CC @jimlambrt for thoughts

internal/auth/service_list_auth_methods_ext_test.go
```diff
@@ -0,0 +1,685 @@
...
+// 		t.Parallel()
+// 		_, err := auth.NewAuthMethodRepository(ctx, &fakeReader{}, &fakeWriter{}, nil)
+// 		require.Error(t, err)
+// 	})
+// 	t.Run("nil-interface-kms", func(t *testing.T) {
+// 		t.Parallel()
+// 		_, err := auth.NewAuthMethodRepository(ctx, &fakeReader{}, &fakeWriter{}, (*fakeKms)(nil))
+// 		require.Error(t, err)
+// 	})
+// }
```
johanbrandhorst, line 82: :rotating_light:

internal/auth/service_list_auth_methods_ext_test.go
```diff
@@ -0,0 +1,685 @@
...
+
+		// get the rest
+		resp5, err := auth.ListAuthMethodsPage(ctx, []byte("some hash"), 2, filterFunc, resp4.ListToken, repo, []string{org.PublicId})
+		require.NoError(t, err)
+		require.Equal(t, resp5.ListToken.GrantsHash, []byte("some hash"))
+		require.True(t, resp5.CompleteListing)
+		require.Equal(t, resp5.EstimatedItemCount, 6)
+		require.Empty(t, resp5.DeletedIds)
+		require.Len(t, resp5.Items, 2)
+		require.Empty(t, cmp.Diff(resp5.Items[0], ams[4], cmpOpts...))
```
johanbrandhorst, line 474: What about the second item 🤔? This could be
```suggestion
		require.Empty(t, cmp.Diff(resp5.Items, ams[4:], cmpOpts...))
```

internal/auth/service_list_auth_methods_ext_test.go
```diff
@@ -0,0 +1,685 @@
...
+		// Run analyze to update count estimate
+		_, err = sqlDb.ExecContext(ctx, "analyze")
+		require.NoError(t, err)
+
+		resp3, err := auth.ListAuthMethodsRefresh(ctx, []byte("some hash"), 1, filterFunc, resp2.ListToken, repo, []string{org.PublicId})
+		require.NoError(t, err)
+		require.Equal(t, resp3.ListToken.GrantsHash, []byte("some hash"))
+		require.True(t, resp3.CompleteListing)
+		require.Equal(t, resp3.EstimatedItemCount, 2)
+		require.Contains(t, resp3.DeletedIds, deletedAmId4)
```
johanbrandhorst, line 682: This should also contain `deletedAmId3`, no?

internal/auth/service_list_auth_methods_page.go
```diff
@@ -0,0 +1,60 @@
...
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing grants hash")
+	case pageSize < 1:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "page size must be at least 1")
+	case filterItemFn == nil:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing filter item callback")
+	case tok == nil:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing token")
+	case repo == nil:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing repo")
+	}
```
johanbrandhorst, line 43: Validate that `scopeIds` is not empty

internal/auth/service_list_auth_methods.go
```diff
@@ -0,0 +1,47 @@
...
+	const op = "auth.ListAuthMethods"
+
+	switch {
+	case len(grantsHash) == 0:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing grants hash")
+	case pageSize < 1:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "page size must be at least 1")
+	case filterItemFn == nil:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing filter item callback")
+	case scopeIds == nil:
```
johanbrandhorst, line 36: Can we make this compare against the length of the slice instead? This will currently miss an empty, non-nil slice.

internal/auth/service_list_auth_methods_refresh.go
```diff
@@ -0,0 +1,68 @@
...
+	switch {
+	case len(grantsHash) == 0:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing grants hash")
+	case pageSize < 1:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "page size must be at least 1")
+	case filterItemFn == nil:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing filter item callback")
+	case tok == nil:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing token")
+	case scopeIds == nil:
```
johanbrandhorst, line 45: Ditto

internal/auth/service_list_auth_methods_refresh_page.go
```diff
@@ -0,0 +1,73 @@
...
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing grants hash")
+	case pageSize < 1:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "page size must be at least 1")
+	case filterItemFn == nil:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing filter item callback")
+	case tok == nil:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing token")
+	case repo == nil:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing repo")
+	}
```
johanbrandhorst, line 47: Check scopeIds

internal/daemon/controller/handlers/authmethods/authmethod_service_test.go
```diff
@@ -51,6 +55,7 @@ const (
 	testLoginName = "default"
 )
 
+// TODO: fix this
```
johanbrandhorst, line 58: 🚨

internal/daemon/controller/handlers/authmethods/authmethod_service_test.go
```diff
@@ -1550,3 +1599,491 @@ func TestCreate(t *testing.T) {
...
 }
+
+func TestListPagination(t *testing.T) {
+	// Set database read timeout to avoid duplicates in response
+	oldReadTimeout := globals.RefreshReadLookbackDuration
+	globals.RefreshReadLookbackDuration = 0
+	t.Cleanup(func() {
+		globals.RefreshReadLookbackDuration = oldReadTimeout
+	})
+	ctx := context.TODO()
```
johanbrandhorst, line 1610: context.TODO() implies that this will eventually change. I think we always want Background here.
```suggestion
	ctx := context.Background()
```

internal/daemon/controller/handlers/authmethods/authmethod_service_test.go
```diff
@@ -1550,3 +1599,491 @@ func TestCreate(t *testing.T) {
...
+			}),
+			cmpopts.SortSlices(func(a, b protocmp.Message) bool {
+				return a.String() < b.String()
+			}),
+			cmpopts.SortSlices(func(a, b string) bool {
+				return a < b
+			}),
+			cmpopts.SortSlices(func(a, b protocmp.Message) bool {
+				return a.String() < b.String()
+			}),
```
johanbrandhorst, line 2082: 🚨.

This got me thinking though, maybe we should just have a `cmpOpts` variable like we do in the service domain tests for these. Would make this test a bit shorter. What do you think?

internal/db/schema/migrations/oss/postgres/2/10_auth.up.sql
```diff
@@ -27,6 +27,7 @@ where
 -- migration 07_auth.up.sql  This new definition also inserts the sub type's name
 -- into the base type. The name column must be on the base type, so the database
 -- can ensure that auth method names are unique across all sub types.
+-- Replaced in 80/07_auth_method_base_table_updates.up.sql
```
johanbrandhorst, line 30:
```suggestion
-- Replaced in 80/09_auth_method_base_table_updates.up.sql
```

johanbrandhorst (changes requested):
This looks great, some minor comments.

internal/auth/repository_auth_method.go
johanbrandhorst: https://github.com/hashicorp/boundary/pull/4162

internal/auth/service_list_auth_methods_page.go
johanbrandhorst: I think we should also consider adding validation (and testing) for the list token resource type here, similar to what I've added for existing resources in https://github.com/hashicorp/boundary/pull/4162.

internal/auth/repository_auth_method.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	return &AuthMethodRepository{
+		reader: reader,
+		writer: writer,
+		kms:    kms,
+	}, nil
+}
+
+// List lists auth methods across all subtypes.
+func (amr *AuthMethodRepository) List(ctx context.Context, scopeIds []string, afterItem pagination.Item, limit int) ([]AuthMethod, time.Time, error) {
```
tmessi, line 48: Does this need to be exported?

internal/auth/repository_auth_method.go
```diff
@@ -0,0 +1,230 @@
...
+			sql.Named("last_item_create_time", afterItem.GetCreateTime()),
+			sql.Named("last_item_id", afterItem.GetPublicId()),
+		)
+	}
+
+	return amr.queryAuthMethods(ctx, query, args)
+}
+
+// ListRefresh lists auth methods across all subtypes.
+func (amr *AuthMethodRepository) ListRefresh(ctx context.Context, scopeIds []string, updatedAfter time.Time, afterItem pagination.Item, limit int) ([]AuthMethod, time.Time, error) {
```
tmessi, line 80: Does this need to be exported?

internal/auth/repository_auth_method.go
```diff
@@ -0,0 +1,230 @@
...
+	for rows.Next() {
+		if err := amr.reader.ScanRows(ctx, rows, &count); err != nil {
+			return 0, errors.Wrap(ctx, err, op, errors.WithMsg("failed to query total auth methods"))
+		}
+	}
+	return count, nil
+}
+
+// ListDeletedIds lists the deleted IDs of all auth methods since the time specified.
+func (amr *AuthMethodRepository) ListDeletedIds(ctx context.Context, since time.Time) ([]string, time.Time, error) {
```
tmessi, line 133: Does this need to be exported?

tmessi (changes requested):

internal/auth/query.go
mikemountain: I swear I intended to come back and finish them, once I had figured out exactly which ones needed to be added 😂

internal/auth/query.go
mikemountain: There is the possibility of an auth method being stored for a certain subtype while still not being stored as the primary auth method, such as in the sqltests:

`
boundary=# select public_id, scope_id, name from auth_method;
  public_id   |   scope_id   |           name           
--------------+--------------+--------------------------
 apm___colors | o_____colors | Colors Auth Password
 apm_foodtruc | o__foodtruck | Food Truck Auth Password
(2 rows)

boundary=# select is_primary_auth_method, public_id, scope_id, name from auth_password_method_with_is_primary;
 is_primary_auth_method |  public_id   |   scope_id   |           name           
------------------------+--------------+--------------+--------------------------
 f                      | apm___colors | o_____colors | Colors Auth Password
 f                      | apm_foodtruc | o__foodtruck | Food Truck Auth Password
(2 rows)
`

internal/auth/repository_auth_method.go
mikemountain: Leaving them unexported brings in an import cycle in the subtype packages

internal/auth/query.go
johanbrandhorst: I'm not sure why that matters, it would just show up as false here, no?

internal/auth/options.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	return newOpts, nil
+}
+
+// WithLimit provides an option to provide a limit.
+// If WithLimit <= 0, then default limits are used for results.
+func WithLimit(_ context.Context, l int) Option {
+	return func(o *Options) error {
+		o.WithLimit = l
+		return nil
```
johanbrandhorst, line 44: Could we error if `l < 0`?

internal/auth/service_list_auth_methods_ext_test.go
```diff
@@ -0,0 +1,681 @@
...
+		t.Run("missing scope ids", func(t *testing.T) {
+			t.Parallel()
+			filterFunc := func(_ context.Context, am auth.AuthMethod) (bool, error) {
+				return true, nil
+			}
+			tok, err := listtoken.NewPagination(ctx, fiveDaysAgo, resource.AuthMethod, []byte("some hash"), "some-id", fiveDaysAgo)
+			require.NoError(t, err)
+			_, err = auth.ListAuthMethodsPage(ctx, []byte("some hash"), 1, filterFunc, tok, repo, nil, true)
+			require.ErrorContains(t, err, "missing scope ids")
+		})
```
johanbrandhorst, line 272: Add a test with the wrong token resource type

internal/auth/query.go
mikemountain: Got it, misunderstood the purpose of that column

internal/db/schema/migrations/oss/postgres/80/09_auth_method_base_table_updates.up.sql
johanbrandhorst, line 1: This will need to be renumbered to 10 since I merged hosts

johanbrandhorst (changes requested):

internal/auth/options.go
mikemountain: Hmm, we're a bit inconsistent on where we check for the limit and how we handle it across all of pagination - in some domains like auth methods, we'll error in the repository, but in other domains we just override the limit with the repository's default. I think, at the very least, we may as well keep all the checks in the repository methods to keep it consistent in location, for now. Thoughts?

internal/auth/repository_auth_method.go
jimlambrt: I'm curious what @mgaffney thinks.  My opinion: I'm surprised we're returning secrets from an API which aren't HMAC'd

mgaffney: I don't think we should ever return secrets from an API.

internal/db/schema/migrations/oss/postgres/80/10_auth_method_base_table_updates.up.sql
```diff
@@ -0,0 +1,107 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+
+  -- Add create_time, update_time and is_active_public_state to auth_method table.
+  alter table auth_method add column create_time wt_timestamp;
+  alter table auth_method add column update_time wt_timestamp;
+  alter table auth_method add column is_active_public_state boolean;
```
mgaffney, line 9: Why are we allowing this column to be null?

internal/db/schema/migrations/oss/postgres/80/10_auth_method_base_table_updates.up.sql
```diff
@@ -0,0 +1,107 @@
...
+        set create_time            = pw.create_time,
+            update_time            = pw.update_time,
+            is_active_public_state = false
+       from auth_method as am
+  left join auth_password_method as pw
+         on am.public_id = pw.public_id;
+
+  -- Replace the insert trigger to also set the create_time
+  -- Replaces the insert_auth_method_subtype function defined in 2/10_auth.up.sql
+   create or replace function insert_auth_method_subtype() returns trigger
```
mgaffney, line 44:
```suggestion
  create or replace function insert_auth_method_subtype() returns trigger
```

internal/auth/auth_method.go
```diff
@@ -0,0 +1,95 @@
...
+	Urls                     string
+	UserDn                   string
+	UserAttr                 string
+	UserFilter               string
+	EnableGroups             bool
+	UseTokenGroups           bool
+	GroupDn                  string
+	GroupAttr                string
+	GroupFilter              string
+	ClientCertificateKey     []byte
```
johanbrandhorst, line 52: Can we drop this? It looks like a secret to me.

internal/auth/repository_auth_method.go
johanbrandhorst: Seems like we still left some stuff in here that we can remove.

internal/db/schema/migrations/oss/postgres/80/10_auth_method_base_table_updates.up.sql
```diff
@@ -0,0 +1,135 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+
+  -- Add create_time, update_time and is_active_public_state to auth_method table.
+  alter table auth_method add column create_time wt_timestamp;
+  alter table auth_method add column update_time wt_timestamp;
+  alter table auth_method add column is_active_public_state boolean not null;
```
johanbrandhorst, line 9: I don't think you can make this `not null` until after you've done the updates below (unless you also add a default). So normally you'd add this as a boolean here and then do another alter table to set it `not null` after the updates below

johanbrandhorst (commented):
Nice! I think there are a few little things missing still but this is almost there :).

internal/db/schema/migrations/oss/postgres/80/10_auth_method_base_table_updates.up.sql
```diff
@@ -0,0 +1,138 @@
...
+    for each row execute procedure update_auth_method_table_is_active_public_state();
+
+  -- Add new indexes for the create and update time queries.
+  create index auth_method_create_time_public_id_idx
+      on auth_method (create_time desc, public_id asc);
+  create index auth_method_update_time_public_id_idx
+      on auth_method (update_time desc, public_id asc);
+
+  -- Make column not null
+  alter table auth_method alter column is_active_public_state set not null;
```
johanbrandhorst, line 134: Minor nit but maybe move this to just after the data migrations, and also make the comment less generic 😄?

johanbrandhorst (approved):

johanbrandhorst (commented):
There seems to be a test failure

PR #4171, opened 2023-12-18
test(e2e): Temporarily turn off cache daemon in CLI Rate Limit test

moduli:
This PR turns off the client cache daemon during the CLI rate limit e2e test in order to make it pass for now. We will need to remove this once we improve CLI request times when the client cache daemon is enabled.

### DIFF
--- a/testing/internal/e2e/tests/base_plus/rate_limit_test.go
+++ b/testing/internal/e2e/tests/base_plus/rate_limit_test.go
@@ -275,6 +275,14 @@ func TestCliRateLimit(t *testing.T) {
 
 	ctx := context.Background()
 
+	// TODO: Remove this whem CLI request time is improved (when cache daemon is enabled)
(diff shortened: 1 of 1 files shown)


testing/internal/e2e/tests/base_plus/rate_limit_test.go
```diff
@@ -275,6 +275,14 @@ func TestCliRateLimit(t *testing.T) {
 
 	ctx := context.Background()
 
+	// TODO: Remove this whem CLI request time is improved (when cache daemon is enabled)
```
johanbrandhorst, line 278:
```suggestion
	// TODO: Remove this when CLI request time is improved (when cache daemon is enabled)
```
FWIW, I'm not sure it'll get much better than it is right now on the LLB. Todd merged the fix that avoids starting the daemon if it's running, the rest is all keyring access.

moduli: hm...

moduli:
Thinking of another option

PR #4172, opened 2023-12-18
Check in client wrapper if cache daemon is already in use

talanknight:
Starting up a boundary process just to exit takes ~30 ms on my M1. This decreases that time drastically (to <1ms) since it is just a file read and a process lookup.


johanbrandhorst (approved):
👏🏻 awesome!

PR #4182, opened 2023-12-19
simple quote not accepted

zimtls:
Message from terraform init : 
Single quotes are not valid. Use double quotes (") to enclose strings.


jefferai:
Thanks!

PR #4183, opened 2023-12-19
test(e2e): Decrease rate limit in boundary config to resolve failures

moduli:
In `llb-list-pagination2`, TestCliRateLimit was failing for the following reason
```
    rate_limit_test.go:326: Getting rate limit info...
    rate_limit_test.go:343: Sending multiple CLI requests to hit rate limit...
    rate_limit_test.go:347: 1.356015783s
    rate_limit_test.go:347: 1.354126404s
    rate_limit_test.go:347: 1.355090349s
    rate_limit_test.go:347: 1.352092617s
    rate_limit_test.go:347: 1.352088211s
    rate_limit_test.go:347: 1.353169131s
    rate_limit_test.go:354: 
        	Error Trace:	/src/boundary/testing/internal/e2e/tests/base_plus/rate_limit_test.go:354
        	Error:      	An error is expected but got nil.
        	Test:       	TestCliRateLimit
```

`llb-list-pagination` increased CLI request times, which caused `TestCliRateLimit` to send commands after the first quota expires, preventing the test from being rate-limited. The increased CLI request times is due to the client daemon doing an auth token lookup in the keyring. 

This PR updates the controller config used in the test to reduce the limit so that the test will be able to hit the limit still.

### DIFF
--- a/enos/modules/docker_boundary/boundary-config-rate-limit.hcl
+++ b/enos/modules/docker_boundary/boundary-config-rate-limit.hcl
@@ -35,8 +35,8 @@ controller {
     resources = ["host"]
     actions = ["read"]
     per = "auth-token"
-    limit = 5
-    period = "5s"
+    limit = 3
+    period = "10s"
   }
 
   api_rate_limit_max_quotas = 1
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):
Thanks!

PR #4187, opened 2023-12-20
Order public_id descending

johanbrandhorst:
As @talanknight found, ordering the public_id by ascending while using the compound `(create_time, public_id) <` comparison risks missing items with identical update or create times. Order public_id by descending to keep ordering consistent.


mikemountain (approved):
Thanks for this!

johanbrandhorst:
> Most/all of these also have indexes created for these sort orders. Should those be updated also?

The indexes are updated, is it missing any?

PR #4198, opened 2023-12-21
internal/authtoken: join to auth_account to filter

johanbrandhorst:
Instead of doing the scope_id filtering in the second CTE, join the auth_account table to the auth_token table in the query directly.

### DIFF
--- a/internal/authtoken/query.go
+++ b/internal/authtoken/query.go
@@ -8,175 +8,79 @@ const (
 select reltuples::bigint as estimate from pg_class where oid in ('auth_token'::regclass)
 `
 	listAuthTokensTemplate = `
-with auth_tokens as (
-     select public_id,
-            auth_account_id,
-            create_time,
-            update_time,
-            approximate_last_access_time,
-            expiration_time,
-            status
-       from auth_token
-   order by create_time desc, public_id desc
-      limit %d
-),
-auth_accounts as (
-    select public_id,
-           auth_method_id,
-           scope_id,
-           iam_user_id,
-           iam_user_scope_id
-      from auth_account
-     where %s
-       and public_id in (select auth_account_id from auth_tokens)
-),
-final as (
-    select at.public_id,
-           at.auth_account_id,
-           aa.auth_method_id,
-           aa.scope_id,
-           aa.iam_user_id,
-           aa.iam_user_scope_id,
-           at.create_time,
-           at.update_time,
-           at.approximate_last_access_time,
-           at.expiration_time,
-           at.status
-      from auth_tokens at
-      join auth_accounts aa on aa.public_id = at.auth_account_id
-)
-  select *
-    from final
-order by create_time desc, public_id desc;
+  select at.public_id,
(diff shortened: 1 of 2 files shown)


internal/authtoken/query.go
```diff
@@ -10,31 +10,31 @@ select reltuples::bigint as estimate from pg_class where oid in ('auth_token'::r
...
 ),
 auth_accounts as (
     select public_id,
            auth_method_id,
-           scope_id,
            iam_user_id,
            iam_user_scope_id
       from auth_account
-     where %s
-       and public_id in (select auth_account_id from auth_tokens)
```
johanbrandhorst, line 31: This where clause was what concerned me with the old logic. It could lead to auth tokens being excluded after the limit had been applied, which will mess with the pagination logic. Lifting the scope_id into the first CTE applies the limit correctly _after_ excluding tokens from other scopes.

mikemountain (approved):

internal/db/schema/migrations/oss/postgres/80/06_auth_token_base_table_updates.up.sql
```diff
@@ -3,6 +3,53 @@
...
 begin;
 
+  -- Add the scope ID column from the account so we can
+  -- use it in the list queries.
+  -- Note, we cannot make this wt_scope_id as it would require the user
+  -- to provide the scope ID at insert time instead of assigning it
+  -- from the account in a trigger. The problem is that domain type constraints
+  -- run before insert triggers, and the wt_scope_id domain type requires
+  -- the scope ID to be at least 10 characters long, so we can't leave it unset on insert.
+  alter table auth_token add column scope_id text;
```
mgaffney, line 13: Adding `scope_id` to `auth_token` denormalizes the table and nothing in this PR guards against the data anomalies that result from the denormalization. Why not join `auth_token` with `auth_account` in the queries used for listing auth tokens instead?

mgaffney (changes requested):
Adding `scope_id` to the `auth_token` table denormalizes that table

internal/db/schema/migrations/oss/postgres/80/06_auth_token_base_table_updates.up.sql
johanbrandhorst: There is an insert trigger that assigns the scope_id from the auth token (so the user can't set something that isn't the right value), and we make the scope_id immutable after the insert (so the user cannot update it after it's been set), is that not sufficient?

I'm happy to change it to a join in the query but we've been worried that doing a join in the initial CTE would make it less performant. I'll make the update here and we can come back to this option if we find the other one is not performant enough.

johanbrandhorst: I've changed this to join in the query instead

mgaffney (approved):

harshsingh32 (commented):
lgtm

PR #4205, opened 2024-01-02
Manually build the socket url

talanknight:
Url parsing breaks when trying to build a unix socket for windows.  Since we expect unix:// scheme urls to always have the path be an absolute path to the unix socket file, we'll just build the url ourselves always and leave it up to the caller to validate the path is valid.


johanbrandhorst (approved):
Thanks!

PR #4206, opened 2024-01-02
Add more logged events

talanknight:
Logged:
When the list token is invalidated and a fetch is restarted
When an auth token is no longer valid or found in boundary
After resources are updated, logged the deleted and upserted resources for both refreshing the resources and checking if the boundary instance supports search.
Whenever the client cache fails to add an auth token to the cache


johanbrandhorst (approved):
Very nice :)

PR #4207, opened 2024-01-03
Move and export ExpectWorkers

jefferai:
This is a useful test function and keeping it in the cluster package means it can't be used by api/cli tests.


jefferai:
Just a reorg, all test passing, merging

PR #4221, opened 2024-01-04
test(e2e): Increase rate limit in boundary config

johanbrandhorst:
Not sure we'll actually want this but I want to try and see if it'll still pass the tests.


johanbrandhorst:
https://github.com/hashicorp/boundary/actions/runs/7415473226/job/20179010880?pr=4221 looks like we still need this, I'll close this PR.

PR #4230, opened 2024-01-12
Add alias schema changes

talanknight:
This PR adds a postgres extension citext which allows us to have aliases be enforced unique in a case-insensitive way.

While aliases will eventually be allowed in any scope, right now the fk constraint enforces that aliases are only created in the global scope.  The alias_scope_is_parent_to_destination_scope function still is created in a way to allow non global scopes to be used by the alias.

### DIFF
--- a/internal/db/sqltest/tests/alias/targets.sql
+++ b/internal/db/sqltest/tests/alias/targets.sql
@@ -38,7 +38,7 @@ select is(count(*), 0::bigint) from alias_target where public_id = 'alt__t____cb
 select is(count(*), 1::bigint) from alias_target where public_id = 'alt__t____cb' and destination_id is null and host_id is null;
 
 -- validate a host id cant be set if the destination id is not set
-prepare insert_destination_id_not_set_when_host_id_is_set AS
+prepare insert_destination_id_not_set_when_host_id_is_set as
     insert into alias_target (public_id, scope_id, value, host_id)
     values ('unset_destination_id', 'global', 'unset.destination.id', 'h_________cb');
 select throws_like(


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 
 ## Next
 
+### Added dependency
+* postgres citext dependency added to enable aliases to be globally unique
```
irenarindos, line 8: Should this PR also update the system-requirements website doc to include this dependency?

internal/db/schema/migrations/oss/postgres/82/01_citext_extension.up.sql
```diff
@@ -0,0 +1,11 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+
+  -- https://www.postgresql.org/docs/14/citext.html allows us to make
+  -- case-insensitive uniqueness constraints which is useful to us for
+  -- aliases.
+  create extension citext;
```
irenarindos, line 9: If we want to add the citext extension the same way we did for [pgcrypto](https://github.com/hashicorp/boundary/blob/main/internal/db/schema/migrations/oss/postgres/0/60_wh_domain_types.up.sql#L6):
```suggestion
  create extension if not exists "citext";
```

internal/db/schema/migrations/oss/postgres/82/04_alias_target.up.sql
```diff
@@ -0,0 +1,139 @@
...
+  -- Alias delete tracking tables
+  create table alias_target_deleted (
+    public_id wt_public_id primary key,
+    delete_time wt_timestamp not null
+  );
+  comment on table alias_target_deleted is
+  'alias_target_deleted holds the ID and delete_time of every deleted target alias. '
+  'It is automatically trimmed of records older than 30 days by a job.';
+
+  create trigger insert_alias_target_deleted after delete on alias_target
```
irenarindos, line 130: I think this needs to be `insert_deleted_id` - The pgtap tests are failing with
```
/test/purge/deleted_table_tests.sql ..................................................... 1/363 
 Failed test 362: "Table alias_target should have trigger insert_deleted_id"
 Failed test 363: "Trigger insert_deleted_id should call insert_deleted_id()"
         have: NULL
         want: insert_deleted_id
 Looks like you failed 2 tests of 363
```

internal/db/sqltest/tests/alias/wt_target_alias.sql
```diff
@@ -0,0 +1,141 @@
...
+    ('hp--something.test.com'),
+    ('test-for-long-name-which-is-almost-over-the-limit-of-characters'),
+    ('TEST-FOR-LONG-NAME-WHICH-IS-ALMOST-OVER-THE-LIMIT-OF-CHARACTERS'),
+    ('test-for-long-name-which-is-almost-over-the-limit-of-characters.another-label'),
+    ('test.test-for-long-name-which-is-almost-over-the-limit-of-characters'),
+    ('test-for-long-name-which-is-almost-over-the-limit-of-characters.test-for-long-name-which-is-almost-over-the-limit-of-characters'),
+    ('9-things.9-things');
+  select lives_ok('valid_inserts');
+
+  prepare label_to_long as insert into target_alias_testing (v) values
```
irenarindos, line 49: Nit: to -> too here and in the tests below
```suggestion
  prepare label_too_long as insert into target_alias_testing (v) values
```

irenarindos (changes requested):

website/content/docs/install-boundary/system-requirements.mdx
```diff
@@ -154,7 +154,7 @@ When you initialize the database with the `boundary database init` command, the
 
 ### Required PostgreSQL modules
 
-Boundary has a dependency on the PostgreSQL [pgcrypto](https://www.postgresql.org/docs/11/pgcrypto.html) module, which is one of the standard modules that is supplied with PostgreSQL.
+Boundary has a dependency on the PostgreSQL [pgcrypto](https://www.postgresql.org/docs/11/pgcrypto.html) module and [citext](https://www.postgresql.org/docs/11/citext.html), which are part of the standard modules that is supplied with PostgreSQL.
```
irenarindos, line 157:
```suggestion
Boundary has a dependency on the PostgreSQL [pgcrypto](https://www.postgresql.org/docs/11/pgcrypto.html) module and [citext](https://www.postgresql.org/docs/11/citext.html), which are part of the standard modules that are supplied with PostgreSQL.
```

internal/db/sqltest/tests/alias/targets.sql
```diff
@@ -0,0 +1,41 @@
...
+
+select is(count(*), 0::bigint) from alias_target where public_id = 'alt__t____cb' and destination_id = 't_________cb';
+select is(count(*), 1::bigint) from alias_target where public_id = 'alt__t____cb' and destination_id = 't_________cg';
+
+-- validate deleting a target nulls out the destination id
+prepare update_target_alias as
+    delete from target_tcp
+    where public_id = 't_________cg';
+select lives_ok('update_target_alias');
+
```
irenarindos, line 35: Should there also be a test around the destination id being set when host id is set?

irenarindos (changes requested):
Looks good, just two minor things!

mgaffney (changes requested):

internal/db/schema/migrations/oss/postgres/82/01_citext_extension.up.sql
```diff
@@ -0,0 +1,11 @@
+-- Copyright (c) HashiCorp, Inc.
```
irenarindos, line 1: Just dropping a note that these need to be moved from 82 to 84

mgaffney (approved):

PR #4238, opened 2024-01-16
feat (apptoken): add support for periodic expiration interval

jimlambrt:


internal/apptoken/app_token_periodic_expiration.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	return &AppTokenPeriodicExpirationInterval{
+		AppTokenPeriodicExpirationInterval: &store.AppTokenPeriodicExpirationInterval{
+			AppTokenId:                     appTokenId,
+			ExpirationIntervalInMaxSeconds: maxSeconds,
+		},
+	}, nil
+}
+
+// cone an AppTokenPeriodicExpirationInterval
```
cameronperera, line 41:
```suggestion
// clone an AppTokenPeriodicExpirationInterval
```

internal/apptoken/app_token_periodic_expiration_test.go
```diff
@@ -0,0 +1,133 @@
...
+	t.Run("not-equal", func(t *testing.T) {
+		assert, require := assert.New(t), require.New(t)
+		orig, err := NewAppTokenPeriodicExpirationInterval(testCtx, testAppTokenId, uint32(testInterval))
+		require.NoError(err)
+		orig2, err := NewAppTokenPeriodicExpirationInterval(testCtx, testAppTokenId, uint32(testInterval+2))
+		require.NoError(err)
+
+		cp := orig.clone()
+		assert.True(!proto.Equal(cp.AppTokenPeriodicExpirationInterval, orig2.AppTokenPeriodicExpirationInterval))
+	})
```
cameronperera, line 101: Non-blocking:
I just want to make sure I am understanding this test. We create 2 in-memory items, clone the first one, then finally assert that the clone is not referencing the wrong in-memory item?

cameronperera (approved):
LGTM. I left a comment about one of the test and a spelling change. All non-blocking

internal/apptoken/app_token_periodic_expiration.go
jimlambrt: ty!

internal/apptoken/app_token_periodic_expiration_test.go
jimlambrt: correct.

internal/apptoken/options.go
```diff
@@ -55,3 +58,17 @@ func WithLimit(_ context.Context, limit int) Option {
...
+
+// WithExpirationInterval provides an option to specify an expiration interval
+// for an AppToken
+func WithExpirationInterval(ctx context.Context, maxSeconds uint32) Option {
+	const op = "apptoken.WithExpirationInterval"
+	return func(o *options) error {
+		switch {
+		case maxSeconds == 0:
+			return errors.New(ctx, errors.InvalidParameter, op, "invalid expiration interval (equal to zero)")
+		}
```
uruemu, line 70: Curious why we return an error here even when it is a valid case considering it is an optional param. Is it so the caller doesn't unintentionally pass `0` value?

uruemu (approved):
LGTM — Left a comment which is non-blocking but would like to understand the decision

internal/apptoken/options.go
jimlambrt: correct.  The caller is intentionally specifying an interval and zero isn't a
valid interval.

PR #4243, opened 2024-01-17
Fix Unix listener causing 500s on Mac

jefferai:
Some platforms use `@` for a Unix socket address, and some leave it empty. In the empty case this causes the rate limiting code to throw a 500 error. This commit standardizes on using `@` for incoming connections over a Unix socket.

### DIFF
--- a/internal/daemon/common/client_ip.go
+++ b/internal/daemon/common/client_ip.go
@@ -34,7 +34,7 @@ func ClientIpFromRequest(ctx context.Context, listenerCfg *listenerutil.Listener
 		if err != nil {
 			return "", errors.Wrap(ctx, err, op)
 		}
-		if listenerCfg.Type == "unix" {
+		if listenerCfg.Type == "unix" && ip == "" {
 			// Some platforms (Linux) use "@" in this case but some like Mac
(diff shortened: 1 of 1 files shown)


internal/daemon/common/client_ip.go
```diff
@@ -34,6 +34,12 @@ func ClientIpFromRequest(ctx context.Context, listenerCfg *listenerutil.Listener
 		if err != nil {
 			return "", errors.Wrap(ctx, err, op)
 		}
+		if listenerCfg.Type == "unix" {
+			// Some platforms (Linux) use "@" in this case but some like Mac
+			// leave it empty which causes issues with the rate limiting logic,
+			// so standardize on "@" in this case.
+			return "@", nil
+		}
```
tmessi, line 42: Should we only do this if `ip` is `""`? ie:

```golang
if listenerCfg.Type == "unix" && ip == "" {
```

It probably results in the same thing, but it would prevent us from changing the value if for some reason it gets set to something other than `@` by some platform either now or in the future, not that it should really matter much.

jefferai: Sure, will change.

tmessi (approved):

PR #4247, opened 2024-01-17
cmd: default dev postgres version to 16

johanbrandhorst:
This changes the default boundary dev postgres version from 12 to 16 due to observed responsiveness differences between 11 (the old default) and 12.


johanbrandhorst:
We're not going down this path after all

PR #4252, opened 2024-01-18
Use correct argument name in log line

davidsbond:
This commit modifies a log line produced in the WorkerStatusReport function when closing orphaned connections that appears to incorrectly label a worker identifier as a controller identifier.


johanbrandhorst (approved):
oops

louisruch (approved):

PR #4254, opened 2024-01-18
Disable postgres JIT for boudary dev

tmessi:
After updating `boundary dev` to use postgres:12 by default, it was noticed
that the responsiveness of boundary was noticeably slower. This is due to
the introduction of JIT being enabled by default in postgres:12. JIT seems
to incur noticeable overhead without providing benefits, at least for the
instance used for `boundary dev`.

It is also worth noting that for postgres:14 and newer, JIT does not get
triggered for the same queries since they seem to result in a cheaper query
plan that does not exceed the default threshold for JIT.

This also updates the docker image used by tests to also disable JIT,
and adds additional docker images for tests for other postgres versions.

### DIFF
--- a/internal/cmd/base/internal/docker/supported.go
+++ b/internal/cmd/base/internal/docker/supported.go
@@ -41,6 +41,16 @@ func startDbInDockerSupported(dialect string, opt ...Option) (cleanup func() err
 		}
 	}
 
+	runOpts := &dockertest.RunOptions{
+		Tag: tag,
+		Env: []string{"POSTGRES_PASSWORD=password", "POSTGRES_DB=boundary"},
+		Cmd: []string{
+			// JIT seems to cause noticeable overhead without providing noticeable benefit.
+			// See: ICU-12283
+			"-c", "jit=off",
+		},
+	}
+
 	switch dialect {
 	case "postgres", "pgx":
 		switch {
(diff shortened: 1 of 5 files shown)


mgaffney (approved):

johanbrandhorst (approved):
Nice!

internal/cmd/base/internal/docker/supported.go
```diff
@@ -41,20 +41,32 @@ func startDbInDockerSupported(dialect string, opt ...Option) (cleanup func() err
 		}
 	}
 
+	runOpts := &dockertest.RunOptions{
+		Tag: tag,
+		Env: []string{"POSTGRES_PASSWORD=password", "POSTGRES_DB=boundary"},
+		Cmd: []string{
+			// JIT seems to cause noticeable overhead without providing noticeable benefit.
```
jefferai, line 48:
```suggestion
			// JIT seems to cause noticeable overhead without providing noticeable benefit, at least prior to Postgres 14.
```
Just a note in case we want to try it out again as we move to later versions, where possibly it actually becomes an improvement.

jefferai (approved):

internal/cmd/base/internal/docker/supported.go
tmessi: I think JIT would be the same for postgres 14, it just so happens that starting with 14 a different plan is chosen that has a lower cost and so it does not trigger JIT. I can try it out by lowering the threshold so JIT is triggers with the plan for postgres 14 and see if it still produces the same overhead.

tmessi: Yea, just confirmed that I get the same slowness by lowering `jit_above_cost` so that JIT runs for the plan produced by postgres 14.

PR #4271, opened 2024-01-23
No longer enforce version matching between daemon and cli

talanknight:
I have removed the version matching check between daemon and cli.  The reasoning is that a lot of actions happen behind the scenes through the command wrapper and it would be unfortunate for a user to update to a new version (maybe just a patch release) and have all the background actions break without the user noticing, then when they want to search realize that the versions don't match (because of the version check).

Instead I have changed the behavior to always include the version header both from the client and from the daemon.  Additionally the daemon reports whether it is running in background mode.

This PR also adds a logging handler which, when called, logs something to the daemon's event logger.

Here is the output using -output-curl-string from the daemon and in response from the daemon
```
> boundary daemon status -output-curl-string
curl --unix-socket /Users/tak/.boundary/socket/daemon.sock -H "Boundary_version: 0.15.0" -H "Content-Type: application/json" 'http://clientcache.boundary.localhost/v1/status'
> curl --unix-socket /Users/tak/.boundary/socket/daemon.sock -v -H "Boundary_version: 0.15.0" -H "Content-Type: application/json" 'http://clientcache.boundary.localhost/v1/status'
*   Trying /Users/tak/.boundary/socket/daemon.sock:0...
* Connected to clientcache.boundary.localhost (/Users/tak/.boundary/socket/daemon.sock) port 80 (#0)
> GET /v1/status HTTP/1.1
> Host: clientcache.boundary.localhost
> User-Agent: curl/8.1.2
> Accept: */*
> Boundary_version: 0.15.0
> Content-Type: application/json
> 
< HTTP/1.1 200 OK
< Background: true
< Boundary_version: 0.15.0
< Content-Type: application/json
< Date: Wed, 24 Jan 2024 21:47:26 GMT
< Content-Length: 557
< 
* Connection #0 to host clientcache.boundary.localhost left intact
{"uptime":3700195639667,"socket_address":"/Users/tak/.boundary/socket/daemon.sock","log_location":"/Users/tak/.boundary/cache.log","version":"Boundary v0.15.0 (5c8939049fda97493feb4f8dbea61eac92b1070b+CHANGES)","users":[{"id":"u_1234567890","boundary_instance":{"address":"http://127.0.0.1:9200","cache_support":"Supported","last_support_check":825136303000},"auth_tokens":[{"id":"at_GJtLwb92ui","keyring_references":1}],"resources":[{"name":"target","count":2,"refresh_token":{"age":3699955472000,"last_used":825136472000}},{"name":"session","count":0}]}]}% 
```

### DIFF
--- a/internal/clientcache/internal/daemon/log_handler_test.go
+++ b/internal/clientcache/internal/daemon/log_handler_test.go
@@ -110,6 +110,7 @@ func TestLogHandler(t *testing.T) {
 		require.NoError(t, err)
 		gotEvent := &cloudevents.Event{}
 		err = json.Unmarshal(b, gotEvent)
+		require.NoError(t, err)
 
 		gotData := gotEvent.Data.(map[string]any)
 		assert.NotEmpty(t, gotData["error"])


jimlambrt (approved):
Please follow up with product as well

jimlambrt (approved):
ty!

PR #4274, opened 2024-01-24
test(grant): Use ID to validate version check

louisruch:


internal/tests/cluster/recursive_anon_listing_test.go
```diff
@@ -32,7 +32,7 @@ func TestListAnonymousRecursing(t *testing.T) {
 	require.NoError(err)
 	customRole, err = rolesClient.AddPrincipals(tc.Context(), customRole.Item.Id, customRole.Item.Version, []string{"u_anon"})
 	require.NoError(err)
-	_, err = rolesClient.AddGrants(tc.Context(), customRole.Item.Id, customRole.Item.Version, []string{"id=*;type=auth-method;actions=list,authenticate"})
+	_, err = rolesClient.AddGrants(tc.Context(), customRole.Item.Id, customRole.Item.Version, []string{"ids=*;type=auth-method;actions=list,authenticate"})
```
louisruch, line 35: Looks like this was added yesterday using `id` instead of `ids` which started failing at 0.15.0 as well :)

jefferai (approved):

PR #4277, opened 2024-01-24
Add alias resolution to target service

talanknight:


internal/proto/controller/api/resources/targets/v1/target.proto
```diff
@@ -14,6 +14,14 @@ import "google/protobuf/wrappers.proto";
 
 option go_package = "github.com/hashicorp/boundary/sdk/pbs/controller/api/resources/targets;targets";
 
+message Alias {
+  // Output only. The ID of the alias refercing this target.
```
irenarindos, line 18:
```suggestion
  // Output only. The ID of the alias referencing this target.
```

internal/proto/controller/api/resources/targets/v1/target.proto
```diff
@@ -211,6 +219,9 @@ message Target {
     }
   ]; // @gotags: `class:"public"`
 
+  // Output only. The aliases tht point to this Target.
```
irenarindos, line 222:
```suggestion
  // Output only. The aliases that point to this Target.
```

internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -1232,6 +1322,32 @@ func (s Service) getFromRepo(ctx context.Context, id string) (target.Target, []t
 	return u, hs, cl, nil
 }
 
+// resolveAlias replaces the value of in with the destination id of the alias if
```
irenarindos, line 1325: Subbing out "value" for "in" might improve readability here
```suggestion
// resolveAlias returns the alias associated with in if the alias is
```

internal/daemon/controller/controller.go
```diff
@@ -147,6 +147,7 @@ type Controller struct {
 	PluginRepoFn              common.PluginRepoFactory
 	TargetRepoFn              target.RepositoryFactory
 	WorkerAuthRepoStorageFn   common.WorkerAuthRepoStorageFactory
+	TargetAliasRepoFn         common.TargetAliasRepoFactory
 
 	scheduler *scheduler.Scheduler
```
irenarindos, line 152: Need to also add this around line 452:
```	
c.TargetAliasRepoFn = func() (*target2.Repository, error){
		return target2.NewRepository(ctx, dbase, dbase, c.kms)
	}
```

irenarindos: I'm also running into failures on `TestHandleImplementedPaths` for PATCH and DELETE targets; trying to figure that out, but posting in case you might know why

irenarindos:
Looks like `TestAuthorizeSession_Errors/host_defined_in_request_and_alias` is failing now; looks like the expected error message should be updated to `The host id specified in the request does not match the one provided by the alias. Consider omiting the host id in the request`

irenarindos (approved):

PR #4289, opened 2024-01-25
Add new GrantsForUser test

jefferai:
This exercises a wide array of granting possibilities and ensures that we see all of the desired grants and none that we don't expect.


internal/iam/repository_role_grant_test.go
```diff
@@ -568,3 +571,396 @@ func TestRepository_SetRoleGrants_Parameters(t *testing.T) {
...
+	// The first org/project do not have any direct grants to the user. They
+	// contain roles but the user is not a principal.
+	noGrantOrg1, noGrantProj1 := TestScopes(
+		t,
+		repo,
+		WithSkipAdminRoleCreation(true),
+		WithSkipDefaultRoleCreation(true),
+	)
+	noGrantOrg1Role := TestRole(t, conn, noGrantOrg1.PublicId)
+	TestRoleGrant(t, conn, noGrantOrg1Role.PublicId, "id=o_noGrantOrg1;actions=*")
```
moduli, line 597: Do these need to use `ids=` now?

jefferai: Yep. Updated.

louisruch (approved):
LGTM

PR #4305, opened 2024-01-26
return error when creating table fails

talanknight:


internal/clientcache/internal/db/db.go
```diff
@@ -45,7 +45,7 @@ func Open(ctx context.Context, opt ...Option) (*db.DB, error) {
 	switch {
 	case opts.withDbType == dbw.Sqlite:
 		if err := createTables(ctx, conn); err != nil {
-			errors.Wrap(ctx, err, op)
+			return nil, errors.Wrap(ctx, err, op)
```
johanbrandhorst, line 48: 😰

johanbrandhorst (approved):

PR #4311, opened 2024-01-26
testing/e2e: add auth token pagination tests

johanbrandhorst:


testing/internal/e2e/tests/base/paginate_auth_token_test.go
```diff
@@ -0,0 +1,242 @@
...
+	accId, password := boundary.CreateNewAccountCli(t, ctx, amId, "testuser")
+	boundary.SetAccountToUserCli(t, ctx, userId, accId)
+	t.Cleanup(func() {
+		ctx := context.Background()
+		boundary.AuthenticateAdminCli(t, ctx)
+		output := e2e.RunCommand(ctx, "boundary", e2e.WithArgs("auth-methods", "delete", "-id", amId))
+		require.NoError(t, output.Err, string(output.Stderr))
+		output = e2e.RunCommand(ctx, "boundary", e2e.WithArgs("scopes", "delete", "-id", newOrgId))
+		require.NoError(t, output.Err, string(output.Stderr))
+	})
```
moduli, line 47: Should we have the org cleanup occur just after org creation just in case something happens to fail beforehand for whatever reason?

testing/internal/e2e/boundary/authmethod.go
```diff
@@ -0,0 +1,25 @@
...
+
+// CreateNewAuthMethodApi creates a new auth method using the Go api.
+// Returns the id of the new auth method
+func CreateNewAuthMethodApi(t testing.TB, ctx context.Context, client *api.Client, scopeId string) string {
+	aClient := authmethods.NewClient(client)
+	newAMResult, err := aClient.Create(ctx, "password", scopeId)
+	require.NoError(t, err)
+
+	authMethodId := newAMResult.Item.Id
+	t.Logf("Create Auth Method: %s", authMethodId)
```
moduli, line 23: nit
```suggestion
	t.Logf("Created Auth Method: %s", authMethodId)
```

testing/internal/e2e/tests/base/paginate_auth_token_test.go
```diff
@@ -0,0 +1,242 @@
...
+
+	boundary.AuthenticateCli(t, ctx, amId, "testuser", password)
+
+	// Create enough auth tokens to overflow a single page.
+	// We created an auth token when we authenticated, so we
+	// need to remove it from the total auth token count.
+	numPrecreatedAuthTokens := 1
```
moduli, line 54: If a number of tests have already executed prior to this test running, wouldn't there be a number of auth tokens (not just 1) already created? The test is passing in CI, but it might have more auth tokens in the list than expected?

EDIT: But that wouldn't make sense since that would've failed this check below...
```
require.Len(t, newAuthTokens.Items, c.MaxPageSize+1)
```

EDIT2: Ah, we're just listing auth tokens in that org, which only have tokens from the login attempts using this new auth method.

testing/internal/e2e/tests/base/paginate_auth_token_test.go
```diff
@@ -0,0 +1,242 @@
...
+	accId, password := boundary.CreateNewAccountCli(t, ctx, amId, "testuser")
+	boundary.SetAccountToUserCli(t, ctx, userId, accId)
+	t.Cleanup(func() {
+		ctx := context.Background()
+		client.SetToken(adminToken)
+		_, err := amClient.Delete(ctx, amId)
+		require.NoError(t, err)
+		_, err = sClient.Delete(ctx, newOrgId)
+		require.NoError(t, err)
+	})
```
moduli, line 150: Should we have the org cleanup occur just after org creation just in case something happens to fail beforehand for whatever reason?

testing/internal/e2e/tests/base/paginate_auth_token_test.go
```diff
@@ -0,0 +1,242 @@
...
+	for i := 0; i < c.MaxPageSize+1-numPrecreatedAuthTokens; i++ {
+		authtokenIds = append(authtokenIds, createAuthToken(t, ctx, client, amId, "testuser", password))
+	}
+
+	// List auth tokens
+	initialAuthTokens, err := atClient.List(ctx, newOrgId, authtokens.WithRecursive(true))
+	require.NoError(t, err)
+
+	var returnedIds []string
+	// Ignore the two precreated auth tokens, which will appear at the end
```
moduli, line 179: Why `two` here?

testing/internal/e2e/tests/base/paginate_auth_token_test.go
johanbrandhorst: That's fair, yeah :)

testing/internal/e2e/tests/base/paginate_auth_token_test.go
johanbrandhorst: Copy paste error, good find

moduli (approved):

testing/internal/e2e/tests/base/paginate_auth_token_test.go
```diff
@@ -0,0 +1,250 @@
...
+		require.NoError(t, output.Err, string(output.Stderr))
+	})
+	userId := boundary.CreateNewUserApi(t, ctx, client, newOrgId)
+	amId := boundary.CreateNewAuthMethodApi(t, ctx, client, newOrgId)
+	t.Cleanup(func() {
+		ctx := context.Background()
+		boundary.AuthenticateAdminCli(t, ctx)
+		output := e2e.RunCommand(ctx, "boundary", e2e.WithArgs("auth-methods", "delete", "-id", amId))
+		require.NoError(t, output.Err, string(output.Stderr))
+	})
```
moduli, line 49: one more comment... does the auth method not get deleted when the org is deleted?

moduli: Oh right. I think I remember that one.

PR #4320, opened 2024-01-29
Finish sqltests for multi grant scopes

jefferai:


internal/db/sqltest/tests/iam/grant_scope_validation.sql
```diff
@@ -27,113 +30,239 @@ insert into iam_role (public_id, scope_id) values
...
+-- Case 5: "descendants" and "children" are mutually exclusive
+prepare insert_descendants__and_children_fail_global as
+    insert into iam_role_grant_scope
+        (role_id,        scope_id_or_special)
+    values
+        ('r_globeglobe', 'descendants'),
+        ('r_globeglobe', 'children');
+select throws_like(
+    'insert_descendants__and_children_fail_global',
+    'invalid to specify both "children" and "descendants"as a grant scope'
```
tmessi, line 264: It looks like there is a [space missing in the error message](https://github.com/hashicorp/boundary/blob/077f84eb3985d2da1652272823c1ee9b1894391e/internal/db/schema/migrations/oss/postgres/83/01_iam_role_grant_scope.up.sql#L124-L132). Should probably update that and then this assertion can be updated.

```suggestion
    'invalid to specify both "children" and "descendants" as a grant scope'
```

internal/db/sqltest/tests/iam/grant_scope_validation.sql
```diff
@@ -27,113 +30,239 @@ insert into iam_role (public_id, scope_id) values
...
+-- Reset grant scopes
+delete from iam_role_grant_scope;
+
+-- Case 5: "descendants" and "children" are mutually exclusive
+prepare insert_descendants__and_children_fail_global as
+    insert into iam_role_grant_scope
+        (role_id,        scope_id_or_special)
+    values
+        ('r_globeglobe', 'descendants'),
+        ('r_globeglobe', 'children');
```
tmessi, line 261: It looks like there is another variant of this case when `children` was inserted first followed by `descendants`. We should probably cover both variants.

tmessi (changes requested):

internal/db/sqltest/tests/iam/grant_scope_validation.sql
jefferai: Fixed

internal/db/sqltest/tests/iam/grant_scope_validation.sql
jefferai: Fixed

tmessi (approved):

mgaffney (approved):

PR #4325, opened 2024-01-30
Merge PKI & KMS docs

AdamBouhmad:
In Boundary 0.13, KMS workers were deprecated to simplify worker setup for administrators. With a common worker, there are now three methods of registration -- Controller-Led, Worker-Led, and via an external KMS. In 0.15, KMS workers are being removed. This PR's goal is to ensure our docs are clear for administrators looking to understand & setup self-managed workers in their environment


website/content/docs/configuration/worker/worker-configuration.mdx
```diff
@@ -0,0 +1,217 @@
...
+  ]
+
+  public_addr = "myhost.mycompany.com"
+
+  tags {
+    type   = ["prod", "webservers"]
+    region = ["us-east-1"]
+  }
+
+  # use_deprecated_kms_auth_method = true
```
irenarindos, line 163: Should this line be removed?

AdamBouhmad: yes -- good catch

website/content/docs/configuration/worker/worker-configuration.mdx
```diff
@@ -0,0 +1,213 @@
...
+are used to connect to upstream Boundary clusters.
+
+## Resources
+
+For more on how `tags{}` in the above configuration are used to facilitate
+routing to the correct target, refer to the [Worker
+Tags](/boundary/docs/concepts/filtering/worker-tags) page.
+
+
+## KMS configuration
```
irenarindos, line 181: I think this section should be moved/ perhaps become a subsection for the worker led auth, since that's what it applies to.

Dan-Heath:
@AdamBouhmad , thank you for your help with this! I think what you have looks good. I found some additional pages with PKI/KMS references and I am not sure if they need to be updated:

- [Worker stanza](https://developer.hashicorp.com/boundary/docs/configuration/worker)
- [Configure workers](https://developer.hashicorp.com/boundary/docs/install-boundary/configure-workers)
- [Data encryption](https://developer.hashicorp.com/boundary/docs/concepts/security/data-encryption)

AdamBouhmad:
> @AdamBouhmad , thank you for your help with this! I think what you have looks good. I found some additional pages with PKI/KMS references and I am not sure if they need to be updated:
> 
> * [Worker stanza](https://developer.hashicorp.com/boundary/docs/configuration/worker)
> * [Configure workers](https://developer.hashicorp.com/boundary/docs/install-boundary/configure-workers)
> * [Data encryption](https://developer.hashicorp.com/boundary/docs/concepts/security/data-encryption)

@Dan-Heath thanks for finding these, just addressed them in my recent commits -- feel free to take a look when you have a sec

Dan-Heath (approved):
LGTM, thank you Adam!

website/content/docs/configuration/worker/worker-configuration.mdx
AdamBouhmad: this is a good idea -- I wont include this for the initial update as we do call out that it's only available for worker-led & controller-led, but we should figure out if we want to refactor the docs to more easily include this as a subsection (cc @Dan-Heath)

PR #4327, opened 2024-01-30
chore(e2e): Name output based on scenario

moduli:
In GitHub Action `upload-artifact v4.0`, there was a breaking change for uploading multiple files to the same name (https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes). Previously, when you did that, it was bundled into a single zip file. Now, that is no longer allowed and files now need to have unique artifact names.

We will need this change when the action is updated to the new version.

Before:
Screenshot 2024-01-30 at 3 57 47 PM

After:
Screenshot 2024-01-30 at 3 29 16 PM


stasryzhov (approved):
That's a great change, @moduli 👍

PR #4332, opened 2024-01-31
Fix calculating uptime after system sleep

talanknight:


internal/clientcache/internal/daemon/status_handler.go
```diff
@@ -123,7 +123,10 @@ func toApiStatus(in *cache.Status, started time.Time, socketAddr, logLocation st
 	}
 
 	out := &StatusResult{
-		Uptime:        time.Since(started),
+		// Round(0) strips the monotoic clock from results which can drift
```
johanbrandhorst, line 126:
```suggestion
		// Round(0) strips the monotonic clock from results which can drift
```

johanbrandhorst (approved):

PR #4334, opened 2024-01-31
Client-control-plane-support-policy.mdx

PPacent:
Update supported-versions for client <-> control plane support policy

### DIFF
--- a/website/content/docs/enterprise/supported-versions.mdx
+++ b/website/content/docs/enterprise/supported-versions.mdx
@@ -22,7 +22,7 @@ Release updates for Boundary Enterprise can be found at https://releases.hashico
 
 ## Control plane and worker compatability
 
-Although major releases are supported as outlined within the support periods section above, within a Boundary Enterprise deployment API backwards compatibility is only supported between the control plane and workers from the prior “major release”. Using workers on newer versions than the control plane they are registered with is not supported.
+Although major releases are supported as outlined within the support periods section above, within a Boundary Enterprise deployment API backwards compatibility is only supported between the control plane and workers from the prior “major release”. Using a worker with version that is newer than the control plane they connect to is not supported.
 All workers within an environment must be on the same version.
 
 For example, Boundary workers version 0.13.0 are compatible with Boundary control plane running Boundary 0.14.0.
--- a/website/content/docs/enterprise/supported-versions.mdx
+++ b/website/content/docs/enterprise/supported-versions.mdx
@@ -20,13 +20,22 @@ A major release is identified by a change in the first (X) or second (Y) digit i
 As a best practice, HashiCorp expects customers to stay current within two (2) releases from the latest major release in order to receive optimal support.
 Release updates for Boundary Enterprise can be found at https://releases.hashicorp.com.
 
-## Control plane and worker availability
+## Control plane and worker compatability
 
-Although major releases are supported as outlined within the support periods section above, within a Boundary Enterprise deployment API backwards compatibility is only supported between the control plane and workers from the prior “major release”.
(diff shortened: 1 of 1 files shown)


psekar (approved):
Minor: compatability -> compatibility

website/content/docs/enterprise/supported-versions.mdx
```diff
@@ -20,15 +20,24 @@ A major release is identified by a change in the first (X) or second (Y) digit i
...
 All workers within an environment must be on the same version.
 
 For example, Boundary workers version 0.13.0 are compatible with Boundary control plane running Boundary 0.14.0.
 However, they will not have compatibility once the control plane is updated to version 0.15.0 or above.
 Customers are recommended to run the latest versions of Boundary in order to leverage the newest features and bug fixes.
 
+## Control plane and client/cli compatability
+
+The supported version compatability between Boundary's control plane and Boundary clients/cli is the same as control plane and worker compatability. Within a Boundary Enterprise deployment API backwards compatibility is only supported between the control plane and clients from the prior “major release”. Using clients on newer versions than the control plane they are registered with is not supported.
```
stellarsquall, line 34: Is it right to call this a "major release"? I know we don't necessarily adhere to semver, but this could be confused with 1.0.0 vs 2.0.0, for example.

website/content/docs/enterprise/supported-versions.mdx
```diff
@@ -20,15 +20,24 @@ A major release is identified by a change in the first (X) or second (Y) digit i
 As a best practice, HashiCorp expects customers to stay current within two (2) releases from the latest major release in order to receive optimal support.
 Release updates for Boundary Enterprise can be found at https://releases.hashicorp.com.
 
-## Control plane and worker availability
+## Control plane and worker compatability
 
-Although major releases are supported as outlined within the support periods section above, within a Boundary Enterprise deployment API backwards compatibility is only supported between the control plane and workers from the prior “major release”.
+Although major releases are supported as outlined within the support periods section above, within a Boundary Enterprise deployment API backwards compatibility is only supported between the control plane and workers from the prior “major release”. Using workers on newer versions than the control plane they are registered with is not supported.
```
jimlambrt, line 25: I'm not sure I understand what "registered with" means.  Perhaps: "Using a worker with version that is newer than the control plane they connect to is not supported."

website/content/docs/enterprise/supported-versions.mdx
stellarsquall, line 34:
```suggestion
The supported version compatability between Boundary's control plane and Boundary clients/cli is the same as control plane and worker compatability. Within a Boundary Enterprise deployment API backwards compatibility is only supported between the control plane and clients from the prior [major release](https://support.hashicorp.com/hc/en-us/articles/360021185113-Support-Period-and-End-of-Life-EOL-Policy). Using clients on newer versions than the control plane they are registered with is not supported.
```

I think adding this link to our version policy would eliminate confusion here

jimlambrt: I'm with Robin on this one. I think, this would be minor release... basically we support patch version compatibility but not minor version compatibility.

PPacent: @stellarsquall I'm not sure this is required, we already make this explicit elsewhere in the page: https://developer.hashicorp.com/boundary/docs/enterprise/supported-versions see "Support Periods"

PPacent: @stellarsquall we make this explicit elsewhere in the page: https://developer.hashicorp.com/boundary/docs/enterprise/supported-versions see "Support Periods"

stellarsquall: that works, thanks!

website/content/docs/enterprise/supported-versions.mdx
stellarsquall: I like this as well, I think it's more accurate for multi-hop scenarios

stellarsquall (approved):

PR #4345, opened 2024-02-01
boundary daemon add-token to return returns 403 Forbidden when Boundary returns the same

talanknight:
Previously the response from the client cache daemon was set as an internal error. Now it returns a 403 with a snippet that says "Failed to get auth token from Boundary".

This will only happen when the user has had their read:self grants removed for auth tokens.


lisbet-alvarez (approved):
Thank you for this work! Tested locally and I am now seeing the 403 code!

jimlambrt (approved):
Ty

PR #4349, opened 2024-02-01
deps: update runc dependency

johanbrandhorst:
Fixes [GHSA-xr7r-f8xq-vfvv](https://github.com/opencontainers/runc/security/advisories/GHSA-xr7r-f8xq-vfvv). Currently causing our CRT builds to fail. I don't think we're vulnerable in practice so we don't need to backport this.


go.mod
```diff
@@ -193,24 +194,23 @@ require (
 	github.com/oklog/run v1.1.0 // indirect
 	github.com/opencontainers/go-digest v1.0.0 // indirect
 	github.com/opencontainers/image-spec v1.0.2 // indirect
-	github.com/opencontainers/runc v1.1.5 // indirect
+	github.com/opencontainers/runc v1.1.12 // indirect
```
johanbrandhorst, line 197: v1.1.12 is the fixed version.

tmessi:
Looks like you need to run `make gen` since this updated protobuf and a bunch of comments in protobuf files now need updating.

jimlambrt:
should this be back ported to 15.x so our CRT builds will succeed?

johanbrandhorst:
> should this be back ported to 15.x so our CRT builds will succeed?

Yep, fair point

tmessi (approved):

PR #4350, opened 2024-02-01
chore: Fix breaking change in labeler when upgrading to v5

moduli:
The github [labeler](https://github.com/actions/labeler) broke when upgrading to v5 from the [automated](https://github.com/hashicorp/boundary-ui/pull/2065) workflow pin upgrade.

The check for labeler in this PR will still fail as it's using the workflow and config file from `main`

(similar change to https://github.com/hashicorp/boundary-ui/pull/2078)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,126 +1,168 @@
 api:
-  - api/**/*
+  - changed-files:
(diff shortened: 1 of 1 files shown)


johanbrandhorst (approved):
Thanks!

PR #4355, opened 2024-02-02
Use the intercepted token always. If the keyring matches, use it from there

talanknight:
This PR fixes a number of issues with the way we currently add tokens to the client cache daemon.

Previously, If a keyring is not present on the platform the user must specify `-keyring-type none` when calling `boundary daemon add-token`, With this PR it will treat the platform not having a keyring it knows about as if the user provided the keyring-type none.

Previously, if a user called `boundary authenticate -format json` the returned token would not be added to the client cache. With this PR it is intercepted and added.

### DIFF
--- a/internal/clientcache/cmd/daemon/addtoken.go
+++ b/internal/clientcache/cmd/daemon/addtoken.go
@@ -168,12 +168,12 @@ func (c *AddTokenCommand) Add(ctx context.Context, ui cli.Ui, apiClient *api.Cli
 		// doesn't play a part in the request.
 	default:
 		// Just because the keyring type is set doesn't mean the token to add
-		// is contained in it. For example, If the token was intercepted
-		// from an authentication request with '-format json' the token is
+		// is contained in it. For example, if the token was intercepted
+		// from an authentication request with '-format json' then token is
 		// not stored in the keyring, even if a keyring is provided.
 
 		// Try to read the token from the keyring in a best effort way. Ignore
-		// any errors since the keyring may night be present on the system.
+		// any errors since the keyring may not be present on the system.
 		at := base.ReadTokenFromKeyring(ui, keyringType, tokenName)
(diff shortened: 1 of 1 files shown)


moduli (commented):
The test is still failing here
```
_, err = te.Boundary.Resource.Exec(
[]string{
	"boundary", "authenticate", "password",
	"-addr", te.Boundary.UriNetwork,
	"-auth-method-id", te.DbInitInfo.AuthMethod.AuthMethodId,
	"-login-name", te.DbInitInfo.AuthMethod.LoginName,
	"-password", "env://E2E_TEST_BOUNDARY_PASSWORD",
	"-keyring-type", "none",
	"-format", "json",
},
dockertest.ExecOptions{
	StdOut: buf,
	StdErr: ebuf,
	Env:    []string{"E2E_TEST_BOUNDARY_PASSWORD=" + te.DbInitInfo.AuthMethod.Password},
},
)
require.NoError(t, err)
```
for the following reason

```
migration_test.go:355: 
│         	Error Trace:	/home/runner/work/boundary/boundary/testing/internal/e2e/tests/database/migration_test.go:355
│         	            				/home/runner/work/boundary/boundary/testing/internal/e2e/tests/database/migration_test.go:54
│         	Error:      	Should be empty, but was Error opening keyring: Specified keyring backend not available
│         	            	Token must be provided via BOUNDARY_TOKEN env var or -token flag. Reading the token can also be disabled via -keyring-type=none.
```

moduli:
Oh, right. Because of the way this test is structured, it starts with the most recently released version of Boundary (i.e. 0.15.0) and populates its database using that version before trying to migrate it to the current working branch. The failure is happening during that population stage, so it wouldn't get these changes until we release the next version...

jimlambrt (changes requested):
I few minor grammar/typo bits.  I did ask one question about what to do when the client token and the keyring token are not equal that I think might be important.

jimlambrt (approved):
ty!

PR #4362, opened 2024-02-05
Revert "chore(e2e): Remove tparse"

moduli:
Reverts hashicorp/boundary#4326

I didn't realize that when running `go test`, it needed to download other go modules in order to execute. I assumed that the test suite only needed the go mod cache specifically for `tparse`.


johanbrandhorst (approved):
Oops 😁

Issue #4363, opened 2024-02-06
illegal char in event filtering


yongzhang:
**Describe the bug**
Cannot start boundary because of illegal char in event filtering.
Boundary version： 0.15.0

**To Reproduce**
1. Add below event sink:
```
      events {
        audit_enabled        = true
        observations_enabled = true
        sysevents_enabled    = true
        telemetry_enabled    = false

        sink "stderr" {
          name         = "all-events"
          description  = "All events sent to stderr"
          event_types  = ["*"]
          format       = "cloudevents-json"
          deny_filters = [
            '"/data/request_info/path" == "/health"'
          ]
        }
      }
```

2. Start boundary service see error:
```
Feb 06 04:47:28 ip-10-130-19-94 systemd[1]: Started "HashiCorp Boundary - Identity-based access management for dynamic infrastructure".
Feb 06 04:47:28 ip-10-130-19-94 boundary[874]: Error parsing config: Error parsing KMS HCL: At 15:7: illegal char
Feb 06 04:47:28 ip-10-130-19-94 systemd[1]: boundary.service: Main process exited, code=exited, status=3/NOTIMPLEMENTED
Feb 06 04:47:28 ip-10-130-19-94 systemd[1]: boundary.service: Failed with result 'exit-code'.
Feb 06 04:47:33 ip-10-130-19-94 systemd[1]: boundary.service: Scheduled restart job, restart counter is at 1.
Feb 06 04:47:33 ip-10-130-19-94 systemd[1]: Stopped "HashiCorp Boundary - Identity-based access management for dynamic infrastructure".
```

![image](https://github.com/hashicorp/boundary/assets/15604715/879a4def-d1b2-456f-b8a2-e6b0946bbc89)

I don't see any wrong chars here by checking the docs, please help, thanks. I also tried `allow_filters` config from [here](https://developer.hashicorp.com/boundary/docs/concepts/filtering/events), it reports the same error.

**Expected behavior**

**Additional context**

yongzhang:
Ok, found the cause, submitted PR to fix docs:
https://github.com/hashicorp/boundary/pull/4365

PR #4365, opened 2024-02-06
Fix docs for event filters

yongzhang:
Fixes https://github.com/hashicorp/boundary/issues/4363.


stellarsquall (approved):
thanks for opening a PR to fix this!

PR #4367, opened 2024-02-06
docs: Clarify need for permissions to search

Dan-Heath:
This PR attempts to address some confusion around why search results are not displaying in scenarios where a user does not have permission to view them.

[Client cache](https://boundary-bg4tn2t78-hashicorp.vercel.app/boundary/docs/api-clients/client-cache)
[`daemon add-token`](https://boundary-bg4tn2t78-hashicorp.vercel.app/boundary/docs/commands/daemon/add-token)

### DIFF
--- a/website/content/docs/commands/daemon/add-token.mdx
+++ b/website/content/docs/commands/daemon/add-token.mdx
@@ -17,6 +17,15 @@ If you authenticate to multiple Boundary instances, the client cache stores mult
 
 By adding auth tokens to your client cache, you can select which specific Boundary instance you want to search.
 
+<Note>
+
+The `search` operation only displays the resources that you have read permissions to view.
+If you are logged in as a user who does not have the permissions to view a resource, it does not display in the list of results.
+
+For more information, refer to [Identity and access management (IAM)](/boundary/docs/concepts/iam).
+
+</Note>
+
 ## Examples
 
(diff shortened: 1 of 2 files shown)


stellarsquall (approved):

website/content/docs/commands/daemon/add-token.mdx
```diff
@@ -17,6 +17,15 @@ If you authenticate to multiple Boundary instances, the client cache stores mult
 
 By adding auth tokens to your client cache, you can select which specific Boundary instance you want to search.
 
+<Note>
+
+The `search` operation only displays the resources that you have read permissions to view.
```
ZedLi, line 22: For add-token in the client daemo, I'm not sure if we want to call out that you specifically need the `read` or `read:self` grant on `auth-token` to successfully add a token since it's possible to authenticate without actually being able to read the token in boundary which has tripped me up before in testing

Dan-Heath: Thank you @ZedLi ! That does seem important. I modified the note on that page to read:

> The `search` operation only displays the resources that you have permissions to view. You must have the `read` or `read:self` grant on the auth token to successfully add it. If you are logged in as a user who does not have the permissions to view a resource, it does not display in the list of results.

Does it sound OK?

ZedLi (approved):

PR #4372, opened 2024-02-06
Pass our configured random reader into nodee functions

jefferai:
Note for reviewers: in some places where there are already options I'm using options, and in other places I added something to the signature. If this feels odd I can modify it to be consistent one way or the other, happy to take feedback on that.


internal/server/job/options.go
```diff
@@ -49,3 +53,10 @@ func WithCertificateLifetime(with time.Duration) Option {
 		o.withCertificateLifetime = with
 	}
 }
+
+// WithRandomReader provides a way to specify a random reader
```
irenarindos, line 57: Should a opts test be added for this?

internal/server/repository.go
```diff
@@ -26,10 +27,15 @@ type Repository struct {
...
 }
 
-// NewRepository creates a new server Repository. Supports the options: WithLimit
-// which sets a default limit on results returned by repo operations.
+// NewRepository creates a new server Repository.
+//
+// Supported options:
+//
+// * WithLimit: default limit on results returned by repo operations
+// * RandomReader: Specify a specific random source
```
irenarindos, line 38:
```suggestion
// * WithRandomReader: Specify a specific random source
```

PR #4373, opened 2024-02-06
handlers: always return pagination parameters in list

johanbrandhorst:
Ensure that pagination parameters are always returned from successful list calls. Turns out the only affected list endpoint was targets (🙄 of course), but I've also added tests to all the other endpoints. The removed lines are never hit as an error is returned from the called function if there are no scopes available.


jefferai:
LGTM but I don't know the pagination code well. I think the reason only targets was affected is because it's one of the only handlers that does some of the prefetching on scopes to reduce database load so its listing works differently than most others.

PR #4374, opened 2024-02-07
Allow attributes to get updated after resolving targets by alias

talanknight:
This is a draft PR while we discuss if this is the right solution to the problem.

Added is a test case `Update attribute by alias` which passes the request with a non *pb.Target_TcpTargetAttributes attributes field to match how the request actually would arrive to the handler since the attributes translation interceptor wouldn't be able to map the request to the tcp type using the id when the request uses an alias.

### DIFF
--- a/internal/daemon/controller/handlers/targets/target_service.go
+++ b/internal/daemon/controller/handlers/targets/target_service.go
@@ -444,13 +444,13 @@ func (s Service) UpdateTarget(ctx context.Context, req *pbs.UpdateTargetRequest)
 			return nil, errors.Wrap(ctx, err, op)
 		} else {
 			req.Id = a.DestinationId
+			// The first time we tried to Transform the request to a specific
+			// attribute subtypes we didn't have the id prefix. Now that we do
+			// try again.
+			if err := subtypes.TransformRequest(ctx, req); err != nil {
+				return nil, errors.Wrap(ctx, err, op)
+			}
 		}
-	}
-
-	// since we resolved the alias to an id, transform the request to get the
-	// correct subtype attributes
-	if err := subtypes.TransformRequest(ctx, req); err != nil {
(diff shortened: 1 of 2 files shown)


internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -447,6 +447,12 @@ func (s Service) UpdateTarget(ctx context.Context, req *pbs.UpdateTargetRequest)
 		}
 	}
 
+	// since we resolved the alias to an id, transform the request to get the
+	// correct subtype attributes
+	if err := subtypes.TransformRequest(ctx, req); err != nil {
+		return nil, errors.Wrap(ctx, err, op)
+	}
```
johanbrandhorst, line 454: This is unconditional right now, would it not normally go inside the `else` above?

internal/daemon/controller/handlers/targets/tcp/tcp.go
```diff
@@ -92,6 +92,11 @@ func newAttribute(m any) targets.Attributes {
 	}
 	if tcpAttr, ok := m.(*pb.Target_TcpTargetAttributes); ok {
 		a.TcpTargetAttributes = tcpAttr.TcpTargetAttributes
+	} else if attr, ok := m.(*pb.Target_Attributes); ok {
+		_ = attr
+		// We could update this to translate the generic attributes to the TCP
+		// attributes, at least for the getAttributes case, if we wanted to
+		// isolate the change to the targets package.
```
johanbrandhorst, line 99: I see the problem, we either have to duplicate the transformRequest logic here or export it and call it in the handler? I kinda wish we could do the alias resolution here since that would still allow the attribute transformer to do its job.

irenarindos (commented):
Just a comment that we'll also need this for SSH targets/ a similar test case

PR #4377, opened 2024-02-07
Cameronperera/app token delete

cameronperera:


internal/apptoken/repository_delete_test.go
```diff
@@ -0,0 +1,60 @@
...
+			reader:          testRw,
+			writer:          testRw,
+			kms:             testKms,
+			publicId:        testOrg.PublicId,
+			createdBy:       testUserHistoryId,
+			grantsStr:       "id=*;type=*;actions=*",
+			wantRowsDeleted: 1,
+		},
+	}
+	for _, tc := range tests {
```
jimlambrt, line 55: Looks like you're missing some of the test code here.

jimlambrt (changes requested):

internal/apptoken/repository_delete_test.go
cameronperera: Ope. Thanks for catching that. I staged my changes but didn't actually commit them.

jimlambrt (approved):
Ty!

uruemu (approved):
LGTM

PR #4378, opened 2024-02-07
test(e2e): Add test for authentication when no keyring is available

moduli:
There was a bug discovered where the client cache daemon would try to access the keyring to grab an auth token even though there was no keyring on the system (https://github.com/hashicorp/boundary/pull/4355). This PR adds a test to cover this case.

To achieve this, the test executes an `authenticate -format json` command on the docker container running the boundary controller, which has no keyring. Prior to the fix, we would get an error
```
❯ go test -v -count=1  github.com/hashicorp/boundary/testing/internal/e2e/tests/base -run TestCliAuthenticateNoKeyring
=== RUN   TestCliAuthenticateNoKeyring
    # STDOUT
    authenticate_primary_test.go:132: {"status_code":200,"item":{"attributes":{"account_id":"acctpw_wKPEJfkpMr", ...

    # STDERR
    authenticate_primary_test.go:133: Error opening keyring: Specified keyring backend not available
        Token must be provided via BOUNDARY_TOKEN env var or -token flag. Reading the token can also be disabled via -keyring-type=none.
```

This is what the response should look like
```
❯ go test -v -count=1  github.com/hashicorp/boundary/testing/internal/e2e/tests/base -run TestCliAuthenticateNoKeyring
=== RUN   TestCliAuthenticateNoKeyring
    # STDOUT
    authenticate_primary_test.go:133: {"status_code":200,"item":{"attributes":{"account_id":"acctpw_wKPEJfkpMr", ...
    # STDERR
    authenticate_primary_test.go:134: 
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -119,6 +119,7 @@ require (
 )
 
 require (
+	github.com/docker/distribution v2.8.2+incompatible // indirect
 	github.com/frankban/quicktest v1.14.5 // indirect
 	github.com/glebarez/go-sqlite v1.21.2 // indirect
 	github.com/kardianos/osext v0.0.0-20190222173326-2bc1f35cddc0 // indirect
@@ -150,7 +151,7 @@ require (
 	github.com/danieljoos/wincred v1.2.0 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/docker/cli v23.0.1+incompatible // indirect
-	github.com/docker/docker v24.0.7+incompatible // indirect
+	github.com/docker/docker v24.0.7+incompatible
 	github.com/docker/go-connections v0.4.0 // indirect
 	github.com/docker/go-units v0.5.0 // indirect
 	github.com/dvsekhvalnov/jose2go v1.5.1-0.20231206184617-48ba0b76bc88 // indirect
--- a/testing/internal/e2e/tests/base_plus/authenticate_no_keyring_test.go
+++ b/testing/internal/e2e/tests/base_plus/authenticate_no_keyring_test.go
@@ -1,0 +1,132 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package base_plus_test
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"testing"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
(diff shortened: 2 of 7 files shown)


go.mod
moduli, line 1: Are these changes ok?

testing/internal/e2e/tests/base_plus/authenticate_no_keyring_test.go
```diff
@@ -0,0 +1,84 @@
...
+func TestCliAuthenticateNoKeyring(t *testing.T) {
+	e2e.MaybeSkipTest(t)
+	ctx := context.Background()
+
+	bc, err := boundary.LoadConfig()
+	require.NoError(t, err)
+
+	containerID := "boundary"
+
+	docker, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
```
johanbrandhorst, line 35: It looks like we're using https://github.com/ory/dockertest in other places, maybe we can use it here too? Not sure if it's suitable, feel free to say no.

go.mod
johanbrandhorst: See comment about dockertest

testing/internal/e2e/tests/base_plus/authenticate_no_keyring_test.go
moduli: I'll double-check. When I first looked, it didn't seem like there was a way to connect to an existing docker image using `dockertest`.

johanbrandhorst: Ah you're probably right!

johanbrandhorst (approved):

testing/internal/e2e/tests/base_plus/authenticate_no_keyring_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/hashicorp/boundary/testing/internal/e2e/boundary"
+	"github.com/stretchr/testify/require"
+)
+
+// TestCliAuthenticateNoKeyring tests authentication when we're using `-format
+// json` and there is no keyring on the system.
+// This covers a case where the client cache daemon should intercept the auth
+// token in the json response. There was previously a bug where the client cache
+// daemon tried to access the keyring when there wasn't one on the system.
+func TestCliAuthenticateNoKeyring(t *testing.T) {
```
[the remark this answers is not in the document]
moduli: Hmmm. I can try to add a check in the test to confirm that there's no keyring.

PR #4379, opened 2024-02-08
Resolve aliases prior to subtype attribute transforming interceptor

talanknight:
Almost all the target_service_test.go changes are reverting the tests that were added previously to allow the target_service handler to perform the alias resolution on its own.

One negative from this approach is that we had to special case the AuthorizeSessionRequest since that request allows target names to be passed in through the id field.

### DIFF
--- a/internal/proto/controller/api/services/v1/target_service.proto
+++ b/internal/proto/controller/api/services/v1/target_service.proto
@@ -375,6 +375,8 @@ message AuthorizeSessionRequest {
   // The ID of the target. Required unless some combination of scope_id/scope_name and name are set.
   string id = 1 [(custom_options.v1.aliasable) = {
     unless_set: {
+      // scope id and scope name being set indicates that the value in this field
+      // is actually a target name and not an alias.
       fields: [
         'scope_id',
         'scope_name'


internal/alias/interceptor.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+	}
+}
+
+// ResolutionInterceptor returns a grpc.UnaryServerInterceptor that resolves
+// alias values in the request to their corresponding destination ids. If no
+// alias is found or the alias has no destination id, an error is returned.
+// For an field in the request to be considered for alias resolution, it must
+// be annotated with the Aliasable proto option.
+func ResolutionInterceptor(
```
tmessi, line 53: I don't think any of this interceptor implementation should be in the domain package. This causes the domain layer to import things like `internal/daemon/controller/handlers` and `internal/gen/controller/api/services` which reverses the dependency order.

It might be ok to leave the bulk of the interceptor logic here as a domain service, if we can eliminate these imports. It looks like `services` is only imported to handle the special case for AuthorizeSession, see my other comment about that which might eliminate the need for that import. And it looks like `handlers` is only needed for constructing errors. That could be eliminated by a thin wrapper of the domain service to create a grpc interceptor in another layer, which applies formatting of the errors using the `handlers` package.

internal/alias/interceptor.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+
+		// Special case the authorize session request, as it can be ambiguous
+		// if the value is an alias or a target name with a scope id/name
+		if f.FullName() == (&pbs.AuthorizeSessionRequest{}).ProtoReflect().Descriptor().Fields().ByName("id").FullName() {
+			// If the scope is also provided, then we know the value is a target name
+			if r.Has(fields.ByName("scope_id")) || r.Has(fields.ByName("scope_name")) {
+				continue
+			}
+		}
```
tmessi, line 112: I am wondering if there is a way to express this kind of behavior within the proto options as well.

internal/db/schema/migrations/oss/postgres/84/05_alias_view.up.sql
```diff
@@ -0,0 +1,15 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+
+  -- alias_view defines a view that retrieves the common columns from the
+  -- subtype alias tables.
+  create view alias_view as
```
tmessi, line 8: I think we tend to call this something like `_all_subtypes`, ie `alias_all_subtypes`?

tmessi (changes requested):
I think the overall approach here makes sense to me. I just have a few comments mostly about some code organization.

internal/alias/interceptor.go
```diff
@@ -0,0 +1,131 @@
...
+		registerAliasableFields(m.Descriptor())
+		return true
+	})
+}
+
+type registrationInfo struct {
+	unaliableWithFields []string
+}
+
+// globalAliasableRegistry is a map of proto field's full names to struct{} to
```
irenarindos, line 32: I think this comment might be incomplete

irenarindos (approved):

internal/alias/interceptor.go
```diff
@@ -0,0 +1,131 @@
...
+func init() {
+	// register all proto fields that are aliasable.
+	protoregistry.GlobalTypes.RangeMessages(func(m protoreflect.MessageType) bool {
+		registerAliasableFields(m.Descriptor())
+		return true
+	})
+}
+
+type registrationInfo struct {
+	unaliableWithFields []string
```
tmessi, line 29: Is this meant to be `unaliasableWithFields`?

internal/alias/interceptor.go
```diff
@@ -0,0 +1,131 @@
...
+// aliasLookup is an interface for looking up an alias by its value.
+type aliasLookup interface {
+	lookupAliasByValue(ctx context.Context, value string) (*Alias, error)
+}
+
+// ResolveAliasFields transforms the request by replacing alias values with their
+// corresponding destination ids. If no alias is found or the alias has no
+// destination id, an error is returned.
+func ResolveAliasFields(ctx context.Context, req proto.Message, lookup aliasLookup) (context.Context, error) {
+	const op = "alias.TransformRequest"
```
tmessi, line 61: The `op` does not match the function name. Also, you might consider a more self-descriptive function name, perhaps something like `alias.ResolveRequestIds`? Lastly, can you add some more details to the godoc for this function, specifically can you describe how the new proto options that can be set on `req` can impact the behavior of this?

tmessi (changes requested):
Nice, just a couple minor follow up comments/requests.

tmessi (approved):

PR #4384, opened 2024-02-09
Add Error logging and hclog.Logger debug logging throughout the daemon

talanknight:
This passes in a gorm logger when opening up the client cache daemon db.
It also turns off eventing a number of places where we are simply wrapping errors from the repositories.

The DB no longer prints out errors directly, and we have removed -store-debug flag.  If a user passes in -store-debug they are given a warning to use -log-level instead.


jimlambrt (approved):
Looks good.

PR #4393, opened 2024-02-12
fix(bsr): resolve race condition errors in signature file

ddebko:
# Summary

This PR fixes the following race condition errors:
```
Write at 0x00c005a89e50 by goroutine 4628:
  bytes.(*Buffer).Write()
      /opt/homebrew/Cellar/go/1.21.7/libexec/src/bytes/buffer.go:176 +0x38
  io.(*multiWriter).Write()
      /opt/homebrew/Cellar/go/1.21.7/libexec/src/io/multi.go:85 +0xb0
  github.com/hashicorp/boundary/internal/bsr/internal/sign.(*Writer).Write()
      /Users/ddebko/github/boundary-enterprise/internal/bsr/internal/sign/sign.go:60 +0x80
  github.com/hashicorp/boundary/internal/bsr/internal/sign.(*File).Write()
      <autogenerated>:1 +0x20
  github.com/hashicorp/boundary/internal/bsr/internal/checksum.(*File).WriteAndClose()
      /Users/ddebko/github/boundary-enterprise/internal/bsr/internal/checksum/checksum.go:148 +0x6d4
  github.com/hashicorp/boundary/internal/bsr.ChunkEncoder.Encode()
      /Users/ddebko/github/boundary-enterprise/internal/bsr/encode.go:133 +0x7b8
....

Previous write at 0x00c005a89e50 by goroutine 4629:
  bytes.(*Buffer).Write()
      /opt/homebrew/Cellar/go/1.21.7/libexec/src/bytes/buffer.go:176 +0x38
  io.(*multiWriter).Write()
      /opt/homebrew/Cellar/go/1.21.7/libexec/src/io/multi.go:85 +0xb0
  github.com/hashicorp/boundary/internal/bsr/internal/sign.(*Writer).Write()
      /Users/ddebko/github/boundary-enterprise/internal/bsr/internal/sign/sign.go:60 +0x80
  github.com/hashicorp/boundary/internal/bsr/internal/sign.(*File).Write()
      <autogenerated>:1 +0x20
  github.com/hashicorp/boundary/internal/bsr/internal/checksum.(*File).WriteAndClose()
      /Users/ddebko/github/boundary-enterprise/internal/bsr/internal/checksum/checksum.go:148 +0x6d4
  github.com/hashicorp/boundary/internal/bsr.ChunkEncoder.Encode()
      /Users/ddebko/github/boundary-enterprise/internal/bsr/encode.go:133 +0x7b8
...
```

### DIFF
--- a/internal/bsr/internal/sign/sign.go
+++ b/internal/bsr/internal/sign/sign.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"sync"
 
 	"github.com/hashicorp/boundary/internal/bsr/internal/is"
 	"github.com/hashicorp/boundary/internal/bsr/kms"
@@ -29,6 +30,8 @@ type Writer struct {
 	buf *bytes.Buffer
 	w   io.Writer
 	tee io.Writer
+
+	l sync.Mutex
 }
 
 // NewWriter returns a Writer that wraps an io.Writer.
@@ -52,6 +55,8 @@ func NewWriter(_ context.Context, w io.Writer, keys *kms.Keys) (*Writer, error) 
 }
 
 func (w *Writer) Write(b []byte) (int, error) {
+	w.l.Lock()
+	defer w.l.Unlock()
 	return w.tee.Write(b)
 }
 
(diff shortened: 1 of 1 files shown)


internal/bsr/internal/sign/sign.go
```diff
@@ -29,6 +30,8 @@ type Writer struct {
 	buf *bytes.Buffer
 	w   io.Writer
 	tee io.Writer
+
+	l sync.Mutex
```
johanbrandhorst, line 34: Could you add a comment to explain what variables this mutex is protecting? Judging by the implementation it seems to be protecting `buf`, `w` _and_ `tee`? It looks like we're taking a single writer input, then creates a tee that writes to both `buf` and `w`, which implies that writes to `tee` would impact `w` and `buf` both, so I think the logic makes sense, but it's unusual to see a single mutex protecting three variables.

PR #4389, opened 2024-02-09
fix: Proxy closing listener with 0 connections left

elimt:
A bug was introduced in `0.15.0` where sessions with max connection counts were closed when the connections left hit 0. This issue was highlighted by a user: https://github.com/hashicorp/boundary/issues/4364. What was happening was when we have zero connections left the proxy code was closing the listener which closed its connections

To fix this:
- `boundary connect  -target-id=xx -host-id=xx` command with no connections left will now wait for all connections to be closed and before exiting the session just like it did in `0.14.3`
- As soon as there are no more connections left, the listener will close to stop accepting new connections
- Fix issue where the websocket connection was throwing closing errors during the session teardown.

PR #4395, opened 2024-02-12
test(e2e): Add test to validate session connection limits

moduli:
This PR adds an e2e test to validate a target's `-session-connection-limit` setting. This arose from [this fix](https://github.com/hashicorp/boundary/pull/4389). 

The test does the following
- creates a target with a `-session-connection-limit` of 2
- creates connections until the connection limit is exhausted
- creates one more connection and confirms that you are unable to
- confirms that the session is auto-terminated


testing/internal/e2e/tests/base/target_tcp_connect_default_client_port_test.go
```diff
@@ -0,0 +1,60 @@
...
+			"connect",
+			"-target-id", newTargetId,
+			"-exec", "/bin/echo", "--",
+			"{{boundary.port}}",
+		),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+
+	parts := strings.Fields(string(output.Stdout))
+	port := parts[len(parts)-1]
```
johanbrandhorst, line 57: Maybe use `net.SplitHostPort`?

johanbrandhorst (approved):

testing/internal/e2e/tests/base/target_tcp_connect_default_client_port_test.go
moduli: `output.Stdout` contains the following
```
target_tcp_connect_default_client_port_test.go:56: 
        Proxy listening information:
          Connection Limit:    0
          Expiration:          Tue, 13 Feb 2024 17:58:52 EST
          Port:                8356
          Protocol:            tcp
          Session ID:          s_Gh7uSal94c
        8356       <-- `port` is extracting this
```

so I don't think it's applicable in this case? That would be good to know for the future. I'm going to leave this as-is.

testing/internal/e2e/tests/base/target_tcp_connect_connection_limits_test.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	// Check output from session. Session should be automatically terminated due
+	// to connection limit being reached
+	select {
+	case output := <-sessionChannel:
+		t.Log(string(output.Stdout))
+
+		// output includes multiple json objects, separated by newlines
+		parts := strings.Split(string(output.Stdout), "\n")
```
elimt, line 99: What do the JSON objects look like?

moduli: This is what the output is
```
target_tcp_connect_connection_limits_test.go:96: {"address":"127.0.0.1","port":12345,"protocol":"tcp","expiration":"2024-02-13T23:23:23.380921Z","connection_limit":0,"session_id":"s_DL2Vb3q4eR"}
        {"connections_left":1}
        {"connections_left":0}
        {"termination_reason":"No connections left in session"}
```

I *think* this was previous behavior and not introduced by the fix?

elimt: Is there any benefit in validating the last JSON output?

elimt (approved):

testing/internal/e2e/tests/base/target_tcp_connect_connection_limits_test.go
moduli: Sure. We can add a check in.

PR #4401, opened 2024-02-13
Fix linter warnings, part 1

jefferai:
This is part 1 of some work to fix linter warnings, cut off just to keep the PR size reasonable.

### DIFF
--- a/internal/credential/vault/jobs.go
+++ b/internal/credential/vault/jobs.go
@@ -132,7 +132,7 @@ func (r *TokenRenewalJob) Status() scheduler.JobStatus {
 // already running an error with code JobAlreadyRunning will be returned.
 func (r *TokenRenewalJob) Run(ctx context.Context) error {
 	const op = "vault.(TokenRenewalJob).Run"
-	if !r.running.CAS(r.running.Load(), true) {
+	if !r.running.CompareAndSwap(r.running.Load(), true) {
 		return errors.New(ctx, errors.JobAlreadyRunning, op, "job already running")
 	}
 	defer r.running.Store(false)
@@ -270,7 +270,7 @@ func nextRenewal(ctx context.Context, j scheduler.Job) (time.Duration, error) {
 	}
 	defer rows.Close()
 
-	for rows.Next() {
+	if rows.Next() {
 		type NextRenewal struct {
 			RenewalIn time.Duration
 		}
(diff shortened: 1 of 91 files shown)


internal/credential/vault/jobs.go
```diff
@@ -270,7 +270,7 @@ func nextRenewal(ctx context.Context, j scheduler.Job) (time.Duration, error) {
 	}
 	defer rows.Close()
 
-	for rows.Next() {
+	if rows.Next() {
```
johanbrandhorst, line 273: Is this change correct? I am aware that `r.ScanRows` consumes the rows, but isn't it still safer to use a `for` loop?

internal/event/options_test.go
```diff
@@ -5,8 +5,8 @@ package event
 
 import (
 	"context"
+	"crypto/rand"
```
johanbrandhorst, line 8: 😰

internal/server/root_certificate.go
```diff
@@ -89,7 +89,7 @@ func newRootCertificate(ctx context.Context, serialNumber uint64, certificate []
 ) (*RootCertificate, error) {
 	const op = "server.newRootCertificate"
 
-	if &serialNumber == nil {
```
johanbrandhorst, line 92: 😰

johanbrandhorst (approved):
👏🏻 awesome!!

internal/credential/vault/jobs.go
jefferai: I checked with @louisruch and he confirmed that we should only ever want one row (or zero). The linter had been complaining about the fact that there was an unconditional return in the for loop.

If we need to ScanRows to consume all of them or something we can... @louisruch thoughts here?

internal/event/options_test.go
jefferai: It's fine, it was seeded anyways :-D

internal/server/root_certificate.go
jefferai: :-D

internal/credential/vault/jobs.go
johanbrandhorst: Oh, I see. Yeah I mean I'd probably change the return to a break?

louisruch: This looks good to me, a break would also work

louisruch (approved):
this looks great!

internal/credential/vault/jobs.go
jefferai: Adding a break suggests that maybe if something goes wrong with the first row that we want to process a second row? But @louisruch said we should really expect only one, which the `if` will guarantee. So personally I'm leaning towards how it is now if that's okay with you two.

louisruch: `if` sounds good to me

PR #4411, opened 2024-02-14
test(e2e): Add test to check target's `-session-max-seconds`

moduli:
This PR adds an e2e test to validate the `-session-max-seconds` option on a target.

The test does the following
- creates a target with `-session-max-seconds` set to 7
- starts a session on a target
- confirms that connections created within the time limit are successful
- confirms that a connection created after the time limit will be refused

### DIFF
--- a/testing/internal/e2e/tests/base/target_tcp_connect_session_max_seconds_test.go
+++ b/testing/internal/e2e/tests/base/target_tcp_connect_session_max_seconds_test.go
@@ -1,0 +1,83 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package base_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/boundary/internal/session"
+	"github.com/hashicorp/boundary/internal/target"
+	"github.com/hashicorp/boundary/testing/internal/e2e"
+	"github.com/hashicorp/boundary/testing/internal/e2e/boundary"
+	"github.com/stretchr/testify/require"
+)
+
+// TestCliTcpTargetConnectTargetWithSessionMaxSeconds connects to a target that
+// has session max seconds defined. An error should return when trying to
+// connect to the target after the time limit has been reached.
+func TestCliTcpTargetConnectTargetWithSessionMaxSeconds(t *testing.T) {
+	e2e.MaybeSkipTest(t)
+	c, err := loadTestConfig()
+	require.NoError(t, err)
+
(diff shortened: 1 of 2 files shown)


testing/internal/e2e/tests/base/target_tcp_connect_session_max_seconds_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/hashicorp/boundary/internal/target"
+	"github.com/hashicorp/boundary/testing/internal/e2e"
+	"github.com/hashicorp/boundary/testing/internal/e2e/boundary"
+	"github.com/stretchr/testify/require"
+)
+
+// TestCliTcpTargetConnectTargetWithSessionMaxSeconds connects to a target that
+// has session max seconds defined. An error should return when trying to
+// connect to the target after the time limit has been reached.
+func TestCliTcpTargetConnectTargetWithSessionMaxSeconds(t *testing.T) {
```
moduli, line 21: I did have one test variant where...
- started `boundary connect` in a goroutine
- kept creating new channels (using `ssh localhost`) until we reach the time limit
- confirm that trying to create a new channel after the time limit results in an error

Is it worth also having this other variant as a test or this one sufficient?

johanbrandhorst: Probably worth having a test to see that new connections are rejected, in addition to existing ones being torn down, yeah.

moduli: Added another test

testing/internal/e2e/tests/base/target_tcp_connect_session_max_seconds_test.go
```diff
@@ -0,0 +1,159 @@
...
+		if output.Err != nil {
+			break
+		}
+
+		// Ensure that time limit has not been reached yet
+		// (plus a small buffer)
+		require.Less(t, time.Since(start).Seconds(), float64(sessionMaxSeconds+1))
+		time.Sleep(time.Second)
+	}
+}
```
johanbrandhorst, line 159: Should we not test here that enough time has passed? Won't this test pass today if we reject sooner than configured?

moduli: Good catch. I originally had something in there and removed it because I thought it simplified the test... not realizing that didn't test that part.

johanbrandhorst (approved):

stasryzhov (approved):

PR #4421, opened 2024-02-16
Vet/race fixes

jefferai:
This fixes all vet issues and all races. It also enables the race detector as part of the CI workflows.


internal/cmd/base/logging.go
```diff
@@ -59,77 +61,141 @@ func ProcessLogLevelAndFormat(flagLogLevel, flagLogFormat, configLogLevel, confi
...
+		logger: new(atomic.Pointer[grpcLogFakerLogHolder]),
+	}
+	grpclog.SetLoggerV2(grpcLogFaker)
+}
+
 type GRPCLogFaker struct {
-	Logger hclog.Logger
-	Log    bool
+	logger *atomic.Pointer[grpcLogFakerLogHolder]
+	log    *atomic.Bool
```
johanbrandhorst, line 80: Maybe name this something like `loggingEnabled` while we're here? It was a bit confusing to me to see `logger` and `log` in the same struct.
```suggestion
	loggingEnabled    *atomic.Bool
```

internal/cmd/commands/server/listener_reload_test.go
```diff
@@ -83,7 +83,6 @@ listener "tcp" {
 `
 
 func TestServer_ReloadListener(t *testing.T) {
-	t.Parallel()
```
johanbrandhorst, line 86: Could you add a comment with a warning that this test cannot be run in parallel (and why) so we don't reintroduce this accidentally?

internal/daemon/cluster/handlers/upstream_message_service_controller.go
```diff
@@ -19,7 +19,11 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-var upstreamMessageHandler sync.Map
+var upstreamMessageHandler *sync.Map
+
+func init() {
+	upstreamMessageHandler = new(sync.Map)
```
johanbrandhorst, line 25: Can't we do this in the declaration instead of during init?

johanbrandhorst (commented):
Are we also enabling race testing to prevent regressions or new failures? This is great ⭐

jefferai:
>Are we also enabling race testing to prevent regressions or new failures?

Yes, but I think I'll do that when I'm done with the rest of the linting, and enable that at the same time.

internal/cmd/base/logging.go
jefferai: Done.

internal/daemon/cluster/handlers/upstream_message_service_controller.go
jefferai: Ah yeah fair, fixed

internal/cmd/commands/server/listener_reload_test.go
jefferai: That actually I will revert...it was left over from trying to figure out one of the data races.

jimlambrt (approved):
ty!

PR #4423, opened 2024-02-16
test(e2e): Verify long scripts complete successfully on tcp targets

stasryzhov:
This PR adds an e2e-test to verify that SSH requests sent to target can execute long-lasting scripts successfully.

The test saves a ~10s-long script on the target and sends two SSH requests:
- to execute the script saved on the target
- to execute a script sent with the SSH request

This PR relates to this [bug fix](https://github.com/hashicorp/boundary-enterprise/pull/926). The test for ssh-targets will be added to boundary-enterprise in a separate PR.


louisruch (approved):
LGTM

PR #4431, opened 2024-02-20
test(e2e): Add ssh log for failed requests

stasryzhov:
This PR adjusts the e2e-test verifying long lasting scripts can be executed and completed successfully on targets.
It does the following:
- adds more details to failed assertions
- shortens the script to 5 seconds which is sufficient for the test


testing/internal/e2e/tests/base/target_tcp_connect_exec_long_lasting_script_test.go
```diff
@@ -86,34 +85,25 @@ func TestCliTcpTargetConnectExecLongLastingScript(t *testing.T) {
 	require.NoError(t, output.Err, string(output.Stderr))
 
 	// Send SSH requests to the target to execute long-lasting scripts
-	var eg errgroup.Group
-	eg.Go(func() error {
-		t.Log("Executing the long-lasting script saved on the target...")
-		cmd := exec.CommandContext(ctx, "/usr/bin/ssh",
-			"-v",
-			"-l", c.TargetSshUser,
-			"-i", c.TargetSshKeyPath,
```
moduli, line 95: Doesn't this command need credentials for it to work?

testing/internal/e2e/tests/base/target_tcp_connect_exec_long_lasting_script_test.go
```diff
@@ -86,34 +85,25 @@ func TestCliTcpTargetConnectExecLongLastingScript(t *testing.T) {
...
-			"-o", "UserKnownHostsFile=/dev/null",
-			"-o", "StrictHostKeyChecking=no",
-			"-p", proxyPort,
-			"localhost",
-			"sleep 10")
-		return cmd.Run()
-	})
+	t.Log("Executing the long-lasting script saved on the target...")
+	cmd := exec.CommandContext(ctx, "/usr/bin/ssh",
```
moduli, line 89: Is there a specific reason we're using this as opposed to `e2e.RunCommand` (which uses `exec.CommandContext` under the hood)? Not sure if I have a strong opinion, but it would maintain the convention we have throughout most of the test suite.
```
output := e2e.RunCommand(...)
require.NoError(t, output.Err, string(output.Stderr))

```

testing/internal/e2e/tests/base/target_tcp_connect_exec_long_lasting_script_test.go
stasryzhov: Good catch! Got it back 👍

testing/internal/e2e/tests/base/target_tcp_connect_exec_long_lasting_script_test.go
stasryzhov: Agreed.

I was using the e2e.RunCommand for boundary CLI commands, and the standard library for everything else.

moduli (approved):

PR #4436, opened 2024-02-21
test(e2e): Add test for alias

moduli:
This PR adds an e2e test for aliases.

The test does the following
- creates a target with an address
- creates an alias to that target
- connects to the target (`boundary connect`) using the alias
- connects to the target (`boundary connect ssh`) using the alias
- authorize a session (`boundary targets authorize-session`) using the alias
- add an `-authorize-session-host-id` to the alias and authorize a session using the alias again
- attempt to use an alias that doesn't exist

### DIFF
--- a/testing/internal/e2e/tests/base/alias_test.go
+++ b/testing/internal/e2e/tests/base/alias_test.go
@@ -35,6 +35,8 @@ func TestCliAlias(t *testing.T) {
 		require.NoError(t, output.Err, string(output.Stderr))
 	})
 	newProjectId := boundary.CreateNewProjectCli(t, ctx, newOrgId)
+
+	// Create an alias for a target
 	newTargetId := boundary.CreateNewTargetCli(
 		t,
 		ctx,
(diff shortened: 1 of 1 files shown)


irenarindos (commented):
Do you think it'd be worth testing using an alias to do a read on a target?
Could also test deleting a target by alias and checking that the destination ID is cleared out from the alias along with the host ID

moduli:
> Do you think it'd be worth testing using an alias to do a read on a target? Could also test deleting a target by alias and checking that the destination ID is cleared out from the alias along with the host ID

@irenarindos Good ideas! I did notice, though, that we've added bats tests for those already? https://github.com/hashicorp/boundary/blame/llb-alias/internal/tests/cli/boundary/alias.bats. Is there something in an end-to-end setting that (or something that the bats tests doesn't really cover) that we should add?

irenarindos:
> > Do you think it'd be worth testing using an alias to do a read on a target? Could also test deleting a target by alias and checking that the destination ID is cleared out from the alias along with the host ID
> 
> @irenarindos Good ideas! I did notice, though, that we've added bats tests for those already? https://github.com/hashicorp/boundary/blame/llb-alias/internal/tests/cli/boundary/alias.bats. Is there something in an end-to-end setting that (or something that the bats tests doesn't really cover) that we should add?

Would it be worth trying against a host set where we don't specify a host id in the alias?

irenarindos (approved):

PR #4441, opened 2024-02-22
Add boundary connect alias bats tests

talanknight:
Adding tests that validate the authorize-session requests when a host-id is provided when an alias is used, both configured and not configured with a host id already.  Note that a few of the tests were rearranged to not be duplicated.

### DIFF
--- a/internal/tests/cli/boundary/alias.bats
+++ b/internal/tests/cli/boundary/alias.bats
@@ -49,7 +49,7 @@ export NEW_HOST="host_for_alias_test"
   [ "$status" -eq 0 ]
 }
 
-@test "boundary/alias: admin user can connect using passing in a host id" {
+@test "boundary/alias: admin user can connect passing in a host id" {
   run connect_alias_with_host_id $ALIAS_VALUE $DEFAULT_HOST
   [ "$status" -eq 0 ]
 }


internal/tests/cli/boundary/_connect.bash
```diff
@@ -11,4 +11,11 @@ function connect_alias() {
   local alias=$1
   # Note: When this command returns, the session immediately goes into a "canceling" state
   echo "foo" | boundary connect $alias -exec nc -- {{boundary.ip}} {{boundary.port}}
+}
+
+function connect_alias_with_hostid() {
```
moduli, line 16: Looks like other methods use `host_id`?
```suggestion
function connect_alias_with_host_id() {
```

moduli (approved):

internal/tests/cli/boundary/alias.bats
```diff
@@ -36,12 +39,60 @@ export UPDATE_ALIAS_VALUE='test.change'
...
+  run read_target $ALIAS_VALUE
+  [ "$status" -eq 0 ]
+}
+
+@test "boundary/alias: admin user can connect to target using an alias" {
+  run connect_alias $ALIAS_VALUE
+  [ "$status" -eq 0 ]
+}
+
+@test "boundary/alias: admin user can connect using passing in a host id" {
```
irenarindos, line 52:
```suggestion
@test "boundary/alias: admin user can connect passing in a host id" {
```

irenarindos (approved):

PR #4446, opened 2024-02-23
test(e2e): Add multi-scope grants test

stasryzhov:
This PR adds an e2e-test to verify scope grants can be applied to multiple scopes.

The test creates a User and a Role with admin Grants in global scope.
It verifies that the Grants are applied:
- nowhere if the Role does not have the User as a principle
- to the global scope if the Role has the User as a principle
- to the global and org scopes if the Role's grant scopes = `[this, children]`
- to the global, org, and proj scopes if the Role's grant scopes = `[this, descendant]`

Also the test refactors the `CreateRole` function making it more conventional; it now returns `roleId` and `error` that's handled where the function is called.


testing/internal/e2e/boundary/role.go
```diff
@@ -70,4 +80,59 @@ func AddPrincipalToRoleCli(t testing.TB, ctx context.Context, roleId string, pri
...
+}
+
+// SetGrantScopesToRoleCli uses Boundary CLI to override grant scopes for the role with the provided ones.
+// Option WithGrantScopeId can be used multiple times to provide grant scope IDs.
+func SetGrantScopesToRoleCli(t testing.TB, ctx context.Context, roleId string, opt ...RoleOption) error {
+	opts := getRoleOpts(opt...)
+	var args []string
+
+	if len(opts.scopeIds) == 0 {
+		return fmt.Errorf("at least one grant scope id must be provided")
```
moduli, line 93: Not sure what my opinion is yet, but a thought I was having was that it felt weird to have this required parameter that is optional in the method definition

testing/internal/e2e/tests/base/role_multi_scope_grants_test.go
```diff
@@ -0,0 +1,165 @@
...
+	newRoleId, err := boundary.CreateRoleCli(t, ctx, "global")
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		boundary.AuthenticateAdminCli(t, ctx)
+		output := e2e.RunCommand(ctx, "boundary",
+			e2e.WithArgs("roles", "delete", "-id", newRoleId),
+		)
+		require.NoError(t, output.Err, string(output.Stderr))
+	})
+	boundary.AddGrantToRoleCli(t, ctx, newRoleId, "ids=*;type=*;actions=*")
```
moduli, line 66: Would it make sense to move this down after line 84 since that's where it's used? Feels like it makes it more logical to read.

testing/internal/e2e/tests/base/role_multi_scope_grants_test.go
```diff
@@ -0,0 +1,165 @@
...
+func TestCliApplyGrantsForMultipleScopes(t *testing.T) {
+	e2e.MaybeSkipTest(t)
+	bc, err := boundary.LoadConfig()
+	require.NoError(t, err)
+
+	ctx := context.Background()
+	boundary.AuthenticateAdminCli(t, ctx)
+
+	// Create Org and Project
+	newOrgId := boundary.CreateNewOrgCli(t, ctx)
```
moduli, line 25: We could consider removing `new` from these variable names, too

testing/internal/e2e/tests/base/role_multi_scope_grants_test.go
```diff
@@ -0,0 +1,165 @@
...
+	_, err = boundary.CreateRoleCli(t, ctx, newOrgId)
+	require.Error(t, err)
+
+	_, err = boundary.CreateRoleCli(t, ctx, newProjectId)
+	require.Error(t, err)
+
+	// Set Grant Scopes to Role: this, children
+	boundary.AuthenticateAdminCli(t, ctx)
+	err = boundary.SetGrantScopesToRoleCli(t, ctx, newRoleId,
+		boundary.WithGrantScopeId("this"),
```
moduli, line 111: I didn't know `this` was a valid `grant-scope-id`... what if you didn't have `this` and just `children`?

testing/internal/e2e/tests/base/role_multi_scope_grants_test.go
```diff
@@ -0,0 +1,165 @@
...
+	boundary.AuthenticateCli(t, ctx, bc.AuthMethodId, acctName, acctPassword)
+	newGlobalRoleId2, err := boundary.CreateRoleCli(t, ctx, "global")
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		boundary.AuthenticateAdminCli(t, ctx)
+		output := e2e.RunCommand(ctx, "boundary",
+			e2e.WithArgs("roles", "delete", "-id", newGlobalRoleId2),
+		)
+		require.NoError(t, output.Err, string(output.Stderr))
+	})
```
moduli, line 100: Curious if there's any other action we can use to check here besides creating a role (so that we don't have to clean up more stuff)

testing/internal/e2e/tests/base/role_multi_scope_grants_test.go
stasryzhov: Yes, it would 👍

testing/internal/e2e/tests/base/role_multi_scope_grants_test.go
stasryzhov: Then the grants will apply only in children scopes and not in the role's scope

testing/internal/e2e/tests/base/role_multi_scope_grants_test.go
stasryzhov: I've chosen roles as it's a simple entity that can be created in each scope: global, org, proj.

We can get rid of two clean up calls if instead of attempting to create a role in global scope we'd attempt to update an existing one.

Then everything in org and proj scopes would be cleaned up by deleting the org.
And in global scope we'd need to delete the test-admin role only.

I feel that both approaches have its advantages; the alternative approach does save time but checks different operations (i.e. update, create) which does not feel 100% correct.

What do you think?

testing/internal/e2e/boundary/role.go
stasryzhov: I had the same feeling and checked what approach we had in the code base. I found that we used options for required fields when using api clients, e.g. names for targets

```go
newTargetResult, err := tClient.Create(ctx, "tcp", projectId,
		targets.WithName("e2e Target"),
		targets.WithTcpTargetDefaultPort(uint32(targetPort)),
)
```

Alternatively, I thought about adding an array of grant scopes instead of options. I don't dislike it either 🙂

moduli: I see. If it's an existing pattern, then we can keep as-is.

testing/internal/e2e/tests/base/role_multi_scope_grants_test.go
moduli: > but checks different operations (i.e. update, create) which does not feel 100% correct.

Could you elaborate on this? Wouldn't it just be entirely dependent on the permissions we're granting? What if it was something like listing roles or something?

stasryzhov: > Could you elaborate on this?
I want the test to be doing a single operation, for example create role, in each scope under test.
I don't want the test to be doing one operation in global scope and another operation in org and proj scopes.

Listing roles will suffice as well.

moduli (approved):

testing/internal/e2e/boundary/role.go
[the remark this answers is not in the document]
stasryzhov: I double checked, there is no usage of this method in `_ent` suffixed files in `boundary-enterprise`

PR #4454, opened 2024-02-26
dev: use URL for web target and add caveat

irenarindos:
Alias names need to match a web target address; the `web.boundary.dev` alias was failing to connect.


johanbrandhorst (approved):
Ah, nice!

jimlambrt (approved):

PR #4465, opened 2024-02-28
Add sending auth token automatically to ferry daemon

talanknight:
This PR extracts out a common CommandWrapper which can be shared between the client cache daemon's start and add-token commands and ferry's automatic sending the auth token logic

### DIFF
--- a/internal/clientcache/cmd/daemon/wrapper_register.go
+++ b/internal/clientcache/cmd/daemon/wrapper_register.go
@@ -112,7 +112,6 @@ func addTokenToCache(ctx context.Context, baseCmd *base.Command, token string) b
 // provided context is done. It returns an error if the unix socket is not found
 // before the context is done.
 func waitForDaemon(ctx context.Context) error {
-	const op = "wrapper.waitForDaemon"
 	dotPath, err := DefaultDotDirectory(ctx)
 	if err != nil {
 		return err
--- a/internal/cmd/commands/ferry/wrapper_register.go
+++ b/internal/cmd/commands/ferry/wrapper_register.go
@@ -22,14 +22,12 @@ func hook(ctx context.Context, baseCmd *base.Command, token string) {
 	}
 	client, err := baseCmd.Client()
 	if err != nil {
-		// print this error out to stderr?
+		baseCmd.PrintCliError(err)
 		return
 	}
 	if token != "" {
 		client.SetToken(token)
 	}
-	// TODO: Figure out the appropriate reporting of these errors when this is
-	// only run in the background of other commands.
 	_, apiErr, err := addToken(ctx, client, baseCmd.FlagFerryDaemonPort)
 	if err != nil {
 		baseCmd.PrintCliError(err)


internal/cmd/commands/ferry/addtoken.go
```diff
@@ -0,0 +1,187 @@
...
+	*base.Command
+}
+
+func (c *AddTokenCommand) Synopsis() string {
+	return "Add an auth token to a running Boundary ferry daemon"
+}
+
+func (c *AddTokenCommand) Help() string {
+	helpText := `
+Usage: boundary ferry add-token [options]
```
johanbrandhorst, line 33: Should we just make this `boundary ferry authenticate`?

internal/cmd/commands/ferry/addtoken.go
```diff
@@ -0,0 +1,187 @@
...
+		Usage:      "Addr of the Boundary controller, as a complete URL (e.g. https://boundary.example.com:9200).",
+	})
+
+	f = set.NewFlagSet("Client Options")
+
+	f.StringVar(&base.StringVar{
+		Name:   "token-name",
+		Target: &c.FlagTokenName,
+		EnvVar: base.EnvTokenName,
+		Usage:  `If specified, the given value will be used as the name when storing the token in the system credential store. This can allow switching user identities for different commands.`,
```
johanbrandhorst, line 63: How is this used with Ferry?

internal/cmd/commands/ferry/addtoken.go
```diff
@@ -0,0 +1,187 @@
...
+	})
+
+	f.StringVar(&base.StringVar{
+		Name:    "keyring-type",
+		Target:  &c.FlagKeyringType,
+		Default: "auto",
+		EnvVar:  base.EnvKeyringType,
+		Usage:   `The type of keyring to use. Defaults to "auto" which will use the Windows credential manager, OSX keychain, or cross-platform password store depending on platform. Set to "none" to disable keyring functionality. Available types, depending on platform, are: "wincred", "keychain", "pass", and "secret-service".`,
```
johanbrandhorst, line 71: How does this work with Ferry?

internal/cmd/commands/ferry/addtoken.go
```diff
@@ -0,0 +1,187 @@
...
+	case "table":
+		c.UI.Output("The ferry add-token operation completed successfully.")
+	}
+	return base.CommandSuccess
+}
+
+// userTokenToAdd is the request body to this handler.
+type UpsertTokenRequest struct {
+	// BoundaryAddr is a required field for all requests
+	BoundaryUrl string `json:"boundary_url,omitempty"`
```
johanbrandhorst, line 141: Ditto `boundary_addr`

internal/cmd/commands/ferry/addtoken.go
```diff
@@ -0,0 +1,187 @@
...
+// method to pass in a silent UI to suppress any output.
+func (c *AddTokenCommand) Add(ctx context.Context, ui cli.Ui, apiClient *api.Client) (*api.Response, *api.Error, error) {
+	pa := UpsertTokenRequest{
+		BoundaryUrl: apiClient.Addr(),
+	}
+	token := apiClient.Token()
+	if token == "" {
+		return nil, nil, errors.New("The client auth token is empty.")
+	}
+	if parts := strings.SplitN(token, "_", 4); len(parts) != 3 {
```
johanbrandhorst, line 159: Why 4? I think this should be 3? The format we expect is `at_1234567890_awduawodnawodinaw21en1o23h1823h1n` or something like that, right? So we want to split this into at most 3 substrings.

internal/cmd/commands/ferry/addtoken.go
```diff
@@ -0,0 +1,187 @@
...
+
+	ferryClient, err := api.NewClient(&api.Config{
+		Addr:             ferryAddress(c.FlagFerryDaemonPort),
+		OutputCurlString: c.FlagOutputCurlString,
+	})
+	if err != nil {
+		return nil, nil, err
+	}
+
+	req, err := ferryClient.NewRequest(ctx, "POST", "/token", pa)
```
johanbrandhorst, line 177: Changing this to `/v1/tokens`

johanbrandhorst (commented):
Neat!

internal/cmd/commands/ferry/addtoken.go
johanbrandhorst: What is the use case for this command? As a user, sharing the token with Ferry should be automatic or triggered with `boundary authenticate` IMO. This feels like the wrong UX?

internal/cmd/commands/ferry/addtoken.go
johanbrandhorst: Ah, so you want to use `SplitN` where N=number of expected sections+1 because `Split` is potentially more expensive? I get it but I think it obscures the intent. How about just using `Split`?

internal/cmd/commands/ferry/addtoken.go
johanbrandhorst: I see. Do you think we need this for testing? I'd rather we implement multiplexing in Ferry before GA than need a command like this, so maybe we don't merge this new command?

internal/cmd/wrapper/command_wrapper.go
```diff
@@ -0,0 +1,74 @@
...
+	"sync"
+
+	"github.com/hashicorp/boundary/internal/cmd/base"
+	"github.com/mitchellh/cli"
+)
+
+var ErrCallbackAlreadyRegistered = errors.New("registered callback already exists")
+
+// PostSuccessfulCommandCallback is a function that should be run after the
+// wrapped command completes succesfully. token will contain the value for any
```
irenarindos, line 18:
```suggestion
// wrapped command completes successfully. token will contain the value for any
```

internal/cmd/base/base.go
```diff
@@ -68,6 +68,7 @@ const (
 	EnvKeyringType     = "BOUNDARY_KEYRING_TYPE"
 	envRecoveryConfig  = "BOUNDARY_RECOVERY_CONFIG"
 	envSkipCacheDaemon = "BOUNDARY_SKIP_CACHE_DAEMON"
+	envSkipFerry       = "BOUNDARY_SKIP_FERRY"
```
irenarindos, line 71: Should we add _DAEMON for consistency?
```suggestion
	envSkipFerry       = "BOUNDARY_SKIP_FERRY_DAEMON"
```

internal/cmd/commands/ferry/addtoken.go
```diff
@@ -0,0 +1,205 @@
...
+func (c *AddTokenCommand) Flags() *base.FlagSets {
+	set := c.FlagSet(base.FlagSetOutputFormat)
+	f := set.NewFlagSet("Connection Options")
+
+	f.StringVar(&base.StringVar{
+		Name:       base.FlagNameAddr,
+		Target:     &c.FlagAddr,
+		EnvVar:     api.EnvBoundaryAddr,
+		Completion: complete.PredictAnything,
+		Usage:      "Addr of the Boundary controller, as a complete URL (e.g. https://boundary.example.com:9200).",
```
irenarindos, line 59:
```suggestion
		Usage:      "Address of the Boundary controller, as a complete URL (e.g. https://boundary.example.com:9200).",
```

internal/cmd/commands/ferry/addtoken.go
```diff
@@ -0,0 +1,70 @@
...
+	"fmt"
+	"io"
+	"strings"
+	"time"
+
+	"github.com/hashicorp/boundary/api"
+	"github.com/hashicorp/go-retryablehttp"
+)
+
+// userTokenToAdd is the request body to this handler.
```
johanbrandhorst, line 20:
```suggestion
// UpsertTokenRequest is the request body to this handler.
```

johanbrandhorst (approved):
⭐

internal/clientcache/cmd/daemon/wrapper_register.go
```diff
@@ -138,7 +112,7 @@ func (w *CommandWrapper) addTokenToCache(ctx context.Context, token string) bool
 // provided context is done. It returns an error if the unix socket is not found
 // before the context is done.
 func waitForDaemon(ctx context.Context) error {
-	const op = "daemon.waitForDaemon"
+	const op = "wrapper.waitForDaemon"
```
irenarindos, line 115:
```suggestion
	const op = "daemon.waitForDaemon"
```

internal/cmd/commands/ferry/wrapper_register.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+}
+
+func hook(ctx context.Context, baseCmd *base.Command, token string) {
+	if baseCmd.FlagSkipFerry {
+		return
+	}
+	client, err := baseCmd.Client()
+	if err != nil {
+		// print this error out to stderr?
```
irenarindos, line 25: Should this be a TODO or have a print statement here?

irenarindos (approved):
Looks great, just two minor things

PR #4478, opened 2024-03-04
test(e2e): Provide public ca key to UI tests

moduli:
This PR updates the `docker_openssh_server_ca_key` module to return the public CA key so that it can be passed to the UI tests. This is to support the an Admin UI SSH certificate injection test.

I was trying to figure out how to generate the public key from the private key within the Admin UI test itself (using `node.js`), but I wasn't able to find a solution that worked for all keys without adding additional dependencies in the UI project. Also, the built-in `crypto` node module didn't supported OpenSSH keys.

### DIFF
--- a/enos/modules/test_e2e_ui/main.tf
+++ b/enos/modules/test_e2e_ui/main.tf
@@ -63,6 +63,16 @@ variable "target_port" {
   type        = string
   default     = ""
 }
+variable "target_ca_key" {
+  description = "CA Private Key (base64 encoded)"
+  type        = string
+  default     = ""
+}
+variable "target_ca_key_public" {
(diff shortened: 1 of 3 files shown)


enos/modules/test_e2e_ui/main.tf
```diff
@@ -63,6 +63,16 @@ variable "target_port" {
   type        = string
   default     = ""
 }
+variable "target_ca_key" {
```
stasryzhov, line 66: Maybe call it `target_ca_key_private` for the sake of being as explicit as with the `target_ca_key_public`?

stasryzhov (approved):

PR #4488, opened 2024-03-06
Add some more fixes caught by race detector

jefferai:


moduli (approved):
Reviewed the parts that I understood, and it seems to make sense 😅 

Thanks for addressing these!

jimlambrt (approved):
ty!

PR #4504, opened 2024-03-09
Create aliases when creating a target in repo

talanknight:
Follow up PRs will 
* add the relevant fields to the proto and service handlers
* add the sdk and CLI changes


internal/target/options.go
```diff
@@ -275,3 +277,10 @@ func WithStartPageAfterItem(item pagination.Item) Option {
 		o.WithStartPageAfterItem = item
 	}
 }
+
+// WithAliases provides an option to provide aliases.
+func WithAliases(in []*talias.Alias) Option {
```
irenarindos, line 282: Can options test be updated to cover this?

internal/target/tcp/repository_tcp_target_test.go
```diff
@@ -251,6 +252,31 @@ func TestRepository_CreateTarget(t *testing.T) {
 			// add an appropriate assert.
 		})
 	}
+
+	t.Run("create with aliases", func(t *testing.T) {
```
irenarindos, line 256: Can you add error cases where two of the same aliases are added, and an invalid alias?

PR #4508, opened 2024-03-11
Add handler and API support for alias creation with target

talanknight:
Added the a "with_alias" field on targets.

Even though we now have a shared Alias message in the target resource proto package, there are tests to ensure responses only include id and value set and requests only have value, scope id, and host id set.  I am interested in the reviewers opinion on reusing the Alias proto message like this.

### DIFF
--- a/internal/daemon/controller/handlers/targets/tcp/target_service_test.go
+++ b/internal/daemon/controller/handlers/targets/tcp/target_service_test.go
@@ -983,9 +983,10 @@ func TestCreate(t *testing.T) {
 
 	org, proj := iam.TestScopes(t, iamRepo)
 	at := authtoken.TestAuthToken(t, conn, kms, org.GetPublicId())
-	r := iam.TestRole(t, conn, proj.GetPublicId())
+	r := iam.TestRole(t, conn, "global")
 	_ = iam.TestUserRole(t, conn, r.GetPublicId(), at.GetIamUserId())
 	_ = iam.TestRoleGrant(t, conn, r.GetPublicId(), "ids=*;type=*;actions=*")
+	_ = iam.TestRoleGrantScope(t, conn, r.GetPublicId(), globals.GrantScopeDescendants)
 
 	// Ensure we are using the OSS worker filter function
 	workerFilterFn := targets.AuthorizeSessionWorkerFilterFn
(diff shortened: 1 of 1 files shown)


internal/daemon/controller/handlers/targets/tcp/target_service_test.go
```diff
@@ -1035,6 +1035,120 @@ func TestCreate(t *testing.T) {
...
+					{
+						Id:      "alt_1234567890",
+						Value:   "id-specified",
+						ScopeId: "global",
+					},
+				},
+			}},
+			err: handlers.ApiErrorWithCode(codes.InvalidArgument),
+		},
+		{
```
irenarindos, line 1108: Should we test creating an alias with a non-global and other invalid scope?

internal/gen/controller.swagger.json
```diff
@@ -6924,6 +6924,23 @@
...
+        }
+      },
+      "description": "Alias is the alias information related to the aliases associated with the target."
+    },
+    "controller.api.resources.targets.v1.AuthorizeSessionArguments": {
+      "type": "object",
+      "properties": {
+        "host_id": {
+          "type": "string",
+          "description": "host_id is the id of the host that the session will be authorized for.\nWhen specified authorizing a session using this alias will have the same\neffect of authorizing a session to the aliase's destination_id and passing\nin this value through the -host-id flag. If the host-id flag is also\nspecified when calling authorize-session an error will be returned unless\nthe provided host-id matches this value."
```
irenarindos, line 6943:
```suggestion
          "description": "host_id is the id of the host that the session will be authorized for.\nWhen specified authorizing a session using this alias will have the same\neffect of authorizing a session to the alias' destination_id and passing\nin this value through the -host-id flag. If the host-id flag is also\nspecified when calling authorize-session an error will be returned unless\nthe provided host-id matches this value."
```

internal/proto/controller/api/resources/targets/v1/target.proto
```diff
@@ -14,12 +14,31 @@ import "google/protobuf/wrappers.proto";
...
+}
+
+message TargetAliasAttributes {
+  AuthorizeSessionArguments authorize_session_arguments = 1 [json_name = "authorize_session_arguments"]; // @gotags: `class:"public"`
+}
+
+message AuthorizeSessionArguments {
+  // host_id is the id of the host that the session will be authorized for.
+  // When specified authorizing a session using this alias will have the same
+  // effect of authorizing a session to the aliase's destination_id and passing
```
irenarindos, line 37:
```suggestion
  // effect of authorizing a session to the alias' destination_id and passing
```

internal/proto/controller/api/resources/targets/v1/target.proto
```diff
@@ -222,6 +241,10 @@ message Target {
   // Output only. The aliases that point to this Target.
   repeated Alias aliases = 550 [json_name = "aliases"]; // @gotags: `class:"public"`
 
+  // Input only. with_aliases specify the aliases that should be created when
+  // the target is created.  If this field is only usable at target creation time.
```
irenarindos, line 245:
```suggestion
  // the target is created.  This field is only usable at target creation time.
```

internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -1305,7 +1305,21 @@ func (s Service) createInRepo(ctx context.Context, item *pb.Target) (target.Targ
 	if err != nil {
 		return nil, nil, nil, err
 	}
-	out, err := repo.CreateTarget(ctx, u)
+
+	createOptions := []target.Option{}
```
irenarindos, line 1309: Do we need to do some kind of check to ensure the user is authorized to create aliases?

internal/proto/controller/api/resources/targets/v1/target.proto
```diff
@@ -14,12 +14,31 @@ import "google/protobuf/wrappers.proto";
...
 message Alias {
   // Output only. The ID of the alias referencing this target.
   string id = 10; // @gotags: `class:"public" eventstream:"observation"`
 
   // Ouput only. The value of the alias referencing this target.
   string value = 20; // @gotags: `class:"public" eventstream:"observation"`
+
+  string scope_id = 30 [json_name = "scope_id"]; // @gotags: `class:"public" eventstream:"observation"`;
+
+  TargetAliasAttributes attributes = 40;
```
irenarindos, line 27: Is there a reason not to use the alias proto's TargetAliasAttributes here?

irenarindos (approved):
Looks great!

tmessi (approved):

PR #4512, opened 2024-03-12
Add CLI flags for creating an alias with a target

talanknight:
Even though the sdk and HTTP API allows multiple aliases to be created at the same time as a target, the CLI only allows one.
This is because there isn't a clear way in how to associate the value, scope id, and the host id for each alias with multiple aliases without using some delimiting or ordering scheme which will inevitably confuse the CLI user.


irenarindos (approved):
Looks great, should we update the bats tests to use these new flags?

PR #4535, opened 2024-03-15
Fix issue with workers connecting in high latency conditions

jefferai:
At some point we migrated to grpc ConnectParams to control how often retries were happening. One value that we didn't set was the MinConnectTimeout. It turns out that if you don't set this, it won't use the default; instead if will actually use a value of zero, which gRPC internally ups to something that seems to be around 25ms to meet a minimum threshold. This is not documented in the function comments.

This commit simply sets the value of MinConnectTimeout to the unexported gRPC default.


jimlambrt (approved):
Ty

PR #4545, opened 2024-03-20
chore: Add edu team as codeowners for website

moduli:
This PR adds the docs team as codeowners for any changes in the `website/content` directory

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -6,3 +6,7 @@
 
 /.release/                               @hashicorp/github-secure-boundary
 /.github/workflows/build.yml             @hashicorp/github-secure-boundary
+
+# education
+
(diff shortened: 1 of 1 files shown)


CODEOWNERS
moduli, line 1: @tmessi Double-checking something... I've noticed that `boundary` doesn't get tagged in PRs. Is that the intention? Looking at repos like `boundary-local-env`, it looks like the rule should be...
```
* @hashicorp/boundary
```

?

moduli: Chatted with Tim. We'll take this conversation to get opinions from the rest of the team. Doesn't affect this PR.

Dan-Heath (approved):
Thank you!!

PR #4551, opened 2024-03-20
Updated Vault integration section with Client Count

dannyjknights:
Updated the integration section with Vault to include a dedicated section to highlight the Vault client count implications when integrating Boundary and Vault


website/content/docs/integrations/vault/index.mdx
```diff
@@ -53,8 +53,22 @@ Vault is configured to act as the certificate authority (CA), to ensure that the
 
 `/sign` - The Boundary controllers generate the key pair and then send it to Vault to sign.
 
+When you connect to a target, which uses Vault for dynamic SSH certificates, a new certificate is generated for every target connection. As long as the target trusts the CA, then access is granted without you having any visibility into the credentials involved.
+
+# Vault clients
```
Dan-Heath, line 58: I'm wondering if we can get rid of this heading since there is no content below it. Alternatively, we could make this a level 2 heading (there should only be 1 level 1 at the top of the page) and add a short introductory paragraph about how you could use Vault clients for secret management and IdP, and then make the following 2 sections level 3 headings.

Dan-Heath (commented):
This is great Danny, thank you! I'm adding a few suggestions for minor rewrites. Then I think we should see what the rest of the integration work looks like before we merge it. But I think this is a great addition to this topic, and definitely on the right track.

website/content/docs/integrations/vault/index.mdx
dannyjknights: Thanks Dan, yes I didn't know how best to handle this, so just left it how it was as I knew you'd have some recommendations. I am not precious about it, so whatever you think is best, based on your expert knowledge

website/content/docs/integrations/vault/index.mdx
```diff
@@ -53,8 +53,22 @@ Vault is configured to act as the certificate authority (CA), to ensure that the
...
+
+Organizations may have valid reasons for multiple credential stores within Boundary. As you can only create credential stores at the project level, there may be different projects within an org scope that have different requirements from their credential stores. For example, within an org scope you may have two project scopes: Database and Compute. These two projects need to be isolated and therefore would have a dedicated credential store per project.
+
+When it is integrated with Vault, Boundary has to be assigned a periodic, renewable, orphan token from Vault. Each credential store that is created within Boundary that leverages Vault equates to one Vault token being created. The number of Boundary targets that source credentials from the stores, the number of users connecting to the targets, the number of sessions that get created, or how many credential libraries the credential store contains, all have no impact on the client count in Vault.
```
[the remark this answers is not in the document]
dannyjknights: I went more verbose in the list of things that do not affect the client count so it couldn't get any more black and white, just down to the nature of how client counts can confuse things :) Happy to take your lead if you think it reads better

Dan-Heath (approved):
This looks good to me 🎉 Thank you for all your hard work on it @dannyjknights ! This will be a great addition for our users.

PR #4556, opened 2024-03-21
Added sample Terraform code for session recording configuration

carlosvilleg:
Added sample Terraform code for session recording configuration

Also added HCP Boundary *Plus* as requirement for session recording

Added missing group attributes on pre-existing tabs


Dan-Heath (approved):
Thank you for your contribution @carlosvilleg ! This is a great addition and something our users have been asking for. Apologies it took so long to get it merged. We had some other updates and conflicts to resolve first.

Thank you for participating in Docs Days!!

PR #4567, opened 2024-03-22
Update style on core workflow image and light/dark theme

joshuaogle:
Updated some docs images to new style and to add light/dark theming.
- `boundary-network.png` image was originally made to replace the `production.png` but it looks like we missed one. It isn't in the new style, but looks a lot better.
- `access-model` image was the same as `how-boundary-works`, so I replaced that too


Dan-Heath (approved):
These look great, thank you so much for the update @joshuaogle ! I apologize that this has taken so long to get merged, we created some conflicts that had to be resolved first.

Thank you for participating in Docs Days!! These new graphics are a huge improvement ❤️

PR #4572, opened 2024-03-25
Add hidden ferry commands and auto adding tokens to ferry daemon

talanknight:
The background actions that run with every command is silent on error, and only tries the ferry daemon once before stopping.


johanbrandhorst (approved):
:shipit:

PR #4575, opened 2024-03-26
Change 404 errors for aliases to 401s

talanknight:


internal/daemon/controller/interceptor_test.go
```diff
@@ -724,13 +724,13 @@ func Test_aliasResolutionInterceptor(t *testing.T) {
 			name:            "aliasable request with unknown alias",
 			req:             &pbs.GetTargetRequest{Id: "not.a.registered.alias"},
 			wantModifiedReq: &pbs.GetTargetRequest{Id: "not.a.registered.alias"},
-			errorContains:   "resource alias not found with value",
+			errorContains:   "Forbidden",
```
johanbrandhorst, line 727: Is it worth checking that the error is _exactly_ `"Forbidden"` in this and the other case? It doesn't help if we return different error strings because of wrapping or something.

internal/daemon/controller/interceptor.go
```diff
@@ -355,7 +355,14 @@ func aliasResolutionInterceptor(
 		}
 		interceptorCtx, err = alias.ResolveRequestIds(interceptorCtx, reqMsg, r)
 		if err != nil {
-			return nil, err
+			// Since this is intercepted prior to checking that the requester
+			// is authorized to make the request, returning a 404 here and a 403
```
johanbrandhorst, line 359:
```suggestion
			// is authorized to make the request, returning a 404 here and a 401
```

internal/daemon/controller/interceptor.go
```diff
@@ -355,7 +355,14 @@ func aliasResolutionInterceptor(
...
+			// when the request is checked in the handler would expose which
+			// aliases exist. Instead, we return a forbidden error here so there
```
johanbrandhorst, line 361:
```suggestion
			// aliases exist. Instead, we return a unauthenticated error here so there
```

internal/tests/api/targets/target_test.go
```diff
@@ -748,7 +759,18 @@ func TestSet_Errors(t *testing.T) {
...
+	noAliasApiErrReponse := apiErr.Response()
+
+	client.SetToken("at_1234567890_madeupinvalidtoken")
+	tarClient = targets.NewClient(client)
+	_, err = tarClient.Read(tc.Context(), validTargetId)
+	apiErr = api.AsServerError(err)
+	assert.NotNil(apiErr)
+	assert.EqualValues(http.StatusUnauthorized, apiErr.Response().StatusCode())
+	// We should not be able to tell the difference between an unauthorized request
+	// and an request that tries to resolve an alias that points to nothing.
```
johanbrandhorst, line 772:
```suggestion
	// and a request that tries to resolve an alias that points to nothing.
```

internal/daemon/controller/interceptor.go
```diff
@@ -355,7 +355,14 @@ func aliasResolutionInterceptor(
...
+			// is no way to distinguish between a missing alias and an existing
+			// alias that points to a destination the requester is not allowed
+			// to perform an action on.
+			return nil, handlers.UnauthenticatedError()
```
johanbrandhorst, line 365: May be worth adding a little note here to say that `handlers.UnauthenticatedError()` turns in to HTTP Status 401 _Unauthorized_ (thanks HTTP), since the inconsistency of using `Unauthenticated` and `Unauthorized` is easy to be confused by.

internal/tests/api/targets/target_test.go
```diff
@@ -748,7 +759,18 @@ func TestSet_Errors(t *testing.T) {
...
+	assert.EqualValues(noAliasApiErrReponse.Body, apiErr.Response().Body)
```
johanbrandhorst, line 773: 👌🏻

johanbrandhorst (approved):

PR #4578, opened 2024-03-27
test(e2e): Add module for minio docker container

moduli:
This PR adds an enos module to create a minio docker container. There will be a follow-up PR in boundary-enterprise to add tests using this module.


enos/modules/docker_minio/main.tf
```diff
@@ -0,0 +1,160 @@
...
+}
+variable "user_access_key_id" {
+  description = "Access Key Id for user that can access bucket"
+  type        = string
+  default     = "useraccesskeyid"
+}
+variable "user_secret_access_key" {
+  description = "Secret Access Key for user that can access bucket"
+  type        = string
+  default     = "secretaccesskey"
```
stasryzhov, line 74: Is this no-separators-naming style some kind of minio convention?

stasryzhov (approved):

enos/modules/docker_minio/main.tf
moduli: What are you referring to here? The default values? It can be set to anything (as long it as it meets some requirements). If you were to minio generate these values automatically, it would look something like
```
Access Key: 2B7G5EMQAANJX6YQ0H0I
Secret Key: kgd9tH29fx+ymWWDBAZss5KcMxqAa3iZXEw7lxWi
```

stasryzhov: The default is `secretaccesskey`. I was wandering why it's not `secret-access-key` for example.

moduli: It can be set to anything as long as it's greater than 8 characters.

PR #4586, opened 2024-04-02
fix(session): return from workerStatusReport on first error

irenarindos:
Return at first error in WorkerStatusReport.

### DIFF
--- a/internal/session/service_worker_status_report.go
+++ b/internal/session/service_worker_status_report.go
@@ -5,7 +5,6 @@ package session
 
 import (
 	"context"
-	stderrors "errors"
 	"fmt"
 
 	"github.com/hashicorp/boundary/internal/errors"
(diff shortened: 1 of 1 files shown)


johanbrandhorst (approved):

internal/session/service_worker_status_report.go
```diff
@@ -44,27 +43,23 @@ func WorkerStatusReport(ctx context.Context, repo *Repository, connRepo *Connect
 		}
 	}
 
-	var merr error
 	err := connRepo.updateBytesUpBytesDown(ctx, reportedConnections...)
 	if err != nil {
-		merr = stderrors.Join(merr, errors.New(ctx, errors.Internal, op, fmt.Sprintf("failed to update bytes up and down for worker reported connections: %v", err)))
+		return nil, errors.New(ctx, errors.Internal, op, fmt.Sprintf("failed to update bytes up and down for worker reported connections: %v", err))
```
tmessi, line 48: Thinking about this again, I think there was a reason we continued instead of returning early, so that we don't leave things like orphaned connections around if there was some issue updating bytes up/down. I think that is actually still a good idea, but the exception would be if the context has been canceled, in which case we should return early.

PR #4587, opened 2024-04-02
test(credential): Potential fix for race condition

moduli:
This PR makes some fixes to `TestStoreService_List` to hopefully resolve some race conditions in these tests.

https://hashicorp.atlassian.net/browse/ICU-13237


johanbrandhorst (approved):
Ha, looks like Jeff missed these. Did you check the rest of the tests with the same pattern? There are probably another 15 or so files like this.

moduli:
I only addressed these as I saw some test failures in this package. I'll take a quick peek at the others.

PR #4592, opened 2024-04-03
fix(session): Use active transaction

tmessi:
When closing orphaned connections and checking if a session is not
active, the repository was opening a transaction, but then calling the
original reader instead of using the reader/writer from the transaction.
This would result in a transaction being opend, then a separate
transaction being used to run the query, and finally the original
transaction would commit without any statements being executed.

Co-authored-by: Sorawis Nilparuk (Bo) <sorawis.nilparuk@hashicorp.com>

### DIFF
--- a/internal/session/repository_connection.go
+++ b/internal/session/repository_connection.go
@@ -416,8 +416,8 @@ func (r *ConnectionRepository) closeOrphanedConnections(ctx context.Context, wor
 		ctx,
 		db.StdRetryCnt,
 		db.ExpBackoff{},
-		func(reader db.Reader, w db.Writer) error {
-			rows, err := r.reader.Query(ctx, fmt.Sprintf(orphanedConnectionsCte, notInClause), args)
+		func(_ db.Reader, w db.Writer) error {
+			rows, err := w.Query(ctx, fmt.Sprintf(orphanedConnectionsCte, notInClause), args)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


internal/session/repository_connection.go
```diff
@@ -416,8 +416,8 @@ func (r *ConnectionRepository) closeOrphanedConnections(ctx context.Context, wor
 		ctx,
 		db.StdRetryCnt,
 		db.ExpBackoff{},
-		func(reader db.Reader, w db.Writer) error {
-			rows, err := r.reader.Query(ctx, fmt.Sprintf(orphanedConnectionsCte, notInClause), args)
```
johanbrandhorst, line 420: This seems like something we could write a [ruleguard](https://github.com/quasilyte/go-ruleguard) linter for 🤔. Great find!

johanbrandhorst (approved):

internal/session/repository_connection.go
tmessi: A check for unused variables would also catch it. I know some linters provide that, but I don't think we have it enabled.

bosorawis:
FYI same thing here https://github.com/hashicorp/boundary-enterprise/blob/35cef97bd44b26690b3d6b19170c0c5bd2e80876/internal/session/repository_session.go#L835

johanbrandhorst (approved):

ddebko (approved):

PR #4595, opened 2024-04-04
chore: Address CVE-2023-45288

moduli:
This PR updates `golang.org/x/net` to address [GO-2024-2687](https://pkg.go.dev/vuln/GO-2024-2687)/[CVE-2023-45288](https://www.cve.org/CVERecord?id=CVE-2023-45288)


moduli:
I *think* this updates everything needed? Not confident

johanbrandhorst (approved):
Thanks! It's not required, but we could also update the `go` version in https://github.com/hashicorp/boundary/blob/main/go.mod to be consistent with `.go-version`.

louisruch (approved):
LGTM, thanks for getting to this!

PR #4603, opened 2024-04-04
make newSendCtx also handle context.DeadlineExceeded

bosorawis:
When running into `context deadline exceeded`, Boundary attempts to use the same context to publish events which may lead to further failure. Adding a check to also create a new `context` when `ctx.Err() == context.DeadlineExceeded`

### DIFF
--- a/internal/event/context_unexported_test.go
+++ b/internal/event/context_unexported_test.go
@@ -6,6 +6,7 @@ package event
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


internal/event/context_unexported_test.go
```diff
@@ -44,6 +45,28 @@ func Test_newSendCtx(t *testing.T) {
 			}(),
 			wantCancel: true,
 		},
+		{
```
jimlambrt, line 48: should we add an assertion after line #90 of:
`assert.True(ctx.Err() != context.DeadlineExceeded)`

jimlambrt (changes requested):

internal/event/context_unexported_test.go
bosorawis: oooh, good catch! 👍

fixed!

PR #4617, opened 2024-04-09
refact(session connection): remove session connection state table

irenarindos:
WIP

This PR removes the `session_connection_state` table, replacing it with the new field `connected_time_range` on `session_connection`


internal/db/schema/migrations/oss/postgres/87/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,220 @@
...
+  drop function wh_insert_session_connection();
+
+  drop table session_connection_state;
+  drop table session_connection_state_enm;
+
+  --  If the connected_time_range is null, it means the connection is authorized but not connected.
+  --  If the upper value of connected_time_range is > now() (upper range is infinity) then the state is connected.
+  --  If the upper value of connected_time_range is <= now() then the connection is closed.
+  alter table session_connection
+    add column connected_time_range tstzrange;
```
mgaffney, line 34: Why not add the column and migrate the data from `session_connection_state`?

irenarindos: Do you think it's worth migrating? I'm assuming that the control plane would go down for maintenance, and if we regularly remove terminated sessions from the db it seems like we should be fine to drop that table because sessions should be in a terminated state.

tmessi: I think we should migrate the data. We only delete the sessions after they have been terminated for an hour. So once the controllers are back up there will be an hour before the previous sessions are deleted, and I think we should report an accurate connection state for that time period.

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,231 @@
...
+      perform  from
+        session
+      where
+        public_id = new.public_id and
+        public_id not in (
+            select session_id
+              from session_connection
+            where
+              -- open connections will have a connected_time_range with an upper bound of infinity
+              upper(connected_time_range) > now()
```
tmessi, line 105: We should check if this is equal to `'infinity'::timestamptz`.

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,231 @@
...
+        );
+      if not found then
+        raise 'session %s has open connections', new.public_id;
+      end if;
```
tmessi, line 109: I think this can be done without the subquery and without using `not in`:

```sql
perform
   from session_connection
  where session_id                  = new.public_id
    and upper(connected_time_range) = 'infinity'::timestamptz;
if found then
  raise 'session %s has open connections', new.public_id;
end if;
```

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,231 @@
...
+       new.client_tcp_address,
+       new.client_tcp_port,
+       new.endpoint_tcp_address,
+       new.endpoint_tcp_port,
+       new.bytes_up,
+       new.bytes_down
+    from authorized_timestamp,
+       session_dimension
+         returning * into strict new_row;
+    return null;
```
tmessi, line 180: Here, and elsewhere can you use [river formatting](https://github.com/hashicorp/boundary/blob/16d23fa55f6915b388f23c2dbc5a673249ba8823/sql-style-guide.md?plain=1#L130-L216), i.e.:

```suggestion
    with
      authorized_timestamp (date_dim_key, time_dim_key, ts) as (
        select wh_date_key(create_time), wh_time_key(create_time), create_time
          from session_connection
         where public_id = new.public_id
           and connected_time_range is null
    ),
      session_dimension (host_dim_key, user_dim_key, credential_group_dim_key) as (
        select host_key, user_key, credential_group_key
          from wh_session_accumulating_fact
         where session_id = new.session_id
    )
    insert into wh_session_connection_accumulating_fact (
                connection_id,
                session_id,
                host_key,
                user_key,
                credential_group_key,
                connection_authorized_date_key,
                connection_authorized_time_key,
                connection_authorized_time,
                client_tcp_address,
                client_tcp_port_number,
                endpoint_tcp_address,
                endpoint_tcp_port_number,
                bytes_up,
                bytes_down
    )
    select new.public_id,
           new.session_id,
           session_dimension.host_dim_key,
           session_dimension.user_dim_key,
           session_dimension.credential_group_dim_key,
           authorized_timestamp.date_dim_key,
           authorized_timestamp.time_dim_key,
           authorized_timestamp.ts,
           new.client_tcp_address,
           new.client_tcp_port,
           new.endpoint_tcp_address,
           new.endpoint_tcp_port,
           new.bytes_up,
           new.bytes_down
      from authorized_timestamp,
           session_dimension
 returning * into strict new_row;
 
    return null;
```

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,231 @@
...
+    begin
+      if new.connected_time_range is null then
+        -- Indicates authorized connection. The update statement in this
+        -- trigger will fail for the authorized state because the row for the
+        -- session connection has not yet been inserted into the
+        -- wh_session_connection_accumulating_fact table.
+        return null;
+      end if;
+
+      if upper(new.connected_time_range) > now() then
```
tmessi, line 205: I think this should check if `= 'infinity'::timestamptz`

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,231 @@
...
+      time_col = 'connection_' || state || '_time_key';
+      ts_col   = 'connection_' || state || '_time';
+
+      q = format('update wh_session_connection_accumulating_fact
+                   set (%I, %I, %I) = (select wh_date_key(%L), wh_time_key(%L), %L::timestamptz)
+                   where connection_id = %L
+                   returning *',
+                   date_col,       time_col,       ts_col,
+                   new.update_time, new.update_time, new.update_time,
+                   new.public_id);
```
tmessi, line 221: Can we just do these updates within the two separate cases of the `if` above and avoid the use of `format` and `execute`?

internal/db/sqltest/tests/session/session_connection_state_transition.sql
```diff
@@ -0,0 +1,35 @@
...
+    -- Ensure session connection table is populated
+    select is(count(*), 2::bigint) from session_connection;
+
+    -- Check that both session connections are in the authorized state (null connected_time_range)
+    select is(count(*), 2::bigint) from session_connection where connected_time_range is null;
+
+    -- Connect one of the session connections
+    update session_connection
+        set connected_time_range=tstzrange(now(),'infinity')
+        where public_id = 's1c1___clare';
```
tmessi, line 16: Likewise, can you format blocks like these to form a river?:


```suggestion
    update session_connection
       set connected_time_range = tstzrange(now(), 'infinity')
     where public_id = 's1c1___clare';
```

tmessi (changes requested):
I still need to review the go code, but I have a few comments/requests on the sql.

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
irenarindos: sure, thanks!

internal/db/schema/migrations/oss/postgres/87/01_remove_session_connection_state.up.sql
irenarindos: I've added a migration; unfortunately I can't add a migration test because of the code changes, but I've manually tested it.

tmessi (changes requested):
I'm still reviewing the go changes, but I have some comments on the sql if you want to take a look at those first.

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,228 @@
...
+  -- "Connected" is handled by the function ConnectConnection, which sets the connected_time_range lower bound to now() and upper bound to infinity
+  -- "Closed" is handled by the trigger function, update_connected_time_range_on_closed_reason, which sets the connected_time_range upper bound to now()
+  -- State transitions are guarded by the trigger function, check_connection_state_transition, which ensures that the state transitions are valid
+  create function check_connection_state_transition() returns trigger
+  as $$
+    begin
+      -- Authorized state
+      if new.connected_time_range is null then
+        return new;
+      end if;
```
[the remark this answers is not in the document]
irenarindos: thanks!

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,228 @@
...
+      -- If the old state was authorized, any transition is valid
+      if old.connected_time_range is null then
+        return new;
+      end if;
```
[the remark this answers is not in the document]
irenarindos: A lower bound of null is valid if the connection went from authorized to close and skipped over connected but I'll check for -infinity, thanks!

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,228 @@
...
+            and (connected_time_range is null
+             or connected_time_range = tstzrange(lower(connected_time_range), 'infinity'::timestamptz)
+            );
+      end if;
+    return new;
+    end;
+  $$ language plpgsql;
+
+  create trigger update_connected_time_range_closed_reason after update of closed_reason on session_connection
+    for each row execute procedure update_connected_time_range_on_closed_reason();
```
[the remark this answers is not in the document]
irenarindos: Thanks!

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,228 @@
...
+
+  create trigger check_connection_state_transition before update of connected_time_range on session_connection
+    for each row execute procedure check_connection_state_transition();
+
+  create function update_connected_time_range_on_closed_reason() returns trigger
+  as $$
+    begin
+      if new.closed_reason is not null then
+          update session_connection
+             set connected_time_range = tstzrange(lower(connected_time_range), now())
```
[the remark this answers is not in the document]
irenarindos: thanks!

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,223 @@
...
+        return new;
+      end if;
+      -- Prevent transitions from connected to connected
+      if upper(old.connected_time_range) = 'infinity' and upper(new.connected_time_range) = 'infinity' then
+        raise exception 'Invalid state transition from connected to connected';
+      end if;
+      -- Prevent transitions from closed to connected
+      if lower(new.connected_time_range) >= upper(old.connected_time_range) then
+        raise exception 'Invalid state transition from closed to connected';
+      end if;
```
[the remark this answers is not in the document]
irenarindos: Sure, I've added this change

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,224 @@
...
+      -- new one.
+      perform
+         from session_state ss
+        where ss.session_id = new.public_id and
+                   ss.state = 'terminated';
+      if found then
+        return new;
+      end if;
+    insert into session_state (session_id,    state)
+                       values (new.public_id, 'terminated');
```
tmessi, line 111: nit:
```suggestion
      insert into session_state (session_id,    state)
           values               (new.public_id, 'terminated');
```

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,224 @@
...
+
+  create function wh_insert_session_connection_state() returns trigger
+  as $$
+    declare
+         state text;
+      date_col text;
+      time_col text;
+        ts_col text;
+             q text;
+      connection_row wh_session_connection_accumulating_fact%rowtype;
```
tmessi, line 185: nit:
```suggestion
               state text;
            date_col text;
            time_col text;
              ts_col text;
                   q text;
      connection_row wh_session_connection_accumulating_fact%rowtype;
```

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,224 @@
...
+      if upper(new.connected_time_range) = 'infinity'::timestamptz then
+            update wh_session_connection_accumulating_fact
+               set (connection_connected_date_key,
+                    connection_connected_time_key,
+                    connection_connected_time) = (
+             select wh_date_key(new.update_time),
+                    wh_time_key(new.update_time),
+                    new.update_time::timestamptz)
+              where connection_id = new.public_id;
+        --  returning *;
```
tmessi, line 204: Can this be removed?

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,224 @@
...
+      else
+          update wh_session_connection_accumulating_fact
+                set (connection_closed_date_key,
+                     connection_closed_time_key,
+                     connection_closed_time) = (
+              select wh_date_key(new.update_time),
+                     wh_time_key(new.update_time),
+                     new.update_time::timestamptz)
+               where connection_id = new.public_id;
```
tmessi, line 213: Same thing here:

```suggestion
          update wh_session_connection_accumulating_fact
             set (connection_closed_date_key,
                  connection_closed_time_key,
                  connection_closed_time) = (select wh_date_key(new.update_time),
                                                    wh_time_key(new.update_time),
                                                    new.update_time::timestamptz)
           where connection_id = new.public_id;
```

internal/session/query.go
```diff
@@ -337,21 +340,32 @@ where
...
+	// convert connected_time_range range into state string
+	fetchConnectionStatusCte = `
+select
+  case
+    when connected_time_range is null then 'authorized'
+	when upper(connected_time_range) > now() then 'connected'
+    else 'closed'
+	end status
+from session_connection
+where public_id = ?;
```
tmessi, line 352:
```suggestion
select case
         when connected_time_range is null        then 'authorized'
         when upper(connected_time_range) > now() then 'connected'
         else 'closed'
       end as status
  from session_connection
 where public_id = ?;
```

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,224 @@
...
+          -- returning *;
```
tmessi, line 214: can this be removed?

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,224 @@
...
+  $$ language plpgsql;
+
+  create trigger check_connection_state_transition before update of connected_time_range on session_connection
+    for each row execute procedure check_connection_state_transition();
+
+  create function update_connected_time_range_on_closed_reason() returns trigger
+  as $$
+    begin
+      if new.closed_reason is not null then
+          if old.connected_time_range is null or old.connected_time_range = tstzrange(lower(old.connected_time_range), 'infinity'::timestamptz) then
```
tmessi, line 79: This conditional seems a little odd. In `check_connection_state_transition` we just compare the `upper` value, so I think the same check could be done with:

```suggestion
          if old.connected_time_range is null or upper(old.connected_time_range) = 'infinity'::timestamptz then
```

I think that is a little easier to read.

internal/session/query.go
```diff
@@ -124,6 +124,15 @@ from
 	session_connection_limit, session_connection_count;
 `
 
+	// connectConnection sets the connected time range to (now, infinity) to
+	// indicate the connection is connected.
+	connectConnection = `
+    update session_connection 
+        set connected_time_range=tstzrange(now(),'infinity') 
+    where 
+        public_id=@public_id
```
tmessi, line 133: At least for these new queries, can we format them as per the style guide, ie:

```suggestion
    update session_connection 
       set connected_time_range = tstzrange(now(), 'infinity') 
     where public_id=@public_id
```

tmessi (changes requested):
Couple minor things, mostly style related for the sql.

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
irenarindos: Is it possible we could have an autoformatter for sql? I'm going to admit sometimes I think I'm doing river and then I'm not sure why something isn't, and it can get frustrating.

tmessi: > Is it possible we could have an autoformatter for sql

We don't have anything currently :(.

> sometimes I think I'm doing river and then I'm not sure why something isn't

In this case, you want to separate the sub-query into its own river so it is more obvious where the sub-query begins and ends. See [this example](https://github.com/hashicorp/boundary/blob/ad4957a3c7d459be1fbbd8b9d24ff9f6b666ede1/sql-style-guide.md?plain=1#L173-L216).

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,222 @@
...
+  alter table session_connection
+    add column connected_time_range tstzrange;
+
+  -- Migrate existing data from session_connection_state to session_connection
+   update session_connection
+      set connected_time_range =
+  (select tstzrange(min(start_time), max(start_time))
+     from session_connection_state
+    where session_connection_state.connection_id = session_connection.public_id
+ group by connection_id );
```
tmessi, line 39: This is another sub-query that I missed for formating:

```suggestion
  update session_connection
     set connected_time_range = (select tstzrange(min(start_time), max(start_time))
                                   from session_connection_state
                                  where session_connection_state.connection_id = session_connection.public_id
                               group by connection_id);
```

tmessi (changes requested):
Alright, I looked through the rest, so hopefully this is the last of my feedback.

internal/session/query.go
```diff
@@ -197,17 +205,11 @@ from
...
-            session_connection_state
-          where 
-            state != 'closed' and
-            end_time is null
-        )
-    )
+          session_connection
+        where 
+          upper(connected_time_range) > now() or 
+          connected_time_range is null
```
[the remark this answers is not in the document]
irenarindos: Do we have any existing perf tests we can add to? Or any suggestions on how to start?

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,221 @@
...
+                   q text;
+      connection_row wh_session_connection_accumulating_fact%rowtype;
+    begin
+      if new.connected_time_range is null then
+        -- Indicates authorized connection. The update statement in this
+        -- trigger will fail for the authorized state because the row for the
+        -- session connection has not yet been inserted into the
+        -- wh_session_connection_accumulating_fact table.
+        return null;
+      end if;
```
tmessi, line 194: This new condition is why the [wh/session_connection/update](https://github.com/hashicorp/boundary/blob/irindos-session-connection-states/internal/db/sqltest/tests/wh/session_connection/update.sql) sqltest is failing. I think that test just needs to be updated so it is setting the new `connection_time_range` column when it updates the session_connection.

internal/session/repository_connection.go
```diff
@@ -166,31 +165,29 @@ func (r *ConnectionRepository) AuthorizeConnection(ctx context.Context, sessionI
...
-			}
+
 			return nil
 		},
 	)
 	if err != nil {
-		return nil, nil, errors.Wrap(ctx, err, op)
+		return nil, errors.Wrap(ctx, err, op)
 	}
+	connection.ConnectionStatus = StatusAuthorized.String()
```
tmessi, line 175: I don't think we should just set this status directly like this. We should retrieve it from the database in the same way we do in LookupConnection.

Also see my other comment about using a view, which would allow the `LookupById` call above to retrieve the status.

internal/session/repository_connection.go
```diff
@@ -442,14 +436,22 @@ func (r *ConnectionRepository) closeOrphanedConnections(ctx context.Context, wor
...
 }
 
-func fetchConnectionStates(ctx context.Context, r db.Reader, connectionId string, opt ...db.Option) ([]*ConnectionState, error) {
-	const op = "session.fetchConnectionStates"
-	var states []*ConnectionState
-	if err := r.SearchWhere(ctx, &states, "connection_id = ?", []any{connectionId}, opt...); err != nil {
-		return nil, errors.Wrap(ctx, err, op)
+func fetchConnectionStatus(ctx context.Context, r db.Reader, connectionId string, opt ...db.Option) (ConnectionStatus, error) {
+	const op = "session.fetchConnectionStatus"
+	rows, err := r.Query(ctx, fetchConnectionStatusCte, []interface{}{connectionId}, opt...)
```
tmessi, line 441: It might be a little cleaner to make a view for retrieving a connection with the status. For example a view like:

```sql
create view session_connection_with_status_view as
     select public_id,
            session_id,
            client_tcp_address,
            client_tcp_port,
            endpoint_tcp_address,
            endpoint_tcp_port,
            bytes_up,
            bytes_down,
            closed_reason,
            version,
            create_time,
            update_time,
            user_client_ip,
            worker_id,
            case
               when connected_time_range is null        then 'authorized'
               when upper(connected_time_range) > now() then 'connected'
               else                                          'closed'
            end as status
       from session_connection;
```

Then we can eliminate this `fetchConnectionStates` function and `fetchConnectionStausCte` (which is also a misnomer since it is not a CTE). Instead we can just use `LookupById` and the status would appear just like any other column in the sql result.

internal/session/connection.go
```diff
@@ -44,6 +44,8 @@ type Connection struct {
 	UpdateTime *timestamp.Timestamp `json:"update_time,omitempty" gorm:"default:current_timestamp"`
 	// Version of the connection
 	Version uint32 `json:"version,omitempty" gorm:"default:null"`
+	// ConnectionStatus is a field derived from connected_time_range
+	ConnectionStatus string `json:"connection_state,omitempty" gorm:"-"`
```
tmessi, line 48: Should the json be `connection_state` when the field is `ConnectionStatus`?

Also, can this just be `Status` since the type already is `Connection`?

tmessi (changes requested):

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
```diff
@@ -0,0 +1,221 @@
+-- Copyright (c) HashiCorp, Inc.
```
louisruch, line 1: Just a heads up that the migration number will need to be bumped

irenarindos: Thanks!

internal/db/schema/migrations/oss/postgres/88/01_remove_session_connection_state.up.sql
irenarindos: ty!

internal/session/connection.go
irenarindos: Sure!

internal/session/repository_connection.go
```diff
@@ -266,31 +255,31 @@ func (r *ConnectionRepository) ConnectConnection(ctx context.Context, c ConnectW
...
 			}
 			return nil
 		},
 	)
 	if err != nil {
-		return nil, nil, errors.Wrap(ctx, err, op)
+		return nil, errors.Wrap(ctx, err, op)
 	}
-	return &connection, connectionStates, nil
+	connection.Status = StatusConnected.String()
```
tmessi, line 274: Can we read the status back from the database instead of just assuming this is correct if there were no errors?

irenarindos: thanks!

tmessi (approved):

PR #4637, opened 2024-04-11
docs: Revise HTTP header documentation

Dan-Heath:
This PR adds the `x-correlation-id` HTTP header to the API Overview docs.

View the update in the preview deployment:

https://boundary-gwxwnbon6-hashicorp.vercel.app/boundary/docs/api-clients/api#http-headers

### DIFF
--- a/website/content/docs/api-clients/api/index.mdx
+++ b/website/content/docs/api-clients/api/index.mdx
@@ -89,7 +89,10 @@ Responses may contain the following headers:
 
 - `RateLimit` - Provides the current limit, number of remaining requests, and the time at which the quota will reset for the limit that is closest to being exhausted for the requested resource and action.
 - `RateLimit-Policy` - Describes the limits for the requested resource and action.
-- `x-correlation-id` - Indicates that an API request that came in to Boundary is also related to an audit event in Vault.
+- `X-Correlation-ID` - Identifies a transaction over a series of requests and responses.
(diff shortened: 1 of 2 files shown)


website/content/docs/api-clients/api/index.mdx
```diff
@@ -82,6 +82,15 @@ The following method conventions are used within Boundary's API:
...
 `DELETE` is used for deleting a specific resource, and is only used against a particular resource path.
 
+## HTTP headers
+
+Clients that make requests to the controller API can inspect HTTP response headers to understand the configured limits and current usage.
+Responses may contain the following headers:
+
+- `RateLimit` - Provides the current limit, number of remaining requests, and the time at which the quota will reset for the limit that is closest to being exhausted for the requested resource and action.
+- `RateLimit-Policy` - Describes the limits for the requested resource and action.
+- `x-correlation-id` - Indicates that an API request that came in to Boundary is also related to an audit event in Vault.
```
louisruch, line 92: Might want to clarify what this is a bit, I am not sure on the best wording but:

If the user provides a `x-correlation-id` header in the HTTP request, Boundary will log that value in all audit events related to the request. It will also use that value as the `x-correlation-id` header when making requests to Vault, which will allow users to correlated Boundary and Vault Audit logs.

One thing to note here, this will eventually be for ALL external systems not just Vault. It just so happens that our first iteration of this only really targets Vault.

louisruch (approved):
This is great thanks!

PR #4645, opened 2024-04-15
bug: fix data race in hclogger

irenarindos:


johanbrandhorst (approved):
Mind adding a comment to the `l` field in the struct to explain that it's protecting the `signer` field? It wasn't clear to me until I'd read this whole file why we needed the lock.

PR #4652, opened 2024-04-16
Update worker-tags.mdx to remove Enterprise tag on Vault worker filter

vanphan24:
Vault worker filters are not an Enterprise feature, so we're removing it.


stellarsquall (approved):
thank you!

Dan-Heath (approved):
Great catch, thank you @vanphan24 !

PR #4653, opened 2024-04-17
Add ListResolvableAliases SDK changes

talanknight:
These changes include changing Response to uppercase so the alias list response can be parsed from the response body from a different package (the aliases.AliasListResult populated from the users package).

### DIFF
--- a/api/users/custom.go
+++ b/api/users/custom.go
@@ -13,6 +13,10 @@ import (
 	"github.com/hashicorp/boundary/api/aliases"
 )
 
+// ListResolvableAliases builds and sends a request to the API for listing
+// resolvable aliases for the specified user. It retrieves all remaining pages
+// and includes in the result the list token for paginating through future
+// updates. To use the list token use the users.WithListToken option.
 func (c *Client) ListResolvableAliases(ctx context.Context, userId string, opt ...Option) (*aliases.AliasListResult, error) {
 	if userId == "" {
 		return nil, fmt.Errorf("empty userId value passed into ListResolvableAliases request")


internal/api/genapi/templates.go
```diff
@@ -788,7 +786,7 @@ func (n {{ .Name }}ListResult) GetResponseType() string {
 }
 
 func (n {{ .Name }}ListResult) GetResponse() *api.Response {
-	return n.response
+	return n.Response
```
johanbrandhorst, line 789: Since this already returns a pointer to the underlying value... do we actually need this change? Can't we just mutate the returned value through the pointer?

internal/api/genapi/templates.go
```diff
@@ -719,33 +719,31 @@ import (
 
 type {{ .Name }} struct { {{ range .Fields }}
 {{ .Name }}  {{ .FieldType }} `, "`json:\"{{ .ProtoName }}{{ if ( ne ( len ( .JsonTags ) ) 0 ) }},{{ stringsjoin .JsonTags \",\" }}{{ end }}{{ if ( not .AllowEmpty ) }},omitempty{{ end }}\"`", `{{ end }}
-{{ if ( not ( eq ( len ( .CreateResponseTypes ) ) 0 ) )}}
-	response *api.Response
```
johanbrandhorst, line 723: Do any types now have an *api.Response that didn't before?

internal/api/genapi/templates.go
johanbrandhorst: I see, and we can't dereference the pointer to assign because it's nil.

johanbrandhorst (approved):

PR #4668, opened 2024-04-18
test(worker): Address flake in TestMonitorUpstreamConnectionState

moduli:
We have observed `TestMonitorUpstreamConnectionState` failing due to the following
```
--- FAIL: TestMonitorUpstreamConnectionState (0.02s)
  --- FAIL: TestMonitorUpstreamConnectionState/connection_with_multiple_good_addresses (0.00s)

    controller_connection_test.go:105: 
        	Error Trace:	/home/runner/work/boundary/boundary/internal/daemon/worker/controller_connection_test.go:105
        	Error:      	Not equal: 
        	            	expected: 2
        	            	actual  : 1
        	Test:       	TestMonitorUpstreamConnectionState/connection_with_multiple_good_addresses
```
where...
`1` = Connecting
`2` = Ready

This PR modifies the test to remove the extra check at the end of the test. `waitForConnectionStateCondition` will already have observed the expected state.

https://hashicorp.atlassian.net/browse/ICU-13235

### DIFF
--- a/internal/daemon/worker/controller_connection_test.go
+++ b/internal/daemon/worker/controller_connection_test.go
@@ -12,7 +12,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"golang.org/x/net/nettest"
 	"google.golang.org/grpc"
@@ -92,7 +91,6 @@ func TestMonitorUpstreamConnectionState(t *testing.T) {
 
 			require.NoError(t, cc.Close())
 			cancelStateCtx()
-			assert.Equal(t, connectivity.Shutdown, cc.GetState())
 		})
 	}
 }
--- a/internal/daemon/worker/controller_connection_test.go
+++ b/internal/daemon/worker/controller_connection_test.go
@@ -21,38 +21,13 @@ import (
 )
 
 func TestMonitorUpstreamConnectionState(t *testing.T) {
-	ctx := context.Background()
-	stateCtx, cancelStateCtx := context.WithCancel(ctx)
-
-	upstreamConnectionState := new(atomic.Value)
-
 	servers, err := createTestServers(t)
 	require.NoError(t, err)
 
 	scheme := strconv.FormatInt(time.Now().UnixNano(), 36)
 	res := manual.NewBuilderWithScheme(scheme)
-	grpcResolver := &grpcResolverReceiver{res}
-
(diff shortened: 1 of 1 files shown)


internal/daemon/worker/controller_connection_test.go
```diff
@@ -100,9 +98,6 @@ func TestMonitorUpstreamConnectionState(t *testing.T) {
 			case <-time.After(2 * time.Second):
 				t.Error("Time out waiting for condition")
 			}
-
-			got := upstreamConnectionState.Load()
-			assert.Equal(t, tt.expectedState, got)
```
moduli, line 105: Is going from `READY` -> `CONNECTING` possible?

internal/daemon/worker/controller_connection_test.go
```diff
@@ -60,10 +59,9 @@ func TestMonitorUpstreamConnectionState(t *testing.T) {
 	})
 
 	tests := []struct {
-		name             string
-		expectedResponse *opsservices.GetHealthResponse
```
moduli, line 64: Removed this as it looks like it's unused

internal/daemon/worker/controller_connection_test.go
elimt: Going from `READY` -> `CONNECTING` is not possible https://grpc.github.io/grpc/core/md_doc_connectivity-semantics-and-api.html

Going from `READY` -> `TRANSIENT_FAILURE` -> `CONNECTING` is possible but it can happen really quickly

moduli: Interesting... then how is that flake failure occurring?

internal/daemon/worker/controller_connection_test.go
```diff
@@ -101,8 +90,9 @@ func TestMonitorUpstreamConnectionState(t *testing.T) {
 				t.Error("Time out waiting for condition")
 			}
 
-			got := upstreamConnectionState.Load()
-			assert.Equal(t, tt.expectedState, got)
+			require.NoError(t, cc.Close())
+			cancelStateCtx()
+			assert.Equal(t, connectivity.Shutdown, cc.GetState())
```
johanbrandhorst, line 95: Aren't we just testing the gRPC library at this point? Do we need this check?

elimt: Yeah that's true. I added this this when verifying everything shutdown correctly. I can take this check out. We only care about the states we expect.

johanbrandhorst (approved):

PR #4691, opened 2024-04-19
test(e2e): Verify host id and invalid input for aliases

stasryzhov:
This PR updates the e2e-test for aliases. It adds the following verifications:
- When there is multiple hosts are added to the host set used for the target, an alias with a host id points to the specified one


moduli (commented):
What about those other checks? Were those checks already in the bats tests?

stasryzhov:
> What about those other checks? Were those checks already in the bats tests?

It's investigated separately and will be added in a separate PR if needed.

moduli (approved):

PR #4742, opened 2024-04-29
chore: update depedency

ddebko:


moduli (approved):
I would like to see some confirmation from others that this is ok to do.

johanbrandhorst (approved):
I think this should be fine, but we should try to come back to this once there is a stable version of 1.2, or if they backport it to a 1.1.x release. I left a comment: https://github.com/opencontainers/runc/pull/4217#issuecomment-2083155587. Could you please detail in the commit message what this fixes?

PR #4786, opened 2024-05-09
Auto generate the permissions pages.

talanknight:
This goes a good way towards making sure we do not miss any new resources in the permissions table documentation, and will even document all actions that can be performed on them.  The scope and description could be left blank if the page is not updated when new actions or resources are added.

For add-, set-, remove- actions there are additional steps that can be taken to automate the description process. And that can come later.

### DIFF
--- a/internal/website/permstable/permstable.go
+++ b/internal/website/permstable/permstable.go
@@ -339,13 +339,19 @@ func sortedKeys(in map[string]string) []string {
 	return out
 }
 
+// info holds information for a specific resource
 type info struct {
-	scopes             []string
-	actionDescriptions map[action.Type]string
+	// The scopes this resource can be in
+	scopes []string
+	// If the auto generated descriptions do not correctly cover these actions
+	// for this resource, including the action and a description here will
+	// cause this to be used instead of the auto generated one.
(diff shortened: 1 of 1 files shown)


Dan-Heath (approved):
This is great, thank you Todd! The table looks correct to me in the preview deployment. This will be a big help. 🎉

internal/website/permstable/permstable.go
```diff
@@ -230,725 +328,139 @@ func sortedKeys(in map[string]string) []string {
...
-	return ret
+	// TODO: Add something here which follows the template
+	// "Add Xs to an R"
+	// "Set the full set of Xs on an R"
+	// "Remove Xs from an R"
+	switch {
+	case strings.HasPrefix(t.String(), "add-"):
+	case strings.HasPrefix(t.String(), "set-"):
+	case strings.HasPrefix(t.String(), "remove-"):
+	}
```
johanbrandhorst, line 361: 😅 Maybe best to comment this out for now

johanbrandhorst (approved):
Should we make resource table generation run during `make gen`? Thanks for this!

PR #4787, opened 2024-05-09
chore: Update to use Node20 version of Actions

moduli:
This PR updates some GitHub Actions to use a Node20 version of the Action in order to avoid deprecation warnings.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
       - uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4.1.4
       - name: Set Product version
         id: set-product-version
-        uses: hashicorp/actions-set-product-version@v1  # TSCCR: loading action configs: failed to query HEAD reference: failed to get advertised references: authorization failed
+        uses: hashicorp/actions-set-product-version@v2  # TSCCR: loading action configs: failed to query HEAD reference: failed to get advertised references: authorization failed
 
   product-metadata:
     needs: set-product-version
@@ -175,7 +175,7 @@ jobs:
           CGO_ENABLED: "0"
           PRERELEASE_PRODUCT_VERSION: ${{ needs.set-product-version.outputs.prerelease-product-version }}
           METADATA_PRODUCT_VERSION: ${{ needs.product-metadata.outputs.product-edition }}
-        uses: hashicorp/actions-go-build@v0.1.9  # TSCCR: loading action configs: failed to query HEAD reference: failed to get advertised references: authorization failed
+        uses: hashicorp/actions-go-build@v1  # TSCCR: loading action configs: failed to query HEAD reference: failed to get advertised references: authorization failed
         with:
           product_name: ${{ env.PKG_NAME }}
           product_version: ${{ needs.set-product-version. outputs.product-version }}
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -185,10 +185,6 @@ jobs:
           reproducible: report
           instructions: |-
             make build
-      - uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
-        with:
-          name: ${{ env.PKG_NAME }}_${{ needs.set-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip
-          path: out/${{ env.PKG_NAME }}_${{ needs.set-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip
```
moduli, line 191: `upload-artifact-v4` adds a check where you can't upload something if it already exists. It looks like `actions-go-build` already has an upload step, so we're removing the step here.

.github/workflows/milestone-checker.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         with:
           repository: "grafana/grafana-github-actions"
           path: ./actions
-          ref: be89ad434792280ebaa4d982ac72ba548b6f7095
+          ref: 3e07b3b88b3984f28067a76dce5a052f0c5d3c73
```
moduli, line 30: This isn't sufficient. That action still hasn't been updated to use Node20. I've posted an issue in that repo, but not sure if there are alternate paths here.

calcaide:
@moduli just heads up, this work has been started and will be covered [here](https://github.com/hashicorp/boundary-ui/pull/2282) perhaps sync with @cameronperera to not duplicate work?

BTW Thank you to being proactive and help us on fixing deprecation messages!!

moduli:
@calcaide Just to double-check... that seems like a separate effort? I was only looking at node versions used in various github actions, whereas Cameron's work looks to be updating the application itself?

calcaide:
> @calcaide Just to double-check... that seems like a separate effort? I was only looking at node versions used in various github actions, whereas Cameron's work looks to be updating the application itself?

@moduli I deleted my messare right away, this is my bad, for some reason I though this PR was affecting `boundary-ui` but it does to `boundary`. You are totally right, this is a separate effort, sorry for the confusion here!

.github/workflows/milestone-checker.yml
johanbrandhorst: TBH the milestone checker has been broken for a while, (it never updates once a milestone has been set for me), I think we can retire it.

johanbrandhorst (approved):
Thanks for your hard work on this!

.github/workflows/milestone-checker.yml
moduli, line 1: Note to self: If we keep this workflow, we should update the node reference `node-version: '16.x'`

.github/workflows/milestone-checker.yml
moduli: cc: @jefferai @tmessi Any opinions?

moduli: cc: @jefferai @tmessi Any opinions?

tmessi: Can we investigate why it isn't working? Perhaps it needs additional `types` of events so that it triggers?

moduli: Ok. I think I addressed both things
- I borrowed an implementation from vault: https://github.com/hashicorp/vault/blob/main/.github/workflows/milestone-checker.yml
- Triggers have been updated to include `milestoned` and `demilestoned`. We needed to update the version of `actionlint` to pass a linter error -- it would previously complain that those trigger types didn't exist (https://github.com/rhysd/actionlint/releases/tag/v1.6.27)
- Vault's implementation does not require a third party action, so we don't have to worry about the node version they are using anymore

johanbrandhorst (approved):

tmessi (approved):

PR #4791, opened 2024-05-09
Add GPT-4 support

johanbrandhorst:
### [internal/help: add Embedder interface](https://github.com/hashicorp/boundary/commit/7066184f7cfaa8bedeef7c829be5dc133c74d96c)

The Embedder interface helps decouple the searcher from
the specific LLM implementation.

### [internal/help: remove gemini package](https://github.com/hashicorp/boundary/commit/cbfe5181691d7ffd3b865c87c04dbc6bd223133c)

It was a very thin wrapper and we can simplify this by just removing it

### [internal/daemon/controller: add GPT-4 support](https://github.com/hashicorp/boundary/commit/69b5075b299bc65e5b19d600204c0cbd6e002960)


internal/daemon/controller/controller.go
```diff
@@ -489,19 +489,48 @@ func New(ctx context.Context, conf *Config) (*Controller, error) {
...
 			return nil, fmt.Errorf("unable to initialize gemini model: %w", err)
 		}
-		c.llm = googleai
-		c.searcher, err = help.NewSearcher(ctx, eventLogger.Named("help"), dbase, dbase, googleai, c.conf.RawConfig.Controller.Help.NumHintDocuments)
+		c.llm = llm
+		c.searcher, err = help.NewSearcher(ctx, eventLogger.Named("help"), dbase, dbase, llm, c.conf.RawConfig.Controller.Help.NumHintDocuments)
+		if err != nil {
+			return nil, fmt.Errorf("unable to create searcher: %w", err)
+		}
+		if err := c.searcher.CreateEmbeddings(ctx); err != nil {
```
elimt, line 512: If we start creating the embeddings, wouldn't that be a long blocking call because we have a lot of docs?

internal/daemon/controller/controller.go
```diff
@@ -489,19 +489,48 @@ func New(ctx context.Context, conf *Config) (*Controller, error) {
...
+		if c.conf.RawConfig.Controller.Help.ApiKey == "" {
+			return nil, fmt.Errorf(`api key is required with help model "openai"`)
+		}
+		opts := []openai.Option{
+			openai.WithToken(c.conf.RawConfig.Controller.Help.ApiKey),
+		}
+		if c.conf.RawConfig.Controller.Help.Model != "" {
+			opts = append(opts, openai.WithModel(c.conf.RawConfig.Controller.Help.Model))
+		}
+		if c.conf.RawConfig.Controller.Help.EmbeddingModel != "" {
```
elimt, line 525: What is the `EmbedddingModel` for openai? We also don't create embeddings when the llm is "openai"

johanbrandhorst: This relies on using the default embedding model for openai, I'm not sure which one it is, but I don't think it should matter so much.

We do create embeddings further down, it's just not showing up in this diff because the logic is repeated.

internal/daemon/controller/controller.go
johanbrandhorst: Yep, it takes 10-15s or so on startup.

PR #4794, opened 2024-05-09
Wait to select tokens until we know we need to

talanknight:
AuthToken selection involves checking to seeing if the one we are picking has been canceled or expired on the boundary controller. If we are only searching locally and not performing a refresh, there is no need to check if the tokens have expired or not.


johanbrandhorst (approved):
Nice!

PR #4803, opened 2024-05-15
fix(session): Move encryption/decryption out of database transactions

tmessi:
Several session.Repository methods need to encrypt or decrypt fields
when persisting or retrieving a Session. These encrypt/decrypt
operations where being performed while within a database transaction.

The first step in the encrypt/decrypt operation is to retrieve the
kms.Wrapper. While the wrappers get cached in memory, the retrieval
still results in a database query to check that the kms collection
version has not changed. This query happens with a separate db.Reader,
and therefore uses a separate database connection from the transaction
being used by the session.Repository.

If the controllers are configured with a relatively low
`max_open_connections`, and be under sufficient load interactive with
sessions, the controller can consume all of the available database
connections, and get stuck with open transactions while waiting for a
database connection to execute the kms related queries. This can be see
by checking pg_stat_activity and seeing transactions stuck in `idle in
transaction` with a wait_event of ClientRead.

This commit changes all of the Repository methods the perform any
encryption or decryption operations, including the retrieval of the
kms.Wrapper, outside of a database transaction. An alternative approach
would be to unsure that the kms.Wrapper retrieval happens within the
same database transaction. However, it seems preferable to avoid doing
these relatively expensive operations within a transaction when it is
not necessary.

As part of this fix, it also eliminates some logic that is no longer
needed. During previous would with KMS wrapping, the method of how
sessions were encrypted was changed, and some logic to lazily re-encrypt
sessions was introduced. However, given that terminated sessions are
deleted, and the KMS changes was introduced several versions ago
(0.13.0), this lazy re-encrypt logic should no longer be necessary.

Refs: edd323b73a36a61f37f349e214d4d84515e44ac2
See:
    https://www.postgresql.org/docs/current/monitoring-stats.html#MONITORING-PG-STAT-ACTIVITY-VIEW
    https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-IDLE-IN-TRANSACTION-SESSION-TIMEOUT
    https://developer.hashicorp.com/boundary/docs/configuration/controller#max_open_connections

### DIFF
--- a/internal/session/repository_session.go
+++ b/internal/session/repository_session.go
@@ -252,15 +252,6 @@ func (r *Repository) LookupSession(ctx context.Context, sessionId string, opt ..
 				return errors.Wrap(ctx, err, op)
 			}
 			session.Connections = connections
-			if session.ProjectId == "" || session.UserId == "" {
-				// Skip decryption if Project ID or UserId is missing,
-				// since it will just lead to errors, and the session
-				// is already canceled if either of those are empty.
-				return nil
-			}
-			if err := decryptAndMaybeUpdateSession(ctx, r.kms, &session, w); err != nil && !opts.withIgnoreDecryptionFailures {
-				return errors.Wrap(ctx, err, op)
-			}
 			return nil
 		},
 	)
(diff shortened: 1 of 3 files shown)


johanbrandhorst (approved):
Great job getting to the bottom of this 😲! Now we just need to ram it into our collective culture never to use a wrapper in a transaction 🤔. Mind posting a PSA or something in the foundations channel?

PR #4805, opened 2024-05-16
fix(controller): Use max request duration cluster listener

tmessi:
The cluster listener supports a `max_request_duration` option, and
defaults to 90 seconds. However, this configuration was not being use
when processing grpc requests from the listener. This meant that
requests from worker to controller could hang indefinitely and
contributed to the recently fixed issue with database connections being
stuck in `idle in transaction`.

This commit adds a grpc interceptor that ensures each request has a
deadline set to the max request duration for the listener.

Refs: 29cd7bc9d6128502f918b2469c71eb6bc3751d72

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,24 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 
 ## Next
 
+### Bug Fixes
+
+* Fix a dead lock issue where the controller could get stuck with all of its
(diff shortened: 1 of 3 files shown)


johanbrandhorst (approved):
Nice!

ajayreshc (approved):
I think this is fine from an HCP perspective. If these requests are taking 90+ I think there is a broader issue.

PR #4821, opened 2024-05-23
Rab update filtering managed groups

stellarsquall:
[Deploy preview](https://boundary-h7wd9uqqs-hashicorp.vercel.app/boundary/docs/concepts/filtering/managed-groups)

This is a first pass at updating the managed groups usage docs to add LDAP support.

The organization of this content could be challenging, since OIDC filters are applied on the managed group resource, while LDAP filters are applied on the auth method resource. From a domain model perspective, it may be more appropriate to create a new LDAP auth methods filters page than to update the managed groups page with this content. But this is likely to confuse users, who at just trying to set up a managed group.


johanbrandhorst (approved):
This LGTM but I don't really know enough about LDAP to validate the information, will defer to Jim.

website/content/docs/concepts/filtering/managed-groups.mdx
```diff
@@ -0,0 +1,167 @@
...
+
+## LDAP managed groups
+
+Membership is determined by matching the LDAP group names against an LDAP account's associated groups, and every authentication updates the group membership comparison.
+
+To enable managed groups, the LDAP auth method `enable_groups` attribute should be set to `true`. 
+
+After enabling groups on the auth method, the managed group membership is defined as a list with the `group-names` attribute on the managed group itself.
+
+### LDAP auth method filter attributes
```
jimlambrt, line 86:
```suggestion
### LDAP auth method search filter attributes
```

jimlambrt (approved):
Looks great and it's clear to me.

Dan-Heath (approved):
I pushed some minor edits, but otherwise this looks good and makes sense to me. Thank you!!

PR #4822, opened 2024-05-23
worker auth: load correct records and ability to remove unused

irenarindos:
Note: this will fail until the corresponding Nodeenrollment change is released


internal/db/schema/migrations/oss/postgres/88/01_worker_auth_invariant_trigger.up.sql
```diff
@@ -0,0 +1,42 @@
...
+
+begin;
+-- update_worker_auth_authorized is a before update trigger function for the
+-- worker_auth_authorized table. The worker_auth_authorized table is a child
+-- table of server_worker. A row contains a set encryption keys for a
+-- server_worker that are unique to that worker. A server_worker can only have
+-- two rows in the worker_auth_authorized table: one with a state of 'current'
+-- and one with the state of 'previous'.
+--
+--- This trigger function ensures that there is only ever one entry for
```
johanbrandhorst, line 12:
```suggestion
--- This trigger function ensures that there is only ever one entry
```

internal/db/schema/migrations/oss/postgres/88/01_worker_auth_invariant_trigger.up.sql
```diff
@@ -0,0 +1,42 @@
...
+--  with a state of 'current' and one with a state of 'previous'.
+create function update_worker_auth_authorized() returns trigger
+as $$
+begin
+    if new.state = 'current' then
+        perform
+        from worker_auth_authorized
+          where state = 'current';
```
johanbrandhorst, line 20: Isn't this per-worker? Should there be a `where` clause in here somewhere? Does this work if there are multiple workers?

internal/db/sqltest/tests/server/server_worker_worker_auth.sql
```diff
@@ -0,0 +1,58 @@
...
+    select is(count(*), 1::bigint) from worker_auth_authorized where worker_key_identifier = 'key_id_1' and state='current';
+
+    -- Test rotation logic. Insert another worker auth record, expect it to be current.
+    -- The previous record should be marked as previous
+    insert into worker_auth_authorized
+        (worker_key_identifier, worker_id, worker_signing_pub_key, worker_encryption_pub_key, controller_encryption_priv_key, key_id)
+    values
+        ('key_id_2', 'w_1234567891', 'signing_pub_key_2', 'encryption_pub_key_2', 'controller_encryption_priv_key_2', 'kdkv___widget');
+    select is(count(*), 1::bigint) from worker_auth_authorized where worker_key_identifier = 'key_id_1' and state='previous';
+    select is(count(*), 1::bigint) from worker_auth_authorized where worker_key_identifier = 'key_id_2' and state='current';
```
johanbrandhorst, line 30: Could you add a test where there are two workers each with a `current` and `previous` stored?

internal/server/repository_workerauth.go
```diff
@@ -8,6 +8,7 @@ import (
 	"crypto/x509"
 	"database/sql"
 	"fmt"
+	"github.com/hashicorp/go-dbw"
```
johanbrandhorst, line 11: I think the import formatter went rogue here

johanbrandhorst (commented):
This should fix the subsequent node rotation attempt, but will it fix the dialing errors we were seeing too? This relies on us having at least two active controller connections when the split brain error happens, since one will be lost during the EOF error, and the other will be required to try to rotate again. In a perfect world, I think that means the second rotation should go through and the split brain scenario should be resolved, but what if there's only a single controller and this happens? Since the worker can't tell the controller to delete unused credentials until it's able to complete dialing, I think we'll need a fix on the controller side of the dialing too.

internal/db/schema/migrations/oss/postgres/88/01_worker_auth_invariant_trigger.up.sql
irenarindos: my bad, I've been too focused thinking about a single instance 😮‍💨  thanks!

irenarindos:
> This should fix the subsequent node rotation attempt, but will it fix the dialing errors we were seeing too? This relies on us having at least two active controller connections when the split brain error happens, since one will be lost during the EOF error, and the other will be required to try to rotate again. In a perfect world, I think that means the second rotation should go through and the split brain scenario should be resolved, but what if there's only a single controller and this happens? Since the worker can't tell the controller to delete unused credentials until it's able to complete dialing, I think we'll need a fix on the controller side of the dialing too.

It will fix the dialing errors; those were a result of `GenerateServerCertificates` calling `LoadNodeInformation` for a key id and the old logic there. We incorrectly assumed that the key id passed in was `current` and assigned the current auth record creds to the returned node information.

internal/db/schema/migrations/oss/postgres/88/01_worker_auth_invariant_trigger.up.sql
irenarindos: thanks!

PR #4808, opened 2024-05-17
Rename daemon CLI command to cache

johanbrandhorst:
Deprecates the `boundary daemon` command in favor of `boundary cache`. It also introduces a warning to users using the old command:

```shell
$ boundary daemon
The 'daemon' command is deprecated and will be removed in a future release. Please use the 'cache' command instead.
Usage: boundary cache [sub command] [options]

  This command allows interacting with the Boundary cache.

  Start a cache:

      $ boundary cache start

  For a full list of examples, please see the documentation.

Subcommands:
    add-token    Add an auth token to a running boundary cache
    start        Start a Boundary cache
    status       Get the status information of the running boundary cache
    stop         Stop the Boundary cache
```

PR #4833, opened 2024-05-29
test(e2e): Rename daemon command to cache

moduli:
This PR is a follow up to https://github.com/hashicorp/boundary/pull/4808

This PR updates an e2e test that uses the `daemon` command in favor of the new name `cache`.


johanbrandhorst (approved):
🙏🏻 thank you!

PR #4847, opened 2024-05-30
chore: point to new api & sdk tagged versions

elimt:


johanbrandhorst (commented):
What version of `api` was 0.0.51 made with? I know we recently release `0.0.50` from `main`.

elimt:
> What version of `api` was 0.0.51 made with? I know we recently release `0.0.50` from `main`.

This was made with 0.16.1

johanbrandhorst:
> > What version of `api` was 0.0.51 made with? I know we recently release `0.0.50` from `main`.
> 
> @johanbrandhorst This was made with 0.16.1

Can you check the difference between release 0.0.50 and 0.0.51, if any? I hope we didn't accidentally break any users by removing features.

elimt:
> Can you check the difference between release 0.0.50 and 0.0.51, if any? I hope we didn't accidentally break any users by removing features. We should be careful about tagging releases on release branches.

Shouldn't API & SDK releases correspond with the changes in our official releases? `api/v0.0.50` was made by @talanknight but he added the changes made with that to `release/0.16.x`. I did a diff between both and did not see anything in `api/v0.0.51` that removes features: https://github.com/hashicorp/boundary/compare/api/v0.0.50...api/v0.0.51

johanbrandhorst:
> Shouldn't API & SDK releases correspond with the changes in our official releases? `api/v0.0.50` was made by @talanknight but he added the changes made with that to `release/0.16.x`. I did a diff between both and did not see anything in `api/v0.0.51` that removes features: [api/v0.0.50...api/v0.0.51](https://github.com/hashicorp/boundary/compare/api/v0.0.50...api/v0.0.51)

The `api` module is released whenever we need it to be. I think the 0.0.50 tag in this case was made from `main`, but we did backport the changes so indeed there was no harm done here. As long as you checked before making the new tag it's all good, but obviously we definitely don't want to make a 0.0.51 tag with features from 0.0.50 removed (unless we do so intentionally).

johanbrandhorst (approved):

irenarindos:
Closing as old

PR #4850, opened 2024-05-30
Close proxy listener if session is no longer active

talanknight:
This only fixes the api packages. A follow up PR needs to add this to `boundary connect` command, along with some refactoring since that does not seem to return when the proxy exits.

### DIFF
--- a/api/proxy/option.go
+++ b/api/proxy/option.go
@@ -133,6 +133,9 @@ func WithSessionTeardownTimeout(with time.Duration) Option {
 }
 
 // WithApiClient provides an optional Boundary API client
+// Experimental: It is unclear whether the current usage of this option is the
+// approach that we want to take in the long term. This may be removed at any
+// point going forward.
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):
How hard would it be to write a test for this?

api/proxy/option.go
```diff
@@ -129,3 +131,11 @@ func WithSessionTeardownTimeout(with time.Duration) Option {
 		return nil
 	}
 }
+
+// WithApiClient provides an optional sessions client
```
johanbrandhorst, line 135:
```suggestion
// WithApiClient provides an optional Boundary API client
```

johanbrandhorst (changes requested):

johanbrandhorst (approved):

jefferai (approved):
As discussed, this will probably change in the future so please add a comment about that, but otherwise I'm good with this!

PR #4852, opened 2024-05-31
Instead of relying on termination reason use status

talanknight:


api/proxy/proxy.go
```diff
@@ -273,14 +273,14 @@ func (p *ClientProxy) Start(opt ...Option) (retErr error) {
...
 						return
 					}
 
 					// We got a valid session response for the session we just
-					// closed a connection for. Since there is a termination reason
-					// we can treat the session as being terminated so no more
-					// connections will be able to be established.
+					// closed a connection for. Since the status isn't active
+					// we can treat the session as no longer being able to
+					// suport connections so close this proxy.
```
johanbrandhorst, line 283:
```suggestion
					// support connections, so close this proxy.
```

johanbrandhorst (approved):

PR #4855, opened 2024-05-31
test: Fix tests due to time

moduli:
This PR addresses two failing tests.

### Test 1
`TestRepository_MonthlyActiveUsers` was failing for the following reason
```
=== RUN   TestRepository_MonthlyActiveUsers/valid-with-start-time
    repository_test.go:102: 
                Error Trace:    /Users/mycow/Developer/boundary/internal/billing/repository_test.go:102
                Error:          "[{2024-05-01 00:00:00 +0000 UTC 2024-05-31 14:00:00 +0000 UTC 0} {2024-04-01 00:00:00 +0000 UTC 2024-05-01 00:00:00 +0000 UTC 6} {2024-03-01 00:00:00 +0000 UTC 2024-04-01 00:00:00 +0000 UTC 6}]" should have 4 item(s), but has 3
                Test:           TestRepository_MonthlyActiveUsers/valid-with-start-time
=== RUN   TestRepository_MonthlyActiveUsers/valid-with-start-and-end-time
    repository_test.go:131: 
                Error Trace:    /Users/mycow/Developer/boundary/internal/billing/repository_test.go:131
                Error:          Not equal: 
                                expected: time.Date(2024, time.March, 1, 0, 0, 0, 0, time.UTC)
                                actual  : time.Date(2024, time.April, 1, 0, 0, 0, 0, time.UTC)
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,2 +1,2 @@
                                -(time.Time) 2024-03-01 00:00:00 +0000 UTC
                                +(time.Time) 2024-04-01 00:00:00 +0000 UTC
                                 
                Test:           TestRepository_MonthlyActiveUsers/valid-with-start-and-end-time
    repository_test.go:138: 
                Error Trace:    /Users/mycow/Developer/boundary/internal/billing/repository_test.go:138
                Error:          Not equal: 
                                expected: time.Date(2024, time.March, 1, 0, 0, 0, 0, time.UTC)
                                actual  : time.Date(2024, time.April, 1, 0, 0, 0, 0, time.UTC)
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,2 +1,2 @@
                                -(time.Time) 2024-03-01 00:00:00 +0000 UTC
                                +(time.Time) 2024-04-01 00:00:00 +0000 UTC
                                 
                Test:           TestRepository_MonthlyActiveUsers/valid-with-start-and-end-time
```
This was due to an issue with calculating times relative to the current date. In this case, with today being May 31, `threeMonthsAgo` would return `March` instead of the expected `February`. This PR forces those calculations to use the start of the month to avoid differences in month duration.

### Test 2
`monthly_active_users_all_timezone.sql` was failing for the following reason
```
# Failed test 5
#     Results differ beginning at row 1:
#         have: (13)
#         want: (14)
# Failed test 6
#     Results differ beginning at row 2:
#         have: ("2024-04-01 13:00:00+13","2024-05-01 12:00:00+12",6)
#         want: ("2024-04-01 13:00:00+13","2024-05-01 12:00:00+12",0)
# Failed test 8
#     Results differ beginning at row 2:
#         have: ("2024-04-01 13:00:00+13","2024-05-01 12:00:00+12",6)
#         want: ("2024-04-01 13:00:00+13","2024-05-01 12:00:00+12",0)
# Failed test 10
#     Results differ beginning at row 10:
#         have: NULL
#         want: ("2023-04-01 13:00:00+13","2023-05-01 12:00:00+12",6)
```
This was due to an issue with timezones. The test sets up data in New Zealand time (e.g. June 1), but it's still May 31 in UTC time, resulting in some conflicts when querying for expected data. The fix follows a similar implementation as https://github.com/hashicorp/boundary/pull/3273/files to skip this test from running when it meets this condition.

### DIFF
--- a/internal/billing/repository_test.go
+++ b/internal/billing/repository_test.go
@@ -74,8 +74,16 @@ func TestRepository_MonthlyActiveUsers(t *testing.T) {
 	TestGenerateActiveUsers(t, conn)
 
 	today := time.Now().UTC()
-	threeMonthsAgo := time.Date(today.AddDate(0, -3, 0).Year(), today.AddDate(0, -3, 0).Month(), 1, 0, 0, 0, 0, time.UTC)
-	oneMonthAgo := time.Date(today.AddDate(0, -1, 0).Year(), today.AddDate(0, -1, 0).Month(), 1, 0, 0, 0, 0, time.UTC)
+	// Some time calculations are impacted when using the current day vs. the
+	// start of the month. For example, if...
+	// today -> May 30th
+	// today.AddDate(0, -3, 0).Month() -> March
+	// February was expcted here, but we get March. This seems to be a
+	// rounding thing since February 30th is not a valid date. Instead, the
+	// start of the month is used to ensure the correct months are calculated.
+	monthStart := time.Date(today.Year(), today.Month(), 1, 0, 0, 0, 0, time.UTC)
+	threeMonthsAgo := time.Date(monthStart.AddDate(0, -3, 0).Year(), monthStart.AddDate(0, -3, 0).Month(), 1, 0, 0, 0, 0, time.UTC)
+	oneMonthAgo := time.Date(monthStart.AddDate(0, -1, 0).Year(), monthStart.AddDate(0, -1, 0).Month(), 1, 0, 0, 0, 0, time.UTC)
 	midMonth := time.Date(today.Year(), today.Month(), 15, 0, 0, 0, 0, time.UTC)
 
 	t.Run("valid-no-options", func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


internal/billing/repository_test.go
```diff
@@ -74,8 +74,16 @@ func TestRepository_MonthlyActiveUsers(t *testing.T) {
 	TestGenerateActiveUsers(t, conn)
 
 	today := time.Now().UTC()
-	threeMonthsAgo := time.Date(today.AddDate(0, -3, 0).Year(), today.AddDate(0, -3, 0).Month(), 1, 0, 0, 0, 0, time.UTC)
-	oneMonthAgo := time.Date(today.AddDate(0, -1, 0).Year(), today.AddDate(0, -1, 0).Month(), 1, 0, 0, 0, 0, time.UTC)
+	// Some time calculations are impacted when using the current day vs. the
+	// start of the month. For example, if...
+	// today -> May 30th
+	// today.AddDate(0, -3, 0).Month() -> March
+	// February was expcted here, but we get March. This seems to be a
```
johanbrandhorst, line 81:
```suggestion
	// February was expected here, but we get March. This seems to be a
```

internal/billing/repository_test.go
```diff
@@ -74,8 +74,16 @@ func TestRepository_MonthlyActiveUsers(t *testing.T) {
...
+	// rounding thing since February 30th is not a valid date. Instead, the
+	// start of the month is used to ensure the correct months are calculated.
+	monthStart := time.Date(today.Year(), today.Month(), 1, 0, 0, 0, 0, time.UTC)
```
johanbrandhorst, line 84: Not that we risk having any months with fewer than 1 days anytime soon, but why make this relative at all? Why not just
```suggestion
	monthStart := time.Date(2000, time.May, 1, 0, 0, 0, 0, time.UTC)
```
etc?

johanbrandhorst (approved):

internal/billing/repository_test.go
moduli: There would need to be a lot more modifications to the test. `TestGenerateActiveUsers` generates data for the last twelve months.

tmessi (approved):

PR #4859, opened 2024-06-03
Citex migration the safe way

burdandrei:


jimlambrt:
`citext` is an optional extension, so we'd recommend just create the initial boundary database without it.  

As a first principal, we don't use "if exist" clauses unless we're modifying existing bits like indexes, views and constraints.

mgaffney:
I agree with @jimlambrt. We strictly limit the use of `if not exists` clauses because we want a migration to fail if something exists in the database that we didn't put there. There is no way for us to prevent customers from doing things in the database directly but failing because something exists that we don't know about allows a customer to safely fix whatever they did before running the migrations again.

burdandrei:
Thanks for the fast feedback, but.....
I had Boundary installed and ran `database init`
When I tried to start the service it said - DB should be migrated.
I ran `database migrate` and received the error that citext is already installed.

This is what made me open the PR.

shouldn't `init` and `migrate` be idempotent?

burdandrei:
thanks for re-opening @mgaffney,
to be fair I just took a fresh linux box, and `database init` told me it was already ran when I tried to run it for the second time.

also was able to run migrate 2-3 times without any issues.

Still I'd add `if not exists` as a failsafe.

mgaffney:
Thanks @burdandrei for retesting using a clean env. I was also successful using a clean environment. I ran `database init` on a local build from `main` and then started the controller with no issues. 

I'm going to close this PR since we validated that things work as expected.

tgahlx:
Adding `if not exists` would help other users that are not in charge of their database. For example: we are provided a managed service were we don't get any privileges besides writing within our database schema. Extensions can be requested and will be provided to the schema if required. Boundary will not work for us any more unless we maintain our own database which adds additional effort including backup and other operations to the maintaining team.

I also don't understand your reasoning. On the one hand you demand administrative privileges for the database, on the other hand you don't want users to change things. And what's the reason you're including it for the other extensions?

PR #4860, opened 2024-06-03
test(db): Fix skip command

moduli:
This PR is a follow-up to https://github.com/hashicorp/boundary/pull/4855

I observed today that the sql tests are failing for the following reason
```
Test Summary Report
-------------------
/test/hcp/billing/monthly_active_users_all_timezone.sql                               (Wstat: 0 Tests: 11 Failed: 0)
  Parse errors: Bad plan.  You planned 27 tests but ran 11.
Files=147, Tests=2175, 21 wallclock secs ( 0.38 usr  0.18 sys +  0.41 cusr  0.56 csys =  1.53 CPU)
```
https://github.com/hashicorp/boundary/actions/runs/9349282645/job/25730230873?pr=4858

Looking back at the PR, I think I had an error using the `skip` command and that it should be set to `skip('', 1)`. I think this makes sense -- by setting it to `skip('', 3`) in 8 tests, that would have increased the number by 16 (i.e. 8 * 2), changing the original plan from `11` to `27`.

### DIFF
--- a/internal/db/sqltest/tests/hcp/billing/monthly_active_users_all_timezone.sql
+++ b/internal/db/sqltest/tests/hcp/billing/monthly_active_users_all_timezone.sql
@@ -2,7 +2,7 @@
 -- SPDX-License-Identifier: BUSL-1.1
 
 begin;
-  select plan(27);
+  select plan(11);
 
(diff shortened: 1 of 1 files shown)


tmessi (approved):

tmessi:
Good catch

PR #4867, opened 2024-06-03
test(e2e): Add check for ldap account attribute mapping

moduli:
This PR adds a check to an e2e test for LDAP account attribute mapping implemented here: https://github.com/hashicorp/boundary/pull/4788.

https://hashicorp.atlassian.net/browse/ICU-13979


johanbrandhorst (approved):
very cool!

PR #4870, opened 2024-06-03
workerauth: satisfy NodeIdLoader interface

irenarindos:
Note: tests will fail until the corresponding nodeenrollment changes are merged

Changes to the worker auth repository:

- Update LodeNodeInformation to correctly populate previous nodeInformation
- Satisfy the NodeIdLoader interface with `LoadByNodeId`
- Update store so that if worker NodeInformation has a previous key set, ensure that the control plane's view of what is current and previous matches
- Add test for "split brain" scenario

Also added a database trigger to ensure we don't store multiple current or previous records for a worker


internal/db/sqltest/tests/server/server_worker_local_storage_state.sql
johanbrandhorst, line 1: I think moving this file broke the SQL tests. You probably need to remove `worker/*.sql` from the test script.

internal/db/schema/migrations/oss/postgres/88/01_worker_auth_invariant_trigger.up.sql
```diff
@@ -0,0 +1,42 @@
...
+--
+--- This trigger function ensures that there is only ever one entry
+--  with a state of 'current' and one with a state of 'previous'.
+create function update_worker_auth_authorized() returns trigger
+as $$
+begin
+    if new.state = 'current' then
+        perform
+        from worker_auth_authorized
+          where state = 'current' and worker_id = new.worker_id and worker_key_identifier != new.worker_key_identifier;
```
johanbrandhorst, line 20: This statement is a bit hard to read for me, do you mind adding a comment to elaborate on what we're checking? It looks like we can't update a row to 'current' if there is already a row with a state of 'current' with the same worker ID but a different key identifier? Isn't that always going to be true? Judging by the tests the key identifier changes every rotation.

internal/server/repository_workerauth.go
```diff
@@ -250,6 +246,43 @@ func StoreNodeInformationTx(ctx context.Context, reader db.Reader, writer db.Wri
...
+		// If it's previous... delete current, as it's incorrect and set this one to current to
+		// ensure proper state rotation on store
+		if state == previousWorkerAuthState {
+			query := deleteWorkerAuthByKeyId
+			_, err := writer.Exec(ctx, query, []any{sql.Named("worker_key_identifier", node.PreviousEncryptionKey.KeyId)})
+			if err != nil {
+				return errors.Wrap(ctx, err, op)
+			}
+			query = updateWorkerAuthStateByKeyId
+			_, err = writer.Exec(ctx, query, []any{sql.Named("worker_key_identifier", node.PreviousEncryptionKey.KeyId)})
```
johanbrandhorst, line 279: Is `writer` a transaction? I would want to make sure both of these happen in the same transaction.

johanbrandhorst (changes requested):
Really cool, thank you!

internal/server/repository_workerauth.go
irenarindos: It is, this is in `StoreNodeInformationTx` which is only ever called as part of a transaction

internal/db/sqltest/tests/server/server_worker_local_storage_state.sql
irenarindos: thanks!

internal/db/schema/migrations/oss/postgres/88/01_worker_auth_invariant_trigger.up.sql
irenarindos: Er, and it won't always be true with the changes in this PR that detect when we have split brain creds and remove an incorrect `current` and update a `previous` to `current`

johanbrandhorst (approved):
🤞🏻 would love to have Jeffs review on this as well but LGTM!

irenarindos:
Adding @tmessi and @mgaffney to get an eye on the DB changes. Thanks!

tmessi (changes requested):

internal/server/repository_workerauth.go
```diff
@@ -250,6 +246,43 @@ func StoreNodeInformationTx(ctx context.Context, reader db.Reader, writer db.Wri
 		return new(types.DuplicateRecordError)
 	}
 
+	// It's possible a connection dropped during a rotate credentials response, so the control plane's stored
+	// previous and current WorkerAuth records may not match what the worker has stored.
+	// Check what the worker indicates is its previous key and fix what we have stored before inserting the new record.
+	if node.PreviousEncryptionKey != nil {
```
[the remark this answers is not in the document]
irenarindos: We were running into this when the response from cred rotation from the controller is dropped- the controller stores what it thinks is the correct previous/ next pair for the worker, while the worker discards the current cred it proposed since it failed to get a response. There's no way for the worker to know whether the controller got its request or only dropped the response, so we can't fix the issue there. We can only detect it when the worker sends another rotation request (which it will, since it assumes the first one failed), so that's what we're doing here.

internal/server/query.go
```diff
@@ -34,6 +34,25 @@ const (
...
+		delete from worker_auth_authorized where state = 'current' and worker_id in (select * from key_id_to_worker_id)
+	`
+
+	updateWorkerAuthStateByKeyId = `
+		update worker_auth_authorized set state = 'current' where worker_key_identifier = @worker_key_identifier
+	`
+
+	getWorkerAuthsByWorkerIdQuery = `
+		select * from worker_auth_authorized where worker_id = @worker_id
+	`
```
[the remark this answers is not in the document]
irenarindos: Sure, thanks!

tmessi (approved):

PR #4872, opened 2024-06-04
chore(e2e): Use public version of enos

moduli:
This PR updates the e2e test suite to utilize the public version of enos. Enos (both the CLI and the Terraform Provider) were moved to public repositories and, as a result, no longer require a user to need a TFC token.

For anyone with an existing setup, you can run the following
```
# Install enos from the new location
brew tap hashicorp/tap && brew update && brew install hashicorp/tap/enos

# Remove the TFC token from your enos vars
```

### Follow-up
- [ ] Update enterprise's `enos-run-ent.yml` to remove tfc_api_token
- [ ] Inform others of this change

https://hashicorp.atlassian.net/browse/ICU-13987

### DIFF
--- a/enos/README.md
+++ b/enos/README.md
@@ -13,17 +13,12 @@ for further information regarding installation, execution or composing Enos scen
 ## Requirements
 * AWS access. HashiCorp Boundary developers should use Doormat.
 * Terraform >= 1.0
-* Enos >= v0.0.10. You can [install it from a release channel](https://github.com/hashicorp/Enos-Docs/blob/main/installation.md) or use `make tools` install it into `$GOBIN`.
-* Access to the QTI org in Terraform Cloud. HashiCorp Boundary developers can
-  access this token in 1Password or request their own in #team-quality on slack.
+* Enos >= v0.0.28 (`brew tap hashicorp/tap && brew update && brew install hashicorp/tap/enos`)
 * An SSH keypair in the AWS region you wish to run the scenario. You can use
   doormat to login to the AWS console to create or upload an existing keypair.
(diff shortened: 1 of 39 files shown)


enos/README.md
```diff
@@ -13,17 +13,12 @@ for further information regarding installation, execution or composing Enos scen
 ## Requirements
 * AWS access. HashiCorp Boundary developers should use Doormat.
 * Terraform >= 1.0
-* Enos >= v0.0.10. You can [install it from a release channel](https://github.com/hashicorp/Enos-Docs/blob/main/installation.md) or use `make tools` install it into `$GOBIN`.
-* Access to the QTI org in Terraform Cloud. HashiCorp Boundary developers can
-  access this token in 1Password or request their own in #team-quality on slack.
+* Enos >= v0.0.10 (`brew tap hashicorp/tap && brew update && brew install hashicorp/tap/enos`)
```
ryancragun, line 16: Oh boy, 0.0.10 is ancient. I think the first public version was 0.0.28.

moduli: I'll update that.

Also, I need to update our reference to not use the qti vault module.

enos/enos-modules.hcl
```diff
@@ -104,7 +104,7 @@ module "aws_target" {
 }
 
 module "vault" {
-  source = "app.terraform.io/hashicorp-qti/aws-vault/enos"
+  source = "./modules/aws_vault"
```
moduli, line 107: For now, I'm going to just copy the contents of the of the shared terraform provider as opposed to updating to use the copy from the vault repo. There have been many a number of changes to this module that exist in the vault repo, so I'd like to save that for a separate activity.

PR #4877, opened 2024-06-05
Use appropriate gRPC error codes for OIDC

davidsbond:
This commit makes modifications to the OIDC request handlers to return appropriate gRPC status codes on errors. Before this change, boundary presents the internal error codes to clients. This causes problems for reverse proxies, as these status codes are mapped to non-standard HTTP response codes which are ultimately all turned into 500 responses when 4xx status codes are more appropriate and descriptive of the error returned.

Some modifications in this commit include calls to `event.WriteError` so that the internal error is written to the controller log for visibility. Comments about potential duplicate logs are now removed as `event.WriteError` is not called when using functions from the `handlers` package.


davidsbond:
@johanbrandhorst I'm personally happy for this change to go in without checking PRDE just because there is a unit test covering this scenario that previously expected the internal error code and now expects permission denied:

https://github.com/hashicorp/boundary/pull/4877/files#diff-8dc688d05107ea1a68ab36833787b6df37185c190c6a22bca9f58455a2e4438bL1799

davidsbond:
I haven't spun up PRDE in a while, but can take a look. Or perhaps just setting up OIDC in a dev cluster could work

johanbrandhorst (approved):
:shipit:

jimlambrt (approved):
Ty!

PR #4878, opened 2024-06-05
test(e2e): Add support for dynamic credentials

moduli:
This PR adds support for testing dynamic credentials when creating storage buckets.

Summary of changes
- The bucket module now takes a username as an optional argument. If a user is provided, it will set up the user policy to have access to the bucket. This is needed when testing static credentials
- The worker module now takes a bucket arn as an optional argument. If a bucket arn is provided, it will set up a role to gain access to that bucket, create an instance profile for that role, and attach it to the worker. This is needed when testing dynamic credentials]

There will be a follow-up PR to add a test in `boundary-enterprise`

https://hashicorp.atlassian.net/browse/ICU-13420

### DIFF
--- a/enos/modules/aws_bucket/variables.tf
+++ b/enos/modules/aws_bucket/variables.tf
@@ -12,7 +12,14 @@ variable "cluster_tag" {
   type        = string
 }
 
+variable "is_user" {
+  description = "Boolean to specify if a user was provided to this module."
+  type        = bool
+  default     = false
+}
+
 variable "user" {
   description = "A username that will be allowed access to this module's bucket."
   type        = string
+  default     = ""
 }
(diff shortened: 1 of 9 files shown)


enos/modules/aws_bucket/variables.tf
```diff
@@ -12,7 +12,14 @@ variable "cluster_tag" {
   type        = string
 }
 
+variable "is_user" {
+  description = "Boolean to specify if a user was provided to this module."
+  type        = bool
+  default     = false
+}
+
```
moduli, line 20: I'm not super happy about this implementation, but after discussing with others, there doesn't look to be an alternative at the moment.

I originally only had the modification to the `user` variable and added the check...
```
count = var.user != "" ? 1 : 0
```
but Terraform was not happy with that because it couldn't determine the value of `var.user` during the plan.

PR #4886, opened 2024-06-07
api: remove deprecated grant_scope_id field

johanbrandhorst:
The `grant_scope_id` field was deprecated in 0.15.0, so we can remove it for 0.17.0.


internal/cmd/commands/rolescmd/funcs.go
```diff
@@ -207,20 +205,6 @@ func extraFlagsFuncImpl(c *Command, _ *base.FlagSets, f *base.FlagSet) {
...
-		case 1:
-			if c.flagGrantScopeIds[0] == "null" {
-				*opts = append(*opts, roles.DefaultGrantScopeId())
-			} else {
-				*opts = append(*opts, roles.WithGrantScopeId(c.flagGrantScopeIds[0]))
-			}
-		default:
-			c.UI.Error("-grant-scope-id cannot be specified multiple times when using the deprecated create/update grant_scope_id mechanism")
-			return false
-		}
```
johanbrandhorst, line 222: Sanity check: we're ok to remove this now, right? This flag isn't supported for `create` and `update` anymore.

johanbrandhorst:
We may separately want to make a change to the boundary terraform provider, but I don't know if we have the same compatibility promise for that. CC @louisruch

jefferai (approved):

PR #4892, opened 2024-06-13
Remove trailing slash when using openid-configuration

### LINKED ISSUE #4710: Adding OIDC Keycloak with .well-known/openid-configuration failed because of trailing slash
Hi, here is another bug that I discovered while was trying to add OIDC Keycloak from UI Admin console in Boundary. 

If you add issuer with `.well-known/openid-configuration`, in my example it is `http://172.21.33.27:9999/realms/inventory/.well-known/openid-configuration`: 
Screenshot from 2024-04-24 17-57-44

And fill out the rest, you'll be able to add the OIDC Provider as expected, but, once you'll try to set it's state for example **Public**, it will fail:
Screenshot from 2024-04-24 17-58-48

Cuz the boundary set the issuer as - http://172.21.33.27:9999/realms/inventory/ - with the trailing slas in the end, and with "**/**" it does not match with the OIDC provider which is:
Screenshot from 2024-04-24 18-22-06

It doesn't include `**/**`

To resolve this issue you just need to delete that slash and the OIDC can be set to **Public** and after been used without any problems:

Screenshot from 2024-04-24 18-10-19

Screenshot from 2024-04-24 18-10-10

davidsbond:
This commit fixes #4710 by modifying the OIDC issuer parsing code to also strip the trailing slash when specifying an issuer value that uses the .well-known/openid-configuration suffix for configuration discovery.

Originally, this code keeps any trailing slash left on the issuer value once the suffix is removed. This does not appear to be in line with the specification stating that the issuer value must be the URL used as the prefix to /.well-known etc.


davidsbond:
Closing for now as I don't think this is the right solution

PR #4893, opened 2024-06-14
Add grant scope IDs to role list

jefferai:
This adds `grant_scope_ids` in the output of role listing. The lookup for grant scope IDs was changed to allow taking in a set of roles instead of just a single role, and the results are collated afterwards.


johanbrandhorst (approved):
LGTM!

irenarindos (approved):

PR #4903, opened 2024-06-25
chore(e2e): Clean up comments in enos.vars

moduli:
This PR makes some updates to `enos.vars.hcl` to clarify comments and make it easier to onboard.

Some additional changes were included to accommodate these changes
- `enos-local.vars.hcl` was added to `.gitignore` to allow a local copy of variables without potentially committing them
- Some variables were made optional so that users who don't need to run e2e tests can ignore them.

https://hashicorp.atlassian.net/browse/ICU-14090

### DIFF
--- a/enos/enos.vars.hcl
+++ b/enos/enos.vars.hcl
@@ -4,41 +4,45 @@
 // Example variable inputs. Set these to valid values and uncomment them before
 // running scenarios.
 
-// The AWS region you want to create the resources in. Make sure you choose a
-// region where you've got an AWS keypair.
-// aws_region = "us-east-1"
+// Recommended to copy this file to enos-local.vars.hcl and modify the values
+// there to avoid accidentally committing sensitive information.
 
-// The name of the AWS keypair. You can look them up in the AWS console on a per-
-// region basis. E.g. https://us-east-1.console.aws.amazon.com/ec2/v2/home?region=us-east-1#KeyPairs:
-// aws_ssh_keypair_name = "mykeypair"
-
-// The path to the private key associated with your keypair.
-// aws_ssh_private_key_path = "/Users/<user>/.ssh/mykeypair.pem
-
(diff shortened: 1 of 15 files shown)


enos/enos.vars.hcl
moduli, line 1: @calcaide @mallika-hc Please review to see if the changes here make more sense for you

mallika-hc (approved):
Looks good and the enos-local-vars.hcl is a great addition

enos/enos.vars.hcl
```diff
@@ -47,14 +51,24 @@
...
+// target_count = 1
 
-// Timeout for `go test` execution in the e2e tests, 10m default
-// go_test_timeout = "10m"
+// The directory that contains the copy of the boundary cli that the e2e tests
+// will use in CI. Only needed if e2e_debug_no_run = false.
+// local_boundary_dir = "/Users/<user>/.go/bin"
+
+// The directory that contains the source code of boundary/boundary-enterprise.
+// This is used in docker scenarios in CI in order to mount the source code into
```
calcaide, line 68: Is this only need when running in docker? or is need always?

moduli: 1. I put this in the `Optional Variables` section
2. It's only needed if `e2e_debug_no_run = false` (where enos would run the tests automatically). This variable is specifically for the backend tests. `local_boundary_ui_src_dir` is a similar thing, but for the Admin UI tests.

calcaide: No, all good, yeah I notice was on Optional variables, and now I realize the `e2e_debug_no_run`. I think is clear enough, I wanted to double check 😃

calcaide (approved):
This looks very good to me, much more context and information about enos vars and helping to avoid human errors (as I do). 

Thank you for the work!!! 🙌

PR #4925, opened 2024-07-10
cmd/connect: Ensure passthroughArgs are at the end.

the-maldridge:
The documentation suggests that any arguments supplied after a `--` will be appended to the end of the command.  This is not the case with current released builds, where boundary appends further information relevant to the connection type onto the end of the command.

This commit changes the argument order to match what the docs detail where anything after `--` is passed through unmodified to the command that's being started as a child of the boundary process.  Notably, this is important for tools that expect boundary to behave like the tool that its wrapping, since with the reversed argument ordering commands that work without boundary break when it is introduced.

### DIFF
--- a/internal/cmd/commands/connect/connect.go
+++ b/internal/cmd/commands/connect/connect.go
@@ -686,7 +686,7 @@ func (c *Command) handleExec(clientProxy *apiproxy.ClientProxy, passthroughArgs 
 		return
 	}
 
-	args = append(passthroughArgs, args...)
+	args = append(args, passthroughArgs...)
 
 	stringReplacer := func(in, typ, replacer string) string {
 		for _, style := range []string{


louisruch:
Hey @the-maldridge,

Thanks for submitting this PR and being an active user of Boundary. With regards to the connect helper commands (ssh, rdp, etc), Boundary constructs the command to `exec` based on a number of pieces of information (brokered credentials, the target type, flags etc). With the `passthroughArgs` we are aware that we will need to have more control for end users as to prepending or appending them to the Boundary constructed command to exec. Some binaries are more picky than others around ordering of args, for example the `ssh` docs specify the following format:

```ssh [arguments] destination [command [arguments]]```

Following this format the Boundary `ssh` helper command needs to insert the `destination` in between the `[arguments]` and the `[command [arguments]]`, since we are not aware of what the `passthroughArgs` contain we prepend all arguments and then added a `remote-command` flag specifically to the `ssh` helper to allow users to add a command and arugment to run on the target. While it is true that the SSH clients we have tested do not seem to care about the format shown above and do accept `[arguments]` after the `destination`, not all connect helper commands will play nicely with the change proposed in this PR.

You mention:
> The documentation suggests that any arguments supplied after a -- will be appended to the end of the command

Do you mind linking me where this is noted as this is not correct, all we claim is that the passthrough arguments will be set directly to the binary that is exec'ed.

One alternative here, if the `remote-command` does not do what you need, is to not use the `ssh` helper command and instead use `-exec` to construct your own command bypassing the arguments that Boundary adds on. Something like the following should work for the default tcp target Boundary creates:

```boundary connect ttcp_1234567890 -exec ssh -- username@{{boundary.ip}} -p {{boundary.port}} hostname```

the-maldridge:
Hi Louis, thanks for taking a look.

As written today `passthroughArgs` is not particularly useful, since you can't assert what is going to happen to your arguments without reading the source code which violates the principle of least surprise.  While the mitchellh CLI parser may have more flexibility in argument position, since the vast majority of applications out there today use POSIX flag parsing semantics, the having additional parameters bolted on to the end breaks things.

You raise a good point that other command helpers may not play nice with the change I've proposed, and I should have checked that prior to proposing this change.

Try as I might I cannot find anything that says specifically that arguments are appended.  I most likely assumed that was the behavior as, bluntly, its the only behavior that makes sense with tools like SSH where further commands may be chained onto it.  Reading the docs more thoroughly now I clearly see that it as you describe that they are sent directly to the binary.  Again, to be very blunt, this behavior as implemented is less than useless since I have to go read the boundary source code to understand what the actual command is that's going to be run.  A huge amount of troubleshooting could have been avoided if there was a way to get boundary to dump the entire string that's being passed to `exec`.

I am aware of remote-command, but at the risk of making a cheesy pun its not even remotely useful here.  My end goal is to have Ansible connect to machines and manage them by using boundary as the means of connection instead of directly calling SSH.  In order for this to work, ansible needs to be able to programatically drive boundary to setup tunnels and push traffic across them.  If you're not familiar with Ansible internals, you can think of it as an unmaintainably complex templating engine that spits out shell scripts and python scripts that when executed against a remote machine coerce that machine into a desired state. 

I had not considered calling `connect` directly, but that is probably the correct answer here since Ansible needs functionally complete control over the command line.  I had started this PR based on the assumption that target aliases were only processed in the command helper, but they appear to be processed in the general connect context as well.  I'm going to close this PR in favor of that approach, since that also works with SCP and SFTP which are necessary components of ansible's internals.

louisruch:
@the-maldridge thanks for the in-depth feedback, as I mentioned we are aware that we have work on this path to improve the UX. I will bring your thoughts to the team.
Please reach out if there are any other issues or concerns around using `connect` directly without the `ssh` helper.

PR #4926, opened 2024-07-10
chore: Fail workflow if request does not succeed

moduli:
The `trigger-merge-to-downstream` workflow had been failing for quite some time, but the workflow was still returning success. As a result, automatic merges from `boundary` to `boundary-enterprise` were not happening and there was no notification of a problem. Example: https://github.com/hashicorp/boundary/actions/runs/9843985789/job/27176402958

This PR makes an adjustment to the script to return a non-zero exit code if the curl request to the downstream repo returns a non 2xx exit code.

Testing locally
```
❯ bash trigger-merge-to-downstream-gha main
{ "message": "Resource not accessible by personal access token", "documentation_url": "https://docs.github.com/rest/actions/workflows#create-a-workflow-dispatch-event", "status": "403" } 403
ERROR: failed to trigger downstream workflow
```

https://hashicorp.atlassian.net/browse/ICU-14358

### DIFF
--- a/scripts/trigger-merge-to-downstream-gha
+++ b/scripts/trigger-merge-to-downstream-gha
@@ -21,8 +21,9 @@ if [[ -z ${DOWNSTREAM_WORKFLOW} ]]; then
     exit 0
 fi
 
-curl \
+CURL_RESPONSE=$(curl \
   --silent \
+  -w '%{http_code}\n' \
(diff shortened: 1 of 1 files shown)


johanbrandhorst (approved):
Thank you!

tmessi (approved):

PR #4937, opened 2024-07-17
Update attributes description for AuthenticateRequest to match proto

guygrigsby:
I was reading over the Swagger generated code while trying to auth via an API call and found the `attributes.login` field name to be out of date. This updates it to `attributes.login_name` to correspond to changes made to `PasswordLoginAttributes `.


johanbrandhorst (approved):
Thanks!!

johanbrandhorst (approved):
Thanks!

PR #4948, opened 2024-07-22
Add Google plugin for dynamic host catalog

joatmon08:
Add Google plugin for dynamic host catalog.


elimt:
The changes for adding the GCP Plugin was added in this #5229. Thanks Rosemary!

PR #4958, opened 2024-07-23
docs: Update roles in permissions concept

Dan-Heath:
Updates the description of roles to reflect multi-scope capabilities.

[View the page in the preview deployment](https://boundary-a30xemddz-hashicorp.vercel.app/boundary/docs/concepts/security/permissions#roles)

### DIFF
--- a/website/content/docs/concepts/security/permissions/index.mdx
+++ b/website/content/docs/concepts/security/permissions/index.mdx
@@ -62,7 +62,7 @@ Grant strings can be supplied via a human-friendly string syntax or via JSON.
 ## Roles
 
 Roles map grant strings to principals, which are users, groups, and managed groups.
-Every role assigns grants within a scope, as determined by the role's `grant-scope-ID` value.
+Every role assigns grants within a scope, as determined by the role's grant scope IDs.
 
 You can assign roles to multiple scopes to grant permissions to users who need access to resources across multiple scopes.
 You can also configure child scopes to inherit roles from parents.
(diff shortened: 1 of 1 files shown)


website/content/docs/concepts/security/permissions/index.mdx
```diff
@@ -61,16 +61,16 @@ Grant strings can be supplied via a human-friendly string syntax or via JSON.
...
-exists, or a scope that is a child of the scope in which the role exists,
-controlled by the role's "grant scope ID" value
-
-When a request is made, the scope in which to discover grants is either provided
-by the client (if against a resource collection itself) or is looked up using
-the resource's ID. This scope ID, along with the user's ID and the IDs of the
-groups the user belongs to, controls which roles are fetched to provide grants
-for the request.
+Roles map grant strings to principals, which are users, groups, and managed groups.
+Every role assigns grants within a scope, as determined by the role's `grant-scope-ID` value.
```
stellarsquall, line 65: should we say this using the CLI option, the read output, or plain english?

- `grant-scope-id`
- `Grant Scope IDs`
- as determined by the roles's grant scope IDs

Dan-Heath: Good question. I think I like the third option better. I defaulted to the CLI option, but I guess it would be different if they used the API. Maybe better to use plain English. It would be nice to add a link to the usage doc here, if we can complete the one for roles in time.

I updated this sentence to the plain English version, but let me know if you think there's a better option.

stellarsquall (approved):

PR #4980, opened 2024-07-29
Rab manage roles refactor

stellarsquall:
This PR refreshes the `common-workflows/manage-roles` usage docs, adds new screenshots, and adds them as an href to the `configuration/identity-access-management/` book.


website/public/img/ui/add-role-scopes_dark.png
Dan-Heath, line 1: I think this one is identical to the light version.

stellarsquall: good catch - that should be fixed!

Dan-Heath (approved):
Great work on this!! Thank you.

PR #4993, opened 2024-08-01
fix(db): remove old rewrapping-job table

ddebko:
# Summary

The `storage_plugin_storage_bucket_secret-rewrapping-job` table needs to be dropped because the `storage_plugin_storage_bucket_secret` table no longer exists and was replaced by the `storage_bucket_credential_managed_secret` table.


johanbrandhorst (approved):
LGTM but lets confirm by testing

elimt (approved):

PR #4998, opened 2024-08-06
docs: Fix broken anchor link in Credential management

Dan-Heath:
Fixes an old broken anchor link. This is from when titles used to include "HCP only" instead of using the banner.

[Test the update in the preview deployment](https://boundary-r0oo7hcy2-hashicorp.vercel.app/boundary/docs/concepts/credential-management)

image


stellarsquall (approved):
thanks!

PR #5005, opened 2024-08-09
test(e2e): Add test for worker api tags

moduli:
This PR updates an e2e test to add checks for worker API tags. The checks do the following
- Adds an API tag `k=v` to the worker
- Updates the target to use `k=v` in a worker filter and confirm you can connect to the target
- Modifies the API tag to `a=v`
- Confirm you cannot connect to the target with the `k=v` worker filter
- Update the target to use `a=v` in the worker filter and confirm you can connect to the target
- Remove the worker tags

https://hashicorp.atlassian.net/browse/ICU-14465

### DIFF
--- a/testing/internal/e2e/boundary/worker.go
+++ b/testing/internal/e2e/boundary/worker.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"slices"
 	"testing"
 
 	"github.com/hashicorp/boundary/api/workers"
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):
Nice! Do we plan to have some tests for the interactions between worker tags provided in config and via API? E.g. which one takes precedence, what happens if you remove one, etc?

moduli:
> Nice! Do we plan to have some tests for the interactions between worker tags provided in config and via API? E.g. which one takes precedence, what happens if you remove one, etc?

@johanbrandhorst I could add one more check here where the test creates an api tag of `type = egress` (same as the config tag) and confirms that you can connect to the target. Then, remove the api tag and confirm you can still connect to the target.

johanbrandhorst:
> @johanbrandhorst I could add one more check here where the test creates an api tag of `type = egress` (same as the config tag) and confirms that you can connect to the target. Then, remove the api tag and confirm you can still connect to the target.

Good idea, we should try to test:
1. That config values take precedence (what you suggested should cover this)
2. That worker tags are a union of the config and API values (I think this current test covers this, since it adds a tag that isn't in the config)

moduli:
I'm not sure how it validates precedence? Doesn't it just combine the config and api tags?

PR #5006, opened 2024-08-12
fix(daemon): Return 404 if alias interceptor ResolveRequestIds fails

hugoghx:


jefferai (approved):

moduli (approved):
Tested and verified that this is the following behavior. Just want to double-check with Adam that this is desired. 

Current Behavior
- alias doesn’t exist, logged in -> 401
- alias doesn’t exist, logged out -> 401
- alias exists, but logged out -> 401
- alias exists, but no permissions -> 403

New Behavior
- alias doesn’t exist, logged in -> **404**
- alias doesn’t exist, logged out -> **404**
- alias exists, but logged out -> 401
- alias exists, but no permissions -> 403

PR #5014, opened 2024-08-13
fix(policy): Cascade delete policy when scope is deleted

louisruch:


louisruch:
@hugoghx PTAL, this came up as a potential bug since a user could not delete a org without first deleting all policies. Since we cannot delete the global scope this will only impact orgs and since the policy can only be used by the org and is not directly associated with any session recordings I believe this change is sound. Please let me know if you can think of any reason to restrict delete.

hugoghx (approved):
LGTM I can't think of any issues with this change.

PR #5032, opened 2024-08-21
Update changelog for `0.16.3` & `0.17.1`

elimt:
I will add the changelog changes to the release branches after this PR is approved. Adding all authors of the PRs to verify the changelog additions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,23 +14,13 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 
 ### Bug Fixes
 
-* Minio large file support: Disable multipart uploads via minio to fix an issue where the file checksum is set incorrectly on each part 
-  of the upload, causing it to fail. This change fixes file uploads larger than 16MB and limits upload sizes to 5GB.
-  ([PR](https://github.com/hashicorp/boundary/pull/5013)) and ([PR](https://github.com/hashicorp/boundary-plugin-minio/pull/21))
-* Resolved an issue where session authorization was returning a `401` if the alias is non-existent or the alias does not resolve to anything.
-  A `404` status code is now returned.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -4,9 +4,56 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
...
+### New and Improved
+
+* Add `GetDownstreamWorkersTimeout` config option which represents the period of time (as a duration) timeout
+  for GetDownstreamWorkers call in DownstreamWorkerTicker. This is currently not documented and considered internal.
+  ([PR](https://github.com/hashicorp/boundary/pull/5007))
+
 ### Bug Fixes
 
-* fix(minio): disable multipart uploads to allow session recordings up to 5GB
+* Minio large file support: Disable multipart uploads via minio to fix an issue where the file checksum is set incorrectly on each part
```
louisruch, line 17: I had a quick sync with Jeff and it is proposed that instead of duplicating all the Bug Fixes here rather just add a one liner like:
`Contains Bug Fixes from 0.16.3` I am not sure if we want to include the Security changes in there as well, but it is strange reading a changelog that contains duplicate entries moving from 0.16.3 - 0.17.1
@psekar @AdamBouhmad thoughts?

psekar: I think it makes to have one entry in the latest verison 0.17.1 but mark it as backported to 0.16.3 Enterprise and Community editions. Probably this is the way forward for other backport releases too.

elimt: Yeah that's better. I for security changes, we could do the same. Something similar like `Contains Security Fixes from 0.16.3` under the `Security` heading

louisruch (approved):

PR #5035, opened 2024-08-21
ci: refuse to backport if migrations changed

johanbrandhorst:
I was able to test this successfully here: https://github.com/hashicorp/boundary/actions/runs/10498363248/job/29083143612?pr=5035


moduli (approved):
Seems fine.

Wondering if it's worth somehow posting a comment to the PR when it fails. Just an idea.

johanbrandhorst:
> Seems fine.
> 
> Wondering if it's worth somehow posting a comment to the PR when it fails. Just an idea.

Great idea, I stole the logic from the schema diff workflow, PTAL.

PR #5044, opened 2024-08-26
feat: Add support for capturing URL actions in request info

elimt:
Auth methods that do not support `:callback` were successfully authenticating with a token returned. This change prevents that behaviour

- Pass request context to services to validate actions.
- Validate actions for auth methods during authentication. `/v1/auth-methods/{{AUTH_METHOD_ID}}:authenticate:callback` is currently only supported on `OIDC` auth methods. An error is thrown if `:callback` is not supported
An error is thrown if not supported:

```
curl --location 'http://127.0.0.1:9200/v1/auth-methods/ampw_1234567890:authenticate:callback' \
--header 'Content-Type: application/json' \
--data '{
    "attributes": {
        "login_name": "admin",
        "password": "password"
    },
    "command": "login"
}'
```

```
{
    "kind": "InvalidArgument",
    "message": "Invalid fields provided in request.",
    "details": {
        "request_fields": [
            {
                "name": "request_path",
                "description": "callback is not a valid action for this auth method."
            }
        ]
    }
}
```


internal/daemon/controller/handler.go
```diff
@@ -737,3 +738,17 @@ func wrapHandlerWithCallbackInterceptor(h http.Handler, c *Controller) http.Hand
...
+	lastPart := parts[len(parts)-1]
+
+	items := strings.Split(lastPart, ":")
+
+	if len(items) > 1 {
+		return items[1:]
+	}
+
+	return []string{}
+}
```
johanbrandhorst, line 754: Arguably for a URL like `/v1/auth-methods/amoidc_1234567890:authenticate:callback` this should return both `authenticate` and `callback`, no? Could you add some tests for this function?

johanbrandhorst (commented):
I think we should be able to avoid exporting the verifier, which would make this much simpler.

internal/daemon/controller/auth/auth.go
```diff
@@ -109,7 +111,7 @@ type verifier struct {
 	oidcAuthRepoFn     common.OidcAuthRepoFactory
 	ldapAuthRepoFn     common.LdapAuthRepoFactory
 	kms                *kms.Kms
-	requestInfo        *authpb.RequestInfo
+	RequestInfo        *authpb.RequestInfo
```
johanbrandhorst, line 114: I don't think we need to export this either now, do we?

internal/daemon/controller/auth/auth.go
```diff
@@ -945,3 +947,11 @@ func (r *VerifyResults) ScopesAuthorizedForList(ctx context.Context, rootScopeId
 func (r *VerifyResults) GrantsHash(ctx context.Context) ([]byte, error) {
 	return r.grants.GrantHash(ctx)
 }
+
+func GetRequestInfo(ctx context.Context) (*authpb.RequestInfo, bool) {
```
johanbrandhorst, line 951: Please add a comment above this function to explain how it's used.
```suggestion
// GetRequestInfo extracts the request info stored in the context, if it exists.
// This returns nil, false if the request info could not be found.
func GetRequestInfo(ctx context.Context) (*authpb.RequestInfo, bool) {
```

internal/daemon/controller/handler.go
```diff
@@ -649,7 +650,7 @@ func wrapHandlerWithCallbackInterceptor(h http.Handler, c *Controller) http.Hand
 			return
 		}
 
-		req.URL.Path = strings.TrimSuffix(req.URL.Path, ":callback")
+		req.URL.Path = strings.TrimSuffix(req.URL.Path, fmt.Sprintf(":%s", auth.CallbackAction))
```
johanbrandhorst, line 653: Maybe even simpler:
```suggestion
		req.URL.Path = strings.TrimSuffix(req.URL.Path, ":"+auth.CallbackAction)
```

internal/daemon/controller/handler.go
```diff
@@ -737,3 +738,20 @@ func wrapHandlerWithCallbackInterceptor(h http.Handler, c *Controller) http.Hand
...
 	})
 }
+
+// getActions takes in a URL Path and returns the actions from the URL
+func getActions(path string) []string {
+	// Remove any query parameters
+	path = strings.Split(path, "?")[0]
+
+	parts := strings.Split(path, "/")
+	lastPart := parts[len(parts)-1]
```
johanbrandhorst, line 748: If this is a slash separated path, you can use `path.Base` to get the last element. You'll need to rename the `path` variable to avoid shadowing the `path` standard library package.
```suggestion
	lastPart := path.Base(urlPath)
```

internal/daemon/controller/handler.go
```diff
@@ -737,3 +738,20 @@ func wrapHandlerWithCallbackInterceptor(h http.Handler, c *Controller) http.Hand
...
+
+	items := strings.Split(lastPart, ":")
+
+	if len(items) > 1 {
+		return items[1:]
+	}
+
+	return []string{}
```
johanbrandhorst, line 756: Maybe a little simpler. Up to you.
```suggestion
	// Cut off everything before the first ":"
	_, rest, _ := strings.Cut(lastPart, ":")
	
	// Split the rest on ":", returning all actions and subactions
	return strings.Split(rest, ":")
```

internal/event/event.go
```diff
@@ -14,12 +14,13 @@ type (
...
-	Path     string `json:"path,omitempty" class:"public"`
-	PublicId string `json:"public_id,omitempty" class:"public"`
-	ClientIp string `json:"client_ip,omitempty" class:"public"`
+	EventId  string   `json:"-"`
+	Id       string   `json:"id,omitempty" class:"public"`
+	Method   string   `json:"method,omitempty" class:"public"`
+	Path     string   `json:"path,omitempty" class:"public"`
+	PublicId string   `json:"public_id,omitempty" class:"public"`
+	ClientIp string   `json:"client_ip,omitempty" class:"public"`
+	Actions  []string `json:"actions,omitempty" class:"public"`
```
johanbrandhorst, line 23: Do we need to add this? This will add this line to every event we send, right? Perhaps we could parse the actions out of the path in the handler instead?

internal/proto/controller/auth/v1/auth.proto
```diff
@@ -51,4 +51,7 @@ message RequestInfo {
 
   // the client ip for the request
   string client_ip = 140;
+
+  // url actions from the request path
+  repeated string actions = 141;
```
johanbrandhorst, line 56: Ditto is this required?

internal/event/event.go
elimt: Yeah we actually do not need this added to eventing anymore. That was part of another implementation where I wanted to use the RequestInfo event to set the actions

internal/proto/controller/auth/v1/auth.proto
elimt: This is required to set the actions before https://github.com/hashicorp/boundary/blob/96688b36d577a8769f823e346f52d7e63abab2ac/internal/daemon/controller/handler.go#L652 interceptor trims the `:callback` action.
For example, if we get a request like `http://127.0.0.1:9200/v1/auth-methods/ampw_1234567890:authenticate:callback`, when it gets to the service the path will be `http://127.0.0.1:9200/v1/auth-methods/ampw_1234567890:authenticate` because `:callback` is trimmed. This field basically captures all actions before changes are made to the url path

internal/daemon/controller/auth/auth.go
```diff
@@ -55,6 +55,8 @@ const (
 	AuthTokenTypeRecoveryKms
 )
 
+const CallbackAction = "callback"
```
johanbrandhorst, line 58: Please add a comment explaining what this is used for

johanbrandhorst (approved):
Great, thanks for this!

PR #5045, opened 2024-08-26
test(hcp): Define TF module for long-lived resources

moduli:
This PR adds a Terraform module that creates resources for testing against a long-lived HCP cluster. Right now, the resources include those necessary to perform Session Recording, including a worker, storage bucket, and target. An internal TFC account is used to store the state.

https://hashicorp.atlassian.net/browse/ICU-14783


johanbrandhorst (approved):

enos/ci/hcp-resources/main.tf
```diff
@@ -0,0 +1,226 @@
...
+provider "enos" {
+  transport = {
+    ssh = {
+      user             = "ubuntu"
+      private_key_path = abspath(var.aws_ssh_private_key_path)
+    }
+  }
+}
+
+variable "aws_region" {
```
guygrigsby, line 39: vars are often in a dedicated file.

enos/ci/hcp-resources/main.tf
```diff
@@ -0,0 +1,226 @@
...
+  enos_user            = local.cluster_tag
+  environment          = local.environment_tag
+  project_name         = local.project_tag
+  ami_id               = module.base_infra.ami_ids["ubuntu"]["amd64"]
+  vpc_id               = module.base_infra.vpc_id
+  subnet_ids           = module.worker.subnet_ids
+  additional_tags      = module.target_tags.tag_map
+}
+
+output "bucket_access_key_id" {
```
guygrigsby, line 186: outputs are often in a dedicated file.

guygrigsby (commented):
This is awesome! Thanks for doing this. 

Let me start by saying, I am not a TF expert at all. Not even close. 

That being said, Is this for a single set of shared resources? It wasn't clear based on the ticket and it seems like it's kind of a mix of that and having the modules defined for external users spin up their own things. I don't the former is suited for the Boundary community repo. We call out explicitly the hashicorp-qti org for TFC so that confused me a bit. Is the hope that other outside of HashiCorp would be able to use it too or that it maintains our set of long lived resources?

Maybe we need the modules defined publicly and our "test bed" of resources use those public modules in a private repo.

moduli:
> That being said, Is this for a single set of shared resources? It wasn't clear based on the ticket and it seems like it's kind of a mix of that and having the modules defined for external users spin up their own things. I don't think the former is suited for the Boundary community repo. We call out explicitly the hashicorp-qti org for TFC so that confused me a bit. Is the hope that other outside of HashiCorp would be able to use it too or that it maintains our set of long lived resources?
> 
> Maybe we need the modules defined publicly and our "test bed" of resources use those public modules in a private repo.

@guygrigsby 
- This is intended to define a single set of shared resources for HCP testing.
- The modules used here are not intended/designed for external use. They are used in our end-to-end testing. The idea was to borrow those same modules to set up the resources needed for our long-running tests.
- We could eventually move towards a paradigm of having those modules in some public repository, but this was the simplest way to move forward with the initiative.

Thoughts?

guygrigsby:
@moduli that makes sense and was what I was hoping. I am on board.

enos/ci/hcp-resources/.gitignore
```diff
@@ -0,0 +1 @@
+*.tfvars
```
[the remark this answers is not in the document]
moduli: Discussed with Guy. We'll keep this separate for now.

PR #5046, opened 2024-08-26
feat: Validate LDAP URL port number

elimt:
Add validation for the port number in LDAP URLs. If a port number is provided, it is checked to ensure it is a valid integer and within the valid range. This validation helps prevent errors when configuring LDAP authentication.


jimlambrt (approved):
Looks good. ty!

louisruch (approved):
LGTM

PR #5048, opened 2024-08-27
Return HTTP 400 status for inactive OIDC auth methods

davidsbond:
This commit modifies the OIDC authentication start method to return a 400 (Bad Request) error if a client attempts to start the authentication flow against an OIDC auth method that is set to inactive. This is something we noticed in our error logs in HCP Boundary that is impacting our SLOs.

Previously this returned its own error code which is mapped to a 500 response. Tests have been updated to catch this new scenario.

### DIFF
--- a/internal/daemon/controller/handlers/authmethods/oidc.go
+++ b/internal/daemon/controller/handlers/authmethods/oidc.go
@@ -178,7 +178,14 @@ func (s Service) authenticateOidcStart(ctx context.Context, req *pbs.Authenticat
 	}
 
 	authUrl, tokenId, err := oidc.StartAuth(ctx, s.oidcRepoFn, req.GetAuthMethodId(), opts...)
-	if err != nil {
+	switch {
+	case errors.Match(errors.T(errors.AuthMethodInactive), err):
+		return nil, handlers.ApiErrorWithCodeAndMessage(codes.FailedPrecondition, "Cannot start authentication against an inactive OIDC auth method")
+	case errors.Match(errors.T(errors.RecordNotFound), err):
+		return nil, handlers.ApiErrorWithCodeAndMessage(codes.NotFound, "Auth method %s was not found", req.GetAuthMethodId())
+	case errors.Match(errors.T(errors.InvalidParameter), err):
+		return nil, handlers.ApiErrorWithCodeAndMessage(codes.InvalidArgument, err.Error())
+	case err != nil:
 		event.WriteError(ctx, op, err, event.WithInfoMsg("error starting the oidc authentication flow"))
 		return nil, handlers.ApiErrorWithCodeAndMessage(codes.Internal, "Error generating parameters for starting the OIDC flow. See the controller's log for more information.")
 	}
--- a/internal/daemon/controller/handlers/authmethods/oidc_test.go
+++ b/internal/daemon/controller/handlers/authmethods/oidc_test.go
@@ -1624,9 +1624,10 @@ func TestAuthenticate_OIDC_Start(t *testing.T) {
 	s := getSetup(t)
 
 	cases := []struct {
-		name    string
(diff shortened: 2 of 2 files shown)


johanbrandhorst (approved):
Amazing! My only question is if using 412 has precedent for auth methods. Why did we choose that status code specifically?

davidsbond:
@johanbrandhorst I went with failed precondition as the situation we're in seems to line up with the grpc definition:

> The operation was rejected because the system is not in a state required for the operation’s execution.

In this case, the state of the system (i.e the status of the OIDC auth method) is not in a state where we can start the auth flow. So the OIDC auth method being public feels like a precondition to its use.

johanbrandhorst:
> @johanbrandhorst I went with failed precondition as the situation we're in seems to line up with the [grpc definition](https://grpc.io/docs/guides/status-codes/):
> 
> > The operation was rejected because the system is not in a state required for the operation’s execution.
> 
> In this case, the state of the system (i.e the status of the OIDC auth method) is not in a state where we can start the auth flow. So the OIDC auth method being public feels like a precondition to its use.
> 
> Ultimately for HCP's purposes any 4xx status code is fine.

I agree that the error code makes sense, but generally in Boundary we care more about using the correct HTTP status code than the correct gRPC status code. I just want to make sure that we're not going against some existing pattern. I did a search and it seems we already use this code for various errors, so I think this is fine :)

internal/daemon/controller/handlers/authmethods/oidc.go
```diff
@@ -178,7 +178,14 @@ func (s Service) authenticateOidcStart(ctx context.Context, req *pbs.Authenticat
 	}
 
 	authUrl, tokenId, err := oidc.StartAuth(ctx, s.oidcRepoFn, req.GetAuthMethodId(), opts...)
-	if err != nil {
+	switch {
+	case errors.Match(errors.T(errors.AuthMethodInactive), err):
+		return nil, handlers.ApiErrorWithCodeAndMessage(codes.FailedPrecondition, "Cannot start authentication against an inactive OIDC auth method")
```
uruemu, line 183: A note that this actually returns a HTTP 400 error from the [GRPC gateway](https://github.com/grpc-ecosystem/grpc-gateway/blob/93189d015a3e23872723445862dc6fe1c4502913/runtime/errors.go#L58-L60). I assume this would still make sense since it is on the client to activate the OIDC auth method, does that sound correct?

uruemu: Alternatively, could the GRPC `Aborted` error could be more suitable here? It maps to the `HTTP.StatusConflict`,` which is more descriptive to the case but not sure `Aborted` is more suitable than `FailedPrecondition` here

https://github.com/grpc-ecosystem/grpc-gateway/blob/93189d015a3e23872723445862dc6fe1c4502913/runtime/errors.go#L61-L62

uruemu: I just saw that @johanbrandhorst was having a relevant discussion. This might be useful context, not sure if 400 is suitable here

davidsbond: Oh that's interesting, I wasn't aware that was the mapping behaviour. I'm personally fine with a regular 400.

johanbrandhorst: Yes, the mapping is 400, I think that's fine here.

davidsbond: I've updated the commit & PR title/description to reflect this is a 400 not a 412.

uruemu (approved):

davidsbond:
Going to merge, the tests that are failing seem unrelated and they are failing across all other PRs

PR #5049, opened 2024-08-27
Restrict cache results to 250 entries by default

jefferai:
A query parameter `max_result_set_size` can be specified to set a per-request limit, with `-1` meaning all values. If there were more entries than the max size used for the request, an `incomplete` boolean is set in the response.


johanbrandhorst:
I think the lint failure is some sort of cache problem, `typecheck` errors usually are

johanbrandhorst (approved):
LGTM

jimlambrt (approved):

PR #5051, opened 2024-08-28
feat (clientcache): use a persistent cache

jimlambrt:
By default, we will use a persistent cache for the client cache. This will allow us to keep the cache across restarts of the server and it will also reduce the amount of time it takes to start the server while reducing the amount of memory used.

This change also includes a validation of the cache schema version. If the schema version is different from the one expected, the cache will be reset/recreated.

### DIFF
--- a/internal/clientcache/internal/daemon/server.go
+++ b/internal/clientcache/internal/daemon/server.go
@@ -30,7 +30,6 @@ import (
 	"github.com/hashicorp/boundary/version"
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-secure-stdlib/parseutil"
-	"github.com/mitchellh/go-homedir"
 	"golang.org/x/text/cases"
 	"golang.org/x/text/language"
 )
@@ -543,7 +542,7 @@ func defaultDbUrl(ctx context.Context, opt ...Option) (string, error) {
 		return "", errors.Wrap(ctx, err, op)
 	}
 	if opts.withHomeDir == "" {
-		opts.withHomeDir, err = homedir.Dir()
+		opts.withHomeDir, err = os.UserHomeDir()
 		if err != nil {
 			return "", errors.Wrap(ctx, err, op)
 		}
--- a/internal/clientcache/internal/daemon/server.go
+++ b/internal/clientcache/internal/daemon/server.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"path/filepath"
 	"sort"
 	"strings"
 	"sync"
@@ -29,6 +30,7 @@ import (
 	"github.com/hashicorp/boundary/version"
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-secure-stdlib/parseutil"
+	"github.com/mitchellh/go-homedir"
 	"golang.org/x/text/cases"
 	"golang.org/x/text/language"
 )
@@ -496,6 +498,10 @@ func setupEventing(ctx context.Context, logger hclog.Logger, serializationLock *
 	return nil
 }
 
+// openStore will open the underlying store for the db. If no options are
+// provided, it will default to an on disk store using the user's home dir +
+// ".boundary/cache.db". If a url is provided, it will use that as the store.
+// Supported options: WithUrl, WithLogger, WithHomeDir
 func openStore(ctx context.Context, opt ...Option) (*db.DB, error) {
 	const op = "daemon.openStore"
 	opts, err := getOpts(opt...)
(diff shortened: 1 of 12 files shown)


mgaffney (approved):

internal/clientcache/internal/daemon/server.go
```diff
@@ -29,6 +30,7 @@ import (
 	"github.com/hashicorp/boundary/version"
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-secure-stdlib/parseutil"
+	"github.com/mitchellh/go-homedir"
```
jefferai, line 33: I think go-homedir may be out of date. There was a proposal later for `os.UserHomeDir` (https://github.com/golang/go/issues/26463) and I am not seeing `cgo` in the implementation (https://cs.opensource.google/go/go/+/refs/tags/go1.23.0:src/os/file.go;l=574) so maybe we can just use that instead?

jefferai (changes requested):
Looks pretty good but seems like it's still using inmem by default? Or am I just missing something.

Also I think if we feel good about the `os.UserHomeDir` option not using `cgo` we should use that and not use `go-homedir`.

internal/clientcache/internal/db/db.go
```diff
@@ -50,16 +57,38 @@ func Open(ctx context.Context, opt ...Option) (*db.DB, error) {
 	conn.Debug(opts.withDebug)
 
 	switch {
-	case opts.withDbType == dbw.Sqlite:
+	case opts.withDbType == dbw.Sqlite && url == DefaultStoreUrl:
```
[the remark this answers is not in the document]
jimlambrt: Yes, since there are a ton of tests that rely on this (don't setup a dir/filename).  There's only one place we need to specify the file based url and it's clear in the code.

internal/clientcache/internal/daemon/server.go
jimlambrt: ty!

jimlambrt:
> Looks pretty good but seems like it's still using inmem by default? Or am I just missing something.
> 


We're using the home dir for the server startup.  See `defaultDbUrl(..)`

jefferai:
Sorry, right, that was obvious in hindsight. We're always passing in the option.

jefferai (approved):

internal/clientcache/internal/daemon/server.go
```diff
@@ -521,3 +532,31 @@ func openStore(ctx context.Context, opt ...Option) (*db.DB, error) {
 	}
 	return store, nil
 }
+
+// defaultDbUrl returns the default db name including the path. It will ensure
+// the directory exists by creating it if it doesn't.
+func defaultDbUrl(ctx context.Context, opt ...Option) (string, error) {
+	const op = "daemon.DefaultDotDirectory"
```
johanbrandhorst, line 539:
```suggestion
	const op = "daemon.defaultDbUrl"
```

internal/clientcache/internal/daemon/server.go
```diff
@@ -521,3 +532,31 @@ func openStore(ctx context.Context, opt ...Option) (*db.DB, error) {
...
+		return "", errors.Wrap(ctx, err, op)
+	}
+	fileName := filepath.Join(dotDir, dbFileName)
+	return fmt.Sprintf("%s%s", fileName, fkPragma), nil
+}
+
+const (
+	dotDirname = ".boundary"
+	dbFileName = "cache.db"
+	fkPragma   = "?_pragma=foreign_keys(1)"
```
johanbrandhorst, line 561: Mind adding a comment explaining the source of this? I assume it's from the [SQLite docs](https://www.sqlite.org/foreignkeys.html#fk_enable)?

internal/clientcache/internal/db/db.go
```diff
@@ -68,3 +97,56 @@ func createTables(ctx context.Context, conn *db.DB) error {
...
+	Version string
+	// UpdateTime is the last update of the version
+	UpdateTime time.Time
+	// CreateTime is the create time of the initial version
+	CreateTime time.Time
+}
+
+const (
+	schemaTableName      = "schema_version"
+	schemaCurrentVersion = "v0.0.1"
```
johanbrandhorst, line 146: Should we not just use incrementing integers for this? I can't imagine semantic versioning making sense for database schemas, though in practice I guess it just needs to be a different string from any previous version.

internal/clientcache/internal/db/schema.sql
```diff
@@ -2,6 +2,40 @@
...
+
+
+create trigger update_time_column_schema_version
+before update on schema_version
+for each row 
+when 
+  new.version <> old.version 
+  begin
+    update schema_version set update_time = datetime('now','localtime') where rowid == new.rowid;
+  end;
```
johanbrandhorst, line 34: The combination of these two triggers confuses me. One forbids the create_time from being set, the other automatically sets the update time. Is the table expected to be updated or not? Could you perhaps add a comment explaining the need for both of these?

jefferai: The first one forbids create time from being changed, not set.

internal/clientcache/internal/db/db.go
jimlambrt: this is prior art from go-kms-wrapping/extras and I think it's sufficient.

PR #5052, opened 2024-08-28
test(host): validate ipv6 address for hosts

ddebko:
# Summary

A new set of tests were added to the create/update repository methods of the static host resource. These tests were added to validate that an IPv6 address can be used by a host. The existing UpsertHost test for the dynamic host resource was updated to validate ensure an IPv6 address can be used by a host.


kheina (approved):
LGTM

PR #5053, opened 2024-08-28
Add `implicit-scopes` resource type to cache

jefferai:
This allows getting a list of scope IDs known by the cache via cached targets and sessions. It is not refreshed from the controller. The name contains `implicit` in case we ever want e.g. `scopes`.

### DIFF
--- a/internal/clientcache/internal/cache/repository_implicit_scopes.go
+++ b/internal/clientcache/internal/cache/repository_implicit_scopes.go
@@ -1,0 +1,115 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package cache
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/boundary/api/scopes"
+	"github.com/hashicorp/boundary/internal/errors"
+)
+
+func (r *Repository) ListImplicitScopes(ctx context.Context, authTokenId string, opt ...Option) (*SearchResult, error) {
+	const op = "cache.(Repository).ListImplicitScopes"
+	switch {
+	case authTokenId == "":
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "auth token id is missing")
+	}
+	ret, err := r.searchImplicitScopes(ctx, "true", nil, append(opt, withAuthTokenId(authTokenId))...)
+	if err != nil {
+		return nil, errors.Wrap(ctx, err, op)
+	}
+	return ret, nil
+}
+
+// QueryImplicitScopes does nothing currently; it's not its own table, but a
+// union of scope info from other types. So it just calls searchImplicitScopes
+// as if it was a list. It is required to fulfill the interface, but will return
+// an error if used from the API.
+func (r *Repository) QueryImplicitScopes(ctx context.Context, authTokenId, query string, opt ...Option) (*SearchResult, error) {
+	const op = "cache.(Repository).QueryImplicitScopes"
+	switch {
+	case authTokenId == "":
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "auth token id is missing")
+	}
+	ret, err := r.searchImplicitScopes(ctx, "true", nil, append(opt, withAuthTokenId(authTokenId))...)
+	if err != nil {
+		return nil, errors.Wrap(ctx, err, op)
+	}
+	return ret, nil
+}
+
+func (r *Repository) searchImplicitScopes(ctx context.Context, condition string, searchArgs []any, opt ...Option) (*SearchResult, error) {
+	const op = "cache.(Repository).searchImplicitScopes"
+	switch {
+	case condition == "":
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "condition is missing")
+	}
+
+	opts, err := getOpts(opt...)
+	if err != nil {
+		return nil, errors.Wrap(ctx, err, op)
+	}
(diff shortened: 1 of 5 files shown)


internal/clientcache/internal/cache/repository_implicit_scopes.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return nil, errors.Wrap(ctx, err, op)
+	}
+	return ret, nil
+}
+
+// QueryImplicitScopes does nothing currently; it's not its own table, but a
+// union of scope info from other types. So it just calls searchImplicitScopes
+// as if it was a list. It is required to fulfill the interface, but will return
+// an error if used from the API.
```
johanbrandhorst, line 30: Why not return an error here too and make it impossible to accidentally use?

internal/clientcache/internal/cache/repository_implicit_scopes.go
```diff
@@ -0,0 +1,115 @@
...
+
+	rows, err := r.rw.Query(ctx, unionQuery, searchArgs)
+	if err != nil {
+		return nil, errors.Wrap(ctx, err, op)
+	}
+	defer rows.Close()
+
+	type ScopeIdsResult struct {
+		FkUserId string `gorm:"primaryKey"`
+		ScopeId  string `gorm:"default:null"`
```
johanbrandhorst, line 87: Do these tags matter for reading? Especially the null tag seems unnecessary.

internal/clientcache/internal/cache/repository_implicit_scopes.go
```diff
@@ -0,0 +1,115 @@
...
+	dedupMap := make(map[string]struct{}, len(scopeIdsResults))
+	for _, res := range scopeIdsResults {
+		dedupMap[res.ScopeId] = struct{}{}
+	}
+
+	sr := &SearchResult{
+		ImplicitScopes: make([]*scopes.Scope, 0, len(dedupMap)),
+	}
+	for k := range dedupMap {
+		sr.ImplicitScopes = append(sr.ImplicitScopes, &scopes.Scope{Id: k})
```
johanbrandhorst, line 111: These scopes will appear in a non-deterministic order, perhaps sort before returning?

internal/clientcache/internal/cache/repository_implicit_scopes_test.go
```diff
@@ -0,0 +1,126 @@
...
+			UserId:   "u_123",
+			Type:     "tcp",
+		},
+	}
+	require.NoError(t, r.refreshSessions(ctx, u1, map[AuthToken]string{{Id: "id"}: "something"},
+		WithSessionRetrievalFunc(testStaticResourceRetrievalFunc(t, [][]*sessions.Session{ss}, [][]string{nil}))))
+
+	expectedScopes = append(expectedScopes, &scopes.Scope{
+		Id: ss[0].ScopeId,
+	})
```
johanbrandhorst, line 107: Could we test this with more than 1 result so we discover non-deterministic ordering?

internal/clientcache/internal/cache/repository_implicit_scopes.go
jefferai: I copied this from Targets, which also has the same tags. I'm not sure the tags are right on Targets - there will always be a scope ID for instance - but I figured it was that way for a reason.

internal/clientcache/internal/cache/repository_implicit_scopes.go
jefferai: I could, I was leaving it open for the future. But I can do that now!

internal/clientcache/internal/cache/repository_implicit_scopes.go
jefferai: I thought about that but didn't see there being an advantage in terms of what it's used for by downstream clients. If you think that's useful though I can.

internal/clientcache/internal/cache/repository_implicit_scopes_test.go
jefferai: It will be non-deterministic on purpose, but I can change that if we want.

internal/clientcache/internal/cache/search.go
```diff
@@ -23,11 +24,12 @@ const (
 	ResolvableAliases SearchableResource = "resolvable-aliases"
 	Targets           SearchableResource = "targets"
 	Sessions          SearchableResource = "sessions"
+	ImplicitScopes    SearchableResource = "implicit-scopes"
```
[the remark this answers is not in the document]
jefferai: I'm not sure where we explain any of it to users. :-D If we add scopes that will not change the behavior of this.

internal/clientcache/internal/cache/repository_implicit_scopes.go
johanbrandhorst: I see, I guess they don't hurt

internal/clientcache/internal/cache/repository_implicit_scopes.go
johanbrandhorst: I think it should be sorted on the API at least, to make API calls deterministic

jefferai: OK

johanbrandhorst (approved):

jimlambrt (approved):

PR #5054, opened 2024-08-29
Ensure only one refresh is running for a user/resource at a time

jefferai:
This ensures we don't have competing update processes running. When doing the inline RefreshForSearch, a specific error will be returned so we can hint in the API that we're in the middle of a refresh.

### DIFF
--- a/internal/clientcache/internal/cache/refresh.go
+++ b/internal/clientcache/internal/cache/refresh.go
@@ -19,6 +19,8 @@ import (
 	"github.com/hashicorp/go-hclog"
 )
 
+// This is used as an internal error to indicate that a refresh is already in
+// progress, as a signal that the caller may want to handle it a different way.
 var ErrRefreshInProgress error = stderrors.New("cache refresh in progress")
 
 type RefreshService struct {
--- a/internal/clientcache/internal/cache/refresh.go
+++ b/internal/clientcache/internal/cache/refresh.go
@@ -230,8 +230,9 @@ func (r *RefreshService) RefreshForSearch(ctx context.Context, authTokenid strin
 				args = append(args, "alias staleness", time.Since(rtv.UpdateTime))
 			}
 
-			if opts.withTestRefreshSleepDuration > 0 {
-				time.Sleep(opts.withTestRefreshSleepDuration)
+			if opts.withTestRefreshWaitChs != nil {
+				close(opts.withTestRefreshWaitChs.firstSempahore)
+				<-opts.withTestRefreshWaitChs.secondSemaphore
 			}
 
 			r.logger.Debug("refreshing aliases before performing search", args...)
@@ -260,8 +261,9 @@ func (r *RefreshService) RefreshForSearch(ctx context.Context, authTokenid strin
 				args = append(args, "target staleness", time.Since(rtv.UpdateTime))
 			}
 
-			if opts.withTestRefreshSleepDuration > 0 {
-				time.Sleep(opts.withTestRefreshSleepDuration)
+			if opts.withTestRefreshWaitChs != nil {
+				close(opts.withTestRefreshWaitChs.firstSempahore)
+				<-opts.withTestRefreshWaitChs.secondSemaphore
 			}
 
 			r.logger.Debug("refreshing targets before performing search", args...)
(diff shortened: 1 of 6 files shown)


internal/clientcache/internal/cache/refresh.go
```diff
@@ -135,21 +145,25 @@ func (r *RefreshService) cacheSupportedUsers(ctx context.Context, in []*user) ([
...
+// we should not be making an initial load of all resources while blocking a
+// search query, in case we have not yet even attempted to load the resources
+// for this user yet. Note: Currently, if the context timesout we stop
+// refreshing completely and return to the caller.  A possible enhancement in
+// the future would be to return when the context is Done, but allow the refresh
+// to proceed in the background.
+//
+// If a refresh is already running for that user and resource type we will
+// return ErrRefreshInProgress. If so we will not start another one and we can
+// make it clear in the response that cache filling is ongoing.
```
johanbrandhorst, line 166: Instead of this, what about using https://pkg.go.dev/golang.org/x/sync/singleflight to delay the response until the previous refresh has finished? This could be coupled with a timeout for time sensitive callers.

internal/clientcache/internal/cache/refresh.go
```diff
@@ -17,9 +19,17 @@ import (
 	"github.com/hashicorp/go-hclog"
 )
 
+var ErrRefreshInProgress error = stderrors.New("cache refresh in progress")
```
johanbrandhorst, line 22: Mind adding a comment to explain the use of this and how callers should handle it?

internal/clientcache/internal/cache/refresh_test.go
```diff
@@ -803,6 +803,196 @@ func TestRefreshForSearch(t *testing.T) {
...
+		go func() {
+			defer wg.Done()
+			blockingRefreshError := rs.RefreshForSearch(ctx, at.Id, Targets, append(opts, extraOpts...)...)
+			assert.NoError(t, blockingRefreshError)
+		}()
+		go func() {
+			defer wg.Done()
+			// Sleep here to ensure ordering of the calls since both goroutines
+			// are spawned at the same time
+			time.Sleep(testRefreshDuration / 2)
```
johanbrandhorst, line 867: Sleeps in tests suck, they make tests slower and at high load they introduce flakiness, could we use a channel to synchronize instead? Perhaps the first call could take an option with a channel that must be closed before continuing, and the second call could take an option with the same channel and close it after taking the sync.Map semaphore. No flakiness, minimal time waiting.

internal/clientcache/internal/cache/refresh_test.go
```diff
@@ -803,6 +803,196 @@ func TestRefreshForSearch(t *testing.T) {
...
+			nonblockingRefreshError := rs.RefreshForSearch(ctx, at.Id, Targets, append(opts, extraOpts...)...)
+			assert.ErrorIs(t, nonblockingRefreshError, ErrRefreshInProgress)
+		}()
+		wg.Wait()
+
+		// Unlike in the TestRefreshForSearch test, since we are did a force
+		// refresh we do expect to see values
```
johanbrandhorst, line 874:
```suggestion
		// Unlike in the TestRefreshForSearch test, since we did a force
		// refresh we do expect to see values
```

internal/clientcache/internal/daemon/search_handler.go
```diff
@@ -22,13 +23,23 @@ import (
 	"github.com/hashicorp/go-hclog"
 )
 
+type RefreshStatus string
+
+const (
+	NotRefreshing RefreshStatus = "not-refreshing"
+	Refreshing    RefreshStatus = "refreshing"
+	RefreshError  RefreshStatus = "refresh-error"
+)
```
johanbrandhorst, line 32: Mind adding comments to these exported variables?

internal/clientcache/internal/cache/refresh.go
jefferai: Callers don't handle it - this is internal to the cache.

johanbrandhorst: I mean callers of this function, not just CLI consumers 😛

jefferai: Added a comment

internal/clientcache/internal/daemon/search_handler.go
jefferai: Done.

internal/clientcache/internal/cache/refresh_test.go
jefferai: You'd either need two channels or you'd still need a timer of some kind because it's three-part: we need to ensure that one goroutine gets to refreshing the cache before the other, then that it waits until the other is closed.

johanbrandhorst (approved):

jimlambrt (approved):
Ty!

johanbrandhorst (approved):
Amazing!

internal/clientcache/internal/cache/refresh_test.go
jefferai: Done!

PR #5066, opened 2024-09-03
Two fixes in our Go API code

jefferai:
* Replace `interface{}` with `any`
* For options that are bools, don't always set true; set to the actual value passed in (some options already behaved this way, but they all should)


jefferai:
Note: I plan to backport this to incorporate it into some other API work (mainly the `bool` consistency)

louisruch (approved):
Haha, love how big these gen file changes always look in change counts

johanbrandhorst (approved):

PR #5076, opened 2024-09-05
test(target): validate ipv6 address for tcp targets

ddebko:
A new set of tests were added to the create/update repository methods of the tcp target resources. These tests were added to validate that an IPv6 address can be associated directly to the tcp target.


kheina (approved):
LGTM

PR #5079, opened 2024-09-05
test(server): validate ipv6 address for upsert worker status

ddebko:
# Summary

This PR adds test to validate that an ipv4, ipv6, and an abbreviated ipv6 can be use to update the address field of a worker.


kheina (approved):
looks good!

PR #5083, opened 2024-09-06
internal/daemon/controller+worker: propagate downstream timeout

johanbrandhorst:
Propagate the downstream worker timeout into the downstream receiver and downstream ticker so we have a configurable consistent timeout value to use for worker to worker interactions.

Also change the atomic.Pointer to an atomic.Int64 for consistency with other values.


internal/daemon/worker/status_test.go
```diff
@@ -24,6 +24,7 @@ func TestWorkerWaitForNextSuccessfulStatusUpdate(t *testing.T) {
 	})
 	err := event.InitSysEventer(testLogger, testLock, "TestWorkerWaitForNextSuccessfulStatusUpdate", event.WithEventerConfig(testConfig))
 	require.NoError(t, err)
+	t.Cleanup(func() { event.TestResetSystEventer(t) })
```
johanbrandhorst, line 27: This fixes a test-to-test order dependency in this package

jefferai (approved):

PR #5084, opened 2024-09-06
Some test fixes

johanbrandhorst:
### [internal/session: update variable name in test](https://github.com/hashicorp/boundary/commit/23eeda7c393d2ce72f7892e588b1fc885634b108)

This was missed in https://github.com/hashicorp/boundary/pull/5071

### [internal/tests/cluster: skip failing tests](https://github.com/hashicorp/boundary/commit/d6f0887eadd773b94fd1928e04e93c0ad437ef9c)

These tests will need to be fixed or removed. It is not
clear why they are failing, but they are always failing.

### [internal/host/plugin: fix flaky test](https://github.com/hashicorp/boundary/pull/5084/commits/f980ca9f4ad98161c3bbcd421776dd807f8a2ee1)

Hosts contain IP addresses in string form, whose
order is not defined.

### DIFF
--- a/internal/tests/cluster/x509_verification_test.go
+++ b/internal/tests/cluster/x509_verification_test.go
@@ -10,6 +10,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"sync"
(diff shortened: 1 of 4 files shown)


internal/tests/cluster/x509_verification_test.go
```diff
@@ -185,13 +186,22 @@ func TestCustomX509Verification_Client(t *testing.T) {
...
 		Name:  "test",
+		Level: hclog.Trace,
 	})
 	require.NoError(t, event.InitSysEventer(logger, testLock, "use-TestCustomX509Verification_Server", event.WithEventerConfig(&ec.EventerConfig)))
+	t.Cleanup(func() { event.TestResetSystEventer(t) })
+	t.Cleanup(func() {
+		all, err := io.ReadAll(ec.AllEvents)
+		require.NoError(t, err)
+		t.Log(string(all))
+	})
```
moduli, line 204: Is the purpose of having multiple cleanups back-to-back is just to group like things together?

moduli (approved):

internal/tests/cluster/x509_verification_test.go
```diff
@@ -185,13 +186,22 @@ func TestCustomX509Verification_Client(t *testing.T) {
 }
 
 func TestCustomX509Verification_Server(t *testing.T) {
+	t.Skip("These tests are currently not working and will need further investigation")
```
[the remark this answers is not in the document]
johanbrandhorst: Yeah I raised and issue: https://hashicorp.atlassian.net/browse/ICU-15052. I think they were broken by 1.23 and something in the TLS stack, I had a quick look and couldn't see anything obvious in the release notes but it would be nice to fix and reenable them.

internal/tests/cluster/x509_verification_test.go
johanbrandhorst: I see them as distinct cleanups, one is reading the output of the events at the end of the test, and the other one is resetting the syseventer (the inverse of the line above it). Putting them in one cleanup is mixing concerns a bit IMO, but it's purely a matter of style, as long as the read runs before the resetter.

PR #5085, opened 2024-09-06
Some test fixes

johanbrandhorst:
### [internal/session: update variable name in test](https://github.com/hashicorp/boundary/pull/5085/commits/a118c6ce8df9411372512c3addd5e8b718ea1ae0)

This was missed in https://github.com/hashicorp/boundary/pull/5071

### [internal/host/plugin: fix flaky test](https://github.com/hashicorp/boundary/pull/5085/commits/cece7dbdc2428bbab8d250707316766cfaa587c1)

Hosts contain IP addresses in string form, whose
order is not defined.


mgaffney (approved):

internal/session/job_session_cleanup_test.go
```diff
@@ -95,7 +95,7 @@ func TestSessionConnectionCleanupJob(t *testing.T) {
 
 	// Create the job.
 	job, err := newSessionConnectionCleanupJob(ctx, rw, gracePeriod)
-	job.gracePeriod = gracePeriod // by-pass factory assert so we dont have to wait so long
+	job.workerStatusGracePeriod = gracePeriod // by-pass factory assert so we dont have to wait so long
```
mgaffney, line 98: This is also needed on the `main` branch

jefferai: I would think the other change (not swagger) is as well. Maybe make this against `main` and then backport?

jefferai (approved):

internal/session/job_session_cleanup_test.go
moduli: Looks like those changes are part of this PR to `main`: https://github.com/hashicorp/boundary/pull/5084.

PR #5087, opened 2024-09-08
fix (cache): support for DB debug logging

jimlambrt:


jefferai (approved):

internal/clientcache/internal/daemon/server.go
```diff
@@ -560,3 +568,17 @@ const (
 	dbFileName = "cache.db"
 	fkPragma   = "?_pragma=foreign_keys(1)"
 )
+
+type gormDebugLogger struct {
+	hclog.Logger
+}
+
+func (g gormDebugLogger) Printf(msg string, values ...interface{}) {
```
johanbrandhorst, line 576:
```suggestion
func (g gormDebugLogger) Printf(msg string, values ...any) {
```

internal/clientcache/internal/db/db.go
```diff
@@ -54,8 +54,11 @@ func Open(ctx context.Context, opt ...Option) (*db.DB, error) {
 	if err != nil {
 		return nil, errors.Wrap(ctx, err, op)
 	}
-	conn.Debug(opts.withDebug)
-
+	defer func() {
+		// let's not capture the output of resetSchema and createTables, so
+		// we'll defer turning on debug
+		conn.Debug(opts.withDebug)
+	}()
```
johanbrandhorst, line 61: Is this that noisy? This seems useful to leave enabled, and I could see it confusing some future dev that they enabled debug log but can't see the initial queries.

jimlambrt: It just doesn't seem to provide much value to see a wall of create this, that and the other db obj.  Typically when you're debugging, you're looking for the queries I think.

jimlambrt: In the future, it's pretty simple to discover where `createTable(...)` is called and to see the comment I left.

johanbrandhorst (approved):

PR #5088, opened 2024-09-09
ci: update security-scanner token

johanbrandhorst:
The old token had expired, this is the org-wide one we should be using.


picatz:
Superseded by https://github.com/hashicorp/boundary/pull/5089 instead.

PR #5097, opened 2024-09-11
(cmd): unhide client agent commands

irenarindos:


mikemountain (approved):

internal/cmd/commands/clientagentcmd/addtoken.go
```diff
@@ -48,10 +48,8 @@ func addToken(ctx context.Context, apiClient *api.Client, port uint) (*api.Respo
 	client.RetryWaitMin = 100 * time.Millisecond
 	client.RetryWaitMax = 1500 * time.Millisecond
 
-	// TODO (ICU-13140): Until we release the client agent, do not retry attempts
-	// to connect to the client agent since it adds a noticeably long delay to
-	// the command.
-	client.RetryMax = 0
+	// Explicitly setting this to 1, since this runs after every command and we don't want any delays
+	client.RetryMax = 1
```
johanbrandhorst, line 52: This retry would potentially add 1.5 seconds to every CLI command when ferry isn't running on the system, no? I wonder if we should leave this at 0?

internal/cmd/commands/clientagentcmd/wrapper_register.go
```diff
@@ -5,14 +5,23 @@ package clientagentcmd
...
+var allowErrorOutput = false
+
+const EnvBoundaryClientAgentCliErrorOutput = "BOUNDARY_CLIENT_AGENT_CLI_ERROR_OUTPUT"
+
+func init() {
+	errOutput := os.Getenv(EnvBoundaryClientAgentCliErrorOutput)
+	if strings.ToLower(errOutput) == "true" {
+		allowErrorOutput = true
+	}
+}
```
johanbrandhorst, line 24: It would be nice if we could pass this down through flags like other configuration values. What do you think?

internal/cmd/commands/clientagentcmd/addtoken.go
irenarindos: Sure, I'll boop it back to 0

internal/cmd/commands/clientagentcmd/wrapper_register.go
irenarindos: Sure, I'll do that. I'll open a PR for those changes

PR #5109, opened 2024-09-16
Worker Connection Failure Add Controller Target

RyanDerr:
# Description 
In the event that a worker fails to connect to a controller, update in the worker logs to provide the controllers target that the worker attempted to initialize connection to.

### DIFF
--- a/internal/daemon/worker/status.go
+++ b/internal/daemon/worker/status.go
@@ -230,7 +230,7 @@ func (w *Worker) sendWorkerStatus(cancelCtx context.Context, sessionManager sess
 		UpdateTags:                            w.updateTags.Load(),
 	})
 	if err != nil {
-		event.WriteError(cancelCtx, op, err, event.WithInfoMsg("error making status request to controller"))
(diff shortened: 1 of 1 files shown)


internal/daemon/worker/status.go
```diff
@@ -230,7 +230,7 @@ func (w *Worker) sendWorkerStatus(cancelCtx context.Context, sessionManager sess
 		UpdateTags:                            w.updateTags.Load(),
 	})
 	if err != nil {
-		event.WriteError(cancelCtx, op, err, event.WithInfoMsg("error making status request to controller"))
```
davidsbond, line 233: Since we call `.Load` already on line 185, is it possible to just assign that to a variable and use it in both places? Not sure what the load semantics are.

RyanDerr: I'll push up a change but an item to consider. Would this result in a thread safety/race condition issues?

irenarindos (approved):

PR #5110, opened 2024-09-16
test: ensure ipv6 compatibility for cli and config

ddebko:
# Summary

This PR ensures that the CLI commands inputed by the user can include field values that contain IPv6 addresses. This PR also ensures that the config file may contain field values that have IPv6 addresses.

### DIFF
--- a/internal/cmd/config/config.go
+++ b/internal/cmd/config/config.go
@@ -15,6 +15,7 @@ import (
 	"net"
 	"os"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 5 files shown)


internal/cmd/config/config.go
```diff
@@ -1278,6 +1285,16 @@ func (c *Config) SetupControllerPublicClusterAddress(flagValue string) error {
...
+	// net.JoinHostPort() will add a second pair of square brackets.
+	host = squareBrackets.ReplaceAllString(host, "")
+	if host == "" {
+		// return early if a empty host was provided
+		c.Controller.PublicClusterAddr = net.JoinHostPort(host, port)
+		return nil
+	}
+
 	c.Controller.PublicClusterAddr = net.JoinHostPort(host, port)
 	return nil
```
kheina, line 1299: I am curious what the purpose is of this return early clause if we return immediately afterwards

kheina (approved):
LGTM, just had one question

internal/cmd/config/config.go
ddebko: Thanks for catching this. My original implementation parsed the host value to check if it was a valid IPv4 or IPv6 address, but I had to remove this logic because it would break some unix listener and enos tests. This if conditional was an edge case to ensure that these tests passed as expected, but now I can remove this logic.

kheina (approved):
LGTM

PR #5111, opened 2024-09-16
Indexes and trigger changes to improve performance

tmessi:
This PR adds several performance related changes in the form of indexes,
and changes to a delete trigger to run more efficiently. These
improvements should help with several operations, notably:

- The grants query used to fetch grants for a user for controller API
  requests.
- Deletion of targets, auth tokens, and sessions.

See the individual commits for additional details.

### DIFF
--- a/internal/db/schema/migrations/oss/postgres/91/02_indexes_fk_delete.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/91/02_indexes_fk_delete.up.sql
@@ -1,0 +1,46 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+  -- Index to help when setting target_id to null on a session
+  -- when the corresponding target is deleted.
+  drop index if exists session_target_id_ix;
+  create index session_target_id_ix
+            on session (target_id);
+
+  -- Index to help when setting auth_token_id to null on a session
+  -- when the corresponding auth_token is deleted.
+  drop index if exists session_auth_token_id_ix;
+  create index session_auth_token_id_ix
+            on session (auth_token_id);
+
+  -- Index to help when setting session_id to null on a credential_vault_credential
+  -- when the corresponding session is deleted.
+  drop index if exists credential_vault_credential_session_id_ix;
+  create index credential_vault_credential_session_id_ix
+            on credential_vault_credential (session_id);
+
+  -- Index to help delete cascade of session_worker_protocol
+  -- when the corresponding session is deleted.
+  drop index if exists session_worker_protocol_session_id_ix;
+  create index session_worker_protocol_session_id_ix
+            on session_worker_protocol (session_id);
+
+  -- Index to help when setting session_id to null on recording_connection
+  -- when the corresponding session is deleted.
+  drop index if exists recording_connection_session_id_ix;
+  create index recording_connection_session_id_ix
+            on recording_connection (session_id);
+
+  -- Index to help delete of terminated sessions.
+  drop index if exists session_state_state_start_time_ix;
+  create index session_state_state_terminated_start_time_ix
+            on session_state (state, start_time)
+         where state = 'terminated';
+
+  analyze session,
+          credential_vault_credential,
+          session_worker_protocol,
+          recording_connection,
+          session_state;
+commit;
--- a/internal/db/schema/migrations/oss/postgres/91/03_indexes_grants_query.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/91/03_indexes_grants_query.up.sql
@@ -1,0 +1,39 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+  -- For each of these tables, swap the ordering of the
+  -- columns in the index for the primary key.
+  -- This helps the grants query that contains
+  -- several where clauses on what is currently the second
+  -- column in these indexes. By swapping the order, this
+  -- will make it more likely that the query planner will
+  -- choose to use the index.
+  -- See: https://www.postgresql.org/docs/current/indexes-multicolumn.html
+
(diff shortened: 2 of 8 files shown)


internal/db/schema/migrations/oss/postgres/91/03_indexes_grants_query.up.sql
```diff
@@ -0,0 +1,44 @@
...
+  -- several where clauses on what is currently the second
+  -- column in these indexes. By swapping the order, this
+  -- will make it more likely that the query planner will
+  -- choose to use the index.
+  -- See: https://www.postgresql.org/docs/current/indexes-multicolumn.html
+
+      alter table auth_oidc_managed_group_member_account
+  drop constraint auth_oidc_managed_group_member_account_pkey;
+      alter table auth_oidc_managed_group_member_account
+  add primary key (member_id, managed_group_id);
```
mgaffney, line 17: Can these be combined into one `alter table` statement?

mgaffney (approved):

internal/db/schema/migrations/oss/postgres/91/03_indexes_grants_query.up.sql
tmessi: Sure

mgaffney (approved):

internal/db/schema/migrations/oss/postgres/91/02_indexes_fk_delete.up.sql
```diff
@@ -0,0 +1,46 @@
...
+  create index recording_connection_session_id_ix
+            on recording_connection (session_id);
+
+  -- Index to help delete of terminated sessions.
+  drop index if exists session_state_state_start_time_ix;
+  create index session_state_state_terminated_start_time_ix
+            on session_state (state, start_time)
+         where state = 'terminated';
+
+  analyze session,
```
jefferai, line 41: Is it possible that this might take a long time to run? If we have users used to a rapid migration and have worked that into their upgrade methodology we may want to warn them about the fact that this could cause status updates to time out and thus not be able to keep connections on workers during the migration. (In future releases there will be `HUP` support to make this transition easier, but alas, not now.)

internal/db/schema/migrations/oss/postgres/91/03_indexes_grants_query.up.sql
```diff
@@ -0,0 +1,39 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+  -- For each of these tables, swap the ordering of the
```
jefferai, line 5: Do you want to just change the ordering of the where clauses in the grants query instead?

internal/db/schema/migrations/oss/postgres/91/05_deletion_tables_view.up.sql
```diff
@@ -0,0 +1,15 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+  -- Originially added in 81/01_deleted_tables_and_triggers.up.sql
```
jefferai, line 5: We usually note in the old migration where this was replaced

internal/db/schema/migrations/oss/postgres/91/05_deletion_tables_view.up.sql
```diff
@@ -0,0 +1,15 @@
...
+  -- This is being replaced with a view.
+  drop function get_deletion_tables;
+
+  create view deletion_table as
```
jefferai, line 9: For those of us that don't know how this works, would you mind adding some comments to explain it?

internal/db/schema/migrations/oss/postgres/91/02_indexes_fk_delete.up.sql
tmessi: Since the migrations are offline and we shouldn't have to deal with contention, I would not expect this to take too long, no longer than some other things we do in migrations like `alter table`.

internal/db/schema/migrations/oss/postgres/91/03_indexes_grants_query.up.sql
tmessi: I don't think we could change the grants query instead. In most cases the CTEs in the grants query have a where clause on a single column, and are selecting on the other column. Take for example the `user_oidc_managed_groups` cte:

```sql
 user_oidc_managed_groups (id) as (
   select managed_group_id
     from auth_oidc_managed_group_member_account
    where member_id in (select id from user_accounts)
 ),
```

The `where` clause is on `member_id`, and we are selecting `managed_group_id`. The existing index on `auth_oidc_managed_group_member_account` is on `(managed_group_id, member_id)`, which generally won't get used by a query with a where clause that only uses `member_id`. By swapping the index column order to `(member_id, managed_group_id)`, that grants query can now use the index.

internal/db/schema/migrations/oss/postgres/91/05_deletion_tables_view.up.sql
tmessi: Good catch, added a comment in the old migration.

internal/db/schema/migrations/oss/postgres/91/05_deletion_tables_view.up.sql
tmessi: I added some comments. I would also note that this is the same query that was used by the function, just now it is used in a view.

jefferai: You think I understood it when it was a function 😂

tmessi: Fair point. Do the comments help clarify things?

jefferai (approved):

PR #5138, opened 2024-09-27
fix(session): Fix query for orphaned connections

mgaffney:


internal/session/repository_connection.go
```diff
@@ -399,12 +399,13 @@ func (r *ConnectionRepository) closeOrphanedConnections(ctx context.Context, wor
 		notInClause = fmt.Sprintf(notInClause, strings.Join(params, ","))
 	}
 
+	query := fmt.Sprintf(orphanedConnectionsCte, notInClause)
```
louisruch, line 402: Unsolicited nit: should we rename this to something like closeOrphanedConnections since this does not appear to be a cte anymore

mgaffney: Fixed

internal/session/query.go
```diff
@@ -329,38 +329,17 @@ where
...
-	public_id in (select public_id from connections_to_close)
+   set closed_reason = 'system error'
+ where worker_id = @worker_id
+   and update_time < wt_sub_seconds_from_now(@worker_state_delay_seconds)
+   and (
+        connected_time_range is null
+        or
+        upper(connected_time_range) > now() 
+       )
+   %s
```
tmessi, line 342: I think we should see if we can avoid this `%s` by always including this last clause, but also switch it to an `any`, ie:

```sql
and not (public_id = any(@reported_connection_ids))
```

irenarindos (approved):

PR #5143, opened 2024-10-01
Bump boundary version & update changelog

irenarindos:


moduli (commented):
Is there anything we need to put in the change log?

PR #5149, opened 2024-10-03
test(e2e): Add user name to AWS resources

moduli:
This PR updates several enos modules to add the user's email address to AWS resources in order to make it easier to identify who generated them when looking at the AWS console. This is to assist when investigating who has left lingering resources that can be cleaned up when we are about to reach vpc limits or when costs are high.

Example:
Screenshot 2024-10-03 at 2 32 07 PM

https://hashicorp.atlassian.net/browse/ICU-15225


johanbrandhorst (approved):
🙏🏻 I see I'm not the only one annoyed by this 😂

PR #5151, opened 2024-10-07
Update controller.mdx - public_cluster_addr

vanphan24:
Updating public_cluster_addr to indicate that it is used if a load balancer is used for controllers. Also added a note about not including https:// and http:// in the address


irenarindos (approved):

stellarsquall (approved):
thank you! these changes look good to me

Dan-Heath (approved):
Thank you for this update Van and Irena!

PR #5165, opened 2024-10-09
internal/clientcache/cmd/cache: restart cache on version mismatch

johanbrandhorst:
Restart the client cache if it is found to be a version below the version of the CLI during client cache hook execution.


jefferai (approved):
I did not review the code in depth but hitting approve on the direction

PR #5173, opened 2024-10-12
feat (cache): impl. soft-delete of users in cache

jimlambrt:
Previously, when all the auth tokens for a user
where deleted, we deleted the user along with all
the cached resources for that user.  This can
create a lot of churn in the cache, especially if
the auth tokens have a short TTL (like 8 hrs), which 
can cause all the resources to be reloaded every
morning when the user logs in.

This change introduces a soft-delete of users in
the cache.  When all the auth tokens for a user are 
deleted, but there's still a valid refresh token
that's less than 20 days old, we mark the user as
deleted in the cache and do not return the user in 
the list of users returned by the cache, but we
still keep the user in the cache (all all their
cached resources). This way, when the user logs in 
again, we can just mark the user as active again
and not have to reload all the resources.  If
the refresh token is older than 20 days, we delete 
the user from the cache along with all their cached 
resources.

### DIFF
--- a/internal/clientcache/internal/cache/repository_token.go
+++ b/internal/clientcache/internal/cache/repository_token.go
@@ -75,7 +75,8 @@ func upsertUserAndAuthToken(ctx context.Context, reader db.Reader, writer db.Wri
 	}
 
 	var users []*user
-	if err := reader.SearchWhere(ctx, &users, "true", []any{}, db.WithLimit(-1)); err != nil {
+	// we only want users that have not been soft deleted
+	if err := reader.SearchWhere(ctx, &users, "true", []any{}, db.WithLimit(-1), db.WithTable(activeUserTableName)); err != nil {
 		return errors.Wrap(ctx, err, op)
 	}
 	if len(users) <= usersLimit {
@@ -382,6 +383,8 @@ func cleanExpiredOrOrphanedAuthTokens(ctx context.Context, writer db.Writer, idT
 	return nil
 }
 
+const activeUserTableName = "user_active" // users that have not been soft deleted
+
 // lookupUser returns a user if one is present in the repository or nil if not.
 func (r *Repository) lookupUser(ctx context.Context, id string) (*user, error) {
 	const op = "cache.(Repository).lookupUser"
@@ -390,7 +393,8 @@ func (r *Repository) lookupUser(ctx context.Context, id string) (*user, error) {
 		return nil, errors.New(ctx, errors.InvalidParameter, op, "empty id")
 	}
 	ret := &user{Id: id}
-	if err := r.rw.LookupById(ctx, ret); err != nil {
+	// we only want users that have NOT been soft deleted
+	if err := r.rw.LookupById(ctx, ret, db.WithTable(activeUserTableName)); err != nil {
 		if errors.IsNotFoundError(err) {
 			return nil, nil
 		}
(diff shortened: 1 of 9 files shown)


internal/clientcache/internal/db/db.go
```diff
@@ -146,7 +146,7 @@ type schema struct {
 
 const (
 	schemaTableName      = "schema_version"
-	schemaCurrentVersion = "v0.0.2"
+	schemaCurrentVersion = "v0.0.4"
```
johanbrandhorst, line 149: Do we need to bump this two patch versions?

internal/clientcache/internal/db/schema.sql
```diff
@@ -44,9 +44,17 @@ create table if not exists user (
     check (length(id) > 0),
   -- The address of the boundary instance that this user id comes from
   address text not null
-    check (length(address) > 0)
+    check (length(address) > 0),
+    -- deleted indicate if the user has been soft-deleted when the all
+    -- auth_tokens associated with the user are deleted.
+  deleted_at timestamp not null default 'infinity'
```
johanbrandhorst, line 50:
```suggestion
  -- deleted_at indicates when the user was soft-deleted because all
  -- auth_tokens associated with the user were deleted. It is set to 'infinity'
  -- for users that have not been soft-deleted.
  deleted_at timestamp not null default 'infinity'
```

internal/db/read_writer.go
```diff
@@ -471,14 +471,15 @@ func (rw *Db) IsTx(_ context.Context) bool {
...
+// must be unique. WithTable and WithDebug are the only valid options, all other
+// options are ignored.
 func (rw *Db) LookupById(ctx context.Context, resourceWithIder any, opt ...Option) error {
 	const op = "db.LookupById"
 	if rw.underlying == nil {
 		return errors.New(ctx, errors.InvalidParameter, op, "missing underlying db")
 	}
 	opts := GetOpts(opt...)
-	if err := dbw.New(rw.underlying.wrapped.Load()).LookupBy(ctx, resourceWithIder, dbw.WithDebug(opts.withDebug)); err != nil {
+	if err := dbw.New(rw.underlying.wrapped.Load()).LookupBy(ctx, resourceWithIder, dbw.WithDebug(opts.withDebug), dbw.WithTable(opts.withTable)); err != nil {
```
johanbrandhorst, line 482: Can we really set `dbw.WithTable` unconditionally? What happens if `opts.withTable` is unset?

johanbrandhorst (commented):
Looks great, some minor questions

internal/clientcache/internal/cache/repository_token.go
```diff
@@ -482,16 +487,31 @@ func deleteUser(ctx context.Context, w db.Writer, u *user) (int, error) {
...
+		// fallback to soft deleting the user
+		softDeleteStmt = "update user set deleted_at = (strftime('%Y-%m-%d %H:%M:%f','now')) where id = ?"
+	)
+	// see if we should delete the user
+	rowsAffected, err := w.Exec(ctx, deleteStmt, []any{u.Id})
+	switch {
+	case err != nil:
+		return db.NoRowsAffected, errors.Wrap(ctx, err, op)
+	case rowsAffected > 0:
+		// if we deleted the user, so we're done.
```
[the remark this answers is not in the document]
jimlambrt: ty

internal/clientcache/internal/cache/repository_token_test.go
```diff
@@ -650,6 +657,250 @@ func TestRepository_lookupUpser(t *testing.T) {
...
+	if d.IsInfinity {
+		return "infinity", nil
+	} else if d.IsNegInfinity {
+		return "-infinity", nil
+	}
+	return d.Time.Format(sqliteDatetimeLayout), nil
+}
+
+// testUser is used by testLookupUser to lookup a user from the database and
+// supports returing the user's DeletedAt time (soft delete).
```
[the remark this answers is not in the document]
jimlambrt: ty!

internal/clientcache/internal/db/db.go
jimlambrt: Probably not.  ty!

internal/clientcache/internal/db/schema.sql
jimlambrt: ty!

internal/db/read_writer.go
jimlambrt: it's a noop if it's unset.

johanbrandhorst (approved):

mgaffney (approved):

PR #5179, opened 2024-10-15
chore: update version and add Next to changelog post-release

irenarindos:


moduli (approved):
LGTM pending pipeline success

irenarindos:
> LGTM pending pipeline success

@moduli I'm adding a test flake for the test-race failure; I can't repro it locally and it's not obvious to me why that case failed

PR #5187, opened 2024-10-18
test(e2e): Add retry to search test

moduli:
Potentially with the implementation of https://github.com/hashicorp/boundary/pull/5173, the e2e test for search was impacted and started to fail on `main` (in CI only (?). I couldn't repro locally). 

```
 │ === RUN   TestCliSearch
│     search_test.go:40: Stopping daemon...
│     search_test.go:58: Waiting for daemon to start...
│     scope.go:85: Created Org Id: o_Mi5dwTRE8G
│     scope.go:127: Created Project Id: p_swelCH9fAZ
│     search_test.go:109: 
│         	Error Trace:	/src/boundary/testing/internal/e2e/tests/base/search_test.go:109
│         	Error:      	"[]" should have 1 item(s), but has 0
│         	Test:       	TestCliSearch
│ --- FAIL: TestCliSearch (6.15s)
```

This PR makes some modifications to the e2e test to perform some retries for that assertion.

### DIFF
--- a/testing/internal/e2e/tests/base/search_test.go
+++ b/testing/internal/e2e/tests/base/search_test.go
@@ -102,19 +102,42 @@ func TestCliSearch(t *testing.T) {
 	require.NoError(t, err)
 
 	// Get current number of targets
-	output = e2e.RunCommand(ctx, "boundary", e2e.WithArgs("cache", "status", "-format", "json"))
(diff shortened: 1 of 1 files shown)


johanbrandhorst (approved):
Thanks for looking into this!

PR #5189, opened 2024-10-21
feat(workers): add VerifyKnownWorkers repo function

irenarindos:
The repository function VerifyKnownWorkers replaces the use of ListWorkers in two instances where we are only concerned in confirming the existence of workers


hugoghx:
Q: Should we use an LLB for the entire effort?

irenarindos:
> Q: Should we use an LLB for the entire effort?

@hugoghx I spoke to @louisruch and this work is to go in separate from a branch so that it can be included in any releases we might have before 0.19.x; the other work we'll do will go in a branch

hugoghx (approved):

PR #5197, opened 2024-10-23
chore: update security scanner

irenarindos:
The security-scanner started failing with `no assets to download` (see https://github.com/hashicorp/boundary/actions/runs/11485270903/job/31967837536)


mikemountain (approved):

johanbrandhorst (approved):
I went and looked at https://github.com/github/codeql-action/releases to understand the change, I guess we need to remove any releases that aren't the bundle from consideration? If so, LGTM!

PR #5213, opened 2024-10-28
Clarify client agent config reload docs

johanbrandhorst:
### [docs/client-agent: clarify configuration reload on MacOS](https://github.com/hashicorp/boundary/commit/4ebdf3fa6b2420b2ad31b80ef4cbe1892a6794e6)

The previous configuration was known to confuse at least one reader.
It's easy to just copy the first command you see and run it, so
make that the non-destructive reload command.

### [docs/client-agent: correct log level config example](https://github.com/hashicorp/boundary/commit/d063628307ea80048210bf7d4ee57954d602e7d7)

### DIFF
--- a/website/content/docs/api-clients/client-agent.mdx
+++ b/website/content/docs/api-clients/client-agent.mdx
@@ -127,7 +127,7 @@ The configuration file contains the following fields:
 
    Example:
    ```hcl
-   log_level=DEBUG
+   log_level="DEBUG"
    ```
 
 - `log_to_stdout` - Logs to STDOUT in addition to the `boundary-client-agent.log` file.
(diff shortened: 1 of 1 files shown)


website/content/docs/api-clients/client-agent.mdx
```diff
@@ -172,20 +172,17 @@ By default, it is located in the following directory:
...
+   </Note>
 
-   ```shell-session
-   $ sudo launchctl stop com.hashicorp.boundary.boundary-client-agent
-   $ sudo launchctl start com.hashicorp.boundary.boundary-client-agent
-   ```
+1. Either reload the configuration file or restart the Client Agent.
 
-   Alternatively, you can update the following configuration values by reloading the configuration file, which will not disrupt any existing sessions:
+   The following configuration values can be changed by reloading the configuration file, which will not disrupt any existing sessions:
```
Dan-Heath, line 185:
```suggestion
   You can change the following configuration values by reloading the configuration file, which will not disrupt any existing sessions:
```

Dan-Heath: Suggest we make this active voice ("you can change" vs "can be changed") for clarity about who performs the action.

Dan-Heath (approved):
I added one suggested rewrite for active voice, but otherwise this looks great. Thanks Johan!

PR #5224, opened 2024-10-31
docs: Update meta content - Concepts

Dan-Heath:
Updates the page level meta content for all pages in the **Concepts** section in the docs.

### DIFF
--- a/website/content/docs/concepts/connection-workflows/exec-flag.mdx
+++ b/website/content/docs/concepts/connection-workflows/exec-flag.mdx
@@ -2,7 +2,7 @@
 layout: docs
 page_title: -exec flag
 description: >-
-  Learn how to use the `-exec` flag to execute TCP sessions or pass flags using your preferred client, even when there is no built-in support for it.
+  Learn how to use the `-exec` flag to execute TCP sessions or pass flags using your preferred client.
 ---
 
 # Exec flag
--- a/website/content/docs/concepts/connection-workflows/exec-flag.mdx
+++ b/website/content/docs/concepts/connection-workflows/exec-flag.mdx
@@ -2,7 +2,7 @@
 layout: docs
 page_title: -exec flag
 description: >-
-  Learn how to use the `-exec` flag to execute Boundary TCP sessions or pass flags using your preferred client, even when there is no built-in support for it.
+  Learn how to use the `-exec` flag to execute TCP sessions or pass flags using your preferred client, even when there is no built-in support for it.
 ---
 
 # Exec flag
--- a/website/content/docs/concepts/connection-workflows/exec-flag.mdx
+++ b/website/content/docs/concepts/connection-workflows/exec-flag.mdx
@@ -1,8 +1,8 @@
 ---
 layout: docs
 page_title: -exec flag
-description: |-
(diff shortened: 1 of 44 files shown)


website/content/docs/concepts/connection-workflows/exec-flag.mdx
```diff
@@ -1,8 +1,8 @@
 ---
 layout: docs
 page_title: -exec flag
-description: |-
-  Learn how the `-exec` flag enables you to execute Boundary TCP sessions using your preferred client, even when there is no built-in support for it.
+description: >-
+  Learn how to use the `-exec` flag to execute TCP sessions or pass flags using your preferred client, even when there is no built-in support for it.
```
stellarsquall, line 5: curious if you think this part is useful?

> even when there is no built-in support for it.

I'm curious how in-depth we should make the description, since it's never exposed to the reader. Will this statement improve SEO or discoverability?

website/content/docs/concepts/domain-model/credentials.mdx
```diff
@@ -1,8 +1,8 @@
 ---
 layout: docs
 page_title: Domain model - credentials
-description: |-
-  The anatomy of a Boundary credential
+description: >-
+  Use credentials to define a user's permissions on a host. Learn about username password, SSH private key, SSH certificate, and JSON credential types.
```
stellarsquall, line 5:
```suggestion
  Use credentials to define secrets for a host. Learn about username password, SSH private key, SSH certificate, and JSON credential types.
```

website/content/docs/concepts/domain-model/scopes.mdx
```diff
@@ -1,8 +1,8 @@
 ---
 layout: docs
 page_title: Domain model - scopes
-description: |-
-  Use scopes to group and manage resources in Boundary. Learn how to configure global scopes, org scopes, and projects to logically group resources.
+description: >-
+  Use scopes to group and manage resources. Learn how to configure global, org, and project scopes to logically group resources.
```
stellarsquall, line 5: I would consider minimizing the use of words that could be mistaken for other boundary resources, like group in this context

```suggestion
  Use scopes to organize and manage resources. Learn how to configure global, org, and project scopes to logically group resources.
```

Dan-Heath: Good point!

website/content/docs/concepts/connection-workflows/exec-flag.mdx
Dan-Heath: No, it probably isn't. I removed it. Thanks!

website/content/docs/concepts/filtering/events.mdx
```diff
@@ -1,8 +1,8 @@
 ---
 layout: docs
 page_title: Filtering - events
-description: |-
-  How to filter events emitted by Boundary.
+description: >-
+  Learn how to filter audit, observation, system, and telemetry events to easily find information written to any event sinks you configured.
```
stellarsquall, line 5: I usually avoid using words like "easily" or "simply"

```suggestion
  Learn how to filter audit, observation, system, and telemetry events to find information written to any event sinks you configured.
```

yhyakuna (approved):

PR #5228, opened 2024-11-06
feat(server): split worker tags table

irenarindos:
This PR splits the server_worker_tag table into two new tables- server_worker_api_tag and server_worker_config_tag
Config tags are only affected by UpsertWorkerStatus
SetWorkerTags, AddWorkerTags and DeleteWorkerTags only target API tags.

Note: Using migration folder 99 for this llb. When the llb is merged to main, we'll update it to the appropriate migration number. The aggregate view will also be dropped in favor of custom queries for the list and lookup worker functions.


internal/db/schema/migrations/oss/postgres/99/01_split_worker_tag_table.up.sql
```diff
@@ -0,0 +1,88 @@
...
+             (worker_id, key, value)
+      select worker_id, key, value
+        from server_worker_tag
+       where source = 'api';
+
+
+drop view server_worker_aggregate;
+-- Replaces view created in 86/01_server_worker_local_storage_state.up.sql to use the disparate worker tag tables
+-- View also switches to using json_agg to build the tags for consumption
+create view server_worker_aggregate as
```
louisruch, line 50: I assume the view is just temporary while we move to all the individual queries? If so perhaps a comment would be helpful

internal/server/repository_worker.go
```diff
@@ -431,21 +431,19 @@ func (r *Repository) UpsertWorkerStatus(ctx context.Context, worker *Worker, opt
...
 
-// setWorkerTags removes all existing tags from the same source and worker id
+// setWorkerConfigTags removes all existing config tags from the same source and worker id
 // and creates new ones based on the ones provided.  This function should be
 // called from inside a db transaction.
 // Workers/worker tags are intentionally not oplogged.
-func setWorkerTags(ctx context.Context, w db.Writer, id string, ts TagSource, tags []*Tag) error {
-	const op = "server.setWorkerTags"
+func setWorkerConfigTags(ctx context.Context, w db.Writer, id string, tags []*Tag) error {
+	const op = "server.setWorkerApiTags"
```
louisruch, line 439:
```suggestion
	const op = "server.setWorkerConfigTags"
```

internal/server/worker.go
```diff
@@ -115,8 +114,8 @@ type Worker struct {
 	*store.Worker
 
 	activeConnectionCount uint32 `gorm:"-"`
-	apiTags               []*Tag `gorm:"-"`
-	configTags            []*Tag `gorm:"-"`
+	apiTags               Tags
```
louisruch, line 117: I guess we currently never use the worker struct directly with `gorm` since its only being populated from the aggregate, but these will need to be exported when we do since you are dropping the gorm:"-" now we could do now or later just calling it out.

louisruch (commented):
Couple questions around the migration, but nothing major there. Otherwise everything here LGTM

internal/db/schema/migrations/oss/postgres/99/01_split_worker_tag_table.up.sql
irenarindos: it is temporary- I'll add a comment, thanks!

internal/server/worker.go
irenarindos: Thanks! I've got a PR pending on top of this that takes care of that

internal/server/repository_worker.go
irenarindos: thanks!

tmessi (approved):

internal/db/schema/migrations/oss/postgres/99/01_split_worker_tag_table.up.sql
```diff
@@ -0,0 +1,88 @@
...
+           cc.count as active_connection_count,
+           wt.tags as api_tags,
+           ct.tags as worker_config_tags
+      from server_worker w
+ left join (select worker_id, json_agg(json_build_object('key', key, 'value', value)) as tags from server_worker_api_tag group by worker_id) wt
+        on w.public_id = wt.worker_id
+ left join (select worker_id, json_agg(json_build_object('key', key, 'value', value)) as tags from server_worker_config_tag group by worker_id) ct
+        on w.public_id = ct.worker_id
+ left join connection_count as cc
+        on w.public_id = cc.worker_id;
```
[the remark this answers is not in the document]
irenarindos: We have a task to look at the query perf and compare it to Boundary without these changes; I'll be sure to share what we find with you!

PR #5230, opened 2024-11-08
feat(server): create query for list and lookup worker to use

irenarindos:
This PR creates a query for ListWorker and LookupWorker to use and directly populates the worker struct.

A future PR will remove the view the query replicates and remove the use of the workerAggregate struct entirely.

### DIFF
--- a/internal/server/query.go
+++ b/internal/server/query.go
@@ -4,6 +4,39 @@
 package server
 
 const (
+	listWorkersQuery = `
+	with connection_count (worker_id, count) as (
+    select worker_id,
+           count(1) as count
+      from session_connection
+      where closed_reason is null
+   group by worker_id
+)
+    select w.public_id,
+           w.scope_id,
+           w.description,
+           w.name,
+           w.address,
+           w.create_time,
+           w.update_time,
+           w.version,
+           w.last_status_time,
+           w.type,
+           w.release_version,
+           w.operational_state,
+           w.local_storage_state,
+           cc.count as active_connection_count,
+           wt.tags as api_tags,
+           ct.tags as config_tags
+      from server_worker w
+ left join (select worker_id, json_agg(json_build_object('key', key, 'value', value)) as tags from server_worker_api_tag group by worker_id) wt
+        on w.public_id = wt.worker_id
+ left join (select worker_id, json_agg(json_build_object('key', key, 'value', value)) as tags from server_worker_config_tag group by worker_id) ct
+        on w.public_id = ct.worker_id
+ left join connection_count as cc
+        on w.public_id = cc.worker_id
+	`
+
 	getStorageBucketCredentialStatesByWorkerId = `
 		select spsb.public_id as storage_bucket_id,
 			   wsbcs.permission_type, wsbcs.state, 
--- a/internal/server/repository_worker.go
+++ b/internal/server/repository_worker.go
@@ -5,6 +5,7 @@ package server
 
 import (
 	"context"
+	"database/sql"
 	stderrors "errors"
 	"fmt"
 	"strings"
@@ -117,20 +118,22 @@ func lookupWorker(ctx context.Context, reader db.Reader, id string) (*Worker, er
 	case id == "":
 		return nil, errors.New(ctx, errors.InvalidParameter, op, "id is empty")
 	}
-	wAgg := &workerAggregate{}
-	wAgg.PublicId = id
-	err := reader.LookupById(ctx, wAgg)
+
+	lookupQuery := fmt.Sprintf("%s where w.public_id = @worker_id", listWorkersQuery)
+
+	rows, err := reader.Query(ctx, lookupQuery, []any{sql.Named("worker_id", id)})
 	if err != nil {
-		if errors.IsNotFoundError(err) {
-			return nil, nil
+		return nil, errors.Wrap(ctx, err, op)
+	}
+	defer rows.Close()
+	for rows.Next() {
+		var worker Worker
+		if err := reader.ScanRows(context.Background(), rows, &worker); err != nil {
+			return nil, err
 		}
-		return nil, errors.Wrap(ctx, err, op)
+		return &worker, nil
 	}
-	w, err := wAgg.toWorker(ctx)
-	if err != nil {
-		return nil, errors.Wrap(ctx, err, op)
-	}
-	return w, nil
+	return nil, nil
 }
 
 // ListWorkers will return a listing of Workers and honor the WithLimit option.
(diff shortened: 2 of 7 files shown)


internal/server/worker.go
```diff
@@ -190,19 +189,19 @@ func (w *Worker) clone() *Worker {
 
 // ActiveConnectionCount is the current number of sessions this worker is handling
 // according to the controllers.
-func (w *Worker) ActiveConnectionCount() uint32 {
-	return w.activeConnectionCount
+func (w *Worker) GetActiveConnectionCount() uint32 {
+	return w.ActiveConnectionCount
```
tmessi, line 193: Do we still need this accessor method? Instead can the worker service just access ActiveConnectionCount directly?

internal/server/worker.go
```diff
@@ -190,19 +189,19 @@ func (w *Worker) clone() *Worker {
...
 	dedupedTags := make(map[Tag]struct{})
-	for _, t := range w.apiTags {
+	for _, t := range w.ApiTags {
 		dedupedTags[*t] = struct{}{}
 	}
-	for _, t := range w.configTags {
+	for _, t := range w.ConfigTags {
 		dedupedTags[*t] = struct{}{}
 	}
 	tags := make(map[string][]string)
```
tmessi, line 207: It might be nice to move this logic of converting from worker.Tags to a `map[string][]string` onto `worker.Tags`.

internal/server/worker.go
```diff
@@ -167,16 +166,16 @@ func (w *Worker) clone() *Worker {
...
 		Worker: cw.(*store.Worker),
 	}
-	if w.apiTags != nil {
-		cWorker.apiTags = make([]*Tag, 0, len(w.apiTags))
-		for _, t := range w.apiTags {
-			cWorker.apiTags = append(cWorker.apiTags, &Tag{Key: t.Key, Value: t.Value})
+	if w.ApiTags != nil {
+		cWorker.ApiTags = make([]*Tag, 0, len(w.ApiTags))
+		for _, t := range w.ApiTags {
+			cWorker.ApiTags = append(cWorker.ApiTags, &Tag{Key: t.Key, Value: t.Value})
```
tmessi, line 172: It might be nice to create a `clone` method on `worker.Tags` which could then be used here in both places where we are cloning Tags.

internal/server/query.go
```diff
@@ -4,6 +4,39 @@
...
+           cc.count as active_connection_count,
+           wt.tags as api_tags,
+           ct.tags as config_tags
+      from server_worker w
+ left join (select worker_id, json_agg(json_build_object('key', key, 'value', value)) as tags from server_worker_api_tag group by worker_id) wt
+        on w.public_id = wt.worker_id
+ left join (select worker_id, json_agg(json_build_object('key', key, 'value', value)) as tags from server_worker_config_tag group by worker_id) ct
+        on w.public_id = ct.worker_id
+ left join connection_count as cc
+        on w.public_id = cc.worker_id
```
tmessi, line 37:
```suggestion
	with connection_count (worker_id, count) as (
     select worker_id,
            count(1) as count
       from session_connection
      where closed_reason is null
   group by worker_id
)
    select w.public_id,
           w.scope_id,
           w.description,
           w.name,
           w.address,
           w.create_time,
           w.update_time,
           w.version,
           w.last_status_time,
           w.type,
           w.release_version,
           w.operational_state,
           w.local_storage_state,
           cc.count as active_connection_count,
           wt.tags as api_tags,
           ct.tags as config_tags
      from server_worker w
 left join (   select worker_id, 
                      json_agg(json_build_object('key', key, 'value', value)) as tags
                 from server_worker_api_tag
             group by worker_id) as wt
        on w.public_id = wt.worker_id
 left join (   select worker_id, 
                      json_agg(json_build_object('key', key, 'value', value)) as tags
                 from server_worker_config_tag
             group by worker_id) as ct
        on w.public_id = ct.worker_id
 left join connection_count as cc
        on w.public_id = cc.worker_id
```

internal/server/repository_worker.go
```diff
@@ -198,27 +204,30 @@ func ListWorkers(ctx context.Context, reader db.Reader, scopeIds []string, opt .
...
 	case KmsWorkerType, PkiWorkerType:
-		where = append(where, "type = ?")
-		whereArgs = append(whereArgs, opts.withWorkerType.String())
+		where = append(where, "w.type = @type")
+		whereArgs = append(whereArgs, sql.Named("type", opts.withWorkerType.String()))
 	default:
 		return nil, errors.New(ctx, errors.InvalidParameter, op, fmt.Sprintf("unknown worker type %v", opts.withWorkerType))
 	}
 
+	// Note: this option will be removed in a future PR for this llb
```
tmessi, line 221: Is there a reason to wait for a future PR? It seems like this is making changes that will just be dropped soon which just leads to more changes to review that might not even remain.

internal/server/repository_worker.go
```diff
@@ -117,20 +118,22 @@ func lookupWorker(ctx context.Context, reader db.Reader, id string) (*Worker, er
...
-		}
 		return nil, errors.Wrap(ctx, err, op)
 	}
-	w, err := wAgg.toWorker(ctx)
-	if err != nil {
-		return nil, errors.Wrap(ctx, err, op)
+	defer rows.Close()
+	for rows.Next() {
+		var worker Worker
+		if err := reader.ScanRows(context.Background(), rows, &worker); err != nil {
```
tmessi, line 131: Should this use `ctx` instead of creating a new context?

internal/server/repository_worker.go
```diff
@@ -117,20 +118,22 @@ func lookupWorker(ctx context.Context, reader db.Reader, id string) (*Worker, er
 	case id == "":
 		return nil, errors.New(ctx, errors.InvalidParameter, op, "id is empty")
 	}
-	wAgg := &workerAggregate{}
-	wAgg.PublicId = id
-	err := reader.LookupById(ctx, wAgg)
+
+	lookupQuery := fmt.Sprintf("%s where w.public_id = @worker_id", listWorkersQuery)
```
tmessi, line 122: I think we should make a complete `lookupQuery` const in `query.go` instead of using Sprintf here, even if that means duplicating some of the sql.

tmessi (changes requested):

internal/server/repository_worker.go
irenarindos: We have incoming work to create domain functions that replace usages of ListWorkers where these options are passed; we can't remove these options until those are created, and some of the refactoring to make those domain functions will be significant.

internal/server/worker.go
irenarindos: Good call, thanks!

PR #5250, opened 2024-11-18
docs: add ssh host key error to known issues

amitsk6:
host key error for man in the middle attack is commonly encountered while connecting ssh target, updated docs to help troubleshoot this issue.

### DIFF
--- a/website/content/docs/api-clients/client-agent.mdx
+++ b/website/content/docs/api-clients/client-agent.mdx
@@ -512,10 +512,10 @@ Follow the troubleshooting steps to understand why the Client Agent is not able 
 
 #### WARNING! Remote host indentification has changed! It is possible that someone is doing something nasty!
 
-This issue typically arises when a user resumes the client-agent after a pause and attempts to connect to a target. 
-The error message is a result of the client-agent detecting a change in the remote host's identity due to change in remote worker IP address.
+This issue arises when using SSH to connect to an SSH target through an alias after the first successful connection to that alias. The issue is due to the way Boundary workers generate a new host key on every new SSH connection. The warning can safely be ignored using the `StrictHostKeyChecking=no` command line option:
 
-HashiCorp advises removing the existing server host key from the `~/.ssh/known_hosts` file.
+```shell-session
+$ ssh -o StrictHostKeyChecking=no targetalias.boundary.dev
 
 ## Conflicting software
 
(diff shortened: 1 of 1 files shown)


website/content/docs/api-clients/client-agent.mdx
```diff
@@ -510,6 +510,13 @@ $ boundary client-agent pause
 
 Follow the troubleshooting steps to understand why the Client Agent is not able to reach the controller.
 
+#### WARNING! Remote host indentification has changed! It is possible that someone is doing something nasty!
+
+This issue typically arises when a user resumes the client-agent after a pause and attempts to connect to a target. 
+The error message is a result of the client-agent detecting a change in the remote host's identity due to change in remote worker IP address.
+
+HashiCorp advises removing the existing server host key from the `~/.ssh/known_hosts` file.
```
johanbrandhorst, line 518: It's probably worth having an entry for this here, but I would change the explanation a bit. What do you think about

```suggestion
#### WARNING! Remote host indentification has changed! It is possible that someone is doing something nasty!

This issue arises when using SSH to connect to an SSH target through an alias after the first successful connection to that alias. The issue is due to the way Boundary workers generate a new host key on every new SSH connection. The warning can safely be ignored using the `StrictHostKeyChecking=no` command line option:

```shell-session
$ ssh -o StrictHostKeyChecking=no targetalias.boundary.dev
```

You can also remove the existing server host key from the `~/.ssh/known_hosts` file to avoid the error.

website/content/docs/api-clients/client-agent.mdx
```diff
@@ -510,6 +510,16 @@ $ boundary client-agent pause
 
 Follow the troubleshooting steps to understand why the Client Agent is not able to reach the controller.
 
+#### WARNING! Remote host indentification has changed! It is possible that someone is doing something nasty!
+
+This issue arises when using SSH to connect to an SSH target through an alias after the first successful connection to that alias. The issue is due to the way Boundary workers generate a new host key on every new SSH connection. The warning can safely be ignored using the `StrictHostKeyChecking=no` command line option:
```
Dan-Heath, line 515:
```suggestion
This error arises when you use an alias to connect to an SSH target after the first successful connection using that alias. The issue occurs because Boundary workers generate a new host key on every new SSH connection. You can safely ignore the warning using the `StrictHostKeyChecking=no` command line option:
```

Dan-Heath (changes requested):
Thank you for this update @amitsk6 ! I suggested a revision to use active voice and simplify the first sentence a bit. If it sounds good to you, I will get this published.

website/content/docs/api-clients/client-agent.mdx
amitsk6: Looks good, thank you @Dan-Heath!

Dan-Heath (approved):
Thank you @amitsk6 !

PR #5262, opened 2024-11-19
Dockerfile: remove wget

johanbrandhorst:
wget shouldn't be required at runtime and the latest release contains an unfixed vulnerability (https://www.cve.org/CVERecord?id=CVE-2024-10524).


johanbrandhorst:
wget is sometime we include for health checks and document for our customers: https://developer.hashicorp.com/boundary/docs/operations/health#check-the-health-endpoint-using-wget

PR #5265, opened 2024-11-20
docs: clarify behavior of launchctl stop and unload for client agent

Mohit0233:
This pull request updates the macOS client agent documentation to clarify the behavior of `launchctl stop` and `launchctl unload`. It adds a note explaining that `launchctl stop` stops the agent temporarily, but the agent will automatically restart after 60 seconds due to the `StartInterval` setting. Additionally, the update provides information on using `launchctl unload` to permanently disable the client agent, in case users need to do so for specific reasons.

### DIFF
--- a/website/content/docs/api-clients/client-agent.mdx
+++ b/website/content/docs/api-clients/client-agent.mdx
@@ -277,7 +277,7 @@ If you want to disable the Boundary Client Agent, you can stop it with the follo
 $ sudo launchctl unload -w /Library/LaunchDaemons/com.hashicorp.boundary.boundary-client-agent.plist
 ```
 
-Unloading the Boundary client agent removes its launch daemon configuration. To restart the agent, use:
+Unloading the Boundary Client Agent removes its launch daemon configuration. To restart the Client Agent, use:
 
 ```shell-session
 $ sudo launchctl load -w /Library/LaunchDaemons/com.hashicorp.boundary.boundary-client-agent.plist
--- a/website/content/docs/api-clients/client-agent.mdx
+++ b/website/content/docs/api-clients/client-agent.mdx
@@ -274,14 +274,12 @@ If you want to disable the Boundary Client Agent, you can stop it with the follo
 <Tab heading="MacOS" group="macos">
 
 ```shell-session
-$ sudo launchctl stop com.hashicorp.boundary.boundary-client-agent
+$ sudo launchctl unload -w /Library/LaunchDaemons/com.hashicorp.boundary.boundary-client-agent.plist
 ```
 
 <Note>
 
-Stopping the Boundary client agent using `launchctl stop` only halts it temporarily, as the `StartInterval` setting causes it to restart automatically after 60 seconds.
-
-**To disable it permanently**, you must unload the service using `launchctl unload`. Unloading not only stops the agent but also removes its configuration from `launchd`. To restart the agent after unloading, you must reload it using `launchctl load`, as the `start` command will not work in this case.
+Unloading the Boundary client agent removes its launch daemon configuration. To restart the agent, use `sudo launchctl load ...`
 
 </Note>
 
--- a/website/content/docs/api-clients/client-agent.mdx
+++ b/website/content/docs/api-clients/client-agent.mdx
@@ -277,6 +277,14 @@ If you want to disable the Boundary Client Agent, you can stop it with the follo
 $ sudo launchctl stop com.hashicorp.boundary.boundary-client-agent
 ```
 
+<Note>
+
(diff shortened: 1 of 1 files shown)


website/content/docs/api-clients/client-agent.mdx
```diff
@@ -277,6 +277,14 @@ If you want to disable the Boundary Client Agent, you can stop it with the follo
 $ sudo launchctl stop com.hashicorp.boundary.boundary-client-agent
 ```
 
+<Note>
+
+Stopping the Boundary client agent using `launchctl stop` only halts it temporarily, as the `StartInterval` setting causes it to restart automatically after 60 seconds.
+
+**To disable it permanently**, you must unload the service using `launchctl unload`. Unloading not only stops the agent but also removes its configuration from `launchd`. To restart the agent after unloading, you must reload it using `launchctl load`, as the `start` command will not work in this case.
+
+</Note>
```
johanbrandhorst, line 286: Thanks for writing this, however, I think we should just include the `unload` command in the above shell session to avoid confusion. Is both `stop` and `unload` required, or just `unload`?

Mohit0233: Makes sense, adding the `unload` command directly will make the instructions clearer and avoid confusion.  

On the other hand, there might be a couple of minor points that could potentially cause confusion:  
1. Users familiar with `launchd` might use `stop`, not realizing it restarts after 60 seconds due to the `StartInterval` setting.  
2. Users who unload the service might not know they need to reload it with `launchctl load` to start it again, as this behavior isn’t explicitly mentioned.  

These might not be significant issues, but do you think it’s worth briefly addressing them?

johanbrandhorst: Yeah, I think we could have the shell command say how to unload it, and then have a note to say how to start it again. Do you mind adding the equivalent for Windows too?

Mohit0233: For Windows, there isn’t an equivalent to `StartInterval` like on macOS. Once the BoundaryClientAgent is stopped using `net stop BoundaryClientAgent`, it remains stopped permanently unless explicitly started again. Therefore, there’s no need to update anything specific for Windows. The `net stop BoundaryClientAgent` command works perfectly fine as expected.

website/content/docs/api-clients/client-agent.mdx
```diff
@@ -274,9 +274,15 @@ If you want to disable the Boundary Client Agent, you can stop it with the follo
 <Tab heading="MacOS" group="macos">
 
 ```shell-session
-$ sudo launchctl stop com.hashicorp.boundary.boundary-client-agent
+$ sudo launchctl unload -w /Library/LaunchDaemons/com.hashicorp.boundary.boundary-client-agent.plist
 ```
 
+<Note>
+
+Unloading the Boundary client agent removes its launch daemon configuration. To restart the agent, use `sudo launchctl load ...`
```
johanbrandhorst, line 282: I think we have enough space here to show the whole command, to make it easier for the user:

Unloading the Boundary client agent removes its launch daemon configuration. To restart the agent, use:

```shell-session
 $ sudo launchctl load -w /Library/LaunchDaemons/com.hashicorp.boundary.boundary-client-agent.plist
```

johanbrandhorst (changes requested):
Looks good, lets just make the reload command easier too

website/content/docs/api-clients/client-agent.mdx
Mohit0233: I initially tried wrapping the code snippet under the Note, but it overflowed and didn't display properly—I think the Note tag doesn't support wrapping code snippets. So I ellipsised it (...) to keep things concise. I also considered adding the entire command as an inline code snippet, but it looked too congested.

image

johanbrandhorst: Thank you for checking, that does indeed look very odd. Maybe we should remove the "Note" wrapper and just include the text and the command? That should still look ok, I think?

Mohit0233: I’ve updated the code as suggested. The "Note" wrapper has been removed, and the text now directly includes the explanation alongside the command.

johanbrandhorst (dismissed):

johanbrandhorst:
Thanks for the docs update, we'll be able to merge once @Dan-Heath has had a look.

Dan-Heath (approved):
Thanks for all your work on this @Mohit0233 and @johanbrandhorst ! It makes sense to me. I pushed an update to capitalize Client Agent, but otherwise it LGTM. I will get this published.

stellarsquall (approved):

PR #5277, opened 2024-11-22
test(e2e): Fix failure in migration test

moduli:
This PR addresses an e2e test failure that started appearing around October 14, 2024. The PR adjusts the test to wait for the controller to finish starting up before proceeding.  

https://hashicorp.atlassian.net/browse/ICU-15656

### DIFF
--- a/testing/internal/e2e/tests/database/migration_test.go
+++ b/testing/internal/e2e/tests/database/migration_test.go
@@ -181,16 +181,17 @@ func setupEnvironment(t testing.TB, c *config, boundaryRepo, boundaryTag string)
 
 	t.Log("Waiting for Boundary to finish loading...")
 	err = pool.Retry(func() error {
-		response, err := http.Get(b.UriLocalhost)
+		response, err := http.Get(fmt.Sprintf("%s/health", b.UriLocalhost))
 		if err != nil {
-			t.Logf("Could not access Boundary URL: %s. Retrying...", err.Error())
+			t.Logf("Could not access health endpoint: %s. Retrying...", err.Error())
 			return err
 		}
 
 		if response.StatusCode != http.StatusOK {
-			return fmt.Errorf("Could not connect to %s. Status Code: %d", b.UriLocalhost, response.StatusCode)
+			return fmt.Errorf("Health check returned an error. Status Code: %d", response.StatusCode)
 		}
 
+		response.Body.Close()
 		return nil
 	})
 	require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


testing/internal/e2e/tests/database/migration_test.go
```diff
@@ -180,17 +180,24 @@ func setupEnvironment(t testing.TB, c *config, boundaryRepo, boundaryTag string)
 	t.Logf("Using Boundary Version: %s", buf.String())
 
 	t.Log("Waiting for Boundary to finish loading...")
+	// Adding a manual sleep here to give the Boundary server time to start up.
+	// Sending an HTTP request to the controller too early can cause it to crash.
+	// This is a temporary solution until we can find a better way to determine
+	// when the server is ready.
+	time.Sleep(5 * time.Second)
```
moduli, line 187: ew. manual sleep, but I couldn't figure any other way. Ideas?

johanbrandhorst: Yeah this is going to be flaky or waste time. Honestly I think we should just wait until we can fix the issue in Boundary and leave this test failing. It is, after all, actually pointing out a bug in Boundary. We can still merge the other fixes if they are necessary but this sleep is just wrong 😬.

johanbrandhorst (changes requested):
Lets let this test fail, we should fix the bug soon.

testing/internal/e2e/tests/database/migration_test.go
moduli: Do you think a fix is something that can be done in the near-ish term? I'd like to try to squash these failures.

johanbrandhorst: I hope we will fix it in the next week. It's a DOS vulnerability, if small.

testing/internal/e2e/tests/database/migration_test.go
```diff
@@ -195,6 +195,22 @@ func setupEnvironment(t testing.TB, c *config, boundaryRepo, boundaryTag string)
...
+		}
+
+		if response.StatusCode != http.StatusOK {
+			return fmt.Errorf("Health check returned an error. Status Code: %d", response.StatusCode)
+		}
+
+		response.Body.Close()
+		return nil
+	})
+	require.NoError(t, err)
```
johanbrandhorst, line 212: Should we do this and the non-health one above or just one of them? I feel like we could just remove the first one?

moduli: Done

johanbrandhorst (approved):

PR #5284, opened 2024-11-25
docs: Update meta content - HCP and Enterprise

Dan-Heath:
Updates the page level meta content for all pages in the **HCP Boundary** and **Boundary Enterprise*** sections in the docs.

### DIFF
--- a/website/content/docs/enterprise/supported-versions.mdx
+++ b/website/content/docs/enterprise/supported-versions.mdx
@@ -2,7 +2,7 @@
 layout: docs
 page_title: Boundary Enterprise supported version policies
 description: >-
-  Learn about Enterprise support policies. Includes information about support periods, control plane and worker compatibility, and Postgres database versions.
+  Learn about Enterprise support policies. Includes information about support periods, control plane and worker compatibility, and PostgreSQL database versions.
 ---
 
 # Boundary Enterprise supported versions policy


website/content/docs/enterprise/licensing.mdx
```diff
@@ -2,7 +2,7 @@
 layout: docs
 page_title: License Boundary Enterprise
 description: >-
-  How to license Boundary Enterprise.
+  Learn how to enable Boundary Enterprise, once you have a valid license. Request access from an account team to enable enterprise features.
```
stellarsquall, line 5:
```suggestion
  Learn how to enable Boundary Enterprise with a valid license. Request access from an account team to enable enterprise features.
```

website/content/docs/enterprise/supported-versions.mdx
```diff
@@ -1,8 +1,8 @@
 ---
 layout: docs
-page_title: Boundary Enterprise supported versions
+page_title: Boundary Enterprise supported version policies
 description: >-
-  The supported versions policy for Boundary Enterprise. Includes information about support periods, control plane and worker compatibility, and Postgres database version support.
+  Learn about Enterprise support policies. Includes information about support periods, control plane and worker compatibility, and Postgres database versions.
```
stellarsquall, line 5: I think both names are acceptable here, just not sure which the tech writers are using? 

```suggestion
  Learn about Enterprise support policies. Includes information about support periods, control plane and worker compatibility, and PostgreSQL database versions.
```

stellarsquall (dismissed):

website/content/docs/enterprise/supported-versions.mdx
Dan-Heath: Good question, I don't think we have any kind of word list for product names like that. In the docs, I have been trying to use the full version you used in your suggestion `PostgreSQL`, but I missed it here.

stellarsquall (approved):

PR #5292, opened 2024-11-28
Suppress wget vulnerability in 0.18.x release

louisruch:


louisruch:
@johanbrandhorst I think you were correct that we should get this into the CE release branch. While it does seem like the validate oss check passes if we end up needing another CE release before 19 it may become a conflict that someone needs to resolve.

elimt (approved):

PR #5294, opened 2024-12-02
test(e2e): add GCP dynamic host catalog e2e tests

elimt:
- test(e2e): rename CreateAwsHostSetCli to CreatePluginHostSetCli: renaming to allow other plugins like GCP to invoke it for testing
- test(e2e): add GCP infra resources for enos testing:
This adds 2 new modules:
- `gcp_iam_setup`: This creates a GCP service account and assigns it with the required roles. All GCP auth uses the created service account
- `gcp_target`: creates a GCP VM instance with labels for filtering
- added new `gcp_project_id` variable which defines the project to use. This is only required for GCP testing

**Note**: This will fail until we have a GCP project we can use for CI testing. This PR will be open until we have that ready.


enos/enos-scenario-e2e-docker-base-with-gcp.hcl
```diff
@@ -0,0 +1,160 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: BUSL-1.1
+
+# For this scenario to work, add the following line to /etc/hosts
+# 127.0.0.1 localhost boundary
```
johanbrandhorst, line 5: This might just be copied from the other TF files, but is there any way we can make the scenario error with this error visible to the user instead of just adding a note here? I worry that this is sort of difficult to discover.

johanbrandhorst (commented):
The code looks great. I take it you've tested this manually against a project? What do you think about leaving this as a draft until we have a GCP project to test against, so we don't merge a failing tests accidentally?

enos/enos-scenario-e2e-docker-base-with-gcp.hcl
elimt: I also had this issue when testing locally :sweat_smile:. It took me some time to figure out . We could probably add something when we authenticate and cannot reach boundary. An message that could suggest this is the issue. Maybe in this function: https://github.com/hashicorp/boundary/blob/ae0ca0c5aadcf87b9d82c13bbb31327b66993c70/testing/internal/e2e/boundary/authenticate.go#L56

@moduli thoughts?

elimt:
@johanbrandhorst yeah I've tested it locally. yeah sure I am happy to leave it as draft for now

enos/enos-scenario-e2e-docker-base-with-gcp.hcl
moduli: Sure. We could put a message there if `output := e2e.RunCommand(ctx, "boundary", options...)` returns an error. Just want to make sure the error states that it's also possible that the address is no longer valid.

enos/modules/test_e2e_docker/main.tf
```diff
@@ -284,6 +343,16 @@ resource "enos_local_exec" "run_e2e_test" {
     E2E_LDAP_USER_NAME            = var.ldap_user_name
     E2E_LDAP_USER_PASSWORD        = var.ldap_user_password
     E2E_LDAP_GROUP_NAME           = var.ldap_group_name
+    E2E_GCP_PRIVATE_KEY_ID        = var.gcp_private_key_id
+    E2E_GCP_PRIVATE_KEY           = var.gcp_private_key
+    E2E_GCP_PROJECT_ID            = var.gcp_project_id
+    E2E_GCP_CLIENT_EMAIL          = var.gcp_client_email
+    E2E_GCP_REGION                = var.gcp_region
```
moduli, line 350: Was this used anywhere?

testing/internal/e2e/tests/gcp/dynamichostcatalog_host_set_empty_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	// Attempt to connect to target
+	output := e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"connect",
+			"-target-id", targetId,
+			"-format", "json",
+			"-exec", "/usr/bin/ssh", "--",
+			"-l", c.GcpTargetSshUser,
+			"-i", c.GcpTargetSshKey,
```
moduli, line 112: Does this work? Based on the other test, this contains the key contents? Do we need to create a file and specify the file here?

testing/internal/e2e/boundary/host.go
```diff
@@ -359,3 +359,53 @@ func WaitForNumberOfHostsInHostSetCli(t testing.TB, ctx context.Context, hostSet
 	)
 	require.NoError(t, err)
 }
+
+// CreateGCPHostCatalogCli uses the cli to create a new GCP dynamic host catalog.
+// Returns the id of the new host catalog.
+func CreateGCPHostCatalogCli(
```
moduli, line 365: I *think* the convention is Gcp?
```suggestion
func CreateGcpHostCatalogCli(
```

enos/modules/test_e2e_docker/main.tf
```diff
@@ -284,6 +343,16 @@ resource "enos_local_exec" "run_e2e_test" {
...
+    E2E_GCP_ZONE                  = var.gcp_zone
+    E2E_GCP_TARGET_SSH_KEY        = var.gcp_target_ssh_key
+    E2E_GCP_HOST_SET_FILTER1      = var.gcp_host_set_filter1
+    E2E_GCP_HOST_SET_FILTER2      = var.gcp_host_set_filter2
+    E2E_GCP_HOST_SET_IPS          = jsonencode(var.gcp_host_set_ips)
```
moduli, line 355: Another thing - not the cleanest thing, but you will also need to modify `enos/modules/test_e2e_docker/test_runner.sh` with these new variables.

enos/enos-variables.hcl
```diff
@@ -200,3 +200,33 @@ variable "hcp_boundary_cluster_id" {
...
+  description = "GCP project where the resources will be created"
+  type        = string
+  default     = ""
+}
+
+variable "gcp_client_email" {
+  description = "GCP client email"
+  type        = string
+  default     = ""
+}
```
moduli, line 232: Is this needed? It seems like it gets automatically set in `gcp_iam_setup`?

enos/modules/test_e2e_docker/main.tf
elimt: Do you mean all the variables or just `E2E_GCP_REGION`?

moduli: Just REGION

testing/internal/e2e/tests/gcp/dynamichostcatalog_host_set_empty_test.go
elimt: Actually I have to pass the key as a file. I'll fix that. The target connection should still. I'll fix it and test it again

enos/modules/test_e2e_docker/main.tf
elimt: It worked for me without that. When is it necessary?

moduli: In CI, we needed to run the tests from within a docker container due to some permission restrictions on github runners (you know that whole `/etc/hosts: 127.0.0.1 localhost boundary` thing -- we can't apply that change on the runners that we use). When you run it locally, we're not using this docker container unless you invoke this test_runner.sh script)

enos/enos-variables.hcl
elimt: This is required to set the client email variable here: https://github.com/hashicorp/boundary/blob/6ed1fab838e440242b169b626486ed3f1c86de37/enos/enos-scenario-e2e-docker-base-with-gcp.hcl#L151

Unless the enos tests do not work that way?

moduli: If I'm understanding right...

- the enos-variables.hcl defines some "global" variables that are can be populated by a user prior to kicking off enos (for example, by doing `export ENOS_VAR_my_variable=123`). This would allow you to use `var.my_variable` in a scenario file.
- For that line you referenced, you're populating the `gcp_client_email` input of one module with the output from the `gcp_iam_setup` module

testing/internal/e2e/tests/gcp/env_test.go
```diff
@@ -0,0 +1,32 @@
...
+	GcpProjectId      string `envconfig:"E2E_GCP_PROJECT_ID" required:"true"` // e.g. "my-project"
+	GcpClientEmail    string `envconfig:"E2E_GCP_CLIENT_EMAIL" required:"true"`
+	GcpHostSetFilter1 string `envconfig:"E2E_GCP_HOST_SET_FILTER1" required:"true"`
+	GcpHostSetFilter2 string `envconfig:"E2E_GCP_HOST_SET_FILTER2" required:"true"`
+	GcpHostSetIps     string `envconfig:"E2E_GCP_HOST_SET_IPS" required:"true"`
+	GcpTargetSshKey   string `envconfig:"E2E_GCP_TARGET_SSH_KEY" required:"true"`
+	GcpTargetAddress  string `envconfig:"E2E_TARGET_ADDRESS" required:"true"` // e.g. "192.168.0.1"
+	GcpTargetSshUser  string `envconfig:"E2E_SSH_USER" required:"true"`       // e.g. "ubuntu"
+	GcpTargetPort     string `envconfig:"E2E_TARGET_PORT" required:"true"`    // e.g. "22"
+	MaxPageSize       int    `envconfig:"E2E_MAX_PAGE_SIZE" default:"1000"`
```
moduli, line 21: Seems like this isn't used. Can probably remove

enos/enos.vars.hcl
```diff
@@ -60,6 +60,23 @@
...
+// The GCP private_key. This is used to authenticate with GCP. Only needed
+// if running GCP scenarios. This should not be used in combination with gcp_private_key_path.
+// gcp_private_key = ""
+
+// The GCP private_key_id. Only needed if running GCP scenarios. 
+// gcp_private_key_id = ""
+
+// The GCP client_email used to authenticate with GCP
+// gcp_client_email = "my-gcp-client-email"
+
```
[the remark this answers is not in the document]
moduli: Maybe the gcp_zone and gcp_region could live here, though. That's something that is potentially user configurable.

.github/workflows/enos-run.yml
```diff
@@ -101,6 +102,10 @@ jobs:
       ENOS_VAR_boundary_docker_image_name: ${{ inputs.docker-image-name }}
       ENOS_VAR_boundary_docker_image_file: ./support/boundary_docker_image.tar
       ENOS_VAR_go_version: ${{ inputs.go-version }}
+      ENOS_VAR_gcp_project_id: ${{ secrets.GCP_PROJECT_ID_CI }}
+      ENOS_VAR_gcp_client_email: ${{ secrets.GCP_CLIENT_EMAIL_CI }}
+      ENOS_VAR_gcp_private_key_id: ${{ secrets.GCP_PRIVATE_KEY_ID_CI }}
+      ENOS_VAR_gcp_private_key: ${{ secrets.GCP_PRIVATE_KEY_CI }}
```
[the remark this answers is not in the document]
elimt: Yes we do have the values for this. Once finalized in this PR, we can put it there too

enos/modules/test_e2e_docker/main.tf
```diff
@@ -222,6 +222,69 @@ variable "test_timeout" {
...
+  description = "GCP project where the resources will be created"
+  type        = string
+  default     = ""
+}
+
+variable "gcp_region" {
+  description = "GCP region where the resources will be created"
+  type        = string
+  default     = ""
+}
```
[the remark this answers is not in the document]
elimt: Yes it's not necessary. The GCP TF resources can just get the region from the provider if not passed in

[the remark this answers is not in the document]
elimt: I'll take it out

.github/workflows/enos-run.yml
```diff
@@ -147,6 +152,16 @@ jobs:
           role-to-assume: ${{ secrets.AWS_ROLE_ARN_CI }}
           role-skip-session-tagging: true
           role-duration-seconds: 3600
+      - name: Configure GCP credentials
+        id: gcp_auth
+        uses: google-github-actions/auth@55bd3a7c6e2ae7cf1877fd1ccb9d54c0503c457c # v2.1.2
+        with:
+          credentials_json: ${{ secrets.GCP_CREDENTIALS }}
+          access_token_lifetime: '3600s'
+          project_id: ${{ secrets.GCP_PROJECT_ID_CI }}
```
moduli, line 161: Not sure if we also want to add the if conditional here, too

enos/enos-scenario-e2e-docker-base-with-gcp.hcl
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  step "create_gcp_target" {
+    module = module.gcp_target
+
+    variables {
+      enos_user     = var.enos_user
+      instance_type = var.gcp_target_instance_type
+      gcp_zone      = var.gcp_zone
+      target_count  = 1
```
moduli, line 106: Question: In the AWS scenario, we created two targets with one filter and one target with another filter so as to verify that when you set a different filter on the host set, you can see that the number of hosts in that host set change. Do you that think that's worth it or just keep it as is?

Alternatively, we should this to `var.target_count`

moduli (commented):
Sorry, a couple of more minor comments

enos/enos-scenario-e2e-docker-base-with-gcp.hcl
elimt: I think that will be a good enhancement. For now I can set the target count using `var.target_count`

moduli: Let's chat during poddle

moduli (approved):

PR #5295, opened 2024-12-02
test(e2e): Update openssh module to account for image changes

moduli:
The linux openssh server docker container made some directory changes in a recent update. This broke some of our e2e tests. These changes assist in resolving those failures.

There will be a follow-up PR in boundary-enterprise.


enos/modules/docker_openssh_server_ca_key/custom-cont-init.d/01-allow-tcp-forwarding
moduli, line 1: This was a file in boundary-enterprise, but I think it makes sense to have it live here. The follow-up PR in boundary-enterprise will remove the `_ent` version of the file.

johanbrandhorst (approved):
Nice work!

PR #5303, opened 2024-12-03
chore: update go dependency

ddebko:


johanbrandhorst (approved):

johanbrandhorst (changes requested):
Hm, make gen is failing

johanbrandhorst (dismissed):

moduli (commented):
Dumb question... does this actually contain the change to the awssdk that we're looking for?

johanbrandhorst (approved):

PR #5304, opened 2024-12-03
internal/event: avoid reusing canceled context

johanbrandhorst:
Don't error in ReleaseGate if the context used for the original event has been cancelled. Any error in ReleaseGate causes the controller to terminate its startup. If a user was trying to send a request to the Boundary controller as it started up, the context tied to the request would be reused to attempt logging the observation associated with the request after the logging gate was released. This would always fail, as the context associated with the request was canceled.

We now use a new context timeout for events logged after the release gate is released, if the original context was canceled.

Fixes ICU-15809


moduli (approved):
Ran the failing e2e test and it seems like this has resolved that error

PR #5334, opened 2024-12-05
chore: Clarify GrantHash behavior when there are no GrantTuples

dkanney:
- Add clarifying comment to `GrantHash` --> if GrantTuples is nil, return a slice of empty bytes: `[]byte{0000 0000}`
- Enforce this behavior with additional test cases

### DIFF
--- a/internal/perms/grants.go
+++ b/internal/perms/grants.go
@@ -53,9 +53,10 @@ type GrantTuple struct {
 type GrantTuples []GrantTuple
 
 // GrantsHash returns a stable hash of all the grants in the GrantTuples.
+//
+// In the event that GrantTuples is nil, return a slice of empty bytes: []byte{0,0,0,0,0,0,0,0}
 func (g GrantTuples) GrantHash(ctx context.Context) ([]byte, error) {
 	const op = "perms.(GrantTuples).GrantHash"
-	// TODO: Should this return an error when the GrantTuples is empty?
 	var values []string
 	for _, grant := range g {
 		values = append(values, grant.Grant, grant.RoleId, grant.GrantScopeId)
--- a/internal/perms/grants.go
+++ b/internal/perms/grants.go
@@ -207,9 +207,7 @@ func (g Grant) clone() *Grant {
 	}
 	if outFields, hasSetFields := g.OutputFields.Fields(); hasSetFields {
 		fieldsToAdd := make([]string, 0, len(outFields))
-		for _, v := range outFields {
-			fieldsToAdd = append(fieldsToAdd, v)
-		}
+		fieldsToAdd = append(fieldsToAdd, outFields...)
 		ret.OutputFields = ret.OutputFields.AddFields(fieldsToAdd)
 	}
 	return ret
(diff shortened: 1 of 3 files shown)


johanbrandhorst (approved):
Thanks!

internal/daemon/controller/auth/auth.go
```diff
@@ -644,6 +647,9 @@ func (v verifier) performAuthCheck(ctx context.Context) (
 		aclResults.AuthenticationFinished = true
 		aclResults.Authorized = true
 		retErr = nil
+		grantTuples = []perms.GrantTuple{
+			{},
+		}
```
johanbrandhorst, line 652: Incredible, this must've been broken for almost a year (since we released pagination). I guess we never tried listing as the recovery user? Great spot!

internal/perms/grants.go
```diff
@@ -55,7 +55,11 @@ type GrantTuples []GrantTuple
 // GrantsHash returns a stable hash of all the grants in the GrantTuples.
 func (g GrantTuples) GrantHash(ctx context.Context) ([]byte, error) {
 	const op = "perms.(GrantTuples).GrantHash"
-	// TODO: Should this return an error when the GrantTuples is empty?
+
+	if len(g) == 0 {
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "empty GrantTuples receiver passed into GrantHash")
+	}
```
johanbrandhorst, line 61: So, currently, we return a slice of empty bytes when there are no grant tuples? I'm not sure if that is wrong, at least it's a unique mapping from the input to the output. As long as the slice isn't actually empty, that should keep working, right? What happens if we stop making this an error?

johanbrandhorst (commented):
This is a weird one, curious to hear what Tim thinks we should do here. Thanks for looking into it!

internal/perms/grants.go
dkanney: Yes

Technically there's a non-nil input of GrantTuples that'll hash to a slice of empty bytes (possible but very unlikely). So by also mapping nil GrantTuples to the empty bytes slice, this function would _technically_ no longer produce unique hashes

If we stop making this an error & return nil, there will be an error in some of the pagination functions (`List`, `ListPage`, etc) where we check for `len(grantsHash) == 0`. If we return the empty byte slice instead, things will work normally, but I guess there would be the possibility of some non-nil GrantTuple input hashing to an empty byte slice

johanbrandhorst: I personally think we're OK with the remote possibility of a valid grants combination hashing to 32 bytes of 0s as a collision in this case. I'm tempted to say we should keep mapping (no grants) -> (0000000). Does that sound OK?

internal/daemon/controller/auth/auth.go
dkanney: We test `ListSessions` with the recovery user but it worked due to `GrantHash` returning a slice of empty bytes

internal/perms/grants.go
dkanney: @johanbrandhorst That sounds good to me 👍

tmessi: My opinion on this is that a `nil` or empty `GrantsTuples` should still be usable. So calls to `Hash` should not result in an error in either case.

dkanney: Reverted back to normal & added a comment clarifying this behavior: [00550a0](https://github.com/hashicorp/boundary/pull/5334/commits/00550a064a070ee1db89ce9b0e8728e0205d1d1d)

johanbrandhorst (approved):
Sweet, thank you, I particularly like the comment additions!

PR #5344, opened 2024-12-10
chore(e2e): Update AWS service user with required permission

moduli:
This PR updates the terraform that manages the AWS service user used by CI with a new permission. This was needed after an update to the s3 bucket module that added an expiration date for objects in the bucket.

The change has already been applied to the service user and has resolved the failing e2e tests. This PR is just checking in those changes.


johanbrandhorst (approved):
Thanks!

PR #5360, opened 2024-12-12
cmd: propagate keyring reading errors

johanbrandhorst:
Reading a token from the keyring can sometimes fail for reasons other than the token not being present in the keyring. This PR introduces a new error used to indicate that the token requested could not be found in the keyring and propagates errors from the keyring up to the caller, so that the caller can act on them.

Additionally, this fixes a bug in the client cache where it would incorrectly assume that a token had been removed from the keyring, when in fact there was an error reading from the keyring.

### DIFF
--- a/internal/clientcache/internal/cache/repository_token.go
+++ b/internal/clientcache/internal/cache/repository_token.go
@@ -196,7 +196,10 @@ func (r *Repository) AddKeyringToken(ctx context.Context, bAddr string, token Ke
 		return errors.New(ctx, errors.InvalidParameter, op, "boundary address is empty", errors.WithoutEvent())
 	}
 	kt := token.clone()
-	keyringStoredAt := r.tokenKeyringFn(kt.KeyringType, kt.TokenName)
+	keyringStoredAt, err := r.tokenKeyringFn(kt.KeyringType, kt.TokenName)
+	if err != nil {
+		return errors.Wrap(ctx, err, op, errors.WithMsg("unable to lookup token in keyring, keyring type: %q, token name: %q", kt.KeyringType, kt.TokenName), errors.WithoutEvent())
+	}
 	if keyringStoredAt == nil {
 		return errors.New(ctx, errors.InvalidParameter, op, "unable to find token in the keyring specified", errors.WithoutEvent())
 	}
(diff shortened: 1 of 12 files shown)


moduli (commented):
I've ran the e2e tests a number of times and it seems to have resolved the failure.

moduli (approved):

internal/cmd/base/keyring.go
```diff
@@ -109,24 +112,26 @@ func (c *Command) DiscoverKeyringTokenInfo() (string, string, error) {
 	return keyringType, tokenName, nil
 }
 
-func ReadTokenFromKeyring(ui cli.Ui, keyringType, tokenName string) *authtokens.AuthToken {
+// ReadTokenFromKeyring will attempt to read the token with the name tokenName from the keyring
```
jimlambrt, line 115: Hum... should we consider writing a mock keyring provider (like the pkg does),
so we can write some unit tests?

internal/clientcache/internal/cache/repository_token.go
```diff
@@ -196,7 +196,10 @@ func (r *Repository) AddKeyringToken(ctx context.Context, bAddr string, token Ke
 		return errors.New(ctx, errors.InvalidParameter, op, "boundary address is empty", errors.WithoutEvent())
 	}
 	kt := token.clone()
-	keyringStoredAt := r.tokenKeyringFn(kt.KeyringType, kt.TokenName)
+	keyringStoredAt, err := r.tokenKeyringFn(kt.KeyringType, kt.TokenName)
+	if err != nil {
+		return errors.Wrap(ctx, err, op, errors.WithMsg("unable to lookup token in keyring, keyring type: %q, token name: %q", kt.KeyringType, kt.TokenName), errors.WithoutEvent())
```
jimlambrt, line 201: Not a big deal, but it appears to be straightforward to test these new bits.

jimlambrt (approved):
Looks great, just a few minor suggestions for unit tests.  Ty!

internal/cmd/base/keyring.go
johanbrandhorst: I think this file could do with a refactor to have `DiscoverKeyringTokenInfo` return a keyring provider interface or something, which we could pass around instead of the keyring type, but I think that change is outside the scope of this PR. I've created https://hashicorp.atlassian.net/browse/ICU-16114 to track a future effort to refactor and test this logic.

internal/clientcache/internal/cache/repository_token.go
johanbrandhorst: Added a quick test, thanks!

moduli (approved):

PR #5370, opened 2024-12-17
internal/daemon/worker: speed up multi-worker tests

johanbrandhorst:
Instead of sleeping, we can poll for the desired state. Also, run shutdowns in parallel. This mostly affects the enterprise tests that start multiple workers.

### DIFF
--- a/internal/daemon/worker/testing.go
+++ b/internal/daemon/worker/testing.go
@@ -456,9 +456,6 @@ func NewTestMultihopWorkers(t testing.TB,
 	})
 	t.Cleanup(kmsWorker.Shutdown)
 
-	// Give time for it to be inserted into the database
-	time.Sleep(2 * time.Second)
-
 	// names should not be set when using pki workers
 	pkiWorkerConf, err := config.DevWorker()
 	require.NoError(err)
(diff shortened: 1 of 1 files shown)


irenarindos (approved):

internal/daemon/worker/testing.go
```diff
@@ -548,16 +539,62 @@ func NewTestMultihopWorkers(t testing.TB,
...
-	}
+	require.NoError(err)
 
-	// Sleep so that workers can startup and connect.
-	time.Sleep(12 * time.Second)
+	t.Log("Waiting for workers to start up")
+	require.Eventually(
+		func() bool {
+			t.Log("Checking worker status")
+			workers, err := serversRepo.ListWorkers(controllerContext, []string{"global"})
```
moduli, line 552: Question: Is `NewTestMultihopWorkers` ever called by parallel tests? Could this impact the output of ListWorkers?

tmessi (approved):
Nice!

internal/daemon/worker/testing.go
johanbrandhorst: Not as long as each test is using it's own controller, which I think we have to assume the parallel tests are doing.

moduli (approved):

PR #5383, opened 2024-12-18
test(cluster): split tests to improve parallelism

ddebko:
# Summary

There are a number of tests in the cluster package that must run sequentially (due to global variables that are being manipulated), but there are also many tests that can run in parallel. The golang sdk does not allow us to configure how we want to run these tests in a specific order. Therefore, I split the tests into two sub packages. This has greatly improved the runtime of the tests, without introducing flaky behaviors.

Before:
Screenshot 2024-12-18 at 2 25 05 PM

After:
Screenshot 2024-12-18 at 2 24 42 PM

### DIFF
--- a/internal/daemon/controller/testing.go
+++ b/internal/daemon/controller/testing.go
@@ -529,7 +529,6 @@ func NewTestController(t testing.TB, opts *TestControllerOpts) *TestController {
 	var err error
 	tc.c, err = New(ctx, conf)
 	if err != nil {
-		tc.Shutdown()
 		t.Fatal(err)
 	}
 
@@ -552,7 +551,6 @@ func NewTestController(t testing.TB, opts *TestControllerOpts) *TestController {
 
 	if !opts.DisableAutoStart {
 		if err := tc.c.Start(); err != nil {
-			tc.Shutdown()
 			t.Fatal(err)
 		}
 	}
--- a/internal/tests/cluster/parallel/multi_controller_worker_test.go
+++ b/internal/tests/cluster/parallel/multi_controller_worker_test.go
@@ -1,10 +1,10 @@
 // Copyright (c) HashiCorp, Inc.
 // SPDX-License-Identifier: BUSL-1.1
 
-package cluster
+package parallel
 
 import (
-	"context"
+	"slices"
 	"sync"
 	"testing"
 	"time"
@@ -14,12 +14,11 @@ import (
 	"github.com/hashicorp/boundary/internal/daemon/worker"
 	"github.com/hashicorp/boundary/internal/tests/helper"
 	"github.com/hashicorp/go-hclog"
-	"github.com/hashicorp/go-secure-stdlib/strutil"
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 func TestMultiControllerMultiWorkerConnections(t *testing.T) {
+	t.Parallel()
 	require := require.New(t)
 	logger := hclog.New(&hclog.LoggerOptions{
 		Level: hclog.Trace,
@@ -32,12 +31,10 @@ func TestMultiControllerMultiWorkerConnections(t *testing.T) {
 		Config: conf,
 		Logger: logger.Named("c1"),
 	})
-	defer c1.Shutdown()
 
 	c2 := c1.AddClusterControllerMember(t, &controller.TestControllerOpts{
 		Logger: c1.Config().Logger.ResetNamed("c2"),
 	})
-	defer c2.Shutdown()
 
 	wg := new(sync.WaitGroup)
 	wg.Add(2)
@@ -56,7 +53,6 @@ func TestMultiControllerMultiWorkerConnections(t *testing.T) {
 		InitialUpstreams: append(c1.ClusterAddrs(), c2.ClusterAddrs()...),
 		Logger:           logger.Named("w1"),
 	})
-	defer w1.Shutdown()
 
 	wg.Add(2)
 	go func() {
@@ -72,7 +68,6 @@ func TestMultiControllerMultiWorkerConnections(t *testing.T) {
 	w2 := w1.AddClusterWorkerMember(t, &worker.TestWorkerOpts{
 		Logger: logger.Named("w2"),
 	})
-	defer w2.Shutdown()
 
 	wg.Add(2)
 	go func() {
@@ -98,53 +93,46 @@ func TestMultiControllerMultiWorkerConnections(t *testing.T) {
 	}()
 	wg.Wait()
 
-	w1 = worker.NewTestWorker(t, &worker.TestWorkerOpts{
+	w3 := worker.NewTestWorker(t, &worker.TestWorkerOpts{
 		WorkerAuthKms:    c1.Config().WorkerAuthKms,
 		InitialUpstreams: c1.ClusterAddrs(),
 		Logger:           logger.Named("w1"),
 	})
-	defer w1.Shutdown()
 
 	wg.Add(2)
 	go func() {
 		defer wg.Done()
-		helper.ExpectWorkers(t, c1, w1, w2)
+		helper.ExpectWorkers(t, c1, w2, w3)
 	}()
 	go func() {
 		defer wg.Done()
-		helper.ExpectWorkers(t, c2, w1, w2)
+		helper.ExpectWorkers(t, c2, w2, w3)
 	}()
 	wg.Wait()
 
 	require.NoError(c2.Controller().Shutdown())
 
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
-		helper.ExpectWorkers(t, c1, w1, w2)
-	}()
-	wg.Wait()
+	helper.ExpectWorkers(t, c1, w2, w3)
 
-	c2 = c1.AddClusterControllerMember(t, &controller.TestControllerOpts{
-		Logger: c1.Config().Logger.ResetNamed("c2"),
+	c3 := c1.AddClusterControllerMember(t, &controller.TestControllerOpts{
+		Logger: c1.Config().Logger.ResetNamed("c3"),
 	})
-	defer c2.Shutdown()
 
 	wg.Add(2)
 	go func() {
 		defer wg.Done()
-		helper.ExpectWorkers(t, c1, w1, w2)
+		helper.ExpectWorkers(t, c1, w2, w3)
 	}()
 	go func() {
 		defer wg.Done()
-		helper.ExpectWorkers(t, c2, w1, w2)
+		helper.ExpectWorkers(t, c3, w2, w3)
 	}()
 	wg.Wait()
 }
 
 func TestWorkerAppendInitialUpstreams(t *testing.T) {
-	ctx := context.Background()
-	require, assert := require.New(t), assert.New(t)
+	t.Parallel()
+	require := require.New(t)
 	logger := hclog.New(&hclog.LoggerOptions{
 		Level: hclog.Trace,
 	})
@@ -156,7 +144,7 @@ func TestWorkerAppendInitialUpstreams(t *testing.T) {
 		Config: conf,
 		Logger: logger.Named("c1"),
 	})
-	defer c1.Shutdown()
+	t.Cleanup(c1.Shutdown)
 
 	helper.ExpectWorkers(t, c1)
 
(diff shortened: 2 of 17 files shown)


internal/tests/cluster/sequential/session_cleanup_test.go
```diff
@@ -117,11 +118,6 @@ func testWorkerSessionCleanupSingle(burdenCase timeoutBurdenType) func(t *testin
 			Logger:                              logger.Named("w1"),
 			SuccessfulStatusGracePeriodDuration: workerGracePeriod(burdenCase),
 		})
-
-		err = w1.Worker().WaitForNextSuccessfulStatusUpdate()
-		require.NoError(err)
-		err = c1.WaitForNextWorkerStatusUpdate(w1.Name())
-		require.NoError(err)
```
johanbrandhorst, line 124: Was this removed because `ExpectWorkers` performs the same checks?

internal/tests/cluster/sequential/session_cleanup_test.go
```diff
@@ -140,7 +136,9 @@ func testWorkerSessionCleanupSingle(burdenCase timeoutBurdenType) func(t *testin
 		// Create test server, update default port on target
 		ts := helper.NewTestTcpServer(t)
 		require.NotNil(t, ts)
-		t.Cleanup(ts.Close)
+		t.Cleanup(func() {
+			go ts.Close()
```
johanbrandhorst, line 140: Is this safe?

internal/tests/cluster/sequential/session_cleanup_test.go
```diff
@@ -163,21 +161,22 @@ func testWorkerSessionCleanupSingle(burdenCase timeoutBurdenType) func(t *testin
...
 			sess.ExpectConnectionStateOnController(ctx, t, c1.Controller().ConnectionRepoFn, session.StatusConnected)
-
 		default:
 			// Should be closed on both worker and controller. Wait on
 			// worker then check controller.
 			sess.ExpectConnectionStateOnWorker(ctx, t, w1, session.StatusClosed)
 			sess.ExpectConnectionStateOnController(ctx, t, c1.Controller().ConnectionRepoFn, session.StatusClosed)
 		}
 
+		time.Sleep(15 * time.Second)
```
johanbrandhorst, line 171: Why is this necessary?

internal/tests/cluster/sequential/session_cleanup_test.go
```diff
@@ -237,8 +236,18 @@ func testWorkerSessionCleanupMulti(burdenCase timeoutBurdenType) func(t *testing
...
+		wg.Add(2)
+		go func() {
+			defer wg.Done()
+			helper.ExpectWorkers(t, c1)
+		}()
+		go func() {
+			defer wg.Done()
+			helper.ExpectWorkers(t, c2)
+		}()
+		wg.Wait()
```
johanbrandhorst, line 250: _nice_

internal/tests/cluster/sequential/session_cleanup_test.go
```diff
@@ -366,8 +373,17 @@ func testWorkerSessionCleanupMulti(burdenCase timeoutBurdenType) func(t *testing
...
+		wg.Add(2)
+		go func() {
+			defer wg.Done()
+			helper.ExpectWorkers(t, c1, w1)
+		}()
+		go func() {
+			defer wg.Done()
+			helper.ExpectWorkers(t, c2, w1)
+		}()
+		wg.Wait()
```
johanbrandhorst, line 386: Is this equivalent to the old code?

internal/tests/helper/testing_helper.go
```diff
@@ -435,7 +435,12 @@ func NewTestTcpServer(t *testing.T) *TestTcpServer {
 	ts.ln, err = net.Listen("tcp", ":0")
 	require.NoError(err)
 
-	go ts.run()
+	serverStarted := make(chan struct{})
+	go func() {
+		close(serverStarted)
+		ts.run()
+	}()
+	<-serverStarted
```
johanbrandhorst, line 443: Is this really useful? The serverStarted channel really only guarantees that the goroutine has started, which shouldn't give us much more confidence about the server starting. Can't we just try sending a TCP packet or something on the listener in a loop (e.g. `c := net.Dial(); c.Write(), c.Read()`) to see whether the server is ready?

internal/tests/helper/testing_helper.go
```diff
@@ -444,18 +449,25 @@ func NewTestTcpServer(t *testing.T) *TestTcpServer {
...
+			if workerMap[k.(string)] == nil {
+				// We don't remove from updateTimes currently so if we're not
+				// expecting it we'll see an out-of-date entry
+				return true
+			}
+			assert.WithinDuration(t, time.Now(), v.(time.Time), 30*time.Second)
+			delete(workerMap, k.(string))
+			return true
+		})
+		assert.Empty(t, workerMap)
```
johanbrandhorst, line 470: I don't understand the use of this logic. It seems to populate the workerMap with no entries, since `len(workers) == 0`, then range over all the update times and return true for every entry (since workerMap is empty), then assert that workerMap is empty, which of course it is because we never added anything to it? I must be missing something?

internal/tests/helper/testing_helper.go
```diff
@@ -444,18 +449,25 @@ func NewTestTcpServer(t *testing.T) *TestTcpServer {
 // this method will sleep for 10 seconds and then validate that the controller worker status
 // is empty.
```
johanbrandhorst, line 450: Looks like this comment needs to be updated?

johanbrandhorst (changes requested):
Super excited about this change, but I'm a little nervous about running cleanup in goroutines which we are not waiting to finish. I don't think that's safe, generally.

internal/daemon/controller/testing.go
```diff
@@ -523,13 +523,15 @@ func NewTestController(t testing.TB, opts *TestControllerOpts) *TestController {
 		shutdownDoneCh: make(chan struct{}),
 		shutdownOnce:   new(sync.Once),
 	}
-	t.Cleanup(tc.Shutdown)
+	t.Cleanup(func() {
+		go tc.Shutdown()
```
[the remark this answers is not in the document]
ddebko: I believe there are a couple of tests that focus on validating the Shutdown function does not leak resources. For those specific tests, I believe that we should not utilize a cleanup function or a go routine when calling the Shutdown method. If those unit tests pass, I think we should feel comfortable with shutting down a worker/controller with go routines for unit tests that do not necessarily care about what the Shutdown method is doing.

Since the Shutdown method doesn't return an error, I feel like there is nothing to check against.

internal/tests/helper/testing_helper.go
ddebko: That is a great call out.

internal/daemon/controller/testing.go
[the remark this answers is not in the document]
johanbrandhorst: I'm not convinced about this, sure, most of the time this probably works fine because the test process is still running while the cleanup is happening, but I really think we should at the very least ensure `TestMain` does not exist until all controllers have shut down successfully. I think I'd prefer if this PR did not run all these functions in goroutines, and we make any such changes in a separate PR. I could see us having something like a global waitgroup that we could add to and that `TestMain` could wait for before returning. Even that I'm not sure about, since `tc.Shutdown` can call `t.Error`, which is not allowed after the test has returned. Does that sound okay?

internal/tests/cluster/sequential/session_cleanup_test.go
ddebko: Thanks for pointing that out, I actually wanted to bring this issue up. Below this, there is a command `sConn.TestSendRecvFail(t)` that will randomly fail because the error that gets returned is not a net.ErrClosed type, but a connection reset error.

I'm not sure why it's flaky and I believe this behavior is consistent on main. I feel like adding this sleep command has reduced the chances of the test failing, but I could be wrong.

ddebko: I was hoping that someone might have an idea of what we can poll instead of using Sleep.

johanbrandhorst: Looks to me like we could actually poll on c.testSendRecv(t), but we'll need a new helper function for that. You could copy TestSendRecvFail and modify it to eventually fail instead of immediately fail. Something like `TestSendRecvEventuallyFail`, maybe take a timeout as a parameter.

internal/daemon/controller/testing.go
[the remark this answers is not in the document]
ddebko: I have an idea, but I'm not sure if it's worth the amount of complexity added for the benefit of improving how fast the tests run.

We could use a `TestMain(m *testing.M){}` to ensure that all the go routines in the cleanup functions finish running once the unit tests are done.

Example:

```
func TestMain(t *testing.M) {
  // setup a global variable for a sync.WaitGroup
  t.Run() // this will run the unit tests in the package
  // now we can call Wait() on the global waitgroup.
}
```

But at this point, we probably won't really get much of a performance improvement. Also this could cause more issues with tests timing out if people forget to call Done() on the waitgroup.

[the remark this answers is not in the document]
johanbrandhorst: Yeah that was my thinking too, but I also think it risks us missing test failures when there are errors in shutdown. So for this PR I'd prefer if we don't do the shutdowns in goroutines and ensure they all conclude before the test that spawned them. We can consider making shutdown changes in another PR. Does that make sense?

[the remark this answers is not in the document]
ddebko: Yea that makes sense to me, I will start reverting the changes to exclude the go routine calls for Shutdown.

internal/tests/cluster/parallel/multi_controller_worker_test.go
```diff
@@ -32,12 +31,16 @@ func TestMultiControllerMultiWorkerConnections(t *testing.T) {
 		Config: conf,
 		Logger: logger.Named("c1"),
 	})
-	defer c1.Shutdown()
+	defer func() {
+		go c1.Shutdown()
+	}()
```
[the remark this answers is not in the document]
ddebko: I think I will remove all of these duplicate calls because they already exist inside of the NewTestController() and NewTestWorker() methods.

internal/tests/cluster/sequential/session_cleanup_test.go
```diff
@@ -107,7 +108,7 @@ func testWorkerSessionCleanupSingle(burdenCase timeoutBurdenType) func(t *testin
 		)
 		require.NoError(err)
 		t.Cleanup(func() {
-			proxy.Close()
+			go proxy.Close()
```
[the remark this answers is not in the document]
ddebko: I will revert all the changes that added go routines to the Close() methods.

internal/tests/cluster/sequential/session_cleanup_test.go
ddebko: Yes, that is exactly why I removed these lines.

internal/tests/cluster/sequential/session_cleanup_test.go
ddebko: Reverted

internal/tests/cluster/sequential/session_cleanup_test.go
ddebko: I think the old code was flaky. I believe that we actually want to verify that both controller 1 & 2 are able to connect to the worker. The reason why I think this is the expected behavior is because both proxy 1 and 2 were resumed at the same time.

internal/tests/helper/testing_helper.go
ddebko: I think we should move away from the original implementation, specifically using the sleep command and clearing the worker status update times map.

I feel like we should create a different helper function called `RemoveExpectedWorkers`. In this function, if no workers are provided then we simply just check if the WorkerStatusUpdateTime is empty. If workers are provided, we invoke the `WaitForNextSuccessfulStatusUpdate` method for each worker and then check if the WorkerStatusUpdateTime is empty. This should be a more accurate way to validate.

In the original `ExpectWorkers`, we will fail the test is the worker list is empty.

johanbrandhorst: The old code is a little confusing, but it makes sense to me that `ExpectWorkers` with no workers would assert that there are no active workers. What does RemoveExpectedWorkers do? It can't just check if `WorkerStatusUpdateTime` is empty, because it'll contain updates from workers that were previously active. It could clear the map and check for workers, which is what the existing code does. I think the existing code relying on a sleep isn't great, so it'd be good to change that, but do we need another function?

ddebko:
Work was moved to this PR:
https://github.com/hashicorp/boundary/pull/5390

PR #5384, opened 2024-12-19
test(e2e): Add test for vault oidc provider

moduli:
This PR adds an e2e test that validates setting up [Vault as an OIDC provider](https://developer.hashicorp.com/vault/tutorials/auth-methods/oidc-identity-provider#create-a-vault-oidc-provider) and creating a Boundary auth method for it. This test does not authenticate using this auth method as it requires access to a browser to complete authentication. I plan on exploring an Admin UI e2e test to see if it's feasible to do there.

```
=== RUN   TestAuthMethodOidcVault
    auth_method_oidc_vault_test.go:29: Setting up Vault OIDC provider...
    scope.go:85: Created Org Id: o_QTNEIIlw7l
    auth_method_oidc_vault_test.go:305: Created OIDC Auth Method: amoidc_M9H2YNLS5Y
--- PASS: TestAuthMethodOidcVault (17.41s)
PASS
ok      github.com/hashicorp/boundary/testing/internal/e2e/tests/base_with_vault        18.469s
```

Additionally, some minor changes were done to the networking of the vault docker container. To set up the provider correctly, we needed a vault address such that entities both inside and outside the docker network could access so that redirects could work appropriately). The change moved vault to attach to local port 8300 so that it could be referenced using the same address both inside and outside the docker network (http://vault:8300)

https://hashicorp.atlassian.net/browse/ICU-14377

### DIFF
--- a/enos/modules/docker_vault/main.tf
+++ b/enos/modules/docker_vault/main.tf
@@ -37,6 +37,11 @@ variable "vault_port" {
   type        = string
   default     = "8300"
 }
+variable "vault_port_internal" {
+  description = "External Port to use"
+  type        = string
+  default     = "8300"
+}
 
 resource "docker_image" "vault" {
   name         = var.image_name
@@ -48,7 +53,7 @@ resource "docker_container" "vault" {
   name    = var.container_name
   command = ["vault", "server", "-config", "/vault/config.d/config.json"]
   ports {
-    internal = 8200
+    internal = var.vault_port_internal
     external = var.vault_port
   }
   mounts {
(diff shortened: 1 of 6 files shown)


enos/modules/docker_vault/main.tf
```diff
@@ -37,6 +37,11 @@ variable "vault_port" {
   type        = string
   default     = "8300"
 }
+variable "vault_port_internal" {
+  description = "External Port to use"
```
johanbrandhorst, line 41: ?
```suggestion
  description = "Internal Port to use"
```

johanbrandhorst (approved):
This is super cool, I didn't even know Vault could run as an OIDC server

johanbrandhorst (approved):

moduli:
Discussed with Johan. Will merge this as-is and will later explore a way to "manually" authenticate OIDC in this test.

PR #5385, opened 2024-12-19
fix(api): Correctly apply BOUNDARY_MAX_RETRIES env var

tmessi:
The api.Client was first checking for the environment variable, and then
resetting the max retries to 2, thus overriding the environment variable
value. In addition, the cli would set max retries to zero if the
environment variable was not set. This would mean that by default the
max retries would be zero, and if the environment variable was set, the
max retries would be 2, regardless of the actual value for the
environment variable.

This fixes the order in which the environment variables are read, to
ensure they take precedence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 
 ## Next
 
+### Bug fixes
+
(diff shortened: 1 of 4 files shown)


johanbrandhorst (approved):
Nice!

moduli (approved):

PR #5386, opened 2024-12-19
Revert "internal/daemon: speed up multi-worker tests (#5370)"

johanbrandhorst:
Unfortunately, this started causing the enterprise tests to break in an unexpected way. Reverting to spend some more time testing the changes.


moduli (approved):
😢

PR #5387, opened 2024-12-19
internal/daemon/worker: improve runtime and reduce noise

johanbrandhorst:
TestRotationTicking would previously wait for 5 rotations, which is pretty slow considering each rotation takes 20 seconds. Reduce the number to 3, and use require instead of assert to remove noise during flaky test runs.

This doesn't fix the flakiness of this test, but it should reduce the log spam.

### DIFF
--- a/internal/daemon/worker/auth_rotation_test.go
+++ b/internal/daemon/worker/auth_rotation_test.go
@@ -28,6 +28,8 @@ import (
 // TestRotationTicking ensures that we see new credential information for a
 // worker on both the controller side and worker side in a periodic fashion
 func TestRotationTicking(t *testing.T) {
+	t.Parallel()
+
 	require, assert := require.New(t), assert.New(t)
 	logger := hclog.New(&hclog.LoggerOptions{
 		Level: hclog.Trace,
@@ -95,7 +97,7 @@ func TestRotationTicking(t *testing.T) {
 	// Verify we see authorized credentials now
 	auths, err = workerAuthRepo.List(c.Context(), (*types.NodeInformation)(nil))
 	require.NoError(err)
-	assert.Len(auths, 2)
+	require.Len(auths, 2)
 	// Fetch creds and store current key
 	currNodeCreds, err := types.LoadNodeCredentials(
 		w.Context(),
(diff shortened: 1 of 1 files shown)


internal/daemon/worker/auth_rotation_test.go
```diff
@@ -28,6 +28,8 @@ import (
 // TestRotationTicking ensures that we see new credential information for a
 // worker on both the controller side and worker side in a periodic fashion
 func TestRotationTicking(t *testing.T) {
+	t.Parallel()
```
moduli, line 31: Was this intentional?

johanbrandhorst: yes, I think this should be able to run in parallel. I can't see any global resources used

internal/daemon/worker/auth_rotation_test.go
```diff
@@ -112,21 +114,27 @@ func TestRotationTicking(t *testing.T) {
 	// Make sure we have a failsafe in case the below loop never finds what it's
 	// looking for; at expiration the List will fail and we'll hit the Require,
 	// exiting
-	deadlineCtx, deadlineCtxCancel := context.WithTimeout(c.Context(), 10*time.Minute)
+	testTimeout := 10 * time.Minute
+	testDeadline, ok := t.Deadline()
+	if ok {
+		testTimeout = time.Until(testDeadline) - time.Second
+	}
+	deadlineCtx, deadlineCtxCancel := context.WithTimeout(c.Context(), testTimeout)
```
moduli, line 122: Curious what the reasoning for this change in. If I'm understanding it right, we're allowing for the test to run until the remainder of the test timeout (if there is one) as opposed to the standard 10 minute timeout? Should we expect this test to ever take that long?

moduli (approved):

internal/daemon/worker/auth_rotation_test.go
johanbrandhorst: Hm, fair, I wanted this to respect the test timeout if it's shorter than the 10 minute timeout, but yeah, we should probably only override it when it's smaller, not when it's bigger. I've had this test fail because I use a shorter test timeout locally. I'll push an update 👍🏻

moduli (approved):

PR #5391, opened 2024-12-20
DO NOT MERGE: specify tlskyber=0 go to disable two-packet TLS hello

johanbrandhorst:
@ajayreshc is the MVP for pointing out this option to me rather than downgrading our Go version


johanbrandhorst:
This will no longer be necessary

PR #5392, opened 2024-12-20
test(cluster): small stability improvements to unit tests

ddebko:


internal/tests/helper/testing_helper.go
```diff
@@ -436,26 +436,52 @@ func NewTestTcpServer(t *testing.T) *TestTcpServer {
...
+					// expecting it we'll see an out-of-date entry
+					return true
+				}
+				assert.WithinDuration(t, time.Now(), v.(time.Time), 30*time.Second)
+				delete(workerMap, k.(string))
+				return true
 			})
-			return empty
-		}, 30*time.Second, 2*time.Second)
+			return len(workerMap) == 0
```
johanbrandhorst, line 483: This logic still doesn't make sense to me. `workers` is empty, why do we even use the map?

johanbrandhorst (changes requested):
Thanks for opening a separate PR for these changes!

internal/tests/helper/testing_helper.go
ddebko: Good point, I was trying to revert to the original logic, which seems to work. 

https://github.com/hashicorp/boundary/blob/cd9c5f9e529ff9dfdc0c5ea4d636721d96591309/internal/tests/helper/testing_helper.go#L507

Now I realize I should just make sure the updateTimes map is empty.

ddebko: I updated the code to check if the last update time exceeds the default grace period.

internal/tests/helper/testing_helper.go
```diff
@@ -463,25 +470,17 @@ func ExpectWorkers(t *testing.T, c *controller.TestController, workers ...*worke
...
+				if time.Since(lastStatusTime.(time.Time)) < DefaultSuccessfulStatusGracePeriod {
+					workers = append(workers, workerId.(string))
 				}
-				assert.WithinDuration(t, time.Now(), v.(time.Time), 30*time.Second)
-				delete(workerMap, k.(string))
 				return true
 			})
-			return len(workerMap) == 0
-		}, 30*time.Second, 250*time.Millisecond)
+			return len(workers) == 0
```
johanbrandhorst, line 482: Yeah this makes more sense, thanks!

johanbrandhorst (approved):

PR #5393, opened 2024-12-20
test(cluster): check worker status multiple times

ddebko:
# Summary

The cluster session cleanup test would fail randomly due to a worker status context deadline timeout. This update fixes the flaky test by checking the worker status multiple times using a `require.Eventually` check.


internal/tests/cluster/sequential/session_cleanup_test.go
```diff
@@ -289,7 +294,6 @@ func testWorkerSessionCleanupMulti(burdenCase timeoutBurdenType) func(t *testing
 		// Worker needs some extra time to become ready, otherwise for a
 		// currently-unknown reason the next successful status update fails
 		// because it's not sent before the context times out.
-		time.Sleep(5 * time.Second)
```
johanbrandhorst, line 292: If removing this sleep fixes the error, we should remove the comment too

johanbrandhorst (approved):

PR #5394, opened 2024-12-23
chore: Install tools if cache restore fails

moduli:
This PR updates the unit test workflow to resolve an error when the cache restores to fail. Here is an example run: https://github.com/hashicorp/boundary/actions/runs/12439588543/job/34735079491.

When the cache fails to restore, the pipeline does not have access to the `tparse` tool, resulting in the error
```
Run nick-fields/retry@7152eba30c6575329ac0576536151aca5a72780e
go test "./..." -tags="" -v -json -cover -timeout 40m | tparse -follow
/bin/sh: 1: tparse: not found
```

This PR adds a step to install tools if the restore happens to fail.

### DIFF
--- a/.github/workflows/test-race.yml
+++ b/.github/workflows/test-race.yml
@@ -132,6 +132,7 @@ jobs:
           go-version: "${{ needs.setup.outputs.go-version }}"
           cache: false
       - name: Set up Go modules cache
+        id: go-cache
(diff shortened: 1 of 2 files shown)


johanbrandhorst (dismissed):
Ah, great thinking!

ddebko (approved):

moduli (commented):
Needed to also put it on the `test-race` workflow

johanbrandhorst (approved):

PR #5396, opened 2024-12-23
fix(target): Assign session connection limit to SessionAuthorization.ConnectionLimit

dkanney:
- We weren’t assigning the session connection limit we got from the DB to the `ConnectionLimit` field on the `SessionAuthorization` response object before returning from `AuthorizeSession` [[link](https://github.com/hashicorp/boundary-enterprise/blob/62c41191ffce6ab68fbed810559dd857fa8a2528/internal/daemon/controller/handlers/targets/target_service.go#L1203)]. As a result, the connection limit has a default value of `0` when the connect cmd prints the session info
- We _do_ assign `ConnectionLimit` in a similar object, `SessionAuthorizationData`, but it gets marshalled into an auth token. Instead of decoding & unmarshalling the entire `SessionAuthorizationData` payload for the connection limit, I decided to just copy its value onto the `SessionAuthorization` struct.

### DIFF
--- a/internal/daemon/controller/handlers/targets/target_service.go
+++ b/internal/daemon/controller/handlers/targets/target_service.go
@@ -1214,6 +1214,7 @@ func (s Service) AuthorizeSession(ctx context.Context, req *pbs.AuthorizeSession
 		HostSetId:          hostSetId,
 		Endpoint:           endpointUrl.String(),
 		Credentials:        creds,
+		ConnectionLimit:    t.GetSessionConnectionLimit(),
 	}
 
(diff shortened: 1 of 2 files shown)


dkanney:
TODO:
- [ ] Add test case

johanbrandhorst (dismissed):
Sorry for the drive by review, I was curious when Damian mentioned it 😁. The fix looks great to me, thank you! It made me curious why we have this information both in the SessionAuthorization and SessionAuthorizationData 🤔.

ddebko (approved):
This looks great, feel free to merge once all the tests are passing.

PR #5397, opened 2024-12-30
test(server): optimize multihop worker helper func

ddebko:
# Summary

Replacing the Sleep duration call and utilizing more accurate methods of validating when a worker-controller connection is established for the helper method `NewTestMultihopWorkers`. This should help speed up some of the tests.


internal/daemon/worker/testing.go
```diff
@@ -552,13 +537,59 @@ func NewTestMultihopWorkers(t testing.TB,
...
+	wg.Wait()
+
+	// validate connectivity between worker pool
+	require.Eventually(func() bool {
+		pkiPath, err := c.DownstreamWorkers().PathTo(c.Context(), pkiWorker.Worker().LastStatusSuccess().WorkerId)
+		if err != nil {
+			return false
+		}
+		childPkiPath, err := c.DownstreamWorkers().PathTo(c.Context(), childPkiWorker.Worker().LastStatusSuccess().WorkerId)
+		if err != nil {
```
johanbrandhorst, line 588: Maybe this should go in a testing_ent instead if we can't do this in community 🤔

ddebko: Should we move all the tests that reference this helper function into `_ent` files too?

johanbrandhorst: Yeah that'd be a requirement too.

johanbrandhorst (approved):
It's a shame to move these tests out of community, but it'll be worth it if the tests are more stable, IMO

PR #5402, opened 2025-01-03
internal/session: fix race condition in connection cleanup

johanbrandhorst:
The totalClosed variable was access in both the Status and the Run functions, which may be called from independent goroutines.

### DIFF
--- a/internal/session/job_session_cleanup.go
+++ b/internal/session/job_session_cleanup.go
@@ -40,7 +40,7 @@ type sessionConnectionCleanupJob struct {
 	workerStatusGracePeriod *atomic.Int64
 
 	// The total number of connections closed in the last run.
-	totalClosed int
+	totalClosed *atomic.Int64
 }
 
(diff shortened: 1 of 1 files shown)


louisruch (commented):
Hmm, which test was triggering a datarace? I was trying to see locally but didnt seem to get hit with it. 

The issue with this approach is that we hold the lock for the entirety of the `Run` which is the same time that `Status` needs to be called, this means that will block in `Status` until it is not longer valid to send a status which risks jobs do take a bit longer to run being killed prematurely. 

We are doing this for every job that exists but don't see any datarace warnings for them or hold locks for any of them, that being said we do call read from one routine (monitorJobs) and write in another (Run).

johanbrandhorst:
This particular failure was triggered in a generic enterprise test using the test controller. I sent a link on Slack. If we need to fix this across all jobs I'm happy to do that too.

PR #5403, opened 2025-01-03
test(groups): add grants tests for groups API

bosorawis:
Add tests that validate covering grants in preparation for grants system rework


internal/daemon/controller/handlers/groups/grants_test.go
```diff
@@ -0,0 +1,1079 @@
...
+						grantStrings: []string{"id=*;type=*;actions=*"},
+						grantScopes:  []string{globals.GrantScopeThis, globals.GrantScopeChildren},
+					},
+				},
+				wantErr: nil,
+				// TODO (Bo 20-dec-2024): expect 3 groups but only getting 1 back
+				// need to investigate further
+				// wantIDs: []string{globalGroup.PublicId, org1Group.PublicId, org2Group.PublicId},
+				wantIDs: []string{globalGroup.PublicId},
+			},
```
johanbrandhorst, line 164: Did we get to the bottom of this?

johanbrandhorst (approved):
Very cool, thanks for this!

internal/daemon/controller/handlers/groups/grants_test.go
bosorawis: Yup! found a bug in how the result sets get filtered - parent scope ID isn't getting passed into `authResults.FetchActionSetForId`. Fixed in [this](https://github.com/hashicorp/boundary/pull/5403/commits/d9dbcfd18c618f44c6f2df87f0030ef5f831c937) commit. Will make a separate PR for that specific bug fix and test since this PR is intended for an LLB

johanbrandhorst: Amazing!!

johanbrandhorst:
Looks like you need a "go mod tidy"

internal/daemon/controller/handlers/groups/grants_test.go
```diff
@@ -0,0 +1,1076 @@
...
+					require.NoErrorf(t, err, "failed to delete group in scope %s", scope)
+				}
+			})
+		}
+	})
+
+	t.Run("update", func(t *testing.T) {
+		testcases := []struct {
+			name                      string
+			setupScopesResourcesRoles func(t *testing.T, conn *db.DB, iamRepo *iam.Repository) (*iam.Group, []roleRequest)
```
elimt, line 642: Is there a reason why we can't pass only `roleRequest` and allow the for loop to create a test group to follow the pattern of other tests? is it because we want to create groups in different scopes?

bosorawis: because in some cases we need ID of the group for testing resource-specific grant string

internal/daemon/controller/handlers/groups/grants_test.go
```diff
@@ -0,0 +1,1076 @@
...
+					},
+				},
+				inputWantErrMap: map[*pbs.GetGroupRequest]error{
+					{Id: globalGroup.PublicId}: handlers.ForbiddenError(),
+					{Id: org1Group.PublicId}:   nil,
+					{Id: proj1Group.PublicId}:  nil,
+				},
+			},
+			{
+				name: "global_role_grant_all_specific_permissions",
```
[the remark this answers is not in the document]
bosorawis: good catch! my bad, will push a fix soon

internal/daemon/controller/handlers/groups/grants_test.go
```diff
@@ -0,0 +1,1076 @@
...
+						roleScopeID:  globals.GlobalPrefix,
+						grantStrings: []string{"id=*;type=*;actions=*"},
+						grantScopes:  []string{globals.GrantScopeThis, globals.GrantScopeDescendants},
+					},
+				},
+				wantErr: nil,
+				wantIDs: []string{globalGroup.PublicId, org1Group.PublicId, org2Group.PublicId, proj1Group.PublicId, proj2Group.PublicId, proj3Group.PublicId},
+			},
+			{
+				name: "org role grant children IDs only org children",
```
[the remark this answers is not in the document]
bosorawis: Oops, this is a bad test name. The role is in `global` with `descendant` access - calling `list` at `org2` will return `org2, proj2, proj3`. AFAIK, there's no way to make a `list` exclude the "listing scope". You always get all the groups in `ScopeId` of the list call.

internal/daemon/controller/handlers/groups/grants_test.go
```diff
@@ -0,0 +1,1076 @@
...
+type roleRequest struct {
+	roleScopeID  string
+	grantStrings []string
+	grantScopes  []string
+}
+
+// genAuthTokenCtx creates an auth.VerifierContext which contains a valid auth token
+// for a user which is associated with roles in the roles parameter
+// this function creates an authMethod, account, user at global scope
+func genAuthTokenCtx(t *testing.T,
```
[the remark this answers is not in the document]
bosorawis: Since the function takes `*testing.T` as its first parameter so I don't think this can be accidentally called outside of test but I'll add `test...` prefix to its name 👍

internal/daemon/controller/handlers/groups/grants_test.go
```diff
@@ -0,0 +1,1076 @@
...
+	proj3 := iam.TestProject(t, iamRepo, org2.GetPublicId())
+
+	globalGroup := iam.TestGroup(t, conn, globals.GlobalPrefix, iam.WithDescription("global"), iam.WithName("global"))
+	org1Group := iam.TestGroup(t, conn, org1.GetPublicId(), iam.WithDescription("org1"), iam.WithName("org1"))
+	org2Group := iam.TestGroup(t, conn, org2.GetPublicId(), iam.WithDescription("org2"), iam.WithName("org2"))
+	proj1Group := iam.TestGroup(t, conn, proj1.GetPublicId(), iam.WithDescription("proj1"), iam.WithName("proj1"))
+	proj2Group := iam.TestGroup(t, conn, proj2.GetPublicId(), iam.WithDescription("proj2"), iam.WithName("proj2"))
+	proj3Group := iam.TestGroup(t, conn, proj3.GetPublicId(), iam.WithDescription("proj3"), iam.WithName("proj3"))
+
+	t.Run("List", func(t *testing.T) {
```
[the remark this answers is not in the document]
bosorawis: added tests to cover both cases

internal/daemon/controller/handlers/groups/grants_test.go
```diff
@@ -0,0 +1,1076 @@
...
+					},
+				},
+				inputWantErrMap: map[*pbs.GetGroupRequest]error{
+					{Id: globalGroup.PublicId}: handlers.ForbiddenError(),
+					{Id: org1Group.PublicId}:   nil,
+					{Id: proj1Group.PublicId}:  nil,
+				},
+			},
+			{
+				name: "global_role_grant_all_specific_permissions",
```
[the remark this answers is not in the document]
bosorawis: another test I forgot to rename 🤦

johanbrandhorst (approved):

internal/daemon/controller/handlers/groups/grants_test.go
```diff
@@ -0,0 +1,1123 @@
...
+func TestGrants_ReadActions(t *testing.T) {
+	ctx := context.Background()
+	conn, _ := db.TestSetup(t, "postgres")
+	wrap := db.TestWrapper(t)
+	iamRepo := iam.TestRepo(t, conn, wrap)
+	repoFn := func() (*iam.Repository, error) {
+		return iamRepo, nil
+	}
+	s, err := groups.NewService(ctx, repoFn, 1000)
+	require.NoError(t, err)
```
dkanney, line 112: We're performing this test setup in a few places -- we could reduce code by moving this into a shared function

bosorawis: 🤔 We're using most of these setup objects throughout the test though. I'm not sure if we could abstract this away while keeping those available

elimt (approved):

internal/daemon/controller/handlers/groups/grants_test.go
```diff
@@ -0,0 +1,1154 @@
...
+
+		org1, proj1 := iam.TestScopes(t, iamRepo)
+		org2, proj2 := iam.TestScopes(t, iamRepo)
+		proj3 := iam.TestProject(t, iamRepo, org2.GetPublicId())
+
+		allScopeIDs := []string{globals.GlobalPrefix, org1.PublicId, org2.PublicId, proj1.PublicId, proj2.PublicId, proj3.PublicId}
+		testcases := []struct {
+			name              string
+			userFunc          func() *iam.User
+			canCreateInScopes []string
```
elimt, line 554: For consistency, why don't we follow what we do for `Get` between tests where we attempt an action and evaluate the result for each group. Makes it easier add to it and to follow

```
inputWantErrMap: map[*pbs.CreateGroupRequest]error{
					{Id: globalGroup.PublicId}: nil,
					{Id: org1Group.PublicId}:   nil,
					{Id: proj1Group.PublicId}:  nil,
					{Id: org2Group.PublicId}:   handlers.ForbiddenError(),
					{Id: proj2Group.PublicId}:  handlers.ForbiddenError(),
				}
```

internal/iam/testing.go
```diff
@@ -315,6 +315,98 @@ func TestManagedGroupRole(t testing.TB, conn *db.DB, roleId, managedGrpId string
...
+	return func() *User {
+		t.Helper()
+		ctx := context.Background()
+		rw := db.New(conn)
+		repo, err := NewRepository(ctx, rw, rw, kmsCache)
+		require.NoError(t, err)
+		u, err := NewUser(ctx, scopeID)
+		require.NoError(t, err)
+		user, err := repo.CreateUser(ctx, u)
+		require.NoError(t, err)
```
elimt, line 366: This part of it can be done with [iam.TestUser](https://github.com/hashicorp/boundary/blob/e707cff801c6ec2fa5d5a94242f00300288eb6a1/internal/iam/testing.go#L167)

internal/iam/testing.go
```diff
@@ -315,6 +315,98 @@ func TestManagedGroupRole(t testing.TB, conn *db.DB, roleId, managedGrpId string
...
+		g, err := NewGroup(ctx, scopeID)
+		require.NoError(t, err)
+		group, err := repo.CreateGroup(ctx, g)
+		require.NoError(t, err)
+		u, err := NewUser(ctx, scopeID)
+		require.NoError(t, err)
+		user, err := repo.CreateUser(ctx, u)
+		require.NoError(t, err)
+		for _, trg := range testRoleGrants {
```
elimt, line 400: What do you think about splitting this up some more. Have a function that only creates the target resource(direct,group,managed_group) and another function that does the associating of the target resources to roles set in `TestRoleGrantsRequest` by type

bosorawis: IMO that makes writing the test table more complicated since they cannot be normalized across all types of grants associations. 

I wanted to incorporate all grants type to under one setup interface (`func() *iam.User`) so the table-driven tests can use the same table so tests don't need to be repeated. I'm happy to make them more distinct though but I'm afraid people might ended up ignoring the more complicated test setup (like LDAP/OIDC)

internal/daemon/controller/handlers/groups/grants_test.go
```diff
@@ -0,0 +1,1154 @@
...
+					{Id: org1Group.PublicId}:   nil,
+					{Id: proj1Group.PublicId}:  nil,
+					{Id: org2Group.PublicId}:   handlers.ForbiddenError(),
+					{Id: proj2Group.PublicId}:  handlers.ForbiddenError(),
+				},
+			},
+		}
+
+		for _, tc := range testcases {
+			t.Run(tc.name, func(t *testing.T) {
```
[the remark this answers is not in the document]
bosorawis: IMO `output_fields` should be tested across all CRUDL methods, not just the the read operations. I'm still working on adding a specific `output_fields` tests which encompasses all the method types.

internal/iam/testing.go
```diff
@@ -345,6 +315,148 @@ func TestManagedGroupRole(t testing.TB, conn *db.DB, roleId, managedGrpId string
 	return r
 }
 
+// TestRoleWithGrants creates a role suitable for testing along with grants
+// Functional options for GrantScopes aren't used to express that
+// this function does not provide any default grant scope unlike TestRole
+func TestRoleWithGrants(t testing.TB, conn *db.DB, scopeId string, grantScopeIDs []string, grants []string) *Role {
```
bosorawis, line 321: this will be moved to an unexported function if the setup pattern is accepted

elimt (approved):
Approving with the assumption the minor comments will be addressed

internal/daemon/controller/handlers/groups/grants_test.go
```diff
@@ -116,4 +252,1948 @@ func TestGrants_ReadActions(t *testing.T) {
...
+			})
+		}
+	})
+
+}
+
+// assertOutputFields asserts that the output fields of a group match the expected fields
+// fields that is nil or empty in the result will throw an error if they are listed in expectedFields
+// e.g. members when group does not contian any members
+func assertOutputFields(t *testing.T, g *pb.Group, expectFields []string) {
```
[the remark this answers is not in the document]
bosorawis: moved to `handlers.AssertOutputFields`

internal/iam/testing.go
```diff
@@ -345,6 +315,148 @@ func TestManagedGroupRole(t testing.TB, conn *db.DB, roleId, managedGrpId string
...
+// the request grants through managed group.
+// Note: This method is not responsible for associating the user to the managed group. That action needs to be done
+// by the caller
+// This function returns iam.User and the AccountID from the account setup func
+func TestUserManagedGroupGrantsFunc(
+	t *testing.T,
+	conn *db.DB,
+	kmsCache *kms.Kms,
+	scopeID string,
+	managedGroupAccountSetupFunc func() (string, string),
```
tmessi, line 364: I think we could define some types for some of these functions to get stricter type checking. Let's sync up since it would involve several changes in various places, so it isn't very easy to show in a PR comment.

tmessi (changes requested):

internal/daemon/controller/handlers/groups/grants_test.go
```diff
@@ -116,4 +252,1948 @@ func TestGrants_ReadActions(t *testing.T) {
...
+				require.NoError(t, err)
+				require.Equal(t, uint32(2), got.Item.Version)
+				require.True(t, got.Item.UpdatedTime.AsTime().After(original.UpdateTime.AsTime()))
+			})
+		}
+	})
+}
+
+// TestGroupMember tests actions performed on the group-members (child-resources)
+func TestGrants_ChildResourcesActions(t *testing.T) {
```
dkanney, line 855:
```suggestion
// TestGrants_ChildResourcesActions tests actions performed on the group-members (child-resources)
func TestGrants_ChildResourcesActions(t *testing.T) {
```

internal/daemon/controller/handlers/groups/grants_test.go
dkanney: I think what I meant by this comment was something like [this](https://github.com/hashicorp/boundary/pull/5569/commits/742d64ac155d1bbe04450f2298c79ba038fd05f7#diff-8dc688d05107ea1a68ab36833787b6df37185c190c6a22bca9f58455a2e4438bR141) -- on my branch for auth methods I created a `TestCore()` function that returns all this stuff to reduce LOC:

```go
// getTestCore returns the core components required to run authmethod tests.
func getTestCore(t *testing.T) (
	ctx context.Context,
	rootWrapper wrapping.Wrapper,
	conn *db.DB,
	rw *db.Db,
	kmsCache *kms.Kms,
	iamRepo *iam.Repository,
) {
	t.Helper()
	ctx = context.Background()
	rootWrapper = db.TestWrapper(t)
	conn, _ = db.TestSetup(t, "postgres")
	rw = db.New(conn)
	kmsCache = kms.TestKms(t, conn, rootWrapper)
	iamRepo = iam.TestRepo(t, conn, rootWrapper)
	return
}
```

So now I initialize all the dependencies with a single function call instead of initializing each line-by-line:

```go
func TestGrants_WriteActions(t *testing.T) {
	ctx, wrap, conn, rw, kmsCache, iamRepo := getTestCore(t)
	...
}
```

internal/iam/testing.go
bosorawis: That function in `authtoken` was added as a part of the `children` grants fix in `0.19.0`. The intention is to remove the function in `authtoken` that calls this function. Other than that, there's no external caller of this function.

internal/auth/db_test.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) HashiCorp, Inc.
 // SPDX-License-Identifier: BUSL-1.1
 
-package auth
+package auth_test
```
bosorawis, line 4: Had to move this to a test package to avoid circular import between `iam` and `auth` (`iam` now imports `auth`)

internal/daemon/controller/handlers/groups/grants_test.go
```diff
@@ -116,4 +253,1948 @@ func TestGrants_ReadActions(t *testing.T) {
...
+				require.NoError(t, err)
+				require.Equal(t, uint32(2), got.Item.Version)
+				require.True(t, got.Item.UpdatedTime.AsTime().After(original.UpdateTime.AsTime()))
+			})
+		}
+	})
+}
+
+// TestGrants_ChildResourcesActions tests actions performed on the group-members (child-resources)
+func TestGrants_ChildResourcesActions(t *testing.T) {
```
tmessi, line 856: I don' think "child resources" is the correct terminology for this. This seems to just be testing the various group membership actions, which are additional ID actions that groups support, but they still operate on a group. A "child resource" would refer to something like a credential library which has a parent resource of a credential store.

tmessi (changes requested):
It's nice to see this robust testing. I just have one minor thing to adjust.

internal/daemon/controller/handlers/groups/grants_test.go
bosorawis: Renamed that to `TestGrants_GroupMembership`. I also reorganized tests to `TestGrants_<api-name>` instead of `TestGrants_<Read|Write>actions`

tmessi (approved):

PR #5405, opened 2025-01-03
mod: update golang.org/x/net dependency

johanbrandhorst:
Fixes CVE-2024-45338, GO-2024-3333


dkanney (approved):

dkanney:
For the plugins -- do we need to bump `golang.org/x/net` to v.33 in their respective repos as well?

johanbrandhorst:
> For the plugins -- do we need to bump `golang.org/x/net` to v.33 in their respective repos as well?

I don't think so, I don't think we're actually affected by this, but it's blocking us from doing any sort of releases right now.

PR #5412, opened 2025-01-07
feat(cmd): Make db migrate plugin configurable and default to azure

hugoghx:


ddebko (approved):

internal/cmd/commands/database/migrate.go
```diff
@@ -125,6 +126,13 @@ func (c *MigrateCommand) Flags() *base.FlagSets {
 		Usage:  `Run the repair function for the provided migration version.`,
 	})
 
+	f.StringVar(&base.StringVar{
+		Name:    "plugin-name",
+		Target:  &c.flagPluginName,
+		Usage:   "The 'database migrate' command runs a Boundary plugin environment check by creating a client to a plugin. This flag controls which plugin gets instantiated for that check.",
+		Default: "azure",
+	})
```
johanbrandhorst, line 134: I appreciate this option, but since we're only creating a plugin to see that we can actually execute plugins, I don't think this needs to be configurable. If a customer asks for it (why would they?) I think we can add it, but until then, less is more. What do you think?

hugoghx: I discussed this with Louis and the idea was that right now the best plugin to default to may be Azure, but that may not may always be the case into the future. Using a flag allows for flexibility in this.

louisruch: I do agree with Johan that there is no real need to expose this to users, it may just confuse as to what it really is. 
Making a code change for a future change should be sufficient given the lightweight nature of the change. 

We would need to rethink this path in a future ecosystem world regardless as we would likely need to discover what plugins exist.

hugoghx:
@johanbrandhorst @louisruch Given that we won't make this dynamic, I'm closing this PR in favor of this one https://github.com/hashicorp/boundary/pull/5422

PR #5413, opened 2025-01-07
.github: disable github-actions dependabot

johanbrandhorst:
Github actions versions are now managed by TSCCR and these PRs are just noise.


picatz (approved):

moduli (commented):
I believe TSCCR only handles updates to external actions? I think we want to configure this to only update internal actions, if possible.

PR #5414, opened 2025-01-07
chore: Only use dependabot for internal action updates

moduli:
This PR updates the dependabot configuration to only provide updates for internal Github Action updates. We have an internal tool to provide security approved 3rd party Action updates, so the PRs generated by dependbot are redundant.


johanbrandhorst (approved):
Thank you!

picatz (approved):

PR #5423, opened 2025-01-08
test(e2e): Add test for `connect http`

moduli:
This PR adds an e2e test for the usage of `boundary connect http`.

The way this is achieved is a little roundabout. The test first connects to the SSH target and enables port forwarding. It then kicks off a process on a target to start a simple web server. Finally, the test creates a new target pointing to the web server and performs a `connect http`.

https://hashicorp.atlassian.net/browse/ICU-10838

### DIFF
--- a/testing/internal/e2e/tests/base/target_tcp_connect_http_test.go
+++ b/testing/internal/e2e/tests/base/target_tcp_connect_http_test.go
@@ -1,0 +1,94 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package base_test
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os/exec"
+	"testing"
+
+	"github.com/creack/pty"
+	"github.com/hashicorp/boundary/internal/session"
+	"github.com/hashicorp/boundary/internal/target"
+	"github.com/hashicorp/boundary/testing/internal/e2e"
+	"github.com/hashicorp/boundary/testing/internal/e2e/boundary"
+	"github.com/stretchr/testify/require"
+)
+
+// TestCliTcpTargetConnectHttp validates the usage of "boundary connect http".
+// This is done by first connecting to an SSH target, starting a webserver on
+// that target, and then creating another target that connects to the webserver
+func TestCliTcpTargetConnectHttp(t *testing.T) {
+	e2e.MaybeSkipTest(t)
+	c, err := loadTestConfig()
+	require.NoError(t, err)
+
+	ctx := context.Background()
+	boundary.AuthenticateAdminCli(t, ctx)
+	orgId, err := boundary.CreateOrgCli(t, ctx)
+	require.NoError(t, err)
+	t.Cleanup(func() {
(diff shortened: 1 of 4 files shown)


testing/internal/e2e/tests/base/target_tcp_connect_http_test.go
```diff
@@ -0,0 +1,94 @@
...
+		err := f.Close()
+		require.NoError(t, err)
+	})
+
+	t.Log("Starting a webserver on the target...")
+	htmlPage := "<html>Hello World!</html>"
+	go func() {
+		_, err = f.Write([]byte(fmt.Sprintf("echo '%s' > somepage.html\n", htmlPage)))
+		require.NoError(t, err)
+		_, err = f.Write([]byte(fmt.Sprintf("while true; do nc -l -p %s -q 1 < somepage.html; done\n", destPort)))
```
johanbrandhorst, line 72: Does this work? `nc` speaks TCP, not HTTP, right? Don't you need to at least add the HTTP protocol headers to the response? E.g.
```suggestion
	htmlPage := `
HTTP/1.1 200 OK
Content-Type: text/html; charset=UTF-8
Server: netcat-can-you-believe-it

<html>Hello World!</html>
`
	go func() {
		_, err = f.Write([]byte(fmt.Sprintf("echo '%s' > somepage.html\n", htmlPage)))
		require.NoError(t, err)
		_, err = f.Write([]byte(fmt.Sprintf("while true; do nc -l -p %s -q 1 < somepage.html; done\n", destPort)))
```

testing/internal/e2e/tests/base/target_tcp_connect_http_test.go
```diff
@@ -0,0 +1,94 @@
...
+		require.NoError(t, err)
+		_, _ = io.Copy(io.Discard, f) // Not checking error here since it fails in CI for some reason
```
johanbrandhorst, line 74: 🤔?

testing/internal/e2e/tests/base/target_tcp_connect_http_test.go
```diff
@@ -0,0 +1,94 @@
...
+	// Create http target and connect to it
+	httpTargetId, err := boundary.CreateTargetCli(t, ctx, projectId, destPort, target.WithAddress(c.TargetAddress))
+	require.NoError(t, err)
+	output := e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"connect", "http",
+			"-target-id", httpTargetId,
+			"-scheme", "http",
+			"--",
+			"--http0.9",
```
johanbrandhorst, line 89: Is this how we're working around the no HTTP header business 😝?

johanbrandhorst (changes requested):

testing/internal/e2e/tests/base/target_tcp_connect_http_test.go
moduli: The theory: since the command being executed is an infinite loop, this ends up returning an error on session cancellation during cleanup since the command was interrupted.

johanbrandhorst (approved):

PR #5437, opened 2025-01-20
Restrict concurrent password hashes

johanbrandhorst:
### [internal/auth/password: add resizable pool](https://github.com/hashicorp/boundary/pull/5437/commits/5d3ed2ccb8d6784b63ab7e7d97956bcdd92e9467)
The resizable pool wraps the permit pool with a mutex
to allow it to be resized at runtime.

### [internal/auth/password: add concurrency limit](https://github.com/hashicorp/boundary/pull/5437/commits/e75ce91c84ff1af3b95c8f299b56c33f893a24c1)
The default concurrency limit is 1, and it can be
set using the SetHashingPermits function.

### [internal/cmd/config: allow configuring hashing limits](https://github.com/hashicorp/boundary/pull/5437/commits/724cbb1e12a3fe6a6ed68d001d66bf2e52d94ac8)
The new concurrent_password_hash_workers configuration
value can be used to loosen the constraints on the
password auth method hashing operation.

### [internal/cmd: set password hashing constraints](https://github.com/hashicorp/boundary/pull/5437/commits/8285311bc27b1a06ab530cc7979551bedee24efc)
Sets the password hashing concurrency constraints on
startup and config reload.

### [CHANGELOG: add notice about password concurrency limit](https://github.com/hashicorp/boundary/pull/5437/commits/82356e17f0e9c5d34868a9abc94d92ec6c5c5400)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 * Fix bug in parsing IPv6 addresses. Previously setting a target address or the
   initial upstream address in the config file would result in a malformed value.
   ([PR](https://github.com/hashicorp/boundary/pull/5221)).
+* Fix an issue where, when starting a session, the connection limit always displays 0.
+  ([PR](https://github.com/hashicorp/boundary/pull/5396)).
 
 ### New and Improved
 
@@ -33,11 +35,16 @@ maintainability of worker queries, and improve DB performance. ([PR](https://git
   and reliability at large scale. Workers older than v0.19.0 will remain supported
   until the release of v0.20.0, in accordance with
   [our worker/controller compatiblity policy](https://developer.hashicorp.com/boundary/docs/enterprise/supported-versions#control-plane-and-worker-compatibility).
+* Add concurrency limit on the password hashing of all password auth methods.
+  ([PR](https://github.com/hashicorp/boundary-plugin-aws/pull/5437)).
 
-### Bug fixes
+  This avoids bursty memory and CPU use during concurrent password auth method
+  authentication attempts. The number of concurrent hashing operations
+  can be set with the new `concurrent_password_hash_workers` configuration
+  value in the controller stanza, or the new 
+  `BOUNDARY_CONTROLLER_CONCURRENT_PASSWORD_HASH_WORKERS` environment variable.
+  The default limit is 1.
 
-* Fix an issue where, when starting a session, the connection limit always displays 0.
-  ([PR](https://github.com/hashicorp/boundary/pull/5396)).
 
 ## 0.18.2 (2024/12/12)
 ### Bug fixes
--- a/go.mod
+++ b/go.mod
@@ -63,7 +63,7 @@ require (
 	github.com/posener/complete v1.2.3
 	github.com/prometheus/client_golang v1.18.0
 	github.com/ryanuber/go-glob v1.0.0
-	github.com/stretchr/testify v1.9.0
+	github.com/stretchr/testify v1.10.0
 	github.com/zalando/go-keyring v0.2.3
 	go.uber.org/atomic v1.11.0
 	golang.org/x/crypto v0.31.0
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -24,11 +26,14 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
...
+  authentication attempts. The number of concurrent hashing operations
+  can be set with the new `concurrent_password_hash_workers` configuration
+  value in the controller stanza, or the new 
+  `BOUNDARY_CONTROLLER_CONCURRENT_PASSWORD_HASH_WORKERS` environment variable.
+  The default limit is 1.
 
-### Bug fixes
-
-* Fix an issue where, when starting a session, the connection limit always displays 0.
-  ([PR](https://github.com/hashicorp/boundary/pull/5396)).
```
johanbrandhorst, line 31: There were two "Bug Fixes" sections in the changelog under "Next", so I moved this one up to the one above.

go.mod
```diff
@@ -93,6 +93,7 @@ require (
 	github.com/hashicorp/dbassert v0.0.0-20231012105025-1bc1bd88e22b
 	github.com/hashicorp/go-kms-wrapping/extras/kms/v2 v2.0.0-20241126174344-f3b1a41a15fd
 	github.com/hashicorp/go-rate v0.0.0-20231204194614-cc8d401f70ab
+	github.com/hashicorp/go-secure-stdlib/permitpool v0.0.0-20250109173936-1ecdd6ad783f
```
johanbrandhorst, line 96: This will need to be updated before merge, it's currently depending on a branch for https://github.com/hashicorp/go-secure-stdlib/pull/153

internal/auth/password/argon2.go
```diff
@@ -7,16 +7,46 @@ import (
...
 
+var (
+	// hashingPermitPool is the global permit pool used to restrict concurrent
+	// password hashing. It can be configured with SetHashingPermits.
+	hashingPermitPool *permitpool.Pool
+	// hashingPermitPoolLock is used to synchronize access to hashingPermitPool
+	// This is an RWMutex to allow an unlimited number of readers
+	// and a single writer, since allowing a single reader or writer
+	// would effectively make the pool useless.
+	hashingPermitPoolLock *sync.RWMutex
```
tmessi, line 30: If I am following this correctly, this mutex is use to allow the entire permit pool to be replaced safely, allowing it to be resized. Could this mutex be internalized into the pool itself by exposing a Resize method on the pool?

johanbrandhorst: I think I'd prefer not to add this to the pool itself, as it may encourage the use of global pools like this one. Ideally a global pool like this isn't necessary, and all operations can be funneled through a single type or struct, which would then allow each instance of such a struct to have a pool assigned on it (and the size decided at runtime). The only reason we need to resize this pool is because it is a global variable with a default size set at init time. Does that makes sense?

tmessi: Even if it isn't a global variable, wouldn't any resize operation require a mutex? And there could be other times that a resize is necessary or desirable, like if the pool size can be changed via a SIGHUP.

johanbrandhorst: Yes, resizing in general would require a mutex, I agree. But from the permitpool package point of view, a pool can just be considered immutable, and if a user wants to resize it they have to replace it. That does punt some complexity onto the user, but also keeps the pool simpler for users that don't need the resizing option. Vault for example has a [`PermitPoolWithMetrics`](https://github.com/hashicorp/vault/blob/5b4b606c0d80820a117cbbda8247b762fceab44b/physical/dynamodb/dynamodb.go#L128-L132) that wraps the permitpool primitive. I think there are other use cases like this that don't necessarily need to be implemented in the pool itself.

tmessi: Ok, then I would suggest that we implement something like a `ResizablePermitPool` that uses a `PermitPool` but provides a `Resize` method. This allows for the mutex to be coupled with the PermitPool, instead of them both being independent package variables. If we want to implement this just in boundary, it wouldn't need be an exported type for now.

johanbrandhorst: Sounds good 👍🏻

johanbrandhorst: Fixed, please take a look!

PR #5438, opened 2025-01-21
internal/session: ensure orphaned connection cleanup

johanbrandhorst:
Previously, the orphaned connection cleanup job would not detect that a connection that was created before a worker had submitted a successful session info was orphaned. We now fall back to using the worker update time as an indicator if the session info time is not set.


internal/session/query.go
```diff
@@ -292,16 +292,18 @@ where
...
-           from server_worker_session_info_request w
-          where
-			w.last_request_time < wt_sub_seconds_from_now(@grace_period_seconds)
+         select w.public_id as worker_id,
+                coalesce(wsi.last_request_time, w.update_time) as last_update_time
+           from server_worker w
+      left join server_worker_session_info_request wsi on wsi.worker_id = w.public_id 
+          where wsi.last_request_time < wt_sub_seconds_from_now(@grace_period_seconds)
+             or wsi.last_request_time is null
+            and w.update_time < wt_sub_seconds_from_now(@grace_period_seconds)
```
tmessi, line 306: It might be good to explicitly group these two clauses so it is clear how they relate to the first where clause, i.e.

```suggestion
             or (    wsi.last_request_time is null
                 and w.update_time < wt_sub_seconds_from_now(@grace_period_seconds))
```

tmessi (changes requested):

tmessi (approved):

PR #5443, opened 2025-01-21
test(grants): GrantsForUsers tests for Group resource

dkanney:
Goal: Write direct & indirect tests for the GrantsForUser function against the Group resource

- Direct: Directly associating a user to a role
- Indirect: Associating a user to a role via one of the following:
    1. Group
    2. Managed Group (e.g. OIDC, LDAP)

___
  
I created one test for each case: one direct test, one group test, and one managed group test (performed via OIDC). All tests contain grants against the Group resource (and its associated [actions](https://developer.hashicorp.com/boundary/docs/configuration/identity-access-management/resource-table#group))

### DIFF
--- a/internal/iam/repository_role_grant_ext_test.go
+++ b/internal/iam/repository_role_grant_ext_test.go
@@ -724,7 +724,6 @@ func TestGrantsForUser_DirectAssociation(t *testing.T) {
 			_, err := repo.GrantsForUser(ctx, userId, iam.WithTestCacheMultiGrantTuples(multiGrantTuplesCache))
 			require.NoError(t, err)
 
-			// log.Println("multiGrantTuplesCache", pretty.Sprint(*multiGrantTuplesCache))
 			assert.ElementsMatch(t, *multiGrantTuplesCache, expMultiGrantTuples[userId])
 		}
 	})
@@ -1038,14 +1037,7 @@ func TestGrantsForUser_DirectAssociation(t *testing.T) {
 					},
 				},
 			}
-			/*
-				log.Println("org1", directGrantOrg1.PublicId)
-				log.Println("proj1a", directGrantProj1a.PublicId)
-				log.Println("proj1b", directGrantProj1b.PublicId)
-				log.Println("org2", directGrantOrg2.PublicId)
-				log.Println("proj2a", directGrantProj2a.PublicId)
-				log.Println("proj2b", directGrantProj2b.PublicId)
-			*/
+
 			assert.Len(t, directGrants, len(expDirectGrants))
 			for k, v := range directGrants {
 				assert.ElementsMatch(t, v, expDirectGrants[k])
@@ -1395,7 +1387,6 @@ func TestGrantsForUser_Group(t *testing.T) {
 			_, err := repo.GrantsForUser(ctx, userId, iam.WithTestCacheMultiGrantTuples(multiGrantTuplesCache))
 			require.NoError(t, err)
 
-			// log.Println("multiGrantTuplesCache", pretty.Sprint(*multiGrantTuplesCache))
 			assert.ElementsMatch(t, *multiGrantTuplesCache, expMultiGrantTuples[userId])
 		}
 	})
@@ -1709,14 +1700,7 @@ func TestGrantsForUser_Group(t *testing.T) {
 					},
 				},
 			}
-			/*
-				log.Println("org1", directGrantOrg1.PublicId)
-				log.Println("proj1a", directGrantProj1a.PublicId)
-				log.Println("proj1b", directGrantProj1b.PublicId)
-				log.Println("org2", directGrantOrg2.PublicId)
-				log.Println("proj2a", directGrantProj2a.PublicId)
-				log.Println("proj2b", directGrantProj2b.PublicId)
-			*/
+
 			assert.Len(t, directGrants, len(expDirectGrants))
 			for k, v := range directGrants {
 				assert.ElementsMatch(t, v, expDirectGrants[k])
@@ -2088,7 +2072,6 @@ func TestGrantsForUser_ManagedGroup(t *testing.T) {
 			_, err := repo.GrantsForUser(ctx, userId, iam.WithTestCacheMultiGrantTuples(multiGrantTuplesCache))
 			require.NoError(t, err)
 
-			// log.Println("multiGrantTuplesCache", pretty.Sprint(*multiGrantTuplesCache))
 			assert.ElementsMatch(t, *multiGrantTuplesCache, expMultiGrantTuples[userId])
 		}
 	})
@@ -2402,14 +2385,7 @@ func TestGrantsForUser_ManagedGroup(t *testing.T) {
 					},
 				},
 			}
-			/*
-				log.Println("org1", directGrantOrg1.PublicId)
-				log.Println("proj1a", directGrantProj1a.PublicId)
-				log.Println("proj1b", directGrantProj1b.PublicId)
-				log.Println("org2", directGrantOrg2.PublicId)
-				log.Println("proj2a", directGrantProj2a.PublicId)
-				log.Println("proj2b", directGrantProj2b.PublicId)
-			*/
+
 			assert.Len(t, directGrants, len(expDirectGrants))
 			for k, v := range directGrants {
 				assert.ElementsMatch(t, v, expDirectGrants[k])
--- a/internal/iam/repository_role_grant_ext_test.go
+++ b/internal/iam/repository_role_grant_ext_test.go
@@ -389,3 +389,82 @@ func TestGrantsForUserRandomized(t *testing.T) {
 			", roles from ldap managed groups", rolesFromLdapManagedGroups)
 	}
 }
+
+func TestGrantsForUser_Groups(t *testing.T) {
+	ctx := context.Background()
+
+	conn, _ := db.TestSetup(t, "postgres")
+	wrap := db.TestWrapper(t)
+
+	repo := iam.TestRepo(t, conn, wrap)
+	user := iam.TestUser(t, repo, "global") // create a user
+
+	grant := "ids=*;type=*;actions=*"
+
+	group := iam.TestGroup(t, conn, "global") // create a group
+	role := iam.TestRole(t, conn, "global")
+	iam.TestGroupMember(t, conn, group.PublicId, user.PublicId) // add user to the group
+	iam.TestGroupRole(t, conn, role.PublicId, group.PublicId)   // assign a role to the group
+	iam.TestRoleGrant(t, conn, role.PublicId, grant)            // assign a grant to the role
+
+	t.Run("group association", func(t *testing.T) {
+		grantTuples, err := repo.GrantsForUser(ctx, user.PublicId)
+		require.NoError(t, err)
+
+		require.NotEmpty(t, grantTuples)
+		assert.Equal(t, grant, grantTuples[0].Grant)
+	})
+
+	t.Run("managed group association", func(t *testing.T) {
+		ctx := context.Background()
+		conn, _ := db.TestSetup(t, "postgres")
+		wrap := db.TestWrapper(t)
+		iamRepo := iam.TestRepo(t, conn, wrap)
+		kmsCache := kms.TestKms(t, conn, wrap)
+
+		o, _ := iam.TestScopes(
+			t,
+			iamRepo,
+			iam.WithSkipAdminRoleCreation(true),
+			iam.WithSkipDefaultRoleCreation(true),
+		)
+		databaseWrapper, err := kmsCache.GetWrapper(ctx, o.PublicId, kms.KeyPurposeDatabase)
+		require.NoError(t, err)
+
+		oidcAuthMethod := oidc.TestAuthMethod(
+			t, conn, databaseWrapper, o.GetPublicId(), oidc.ActivePrivateState,
+			"alice-rp", "fido",
+			oidc.WithSigningAlgs(oidc.RS256),
+			oidc.WithIssuer(oidc.TestConvertToUrls(t, "https://www.alice.com")[0]),
+			oidc.WithApiUrl(oidc.TestConvertToUrls(t, "https://www.alice.com/callback")[0]),
+		)
+
+		// ldap setup
+		ldapAm := ldap.TestAuthMethod(t, conn, databaseWrapper, o.GetPublicId(), []string{"ldap://test"})
+		ldapManagedGroup := ldap.TestManagedGroup(t, conn, ldapAm, []string{"admin"})
+		ldapRole := iam.TestRole(t, conn, o.GetPublicId())
+		ldapMgRole := iam.TestManagedGroupRole(t, conn, ldapRole.GetPublicId(), ldapManagedGroup.GetPublicId())
+		iam.TestRoleGrant(t, conn, ldapMgRole.GetRoleId(), grant)
+
+		// do the thing
+		grantTuples, err := repo.GrantsForUser(ctx, user.PublicId)
+		require.NoError(t, err)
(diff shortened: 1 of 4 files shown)


internal/iam/repository_role_grant_test.go
```diff
@@ -1536,3 +1485,958 @@ func TestGrantsForUser(t *testing.T) {
...
+					ParentScopeId: directGrantOrg1.PublicId,
+				},
+				act:        action.Update,
+				shouldWork: true,
+			}, testCase{
+				name: "descendants-e",
+				res: perms.Resource{
+					ScopeId:       directGrantProj1a.PublicId,
+					Id:            "cs_abcd1234",
+					Type:          resource.Credential,
```
elimt, line 2368: Do we hae dedicated tests for `Groups` resource

internal/iam/repository_role_grant_ext_test.go
```diff
@@ -389,3 +389,64 @@ func TestGrantsForUserRandomized(t *testing.T) {
...
+func TestGrantsForUser_ManagedGroup(t *testing.T) {
+	ctx := context.Background()
+
+	conn, _ := db.TestSetup(t, "postgres")
+	wrap := db.TestWrapper(t)
+
+	repo := iam.TestRepo(t, conn, wrap)
+	kmsCache := kms.TestKms(t, conn, wrap)
+
+	grant := "ids=*;type=*;actions=*"
```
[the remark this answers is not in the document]
dkanney: Updated these tests to only test against the Groups resource

internal/iam/repository_role_grant_test.go
dkanney: We also have tests against the Groups resource here: https://github.com/hashicorp/boundary/blob/f62e67584e14fa4d706f3289c7483d181333aef6/internal/perms/grants_test.go

internal/iam/repository_role_grant_ext_test.go
```diff
@@ -389,3 +394,1547 @@ func TestGrantsForUserRandomized(t *testing.T) {
...
+
+	// Create a series of scopes with roles in each. We'll create two of each
+	// kind to ensure we're not just picking up the first role in each.
+
+	// The first org/project set contains direct grants, but without
+	// inheritance. We create two roles in each project.
+	directGrantOrg1, directGrantProj1a, directGrantProj1b := iam.SetupDirectGrantScopes(t, conn, repo)
+	directGrantOrg1Role := iam.TestRole(t, conn, directGrantOrg1.PublicId)
+	iam.TestUserRole(t, conn, directGrantOrg1Role.PublicId, user.PublicId)
+	directGrantOrg1RoleGrant1 := "ids=*;type=group;actions=*"
```
elimt, line 415: does this direct association only test `group` resource? The title of the test just says `TestGrantsForUser_DirectAssociation` and the grantString is has type set to  ` "ids=*;type=group;actions=*"`. When we want to test other resource, are you going to use the same test which builds a grantString for each resource or will they be separated?

internal/iam/repository_role_grant_ext_test.go
```diff
@@ -389,3 +394,1547 @@ func TestGrantsForUserRandomized(t *testing.T) {
...
+	childGrantOrg2RoleGrant2 := "ids=*;type=group;actions=delete"
+	iam.TestRoleGrant(t, conn, childGrantOrg2Role.PublicId, childGrantOrg2RoleGrant2)
+
+	// Finally, let's create some roles at global scope with children and
+	// descendants grants
+	childGrantGlobalRole := iam.TestRole(t, conn, scope.Global.String(),
+		iam.WithGrantScopeIds([]string{
+			globals.GrantScopeChildren,
+		}))
+	iam.TestUserRole(t, conn, childGrantGlobalRole.PublicId, globals.AnyAuthenticatedUserId)
```
elimt, line 986: Why not set the `userId` here instead of assigning the role to `u_auth`?

internal/iam/repository_role_grant_ext_test.go
```diff
@@ -389,3 +394,1547 @@ func TestGrantsForUserRandomized(t *testing.T) {
...
+				Grants:        descendantGrantGlobalRoleGrant,
+			},
+			{
+				RoleId:        childGrantGlobalRole.PublicId,
+				RoleScopeId:   scope.Global.String(),
+				GrantScopeIds: globals.GrantScopeChildren,
+				Grants:        childGrantGlobalRoleGrant,
+			},
+		}
+		for i, tuple := range expMultiGrantTuples {
```
elimt, line 1075: Do we assign any roles to a different user to user it does not get returned?

internal/iam/repository_role_grant_ext_test.go
```diff
@@ -389,3 +394,1547 @@ func TestGrantsForUserRandomized(t *testing.T) {
...
+		// local exploding of grants and grant scopes
+		expMultiGrantTuples := []iam.MultiGrantTuple{
+			// No grants from noOrg/noProj
+			// Direct org1/2:
+			{
+				RoleId:            directGrantOrg1Role.PublicId,
+				RoleScopeId:       directGrantOrg1.PublicId,
+				RoleParentScopeId: scope.Global.String(),
+				GrantScopeIds:     globals.GrantScopeThis,
+				Grants:            strings.Join([]string{directGrantOrg1RoleGrant1, directGrantOrg1RoleGrant2}, "^"),
```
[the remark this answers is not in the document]
dkanney: It looks like `^` is used as a delimiter to split grant strings/grant scope ids.

The `grantsForUser` query pulls in multiple grant strings/grant scope ids, aggregating them into a single string using `^` as a delimiter. In the code, we use `strings.Split(..., "^")` to split grant strings/grant scope ids into atomic pieces.

internal/iam/repository_role_grant_ext_test.go
dkanney: Yes - only the group resource for now. This function should probably be renamed to something Group-specific if we don't plan on adding subtests for the other differently scoped resources (e.g. Auth Methods, Targets).

I figured we could create a separate function for each differently scoped resource since the scope combinations and actions will be different. For instance - Auth Methods apply to `global` and `org` scopes only, so we don't need to setup any `project` scopes like we do here.

internal/iam/repository_role_grant_ext_test.go
dkanney: Oof, I didn't notice I was assigning `u_auth` to the role here instead of the test user I created.

Pushed a change to use `user.PublicId` instead of `u_auth`

internal/iam/repository_role_grant_ext_test.go
dkanney: That's a good test case - I'll add another user with different grants to each test

dkanney: OK -- I added a second user/group/managed group to each test. I assign them some grants & check that their grants are separate from the grants assigned to the original user/group/managed group

internal/iam/repository_role_grant_ext_test.go
```diff
@@ -389,3 +394,2026 @@ func TestGrantsForUserRandomized(t *testing.T) {
...
+				},
+			}
+			/*
+				log.Println("org1", directGrantOrg1.PublicId)
+				log.Println("proj1a", directGrantProj1a.PublicId)
+				log.Println("proj1b", directGrantProj1b.PublicId)
+				log.Println("org2", directGrantOrg2.PublicId)
+				log.Println("proj2a", directGrantProj2a.PublicId)
+				log.Println("proj2b", directGrantProj2b.PublicId)
+			*/
```
elimt, line 1048: ~ remove print statement

internal/iam/repository_role_grant_ext_test.go
```diff
@@ -389,3 +394,2026 @@ func TestGrantsForUserRandomized(t *testing.T) {
...
```
elimt, line 1719: ~ remove print statement

internal/iam/repository_role_grant_ext_test.go
```diff
@@ -389,3 +394,2026 @@ func TestGrantsForUserRandomized(t *testing.T) {
...
```
elimt, line 2412: remove print statement

internal/iam/repository_role_grant_test.go
```diff
@@ -801,11 +750,11 @@ func TestGrantsForUser(t *testing.T) {
...
+			tuple.TestStableSort()
 			expMultiGrantTuples[i] = tuple
 		}
-		multiGrantTuplesCache := new([]multiGrantTuple)
-		_, err := repo.GrantsForUser(ctx, user.PublicId, withTestCacheMultiGrantTuples(multiGrantTuplesCache))
+		multiGrantTuplesCache := new([]MultiGrantTuple)
+		_, err := repo.GrantsForUser(ctx, user.PublicId, WithTestCacheMultiGrantTuples(multiGrantTuplesCache))
 		require.NoError(t, err)
 
 		// log.Println("multiGrantTuplesCache", pretty.Sprint(*multiGrantTuplesCache))
```
elimt, line 760: remove print statement

internal/iam/repository_role_grant_test.go
```diff
@@ -1536,3 +1485,961 @@ func TestGrantsForUser(t *testing.T) {
...
+		}
+		for i, tuple := range expMultiGrantTuples {
+			tuple.TestStableSort()
+			expMultiGrantTuples[i] = tuple
+		}
+		multiGrantTuplesCache := new([]MultiGrantTuple)
+		_, err := repo.GrantsForUser(ctx, user.PublicId, WithTestCacheMultiGrantTuples(multiGrantTuplesCache))
+		require.NoError(t, err)
+
+		// log.Println("multiGrantTuplesCache", pretty.Sprint(*multiGrantTuplesCache))
```
elimt, line 1674: remove print statement

internal/iam/repository_role_grant_test.go
```diff
@@ -1536,3 +1485,961 @@ func TestGrantsForUser(t *testing.T) {
...
+				},
+			}
+			/*
+				log.Println("org1", directGrantOrg1.PublicId)
+				log.Println("proj1a", directGrantProj1a.PublicId)
+				log.Println("proj1b", directGrantProj1b.PublicId)
+				log.Println("org2", directGrantOrg2.PublicId)
+				log.Println("proj2a", directGrantProj2a.PublicId)
+				log.Println("proj2b", directGrantProj2b.PublicId)
+			*/
```
elimt, line 1994: remove print statement

internal/iam/repository_role_grant_ext_test.go
```diff
@@ -389,3 +394,2026 @@ func TestGrantsForUserRandomized(t *testing.T) {
...
+		for userId, tuples := range expMultiGrantTuples {
+			for i, tuple := range tuples {
+				tuple.TestStableSort()
+				expMultiGrantTuples[userId][i] = tuple
+			}
+			multiGrantTuplesCache := new([]iam.MultiGrantTuple)
+			_, err := repo.GrantsForUser(ctx, userId, iam.WithTestCacheMultiGrantTuples(multiGrantTuplesCache))
+			require.NoError(t, err)
+
+			// log.Println("multiGrantTuplesCache", pretty.Sprint(*multiGrantTuplesCache))
```
elimt, line 727: remove print statement

internal/iam/repository_role_grant_ext_test.go
```diff
@@ -389,3 +394,2026 @@ func TestGrantsForUserRandomized(t *testing.T) {
 			", roles from ldap managed groups", rolesFromLdapManagedGroups)
 	}
 }
+
+func TestGrantsForUser_DirectAssociation(t *testing.T) {
```
elimt, line 398: The general feedback I have on the tests is the setup for the test. It looks like we setup everything in the beginning and then invoke tests below for assertions. I think because we're setting up everything before hand, we have to create a lot of roles and grants.

What do you think about grouping this some more? Maybe have a table test with a setup function like we have [here](https://github.com/hashicorp/boundary/blob/e8bc0748febf76139253650ef3e64a405e9ad267/internal/authtoken/repository_test.go#L938). In the `setup`, we can create roles relevant for that function and name the test as well so it's easier for anyone looking at it to follow and understand the test coverage we have. Maybe there might be another way of grouping this better.

I also see some overlap here with @bosorawis [PR](https://github.com/hashicorp/boundary/pull/5403) where Bo has created some helper functions for creating roles and assigning grants to it. For example [this](https://github.com/hashicorp/boundary/blob/52b0decf96e78944f32aed8e3dfd74c826d642f5/internal/daemon/controller/handlers/groups/grants_test.go#L88-L94) for Direct Association, [this](https://github.com/hashicorp/boundary/blob/52b0decf96e78944f32aed8e3dfd74c826d642f5/internal/daemon/controller/handlers/groups/grants_test.go#L843-L853) for Group Association and [this](https://github.com/hashicorp/boundary/blob/52b0decf96e78944f32aed8e3dfd74c826d642f5/internal/daemon/controller/handlers/groups/grants_test.go#L103-L109) for Managed Groups. Bo's PR is still under review and subject to change but I think we could share and use some of those helper functions here to reduce the code we have here and may decide to set things up slightly differently if necessary.

dkanney: > What do you think about grouping this some more? Maybe have a table test with a setup function like we have [here](https://github.com/hashicorp/boundary/blob/e8bc0748febf76139253650ef3e64a405e9ad267/internal/authtoken/repository_test.go#L938). In the setup, we can create roles relevant for that function and name the test as well so it's easier for anyone looking at it to follow and understand the test coverage we have.

Pushed what I have so far -- https://github.com/hashicorp/boundary/commit/4c919f41f5ac83fbbb64bc40559c0a30798cdf81

Grouped the role granting logic & role-user association call together into a helper function:

```go
func grantRoleAndAssociate(t *testing.T, conn *db.DB, roleId string, roleAssociationFunc func(), grants ...string) {
	t.Helper()
	for _, grant := range grants {
		iam.TestRoleGrant(t, conn, roleId, grant)
	}
	roleAssociationFunc()
}
```

___

> The general feedback I have on the tests is the setup for the test. It looks like we setup everything in the beginning and then invoke tests below for assertions. I think because we're setting up everything before hand, we have to create a lot of roles and grants.
> ...
> Maybe there might be another way of grouping this better.

I don't think I can group things together much more with the current test structure. Current top-level test setup:
- Create a few scopes
- Create a couple roles at each scope
- Grant each role
- Associate a user to each role
- Create test cases. For each test case:
    - Do a single, big diff/comparison

If I refactor to look more like the other tests, they might look like this:
- Create a few scopes
- Create test cases. For each test case:
    - Create a role at a scope
    - Grant the role
    - Associate a user to the role
    - Do a single, small diff/comparison
  
This might be easier to follow since we'd be able to name each test case after the relationship (e.g. `t.Run("direct grant org 1 role 1", ...)` , `t.Run("direct grant project 1a role 1", ...)`, etc.). However, I think we'd lose the ability to assign different roles to multiple users in a test case (mentioned in an earlier [comment](https://github.com/hashicorp/boundary/pull/5443#discussion_r1938185689)). There might be a way around this with maps, albeit with added complexity

elimt: I see what you mean with the way the tests are setup. I think this already looks better. Grouping it might be complicated after looking at the tests some more

elimt (approved):

internal/iam/repository_role_grant_ext_test.go
dkanney: Yeah, after looking at this more I realized my tests assert on the created role's `PublicId` (via `GrantTuple.RoleId`) whereas `TestUserDirectGrantsFunc()` doesn't return the id(s) its created role(s)

PR #5444, opened 2025-01-21
chore: Update minio plugin version

louisruch:


johanbrandhorst (approved):

hugoghx (approved):
LGTM! Rotation seems to work as it did before 😄

moduli (approved):
Failures in `test-race` look like they currently exist. We can also verify all of the e2e tests pass once this gets into boundary-enterprise.

PR #5451, opened 2025-01-22
fix(authtokens): set parent scope ID for auth token resource

bosorawis:
Set Scope ID for resources when performing final auth check in `newOutputOpt`


johanbrandhorst (approved):

ddebko (changes requested):
There is a small linter issue with the import order for the grants_test file.

PR #5452, opened 2025-01-22
fix(roles): set parent scope id for roles resource

elimt:
set the `ParentScopeId` before fetching authorized actions for role resource


johanbrandhorst (approved):
Assuming linter warnings are fixed

ddebko (approved):

tmessi (approved):

PR #5464, opened 2025-01-22
Update azure plugin to support VMSS

johanbrandhorst:
v0.2.0 of the Azure plugin adds support for Virtual Machine Scale Sets (VMSS) in addition to Virtual Machines.


johanbrandhorst:
Moving this to draft as we're not planning to release it with 0.19.0 anymore.

mikemountain (approved):

PR #5479, opened 2025-01-27
Update index.mdx, added note about credentials command

vanphan24:
Added note about `Boundary credentials` command not supporting vault cred store


Dan-Heath (approved):
LGTM, thank you @vanphan24 !

PR #5486, opened 2025-01-28
WIP: SQL Define new grants tables

elimt:
Create new tables for grants:

1. `iam_role_global`:
  Roles that are placed in the global scope will be persisted in the `iam_role_global` table. A global role has a `grant_scope` which must be one of:
  * descendants
  * children
  * individual

  This enforces that a global role's grants either apply to:
  *  All orgs and projects.
  *  All orgs.
  *  An individual set of orgs and/or projects.

  When the `grant_scope` is set to `individual`, entries for the specific set of orgs and/or projects can be added to the 
  `iam_role_global_individual_grant_scope` table.

  Separately, a global role can be set to also apply its grant to the global scope by setting `grant_this_role_scope` to true.

2. `iam_role_org`:
  Roles that are placed in an org scope will be persisted in the `iam_role_org` table. An org role has a `grant_scope` which must be one of:
  * children
  * individual

  This enforces that an org role's grants either apply to:
  * All projects in the org.
  * An individual set of projects in the org.

  When the `grant_scope` is set to `individual`, entries for the specific set of projects can be added to the 
  `iam_role_org_individual_grant_scope` table.
  **NOTE**: The projects must belong to the org's scope

  Separately, an org role can be set to also apply its grant to the org by setting `grant_this_role_scope` to true.

3. `resource_enm`: 
  Contains all boundary resources. This is used by `iam_grant` to set the resource from a canonical_grant.

3. `iam_grant`
Stores the canonical grant string and the resource for filtering on specific grants.


internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
```diff
@@ -0,0 +1,133 @@
...
+      foreign key (role_id, grant_scope)
+      references iam_role_global(public_id, grant_scope),
+    create_time wt_timestamp
+  );
+
+  create trigger default_create_time_column before insert on iam_role_global_individual_grant_scope
+  for each row execute procedure default_create_time();
+
+  create trigger immutable_columns before update on iam_role_global_individual_grant_scope
+  for each row execute procedure immutable_columns('scope_id', 'create_time');
```
bosorawis, line 131: I think every column in this table is immutable, not sure if we need constraints for all of them though 🤔

internal/db/schema/migrations/oss/postgres/100/02_iam_role_org.up.sql
```diff
@@ -0,0 +1,132 @@
...
+      foreign key (role_id, grant_scope)
+      references iam_role_org(public_id, grant_scope),
+    create_time wt_timestamp
+  );
+
+  create trigger default_create_time_column before insert on iam_role_org_individual_grant_scope
+  for each row execute procedure default_create_time();
+
+  -- ensure the project's parent is the role's scope
+  create or replace function ensure_project_belongs_to_org() returns trigger
```
bosorawis, line 93: nitpick: 
```suggestion
  create or replace function ensure_project_belongs_to_role_org() returns trigger
```

internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
```diff
@@ -0,0 +1,163 @@
...
+    grant_scope text
+      constraint iam_role_global_grant_scope_enm_fkey
+        references iam_role_global_grant_scope_enm(name)
+        on delete restrict
+        on update cascade,
+    version wt_version,
+    grant_this_role_scope_update_time wt_timestamp,
+    grant_scope_update_time wt_timestamp,
+    create_time wt_timestamp,
+    update_time wt_timestamp,
```
elimt, line 98: I realized we have a pattern of using triggers for setting the default timestamp
```
create trigger default_create_time_column before insert on iam_user_role
  for each row execute procedure default_create_time();
``` 
could we have a `default now()` instead of a trigger?

elimt, line 98: We had to separate the timestamps to match what we currently do for iam_role tables so we do not change the behaviour.

- `grant_this_role_scope_update_time`: Is only set when `grant_this_role_scope` is set or updated
- `grant_scope_update_time`: Is only updated when `grant_scope` is set or updated
- `create_time`: when the the role is initially created
- `update_time`: when the name or the description is updated

internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
```diff
@@ -0,0 +1,163 @@
...
+    end if;
+    return new;
+  end;
+  $$ language plpgsql;
+
+  create or replace function insert_grant_this_role_scope_update_time() returns trigger
+  as $$
+  begin
+    if new.grant_this_role_scope is distinct from old.grant_this_role_scope then
+      new.grant_scope_update_time = now();
```
mgaffney, line 49: Should this be `grant_this_role_scope_update_time`?

internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
```diff
@@ -0,0 +1,163 @@
...
+    return new;
+  end;
+  $$ language plpgsql;
+
+  -- Add trigger to update the new column on every iam_role subtype update.
+  -- This is used to update the update_time of the iam_role table
+  -- when either the name or the description of the subtype tables are updated.
+  -- This is only applicable to the name and description columns because we
+  -- do not want the update_time to be updated when the grant_scope or grant_this_role_scope
+  -- columns are updated.
```
mgaffney, line 60: Are we sure about this? I thought we wanted the `update_time` to be updated no matter what changed. If we do what it says here, then the `update_time` column would behave differently from all of our other `update_time` columns. @tmessi do you have any thoughts on this?

internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
```diff
@@ -0,0 +1,163 @@
...
+    end if;  
+  end;
+  $$ language plpgsql;
+  comment on function update_iam_role_table_update_time() is
+    'update_iam_role_table_update_time is used to automatically update the update_time '
+    'of the base table whenever one of the subtype tables are updated';
+
+  -- global iam_role must have a scope_id of global
+  create table iam_role_global (
+    public_id wt_role_id not null primary key
```
mgaffney, line 76:
```suggestion
    public_id wt_role_id primary key
```
It is not necessary to declare a primary key as `not null` in PostgreSQL

internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
```diff
@@ -0,0 +1,163 @@
...
+      constraint iam_role_fkey
+        references iam_role(public_id)
+        on delete cascade
+        on update cascade,
+    scope_id wt_scope_id
```
mgaffney, line 81: When can the `scope_id` be null?

internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
```diff
@@ -0,0 +1,163 @@
...
+      constraint iam_scope_global_fkey
+        references iam_scope_global(scope_id)
+        on delete cascade
+        on update cascade,
+    name text,
+    description text,
+    grant_this_role_scope boolean not null default false,
+    grant_scope text
```
mgaffney, line 89: When can the `grant_scope` be null?

mgaffney (changes requested):

internal/db/schema/migrations/oss/postgres/100/05_iam_grant.up.sql
```diff
@@ -0,0 +1,39 @@
...
+  -- A grant can only reference a single resource, including the special
+  -- strings "*" to indicate "all" resources, and "unknown" when no resource is set.
+  create table iam_grant (
+    canonical_grant text not null primary key,
+    resource text not null
+      constraint resource_enm_fkey
+        references resource_enm(string)
+        on delete restrict
+        on update cascade
+  );
```
mgaffney, line 16: Shouldn't the `iam_role_grant` table have a foreign key to this table?

internal/db/sqltest/Makefile
```diff
@@ -17,28 +17,29 @@ endif
...
-		 tests/auth/*/*.sql \
-		 tests/purge/*.sql \
+		 tests/host/*.sql \
+		 tests/iam/*.sql \
+		 tests/kms/*.sql \
+		 tests/org/*.sql \
 		 tests/pagination/*.sql \
 		 tests/policy/*.sql \
-		 tests/host/*.sql \
-		 tests/server/*.sql
```
mgaffney, line 41: Why change the order of the existing tests?

internal/db/sqltest/tests/iam/iam_role_org.sql
```diff
@@ -0,0 +1,2 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
```
mgaffney, line 2: Can we remove this (almost) empty file?

internal/db/schema/migrations/oss/postgres/100/04_resource.up.sql
```diff
@@ -0,0 +1,38 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+
+  create table resource_enm (
+    string text not null primary key
```
mgaffney, line 7: 1. I think `string` is a poor column name since it is a primitive type in many languages and it doesn't convey what the values of the column represent. 
2. Shouldn't there be a constraint on this column since this is an enum table?

internal/db/sqltest/Makefile
bosorawis: I changed it to match the alphabetical order - similar to how these directories show up in the IDEs to try to make it easier to glance if any directory was missed.

internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
tmessi: The trigger ensures that `create_time` is set to `now()` even if it was explicitly set to some other timestamp as part of the `insert` call. A `default now()` would only be used if the column was *not* explicitly set during an `insert`. So the trigger guards against mistakes from the application.

tmessi: See the [comment above](https://github.com/hashicorp/boundary/pull/5486/files/6f0ae35dcc9182c225f4d37af8e4027f1a6844f2#r1933945495) about `update_time`, it should update when there is any change to the role aggregate.
I would also recommend including some comments about the purpose of these other `_update_time` columns directly in the migration to provide context for others in the future.

internal/db/schema/migrations/oss/postgres/100/05_iam_grant.up.sql
elimt: Actually there should be one on the `iam_role_grant` table to ensure `canonical_grant` can't get updated on only `iam_role_grant` table. I'll add that

internal/db/sqltest/tests/iam/iam_role_global.sql
```diff
@@ -0,0 +1,157 @@
...
+  select plan(14);
+
+    -- Set up a scope  to test against
+  insert into iam_scope
+    (type,      public_id,      parent_id)
+  values
+    ('org',     'o_1111111111', 'global');
+
+  -- remove any existing roles and their grant scopes, so we can test fresh
+  delete from iam_role;
```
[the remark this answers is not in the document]
elimt: Took that out. I switched to using the personas

internal/db/sqltest/tests/iam/iam_role_global.sql
```diff
@@ -0,0 +1,157 @@
...
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+  select plan(14);
+
+    -- Set up a scope  to test against
+  insert into iam_scope
+    (type,      public_id,      parent_id)
+  values
+    ('org',     'o_1111111111', 'global');
```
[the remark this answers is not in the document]
elimt: I switched to using the personas by calling
```
wtt_load('widgets', 'iam');
```

I have not found a need to load only specific parts of `iam` widget yet. Once necessary, I can split it up

internal/db/schema/migrations/oss/postgres/100/02_iam_role_org.up.sql
```diff
@@ -0,0 +1,139 @@
...
+  -- ensure the project's parent is the role's scope
+  create function ensure_project_belongs_to_role_org() returns trigger
+  as $$
+  begin
+    perform
+      from iam_scope_project
+      join iam_role_org 
+        on iam_role_org.scope_id = iam_scope_project.parent_id 
+      where iam_scope_project.scope_id = new.scope_id
+        and iam_role_org.public_id = new.role_id;
```
tmessi, line 126:
```suggestion
    perform
       from iam_scope_project
       join iam_role_org 
         on iam_role_org.scope_id      = iam_scope_project.parent_id
      where iam_scope_project.scope_id = new.scope_id
        and iam_role_org.public_id     = new.role_id;
```

internal/db/sqltest/initdb.d/01_colors_persona.sql
```diff
@@ -139,15 +139,15 @@ begin;
...
-    ('r_pp_gc__mix', 'type=color;action=mix',     'mix color');
+    ('r_go____name', 'type=group;action=name',    'name colors'),
+    ('r_gp____spec', 'type=group;action=inspect', 'inspect colors'),
+    ('r_oo_____art', 'type=group;action=create',  'create color'),
+    ('r_op_bc__art', 'type=group;action=create',  'create color'),
+    ('r_op_rc__art', 'type=group;action=create',  'create color'),
+    ('r_op_gc__art', 'type=group;action=create',  'create color'),
+    ('r_pp_bc__mix', 'type=group;action=mix',     'mix color'),
+    ('r_pp_rc__mix', 'type=group;action=mix',     'mix color'),
+    ('r_pp_gc__mix', 'type=group;action=mix',     'mix color');
```
tmessi, line 150: We should fix these grants to be valid grants. So they should be `actions` not `action` and should use valid actions like `create`, `read`, `list`, `update` etc.

internal/db/sqltest/initdb.d/03_widgets_persona.sql
```diff
@@ -77,12 +79,12 @@ begin;
...
-      ('r_op_sw__eng', 'type=widget;action=tune',   'tune widget'),
-      ('r_op_sw__eng', 'type=widget;action=clean',  'clean widget'),
-      ('r_pp_bw__bld', 'type=widget;action=build',  'build widget'),
-      ('r_pp_sw__bld', 'type=widget;action=build',  'build widget');
+      ('r_oo_____eng', 'type=target;action=design', 'design widget'),
+      ('r_op_sw__eng', 'type=target;action=design', 'design widget'),
+      ('r_op_sw__eng', 'type=target;action=tune',   'tune widget'),
+      ('r_op_sw__eng', 'type=target;action=clean',  'clean widget'),
+      ('r_pp_bw__bld', 'type=target;action=build',  'build widget'),
+      ('r_pp_sw__bld', 'type=target;action=build',  'build widget');
```
tmessi, line 87: Same here, lets make these realistic grants.

internal/db/sqltest/tests/iam/iam_grant.sql
```diff
@@ -0,0 +1,87 @@
...
+
+begin;
+select plan(9);
+select wtt_load('widgets', 'iam');
+
+prepare insert_grant_scope as
+  insert into iam_grant
+    (canonical_grant)
+  values
+    ('type=scope;others=stuff;');
```
tmessi, line 12: I think we need to have tests for other kinds of valid and invalid grants. Some that come to mind:
- Grants with an `ids` component, including wildcard, single, multiple ids.
- Grants with `output_fields`
- Grants with multiple actions.
- Grants with wildcard type.

tmessi (changes requested):

internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
elimt: If we set `grant_scope` to `individual` & `grant_this_role_scope` to `true` when grant scope uses the special keyword `this`, wouldn't that be understood differently by anyone reading the data back from the DB?

We'll always query the `iam_role_global_individual_grant_scope` table when `grant_scope` is set to `individual` correct? Is that expected?

internal/db/sqltest/tests/iam/iam_grant.sql
elimt: Okay just added some more test coverage for this

internal/db/schema/migrations/oss/postgres/100/02_iam_role_org.up.sql
```diff
@@ -0,0 +1,139 @@
...
+      constraint iam_role_org_grant_scope_enm_fkey
+        references iam_role_org_grant_scope_enm(name)
+        on delete restrict
+        on update cascade,
+    version wt_version,
+    grant_this_role_scope_update_time wt_timestamp,
+    grant_scope_update_time wt_timestamp,
+    create_time wt_timestamp,
+    update_time wt_timestamp,
+    unique(public_id, grant_scope)
```
mgaffney, line 50: Can you name this unique constraint with a name that follows our standard naming convention for unique constraints? Our unique constraint naming convention is to concatenate the table name with each column in the unique constraint and then add the suffix `uq`. (ex: `table_name_col1_col2_colN_uq`)

**Example** shows a foreign key constraint and a unique constraint:

```sql
create table widget_shelf (
  public_id wt_public_id primary key,
  scope_id wt_scope_id not null
    constraint iam_scope_fkey
      references iam_scope (public_id)
      on delete cascade
      on update cascade,
  name wt_name,
  constraint widget_shelf_scope_id_name_uq
    unique(scope_id, name)
);
```

Also, consider reversing the order of the columns

internal/db/schema/migrations/oss/postgres/100/02_iam_role_org.up.sql
```diff
@@ -0,0 +1,139 @@
...
+      constraint iam_role_org_fkey
+        references iam_role_org(public_id)
+        on delete cascade
+        on update cascade,
+    -- grant_scope is used for constraint checking.
+    -- This restricts the grant_scope to be 'individual'
+    -- and since it is also a foreign key to the iam_role_org
+    -- grant_scope, it ensures that iam_role_org is set to 'individual'
+    -- if this table is populated for the corresponding role.
+    grant_scope text
```
mgaffney, line 93: Also missing a `not null`

internal/db/schema/migrations/oss/postgres/100/02_iam_role_org.up.sql
```diff
@@ -0,0 +1,139 @@
...
+            grant_scope = 'individual'
+        ),
+    scope_id wt_scope_id not null
+      constraint iam_scope_org_scope_id_fkey
+        references iam_scope_project(scope_id)
+        on delete cascade
+        on update cascade,
+    constraint iam_role_org_grant_scope_fkey 
+      foreign key (role_id, grant_scope)
+      references iam_role_org(public_id, grant_scope),
```
mgaffney, line 105: Should this have `on delete` and `on cascade` statements?

internal/db/schema/migrations/oss/postgres/100/04_resource.up.sql
```diff
@@ -0,0 +1,71 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+
+  create table resource_enm (
```
mgaffney, line 6: Can we add rename this to something like `iam_resource_enm` or `iam_grant_resource_enm`? 
I'm not sure of the exact name but I think it at least needs the `iam_` prefix since that is the subsytem/package that it most belongs with. @tmessi thoughts?

mgaffney (changes requested):

internal/db/sqltest/tests/iam/iam_grant.sql
```diff
@@ -0,0 +1,147 @@
...
+  'primary key (canonical_grant) is enforced'
+);
+
+-- insert a canonical grant string with wildcards for id, type, actions, and output_fields
+-- validate that the resource is set to '*'
+prepare insert_grant_wildcard_actions as
+  insert into iam_grant
+    (canonical_grant)
+  values
+    ('id=*;type=*;actions=*;output_fields=*');
```
tmessi, line 91: The canonical format of a grant uses `ids` instead of `id`.

internal/db/sqltest/tests/iam/iam_grant.sql
tmessi: I think there are still additional combinations of grants to try, some that come to mind are:

- `ids=*;type=*;actions=*`
- Some with `type` that includes dashes like `host-catalog`.
- Some malformed grant strings like `ids=*;type;=target;actions=*` or `ids=*;;type=target;actions=*`.

elimt: I added some more tests for this

internal/db/sqltest/tests/iam/iam_grant.sql
```diff
@@ -0,0 +1,279 @@
...
+  values
+    ('ids=*;type;=credential-library;actions=create;output_fields=id;');
+select lives_ok('insert_grant_malformed_type_with_semicolon_after_type');
+select is(
+  (select resource
+     from iam_grant
+    where canonical_grant = 'ids=*;type;=credential-library;actions=create;output_fields=id;'),
+  'unknown',
+  'resource should default to "unknown" if the type is not found'
+);
```
tmessi, line 276: I don't think these malformed grants should result in a row being inserted. I would expect an error in these cases.

tmessi (changes requested):

internal/db/sqltest/tests/iam/iam_grant.sql
elimt: We do not do any validation of the grant string. We only try to get the type for resource so it won't throw an error. It will just default to unknown if it cannot be parsed

```
  create function set_resource() returns trigger
  as $$
  declare resource text[];
  begin
    select regexp_matches(new.canonical_grant, 'type=([^;]+);')
    into resource;
    if resource is null then
      new.resource = 'unknown';
    else
      new.resource = resource[1];
    end if;
    return new;
  end
  $$ language plpgsql;
  comment on function set_resource() is
    'set_resource is a trigger function that sets the resource column based on the canonical_grant.';
```

elimt: Unless you are saying we should validate the `canonical_grant` string on inserts?

tmessi: Yea I think we might want some validation here. While the application *should* only send well-formed canonical grants to the database layer, for this trigger to be successful we need to be sure we have a valid canonical grant. The case were we set the `resource` to `unknown` should only be in the case where we have a specific kind of grant, one that is still a valid grant but omits the `type` component. I wouldn't want us to persist one of these malformed grants and then treat it like we will other grants with resource of `unknown`.

elimt: Okay I just updated the tests and `set_resource` function to fail if a malformed `canonical_grant` is passed

internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
```diff
@@ -0,0 +1,178 @@
...
+  create trigger update_time_column before update on iam_role_global
+    for each row execute procedure update_time_column();
+
+  create trigger update_version_column after update on iam_role_global 
+    for each row execute procedure update_version_column();
+
+  create trigger immutable_columns before update on iam_role_global
+    for each row execute procedure immutable_columns('scope_id', 'create_time');
+
+  create table iam_role_global_individual_grant_scope (
```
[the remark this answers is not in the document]
elimt: The primary key should be the role_id and scope_id. I just pushed a commit to add this

internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
```diff
@@ -0,0 +1,179 @@
...
+          scope_id != 'global'
+        ),
+    constraint iam_role_global_grant_scope_fkey
+      foreign key (role_id, grant_scope)
+      references iam_role_global(public_id, grant_scope)
+      on delete cascade
+      on update cascade,
+    create_time wt_timestamp,
+    primary key(role_id, scope_id)
+  );
```
mgaffney, line 169:
```suggestion
  create table iam_role_global_individual_grant_scope (
    role_id wt_role_id
      constraint iam_role_global_fkey
        references iam_role_global(public_id)
        on delete cascade
        on update cascade,
    scope_id wt_scope_id not null
      constraint iam_scope_fkey
        references iam_scope(public_id)
        on delete cascade
        on update cascade
      constraint scope_id_is_not_global
        check(
          scope_id != 'global'
        ),
    -- grant_scope is used for constraint checking.
    -- This restricts the grant_scope to be 'individual'
    -- and since it is also a foreign key to the iam_role_global
    -- grant_scope, it ensures that iam_role_global is set to 'individual'
    -- if this table is populated for the corresponding role.
    grant_scope text not null
       constraint only_individual_grant_scope_allowed
         check(
          grant_scope = 'individual'
        ),
    constraint iam_role_global_grant_scope_fkey
      foreign key (role_id, grant_scope)
      references iam_role_global(public_id, grant_scope)
      on delete cascade
      on update cascade,
    create_time wt_timestamp,
    primary key(role_id, scope_id)
  );
```
This may seem like a nit but when creating a table the order of the columns aids in readability. In this case putting the two columns that are the primary key really helps.

internal/db/schema/migrations/oss/postgres/100/02_iam_role_org.up.sql
```diff
@@ -0,0 +1,143 @@
...
+        on delete cascade
+        on update cascade,
+    constraint iam_role_org_grant_scope_fkey 
+      foreign key (role_id, grant_scope)
+      references iam_role_org(public_id, grant_scope)
+        on delete cascade
+        on update cascade,
+    create_time wt_timestamp,
+    primary key(role_id, scope_id)
+  );
```
mgaffney, line 111:
```suggestion
  create table iam_role_org_individual_grant_scope (
    role_id wt_role_id
      constraint iam_role_org_fkey
        references iam_role_org(public_id)
        on delete cascade
        on update cascade,
    scope_id wt_scope_id not null
      constraint iam_scope_org_scope_id_fkey
        references iam_scope_project(scope_id)
        on delete cascade
        on update cascade,
    -- grant_scope is used for constraint checking.
    -- This restricts the grant_scope to be 'individual'
    -- and since it is also a foreign key to the iam_role_org
    -- grant_scope, it ensures that iam_role_org is set to 'individual'
    -- if this table is populated for the corresponding role.
    grant_scope text not null
      constraint only_individual_grant_scope_allowed
        check(
            grant_scope = 'individual'
        ),
    constraint iam_role_org_grant_scope_fkey
      foreign key (role_id, grant_scope)
      references iam_role_org(public_id, grant_scope)
        on delete cascade
        on update cascade,
    create_time wt_timestamp,
    primary key(role_id, scope_id)
  );
```
Same here

mgaffney (changes requested):

internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
elimt: Okay I just updated the order

internal/db/schema/migrations/oss/postgres/100/02_iam_role_org.up.sql
elimt: Okay I just updated the order

internal/db/schema/migrations/oss/postgres/100/05_iam_grant.up.sql
```diff
@@ -0,0 +1,70 @@
...
+-- SPDX-License-Identifier: BUSL-1.1
+
+begin;
+
+  -- wt_canonical_grant domain represents Boundary canonical grant.
+  -- A canonical grant is a semicolon-separated list of key=value pairs.
+  -- e.g. "id=*;type=role;action=read;output_fields=id,name"
+  create domain wt_canonical_grant as text
+    check(
+      value ~ '^(?:[^;=]+=[^;=]+)(?:;[^;=]+=[^;=]+)*;?$'
```
tmessi, line 11: It looks like this regex allows for trailing `;` in the canonical grants, which I don't think is correct. It also looks like several of the sqltests include a trailing `;` for success cases.

internal/db/sqltest/initdb.d/01_colors_persona.sql
```diff
@@ -135,19 +135,19 @@ begin;
...
+    ('r_gg____shop', 'type=*;actions=read',                       'type=*;actions=read'),
+    ('r_go____name', 'type=group;action=create,update,read,list', 'type=group;action=create,update,read,'),
+    ('r_gp____spec', 'type=group;action=delete',                  'type=group;action=delete'),
+    ('r_oo_____art', 'type=group;action=create',                  'type=group;action=create'),
+    ('r_op_bc__art', 'type=group;action=create',                  'type=group;action=create'),
+    ('r_op_rc__art', 'type=group;action=create',                  'type=group;action=create'),
+    ('r_op_gc__art', 'type=group;action=create',                  'type=group;action=create'),
+    ('r_pp_bc__mix', 'type=group;action=add-members',             'type=group;action=add-members'),
+    ('r_pp_rc__mix', 'type=group;action=set-members',             'type=group;action=set-members'),
+    ('r_pp_gc__mix', 'type=group;action=delete-members',          'type=group;action=delete-members');
```
tmessi, line 150: Some of these grants still are not valid. It should be `actions` not `action`. And if the actions are id actions, i.e. actions other than `create` or `list`, there should be an `ids` component to the grant.

internal/db/sqltest/initdb.d/03_widgets_persona.sql
```diff
@@ -77,12 +79,12 @@ begin;
...
-      ('r_pp_bw__bld', 'type=widget;action=build',  'build widget'),
-      ('r_pp_sw__bld', 'type=widget;action=build',  'build widget');
+      ('r_oo_____eng', 'type=target;action=create,update,authorize-session',                                         'type=target;action=create,update,authorize-session'),
+      ('r_op_sw__eng', 'type=target;action=add-credential-sources,remove-credential-sources,set-credential-sources', 'type=target;action=add-credential-sources,remove-credential-sources,set-credential-sources'),
+      ('r_op_sw__eng', 'type=target;action=add-host-sources,remove-host-sources,set-host-sources',                   'type=target;action=add-host-sources,remove-host-sources,set-host-sources'),
+      ('r_op_sw__eng', 'type=target;action=read,list',                                                               'type=target;action=read,list'),
+      ('r_pp_bw__bld', 'type=target;action=create,delete',                                                           'type=target;action=create,delete'),
+      ('r_pp_sw__bld', 'type=target;action=authorize-session',                                                       'type=target;action=authorize-session');
```
tmessi, line 87: Same here, some of these are not valid grants.

internal/db/sqltest/tests/iam/iam_grant.sql
tmessi: This grant isn't valid. It has a trailing `;` and `others` isn't a valid component of a grant.

internal/db/sqltest/tests/iam/iam_grant.sql
```diff
@@ -0,0 +1,277 @@
...
+select lives_ok('insert_grant_scope');
+select is(
+  (select resource
+     from iam_grant
+    where canonical_grant = 'type=scope;others=stuff;'),
+  'scope',
+  'resource should be set to "scope" by set_resource() trigger'
+);
+
+-- insert canonical_grant with no type
```
tmessi, line 24: I think this comment and test case need updating. There can be valid grants that have no type and should be inserted. However, this is an invalid grant in that it 1. has a trailing `;`, and 2. is not a key=value.

internal/db/sqltest/tests/iam/iam_grant.sql
```diff
@@ -0,0 +1,277 @@
...
+  values
+    ('no_type_at_all;');
+select throws_like(
+  'insert_malformed_grant',
+  'value for domain wt_canonical_grant violates check constraint "wt_canonical_grant_check"',
+  'inserting a grant that is malformed should fail'
+);
+
+-- insert canonical_grant with type=role,group
+-- validate the resource is set to 'role'
```
tmessi, line 38: This comment is incorrect, this test is validating that an error is thrown, not that the resource was set to `role`.

internal/db/sqltest/tests/iam/iam_grant.sql
```diff
@@ -0,0 +1,277 @@
...
+  'inserting a resource not in iam_grant_resource_enm should fail because type is not a single value'
+);
+
+-- the set_resource() trigger will set resource='bogus', but we did not insert 'bogus'
+-- into resource_enm, so it should fail.
+prepare insert_grant_bogus as
+  insert into iam_grant
+    (canonical_grant)
+  values
+    ('type=bogus;some=thing;');
```
tmessi, line 56: Apart from the bogus type, can you make this valid grant, i.e. `type=bogus;actions=create`

tmessi (changes requested):

internal/db/schema/migrations/oss/postgres/100/05_iam_grant.up.sql
```diff
@@ -0,0 +1,70 @@
...
+
+  -- set_resource sets the resource column based on the "type" token in the canonical_grant.
+  create function set_resource() returns trigger
+  as $$
+  declare type_matches text[];
+  begin
+    -- Extract all "type" tokens from the canonical_grant string
+    select array_agg(t[1])
+      into type_matches
+    from regexp_matches(new.canonical_grant, '(?<=^|;)type=([^;]+)(?=;|$)', 'g') as t;
```
tmessi, line 41: To avoid having a complex regex here, you might consider doing some more simple splits to break the grant string apart into a table. I think something like the following is a bit easier to understand.


```sql
with
parts (p) as (
  select p
    from regexp_split_to_table(new.canonical_grant, ';') as p
),
kv (k, v) as (
  select part[1] as k,
         part[2] as v
    from parts,
         regexp_split_to_array(parts.p, '=') as part
)
select array_agg(v)
  into type_matches
  from kv
 where k = 'type';
```

internal/db/schema/migrations/oss/postgres/100/05_iam_grant.up.sql
elimt: I thought having a trailing `;` was valid. Okay I'll fix that

tmessi: Based off of the [CanonicalString](https://github.com/hashicorp/boundary/blob/main/internal/perms/grants.go#L218-L247) method, I don't think it should ever have a trailing `;`.

internal/db/sqltest/tests/iam/iam_grant.sql
elimt: I added it but that will still throw an error because the `bogus` type is not valid

internal/db/sqltest/initdb.d/01_colors_persona.sql
elimt: Okay I pushed the changes to address this

internal/db/schema/migrations/oss/postgres/100/05_iam_grant.up.sql
elimt: Okay I replaced the regex with this. I agree this is much simpler to follow

internal/db/sqltest/tests/iam/iam_grant.sql
tmessi: Yea, it should still cause an error, but I just want to ensure the only reason for the error is the invalid resource type.

internal/db/sqltest/initdb.d/03_widgets_persona.sql
```diff
@@ -75,14 +77,14 @@ begin;
...
-      -- ('r_gg_____buy', 'type=*;action=purchase',    'purchase anything'),
-      -- ('r_gg____shop', 'type=*;action=view',        'view anything'),
-      ('r_oo_____eng', 'type=widget;action=design', 'design widget'),
-      ('r_op_sw__eng', 'type=widget;action=design', 'design widget'),
-      ('r_op_sw__eng', 'type=widget;action=tune',   'tune widget'),
-      ('r_op_sw__eng', 'type=widget;action=clean',  'clean widget'),
-      ('r_pp_bw__bld', 'type=widget;action=build',  'build widget'),
-      ('r_pp_sw__bld', 'type=widget;action=build',  'build widget');
+      -- ('r_gg_____buy', 'type=*;actions=purchase',    'purchase anything'),
+      -- ('r_gg____shop', 'type=*;actions=view',        'view anything'),
```
tmessi, line 81: Can we remove these commented out ones?

tmessi (approved):
Good work on all of this!

PR #5487, opened 2025-01-28
ci(security-scanner): Suppress openssl linux CVE

elimt:
Add [CVE-2024-13176](https://ubuntu.com/security/CVE-2024-13176) to the suppress vulnerability list as Alpine hasn't addressed the issue yet

### DIFF
--- a/.release/security-scan.hcl
+++ b/.release/security-scan.hcl
@@ -5,6 +5,16 @@ container {
 	dependencies = true
 	alpine_secdb = true
 	secrets      = false
+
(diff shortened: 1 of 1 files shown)


louisruch (commented):
The one issue with pushing this to main is the cleanup (this things tend to hang around), I wonder if we should rather make the release/0.19.x branch now and suppress the CVE there instead of main

irenarindos (approved):

moduli (approved):

louisruch (approved):

PR #5493, opened 2025-01-29
ci(security-scanner): Suppress GO-2025-3408

elimt:
There is currently not a fix for GO-2025-3408. Adding the vulnerability to the suppress list for now

The issue to address the vulnerability is currently open: https://github.com/hashicorp/yamux/issues/142

### DIFF
--- a/.release/security-scan.hcl
+++ b/.release/security-scan.hcl
@@ -12,6 +12,7 @@ container {
 		suppress {
 			vulnerabilities = [
 				"CVE-2024-13176", # openssl@3.3.2-r4
+				"GO-2025-3408", # yamux@v0.1.1
 			]
 		}
 	}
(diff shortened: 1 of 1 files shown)


irenarindos (approved):

moduli (approved):

.release/security-scan.hcl
```diff
@@ -12,6 +12,7 @@ container {
 		suppress {
 			vulnerabilities = [
 				"CVE-2024-13176", # openssl@3.3.2-r4
+				"GO-2025-3408", # yamux@v0.1.1
```
picatz, line 15: This won't suppress the vulnerability that will likely be reported in the binary artifact, since we're configuring it in the `container` block here. We should move this to the `binary` block below.

picatz (changes requested):

.release/security-scan.hcl
elimt: Moved it to the binary block

moduli (approved):

irenarindos (approved):

picatz (approved):

PR #5499, opened 2025-01-30
chore(e2e): Add note for modifying `/etc/hosts`

moduli:
This PR updates the enos readme to include a note for modifying `/etc/hosts`. This is needed when using docker-based scenarios to allow the use of container names in addresses.


amitsk6 (approved):
LGTM

PR #5515, opened 2025-02-05
docs: add note about MacOS 15.1, 15.2 incompatibility

johanbrandhorst:
On MacOS 15.1 and 15.2, some users report that all their DNS responses incur the error `sendmsg: broken pipe`


mikemountain (approved):

Dan-Heath (approved):
LGTM, thank you Johan!!

PR #5521, opened 2025-02-06
internal: ensure schema manager is closed

johanbrandhorst:
This releases the connection it holds indefinitely


johanbrandhorst:
I don't think this is related to the 0.19 issue, but I don't think it will hurt to merge this. No need to backport it to 0.19.x. though.

internal/cmd/commands/database/funcs.go
```diff
@@ -43,6 +43,7 @@ func migrateDatabase(ctx context.Context, ui cli.Ui, dialect, u string, initiali
 		}
 		return noop, 2
 	}
+	defer man.Close(ctx)
```
tmessi, line 46: Hmm, while it probably would be good to call this `Close`, I think this might actually cause a new issue. A few lines down we define `unlock` function, which will try to use `man` to unlock the exclusive lock. This function is returned and then called by the callter of `migrateDatabase`. If we do this `man.Close`, the connection will be closed prior to the attempt to call `man.ExclusiveUnlock`.

So we need to do a bit more here. Perhaps we could include the call to `man.Close` in the `unlock` function. And we would still need to correctly call close in the error case below if we are unable to get an ExclusiveLock.

tmessi (changes requested):

internal/cmd/commands/database/funcs.go
johanbrandhorst: Ah, good thinking. I'll just close this manually in each case.

internal/cmd/commands/database/funcs.go
```diff
@@ -46,24 +46,29 @@ func migrateDatabase(ctx context.Context, ui cli.Ui, dialect, u string, initiali
...
 	unlock := func() {
 		// We don't report anything since this should resolve itself anyways.
 		_ = man.ExclusiveUnlock(ctx)
+		_ = man.Close(ctx)
 	}
 
 	st, err := man.CurrentState(ctx)
 	if err != nil {
 		ui.Error(fmt.Errorf("Error getting database state: %w", err).Error())
+		_ = man.Close(ctx)
```
tmessi, line 61: This still kind of has the same issue. We will call `Close` prior to `unlock`. If you look at where we happen to call `migrateDatabase`, we always do a `defer clean()` which is what the caller calls `unlock`. We do these even before checking for an error, so we will call `unlock` even in an error case. So just adding the `Close` to `unlock` should be sufficient given how it is used.

johanbrandhorst: 🤦🏻 I didn't notice it was still returning `unlock`.

tmessi (approved):

PR #5522, opened 2025-02-06
fix: resolve small issues in db transactions

ddebko:
# Summary

This PR resolves three things:

1. Remove `context.Background()` from database transactions
2. Propagate the same database transaction reader & writer to helper functions. We need to ensure we are using a single db connection per transaction.
3. Ensure we are using the database transaction reader & writer within the anonymous function call and not the repository reader or writer.

### DIFF
--- a/internal/host/options.go
+++ b/internal/host/options.go
@@ -6,7 +6,9 @@ package host
 import (
 	"errors"
 
+	"github.com/hashicorp/boundary/internal/db"
 	"github.com/hashicorp/boundary/internal/pagination"
+	"github.com/hashicorp/boundary/internal/util"
 )
 
 // GetOpts - iterate the inbound Options and return a struct
@@ -26,6 +28,8 @@ type Option func(*options) error
 // options = how options are represented
 type options struct {
 	WithLimit              int
+	WithReader             db.Reader
+	WithWriter             db.Writer
 	WithOrderByCreateTime  bool
 	Ascending              bool
 	WithStartPageAfterItem pagination.Item
@@ -63,6 +67,22 @@ func WithStartPageAfterItem(item pagination.Item) Option {
 			return errors.New("item cannot be nil")
 		}
 		o.WithStartPageAfterItem = item
+		return nil
+	}
+}
+
+// WithReaderWriter is used to share the same database reader
+// and writer when executing sql within a transaction.
+func WithReaderWriter(r db.Reader, w db.Writer) Option {
+	return func(o *options) error {
+		if util.IsNil(r) {
+			return errors.New("reader cannot be nil")
+		}
+		if util.IsNil(w) {
+			return errors.New("writer cannot be nil")
+		}
+		o.WithReader = r
+		o.WithWriter = w
 		return nil
 	}
 }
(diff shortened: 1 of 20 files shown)


internal/host/options.go
```diff
@@ -66,3 +69,19 @@ func WithStartPageAfterItem(item pagination.Item) Option {
...
+			return errors.New("reader cannot be nil")
+		}
+		if w == nil {
+			return errors.New("writer cannot be nil")
+		}
+		o.WithReader = r
+		o.WithWriter = w
+		return nil
+	}
+}
```
johanbrandhorst, line 87: I think we have tests for this option type, do you mind adding some tests for these new options to this PR?

internal/host/plugin/options.go
```diff
@@ -162,3 +165,12 @@ func WithWorkerFilter(wf string) Option {
...
 }
+
+// WithReaderWriter is used to share the same database reader
+// and writer when executing sql within a transaction.
+func WithReaderWriter(r db.Reader, w db.Writer) Option {
+	return func(o *options) {
+		o.WithReader = r
+		o.withWriter = w
+	}
+}
```
johanbrandhorst, line 176: Ditto some basic tests

internal/server/repository_worker.go
```diff
@@ -600,7 +600,11 @@ func (r *Repository) UpdateWorker(ctx context.Context, worker *Worker, version u
 			if err != nil {
 				return errors.Wrap(ctx, err, op)
 			}
-			ret.RemoteStorageStates, err = r.ListWorkerStorageBucketCredentialState(ctx, ret.GetPublicId())
+			txRepo := Repository{
+				reader: reader,
+				writer: w,
+			}
+			ret.RemoteStorageStates, err = txRepo.ListWorkerStorageBucketCredentialState(ctx, ret.GetPublicId())
```
johanbrandhorst, line 607: Is this the pattern for accomplishing this in this package? If not, I think this would be better served by another WithReaderWriter option. What do you think?

johanbrandhorst (changes requested):

internal/host/options.go
ddebko: I added some tests.

internal/host/options.go
```diff
@@ -66,3 +69,19 @@ func WithStartPageAfterItem(item pagination.Item) Option {
 		return nil
 	}
 }
+
+// WithReaderWriter is used to share the same database reader
+// and writer when executing sql within a transaction.
+func WithReaderWriter(r db.Reader, w db.Writer) Option {
+	return func(o *options) error {
+		if r == nil {
```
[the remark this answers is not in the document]
ddebko: Good catch, I will also do the same check against the WithReader/WithWriter values when we actually want to use them.

internal/server/repository_worker.go
ddebko: I switched over to using options.

johanbrandhorst (approved):

johanbrandhorst (approved):

louisruch (approved):

PR #5531, opened 2025-02-10
chore: update version to 0.17.4

elimt:


moduli (commented):
Do we need to also update the version in `controller.swagger.json`?

moduli (approved):

johanbrandhorst:
> Do we need to also update the version in `controller.swagger.json`?

This will be updated by running `make gen`, which should fix the failing CI job

johanbrandhorst:
> > Do we need to also update the version in `controller.swagger.json`?
> 
> This will be updated by running `make gen`, which should fix the failing CI job

Ah, nvm I see they are expected failures from deprecations, ignore me.

johanbrandhorst (approved):

PR #5543, opened 2025-02-11
docs/client-agent: add warning about MacOS 15

johanbrandhorst:
MacOS 15.3 has been known to sometimes block users from using the client agent altogether. Explicitly allowing it access to incoming network connections resolves the issue.

### DIFF
--- a/website/content/docs/api-clients/client-agent.mdx
+++ b/website/content/docs/api-clients/client-agent.mdx
@@ -523,6 +523,8 @@ upgrade to MacOS version 15.3 or later.
 
 It may be neccessary to explicitly allow the `boundary-client-agent` process access to incoming network connections.
 
+From the Firewall options in System Settings, toggle the `boundary-client-agent' to **Allow incoming connections**.  
+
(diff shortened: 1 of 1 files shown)


website/content/docs/api-clients/client-agent.mdx
```diff
@@ -521,6 +521,8 @@ Follow the troubleshooting steps to understand why the Client Agent is not able
 On MacOS versions 15.1 and 15.2, the firewall may incorrectly block the Client Agent from sending DNS responses. To resolve this issue,
 upgrade to MacOS version 15.3 or later.
 
+It may be neccessary to explicitly allow the `boundary-client-agent` process access to incoming network connections.
```
stellarsquall, line 524: I would add a little more detail on where to enable this setting

```suggestion
It may be neccessary to explicitly allow the `boundary-client-agent` process access to incoming network connections.

From the Firewall options in System Settings, toggle the `boundary-client-agent' to **Allow incoming connections**.  
```

stellarsquall (approved):
Thanks for adding this!

PR #5544, opened 2025-02-13
docs: Fixes examples for alias_refresh_interval & dns_request_timeout

aditya2548:
- Issue: Client agent does not start if the given example configurations are used.
- Both the parameters are parsed using `parseutil.ParseDurationSecond`, which parses a duration from a string or numeric value into a time.Duration.
- When units are missing (such as when a numeric type is provided), the duration is assumed to be in seconds.
- Since providing a time format is more readable, updating the example to the corrected string version.
- Providing a string example configuration is more suitable than providing a number example like `alias_refresh_interval=60`; where it is automatically assumed to be seconds, and can become confusing for the consumers.

### DIFF
--- a/website/content/docs/api-clients/client-agent.mdx
+++ b/website/content/docs/api-clients/client-agent.mdx
@@ -99,14 +99,14 @@ The configuration file contains the following fields:
 
    Example:
    ```hcl
-   alias_refresh_interval=60s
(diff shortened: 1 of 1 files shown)


stellarsquall (approved):
thanks for catching these!

johanbrandhorst (approved):
Thank you!!

PR #5549, opened 2025-02-14
Restore soft deleted users on authentication

johanbrandhorst:
This PR contains a potential fix for the issue of deleted auth tokens preventing re-authentication (introduced with https://github.com/hashicorp/boundary/pull/5173). The biggest change is that previously, if a soft-deleted user authenticated, it would remain soft-deleted. Now we restore the user on authentication.

It also contains some tidying of the token error handling.


mikemountain (approved):

jimlambrt (approved):
Ty!

PR #5551, opened 2025-02-18
test(e2e): Update search test to catch bug

moduli:
This PR updates the e2e search test to catch an issue fixed by this PR: https://github.com/hashicorp/boundary/pull/5549. (it addresses an issue where the desktop client would be unable to search on close/reopen). This PR adds the following...
- log out and confirm you CANNOT use search
- while logged in, restart the cache and confirm you CAN search
- log out and restart cache. log back in and confirm you CAN search

https://hashicorp.atlassian.net/browse/ICU-16380

## Testing
Confirmed that this test fails using a build prior to the linked fix. 
Confirmed that this passes when the fix is included.


johanbrandhorst (approved):
Awesome!

PR #5552, opened 2025-02-18
Update go-plugin dep

johanbrandhorst:
Also remove import groups which I think were accidental.


irenarindos (commented):
Looks like there are some gen delta failures

johanbrandhorst:
Ah yeah the generator is updated, thanks!

irenarindos (approved):

PR #5558, opened 2025-02-20
Create function to define valid set of scopes for each resource

dkanney:


internal/types/scope/scope.go
```diff
@@ -38,3 +41,19 @@ var Map = map[string]Type{
...
+	case resource.Alias, resource.Billing, resource.Worker:
+		return []Type{Global}
+	case resource.Account, resource.AuthMethod, resource.AuthToken, resource.ManagedGroup, resource.Policy, resource.Scope, resource.SessionRecording, resource.StorageBucket, resource.User:
+		return []Type{Global, Org}
+	case resource.All, resource.Group, resource.Role:
+		return []Type{Global, Org, Project}
+	case resource.CredentialLibrary, resource.Credential, resource.CredentialStore, resource.HostCatalog, resource.HostSet, resource.Host, resource.Session, resource.Target:
+		return []Type{Project}
+	default:
+		return []Type{Unknown}
```
johanbrandhorst, line 57: What do you think about having this return an error instead? This should only happen when adding a new resource and forgetting to update this function, right? Maybe even a panic 🤔. Returning this might make it harder to track down where the problem is, but let me know if there's something I'm missing!

internal/types/scope/scope.go
```diff
@@ -3,16 +3,19 @@
...
 
 const (
-	Unknown Type = 0
-	Global  Type = 1
-	Org     Type = 2
-	Project Type = 3
+	Unknown Type = iota
+	Global
+	Org
+	Project
```
johanbrandhorst, line 18: Nice! 👌🏻

internal/iam/account.go
```diff
@@ -58,7 +59,7 @@ func (a *authAccount) VetForWrite(ctx context.Context, r db.Reader, opType db.Op
 }
 
 func (a *authAccount) validScopeTypes() []scope.Type {
-	return []scope.Type{scope.Global, scope.Org}
+	return scope.AllowedIn(resource.Account)
 }
```
johanbrandhorst, line 63: Do we still need the old `validScopeTypes` functions, or should we just replace the call sites with this new function instead?

internal/types/scope/scope.go
```diff
@@ -38,3 +41,19 @@ var Map = map[string]Type{
...
 }
+
+// AllowedIn returns the set of Scopes a Resource is allowed in.
+func AllowedIn(r resource.Type) []Type {
+	switch r {
+	case resource.Alias, resource.Billing, resource.Worker:
+		return []Type{Global}
+	case resource.Account, resource.AuthMethod, resource.AuthToken, resource.ManagedGroup, resource.Policy, resource.Scope, resource.SessionRecording, resource.StorageBucket, resource.User:
+		return []Type{Global, Org}
+	case resource.All, resource.Group, resource.Role:
```
johanbrandhorst, line 52: What is `resource.All` used for?

internal/iam/account.go
dkanney: We still need this function because `validScopeTypes` is a method of the `ResourceWithScope` interface (used in `validateScopeForWrite`)

internal/types/scope/scope.go
dkanney: I like the idea of returning an `InvalidParameter` error! Updated this function to return an error: 98fe2786a7cd34ad5426194a4c8424e38ec26851

internal/types/scope/scope.go
dkanney: `resource.All` corresponds to using the wildcard (`*`) in a grant string, e.g. `ids=hcst_1234567890;type=*;actions=create,read,update`

A couple use cases from our docs: https://developer.hashicorp.com/boundary/docs/configuration/identity-access-management/permission-grant-formats#wildcard-type

johanbrandhorst: Ah, of course. Is this function expected to be used with that input anywhere? Does this mapping need to be defined for that resource type?

internal/types/scope/scope.go
```diff
@@ -38,3 +44,20 @@ var Map = map[string]Type{
 	Org.String():     Org,
 	Project.String(): Project,
 }
+
+// AllowedIn returns the set of Scopes a known Resource type is allowed in.
+func AllowedIn(ctx context.Context, r resource.Type) ([]Type, error) {
```
elimt, line 49: I believe we wanted this to live under the resource package as a `resource.Type` method: https://github.com/hashicorp/boundary/blob/11c19d574f6ab220b481c908c64d182ec00734be/internal/types/resource/resource.go#L14

elimt: There are also some other methods in the `resource` package that are not methods on `resource.Type`. We wanted to update those as well. These methods: `Parent`, `HasChildTypes` & `TopLevelType`

internal/types/scope/scope.go
dkanney: We should never call this function on `resource.All` in `GrantsForUser` (where we'll primarily be using it), so I'm returning an error for the `resource.All` case

internal/types/scope/scope.go
dkanney: Agh, I forgot >.< sorry -- added: a4bafe120dd075cd478c6b50bf2fd6145f104f65

internal/types/scope/scope.go
johanbrandhorst: Thanks, that makes sense!

internal/types/scope/scope.go
```diff
@@ -38,3 +44,24 @@ var Map = map[string]Type{
...
+	case resource.Group, resource.Role:
+		return []Type{Global, Org, Project}, nil
+	case resource.CredentialLibrary, resource.Credential, resource.CredentialStore, resource.HostCatalog, resource.HostSet, resource.Host, resource.Session, resource.Target:
+		return []Type{Project}, nil
+	case resource.Unknown:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "unknown resource type")
+	case resource.All:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "resource type '*' is not supported")
+	default:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "invalid resource type")
```
johanbrandhorst, line 65: Nit: maybe include the value to ease debugging when this happens: 
```suggestion
		return nil, errors.New(ctx, errors.InvalidParameter, op, fmt.Sprintf("invalid resource type: resource.Type(%d)", r))
```

johanbrandhorst (changes requested):
I just remembered something, there's a comment in `internal/types/resource/resource.go` that talks about the places that need to be updated when a new resource type is added. Do we want to add something there to mention the new `AllowedIn` function?

internal/types/scope/scope.go
dkanney: Updated: ef5ae30f730fa43579755d5a863fb7a1d96ab41a

Would you suggest wrapping the value in `resource.Type()`?

johanbrandhorst: I included the type name in the error message to make it slightly easier for users, but I think this is fine.

johanbrandhorst (approved):
LGTM, I think golangci-lint has a bad cache again, don't worry about those errors.

internal/types/scope/scope_test.go
```diff
@@ -51,3 +52,112 @@ func Test_Map(t *testing.T) {
...
+			wantScopes: []Type{Global, Org},
+		},
+		{
+			resource:   resource.Worker,
+			wantScopes: []Type{Global},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.resource.String(), func(t *testing.T) {
+			assert.Equal(t, tt.wantScopes, AllowedIn(tt.resource))
```
louisruch, line 160: Quick fly-by comment: the call here does not match the function signature. Also might be good to add a new test for the `invalid resource type` error path - though you would need to change how you get the test name since the `String` wont work for an unknown type (or just test the invalid resource outside of the test cases

dkanney: Great callouts -- updated the function call here & added a test case for an invalid resource type

internal/types/resource/resource.go
```diff
@@ -142,22 +143,22 @@ var Map = map[string]Type{
 
 // Parent returns the parent type for a given type; if there is no parent, it
 // returns the incoming type
-func Parent(in Type) Type {
-	switch in {
+func (r Type) Parent() Type {
```
louisruch, line 146: Will also need to update the test in resource_test.go to update all the function signature changes in here

dkanney: Updated: 25fd85d75f82e9af09e02b84b6efaed41a9f2843

internal/types/scope/scope.go
elimt: Do you plan on moving AllowedIn to `boundary/internal/types/resource/resource.go`? That is where we wanted this to live so resources can just check which scopes they are allowed in. Not under the scope

internal/iam/account.go
```diff
@@ -57,8 +57,8 @@ func (a *authAccount) VetForWrite(ctx context.Context, r db.Reader, opType db.Op
 	return nil
 }
 
-func (a *authAccount) validScopeTypes() []scope.Type {
-	return []scope.Type{scope.Global, scope.Org}
+func (a *authAccount) getResourceType() resource.Type {
```
elimt, line 60: Is there a reason we had to update this method?

dkanney: Technically no -- I previously had `validScopeTypes()` return `scope.AllowedIn(<some resource type>)`. I'm fine with either change - can revert if that change would be easier to understand

internal/types/scope/scope.go
dkanney: If `AllowedIn()` lived in `resource` package, we'd have to declare a new `scope` domain type in the `resource` package. I didn't like the idea of declaring redundant types (`ScopeType` and `Scope`), hence this function living in the `scope` package.

`scope` indirectly imports the `resource` package as well, so we can't import `scope.Type` into the `resource` package

elimt (approved):
LGTM. Validate existing tests like under `perms` and `iam` pass

PR #5559, opened 2025-02-21
feat: add grants tests for billing resource

elimt:
add test coverage for testing grants with billing resource. This tests `monthly-active-users` action with billing.

billing does not support `output_fields` so there are no tests for that


bosorawis (approved):
nitpick: Do you mind adding a case where you grant `descendant` but not `this` to a global role as well as one with role at the org level for negative tests? 

Other than that, LGTM!

internal/daemon/controller/handlers/aliases/grants_test.go
```diff
@@ -84,6 +86,36 @@ func TestGrants_ReadActions(t *testing.T) {
 				},
 				wantErr: handlers.ApiErrorWithCode(codes.PermissionDenied),
 			},
+			{
```
bosorawis, line 89: are changes to this file accidentally committed?

bosorawis (approved):

internal/daemon/controller/handlers/aliases/grants_test.go
elimt: The tests should be for billing resource rather. Added that mistakenly. This is what happens when you're jumping from one test to another 😅. Just pushed a change to address this

PR #5566, opened 2025-02-26
Add grants tests for accounts

bosorawis:
- add tests to `accounts` package covering grants
- fix error messaging erroneously saying that a type isn't a child type of itself


johanbrandhorst (approved):

internal/daemon/controller/handlers/accounts/grants_test.go
```diff
@@ -109,3 +121,2895 @@ func TestListPassword_Grants(t *testing.T) {
 		})
 	}
 }
+
+func TestGrants_GetAccounts(t *testing.T) {
+	t.Parallel()
+	ctx := context.TODO()
```
dkanney, line 127: I've been defaulting to `context.Background()`, but I've seen `context.TODO()` used in some places.

Which one should we align on, `TODO()` or `Background()`?

internal/daemon/controller/handlers/accounts/grants_test.go
```diff
@@ -5,101 +5,216 @@ package accounts_test
```
elimt, line 5: For the tests, I see we use a lot of wilcards for the type. Can we add at test coverage where the type is `auth-method`

bosorawis: I switched some test cases to use `type=account` in the grants string

internal/daemon/controller/handlers/accounts/grants_test.go
```diff
@@ -109,3 +224,721 @@ func TestListPassword_Grants(t *testing.T) {
...
+						Grants:      []string{fmt.Sprintf("id=%s;type=account;actions=read", orgAM.PublicId)},
+						GrantScopes: []string{globals.GrantScopeThis, globals.GrantScopeDescendants},
+					},
+				})
+			},
+			wantAccountID: orgPWAccount.PublicId,
+			wantErr:       nil,
+		},
+		{
+			name: "grant specific id but not scope global.This cannot read global account",
```
dkanney, line 368: Small nitpick - there should be a space between `global.This` right?

i.e. `grant specific id but not scope global. This cannot read global account`

internal/daemon/controller/handlers/accounts/grants_test.go
```diff
@@ -109,3 +224,721 @@ func TestListPassword_Grants(t *testing.T) {
...
+					{
+						RoleScopeId: globals.GlobalPrefix,
+						Grants:      []string{fmt.Sprintf("id=%s;type=account;actions=read", globalPasswordAM.PublicId)},
+						GrantScopes: []string{globals.GrantScopeDescendants},
+					},
+				})
+			},
+			wantAccountID: "",
+			wantErr:       handlers.ForbiddenError(),
+		},
```
dkanney, line 383: Is it worth having a test case for a Read with _no_ grants? 

We may also want a test case for a Read with incorrect grants (e.g. making a `GetAccountRequest` where the grant string has: `actions=list`)

johanbrandhorst (approved):

internal/daemon/controller/handlers/accounts/grants_test.go
bosorawis: added! nice catch!

internal/daemon/controller/handlers/accounts/grants_test.go
bosorawis: I was going for `global.This` like granting this at global scope 😅. Reading it again, it does seem a bit confusing. Renamed the test.

dkanney (approved):

dkanney:
Did we also want tests for the `/v1/accounts/{id}:change-password` and `/v1/accounts/{id}:set-password` actions? Everything else looks good

PR #5569, opened 2025-02-27
Add grants tests for authmethods

dkanney:
Add grants tests for auth methods against the read, list, create, update, delete, and authenticate actions. Each test case also tests grant output fields

Most testcases utilize an `expectedOutput` struct. I wanted to have one common output type for testcases where the output could only be (1) an error or (2) a set of `output_fields` (i.e. no error)

### DIFF
--- a/internal/daemon/controller/handlers/authmethods/grants_test.go
+++ b/internal/daemon/controller/handlers/authmethods/grants_test.go
@@ -85,17 +85,17 @@ func TestGrants_ReadActions(t *testing.T) {
 	require.NoError(t, err)
 
 	// Create OIDC/LDAP/Password auth methods for each scope
-	oidcGlobal := oidc.TestAuthMethod(t, conn, databaseWrapper, globals.GlobalPrefix, oidc.InactiveState, "client-id", "secret")
-	oidcOrg1 := oidc.TestAuthMethod(t, conn, databaseWrapperOrg1, org1.GetPublicId(), oidc.InactiveState, "client-id", "secret")
-	oidcOrg2 := oidc.TestAuthMethod(t, conn, databaseWrapperOrg2, org2.GetPublicId(), oidc.InactiveState, "client-id", "secret")
+	oidcGlobal := oidc.TestAuthMethod(t, conn, databaseWrapper, globals.GlobalPrefix, oidc.InactiveState, "client-id", "secret", oidc.WithName("oidc-global"), oidc.WithDescription("test oidc desc"))
+	oidcOrg1 := oidc.TestAuthMethod(t, conn, databaseWrapperOrg1, org1.GetPublicId(), oidc.InactiveState, "client-id", "secret", oidc.WithName("oidc-org-1"), oidc.WithDescription("test oidc desc"))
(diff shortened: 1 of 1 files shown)


internal/daemon/controller/handlers/authmethods/grants_test.go
```diff
@@ -192,74 +304,631 @@ func TestGrants_ReadActions(t *testing.T) {
...
+
+		// We need a sys eventer in order to authenticate
+		eventConfig := event.TestEventerConfig(t, "TestGrants_WriteActions", event.TestWithObservationSink(t))
+		testLock := &sync.Mutex{}
+		testLogger := hclog.New(&hclog.LoggerOptions{
+			Mutex: testLock,
+			Name:  "test",
+		})
+
+		require.NoError(t, event.InitSysEventer(testLogger, testLock, "TestGrants_WriteActions", event.WithEventerConfig(&eventConfig.EventerConfig)))
```
elimt, line 790: Can we also add a cleanup after initializing the sysEventer with `event.TestResetSystEventer`. I know it caused some issues in other tests because some tests did not clean it up.
eg. https://github.com/hashicorp/boundary/blob/85c923053e7f036161e9c0239cc931dc7e8bb6df/internal/daemon/controller/rate_limiter_test.go#L580

dkanney: Added: https://github.com/hashicorp/boundary/commit/201a412aac714ac87cee149ebef09bb7fc424a30#diff-16687a08c402841c17886737be686fde30cb22ce18d5cc9079c7041c5869048fR584

internal/daemon/controller/handlers/authmethods/grants_test.go
```diff
@@ -192,74 +339,772 @@ func TestGrants_ReadActions(t *testing.T) {
...
+					oidcGlobal.PublicId: {wantErr: handlers.ForbiddenError()},
+					oidcOrg1.PublicId:   {wantErr: handlers.ForbiddenError()},
+					oidcOrg2.PublicId:   {wantErr: handlers.ForbiddenError()},
+					ldapGlobal.PublicId: {wantErr: handlers.ForbiddenError()},
+					ldapOrg1.PublicId:   {wantErr: handlers.ForbiddenError()},
+					ldapOrg2.PublicId:   {wantErr: handlers.ForbiddenError()},
+				},
+			},
+			{
+				name: "project role can't get auth methods (403)",
```
bosorawis, line 466: nitpick: add test covering specific ids grants and multiple roles 


```
				userFunc: iam.TestUserDirectGrantsFunc(t, conn, kmsCache, globals.GlobalPrefix, password.TestAuthMethodWithAccount, []iam.TestRoleGrantsRequest{
					{
						RoleScopeId: globals.GlobalPrefix,
						Grants:      []string{fmt.Sprintf("ids=;type=*;actions=*;output_fields=id,name,description", oidcOrg2.PublicId)},
						GrantScopes: []string{globals.GrantScopeThis},
					},
					{
						RoleScopeId: globals.GlobalPrefix,
						Grants:   []string{fmt.Sprintf("ids=;type=*;actions=*;output_fields=id,name,description", ldapOrg1.PublicId)},
						GrantScopes: []string{globals.descendants},
					},

				}),

```

bosorawis (approved):
1 small nitpick/suggestion but LGTM!

internal/daemon/controller/handlers/authmethods/grants_test.go
dkanney: Added a couple id-only grant tests: 498afca6b

PR #5573, opened 2025-02-27
test: add grants tests for host_catalogs

elimt:
add grants tests for `host_catalog`. This tests read, list, create, update and delete actions. It also tests output fields from grants


internal/daemon/controller/handlers/host_catalogs/grants_test.go
```diff
@@ -46,6 +54,7 @@ func TestGrants_ReadActions(t *testing.T) {
 	iamRepo := iam.TestRepo(t, conn, wrap)
 	kmsCache := kms.TestKms(t, conn, wrap)
 	sche := scheduler.TestScheduler(t, conn, wrap)
+	atRepo, err := authtoken.NewRepository(ctx, rw, rw, kmsCache)
```
dkanney, line 57:
```suggestion
	atRepo, err := authtoken.NewRepository(ctx, rw, rw, kmsCache)
	require.NoError(t, err)
```

internal/daemon/controller/handlers/host_catalogs/grants_test.go
```diff
@@ -74,65 +83,235 @@ func TestGrants_ReadActions(t *testing.T) {
...
+				input: &pbs.ListHostCatalogsRequest{
+					ScopeId:   proj.GetPublicId(),
+					Recursive: true,
+				},
+				userFunc: iam.TestUserDirectGrantsFunc(t, conn, kmsCache, globals.GlobalPrefix, password.TestAccountFunc(t, conn), []iam.TestRoleGrantsRequest{
+					{
+						RoleScopeID: globals.GlobalPrefix,
+						Grants:      []string{"ids=*;type=host-catalog;actions=no-op,list"},
+						GrantScopes: []string{globals.GrantScopeChildren},
+					},
```
dkanney, line 289: Does this mean we create a role at the global level and grant list & no-op actions to children (in this case, all Orgs)? And the List call fails because we try to list host catalogs on a specific project?

internal/daemon/controller/handlers/host_catalogs/grants_test.go
```diff
@@ -74,65 +83,235 @@ func TestGrants_ReadActions(t *testing.T) {
...
 				wantErr: nil,
 				wantIDs: wantHcs,
+				expectOutfields: map[string][]string{
+					hcs[0].GetPublicId(): {globals.IdField, globals.ScopeIdField, globals.TypeField, globals.CreatedTimeField, globals.UpdatedTimeField},
+					hcs[1].GetPublicId(): {globals.IdField, globals.ScopeIdField, globals.TypeField, globals.CreatedTimeField, globals.UpdatedTimeField},
+					hcs[2].GetPublicId(): {globals.IdField, globals.ScopeIdField, globals.TypeField, globals.CreatedTimeField, globals.UpdatedTimeField},
+					hcs[3].GetPublicId(): {globals.IdField, globals.ScopeIdField, globals.TypeField, globals.CreatedTimeField, globals.UpdatedTimeField},
+					hcs[4].GetPublicId(): {globals.IdField, globals.ScopeIdField, globals.TypeField, globals.CreatedTimeField, globals.UpdatedTimeField},
+				},
 			},
```
dkanney, line 115: What would happen if we pinned a specific project id to a list grant? Like so:

`"ids=<proj.PublicId>;type=host-catalog;actions=list,read"`

internal/daemon/controller/handlers/host_catalogs/grants_test.go
```diff
@@ -145,6 +324,684 @@ func TestGrants_ReadActions(t *testing.T) {
...
+
+	t.Run("create", func(t *testing.T) {
+		ctx := context.Background()
+		conn, _ := db.TestSetup(t, "postgres")
+		wrap := db.TestWrapper(t)
+		rw := db.New(conn)
+		iamRepo := iam.TestRepo(t, conn, wrap)
+		kmsCache := kms.TestKms(t, conn, wrap)
+		sche := scheduler.TestScheduler(t, conn, wrap)
+		atRepo, err := authtoken.NewRepository(ctx, rw, rw, kmsCache)
```
dkanney, line 523:
```suggestion
		atRepo, err := authtoken.NewRepository(ctx, rw, rw, kmsCache)
		require.NoError(t, err)
```

internal/daemon/controller/handlers/host_catalogs/grants_test.go
```diff
@@ -145,6 +324,684 @@ func TestGrants_ReadActions(t *testing.T) {
...
+
+	t.Run("update", func(t *testing.T) {
+		ctx := context.Background()
+		conn, _ := db.TestSetup(t, "postgres")
+		wrap := db.TestWrapper(t)
+		rw := db.New(conn)
+		iamRepo := iam.TestRepo(t, conn, wrap)
+		kmsCache := kms.TestKms(t, conn, wrap)
+		sche := scheduler.TestScheduler(t, conn, wrap)
+		atRepo, err := authtoken.NewRepository(ctx, rw, rw, kmsCache)
```
dkanney, line 702:
```suggestion
		atRepo, err := authtoken.NewRepository(ctx, rw, rw, kmsCache)
		require.NoError(t, err)
```

internal/daemon/controller/handlers/host_catalogs/grants_test.go
elimt: Yes we can create a role like that but if it's not applicable to the project, the list call will just return nothing. I don't think it'll throw an error

internal/daemon/controller/handlers/host_catalogs/grants_test.go
elimt: The ids has to be the id of the a resource not a scope so I don't think we support that

internal/daemon/controller/handlers/host_catalogs/grants_test.go
```diff
@@ -74,65 +84,235 @@ func TestGrants_ReadActions(t *testing.T) {
...
+				expectOutfields: map[string][]string{
+					hcs[0].GetPublicId(): {globals.IdField, globals.TypeField},
+					hcs[1].GetPublicId(): {globals.IdField, globals.TypeField},
+					hcs[2].GetPublicId(): {globals.IdField, globals.TypeField},
+					hcs[3].GetPublicId(): {globals.IdField, globals.TypeField},
+					hcs[4].GetPublicId(): {globals.IdField, globals.TypeField},
+				},
+			},
+			{
+				name: "project role with host-catalog type, list action, grant scope: this returns all created host catalogs",
```
dkanney, line 228: > project role with host-catalog type, list action, grant scope: this returns **all created host catalogs**

It doesn't look like this test case returns all created host catalogs -- I think the description may need to be updated

internal/daemon/controller/handlers/host_catalogs/grants_test.go
```diff
@@ -74,65 +84,235 @@ func TestGrants_ReadActions(t *testing.T) {
...
+				userFunc: iam.TestUserDirectGrantsFunc(t, conn, kmsCache, globals.GlobalPrefix, password.TestAccountFunc(t, conn), []iam.TestRoleGrantsRequest{
+					{
+						RoleScopeID: org.PublicId,
+						Grants:      []string{"ids=*;type=host-catalog;actions=no-op,list"},
+						GrantScopes: []string{globals.GrantScopeThis},
+					},
+				}),
+				wantIDs: nil,
 			},
 		}
```
dkanney, line 309: How about a test case where we assign no grants at all? `wantIDs` will be `nil` and we should receive a forbidden error.

internal/daemon/controller/handlers/host_catalogs/grants_test.go
```diff
@@ -145,6 +325,687 @@ func TestGrants_ReadActions(t *testing.T) {
...
+						RoleScopeID: proj.PublicId,
+						Grants:      []string{"id=*;type=*;actions=*;output_fields=id,name"},
+						GrantScopes: []string{globals.GrantScopeThis},
+					},
+				}),
+				canCreateInScopes: map[*pbs.CreateHostCatalogRequest]error{
+					{Item: &pb.HostCatalog{ScopeId: proj.PublicId, Type: "static", Name: &wrapperspb.StringValue{Value: "hc-direct-test-1"}}}: nil,
+				},
+				expectOutfields: []string{globals.IdField, globals.NameField},
+			},
```
dkanney, line 565: Would there be any value in creating an additional project under `org` to test against? 

For this test, we could use the second project to verify that the role granted to `proj` at `globals.GrantScopeThis` doesn't also get granted to another project in the same `org`/parent scope id

internal/daemon/controller/handlers/host_catalogs/grants_test.go
```diff
@@ -74,65 +84,235 @@ func TestGrants_ReadActions(t *testing.T) {
...
-			input         *pbs.ListHostCatalogsRequest
-			rolesToCreate []authtoken.TestRoleGrantsForToken
-			wantErr       error
-			wantIDs       []string
+			name            string
+			input           *pbs.ListHostCatalogsRequest
+			userFunc        func() (user *iam.User, accountId string)
+			wantErr         error
+			wantIDs         []string
+			expectOutfields map[string][]string
```
dkanney, line 92: I don't think we need `wantIDs` alongside `expectOutfields`. I think these two fields are redundant -- each id in `wantIDs` will have an entry in `expectOutfields`. Therefore, if an id is returned, it'll exist in both fields or neither, so we can just use `expectOutfields`

internal/daemon/controller/handlers/host_catalogs/grants_test.go
```diff
@@ -74,65 +84,235 @@ func TestGrants_ReadActions(t *testing.T) {
 
 	t.Run("List", func(t *testing.T) {
```
bosorawis, line 85: since output_fields tests are included here, could you add more coverages for other fields that are included `attributes`, `worker_filter`, `authorized_collection_actions`, `authorized_actions`, `secrets`, and `secrets_hmac`?

internal/daemon/controller/handlers/host_catalogs/grants_test.go
bosorawis: +1.  this'll also help asserting `output_fields` behavior for all cases (like `output_fields=*`)

elimt: Okay added this

internal/daemon/controller/handlers/host_catalogs/grants_test.go
elimt: fixed

internal/daemon/controller/handlers/host_catalogs/grants_test.go
elimt: Okay added this

internal/daemon/controller/handlers/host_catalogs/grants_test.go
elimt: Added this

dkanney:
- Added a couple additional test cases for attempting actions without any grants (error expected): 23a702ec6a851c5d0de49352c8382bcb0cccf9bf
- Consolidated commonly used code/functions into helper funcs: 6aa28ec41538e0537fea55ceac27a637828cac6a

internal/daemon/controller/handlers/host_catalogs/grants_test.go
```diff
@@ -145,6 +310,721 @@ func TestGrants_ReadActions(t *testing.T) {
...
+			return iam.TestRepo(t, conn, wrap), nil
+		}
+		catalogServiceFn := func() (*host.CatalogRepository, error) {
+			return host.NewCatalogRepository(ctx, rw, rw)
+		}
+		s, err := host_catalogs.NewService(ctx, staticRepoFn, pluginHostRepoFn, pluginRepoFn, iamRepoFn, catalogServiceFn, 1000)
+		require.NoError(t, err)
+
+		org, proj := iam.TestScopes(t, iamRepo)
+		_ = iam.TestProject(t, iamRepo, org.GetPublicId())
```
bosorawis, line 540: nitpick: unused project. You could add an additional this to test `children` and `descendant` grants 

```
				canCreateInScopes: map[*pbs.CreateHostCatalogRequest]error{
					{Item: &pb.HostCatalog{ScopeId: proj.PublicId, Type: "static", Name: &wrapperspb.StringValue{Value: "hc-direct-test-1"}}}: nil,
				},
					{Item: &pb.HostCatalog{ScopeId: proj2.PublicId, Type: "static", Name: &wrapperspb.StringValue{Value: "hc-direct-test-2"}}}: nil,
				},

```

dkanney: Added the second project & some additional test cases to assert `children`/`descendant`/`individual` grant scope behavior: 980384d52ea7c9bff18c9031262cbf38a5926363

bosorawis (approved):

PR #5585, opened 2025-03-06
fix(api): Prevent memory spike after creating many resources

moduli:
Recently, pagination end-to-end tests have been consistently failing in `boundary-enterprise` due to an out of memory panic
```
│ === RUN   TestCliPaginateAuthMethods
│     scope.go:85: Created Org Id: o_1eB3coWMK0
│     authmethod.go:35: Created Auth Method: ampw_XXlYygaK0k
│     authmethod.go:35: Created Auth Method: ampw_bPs7brgJ7V
│     authmethod.go:35: Created Auth Method: ampw_4LrkW9cI1b
│     authmethod.go:35: Created Auth Method: ampw_Dbc7emmAbv
│     authmethod.go:35: Created Auth Method: ampw_mHHk4Pxngw
│     authmethod.go:35: Created Auth Method: ampw_h2NishKsBz
│     authmethod.go:35: Created Auth Method: ampw_oWOcPTluZc
│     authmethod.go:35: Created Auth Method: ampw_3N0Cw5KI3g
│     authmethod.go:35: Created Auth Method: ampw_HD5sqrXOkO
│     authmethod.go:35: Created Auth Method: ampw_WLvKjTFuO8
│     authmethod.go:35: Created Auth Method: ampw_K8H6gNEAHs
│     auth_method_password_test.go:185: 
│         	Error Trace:	/src/boundary/testing/internal/e2e/tests/base/auth_method_password_test.go:185
│         	Error:      	Received unexpected error:
│         	            	exit status 2
│         	Test:       	TestCliPaginateAuthMethods
│         	Messages:   	fatal error: runtime: out of memory
│         	            	
│         	            	runtime stack:
│         	            	...
│         	            	goroutine 1 gp=0xc0000061c0 m=6 mp=0xc000500008 [running]:
│         	            	runtime.systemstack_switch()
│         	            		runtime/asm_amd64.s:479 +0x8 fp=0xc000d5b340 sp=0xc000d5b330 pc=0x4772e8
│         	            	runtime.(*mheap).alloc(0x800000000?, 0x400000?, 0x10?)
│         	            		runtime/mheap.go:956 +0x5b fp=0xc000d5b388 sp=0xc000d5b340 pc=0x429a1b
│         	            	runtime.(*mcache).allocLarge(0x1718?, 0x7fffffff0, 0x0)
│         	            		runtime/mcache.go:234 +0x87 fp=0xc000d5b3d8 sp=0xc000d5b388 pc=0x416447
│         	            	runtime.mallocgc(0x7fffffff0, 0x1eb1a80, 0x1)
│         	            		runtime/malloc.go:1177 +0x5d0 fp=0xc000d5b478 sp=0xc000d5b3d8 pc=0x46bd70
│         	            	runtime.makeslice(0xc00047aee8?, 0x1f6ac20?, 0xc000a9e990?)
│         	            		runtime/slice.go:116 +0x49 fp=0xc000d5b4a0 sp=0xc000d5b478 pc=0x473349
│         	            	github.com/hashicorp/boundary/api/authmethods.(*Client).List(0xc000094088, ***0xb249d18, 0xc000bc80f0***, ***0x7fff344fc821, 0xc***, ***0x0, 0x0, 0x0***)
│         	            		github.com/hashicorp/boundary/api@v0.0.50/authmethods/authmethods.gen.go:379 +0x578 fp=0xc000d5b9c0 sp=0xc000d5b4a0 pc=0x18d91f8
│         	            	github.com/hashicorp/boundary/internal/cmd/commands/authmethodscmd.(*Command).Run(0xc000122f60, ***0xc000154150, 0x3, 0x3***)
│         	            		...
```
Example run: https://github.com/hashicorp/boundary-enterprise/actions/runs/13701871710/job/38318300622

This seemed to start happening after some runner changes were made, presumably the changes to use Ubuntu 22.04 runners. In `boundary-enterprise`, this workflow is set to use `RUNNER_LARGE = m5.8xlarge`, but jobs have been choosing to use `m6a.2xlarge`. My guess is that it's due to this: https://hashicorp.atlassian.net/wiki/spaces/IPS/pages/2368864360/Organization+Runner+Groups#Ubuntu-22.04-x64-runners. There is a large memory difference between the instances.
- m5.8xlarge = 128 GiB
- m6a.2xlarge = 32 GiB

Regardless, it was still surprising that tests would still run out of memory on a 32 GiB instance.

Looking at the stack trace, we noticed that it was occurring during an allocation during the a list call. The theory is that this allocation uses an estimate length coming from postgres, and because the test had just created a number of resources in quick succession, the estimate length large enough to run out of memory. This PR modifies how the allocation is done. 

Here is a run in `boundary-enterprise` with these changes that results in a success: https://github.com/hashicorp/boundary-enterprise/actions/runs/13707265684/job/38336012952

https://hashicorp.atlassian.net/browse/ICU-16613

### DIFF
--- a/internal/api/genapi/templates.go
+++ b/internal/api/genapi/templates.go
@@ -295,7 +295,9 @@ func (c *Client) List(ctx context.Context, {{ .CollectionFunctionArg }} string, 
 		return target, nil
 	}
 
-	allItems := target.Items
+	fmt.Fprintln(os.Stderr, "Estimated item count (postgres):", target.EstItemCount)
+	allItems := make([]*{{ .Name }}, 0, target.EstItemCount)
+	allItems = append(allItems, target.Items...)
 
 	// If there are more results, automatically fetch the rest of the results.
 	// idToIndex keeps a map from the ID of an item to its index in target.Items.
@@ -807,6 +809,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"net/url"
 	"slices"
 	"time"
--- a/internal/api/genapi/templates.go
+++ b/internal/api/genapi/templates.go
@@ -285,7 +285,7 @@ func (c *Client) List(ctx context.Context, {{ .CollectionFunctionArg }} string, 
 
 	// In case we shortcut out due to client directed pagination, ensure these
 	// are set
-{{ if .RecursiveListing }} 
+{{ if .RecursiveListing }}
 	target.recursive = opts.withRecursive
 {{ end }}
 	target.pageSize = opts.withPageSize
@@ -295,8 +295,7 @@ func (c *Client) List(ctx context.Context, {{ .CollectionFunctionArg }} string, 
 		return target, nil
 	}
 
-	allItems := make([]*{{ .Name }}, 0, target.EstItemCount)
-	allItems = append(allItems, target.Items...)
+	allItems := target.Items
 
 	// If there are more results, automatically fetch the rest of the results.
 	// idToIndex keeps a map from the ID of an item to its index in target.Items.
@@ -399,7 +398,7 @@ func (c *Client) ListNextPage(ctx context.Context, currentPage *{{ .Name }}ListR
 	opts, apiOpts := getOpts(opt...)
 	opts.queryMap["{{ snakeCase .CollectionFunctionArg }}"] = currentPage.{{ .CollectionFunctionArg }}
 
-{{ if .RecursiveListing }} 
+{{ if .RecursiveListing }}
 	// Don't require them to re-specify recursive
 	if currentPage.recursive {
 		opts.queryMap["recursive"] = "true"
@@ -446,7 +445,7 @@ func (c *Client) ListNextPage(ctx context.Context, currentPage *{{ .Name }}ListR
 	nextPage.{{ .CollectionFunctionArg }} = currentPage.{{ .CollectionFunctionArg }}
 {{ if .RecursiveListing }}
 	nextPage.recursive = currentPage.recursive
-{{ end }} 
+{{ end }}
 	nextPage.pageSize = currentPage.pageSize
 	// Cache the removed IDs from this page
 	nextPage.allRemovedIds = append(currentPage.allRemovedIds, nextPage.RemovedIds...)
@@ -516,14 +515,14 @@ func (c *Client) Read(ctx context.Context, id string, opt... Option) (*{{ .Name 
 `))
 
 var deleteTemplate = template.Must(template.New("").Parse(`
-func (c *Client) Delete(ctx context.Context, id string, opt... Option) (*{{ .Name }}DeleteResult, error) { 
+func (c *Client) Delete(ctx context.Context, id string, opt... Option) (*{{ .Name }}DeleteResult, error) {
 	if id == "" {
 		return nil, fmt.Errorf("empty id value passed into Delete request")
 	}
 	if c.client == nil {
 		return nil, fmt.Errorf("nil client")
 	}
-	
+
 	opts, apiOpts := getOpts(opt...)
 
 	req, err := c.client.NewRequest(ctx, "DELETE", {{ .ResourcePath }}, nil, apiOpts...)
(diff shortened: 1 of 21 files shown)


moduli:
Not sure if we want to backport this?

internal/api/genapi/templates.go
```diff
@@ -295,8 +295,7 @@ func (c *Client) List(ctx context.Context, {{ .CollectionFunctionArg }} string,
 		return target, nil
 	}
 
-	allItems := make([]*{{ .Name }}, 0, target.EstItemCount)
-	allItems = append(allItems, target.Items...)
+	allItems := target.Items
```
tmessi, line 298: This will potentially lead to lots of re-allocations when this slice needs to grow as each new page is requested and added to it. Given that, I'm not sure this is the fix that we would want for this.

tmessi (changes requested):

internal/api/genapi/templates.go
tmessi: Also, if this allocation size is the issue, we would still end up hitting it as we process each page and increase the allocation size of the slice. We might even allocate more memory depending on how the slice size is increased.

jefferai: Depending on what the test is doing, a better idea may be to not rely on this logic but do the page-by-page fetching downstream in the test -- see the cache code for an example of how this is done. That way you can do verification per page, then re-use the memory.

If the tests are being run in parallel, you may want to remove the parallelization.

johanbrandhorst: I proposed this change. I agree that it will lead to more allocations, but the testing shows that using `target.EstItemCount` can sometimes allocate enough memory to OOM one of our 32 GB test instances (apparently?). `EstItemCount` is based on the postgres query planner estimate for the number of rows in a table, which I theorized could be greatly exaggerated if we've inserted a few items very recently.

Since this function is mostly network bound (with each page doing a request to the controller), I don't think allocating new entries in the slice as each page comes back will be a significant issue. At least it should keep the total amount of memory use within acceptable limits. We could explicitly allocate a page-size chunk of entries with each page, but I don't think that will be different from just appending the items directly.

An alternative we consider was pre-allocating something like

```go
allItems := make([]*{{ .Name }}, 0, min(target.EstItemCount, 1000))
```

But I figured it won't actually need preallocating at all in practice. Do you have another suggestion?

johanbrandhorst: I updated this PR to log the estimated item count as we allocate this slice. Here's a snippet from the recent test, ]full log attached](https://github.com/user-attachments/files/19169109/test-e2e-base_e2e_docker_base.log):

```
=== RUN   TestApiPaginateAuthMethods
    scope.go:33: Created Org Id: o_aDVOvOFb5l
    authmethod.go:35: Created Auth Method: ampw_JI2HEaY7hX
    authmethod.go:35: Created Auth Method: ampw_9nJhGeM54S
    authmethod.go:35: Created Auth Method: ampw_Zv1fmIjKCT
    authmethod.go:35: Created Auth Method: ampw_JZzwuncJih
    authmethod.go:35: Created Auth Method: ampw_fiqvaETFAl
    authmethod.go:35: Created Auth Method: ampw_C04dDP8Gyd
    authmethod.go:35: Created Auth Method: ampw_UBvhsGWfVO
    authmethod.go:35: Created Auth Method: ampw_6pnsyzfLJ7
    authmethod.go:35: Created Auth Method: ampw_nszlepCw1m
    authmethod.go:35: Created Auth Method: ampw_zKnUiNJTxx
    authmethod.go:35: Created Auth Method: ampw_9jFcJ8PQ9x
Estimated item count (postgres): 4294967294
```

Looking at the rest of the pagination tests, the estimated item counts returned are:

* 0
* 0
* 4294967294
* 4294967293
* 0
* 4294967295
* 0
* 0
* 0
* 9
* 3
* 7
* 12
* 3

As can be seen, the estimated item count is sometimes off by an absurd number. I didn't know the number was going to be this big, but looking at this with some data, I'm thinking we should probably detect in the controller when this number is unexpectedly big (how?) and return 0 in those cases (or something else)? Thoughts welcome.

johanbrandhorst:
After some discussion, we believe this is best fixed in Boundary instead of in the API client. I've created a JIRA ticket for this: https://hashicorp.atlassian.net/browse/ICU-16649.

PR #5586, opened 2025-03-07
Normalize Addresses entered across workers, targets, sessions, and ldap

kheina:


internal/cmd/base/servers.go
```diff
@@ -839,6 +839,11 @@ func (b *Server) SetupWorkerPublicAddress(conf *config.Config, flagValue string)
 	if err != nil {
 		return fmt.Errorf("Error splitting public adddress host/port: %w", err)
 	}
+	if host != "" {
+		if host, err = parseutil.NormalizeAddr(host); err != nil {
+			return fmt.Errorf("Error normalizing worker address")
+		}
+	}
```
hugoghx, line 846: Oops seems like we have some [overlapping](https://github.com/hashicorp/boundary/pull/5587/commits/7da73ee596a61196c1398b346d743131afeac196) work since I was looking at the `base` package too

kheina: yeah I think this is what we were talking about the other day regarding worker startup. I think we ended up coming to the same conclusion, though, so we can just handle it as a merge conflict later on

hugoghx:
There's some test failures here that I expected, but some I didn't:

```
--- FAIL: TestTarget_AddressMutualExclusiveRelationship (1.60s)

    target_test.go:384: 
        	Error Trace:	/home/runner/work/boundary/boundary/internal/tests/api/targets/target_test.go:384
        	            	{"kind":"InvalidArgument", "message":"Error in provided request.", "details":{"request_fields":[{"name":"address", "description":"Error parsing address: address cannot be encapsulated by brackets."}]}}
        	Test:       	TestTarget_AddressMutualExclusiveRelationship
```

This is the failing call:
```go
targetResp, err := tClient.Create(tc.Context(), "tcp", proj.GetPublicId(),
	targets.WithName("test-address"), targets.WithAddress("[::1]"), targets.WithTcpTargetDefaultPort(22))
```

`[::1]` is not valid so needs to be changed to `targets.WithAddress("::1")`

---

`--- FAIL: TestCreateTarget_DirectlyAttachedAddress/* (0.00s)`

`--- FAIL: TestRepository_UpdateTcpTarget/* (4.59s)`

`--- FAIL: TestRepository_CreateTarget/* (1.71s)`

These all have problems with some of the IPv6 addrs in the test cases having brackets around them.

internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -1732,7 +1733,14 @@ func toProto(ctx context.Context, in target.Target, opt ...handlers.Option) (*pb
...
 	if outputFields.Has(globals.AddressField) {
-		out.Address = wrapperspb.String(in.GetAddress())
+		addr := in.GetAddress()
+		if addr != "" {
+			var err error
+			if addr, err = parseutil.NormalizeAddr(in.GetAddress()); err != nil {
+				return nil, errors.New(ctx, errors.InvalidParameter, op, fmt.Sprintf("unable to normalize the given target address: %q", in.GetAddress()))
+			}
+		}
+		out.Address = wrapperspb.String(addr)
```
hugoghx, line 1743: This is odd -- Why do we need to normalize on output? Shouldn't it already be normalized from input?

kheina: I don't think I noticed that it was only used for output addresses, just that toProto was called in every function. this may not be necessary as the tests were passing before I fixed the error return issue in 6f728b4891579eb7f438225f3031b9e0d309f962. for clarity, though, create and update were fixed by updating the `util.ParseAddress` function

hugoghx: > create and update were fixed by updating the util.ParseAddress function

Yep, and we call that func in the Target repo so I just assumed that + changing the validation funcs (as you did) + changing any invalid tests should've done the job.

hugoghx (approved):

PR #5588, opened 2025-03-07
test: add grants test for auth-token resource

elimt:
test: add grants test for  auth-token resource

- add test that tests the grants for the auth-token resource with different actions, scopes and output fields
- Fix bug where the `ParentScopeId` is not set when fetching output fields for a resource. Output fields were not being respected for children grant scopes.


internal/daemon/controller/handlers/authtokens/grants_test.go
```diff
@@ -63,68 +63,950 @@ func TestGrants_ReadActions(t *testing.T) {
...
-				var gotIDs []string
-				for _, g := range got.Items {
-					fmt.Println("ID:", g.GetId(), "Scope:", g.GetScopeId())
-					gotIDs = append(gotIDs, g.GetId())
+			}
+		})
+	}
+}
+
+func TestGrants_Read(t *testing.T) {
```
bosorawis, line 296: Could you add a test that covers `output_fields=*` to verify that all the fields are being returned properly in the happy path case?

internal/daemon/controller/handlers/authtokens/grants_test.go
```diff
@@ -63,68 +63,950 @@ func TestGrants_ReadActions(t *testing.T) {
...
-
-		for _, tc := range testcases {
-			t.Run(tc.name, func(t *testing.T) {
-				tok := authtoken.TestAuthTokenWithRoles(t, conn, kmsCache, globals.GlobalPrefix, tc.rolesToCreate)
-				fullGrantAuthCtx := auth.TestAuthContextFromToken(t, conn, wrap, tok, iamRepo)
-				got, finalErr := s.ListAuthTokens(fullGrantAuthCtx, tc.input)
-				if tc.wantErr != nil {
-					require.ErrorIs(t, finalErr, tc.wantErr)
-					return
+	testcases := []struct {
```
bosorawis, line 66: Could you add a few negative tests for missing permissions,  tests with specific scope grants, and tests with specific resource ID grants?

internal/daemon/controller/handlers/authtokens/grants_test.go
```diff
@@ -63,68 +63,950 @@ func TestGrants_ReadActions(t *testing.T) {
...
 				}
-			})
-		}
-	})
+			}
+		})
+	}
+}
+
+func TestGrants_ReadSelf(t *testing.T) {
```
bosorawis, line 555: can you add a case that covers grants with specific resource ID?

internal/daemon/controller/handlers/authtokens/grants_test.go
```diff
@@ -63,68 +63,950 @@ func TestGrants_ReadActions(t *testing.T) {
...
+					require.ErrorIs(t, err, handlers.ForbiddenError())
+					continue
+				}
+				require.NoErrorf(t, err, "failed to delete group in scope %s", scope)
+			}
+		})
+	}
+}
+
+func TestGrants_DeleteSelf(t *testing.T) {
```
bosorawis, line 875: tests for specific resource ID grants

bosorawis (approved):
the changes LGTM! I'd like to see a few more test cases covering grants with specific IDs - specifically for `...-self` calls where the granted resource ID doesn't match the user's token

PR #5591, opened 2025-03-12
test(hostsets): add grants tests

dkanney:
- CRUDL grants tests for host sets
- Added `TestHostSet()` function in order to create test host sets with names & descriptions for validating `output_fields`
- Added `TestHostCatalog()` function in order to create multiple test host catalogs without running into duplicate name errors from the API

### DIFF
--- a/internal/daemon/controller/handlers/host_sets/grants_test.go
+++ b/internal/daemon/controller/handlers/host_sets/grants_test.go
@@ -817,7 +817,166 @@ func TestGrants_WriteActions(t *testing.T) {
 		}
 	})
 
-	t.Run("Add Host Set Hosts", func(t *testing.T) {})
-	t.Run("Remove Host Set Hosts", func(t *testing.T) {})
+	t.Run("Add Host Set Hosts", func(t *testing.T) {
+		hc := static.TestCatalog(t, conn, proj.PublicId, static.WithName("test catalog"), static.WithDescription("test desc"))
+		hcClone := static.TestCatalog(t, conn, proj.PublicId, static.WithName("clone catalog"), static.WithDescription("clone desc"))
+
+		testcases := []struct {
+			name     string
+			userFunc func() (*iam.User, auth.Account)
+			expected expectedOutput
+		}{
+			{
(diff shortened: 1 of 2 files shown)


internal/daemon/controller/handlers/host_sets/grants_test.go
```diff
@@ -41,99 +56,1270 @@ func TestGrants_ReadActions(t *testing.T) {
...
+		return hostplugin.NewRepository(ctx, rw, rw, kmsCache, scheduler, map[string]plgpb.HostPluginServiceClient{})
 	}
 	repoFn := func() (*static.Repository, error) {
 		return static.NewRepository(ctx, rw, rw, kmsCache)
 	}
 	s, err := host_sets.NewService(ctx, repoFn, pluginRepoFn, 1000)
 	require.NoError(t, err)
 
-	org, proj := iam.TestScopes(t, iamRepo)
+	org, proj := iam.TestScopes(t, iamRepo) // TODO: Add a second org/proj scopes? To enforce independent behavior (e.g. grants in org1/proj1 shouldn't be returned when querying org2/proj2)
```
bosorawis, line 75: I don't know if this is necessary since `hostsets` do not support recursive listing. But if we'd like to add additional test, I suggest adding these 2 tests (using the same role setup) and remove the TODO
- create role at `org_1` with `grant_scope = children` 
- - check that you can list host sets in host catalog in `proj_1`
- - check that you cannot list host sets in host catalog in `proj_2`

bosorawis (approved):

internal/daemon/controller/handlers/host_sets/grants_test.go
dkanney: Added here: 13d5c599a

- Created a second project + host-catalog + host-sets
- Grant on `host-catalog 1` at `org_1` scope with `grant_scope = children`
    - Test #1 can list host-sets on the host-catalog on `proj_1`
    - Test #2 cannot list host-sets on the host-catalog on `proj_2` (lacks permission)

PR #5594, opened 2025-03-13
fix(token): utilize a more accurate comparison method

ddebko:
# Summary

Use a constant-time method for comparing secrets. Such an operation will scan the entirety of both strings to be compared rather than returning early as soon as a character is encountered that does not match.


johanbrandhorst (approved):
Hehe, more accurate isn't exactly true :P.

PR #5597, opened 2025-03-13
fix(db): return expected rows err for migration logs

ddebko:
# Summary

Errors encountered during database migrations applied through the database init and database migrate commands do not display the correct error messages. This issue arises from an incorrect variable reference in the GetMigrationLog function.


johanbrandhorst (approved):
You're on a roll today 🥁

PR #5598, opened 2025-03-14
chore: Update AWS and MinIO plugins

hugoghx:


kheina (approved):
LGTM! both plugins are latest

PR #5599, opened 2025-03-14
Normalize fields across various Boundary components

hugoghx:
This PR normalizes various address/ip/host fields accross Boundary to make it compliant with IPv6 specifications.

### DIFF
--- a/internal/daemon/controller/handlers/hosts/host_service.go
+++ b/internal/daemon/controller/handlers/hosts/host_service.go
@@ -772,14 +772,12 @@ func validateCreateRequest(req *pbs.CreateHostRequest) error {
 					len(attrs.GetAddress().GetValue()) > static.MaxHostAddressLength {
 					badFields[globals.AttributesAddressField] = fmt.Sprintf("Address length must be between %d and %d characters.", static.MinHostAddressLength, static.MaxHostAddressLength)
 				} else {
-					host, port, err := util.SplitHostPort(attrs.GetAddress().GetValue())
-					switch {
-					case err == nil && port != "":
-						badFields[globals.AttributesAddressField] = "Address for static hosts does not support a port."
-					case (host != "" && port == "") || strings.Contains(err.Error(), globals.MissingPortErrStr):
-						// Bare hostname, which we want
-					default:
+					_, port, err := util.SplitHostPort(attrs.GetAddress().GetValue())
+					if err != nil && !errors.Is(err, util.ErrMissingPort) {
 						badFields[globals.AttributesAddressField] = fmt.Sprintf("Error parsing address: %v.", err)
+					}
+					if port != "" {
+						badFields[globals.AttributesAddressField] = "Address for static hosts does not support a port."
 					}
 				}
 			}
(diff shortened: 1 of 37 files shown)


internal/daemon/controller/handlers/hosts/host_service.go
```diff
@@ -773,11 +772,11 @@ func validateCreateRequest(req *pbs.CreateHostRequest) error {
...
 					badFields[globals.AttributesAddressField] = fmt.Sprintf("Address length must be between %d and %d characters.", static.MinHostAddressLength, static.MaxHostAddressLength)
 				} else {
-					_, _, err := net.SplitHostPort(attrs.GetAddress().GetValue())
+					host, port, err := util.SplitHostPort(attrs.GetAddress().GetValue())
 					switch {
-					case err == nil:
+					case err == nil && port != "":
 						badFields[globals.AttributesAddressField] = "Address for static hosts does not support a port."
-					case strings.Contains(err.Error(), globals.MissingPortErrStr):
+					case (host != "" && port == "") || strings.Contains(err.Error(), globals.MissingPortErrStr):
```
johanbrandhorst, line 779: Should this not be
```suggestion
					case (host != "" && port == "") || strings.Contains(err.Error(), util.InvalidAddressContainsPort):
```
?
Also, since we own this function, can't we make it return a typed error instead of having to do this string matching? If it means changing how the function is used elsewhere too, I think that's worthwhile. We should use `errors.Is` where we can.

internal/daemon/controller/handlers/targets/target_service.go
```diff
@@ -1839,15 +1839,13 @@ func validateCreateRequest(req *pbs.CreateTargetRequest) error {
...
 		if address := item.GetAddress(); address != nil {
-			if len(address.GetValue()) < static.MinHostAddressLength ||
-				len(address.GetValue()) > static.MaxHostAddressLength {
-				badFields[globals.AddressField] = fmt.Sprintf("Address length must be between %d and %d characters.", static.MinHostAddressLength, static.MaxHostAddressLength)
-			}
-			_, _, err := net.SplitHostPort(address.GetValue())
+			_, err := util.ParseAddress(context.Background(), address.GetValue())
 			switch {
 			case err == nil:
+			case err.Error() == util.InvalidAddressLength:
```
johanbrandhorst, line 1845: Ditto can we make this use errors.Is? Goes for the rest of these comparisons.

johanbrandhorst (changes requested):
Thanks for doing this, I wonder if we should take this opportunity to change the error handling of util.SplitHostPort?

internal/daemon/controller/handlers/hosts/host_service.go
hugoghx: > Should this not be ...

In this case, this switch case is not testing a negative case -- You can see it actually has no code that executes if it falls in this case. The reason this is here at all is due to the `default` case.

---

The logic around this is very convoluted and should be refactored. I have some ideas on improving all of this -- I'll be opening a PR and I'll tag you on it 😃.

johanbrandhorst: Sounds good, I'm happy to approve this pending that separate PR!

johanbrandhorst (commented):
Not actually approving so we don't merge this accidentally, but this LGTM!

johanbrandhorst (approved):
Code LGTM, I'm wondering if we looked at our docs and tutorials to see if we use the `[::1]` format incorrectly anywhere? If not it'd be good to have a look to see that our examples are up to date with this new normalization. Thanks!

hugoghx:
> I'm wondering if we looked at our docs and tutorials to see if we use the [::1] format incorrectly anywhere? If not it'd be good to have a look to see that our examples are up to date with this new normalization. Thanks!

I Ctrl+F'd the codebases for `[::1]` -- Could not find any references. It's still possible that  we got `[something_else_here]` somewhere instead, so I'm going to look a bit more into this.

louisruch (approved):
Great work on this @hugoghx @kheina

PR #5601, opened 2025-03-17
docs/update two pages

mister-ken:
edits on `connect-to-target` and `configure-controllers`

https://boundary-6ww1hl6ns-hashicorp.vercel.app/boundary/docs/hcp/get-started/connect-to-target

https://boundary-6ww1hl6ns-hashicorp.vercel.app/boundary/docs/install-boundary/configure-controllers

### DIFF
--- a/website/content/docs/hcp/get-started/connect-to-target.mdx
+++ b/website/content/docs/hcp/get-started/connect-to-target.mdx
@@ -50,7 +50,7 @@ is why connect helpers can be useful.
 
 The dev-mode default target allows you to make as connections as you want
 within the authorized session. When you finish making connections, a
-`Ctrl-C/Command-C` to the `boundary connect` process to shut down the session.
+`Ctrl-C/Command-C` to the `boundary connect` process will shut down the session.
 
 ## Select targets
 
--- a/website/content/docs/hcp/get-started/connect-to-target.mdx
+++ b/website/content/docs/hcp/get-started/connect-to-target.mdx
@@ -14,7 +14,7 @@ is the fastest way to onboard a target to HCP Boundary.
 ![Setup Wizard](/img/setup-wizard-address-target.png)
 
 The Quick setup wizard creates a default `tcp` target with the port `22` (the default SSH port using TCP).
-The target is created with the address `127.0.0.1`.
+The wizard creates the target with the address `127.0.0.1`.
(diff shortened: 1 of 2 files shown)


website/content/docs/hcp/get-started/connect-to-target.mdx
```diff
@@ -28,44 +28,45 @@ To connect to the initial EC2 Instances target:
...
-For many applications there are still some extra hurdles that can exist, which
+For some applications there are still some extra hurdles that can exist, which
 is why connect helpers can be useful.
 
-The dev-mode default target allows you to make as many connections as you want
-within the authorized session. When you are finished making connections, simply
-`Ctrl-C/Command-C` the `boundary connect` process to shut down the session.
+The dev-mode default target allows you to make as connections as you want
+within the authorized session. When you finish making connections, a
+`Ctrl-C/Command-C` to the `boundary connect` process to shut down the session.
```
johanbrandhorst, line 53:
```suggestion
`Ctrl-C/Command-C` to the `boundary connect` process will shut down the session.
```

website/content/docs/hcp/get-started/connect-to-target.mdx
```diff
@@ -87,11 +88,11 @@ These are written as `boundary connect <subcommand>` and are supported for the f
 
 Refer to the [Connect helpers](/boundary/docs/concepts/connection-workflows/connect-helpers) documentation for more information.
 
-## Built-in vs. exec
+## Built-in verses exec
```
johanbrandhorst, line 91: Is it common practice to write this out like this? I've not seen it before. Isn't it `versus`? In any case I prefer `vs.`, but we might want to reword it altogether.

Dan-Heath: I usually write this as `vs.` also, but I checked the Google style guide and they do actually recommend that it be spelled out (TIL).

I agree it might just be better to reword though; the "vs" construction was from an older version of this content from a tutorial that compared/contrasted built-in and exec, but since this section just focuses on the Exec flag I think we could just rename it as `Exec flag`.

Dan-Heath (approved):
Thank you for these style and standards fixes @mister-ken , they look great! I pushed some updates to fix some other issues with these pages. If you have any questions, please let me know.

johanbrandhorst (approved):

PR #5602, opened 2025-03-17
Controller Liveliness Query Split

RyanDerr:
- Modified the controller's liveliness update queries to the `server_controller` table via the ticking schedule to perform a psql `update` instead of a `insert on conflict`. 
   - When a controller is first started it will attempt the original `upsert` into the `server_controller`.
   - Subsequent updates via the `tickers.go/startStatusTicking` will be done via an `update` as we can safely assume that the controller will already have an entry into the table at this stage.
      - We expect this to perform better during liveliness updates due to postgres not having to detect the conflict before performing an update.
   - This is in assistance of aiding in pg performance for HCP Boundary's [slow queries](https://hashicorp-cloud.datadoghq.com/logs?query=env%3Aboundary-prod%20service%3Apostgresql%20host%3Ahcp-aurora-postgres-boundary-dataplane%20-%22automatic%20analyze%20of%20table%22%20-%22drop%20role%22%20-%22owner%20of%22%20-%22could%20not%20receive%20data%22%20-%22cannot%20be%20dropped%20because%22%20-%22migration%22%20%40duration%3A%3E1&agg_m=count&agg_m_source=base&agg_q=%40msg&agg_q_source=base&agg_t=count&cols=%40duration%2C%40db.statement%2C%40msg&fromUser=true&messageDisplay=inline&refresh_mode=sliding&saved-view-id=2874261&sort_m=count&sort_m_source=base&sort_t=count&start_query=&storage=hot&stream_sort=time%2Casc&top_n=25&top_o=top&viz=toplist&x_missing=true&from_ts=1725481091778&to_ts=1726085891778&live=true)
- Update the terraform labels for [google_compute_instance](https://github.com/hashicorp/boundary/blob/8d182ca576901c8886303b59756228d0a61a5427/enos/modules/gcp_target/main.tf#L128) to resolve error where the label is based off a GH username, but the GCP label cannot contain uppercase characters, causing failing tests.

### DIFF
--- a/enos/modules/gcp_target/main.tf
+++ b/enos/modules/gcp_target/main.tf
@@ -125,7 +125,7 @@ resource "google_compute_instance" "boundary_target" {
     "project" : "enos",
     "project_name" : "qti-enos-boundary",
     "environment" : var.environment,
-    "enos_user" : replace(lower(var.enos_user), "/[\\W]+/", ""),
+    "enos_user" : replace(var.enos_user, "/[\\W]+/", ""),
     "filter_label_1" : random_id.filter_label1.hex
     "filter_label_2" : random_id.filter_label2.hex
   })
--- a/enos/modules/gcp_target/main.tf
+++ b/enos/modules/gcp_target/main.tf
@@ -125,7 +125,7 @@ resource "google_compute_instance" "boundary_target" {
     "project" : "enos",
     "project_name" : "qti-enos-boundary",
     "environment" : var.environment,
-    "enos_user" : replace(var.enos_user, "/[\\W]+/", ""),
+    "enos_user" : replace(lower(var.enos_user), "/[\\W]+/", ""),
     "filter_label_1" : random_id.filter_label1.hex
     "filter_label_2" : random_id.filter_label2.hex
   })
--- a/internal/server/repository_controller.go
+++ b/internal/server/repository_controller.go
@@ -108,7 +108,6 @@ func (r *Repository) UpdateController(ctx context.Context, controller *store.Con
 			return nil
 		},
 	)
-
 	if err != nil {
 		return db.NoRowsAffected, err
 	}
--- a/internal/server/repository_controller.go
+++ b/internal/server/repository_controller.go
@@ -77,3 +77,41 @@ func (r *Repository) UpsertController(ctx context.Context, controller *store.Con
 
 	return int(rowsUpdated), nil
 }
+
+func (r *Repository) UpdateController(ctx context.Context, controller *store.Controller) (int, error) {
+	const op = "server.(Repository).UpdateController"
+
+	if controller == nil {
+		return db.NoRowsAffected, errors.New(ctx, errors.InvalidParameter, op, "controller is nil")
+	}
+	if controller.PrivateId == "" {
+		return db.NoRowsAffected, errors.New(ctx, errors.InvalidParameter, op, "controller private_id is empty")
+	}
+	if controller.Address == "" {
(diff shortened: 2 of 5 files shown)


enos/modules/gcp_target/main.tf
```diff
@@ -125,7 +125,7 @@ resource "google_compute_instance" "boundary_target" {
     "project" : "enos",
     "project_name" : "qti-enos-boundary",
     "environment" : var.environment,
-    "enos_user" : replace(var.enos_user, "/[\\W]+/", ""),
+    "enos_user" : replace(lower(var.enos_user), "/[\\W]+/", ""),
```
johanbrandhorst, line 128: Could this be a separate PR? I'm a little confused about what this is doing here.

internal/server/repository_controller.go
```diff
@@ -77,3 +77,40 @@ func (r *Repository) UpsertController(ctx context.Context, controller *store.Con
...
+	}
+
+	var rowsUpdated int
+	_, err := r.writer.DoTx(
+		ctx,
+		db.StdRetryCnt,
+		db.ExpBackoff{},
+		func(reader db.Reader, w db.Writer) error {
+			var err error
+			rowsUpdated, err = w.Update(ctx, controller, []string{"description", "address"}, nil, db.WithWhere("private_id = ?", controller.PrivateId))
```
johanbrandhorst, line 101: This will update the `description` and `address` values, but shouldn't it also set `update_time`? Is there a trigger that sets the update_time?

internal/server/repository_controller_test.go
```diff
@@ -100,3 +100,140 @@ func TestRepository_UpsertController(t *testing.T) {
...
+
+			// Retrieve the updated controller in the database and assert updated successfully
+			controllerList, err := testRepo.ListControllers(ctx, []Option{}...)
+			require.NoError(err)
+			updatedControllerEntry := controllerList[len(controllerList)-1]
+
+			assert.Equal(tt.updatedController.PrivateId, updatedControllerEntry.PrivateId)
+			assert.Equal(tt.updatedController.Address, updatedControllerEntry.Address)
+			assert.Equal(tt.updatedController.Description, updatedControllerEntry.Description)
+		})
```
johanbrandhorst, line 237: Can we also check that the UpdateTime has been updated?

johanbrandhorst (changes requested):

internal/server/repository_controller.go
RyanDerr: Yep there exists a trigger the table when an update occurs. 
- [Trigger](https://github.com/hashicorp/boundary/blob/8d182ca576901c8886303b59756228d0a61a5427/internal/db/schema/migrations/oss/postgres/34/02_worker_controller_tables.up.sql#L28)
- [Trigger Procedure](https://github.com/hashicorp/boundary/blob/8d182ca576901c8886303b59756228d0a61a5427/internal/db/schema/migrations/oss/postgres/0/01_domain_types.up.sql#L48)

internal/server/repository_controller_test.go
RyanDerr: Yeah, we can add an assertion to validate that the time is present/not nil. Was doing some funky tests earlier and forgot to add it back.

enos/modules/gcp_target/main.tf
johanbrandhorst: Haha that's hilarious

RyanDerr: Yeah lol, I went ahead and created a PR to solve that solely [here](https://github.com/hashicorp/boundary/pull/5605).

johanbrandhorst (approved):
Thanks!

PR #5609, opened 2025-03-18
Migrate Controller Repo Usage From Store To Server

RyanDerr:
# Overview
- Normally we don't use `store` structs directly for exported methods as part of a repo. This updates the server controller to use it's domain package instead which itself embeds the store. 

This PR was created as a result of the discussion [here](https://github.com/hashicorp/boundary/pull/5602#discussion_r1999418660).

### DIFF
--- a/internal/server/controller.go
+++ b/internal/server/controller.go
@@ -5,8 +5,10 @@ package server
 
 import "github.com/hashicorp/boundary/internal/server/store"
 
-// Controllers are a server that provider user auth to a client and contain the permissions and resources.
-// Additionally controllers communicate with external services such as databases, KMS, idp, plugins, etc.
+// Controller is a server that is responsible for understanding configuration,
+// authenticating and authorizing users, and serving user API requests (e.g. to
+// initiate a session). They also assign tasks to workers (session handling,
+// session recording parsing, etc.).
 type Controller struct {
 	*store.Controller
 }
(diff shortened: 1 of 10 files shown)


johanbrandhorst (approved):
Nice, this looks much better!

internal/server/controller.go
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package server
+
+import "github.com/hashicorp/boundary/internal/server/store"
+
+// Controllers are a server that provider user auth to a client and contain the permissions and resources.
+// Additionally controllers communicate with external services such as databases, KMS, idp, plugins, etc.
```
tmessi, line 9: I think some of these details in this doc are quite correct. For example a lot fo that communication to external services can actually happen via workers,
but the controller tells the worker to make the request. Perhaps instead something like:

```suggestion
// Controller is a server that is responsible for understanding configuration,
// authenticating and authorizing users, and serving user API requests (e.g. to
// initiate a session). They also assign tasks to workers (session handling,
// session recording parsing, etc.).
```

tmessi (changes requested):
Thanks for following up with this refactor!

internal/server/controller.go
RyanDerr: Good callout, went ahead and updated the comment.

tmessi (approved):

PR #5612, opened 2025-03-18
feat: grantsForUser for Global Resources

elimt:
add query to fetch grants for a user for resources that are only globally scoped

**Note**:
DO NOT MERGE until we can create roles by subType from domain code instead of using `setupDB_IamRoles` to manually insert. `setupDB_IamRoles` was used to just validate the query

### DIFF
--- a/internal/iam/query.go
+++ b/internal/iam/query.go
@@ -211,7 +211,7 @@ const (
     users (id) as (
       select public_id
         from iam_user
-      %s -- anonUser || authUser
+      where public_id = any(@user_ids)
     ),
     user_groups (id) as (
       select group_id
@@ -268,7 +268,7 @@ const (
       join iam_grant
         on iam_grant.canonical_grant        = iam_role_grant.canonical_grant
       where iam_role.public_id in (select role_id from user_group_roles)
-        and iam_grant.resource              = any('{"%s", "unknown", "*"}')
+        and iam_grant.resource              = any(@resources)
    ),
    global_roles as (
       select iam_role_global.public_id             as role_id,
--- a/internal/iam/query.go
+++ b/internal/iam/query.go
@@ -259,6 +259,13 @@ const (
       select role_id
         from managed_group_roles
     ),
+    individual_grant_scopes (role_id, scope_id) as (
+      select role_id, scope_id
+        from iam_role_global_individual_org_grant_scope
+      union
+      select role_id, scope_id
+        from iam_role_global_individual_project_grant_scope
+    ),
     roles_with_grants (role_id, canonical_grant) as (
       select iam_role_grant.role_id,
         iam_role_grant.canonical_grant
@@ -282,7 +289,7 @@ const (
       from iam_role_global
       join roles_with_grants
         on roles_with_grants.role_id = iam_role_global.public_id
-      left join iam_role_global_individual_grant_scope individual
+      left join individual_grant_scopes individual
         on roles_with_grants.role_id = individual.role_id
       join iam_scope
         on iam_scope.public_id = iam_role_global.scope_id
--- a/internal/iam/query.go
+++ b/internal/iam/query.go
@@ -298,7 +298,7 @@ const (
            role_parent_scope_id,
            grant_scope,
            grant_this_role_scope,
-           null as individual_grant_scopes, -- there can be no individual grants for global roles
+           null as individual_grant_scopes, -- individual grant scopes do not apply to resources in the global scope
            array_agg(distinct(canonical_grant)) as canonical_grants
       from global_roles
      where global_roles.grant_this_role_scope = true
--- a/internal/iam/query.go
+++ b/internal/iam/query.go
@@ -286,8 +286,8 @@ const (
    )
    select role_id,
           role_scope_id,
-          grant_scope,
 		      role_parent_scope_id,
+          grant_scope,
           grant_this_role_scope,
           NULL as individual_grant_scopes, -- there can be no individual grants for global roles
           array_agg(distinct(canonical_grant)) filter (where canonical_grant is not null) as canonical_grants -- only include canonical grants if they are not null
--- a/internal/iam/query.go
+++ b/internal/iam/query.go
@@ -206,6 +206,104 @@ const (
         on grants.role_id = roles.role_id;
     `
 
+	grantsForUserGlobalResourcesQuery = `
+	with
+    users (id) as (
+      select public_id
+        from iam_user
+      %s -- anonUser || authUser
+    ),
+    user_groups (id) as (
+      select group_id
+        from iam_group_member_user
+       where member_id in (select id from users)
+    ),
+    user_accounts (id) as (
+      select public_id
+        from auth_account
+       where iam_user_id in (select id from users)
+    ),
+    user_oidc_managed_groups (id) as (
+      select managed_group_id
+        from auth_oidc_managed_group_member_account
+       where member_id in (select id from user_accounts)
+    ),
+    user_ldap_managed_groups (id) as (
+      select managed_group_id
+        from auth_ldap_managed_group_member_account
+       where member_id in (select id from user_accounts)
+    ),
+    managed_group_roles (role_id) as (
+      select distinct role_id
+        from iam_managed_group_role
+       where principal_id in (select id from user_oidc_managed_groups)
+          or principal_id in (select id from user_ldap_managed_groups)
+    ),
+    group_roles (role_id) as (
+      select role_id
+        from iam_group_role
+       where principal_id in (select id from user_groups)
+    ),
+    user_roles (role_id) as (
+      select role_id
+        from iam_user_role
+       where principal_id in (select id from users)
+    ),
+    user_group_roles (role_id) as (
+      select role_id
+        from group_roles
+      union
+      select role_id
+        from user_roles
+      union
+      select role_id
+        from managed_group_roles
+    ),
+    roles_with_grants (role_id, canonical_grant) as (
+      select iam_role_grant.role_id,
+        iam_role_grant.canonical_grant
+      from iam_role_grant
+      join iam_role
+        on iam_role.public_id               = iam_role_grant.role_id
+      join iam_grant
+        on iam_grant.canonical_grant        = iam_role_grant.canonical_grant
+      where iam_role.public_id in (select role_id from user_group_roles)
+        and iam_grant.resource              = any('{"%s", "unknown", "*"}')
+   ),
+   global_roles as (
+      select iam_role_global.public_id             as role_id,
+             iam_role_global.scope_id              as role_scope_id,
+             iam_scope.type                        as role_type,
+             'global'                              as role_parent_scope_id, -- manually set to global because we are only looking at global roles and the parent scope is always global
+             iam_role_global.grant_scope           as grant_scope,
+             iam_role_global.grant_this_role_scope as grant_this_role_scope,
+             individual.scope_id                   as individual_grant_scope,
+             roles_with_grants.canonical_grant     as canonical_grant
+      from iam_role_global
+      join roles_with_grants
+        on roles_with_grants.role_id = iam_role_global.public_id
+      left join iam_role_global_individual_grant_scope individual
+        on roles_with_grants.role_id = individual.role_id
+      join iam_scope
+        on iam_scope.public_id = iam_role_global.scope_id
+   )
+   select role_id,
+          role_scope_id,
+          grant_scope,
+		      role_parent_scope_id,
+          grant_this_role_scope,
+          array_agg(distinct(individual_grant_scope)) filter (where individual_grant_scope is not null) as individual_grant_scopes, -- only include individual grant scopes if they are not null
+          array_agg(distinct(canonical_grant)) filter (where canonical_grant is not null)               as canonical_grants -- only include canonical grants if they are not null
+    from global_roles
+   group by (
+          role_id,
+          role_scope_id,
+		      role_parent_scope_id,
+          grant_scope,
+          grant_this_role_scope
+   );
+   `
+
 	estimateCountRoles = `
 		select reltuples::bigint as estimate from pg_class where oid in ('iam_role'::regclass)
 	`
--- a/internal/iam/query.go
+++ b/internal/iam/query.go
@@ -207,70 +207,79 @@ const (
     `
 
 	grantsForUserGlobalResourcesQuery = `
-	with
+    with
     users (id) as (
       select public_id
         from iam_user
-      where public_id = any(@user_ids)
+       where public_id = any(@user_ids)
     ),
     user_groups (id) as (
       select group_id
         from iam_group_member_user
-       where member_id in (select id from users)
+       where member_id in (select id
+                             from users)
     ),
     user_accounts (id) as (
       select public_id
         from auth_account
-       where iam_user_id in (select id from users)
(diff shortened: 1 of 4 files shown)


internal/iam/query.go
```diff
@@ -206,6 +206,111 @@ const (
...
+      select role_id
+        from managed_group_roles
+    ),
+    individual_grant_scopes (role_id, scope_id) as (
+      select role_id, scope_id
+        from iam_role_global_individual_org_grant_scope
+      union
+      select role_id, scope_id
+        from iam_role_global_individual_project_grant_scope
+    ),
```
elimt, line 268: If this query is only for resources that are globally scoped like `Alias`, do we need to query for individual scopes of children and descendants?

I thought we'd have another query that will do that for resources that could be `global`, `org` or `project` like `Roles` resource

dkanney: Hmm... we do not

My thinking was to reconcile the latest changes to our schema with the query on this branch. But yes, the `individual` sections of this query can be removed -- there's no way to grant "individual" actions at the global scope, only on org and project scopes.

internal/iam/repository_role_grant.go
```diff
@@ -494,3 +519,101 @@ func (m *MultiGrantTuple) TestStableSort() {
...
 	m.Grants = strings.Join(gts, "^")
 }
+
+// grantsForUserGlobalResources returns the grants for the user for resources that can
+// only be globally scoped.
+func (r *Repository) grantsForUserGlobalResources(
+	ctx context.Context,
+	userId string,
+	res resource.Type,
+	reqScope Scope,
```
dkanney, line 529: @elimt Do we need `reqScope` in this function? From what I understand, we use `reqScope` to determine which one of these subfunctions to call (e.q. if `reqScope` is global, we'd call `grantsForUserGlobalResources`), but beyond that purpose, I don't know how else it's used here.

___

[Originally it was being used as the `GrantScopeId`](a8620d9a4c99552d058c4fe4079420252d2d1af7) but... I think in this context, `GrantScopeId` would always be `global`. And for the other subfunctions,  I think `GrantScopeId` would be `grant.grantScope`

elimt: Yeah, I think it can be dropped from the function. I don't think it's used.

internal/iam/query.go
dkanney: Removed this subquery for individual grant scopes

elimt: I think you still need individual for global. For example: `ids=al_1234;type=alias;actions=*`, where you are restricting the grant string to one particular alias only. You can cross-check this with Tim & Bo

dkanney: For granting to a specific resource, e.g. `ids=al_1234;type=alias;actions=*`, wouldn't `grant_this_role_scope` need to be `true` in order for the grant to work? Otherwise we won't select anything at the current scope

dkanney: No problem! I appreciate all the help 🙂

I had to double-check to make sure -- we _do_ need `this` grant scope when granting to a specific alias id, otherwise we get `handlers.ForbiddenError()`

internal/iam/query.go
```diff
@@ -206,6 +206,102 @@ const (
         on grants.role_id = roles.role_id;
     `
 
+	grantsForUserGlobalResourcesQuery = `
+	with
+    users (id) as (
+      select public_id
+        from iam_user
+      %s -- anonUser || authUser
```
tmessi, line 214: I believe we can eliminate the `%s` here by making this something like:

```sql
   where public_id = any(@user_ids)
```

And then when we use this query, constructing the args like:

```go
var args []any
switch userId {
case globals.AnonymousUserId:
  args = append(args, sql.Named("user_ids", "{'u_anon', 'u_auth'}"))
default:
  args = append(args, sql.Named("user_ids", fmt.Sprintf("{'%s', 'u_anon', 'u_auth'}", userId)))
}
```

internal/iam/repository_role_grant.go
```diff
@@ -494,3 +519,100 @@ func (m *MultiGrantTuple) TestStableSort() {
...
+		authUser = `where public_id in ('u_anon', 'u_auth', ?)`
+	)
+
+	var query string
+	switch userId {
+	case globals.AnonymousUserId:
+		query = fmt.Sprintf(grantsForUserGlobalResourcesQuery, anonUser, res.String())
+	default:
+		query = fmt.Sprintf(grantsForUserGlobalResourcesQuery, authUser, res.String())
+	}
```
tmessi, line 547: See my comment above on this query, I think this can be updated to eliminate the `%s` and the use of `fmt.Sprintf` for constructing this query.

internal/iam/query.go
```diff
@@ -206,6 +206,102 @@ const (
...
+    roles_with_grants (role_id, canonical_grant) as (
+      select iam_role_grant.role_id,
+        iam_role_grant.canonical_grant
+      from iam_role_grant
+      join iam_role
+        on iam_role.public_id               = iam_role_grant.role_id
+      join iam_grant
+        on iam_grant.canonical_grant        = iam_role_grant.canonical_grant
+      where iam_role.public_id in (select role_id from user_group_roles)
+        and iam_grant.resource              = any('{"%s", "unknown", "*"}')
```
tmessi, line 271: Likewise this could be done with a parameter, ie:

```sql
and iam_grant.resource = any(@resources)
```

And a named arg:

```go
sql.Named("resources", fmt.Sprintf("{'%s', 'unknown', '*'}", res.String()))

**Review by tmessi — CHANGES_REQUESTED** [review]

File: internal/iam/query.go (hunk 3 above)
Comment (dkanney) [reply internal/iam/query.go:214]: Updated to use `sql.Named()` for the userId & resource query args: 5e8d9fb8bdb83b940ea6297fbb1f4c3d27b86596

File: internal/iam/repository_role_grant.go [hunk 6]
```diff
@@ -494,3 +520,67 @@ func (m *MultiGrantTuple) TestStableSort() {
...
+		return nil, errors.Wrap(ctx, err, op)
+	}
+	defer rows.Close()
+	for rows.Next() {
+		var g grantsForUserResults
+		if err := rows.Scan(
+			&g.roleId,
+			&g.roleScopeId,
+			&g.grantScope,
+			&g.roleParentScopeId,
```
dkanney, line 558: Found a bug here. These two fields should be switched

internal/iam/repository_role_grant.go
```diff
@@ -536,3 +562,67 @@ func (m *MultiGrantTuple) TestStableSort() {
...
+	if userId == "" {
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing user id")
+	}
+
+	var args []any
+	switch userId {
+	case globals.AnonymousUserId:
+		args = append(args, sql.Named("user_ids", fmt.Sprintf("{ %s }", userId)))
+	default:
+		args = append(args, sql.Named("user_ids", fmt.Sprintf("{ u_anon, u_auth, %s }", userId)))
```
tmessi, line 584: We should probably use the consts for `u_auth` and `u_anon`, ie:

```suggestion
		args = append(args, sql.Named("user_ids", fmt.Sprintf("{ %s, %s, %s }", globals.AnonymousUserId, globals.AnyAuthenticatedUserId, userId)))
```

tmessi (changes requested):

internal/proto/controller/storage/iam/store/v1/role_global.proto
```diff
@@ -36,7 +36,7 @@ message GlobalRole {
   }];
 
   // control if this role is granted access to its role scope
-  // @inject_tag: `gorm:"default:true"`
+  // @inject_tag: `gorm:"default:false"`
```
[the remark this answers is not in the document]
dkanney: I figured I'd include it in this PR since it was causing issues with the tests --

I wasn't able to create roles where `grant_this_role_scope = false` since this tag would always default the field to `true`

internal/iam/query.go
```diff
@@ -206,6 +206,102 @@ const (
...
+        join iam_scope
+          on iam_scope.public_id = iam_role_global.scope_id
+    )
+    select role_id,
+           role_scope_id,
+		       role_parent_scope_id,
+           grant_scope,
+           grant_this_role_scope,
+           NULL as individual_grant_scopes, -- there can be no individual grants for global roles
+           array_agg(distinct(canonical_grant)) filter (where canonical_grant is not null) as canonical_grants -- only include canonical grants if they are not null
```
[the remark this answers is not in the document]
dkanney: I don't think it can. Since `iam_grant.canonical_grant` is a PK, it can't be null.

Removed this filter

internal/iam/query.go
```diff
@@ -206,6 +206,102 @@ const (
...
+          on roles_with_grants.role_id = iam_role_global.public_id
+        join iam_scope
+          on iam_scope.public_id = iam_role_global.scope_id
+    )
+    select role_id,
+           role_scope_id,
+		       role_parent_scope_id,
+           grant_scope,
+           grant_this_role_scope,
+           NULL as individual_grant_scopes, -- there can be no individual grants for global roles
```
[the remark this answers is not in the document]
dkanney: No more SHOUTCASE 👍

internal/iam/query.go
```diff
@@ -206,6 +206,102 @@ const (
...
+    ),
+    user_group_roles (role_id) as (
+      select role_id
+        from group_roles
+      union
+      select role_id
+        from user_roles
+      union
+      select role_id
+        from managed_group_roles
```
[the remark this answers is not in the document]
dkanney: Updated this spacing

internal/iam/query.go
```diff
@@ -206,6 +206,102 @@ const (
         on grants.role_id = roles.role_id;
     `
 
+	grantsForUserGlobalResourcesQuery = `
+	  with
+    users (id) as (
+      select public_id
+        from iam_user
+      where public_id = any(@user_ids)
```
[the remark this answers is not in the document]
dkanney: Updated to follow river formatting

internal/iam/query.go
```diff
@@ -206,6 +206,102 @@ const (
         on grants.role_id = roles.role_id;
     `
 
+	grantsForUserGlobalResourcesQuery = `
+	  with
```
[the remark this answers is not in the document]
dkanney: Using only spaces now

internal/iam/repository_role_grant.go
```diff
@@ -462,6 +465,29 @@ type MultiGrantTuple struct {
...
+	// roleParentScopeId is the parent scope ID of the role.
+	roleParentScopeId string
+	// grantScope is the grant scope of the role.
+	// The valid values are: "individual", "children" and "descendants".
+	grantScope string
+	// grantThisRoleScope is a boolean that indicates if the role has a grant
+	// for itself aka "this" or "individual" scope.
+	grantThisRoleScope bool
+	// individualGrantScopes represents the individual grant scopes for the role.
+	// This is a slice of strings that may be NULL values if the role does
```
tmessi, line 482:
```suggestion
	// This is a slice of strings that may be empty if the role does
```

Go does not use "null", rather it has `nil`, but in this case I think the slice will be empty. Most of the time a nil slice or an empty slice can be treated as equivalent.

internal/iam/repository_role_grant.go
```diff
@@ -462,6 +465,29 @@ type MultiGrantTuple struct {
...
+	// not have individual grants.
+	individualGrantScopes []string
+	// canonicalGrants represents the canonical grants for the role.
+	// This is a slice of strings that may be NULL if the role does
+	// not have canonical grants associated with it.
```
tmessi, line 487:
```suggestion
	// This is a slice of strings that may be empty if the role does
	// not have canonical grants associated with it.
```

internal/iam/repository_role_grant.go
```diff
@@ -536,3 +562,67 @@ func (m *MultiGrantTuple) TestStableSort() {
...
+	}
+
+	var args []any
+	switch userId {
+	case globals.AnonymousUserId:
+		args = append(args, sql.Named("user_ids", fmt.Sprintf("{ %s }", userId)))
+	default:
+		args = append(args, sql.Named("user_ids", fmt.Sprintf("{ %s, %s, %s }", globals.AnonymousUserId, globals.AnyAuthenticatedUserId, userId)))
+	}
+	args = append(args, sql.Named("resources", fmt.Sprintf("{ %s, unknown, * }", res.String())))
```
tmessi, line 586: Looking at these again, I think we could use `pq.Array` instead of directly doing a `fmt.Sprintf`.

internal/iam/repository_role_grant_test.go
```diff
@@ -2429,3 +2429,77 @@ func SetupChildGrantScopes(t *testing.T, conn *db.DB, repo *Repository) (childGr
...
+	TestRoleGrant(t, conn, role4.PublicId, "ids=*;type=group;actions=read;output_fields=id")
+
+	// Add user to created roles
+	for _, roleId := range []string{role1.PublicId, role2.PublicId, role3.PublicId, role4.PublicId} {
+		_, err := repo.AddPrincipalRoles(ctx, roleId, 1, []string{user.PublicId})
+		require.NoError(t, err)
+	}
+
+	t.Run("Alias", func(t *testing.T) {
+		got, err := repo.grantsForUserGlobalResources(ctx, user.PublicId, resource.Alias)
```
tmessi, line 2465: Can we expand on these tests? It would be good to cover other cases including:
- error cases
- multiple grant tuples being returned
- grants for the special `u_auth` and `u_anon` users

internal/proto/controller/storage/iam/store/v1/role_global.proto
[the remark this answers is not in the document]
tmessi: Are we relying on this to provide the default behavior of setting the grant scope of `this` when creating a role? I believe that is the desired behavior, but if we can enforce that in the repository method instead of a gorm tag that would be fine.

internal/iam/repository_role_grant.go
dkanney: Updated: 39dd7710e

dkanney:
ℹ️ Changed the base branch to `bosorawis-domain-iam-implement-role-grant-scopes-all` to get Bo's grant_scope changes

internal/iam/repository_role_grant_test.go
dkanney: Added more test coverage: 60181c3c287a3f3947c54ee174b9fba4b03237f1

internal/iam/query.go
```diff
@@ -206,6 +206,109 @@ const (
...
+          on roles_with_grants.role_id = iam_role_global.public_id
+        join iam_scope
+          on iam_scope.public_id = iam_role_global.scope_id
+    )
+    select role_id,
+           role_scope_id,
+           role_parent_scope_id,
+           grant_scope,
+           grant_this_role_scope,
+           null as individual_grant_scopes, -- there can be no individual grants for global roles
```
tmessi, line 301: I think the comment here isn't quite correct. It is not that global role can't have individual grants, but rather that any individual grant scopes will not be applicable for a resource that is in the global scope.

tmessi (changes requested):

internal/iam/query.go
dkanney: Updated to: `individual grant scopes do not apply to resources in the global scope`

PR #5626, opened 2025-03-24
Bosorawis domain iam role subtypes

bosorawis:
Implement role subtypes 
- globalRole
- orgRole
- projectRole

These structs will be used to access the DB through gorm. 

This PR does not remove `store.Role` from `iam.Role` struct yet and still rely on it to query `iam_role` table to validate data integrity in tests. This will change in the follow up PR


internal/iam/role.go
```diff
@@ -129,3 +144,186 @@ type deletedRole struct {
 func (s *deletedRole) TableName() string {
 	return "iam_role_deleted"
 }
+
+type globalRole struct {
```
tmessi, line 148: It might be good to add some docs to these un-exported structs. We don't always add docs for un-exported types, but in this case I think it could help clarify the purpose of these structs and how they relate to the database layer.

tmessi (changes requested):

internal/iam/role.go
bosorawis: added go-docs comment

internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
```diff
@@ -65,6 +65,31 @@ begin;
...
+  as $$
+  begin
+    delete from iam_role
+    where public_id = old.public_id;
+    return null; -- result is ignored since this is an after trigger
+  end;
+    $$ language plpgsql;
+    comment on function delete_associated_iam_role_entry() is
+      'delete_associated_iam_role_entry is used to automatically delete associated iam_role entry'
+      'since domain implementation performs deletion on the child table which does not cleanup the base iam_role table ';
```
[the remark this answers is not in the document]
bosorawis: renamed the functions and all associated triggers

PR #5628, opened 2025-03-26
ICU-16457 Split Liveness Query In Two

RyanDerr:
# Overview
Rework and fix the original logic of #5602 query split. Initial implementation resulted in failures of controller's updating their status within the `server_controller` as `w.Update` does not automatically cause the [update_time trigger](https://github.com/hashicorp/boundary/blob/8d182ca576901c8886303b59756228d0a61a5427/internal/db/schema/migrations/oss/postgres/34/02_worker_controller_tables.up.sql#L28) to be used if the values of `address` or `description` were not updated. To result this we instead perform an `w.Exec` where the statement is defined as [updateController](https://github.com/hashicorp/boundary/blob/fce36ad57cf06452cd1d3ea56bbcbd1f9f65dcaa/internal/server/query.go#L227) query which will always force `update_time` to be the value of `now()`. 

# Additional Notes
- Update the `repository_controller` tests to verify that even when the controller values don't change that `update_time` is still updated. 
- Added `cleanupFunc` to test cases to ensure that after each test the inserted controller is removed to prevent future odd behavior where the `ListControllers` can return `n` number of controllers depending on test cases. 
- Verified via `boundary dev` that `worker.(Worker).upstreamDialerFunc` errors were not present when running.

### DIFF
--- a/internal/server/query.go
+++ b/internal/server/query.go
@@ -225,10 +225,10 @@ const (
 `
 
 	updateController = `
-		update server_controller
-		set address = @controller_address,
-			description = @controller_description,
-			update_time = now()
-		where private_id = @controller_private_id;
+		update server_controller  
+           set address     = @controller_address,  
+               description = @controller_description,  
+               update_time = now()  
+         where private_id = @controller_private_id;  
 `
 )
--- a/internal/server/query.go
+++ b/internal/server/query.go
@@ -223,4 +223,12 @@ const (
          where last_status_time > now() - interval '%d seconds'
            and operational_state = 'active';
 `
+
+	updateController = `
+		update server_controller
+		set address = @controller_address,
+			description = @controller_description,
+			update_time = now()
+		where private_id = @controller_private_id;
+`
 )
--- a/internal/server/repository_controller.go
+++ b/internal/server/repository_controller.go
@@ -71,8 +71,12 @@ func (r *Repository) UpsertController(ctx context.Context, controller *Controlle
 	return int(rowsUpdated), nil
 }
 
-func (r *Repository) UpdateController(ctx context.Context, controller *Controller) (int, error) {
-	const op = "server.(Repository).UpdateController"
+// UpdateControllerStatus updates the controller's status in the repository.
+// This includes updating the address or description of the controller as well
+// as updating the update_time attribute, which is required for liveness checks
+// as part of a controller's status ticking.
+func (r *Repository) UpdateControllerStatus(ctx context.Context, controller *Controller) (int, error) {
+	const op = "server.(Repository).UpdateControllerStatus"
 
 	if controller == nil {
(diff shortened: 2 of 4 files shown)


internal/server/query.go
```diff
@@ -223,4 +223,12 @@ const (
          where last_status_time > now() - interval '%d seconds'
            and operational_state = 'active';
 `
+
+	updateController = `
+		update server_controller
+		set address = @controller_address,
+			description = @controller_description,
+			update_time = now()
+		where private_id = @controller_private_id;
```
tmessi, line 232:
```suggestion
		update server_controller
		   set address     = @controller_address,
			   description = @controller_description,
			   update_time = now()
		 where private_id = @controller_private_id;
```

As per our [sql style guide](https://github.com/hashicorp/boundary/blob/main/sql-style-guide.md#data-manipulation-statements) you want to align the keywords to form a "river".

internal/server/repository_controller_test.go
```diff
@@ -14,6 +13,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const (
+	removeControllerSql = `DELETE FROM server_controller WHERE private_id = $1`
```
tmessi, line 17:
```suggestion
	removeControllerSql = `delete from server_controller where private_id = $1`
```

Again, as per our sql style guide, use lower case for sql keywords.

internal/server/repository_controller.go
```diff
@@ -76,3 +70,48 @@ func (r *Repository) UpsertController(ctx context.Context, controller *Controlle
 
 	return int(rowsUpdated), nil
 }
+
+func (r *Repository) UpdateController(ctx context.Context, controller *Controller) (int, error) {
```
tmessi, line 74: I think we should make this method something more domain specific instead of a generic "update" method. Perhaps "UpdateControllerStatus"? This would help clarify that it has a particular purpose, and is the reason the `update_time` is handled differently compared to other "update" repository methods.

Please also include a go doc for the exported method.

tmessi (changes requested):
Thanks for coming back to this. I added a couple requests. Also can you update the commit message with additional details? There is a lot of good context in the PR message that would be good to also include in the commit history. And you might want to include a `Refs` footer to the original commit.

RyanDerr:
> Thanks for coming back to this. I added a couple requests. Also can you update the commit message with additional details? There is a lot of good context in the PR message that would be good to also include in the commit history. And you might want to include a `Refs` footer to the original commit.

Alright went ahead and pushed up the changes, let me know if the commit message is a little to long or is lacking any context we may find useful.

tmessi (approved):
Looks good, my one nit would be that you update the commit `Ref` footer to reference the commit instead of the PR, ie:

```
Refs: 6c0b3a8a1c5c4e455006ea712f4eb868929c67fa
```

PR #5629, opened 2025-03-27
bosorawis domain iam implement getRoleScopeType

bosorawis:
implement looking up scope ID with role ID


internal/iam/role.go
```diff
@@ -327,3 +327,37 @@ func (p *projectRole) Actions() map[string]action.Type {
 	ret[action.SetPrincipals.String()] = action.SetPrincipals
 	return ret
 }
+
+// getRoleScopeId returns scopeID for the Role from the base type iam_role table
+// use this to get scope ID to determine which of the role subtype tables to operate on
+func getRoleScopeId(ctx context.Context, r db.Reader, roleId string) (string, error) {
+	const (
+		op    = "iam.getRoleScopeId"
+		query = `select scope_id from iam_role where public_id = ?`
```
dkanney, line 336: I think this should live in `internal/iam/query.go` based on our current pattern.

There are some plaintext queries that live outside `query.go` but they seem to only be used in special cases/helper functions (`Db.Now`, Db.hasTable, `InitCommand.verifyOplogIsEmpty`)

internal/iam/role.go
```diff
@@ -327,3 +327,37 @@ func (p *projectRole) Actions() map[string]action.Type {
...
+	)
+	if roleId == "" {
+		return "", errors.New(ctx, errors.InvalidParameter, op, "missing role ID")
+	}
+	if r == nil {
+		return "", errors.New(ctx, errors.InvalidParameter, op, "missing db.Reader")
+	}
+	rows, err := r.Query(ctx, query, []any{roleId})
+	if err != nil {
+		return "", errors.Wrap(ctx, err, op, errors.WithMsg("failed to lookup role scope ID"))
```
dkanney, line 346: Might be helpful to include the `roleId` in this error message, e.g. `"failed to lookup scope ID for role ID r_1234"`

bosorawis: added!

internal/iam/query.go
```diff
@@ -221,4 +221,9 @@ const (
 	estimateCountScopes = `
 		select reltuples::bigint as estimate from pg_class where oid in ('iam_scope'::regclass)
 	`
+
+	scopeIdFromRoleIdQuery = `
+	   select scope_id
+		 from iam_role
+		where public_id = $1;`
```
tmessi, line 228: I believe we prefer either `?` or a named arg like `@public_id`.

internal/iam/role.go
```diff
@@ -327,3 +328,34 @@ func (p *projectRole) Actions() map[string]action.Type {
 	ret[action.SetPrincipals.String()] = action.SetPrincipals
 	return ret
 }
+
+// getRoleScopeId returns scopeID for the Role from the base type iam_role table
```
tmessi, line 332: Since this uses a reader, I would place this in one of the `repository_` files. It should really only be called by a repository, so it would make it easier to find it if is in the same file as the repository.

internal/iam/role.go
```diff
@@ -327,3 +328,34 @@ func (p *projectRole) Actions() map[string]action.Type {
...
+// use this to get scope ID to determine which of the role subtype tables to operate on
+func getRoleScopeId(ctx context.Context, r db.Reader, roleId string) (string, error) {
```
tmessi, line 334: I would consider making this function return the role's scope type, rather than the scope id. I think the common use of this will be for the repository to switch on the scope type. So perhaps make this:

```go
func getRoleScopeType(ctx context.Context, r db.Reader, roleId string) (scope.Type, error)
```

tmessi (changes requested):

internal/iam/repository_role.go
```diff
@@ -389,3 +390,44 @@ func (r *Repository) estimatedRoleCount(ctx context.Context) (int, error) {
...
+	if err != nil {
+		return scope.Unknown, errors.Wrap(ctx, err, op, errors.WithMsg(fmt.Sprintf("failed to lookup role scope for :%s", roleId)))
+	}
+	var scopeId string
+	cnt := 0
+	for rows.Next() {
+		cnt++
+		if err := r.ScanRows(ctx, rows, &scopeId); err != nil {
+			return scope.Unknown, errors.Wrap(ctx, err, op, errors.WithMsg(fmt.Sprintf("failed scan results from querying role scope for :%s", roleId)))
+		}
```
dkanney, line 414: I'd only expect one value to be returned for `scopeId`, but just incase there isn't, what if we added an extra guard around that, like:

```suggestion
		}
        if cnt > 1 {
			return scope.Unknown, errors.New(ctx, errors.MultipleRecords, op, fmt.Sprintf("expected 1 row but got: %d", cnt))
		}
```

dkanney (approved):

PR #5638, opened 2025-04-01
bosorawis sql split global grants scope table

bosorawis:
Split `iam_role_global_individual_grant_scope` to `iam_role_global_individual_org_grant_scope` and `iam_role_global_individual_project_grant_scope` and allow `children` to be used in `iam_role_global_individual_project_grant_scope.grant_scope` to support use case where roles at global scope are granted `children` and individual projects grants. 

Add triggers to 

- propagate changing `iam_role_global_individual_project_grant_scope.grant_scope` from `children` to `individual` and vice-versa when `iam_role_global.grant_scope` is updated
- add trigger to delete all `iam_role_global_individual_project_grant_scope` and `iam_role_global_individual_org_grant_scope` when a `iam_role_global.grant_scope` is changed to `descendants`
- add trigger to delete all `iam_role_global_individual_org_grant_scope` when a `iam_role_global.grant_scope` is changed to `children` 
- add trigger to delete all `iam_role_org_individual_grant_scope` when a `iam_role_org.grant_scope` is changed to `children`

### DIFF
--- a/internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
@@ -78,7 +78,7 @@ begin;
       'update_iam_role_table_update_time is used to automatically update the update_time '
       'of the base table whenever one of the subtype iam_role tables are updated';
 
-  create function delete_base_iam_role() returns trigger
+  create function delete_iam_role_subtype() returns trigger
   as $$
   begin
     delete from iam_role
@@ -86,8 +86,8 @@ begin;
     return null; -- result is ignored since this is an after trigger
   end;
     $$ language plpgsql;
-    comment on function delete_base_iam_role() is
-      'delete_base_iam_role is used to automatically delete associated iam_role entry'
+    comment on function delete_iam_role_subtype() is
+      'delete_iam_role_subtype is used to automatically delete associated iam_role entry'
       'since domain implementation performs deletion on the child table which does not cleanup the base iam_role table ';
 
   -- global iam_role must have a scope_id of global.
@@ -138,10 +138,10 @@ begin;
     for each row execute procedure insert_role_subtype();
 
   create trigger insert_grant_scope_update_time before insert on iam_role_global
-    for each row execute procedure insert_grant_scope_update_time();  
+    for each row execute procedure insert_grant_scope_update_time();
 
   create trigger insert_grant_this_role_scope_update_time before insert on iam_role_global
-    for each row execute procedure insert_grant_this_role_scope_update_time();  
+    for each row execute procedure insert_grant_this_role_scope_update_time();
 
   create trigger update_iam_role_global_grant_scope_update_time before update on iam_role_global
     for each row execute procedure insert_grant_scope_update_time();
@@ -152,8 +152,8 @@ begin;
   create trigger update_iam_role_global_base_table_update_time after update on iam_role_global
     for each row execute procedure update_iam_role_table_update_time();
 
-  create trigger delete_base_iam_role after delete on iam_role_global
-    for each row execute procedure delete_base_iam_role();
+  create trigger delete_iam_role_subtype after delete on iam_role_global
+    for each row execute procedure delete_iam_role_subtype();
 
   create trigger default_create_time_column before insert on iam_role_global
     for each row execute procedure default_create_time();
@@ -161,7 +161,7 @@ begin;
   create trigger update_time_column before update on iam_role_global
     for each row execute procedure update_time_column();
 
-  create trigger update_version_column after update on iam_role_global 
+  create trigger update_version_column after update on iam_role_global
     for each row execute procedure update_version_column();
 
   create trigger immutable_columns before update on iam_role_global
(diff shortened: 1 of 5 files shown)


internal/db/sqltest/tests/iam/iam_role_global.sql
```diff
@@ -2,23 +2,31 @@
 -- SPDX-License-Identifier: BUSL-1.1
 
 begin;
-  select plan(14);
```
dkanney, line 5: Is this the number of tests to run? 🤔

internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
```diff
@@ -65,6 +65,28 @@ begin;
...
+      delete from iam_role_global_individual_project_grant_scope
+        where role_id = new.public_id;
+    end if;
+    return new;
+  end;
+  $$ language plpgsql;
+  comment on function delete_global_role_individual_role_grant_scope() is
+    'delete_global_role_individual_role_grant_scope deletes individual role grants from '
+    'iam_role_global_individual_org_grant_scope and iam_role_global_individual_project_grant_scope to remove '
+    'redundant grants';
```
tmessi, line 87: Is this meant to be a part of this PR? This seems like it would belong in a separate one. Also I'm don't think this should be done via a trigger. I think the repository should handle this via a query within a transaction.

internal/db/schema/migrations/oss/postgres/100/02_iam_role_org.up.sql
```diff
@@ -23,6 +23,24 @@ begin;
...
+    if new.grant_scope = 'children' then
+      delete from iam_role_org_individual_grant_scope
+        where role_id = new.public_id;
+    end if;
+    return new;
+  end;
+  $$ language plpgsql;
+  comment on function delete_org_role_individual_role_grant_scope() is
+    'delete_org_role_individual_role_grant_scope deletes individual role grants from '
+    'iam_role_org_individual_grant_scope to remove redundant grants';
```
tmessi, line 39: Same with this function, I think the repository should handle this, and this should be in a separate PR.

tmessi (changes requested):

internal/db/sqltest/tests/iam/iam_role_global.sql
```diff
@@ -60,87 +68,409 @@ begin;
...
+  -- 3a) insert invalid row: grant_scope = 'descendants'
+  prepare insert_invalid_individual_org_grant_scope_descendants as
+    insert into iam_role_global_individual_org_grant_scope
+        (role_id, grant_scope, scope_id)
+    values
+        ('r_3333333333', 'descendants', 'p____bwidget');
+  select throws_like(
+    'insert_invalid_individual_org_grant_scope_descendants',
+    'new row for relation "iam_role_global_individual_org_grant_scope" violates check constraint "only_individual_grant_scope_allowed"',
+    'check(grant_scope = "individual") is enforced'
```
dkanney, line 111: Does this mean we only allow the value `"individual"` in `iam_role_global_individual_org_grant_scope.grant_scope`? If so, we could drop that column right? Since it'd be the same for every row

internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
```diff
@@ -160,6 +182,47 @@ begin;
        constraint only_individual_grant_scope_allowed
          check(
           grant_scope = 'individual'
+        ),
```
dkanney, line 185: Hmm... is it necessary to store the same value in this column for every row?

internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
bosorawis: Could you expand on why is a trigger not preferable here? Happy to make the change but just curious why this shouldn't be enforced in the schema

internal/db/sqltest/tests/iam/iam_role_global.sql
bosorawis: This is to enforce that `iam_role_global.grant_scope` is also set to `individual` in the schema

internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
bosorawis: this column is used to enforce that the corresponding role in `iam_role_global` also has `grant_scope` set to the allowed value (`individual`)

internal/db/sqltest/tests/iam/iam_role_global.sql
bosorawis: Number of statements being executed in this transaction

internal/db/schema/migrations/oss/postgres/100/01_iam_role_global.up.sql
bosorawis: Removed this trigger along with `delete_org_role_individual_role_grant_scope` mentioned in the other comment.

tmessi (changes requested):

tmessi (approved):

PR #5642, opened 2025-04-03
test: add grants tests for managed groups resource

mikemountain:


bosorawis (commented):
Could you also add these following test cases?

- tests for `GetManagedGroup` API
- tests covering not granting the scope but granting permissions (something like granting `ids=*;type=*;actions=*` but not `this` scope the resource is in
- tests covering pinned ID 
    - Creating an auth method `amoidc_123456`
    - use grants like `ids=amoidc_123456;type=managed-group;actions=list,no-op`
    - verify that the user can list managed groups from `amoidc_123456` but not from another auth method

bosorawis (approved):
Nice! LGTM! 🎉

PR #5644, opened 2025-04-10
test: add grants tests for auth tokens resource

mikemountain:


internal/daemon/controller/handlers/authtokens/grants_test.go
```diff
@@ -65,49 +68,191 @@ func TestGrants_ReadActions(t *testing.T) {
...
+				userFunc: iam.TestUserGroupGrantsFunc(t, conn, kmsCache, globals.GlobalPrefix, password.TestAuthMethodWithAccount, []iam.TestRoleGrantsRequest{
+					{
+						RoleScopeId: org1.PublicId,
+						Grants:      []string{"ids=*;type=*;actions=*"},
+						GrantScopes: []string{globals.GrantScopeThis},
+					},
+				}),
+				wantErr:         nil,
+				wantIDs:         nil,
+				expectOutfields: nil,
```
bosorawis, line 247: Could you add 2 test cases around `recursive=false`? 
1. listing at `global` scope when the role isn't granted global scope - expect forbidden error
2. happy path

internal/daemon/controller/handlers/authtokens/grants_test.go
```diff
@@ -65,49 +68,191 @@ func TestGrants_ReadActions(t *testing.T) {
...
+					globals.UserIdField,
+					globals.AuthMethodIdField,
+					globals.AccountIdField,
+					globals.CreatedTimeField,
+					globals.UpdatedTimeField,
+					globals.ApproximateLastUsedTimeField,
+					globals.ExpirationTimeField,
+					globals.AuthorizedActionsField,
+				},
+			},
```
bosorawis, line 202: could you add a copy of this test case but listing at `global` instead of `org1` to test the recursive list behavior?

bosorawis (changes requested):
requesting a few more test cases

bosorawis (approved):

dkanney (commented):
I might be missing something, but I think we still need a test for `GetAuthTokenRequest{}`, no?

PR #5645, opened 2025-04-10
Add parsing logic for User-Agent headers

aditya2548:
- Adds logic to parse user-agent headers & emit them in telemetry events, which is used to identify API request source (client-agent).
- Adds Go tags for TS telemetry
- RFC: https://go.hashi.co/rfc/icu-198
- Jira: https://hashicorp.atlassian.net/browse/ICU-16735
- Corresponding changes in Boundary client agent: https://github.com/hashicorp/boundary-client-agent/pull/336

**Note:** This code was originally implemented and reviewed in [boundary-enterprise](https://github.com/hashicorp/boundary-enterprise/pull/1472). The commits have been cherry-picked into this repository to reuse the reviewed functionality. No functional changes were made during the port


internal/daemon/controller/const.go
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package controller
+
+const (
+	// userAgentProductKey and userAgentProductVersionKey are used to define user agent in the observation events
+	userAgentProductKey        = "user_agent_product"
+	userAgentProductVersionKey = "user_agent_product_version"
+)
```
tmessi, line 10: We can many consts within this package already, so I don't think adding a `const.go` is advisable. We don't tend to use a `const.go`, and in the few places where we have, it is in some small isolated packages like [kms](https://github.com/hashicorp/boundary/tree/main/internal/kms) and all of the consts for that package are contained in that file.

Instead I think you could just add these to `gateway.go` where the consts are used to set values.

internal/daemon/controller/gateway.go
```diff
@@ -93,6 +95,27 @@ func correlationIdAnnotator(_ context.Context, req *http.Request) metadata.MD {
...
+	}
+	if _, err := version.NewSemver(agentVersion); err != nil {
+		// Invalid version
+		return metadata.MD{}
+	}
+	return metadata.New(map[string]string{
+		userAgentProductKey:        agentName,
+		userAgentProductVersionKey: agentVersion,
+	})
+}
```
tmessi, line 117: This does not seem to align with the format outlined in [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110#name-user-agent). Specifically it does not deal with:

- Comments in user agent strings
- Multiple products and versions being specified.

internal/daemon/controller/gateway_test.go
```diff
@@ -95,6 +95,53 @@ func Test_correlationIdAnnotator(t *testing.T) {
...
+			Header: map[string][]string{
+				"User-Agent": {"Boundary-client-agent/0.1.4"},
+			},
+		}
+
+		md := userAgentHeadersAnnotator(context.Background(), req)
+		require.NotNil(t, md)
+		assert.Equal(t, "Boundary-client-agent", md.Get(userAgentProductKey)[0])
+		assert.Equal(t, "0.1.4", md.Get(userAgentProductVersionKey)[0])
+	})
```
tmessi, line 112: We should have additional tests to cover cases for other valid user agent strings. We should be able to parse user agent strings with multiple product/product versions and that contain comments.

internal/event/event.go
```diff
@@ -52,6 +52,7 @@ type Request struct {
 	Endpoint               string           `json:"endpoint,omitempty" class:"public"`  // std audit field
 	Details                proto.Message    `json:"details,omitempty"`                  // boundary field
 	DetailsUpstreamMessage *UpstreamMessage `json:"details_upstream_message,omitempty"` // boundary field
+	UserAgent              *UserAgent       `json:"user_agent,omitempty"`               // boundary field
```
tmessi, line 55: This doesn't seem to align with the proposed event structure that was outlined in the RFC. I thought the structure would support a list of user agents, ie:

```json
"user_agent": [
  {
    "product": "agent1",
    "product_version": "0.1.3"
  },
  {
    "product": "agent2",
    "product_version": "2.17b3"
    "comments": [
       "comment1",
       "comment2"
    ],
  }
],
```

tmessi (changes requested):

internal/daemon/controller/const.go
aditya2548: Thanks for the feedback! As mentioned in the previous [PR](https://github.com/hashicorp/boundary-enterprise/pull/1472) discussion, I moved the constants to a separate `const.go` file since they're used in both `gateway.go` and `interceptor.go`. I went with this approach to align with a pattern I've seen in a few other packages.

That said, happy to move them into `gateway.go` if you'd prefer to keep them localized despite the cross-file usage, just let me know what works best!

internal/daemon/controller/gateway.go
aditya2548: Thanks for the feedback! Apologies for any confusion, my understanding was that we had aligned on the simpler structure outlined [here](https://docs.google.com/document/d/1wPI1GOqW38XkW2Q1_Bz-Yt_bRKGtvYITiyEicXHLJ8s/edit?tab=t.0#heading=h.r2epnlqr90bx).

I implemented it this way to keep things straightforward and aligned with that proposed format. That said, I definitely see the value in supporting the full format of RFC 9110, including multiple user agents and comments, especially if we anticipate more complex user-agent strings in the future.

Also, do you think we’d need to parse and store the comments portion of the user agent for telemetry purposes? Just trying to get a sense of how much of the RFC format we’d want to support at this stage.

On a related note, how should we think about representing this in Snowflake? Would we have multiple rows per event (one per user agent), or nest them in a single structured field?

Happy to revisit the implementation based on where we want to go with this. Just wanted to get your thoughts before making changes.

internal/daemon/controller/const.go
tmessi: Yea, again, since I don't think we should place all consts from the controller package in this file, it would be better to just place these in `gateway.go`. I don't think it is an issue that other files also use the same consts.

internal/daemon/controller/gateway.go
tmessi: I think, since we decided on using the User-Agent header, that we should support parsing the full format, including comments, and include all of it in the event. This allows any consumer of the event to have access to these details.

aditya2548: Sure, made changes to parse the complete user-agent header as per RFC 9110 format.

internal/daemon/controller/gateway_test.go
aditya2548: I've added more test cases to cover valid, invalid and a mix of both type of agents

internal/event/event.go
aditya2548: Changed the event structure to adhere to this format

internal/daemon/controller/gateway.go
```diff
@@ -93,6 +96,63 @@ func correlationIdAnnotator(_ context.Context, req *http.Request) metadata.MD {
 	})
 }
 
+func userAgentHeadersAnnotator(_ context.Context, req *http.Request) metadata.MD {
+	userAgent := req.Header.Get("User-Agent")
+	if userAgent == "" {
+		return metadata.MD{}
+	}
+	// Regular expression to parse product, version, and comments
+	re := regexp.MustCompile(`(?P<product>[^\s/()]+)/(?P<version>[^\s()]+)(?: \((?P<comments>[^)]+)\))?`)
```
johanbrandhorst, line 105: Could you move this invocation to a package level (unexported) global variable? As it is, this regex will be compiled on every request to the controller. Also, is there a source for this regex, like an RFC or some online reference?

internal/daemon/controller/gateway.go
```diff
@@ -93,6 +96,63 @@ func correlationIdAnnotator(_ context.Context, req *http.Request) metadata.MD {
...
+	matches := re.FindAllStringSubmatch(userAgent, -1)
+
+	var userAgents []map[string]interface{}
```
johanbrandhorst, line 108: I see that we're marshaling this to JSON, but it seems like on the other side of the gRPC connection we're unmarshalling into a `[]*event.UserAgent` value. Could we use that type here too? If not, could we at least define an intermediate type, something like
```go
type parsedUserAgent struct {
    Product string
    ProductVersion string
    Comments []string
}
```

internal/daemon/controller/gateway.go
```diff
@@ -93,6 +96,63 @@ func correlationIdAnnotator(_ context.Context, req *http.Request) metadata.MD {
...
+
+	if len(userAgents) == 0 {
+		return metadata.MD{}
+	}
+	userAgentJSON, err := handlers.JSONMarshaler().Marshal(userAgents)
+	if err != nil {
+		return metadata.MD{}
+	}
+	return metadata.New(map[string]string{
+		"userAgents": string(userAgentJSON),
```
johanbrandhorst, line 152: Now that we're using this key (`userAgents`) as a way to communicate between this code and the code in `interceptor.go`, it would be nice if it was a constant in this package instead. What do you think?

internal/daemon/controller/gateway_test.go
```diff
@@ -95,6 +96,111 @@ func Test_correlationIdAnnotator(t *testing.T) {
...
+
+	t.Run("No user-agent header", func(t *testing.T) {
+		t.Parallel()
+		req := &http.Request{Header: map[string][]string{}}
+		md := userAgentHeadersAnnotator(context.Background(), req)
+		require.NotNil(t, md)
+		assert.Empty(t, md)
+	})
+
+	t.Run("Invalid user-agent formats", func(t *testing.T) {
```
johanbrandhorst, line 157: Nit: did you mean for this test to run in parallel too?
```suggestion
	t.Run("Invalid user-agent formats", func(t *testing.T) {
		t.Parallel()
```

internal/daemon/controller/gateway_test.go
```diff
@@ -95,6 +96,111 @@ func Test_correlationIdAnnotator(t *testing.T) {
...
+		require.NoError(t, err)
+		require.Len(t, parsed, 2)
+
+		assert.Equal(t, "Boundary-client-agent", parsed[0]["product"])
+		assert.Equal(t, "0.1.4", parsed[0]["product_version"])
+		assert.ElementsMatch(t, []string{"foo", "bar"}, parsed[0]["comments"].([]interface{}))
+
+		assert.Equal(t, "AnotherApp", parsed[1]["product"])
+		assert.Equal(t, "2.0.0", parsed[1]["product_version"])
+		assert.ElementsMatch(t, []string{"baz"}, parsed[1]["comments"].([]interface{}))
```
johanbrandhorst, line 146: Can we unmarshal into `[]*events.UserAgent`? I think that'd make this code a bit nicer too.

internal/daemon/controller/interceptor.go
```diff
@@ -448,22 +448,42 @@ func eventsRequestInterceptor(
...
+				if err := handlers.JSONMarshaler().Unmarshal([]byte(rawHeader), &userAgents); err == nil {
+					// Filter out any agents missing required fields
+					var filteredAgents []*event.UserAgent
+					for _, ua := range userAgents {
+						if ua.Product != "" && ua.ProductVersion != "" {
+							filteredAgents = append(filteredAgents, ua)
+						}
+					}
+					userAgents = filteredAgents
+				}
```
johanbrandhorst, line 465: If there was an error unmarshaling this value, it is likely that it was our fault, so maybe we should error in that case?

internal/event/event_observation_test.go
```diff
@@ -503,6 +503,50 @@ func Test_composeFromTelemetryFiltering(t *testing.T) {
...
+			fromOp: Op("with-request-client-headers-with-telemetry"),
+			opts: []Option{
+				WithId("with-request-client-headers-with-telemetry"),
+				WithRequestInfo(TestRequestInfo(t)),
+				WithFlush(),
+				WithRequest(&Request{
+					Operation: "op",
+					Endpoint:  "/auth-tokens/<id>",
+					Details:   testWorkerStatus(t),
+					UserAgent: &UserAgent{
```
johanbrandhorst, line 517: Looks like this should be 
```suggestion
					UserAgents: []*UserAgent{{
```

internal/event/event_observation_test.go
```diff
@@ -503,6 +503,50 @@ func Test_composeFromTelemetryFiltering(t *testing.T) {
...
+						Product:        "Boundary-client-agent",
+						ProductVersion: "0.1.4",
+					},
```
johanbrandhorst, line 520:
```suggestion
					}},
```

johanbrandhorst (changes requested):
Thanks!

internal/daemon/controller/gateway.go
aditya2548: Sure, moved to package level variable. This regex was written by me to match the format specified in [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110#name-user-agent). It aims to capture the product/version tokens and optional comments section, as described in the spec.

internal/daemon/controller/gateway.go
aditya2548: Updated the code to unmarshal into `[]*event.UserAgent`

internal/daemon/controller/gateway.go
aditya2548: Makes sense, done!

internal/daemon/controller/gateway_test.go
aditya2548: Updated the test cases to use `events.UserAgent`

internal/daemon/controller/interceptor.go
aditya2548: Makes sense, Updated to error out in this case

internal/event/event_observation_test.go
aditya2548: Done, thank you!

internal/daemon/controller/gateway.go
```diff
@@ -93,6 +104,61 @@ func correlationIdAnnotator(_ context.Context, req *http.Request) metadata.MD {
...
+			}
+		}
+
+		userAgents = append(userAgents, agentData)
+	}
+
+	if len(userAgents) == 0 {
+		return metadata.MD{}
+	}
+	userAgentJSON, err := handlers.JSONMarshaler().Marshal(userAgents)
```
tmessi, line 153: Rather than parsing, marshaling to JSON string, and then having to unmarshal again in the grpc interceptor, can we just add the original user agent string to the metadata here, and then do this parsing in the interceptor?

internal/daemon/controller/gateway.go
```diff
@@ -93,6 +104,61 @@ func correlationIdAnnotator(_ context.Context, req *http.Request) metadata.MD {
...
+		agentVersion := strings.TrimSpace(match[2])
+
+		if strings.HasPrefix(agentVersion, "v") {
+			// Invalid version format (starting with 'v')
+			continue
+		}
+		if _, err := version.NewSemver(agentVersion); err != nil {
+			// Invalid version
+			continue
+		}
```
tmessi, line 126: As far as I can tell, there is nothing in the spec that defines the format of the version.

aditya2548: Yes agreed. The spec doesn’t enforce a strict version format. However, we agreed in the [RFC](https://docs.google.com/document/d/1wPI1GOqW38XkW2Q1_Bz-Yt_bRKGtvYITiyEicXHLJ8s/edit?tab=t.0#heading=h.rno67ke8h473) to use semver without a leading v for consistency, analytics accuracy and easier parsing. That’s why we’re validating the version this way here.

tmessi: That makes sense for the client-agent's User Agent string. But this is applying the same version validation on all products in the user agent header. If we want to apply this validation, we need to only do it if the product is the boundary client agent.

internal/daemon/controller/gateway.go
```diff
@@ -23,13 +24,22 @@ import (
...
 
-const gatewayTarget = ""
+const (
+	gatewayTarget = ""
+
+	// userAgentsKey defines the user-agent header key for the gRPC gateway
+	userAgentsKey = "userAgents"
+)
+
+// Regular expression to parse user-agent product, version, and comments
```
johanbrandhorst, line 41: Thanks for moving this to the global scope, could you add a reference to the RFC with the definition for this format? It appears to be here: https://datatracker.ietf.org/doc/html/rfc9110#name-user-agent

internal/daemon/controller/gateway_test.go
```diff
@@ -95,6 +97,126 @@ func Test_correlationIdAnnotator(t *testing.T) {
...
+		require.Len(t, parsed, 1)
+		expected := []*event.UserAgent{
+			{
+				Product:        "Boundary-client-agent",
+				ProductVersion: "0.1.4",
+			},
+		}
+		assert.Equal(t, expected, parsed)
+	})
+}
```
johanbrandhorst, line 218: We don't have to use all of these, but could you try some of the more complex user agents from here: https://github.com/mileusna/useragent/blob/master/ua_test.go (MIT licensed).

johanbrandhorst (changes requested):

internal/daemon/controller/gateway.go
aditya2548: Sure, made changes to parse user agent in interceptor

internal/daemon/controller/gateway.go
aditya2548: Updated code to validate version only if the product is Boundary-client-agent

internal/daemon/controller/gateway.go
aditya2548: Done!

internal/daemon/controller/gateway_test.go
aditya2548: Done! Combined a web type of header and an unstandard one, and added that test case

johanbrandhorst (approved):
Thanks!

tmessi (changes requested):

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
 
 * Adds support for Azure Virtual Machine Scale Sets in the Azure plugin
   ([PR](https://github.com/hashicorp/boundary-plugin-azure/pull/22)).
+* Adds support to parse User-Agent headers & emit them in telemetry events
```
johanbrandhorst, line 11: Nit: I don't think we use ampersands like this in prose generally
```suggestion
* Adds support to parse User-Agent headers and emit them in telemetry events
```

johanbrandhorst (dismissed):

CHANGELOG.md
aditya2548: Done!

internal/daemon/controller/interceptor.go
```diff
@@ -464,6 +481,52 @@ func eventsRequestInterceptor(
 	}
 }
 
+// parseUserAgents extracts structured UserAgent data from a raw User-Agent header string.
+// It filters out entries with invalid or non-semver versions (e.g., versions starting with 'v'),
```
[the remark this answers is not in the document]
aditya2548: Made changes to reflect this

tmessi (approved):

johanbrandhorst (approved):

PR #5651, opened 2025-04-16
Test unauthenticated user

dillanb-hashi:
Added a test that checks that certain commands fail if the user is not authenticated. 

jira: https://hashicorp.atlassian.net/browse/ICU-11253


testing/internal/e2e/tests/base/auth_unauthenticated_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// Logout
+	output = e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"logout",
+		),
+	)
+	require.NoError(t, output.Err, string(output.Stderr))
+
+	// Attempt commands that unauthetnicated users can make
```
johanbrandhorst, line 38: Nit:
```suggestion
	// Attempt commands that unauthenticated users can make
```

testing/internal/e2e/tests/base/auth_unauthenticated_test.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/hashicorp/boundary/testing/internal/e2e"
+	"github.com/hashicorp/boundary/testing/internal/e2e/boundary"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCliUnauthenticatedUserAccess(t *testing.T) {
```
johanbrandhorst, line 15: Could you add a quick synopsis of the test as a comment on this function?

testing/internal/e2e/tests/base/auth_unauthenticated_test.go
```diff
@@ -0,0 +1,76 @@
...
+	require.Error(t, output.Err, "Unauthenticated user was able to access private resource")
+
+	output = e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"users", "list",
+			"--recursive",
+		),
+	)
+	require.Error(t, output.Err, "Unauthenticated user was able to access private resource")
+}
```
johanbrandhorst, line 76: It would be interesting if this test (or maybe a separate test) could try changing the permissions for the unauthenticated user and see that they are reflected in what the user is able to do. Maybe a follow up?

testing/internal/e2e/tests/base/auth_unauthenticated_test.go
```diff
@@ -0,0 +1,76 @@
...
+	require.NoError(t, output.Err, string(output.Stderr))
+
+	// Check user cannot make requests that require authentication
+	output = e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"targets", "list",
+			"--recursive",
+		),
+	)
+	require.Error(t, output.Err, "Unauthenticated user was able to access private resource")
```
moduli, line 59: Do we want to include a check on the status code that is returned here? This is an example from another test
```
var response boundary.CliError
err = json.Unmarshal(output.Stderr, &response)
require.NoError(t, err)
require.Equal(t, http.StatusForbidden, int(response.Status))
```

testing/internal/e2e/tests/base/auth_unauthenticated_test.go
moduli: Yea. Let's do that in a separate ticket/PR.

I'm trying to play around with it locally, but I'm not sure I'm seeing the expected behavior

johanbrandhorst (approved):

johanbrandhorst (dismissed):
Thank you!

testing/internal/e2e/tests/base/auth_unauthenticated_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	output = e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"sessions", "list",
+			"--recursive",
+			"-format", "json",
+		),
+	)
+	require.Error(t, output.Err, "Unauthenticated user was able to access private resource")
+	fmtError = strings.Split(string(output.Stderr), "\n")[2]
```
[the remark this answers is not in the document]
dillanb-hashi: The last element is an empty string. so I would have to do 

s := ss[len(ss)-2]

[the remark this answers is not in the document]
moduli: Hm... that doesn't seem any better. We can leave as-is for now.

testing/internal/e2e/tests/base/auth_unauthenticated_test.go
moduli: Looks like this got lost in the rebase?

moduli (approved):

PR #5668, opened 2025-04-21
Database Init Flags Refactor

RyanDerr:
# Overview
This commit refactors the database initialization flags to prevent early exits and ensure explicit handling of flag dependencies. If one flag relies on resources skipped by another, the dependent flag will also be skipped.

Key Changes:
- Introduced a `initFlags` struct to organize existing flags and establish a clear pattern for adding future flags.
- Added receiver functions on the `initFlags` struct to determine whether a flag should be skipped based on provided options or dependencies.
- Refactored the `Run()` function to avoid early returns when a flag skips resource creation. Instead, a log message is provided to indicate the skipped option, and the process continues with resource creation.

### DIFF
--- a/internal/cmd/commands/database/flags.go
+++ b/internal/cmd/commands/database/flags.go
@@ -1,0 +1,150 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package database
+
+import "fmt"
+
+const (
+	flagConfigName                           = "config"
+	flagConfigKmsName                        = "config-kms"
+	flagLogLevelName                         = "log-level"
+	flagLogFormatName                        = "log-format"
+	flagSkipInitialLoginRoleName             = "skip-initial-login-role-creation"
+	flagSkipInitialAuthenticatedUserRoleName = "skip-initial-authenticated-user-role-creation"
+	flagSkipAuthMethodName                   = "skip-auth-method-creation"
+	flagSkipScopesName                       = "skip-scopes-creation"
+	flagSkipHostResourcesName                = "skip-host-resources-creation"
+	flagSkipTargetName                       = "skip-target-creation"
+	flagMigrationUrlName                     = "migration-url"
+)
+
+// initFlags contains the flags for the database init command, including optional
+// flags to control which initialization steps are skipped and the reasons for skipping.
+type initFlags struct {
+	flagConfig                                   []string
+	flagConfigKms                                string
+	flagLogLevel                                 string
+	flagLogFormat                                string
+	flagMigrationUrl                             string
+	flagSkipInitialLoginRoleCreation             bool
+	flagSkipInitialAuthenticatedUserRoleCreation bool
+	flagSkipAuthMethodCreation                   bool
+	flagSkipScopesCreation                       bool
+	flagSkipHostResourcesCreation                bool
+	flagSkipTargetCreation                       bool
+}
+
+// SkipInitialLoginRoleCreation checks if the creation of the initial anonymous login role should be skipped.
+// It returns a boolean indicating whether to skip and a reason string if skipping is applicable.
+//
+// The creation of the initial login role is skipped if:
+// - The flag for skipping the creation of the initial anonymous user role (`-skip-initial-login-role-creation`) is provided.
+//
+// Returns:
+// - bool: True if the creation should be skipped, false otherwise.
+// - string: A reason string explaining why the creation is skipped, or an empty string if not skipped.
+func (f *initFlags) SkipInitialLoginRoleCreation() (bool, string) {
+	if f.flagSkipInitialLoginRoleCreation {
+		return true, reasonFlagWasSet(flagSkipInitialLoginRoleName)
+	}
+	return false, ""
+}
+
+// SkipInitialAuthenticatedUserRoleCreation checks if the creation of the initial authenticated user role should be skipped.
+// It returns a boolean indicating whether to skip and a reason string explaining why the creation is skipped.
+//
+// The creation of the initial authenticated user role is skipped if:
+// - The flag for skipping the creation of the initial authenticated user role (`-skip-initial-authenticated-user-role-creation`) is provided.
+//
+// Returns:
+// - bool: True if the creation of the initial authenticated user role should be skipped, false otherwise.
+// - string: A reason string explaining why the creation is skipped, or an empty string if not skipped.
+func (f *initFlags) SkipInitialAuthenticatedUserRoleCreation() (bool, string) {
+	if f.flagSkipInitialAuthenticatedUserRoleCreation {
+		return true, reasonFlagWasSet(flagSkipInitialAuthenticatedUserRoleName)
+	}
+	return false, ""
+}
+
+// SkipAuthMethodCreation checks if the creation of authentication methods should be skipped.
+// It returns a boolean indicating whether to skip and a reason string explaining why the creation is skipped.
+//
+// Authentication method creation is skipped if:
+// - The flag for skipping authentication method creation (`-skip-auth-method-creation`) is provided.
+//
+// Returns:
+// - bool: True if the creation of authentication methods should be skipped, false otherwise.
+// - string: A reason string explaining why the creation is skipped, or an empty string if not skipped.
+func (f *initFlags) SkipAuthMethodCreation() (bool, string) {
+	if f.flagSkipAuthMethodCreation {
+		return true, reasonFlagWasSet(flagSkipAuthMethodName)
+	}
+	return false, ""
+}
+
+// SkipScopesCreation checks if the creation of default scopes should be skipped.
+// It returns a boolean indicating whether to skip and a reason string explaining why the creation is skipped.
+//
+// Scopes creation is skipped if:
+// - The flag for skipping scopes creation (`-skip-scopes-creation`) is provided.
+//
+// Returns:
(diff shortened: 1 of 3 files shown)


internal/cmd/commands/database/init.go
```diff
@@ -23,6 +23,20 @@ var (
 	_ cli.CommandAutocomplete = (*InitCommand)(nil)
 )
 
+type InitFlags struct {
```
johanbrandhorst, line 26: Does this need to be exported? Seems like we could unexport it. Also, could you add some docs for this struct?

internal/cmd/commands/database/init.go
```diff
@@ -37,17 +51,7 @@ type InitCommand struct {
...
-	flagLogLevel                                 string
-	flagLogFormat                                string
-	flagMigrationUrl                             string
-	flagSkipInitialLoginRoleCreation             bool
-	flagSkipInitialAuthenticatedUserRoleCreation bool
-	flagSkipAuthMethodCreation                   bool
-	flagSkipScopesCreation                       bool
-	flagSkipHostResourcesCreation                bool
-	flagSkipTargetCreation                       bool
+	InitFlags InitFlags
```
johanbrandhorst, line 54: Could we add some docs to this field? We have to make it clear that new flags should be added to the struct, not as a new field. Also, should this field be exported?

internal/cmd/commands/database/init.go
```diff
@@ -23,6 +23,20 @@ var (
...
+	flagLogLevel                                 string
+	flagLogFormat                                string
+	flagMigrationUrl                             string
+	flagSkipInitialLoginRoleCreation             bool
+	flagSkipInitialAuthenticatedUserRoleCreation bool
+	flagSkipAuthMethodCreation                   bool
+	flagSkipScopesCreation                       bool
+	flagSkipHostResourcesCreation                bool
+	flagSkipTargetCreation                       bool
+}
```
johanbrandhorst, line 38: Could you move the definition of this struct to live alongside the methods of the struct? Perhaps in a separate file?

internal/cmd/commands/database/init.go
```diff
@@ -543,3 +545,27 @@ func (c *InitCommand) verifyOplogIsEmpty(ctx context.Context) error {
...
+	return f.flagSkipScopesCreation
+}
+
+func (f *InitFlags) SkipHostResourcesCreation() bool {
+	return f.flagSkipHostResourcesCreation || f.SkipScopesCreation()
+}
+
+func (f *InitFlags) SkipTargetCreation() bool {
+	return f.flagSkipTargetCreation || f.SkipHostResourcesCreation() || f.SkipScopesCreation()
+}
```
johanbrandhorst, line 571: Could you document each of these methods?

internal/cmd/commands/database/init.go
```diff
@@ -321,178 +325,176 @@ func (c *InitCommand) Run(args []string) (retCode int) {
...
-		UserId:         user.PublicId,
-		UserName:       user.Name,
-	}
-	switch base.Format(c.UI) {
-	case "table":
-		c.UI.Output(generateInitialAuthTableOutput(authMethodInfo))
-	case "json":
-		jsonMap["auth_method"] = authMethodInfo
+	if c.InitFlags.SkipInitialLoginRoleCreation() {
+		c.UI.Info("Skipping creation of initial login role.")
```
johanbrandhorst, line 329: I wonder if it would be nice to include a little summary for the user to explain why this was skipped. If each of these helper methods returned a string explaining _why_ instead of just a bool, we could print that here.

```suggestion
	if shouldSkip, reason := c.InitFlags.SkipInitialLoginRoleCreation(); shouldSkip{
		c.UI.Infof("Skipping creation of initial login role: %s.", reason)
```

Could print

```
Skipping creation of initial login role: flag `-skip-initial-login-role` was set
```

What do you think?

johanbrandhorst (changes requested):
Could you add some tests for the InitFlags struct logic? Thanks!

internal/cmd/commands/database/init.go
RyanDerr: Good point, this does not need to be available outsize the `database` package.

internal/cmd/commands/database/init.go
RyanDerr: Love the idea!

RyanDerr: Alright updated some of the logic for this, let me know what ya think!

internal/cmd/commands/database/flags.go
```diff
@@ -0,0 +1,143 @@
...
+// - bool: True if the creation of host resources should be skipped, false otherwise.
+// - string: A reason string explaining why the creation is skipped, or an empty string if not skipped.
+func (f *initFlags) SkipHostResourcesCreation() (bool, string) {
+	if f.flagSkipHostResourcesCreation {
+		return true, reasonSkipHostResourcesCreation
+	}
+	if skip, reason := f.SkipScopesCreation(); skip {
+		return true, reason
+	}
+	return false, ""
```
johanbrandhorst, line 115: Nit, but you could simplify this to
```suggestion
	return f.SkipScopesCreation()
```

internal/cmd/commands/database/flags.go
```diff
@@ -0,0 +1,143 @@
...
+	if f.flagSkipTargetCreation {
+		return true, reasonSkipTargetCreation
+	}
+	if skip, reason := f.SkipHostResourcesCreation(); skip {
+		return true, reason
+	}
+	if skip, reason := f.SkipScopesCreation(); skip {
+		return true, reason
+	}
+	return false, ""
```
johanbrandhorst, line 142: Ditto

internal/cmd/commands/database/flags_test.go
```diff
@@ -0,0 +1,219 @@
...
+		{
+			name:           "SkipInitialLoginRoleCreation is false",
+			flags:          &initFlags{flagSkipInitialLoginRoleCreation: false},
+			expectedSkip:   false,
+			expectedReason: "",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			skip, reason := tt.flags.SkipInitialLoginRoleCreation()
```
johanbrandhorst, line 36: If you want these test cases to run in parallel you'll need to add it here too
```suggestion
		t.Run(tt.name, func(t *testing.T) {
			t.Parallel()
			skip, reason := tt.flags.SkipInitialLoginRoleCreation()
```
Same for the rest of these obviously.

internal/cmd/commands/database/flags_test.go
```diff
@@ -0,0 +1,219 @@
...
+		name           string
+		flags          *initFlags
+		expectedSkip   bool
+		expectedReason string
+	}{
+		{
+			name:           "SkipInitialLoginRoleCreation is true",
+			flags:          &initFlags{flagSkipInitialLoginRoleCreation: true},
+			expectedSkip:   true,
+			expectedReason: reasonSkipInitialLoginRoleCreation,
```
johanbrandhorst, line 25: I appreciate that this reduces boilerplate, but I think this actually devalues the test too. To me, this test should check that the skip and reason return the expected values. Right now, the expected reason is the constant that defines the string returned by the function, which doesn't actually check what the value is!

This is more generally called "a test oracle", which is a hard concept to explain without an example, but this serves as a great one, so let me try. Basically, ask yourself, what is your test _actually_ testing? This test is currently testing that the reason returned is the value of the expected constant. But that doesn't actually tell us anything about the value of the const, which is what the end user sees. If someone changed the const to include a spelling mistake, or changed the name of the flag, this test would keep passing. So in this case, the test oracle is the constant, but what I would like to see is that the test oracle is the string we expect the user to see. The difference is subtle but important. Does that make sense?

```suggestion
			expectedReason: "flag `-skip-initial-login-role` was set",
```

internal/cmd/commands/database/flags.go
```diff
@@ -0,0 +1,143 @@
...
+package database
+
+const (
+	reasonSkipInitialLoginRoleCreation             = "flag `-skip-initial-login-role` was set"
+	reasonSkipInitialAuthenticatedUserRoleCreation = "flag `-skip-initial-authenticated-user-role-creation` was set"
+	reasonSkipAuthMethodCreation                   = "flag `-skip-auth-method-creation` was set"
+	reasonSkipScopesCreation                       = "flag `-skip-scopes-creation` was set"
+	reasonSkipHostResourcesCreation                = "flag `-skip-host-resources-creation` was set"
+	reasonSkipTargetCreation                       = "flag `-skip-target-creation` was set"
+)
```
johanbrandhorst, line 13: Is there some way we can dynamically build these strings from the real flag names? Maybe make the flag names consts instead and use them in both places?

johanbrandhorst (changes requested):

internal/cmd/commands/database/flags_test.go
RyanDerr: That makes perfect sense and helps ensure that even if the constant value changes over time, we can still verify that the expected reason remains consistent with the intended behavior.

internal/cmd/commands/database/flags.go
RyanDerr: Absolutely we can define these as const strings for the flag names as part of this file and have a helper function generate the reason strings based on the flag name.

PR #5673, opened 2025-04-22
ListResolvableAliasesPermissions Hotfix

RyanDerr:
# Overview
Refactor `ACL` function `ListResolvableAliasesPermissions` to handle direct child scopes (e.g., global → org → project) correctly in the scenario a direct grant is given to a scope under your children, and prevent `500` errors when resolving aliases without proper scope permissions where the current only existing alias type,`Target`, lives.

# Changes 
- In the `ListResolvableAliasesPermissions` function for the `ACL`, the default case was removed and replaced with for edge case for a `global` scoped role with the `children` grant scope:
  ```go
  default:
	  // Since direct grants must be the same scope or downstream, the
	  // only possibility left for a children grant is that the parent is
	  // global and the grant is on the org -- if it was for projects it
	  // would need to be a descendants grant
	  if _, ok := childrenScopes[scope.Global.String()]; ok {
		  // We already looked at this scope in the children grants, so skip it
		  continue
	  }
  ```
  This was part of the [switch](https://github.com/hashicorp/boundary/blob/288fa91f7bce9a3bf26330acbc6d2ba45b02e8b4/internal/perms/acl.go#L452) statement within the [scopeMap](https://github.com/hashicorp/boundary/blob/288fa91f7bce9a3bf26330acbc6d2ba45b02e8b4/internal/perms/acl.go#L437) loop. The [replacement](https://github.com/hashicorp/boundary/blob/f15a7e71b9dc74d45dea229d28601c8f3e922b98/internal/perms/acl.go#L471-L472) addresses a scenario where a globally scoped role provides children scopes but can also directly grant permissions to projects under its children. The previous logic skipped these cases where you can assign scopes that live under your direct children, resulting in permissions not being built for projects where targets reside.

- For the functions [listResolvableAliases](https://github.com/hashicorp/boundary/blob/17b9224fa26c93c23efa0c0a60a838d57f2a382d/internal/alias/target/repository_alias_list_resolvable.go#L123), a condition was modified to handle an edge case where permissions derived from the `ACL` do not include any scopes under which the current target resources reside. In such cases, we now return no data for the end user as this is possible given the available permission slice returned from the `ACL`.

### DIFF
--- a/internal/alias/target/repository_alias_list_resolvable.go
+++ b/internal/alias/target/repository_alias_list_resolvable.go
@@ -74,13 +74,6 @@ func (r *Repository) listResolvableAliases(ctx context.Context, permissions []pe
 	}
 
 	directIds, directScopeIds, childAllScopes, allDescendants := splitPermissions(permissions)
-	// This condition checks if there are no direct IDs, no direct scope IDs, and no child scopes,
-	// while also ensuring that the "allDescendants" flag is not set, and if so returns early.
-	// An example scenario that can cause this is when a role on the global scope grants a user
-	// access to list aliases and read targets, but only within the global scope and its immediate children.
-	if !allDescendants && len(directIds) == 0 && len(directScopeIds) == 0 && len(childAllScopes) == 0 {
-		return []*Alias{}, time.Time{}, nil
-	}
 
 	opts, err := getOpts(opt...)
 	if err != nil {
@@ -122,8 +115,12 @@ func (r *Repository) listResolvableAliases(ctx context.Context, permissions []pe
 			destinationIdClauses = append(destinationIdClauses, "destination_id in (select public_id from target where project_id = any(@target_scope_ids))")
 			args = append(args, sql.Named("target_scope_ids", "{"+strings.Join(directScopeIds, ",")+"}"))
 		}
+		// This condition checks if there are no destinations target Ids and no child scopes,
+		// while also ensuring that the "allDescendants" flag is not set, if so we return no data.
+		// An example scenario of when this can happen is when a role on the global scope grants a user
+		// access to list aliases and read targets, but only within the global scope and its immediate children.
 		if len(destinationIdClauses) == 0 && len(childAllScopes) == 0 {
-			return nil, time.Time{}, errors.New(ctx, errors.InvalidParameter, op, "no target ids or scope ids provided")
+			return nil, time.Time{}, nil
 		}
 		whereClause = fmt.Sprintf("destination_id is not null and (%s)", strings.Join(destinationIdClauses, " or "))
 	}
--- a/internal/alias/target/repository_alias_list_resolvable.go
+++ b/internal/alias/target/repository_alias_list_resolvable.go
@@ -73,6 +73,10 @@ func (r *Repository) listResolvableAliases(ctx context.Context, permissions []pe
 		return nil, time.Time{}, errors.New(ctx, errors.InvalidParameter, op, "missing permissions")
 	}
 
+	// This condition checks if there are no direct IDs, no direct scope IDs, and no child scopes,
+	// while also ensuring that the "allDescendants" flag is not set, and if so returns early.
+	// An example scenario that can cause this is when a role on the global scope grants a user
+	// access to list aliases and read targets, but only within the global scope and its immediate children.
 	directIds, directScopeIds, childAllScopes, allDescendants := splitPermissions(permissions)
 	if !allDescendants && len(directIds) == 0 && len(directScopeIds) == 0 && len(childAllScopes) == 0 {
 		return []*Alias{}, time.Time{}, nil
@@ -151,10 +155,6 @@ func (r *Repository) listResolvableAliasesRefresh(ctx context.Context, updatedAf
 	}
 
 	directIds, directScopeIds, childAllScopes, allDescendants := splitPermissions(permissions)
-
-	if !allDescendants && len(directIds) == 0 && len(directScopeIds) == 0 && len(childAllScopes) == 0 {
-		return []*Alias{}, time.Time{}, nil
-	}
 
 	opts, err := getOpts(opt...)
 	if err != nil {
@@ -236,9 +236,6 @@ func (r *Repository) listRemovedResolvableAliasIds(ctx context.Context, since ti
 	}
 
 	directIds, directScopeIds, childAllScopes, allDescendants := splitPermissions(permissions)
-	if !allDescendants && len(directIds) == 0 && len(directScopeIds) == 0 && len(childAllScopes) == 0 {
-		return []string{}, time.Time{}, nil
-	}
 
 	var args []any
 	var destinationIdClauses []string
--- a/internal/alias/target/repository_alias_list_resolvable.go
+++ b/internal/alias/target/repository_alias_list_resolvable.go
@@ -74,6 +74,9 @@ func (r *Repository) listResolvableAliases(ctx context.Context, permissions []pe
 	}
 
 	directIds, directScopeIds, childAllScopes, allDescendants := splitPermissions(permissions)
+	if !allDescendants && len(directIds) == 0 && len(directScopeIds) == 0 && len(childAllScopes) == 0 {
+		return []*Alias{}, time.Time{}, nil
+	}
 
 	opts, err := getOpts(opt...)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


tmessi (changes requested):
I am not sure if this is the best way to resolve this issue and many of the comments being added here are not correct. Aliases can *only* be in the global scope. However, the ListResolveableAliasesPermissions is looking for permissions for the resource that the alias resolves to. Currently, we only have target aliases, so the `requestedType` is always Target. And *Targets* can only ever be in projects. However, if it is possible that we will add new alias subtypes which would resolve to other resources that might be in any scope.

RyanDerr:
> I am not sure if this is the best way to resolve this issue and many of the comments being added here are not correct. Aliases can _only_ be in the global scope. However, the ListResolveableAliasesPermissions is looking for permissions for the resource that the alias resolves to. Currently, we only have target aliases, so the `requestedType` is always Target. And _Targets_ can only ever be in projects. However, if it is possible that we will add new alias subtypes which would resolve to other resources that might be in any scope.

Ah you're 100% right, I missed that when reading over it briefly in our [docs](https://developer.hashicorp.com/boundary/docs/concepts/aliases#scopes) in that case I need to rethink this approach when resolving aliases for targets that the user does not have permission to and is not within the same scope as the global aliases.

internal/alias/target/repository_alias_list_resolvable.go
```diff
@@ -74,6 +74,9 @@ func (r *Repository) listResolvableAliases(ctx context.Context, permissions []pe
 	}
 
 	directIds, directScopeIds, childAllScopes, allDescendants := splitPermissions(permissions)
+	if !allDescendants && len(directIds) == 0 && len(directScopeIds) == 0 && len(childAllScopes) == 0 {
+		return []*Alias{}, time.Time{}, nil
+	}
```
johanbrandhorst, line 79: Could you add a comment here to explain what the scenario is that causes this? Ditto for the other checks you added.

johanbrandhorst (commented):
I don't feel confident enough in my knowledge of the grants system to review this, sorry.

internal/perms/acl_test.go
```diff
@@ -916,6 +942,169 @@ func TestACL_ListResolvableAliasesPermissions(t *testing.T) {
...
+					grants: []string{
+						"type=target;actions=list",
+						"ids=ttcp_1234567890;actions=read",
+					},
+				},
+				{
+					roleScope:  "global",
+					grantScope: "p_1",
+					grants: []string{
+						"type=target;actions=list",
```
bosorawis, line 1073: I don't think `type=target;actions=list` is necessary here. Could you try removing it and see if the results are the same?

`ListResolvableAliasesPermissions` is only called in one place like `ListResolvableAliasesPermissions(resource.Target, targets.IdActions)` so IMHO we should use `targets.IdActions` as the action set input

internal/perms/acl.go
```diff
@@ -468,21 +468,12 @@ func (a ACL) ListResolvableAliasesPermissions(requestedType resource.Type, actio
 				// so skip it
 				continue
 			}
-		default:
```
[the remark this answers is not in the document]
RyanDerr: This appears to fail in an edge case where a role grants permissions to read/list targets and resolve aliases. The issue arises when resolving aliases for a directly assigned `project` grant scope, while the role also includes the `children` grant scope and the role operates at the `global` scope. In this scenario, the `project` grant scope is skipped because the role scope is `global`, and the `children` scope map contains an entry for `global`.

internal/perms/acl_test.go
```diff
@@ -879,6 +888,23 @@ func TestACL_ListResolvableAliasesPermissions(t *testing.T) {
...
+			},
+		},
+		{
+			name:         "global_with_this_no_descendants",
+			resourceType: resource.Target,
+			actionSet:    action.NewActionSet(action.Read, action.Cancel),
+			aclGrants: []scopeGrant{
+				{
+					roleScope:  "global",
+					grantScope: "global",
```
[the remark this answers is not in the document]
RyanDerr: It appears that global is a valid grant scope from a sample output of the default authenticated user role when
```shell
scopeMap: map[
    global:[
        {global  global * scope map[list:true no-op:true] <nil>} 
        {global  global * auth-method map[list:true authenticate:true] <nil>} 
        {global  global * auth-token map[read:self:true delete:self:true] <nil>} 
        {global  global * scope map[read:true] <nil>} 
        {global  global * auth-token map[list:true] <nil>} 
        {global  global acctpw_0987654321 unknown map[read:true change-password:true] 
...
```
Where each of those entries is a [AclGrant](https://github.com/hashicorp/boundary/blob/3b4538d08cba297ef459e06e6e78b86777a480e0/internal/perms/acl.go#L18) where the 3rd entry (counting blank space) is the grant scope.

internal/perms/acl_test.go
RyanDerr: Yep you're correct that for most of these we don't need the permission to list the targets for alias resolution, I'll go ahead and remove them.

internal/perms/acl_test.go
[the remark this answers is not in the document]
jefferai: They're both valid. "this" is a shortcut to not have to look up the scope id; for global, "global" is the id!

internal/perms/acl.go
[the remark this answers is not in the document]
jefferai: That makes sense -- would it make sense to do something like this?
```go
		case p.RoleScopeId == scope.Global.String() && strings.HasPrefix(grantScopeId, scope.Org.Prefix()):
			if _, ok := childrenScopes[p.RoleScopeId]; ok {
				continue
			}
```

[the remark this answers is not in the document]
RyanDerr: Yep that does capture the edge case in our scenario, thanks for the catch @jefferai!

internal/alias/target/repository_alias_list_resolvable.go
jefferai: Can you explain why this is necessary? Is it purely to shortcut performing the database lookup? Or is it because we will return below with `"no target ids or scope ids provided"` and this isn't being handled well in upstream code?

The other thing I'm curious about -- this may be because of the later optimization work that happened in the DB that I'm not full up on -- but if the user is granted access to list aliases and read targets but within a scope where that's invalid, do we not have that scope here anyways? Possibly we used to but now we optimize that out, and as a result we're triggering the error condition below...?

RyanDerr: Your hunch is right we do reach the `500` error condition where there are no target ids or scope ids provided due to upstream code. This is a results of being passed a non-empty slice of permissions into this function from the service, but the scopes of the permissions provided contain no targets themselves. 

For example if I was to configure a role that grants `ids=*;type=target;actions=list,read,authorize-session` & `ids={{.User.Id}};type=user;actions=list-resolvable-aliases` and specified the scopes `global` and `children` (added image for clarity bellow), it would result in the `listResolvableAliases` function being passed this permission slice. 
```shell
[
  {global  children target list-resolvable-aliases [] false, true}
  {global  global target list-resolvable-aliases [] false, true}
]
```

As a result when we call `splitPermissions` with said slice what is returned is all empty slices and false for `allDescendants`. As a result of this we fall through to the condition [no target ids or scope ids provided](https://github.com/hashicorp/boundary/blob/0307e8a0e563011135acc8052186e450f3251b59/internal/alias/target/repository_alias_list_resolvable.go#L126) where in reality we really should just return no data to the user from those available permissions.
image

RyanDerr: Great point! I'll update the current error response to return nil data without an error and document explaining why this condition is checked, along with an example case where it might occur.

internal/alias/target/repository_alias_list_resolvable.go
```diff
@@ -115,8 +115,12 @@ func (r *Repository) listResolvableAliases(ctx context.Context, permissions []pe
 			destinationIdClauses = append(destinationIdClauses, "destination_id in (select public_id from target where project_id = any(@target_scope_ids))")
 			args = append(args, sql.Named("target_scope_ids", "{"+strings.Join(directScopeIds, ",")+"}"))
 		}
+		// This condition checks if there are no destinations target Ids and no child scopes,
+		// while also ensuring that the "allDescendants" flag is not set, if so we return no data.
+		// An example scenario of when this can happen is when a role on the global scope grants a user
+		// access to list aliases and read targets, but only within the global scope and its immediate children.
 		if len(destinationIdClauses) == 0 && len(childAllScopes) == 0 {
-			return nil, time.Time{}, errors.New(ctx, errors.InvalidParameter, op, "no target ids or scope ids provided")
+			return nil, time.Time{}, nil
```
tmessi, line 123: I am a little surprised that this change didn't result in a test failure. I guess we don't have a test covering this case? Can we add some unit tests so we have that going forward?

RyanDerr: No problem, I can add a simple unit test to get coverage on this.

PR #5676, opened 2025-04-23
Bosorawis domain iam role use new model list role

bosorawis:
new query to perform list role across multiple tables

### DIFF
--- a/internal/iam/repository_role.go
+++ b/internal/iam/repository_role.go
@@ -448,7 +448,6 @@ func (r *Repository) queryRoles(ctx context.Context, whereClause string, args []
 		}
 		transactionTimestamp, err = rd.Now(ctx)
 		return err
-
 	})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/iam/repository_role.go
```diff
@@ -443,8 +448,9 @@ func (r *Repository) queryRoles(ctx context.Context, whereClause string, args []
 		}
 		transactionTimestamp, err = rd.Now(ctx)
 		return err
-	}); err != nil {
-		return nil, time.Time{}, err
+	})
+	if err != nil {
+		return nil, time.Time{}, errors.Wrap(ctx, err, op, errors.WithMsg("failed to query roles"))
```
dkanney, line 453: This `errors.WithMsg` is the same as the one inside the transaction. I think it's safe to just return `err` here?

dkanney (approved):

PR #5679, opened 2025-04-24
Domain: iam: Repository: update list-grant-scope and test setup to use new model

bosorawis:


internal/iam/query.go
```diff
@@ -266,5 +266,120 @@ const (
...
     order by update_time desc, public_id desc
-	`
+`
+	roleGrantsScopeQuery = `
+	with 
+	global_roles (role_id) as (
+		select
+			public_id as role_id
+		from iam_role_global
+		where public_id = any($1)
```
tmessi, line 276: The formatting throughout this entire query is a bit off in places. For example this CTE should be:

```suggestion
        select public_id as role_id
          from iam_role_global
         where public_id = any($1)
```

Can you update it to follow the [sql style guide](https://github.com/hashicorp/boundary/blob/main/sql-style-guide.md#data-manipulation-statements)?

tmessi (changes requested):

internal/iam/query.go
bosorawis: done 👍. I also went removed the leading "tab"

tmessi (approved):

PR #5704, opened 2025-04-29
chore(e2e): Added support for boundary license to be an env for enos

dillanb-hashi:
Created an environment variable that allows you to pass in the boundary license for boundary enterprise without using a file. 
Note that the license in the env will overwrite the license in the file. 

Jira - https://hashicorp.atlassian.net/browse/ICU-16630

Testing - 
Tested a couple different scenarios with the license as an env, license in the file and both envs being set

### DIFF
--- a/enos/ci/hcp-resources/main.tf
+++ b/enos/ci/hcp-resources/main.tf
@@ -43,7 +43,6 @@ locals {
   egress_tag = "egress"
 
   license_path      = abspath(var.boundary_license_path)
-  license           = var.boundary_license
   boundary_zip_path = abspath(var.boundary_zip_path)
 
   cluster_tag     = "boundary_hcp_testing"
--- a/enos/ci/hcp-resources/main.tf
+++ b/enos/ci/hcp-resources/main.tf
@@ -43,6 +43,7 @@ locals {
   egress_tag = "egress"
 
   license_path      = abspath(var.boundary_license_path)
+  license           = var.boundary_license
   boundary_zip_path = abspath(var.boundary_zip_path)
 
   cluster_tag     = "boundary_hcp_testing"
--- a/enos/enos-scenario-e2e-aws-base-with-vault.hcl
+++ b/enos/enos-scenario-e2e-aws-base-with-vault.hcl
@@ -17,7 +17,7 @@ scenario "e2e_aws_base_with_vault" {
     aws_ssh_private_key_path = abspath(var.aws_ssh_private_key_path)
     boundary_install_dir     = abspath(var.boundary_install_dir)
     license_path             = abspath(var.boundary_license_path != null ? var.boundary_license_path : joinpath(path.root, "./support/boundary.hclic"))
-    license                    = var.boundary_license
+    license                  = var.boundary_license
     local_boundary_dir       = var.local_boundary_dir != null ? abspath(var.local_boundary_dir) : null
     build_path = {
       "local" = "/tmp",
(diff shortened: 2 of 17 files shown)


johanbrandhorst (approved):

enos/ci/hcp-resources/main.tf
```diff
@@ -43,6 +43,7 @@ locals {
   egress_tag = "egress"
 
   license_path      = abspath(var.boundary_license_path)
+  license           = var.boundary_license
```
moduli, line 46: The `hcp-resources` directory has a different set of variables that it uses (`ci/hcp-resources/variables.tf`), so technically, `var.boundary_license` doesn't exist in this module unless we add it there.

enos/enos-scenario-e2e-aws-base-with-vault.hcl
```diff
@@ -17,6 +17,7 @@ scenario "e2e_aws_base_with_vault" {
     aws_ssh_private_key_path = abspath(var.aws_ssh_private_key_path)
     boundary_install_dir     = abspath(var.boundary_install_dir)
     license_path             = abspath(var.boundary_license_path != null ? var.boundary_license_path : joinpath(path.root, "./support/boundary.hclic"))
+    license                  = var.boundary_license
```
moduli, line 20: Right now, all this does is create a variable `local.license`, what I'm not sure we need. I *think* what you're attempting to do is to pass this variable into the `read_license` step, since you've modified the `read_license` module

```
step "read_license" {
    skip_step = var.boundary_edition == "oss"
    module    = module.read_license
    variables {
      license_path = local.license_path
      license = var.license
    }
}
```

moduli (changes requested):

enos/enos-scenario-e2e-aws-base-with-vault.hcl
dillanb-hashi: Yeah that's what I was trying to do. I just didn't know the scope of where variables were accessible. I'll make this change

enos/ci/hcp-resources/main.tf
dillanb-hashi: I'm not 100% sure if this is even required to be here.... Going to do some more testing

moduli: FYI: This module is outside the scope of the standard e2e testing. This terraform is used to spin up resources for some long-lived testing.

dillanb-hashi: removed

moduli (commented):
Reminder for later: I'll need to show you how to merge to enterprise and apply these changes to enterprise-specific files.

enos/modules/read_license/main.tf
```diff
@@ -2,7 +2,8 @@
 # SPDX-License-Identifier: BUSL-1.1
 
 variable "license_path" {}
+variable "license" {}
```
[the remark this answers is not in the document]
dillanb-hashi: I think it's fine to not set defaults. If the variable isn't set then it's just null and will fall back to the license_path which does have a default.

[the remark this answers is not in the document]
moduli: Then we may want include descriptions on the variables in this module to explain the difference between them and that you'll always need to pass both of them in whenever you use this module.

enos/modules/read_license/main.tf
moduli, line 1: Just thought of another idea. Let me know your thoughts.

Instead of modifying read_license, we would skip the read_license step in scenarios
```
step "read_boundary_license" {
    skip_step = var.boundary_edition == "oss" || var.license != null     <-- 
    module    = module.read_license

    variables {
      license_path = local.boundary_license_path
      license      = var.boundary_license
    }
  }
```
That would mean we would have to add some more logic to where license is used to choose whether or not to use `var.license` vs. `step.read_license.license`. I do like you can just read `step.read_license.license` in this current implementation.

enos/enos-scenario-e2e-docker-base-plus.hcl
```diff
@@ -68,7 +68,8 @@ scenario "e2e_docker_base_plus" {
     module    = module.read_license
 
     variables {
-      license_path = local.license_path
+      license_path = local.license_path
```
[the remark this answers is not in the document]
dillanb-hashi: fixed

moduli (approved):

PR #5739, opened 2025-05-14
fix(enos): Remove brackets from standalone IPv6 addresses

hugoghx:


moduli (approved):
Thanks for the fix!

johanbrandhorst (dismissed):
Is this something we need to inform users about? If this used to work and now no longer does, we may want to include that in the CHANGELOG and release notes for the next release.

hugoghx:
> Is this something we need to inform users about? If this used to work and now no longer does, we may want to include that in the CHANGELOG and release notes for the next release.

Yeah -- Users will not be able to input addresses like `[2001:ab12:cd34::1]`. The brackets were actually required before but now with the IPv6 work, the above (by itself) is not a valid IPv6 address.

Added a CHANGELOG entry.

PR #5746, opened 2025-05-19
docs: Add HCPb grants issue and fix

Dan-Heath:
Updates the documentation as part of the effort to remediate HCPb default roles and grants before the transparent sessions GA. This PR adds the following:

- Updates the release notes with an important change and known issue
- Updates the section that describes the grants required for transparent sessions
- Updates the CLI `database-init` documentation, which was missing the `-skip-initial-authenticated-user-role-creation` parameter
- Updates the transparent sessions configuration doc to call attention to the Client Agent's required grants and the known issue with HCPb.

- 0.19.0 release notes > [Important changes](https://boundary-8krdcy9t8-hashicorp.vercel.app/boundary/docs/release-notes/v0_19_0#important-changes)
- 0.19.0 release notes > [Known issues and breaking changes](https://boundary-8krdcy9t8-hashicorp.vercel.app/boundary/docs/release-notes/v0_19_0#known-issues-and-breaking-changes)
- Client Agent > [Grants](https://boundary-8krdcy9t8-hashicorp.vercel.app/boundary/docs/api-clients/client-agent#grants)
- database init > [-skip-initial-authenticated-user-role-creation](https://boundary-8krdcy9t8-hashicorp.vercel.app/boundary/docs/commands/database/init#skip-initial-authenticated-user-role-creation)
- [Connect using transparent sessions](https://boundary-8krdcy9t8-hashicorp.vercel.app/boundary/docs/configuration/target-aliases/transparent-sessions)

### DIFF
--- a/website/content/docs/api-clients/client-agent.mdx
+++ b/website/content/docs/api-clients/client-agent.mdx
@@ -73,15 +73,8 @@ type=target;ids=*;actions=authorize-session
 type=auth-token;ids=*;actions=read:self
 ```
 
-Boundary also requires the following grant string for transparent sessions so that you can resolve the aliases associated with targets:
-
-```
-type=target;actions=list,read
-```
-
 HashiCorp highly recommends that you also grant users the permission to list resolvable aliases, as the Client Agent periodically fetches a list of aliases to match incoming DNS requests against.
 Without that permission, every DNS request on the system is sent to the Boundary controller, which can easily overwhelm it.
-For HCP Boundary, it is required that you grant users the permission to list resolvable aliases.
 
 You can use the following grant string to grant the permission to list resolvable aliases:
 
@@ -89,13 +82,18 @@ You can use the following grant string to grant the permission to list resolvabl
 type=user;ids=*;actions=list-resolvable-aliases
 ```
 
-<Note>
+<Warning>
 
 A known issue regarding how grants were previously created in HCP Boundary may cause you to receive a 500 error when you attempt to list resolvable aliases.
+Clusters created before May 1, 2025 may be missing the following grants:
+   - `ids={{.User.Id}};type=user;actions=list-resolvable-aliases`
+   - `ids=*;type=target;actions=list,read`
+
+If your cluster is missing these grants, HashiCorp recommends adding them.
 
 For more information, refer to the [known issue](/boundary/docs/release-notes/v0_19_0#hcp-grants).
 
-</Note>
+</Warning>
 
 ## Configuration
 
--- a/website/content/docs/api-clients/client-agent.mdx
+++ b/website/content/docs/api-clients/client-agent.mdx
@@ -86,8 +86,10 @@ type=user;ids=*;actions=list-resolvable-aliases
 
 A known issue regarding how grants were previously created in HCP Boundary may cause you to receive a 500 error when you attempt to list resolvable aliases.
 Clusters created before April 26, 2025 may be missing the following grants:
-   - `ids={{.User.Id}};type=user;actions=list-resolvable-aliases`
-   - `ids=*;type=target;actions=list,read`
+
+   `ids={{.User.Id}};type=user;actions=list-resolvable-aliases`
+
+   `ids=*;type=target;actions=list,read`
 
 If your cluster is missing these grants, HashiCorp recommends adding them.
 
--- a/website/content/docs/api-clients/client-agent.mdx
+++ b/website/content/docs/api-clients/client-agent.mdx
@@ -73,7 +73,7 @@ type=target;ids=*;actions=authorize-session
 type=auth-token;ids=*;actions=read:self
 ```
 
-HCP Boundary also requires the following grant string for transparent sessions:
+Boundary also requires the following grant string for transparent sessions so that you can resolve the aliases associated with targets:
 
 ```
 type=target;actions=list,read
@@ -91,7 +91,7 @@ type=user;ids=*;actions=list-resolvable-aliases
 
 <Note>
 
-A known issue regarding how grants were previously created in HCP Boundary may cause you to receive a 500 error when you attempt to connnect to an alias.
+A known issue regarding how grants were previously created in HCP Boundary may cause you to receive a 500 error when you attempt to list resolvable aliases.
 
 For more information, refer to the [known issue](/boundary/docs/release-notes/v0_19_0#hcp-grants).
 
--- a/website/content/docs/api-clients/client-agent.mdx
+++ b/website/content/docs/api-clients/client-agent.mdx
@@ -85,7 +85,7 @@ type=user;ids=*;actions=list-resolvable-aliases
 <Warning>
 
 A known issue regarding how grants were previously created in HCP Boundary may cause you to receive a 500 error when you attempt to list resolvable aliases.
-Clusters created before May 1, 2025 may be missing the following grants:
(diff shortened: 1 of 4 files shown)


website/content/docs/commands/database/init.mdx
```diff
@@ -115,6 +115,9 @@ The default value is `false`.
 - `-skip-host-resources-creation` - If not set, skips the creation of host resources as part of the initialization, inlcuding host catalog, host set, and hosts.
 The default value is `false`.
 
+- `-skip-initial-authenticated-user-role-creation` - If not set, skips the creation of the role that is associated with `u_auth`.
```
RyanDerr, line 118: If the flag is set then any of the resources are with said flag are not created. As it reads currently if the flag is not set then we skip the creation of said role. 

Additionally it may be more clear exactly what this mean via stating:
`If set, skips the creation of the role and grants provisioned by default for all authenticated users.`

website/content/docs/api-clients/client-agent.mdx
```diff
@@ -74,14 +73,30 @@ type=target;ids=*;actions=authorize-session
 type=auth-token;ids=*;actions=read:self
 ```
 
+HCP Boundary also requires the following grant string for transparent sessions:
```
RyanDerr, line 76: I don't believe this is specifically tied to HCP Boundary, but is required as to list resolvable aliases you must also have access to read/list targets to resolve said alias associated with it. Though I will defer to @johanbrandhorst if this is not explicitly the case.

website/content/docs/api-clients/client-agent.mdx
```diff
@@ -74,14 +73,30 @@ type=target;ids=*;actions=authorize-session
...
+
 You can use the following grant string to grant the permission to list resolvable aliases:
 
 ```
 type=user;ids=*;actions=list-resolvable-aliases
 ```
 
+<Note>
+
+A known issue regarding how grants were previously created in HCP Boundary may cause you to receive a 500 error when you attempt to connnect to an alias.
```
RyanDerr, line 94: I believe this was an issue when you attempted to List Resolvable Aliases (such as with the API or natively via caching for the UI) not when you attempt to connect with said alias.

website/content/docs/api-clients/client-agent.mdx
johanbrandhorst: Yeah this applies to all Boundary deployments, as Ryan says.

Dan-Heath: Updated to apply to all Boundary deployments. I also included a better explanation of why it is required based on Ryan's comment.

website/content/docs/api-clients/client-agent.mdx
```diff
@@ -74,14 +73,30 @@ type=target;ids=*;actions=authorize-session
 type=auth-token;ids=*;actions=read:self
 ```
 
+Boundary also requires the following grant string for transparent sessions so that you can resolve the aliases associated with targets:
+
+```
+type=target;actions=list,read
+```
+
```
johanbrandhorst, line 81: Wait, sorry, where did this requirement come from? I don't think we need this, do we? Was this also a grant missing from the CLI bug? CC @RyanDerr. Also, should this be

```
ids=*;type=target;actions=list,read
```

website/content/docs/api-clients/client-agent.mdx
```diff
@@ -74,14 +73,30 @@ type=target;ids=*;actions=authorize-session
...
 HashiCorp highly recommends that you also grant users the permission to list resolvable aliases, as the Client Agent periodically fetches a list of aliases to match incoming DNS requests against.
 Without that permission, every DNS request on the system is sent to the Boundary controller, which can easily overwhelm it.
+For HCP Boundary, it is required that you grant users the permission to list resolvable aliases.
+
```
johanbrandhorst, line 85: I don't think we should word it this way, since new clusters now have this permission (as we intended back in 0.16.0). I think we should include this in the `Note` below instead, with something like:

```
An issue prevented HCP Boundary clusters created before May 1st 2025 to have this grant added automatically. If your cluster does not have this grant, we strongly recommend adding it.
```

website/content/docs/api-clients/client-agent.mdx
johanbrandhorst: If this is another grant that's usually automatically added, I think we should just have a note about it for older clusters, not imply that all HCPb customers need to do this.

johanbrandhorst: Thanks for confirming Ryan, that makes sense to me :)

website/content/docs/api-clients/client-agent.mdx
RyanDerr: Looks like it was actually slightly earlier, the last commit to resolve the issue [ccc495b](https://github.com/hashicorp/cloud-boundary-service/commit/ccc495b5317b61d3871c3de2c8a4b41c4e2f66d9) was merged on April 25th and was also [released](https://github.com/hashicorp/cloud-boundary-service/actions/runs/14669876288) that day. So I think we should say any clusters before April 26th as a result.

saptaparninandy:
@Dan-Heath can we add the link to the "Grants" page to the Requirement section [here](https://developer.hashicorp.com/boundary/docs/configuration/target-aliases/transparent-sessions) for transparent sessions - maybe add as a note - "Note: If your cluster was created on or before May 2nd, 2025, we recommend you to review the grants required for Transparent Sessions to be able to use this feature". (Please reword as needed and confirm the date with @RyanDerr )

website/content/docs/api-clients/client-agent.mdx
Dan-Heath: Thanks for the clarification! I removed these lines and included that grant string in the Note that Johan suggested below.

website/content/docs/api-clients/client-agent.mdx
Dan-Heath: Thanks Johan! That makes sense. I included a version of your Note with the 2 relevant grant strings that HCPb users should check for. Also, I made the note into a Warning since it could prevent people from using transparent sessions.

Thanks for the clarification on the date! I updated that to April 26 throughout.

Dan-Heath:
> @Dan-Heath can we add the link to the "Grants" page to the Requirement section [here](https://developer.hashicorp.com/boundary/docs/configuration/target-aliases/transparent-sessions) for transparent sessions - maybe add as a note - "Note: If your cluster was created on or before May 2nd, 2025, we recommend you to review the grants required for Transparent Sessions to be able to use this feature". (Please reword as needed and confirm the date with @RyanDerr )

Sure thing. I updated the [Requirements](https://boundary-8krdcy9t8-hashicorp.vercel.app/boundary/docs/configuration/target-aliases/transparent-sessions#requirements) section in that topic. I also added a link to the required Grants section from the Tip in the [Configure targets](https://boundary-8krdcy9t8-hashicorp.vercel.app/boundary/docs/configuration/target-aliases/transparent-sessions#configure-targets) section.

Dan-Heath:
Thanks for your feedback all! I think I addressed all the concerns with this HCPb grants update. Please double check me.  I updated the preview links above, if you want to see how the finished product would look.

johanbrandhorst (approved):
Thanks, this is perfect!

yhyakuna (approved):
LGTM

PR #5748, opened 2025-05-20
Change reqScope parameter type from `Scope` to `string`

dkanney:
- Changing the type of the `reqScope` parameter from `Scope` to `string`
- Simplify `grantsForUserGlobalResourcesQuery` & its repo function. Now querying directly into `perms.GrantTuple`

### DIFF
--- a/internal/iam/repository_role_grant.go
+++ b/internal/iam/repository_role_grant.go
@@ -569,22 +569,20 @@ func (r *Repository) grantsForUserGlobalResources(
 		sql.Named("resources", pq.Array(resources)),
 	)
 
-	var grants []grantsForUserResults
+	var grants []perms.GrantTuple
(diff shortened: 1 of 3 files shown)


internal/iam/repository_role_grant.go
```diff
@@ -593,39 +591,23 @@ func (r *Repository) grantsForUserGlobalResources(
...
+	return grants, nil
 }
 
 // grantsForUserOrgResources returns the grants for the user for resources that can
 // only be org scoped.
 func (r *Repository) grantsForUserOrgResources(
 	ctx context.Context,
-	userId string,
+	userId,
+	scopeId string,
```
bosorawis, line 602: nitpick: calling this request scope ID not to be confused with resource scope ID or role scope ID 
```suggestion
	reqScopeId string,
```

bosorawis (approved):

PR #5750, opened 2025-05-20
feat: grants: add resource type parameters to GrantsForUser

bosorawis:
Add `resource.Type` parameter to GrantsForUser call chain


internal/daemon/controller/auth/auth_test.go
```diff
@@ -243,7 +244,7 @@ func TestVerify_AuditEvent(t *testing.T) {
 			ctx := NewVerifierContext(ctx, iamRepoFn, tokenRepoFn, serversRepoFn, testKms, &requestInfo)
 
 			_ = os.WriteFile(eventConfig.AuditEvents.Name(), nil, 0o666) // clean out audit events from previous calls
-			_ = Verify(ctx, tt.opt...)
+			_ = Verify(ctx, resource.Scope, tt.opt...)
```
bosorawis, line 247: This test tests for audit events so I don't think the resource type matters here as it only cares that audit event information is filled and published. Choosing resource.Scope because the test request is `httptest.NewRequest("GET", "http://127.0.0.1/v1/scopes/o_1", nil)`

internal/daemon/controller/auth/auth_test.go
```diff
@@ -285,6 +286,7 @@ func TestGrantsHash(t *testing.T) {
 	}
 
 	o, _ := iam.TestScopes(t, iamRepo)
+	resourceType := resource.Scope
```
bosorawis, line 289: This test is testing `GrantHash` function so which should stay the same as long as the resource is the same

internal/iam/query.go
```diff
@@ -580,6 +580,230 @@ const (
            canonical_grant;
     `
 
+	grantsForUserProjectResourcesGlobalScopeQuery = resourceRoleGrantsForUsers + `,
```
tmessi, line 583: Is this meant to be a part of this PR or a separate one?

dkanney: This isn't meant to be in this PR; I accidentally pushed a commit from #5783 to `llb-normalized-grants` 😓 

We need to drop the offending commit, ca18f954e00b250ddcff059eb79ef1537546a9d4

bosorawis: fixed

internal/daemon/controller/auth/option.go
```diff
@@ -40,6 +39,12 @@ func getDefaultOptions() options {
 	return options{}
 }
 
+func WithRecursive() Option {
+	return func(o *options) {
+		o.withRecursive = true
+	}
+}
```
dkanney: Although this makes sense in its current state, the pattern we use for the other `bool`s suggests that this function should accept an argument and use that as the value for `o.withRecursive`:

```suggestion

func WithRecursive(recursive bool) Option {
	return func(o *options) {
		o.withRecursive = recursive
	}
}
```

internal/daemon/controller/auth/options_test.go
```diff
@@ -20,16 +19,20 @@ func Test_GetOpts(t *testing.T) {
 	withKms := new(kms.Kms)
 	res := new(perms.Resource)
 
+	// test default values
+	defaultOpts := getDefaultOptions()
+	assert.False(t, defaultOpts.withRecursive)
+
```
dkanney, line 25: Should we also assert the default values of the other struct members here? e.g.

```suggestion
assert.False(t, defaultOpts.withRecoveryTokenNotAllowed)
assert.False(t, defaultOpts.withAnonymousUserNotAllowed)
assert.Equal(t, defaultOpts.withAction, action.Unknown)
```

dkanney (changes requested):

PR #5755, opened 2025-05-21
test(e2e): Test for scopes permission defect

dillanb-hashi:
This test covers the defect https://hashicorp.atlassian.net/browse/ICU-12561

The test modifies some of the default grants and then adds a grant to a user. That user should then only see their one assigned project when they run the command
boundary scopes list --recursive

Command to run just this test
go test github.com/hashicorp/boundary/testing/internal/e2e/tests/base -v -run 'TestScopePermissions'

Test output: 
=== RUN   TestScopePermissions
    scope.go:85: Created Org Id: o_QdBvNI4LV2
    scope.go:127: Created Project Id: p_T5JCyqEu6q
    scope.go:127: Created Project Id: p_fo5CYYKHiz
    account.go:79: Created Account: acctpw_QhFJYTn5jZ
    user.go:65: Created User: u_bVsLbSJ67a
    role.go:64: Created Role: r_0mswbSD9px in scope o_QdBvNI4LV2
--- PASS: TestScopePermissions (4.70s)
PASS
ok      github.com/hashicorp/boundary/testing/internal/e2e/tests/base   5.342s

### DIFF
--- a/testing/internal/e2e/tests/base/scope_permissions_test.go
+++ b/testing/internal/e2e/tests/base/scope_permissions_test.go
@@ -17,6 +17,8 @@ import (
 )
 
 // TestScopePermissions tests an issue with listing scopes
+// When a user is only given permission to only list single project in a scope,
+// the user should only see that project and no error should be displayed
 func TestScopePermissions(t *testing.T) {
 	e2e.MaybeSkipTest(t)
 	bc, err := boundary.LoadConfig()
@@ -39,7 +41,7 @@ func TestScopePermissions(t *testing.T) {
 	_, err = boundary.CreateProjectCli(t, ctx, orgId)
 	require.NoError(t, err)
 
-	//Create new account
+	// Create new account
 	const testAccountName = "test-account"
 	accountId, acctPassword, err := boundary.CreateAccountCli(t, ctx, bc.AuthMethodId, testAccountName)
 	require.NoError(t, err)
@@ -51,7 +53,7 @@ func TestScopePermissions(t *testing.T) {
 		require.NoError(t, output.Err, string(output.Stderr))
 	})
 
-	//create new user
+	// Create new user
 	userId, err := boundary.CreateUserCli(t, ctx, "global")
 	require.NoError(t, err)
 	t.Cleanup(func() {
@@ -64,7 +66,7 @@ func TestScopePermissions(t *testing.T) {
 	err = boundary.SetAccountToUserCli(t, ctx, userId, accountId)
 	require.NoError(t, err)
 
-	//Find Ids of the roles to be modified
+	// Find Ids of the roles to be modified
 	output := e2e.RunCommand(ctx, "boundary", e2e.WithArgs("roles", "list", "-format", "json"))
 	var rolesListResult roles.RoleListResult
 	err = json.Unmarshal(output.Stdout, &rolesListResult)
@@ -89,7 +91,7 @@ func TestScopePermissions(t *testing.T) {
 		}
 	}
 
-	//Modify login and default grants role
+	// Modify login and default grants role
 	output = e2e.RunCommand(ctx, "boundary",
 		e2e.WithArgs("roles", "remove-grants", "-id", loginAndDefaultGrantsRoleId, "-grant", "ids=*;type=scope;actions=list,no-op"))
 	require.NoError(t, output.Err, string(output.Stderr))
@@ -97,18 +99,7 @@ func TestScopePermissions(t *testing.T) {
 		e2e.WithArgs("roles", "add-grants", "-id", loginAndDefaultGrantsRoleId, "-grant", "ids=*;type=scope;actions=list"))
 	require.NoError(t, output.Err, string(output.Stderr))
 
-	t.Cleanup(func() {
-		boundary.AuthenticateAdminCli(t, ctx)
-		output := e2e.RunCommand(ctx, "boundary",
-			e2e.WithArgs("roles", "remove-grants", "-id", loginAndDefaultGrantsRoleId, "-grant", "ids=*;type=scope;actions=list"))
-		require.NoError(t, output.Err, string(output.Stderr))
-		output = e2e.RunCommand(ctx, "boundary",
-			e2e.WithArgs("roles", "add-grants", "-id", loginAndDefaultGrantsRoleId, "-grant", "ids=*;type=scope;actions=list,no-op"),
-		)
-		require.NoError(t, output.Err, string(output.Stderr))
-	})
-
-	//Modify login grants role
+	// Modify login grants role
 	output = e2e.RunCommand(ctx, "boundary",
 		e2e.WithArgs("roles", "remove-grants", "-id", loginGrantsRoleId, "-grant", "ids=*;type=scope;actions=list,no-op"))
 	require.NoError(t, output.Err, string(output.Stderr))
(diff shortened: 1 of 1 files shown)


moduli (changes requested):

moduli (changes requested):
Note the commit format should be `test(e2e)`

testing/internal/e2e/tests/base/scope_permissions_test.go
```diff
@@ -0,0 +1,165 @@
...
+	t.Cleanup(func() {
+		boundary.AuthenticateAdminCli(t, ctx)
+		output := e2e.RunCommand(ctx, "boundary",
+			e2e.WithArgs("roles", "add-grants", "-id", authenticatedUserGrantsRoleId, "-grant", "ids=*;type=scope;actions=read"),
+		)
+		require.NoError(t, output.Err, string(output.Stderr))
+	})
+	//Create new project
+	projectId, err := boundary.CreateProjectCli(t, ctx, "global")
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
dillanb-hashi: The ticket says the new project should be made in the default org. 

I tried your solution with new org and two projects and was getting the project from the default org in the recursive list. Think I'll refactor it to utilize the default org which will solve a couple of the other comments as well.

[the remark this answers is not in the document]
moduli: I kind of like the idea of not using the default org so that the test is creating all of the data that it needs.

We can adjust the assertion to only look at scopes (i.e. projects) in the created org. We could do that by either
- `boundary scopes list --recursive`, only look at scopes that have a "scope id" of the org
- OR `boundary scopes list -scope-id <org>

What do you think?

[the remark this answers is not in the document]
moduli: Note: The ticket says to use the default org as that's what already exists and made it easier for repro.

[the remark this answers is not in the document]
dillanb-hashi: I agree with using a created org. My only reservation to doing, boundary scopes list -scope-id, is I'm not sure if that actually would replicate the defect.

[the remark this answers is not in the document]
dillanb-hashi: >Note: The ticket says to use the default org as that's what already exists and made it easier for repro.

Wasn't sure if there's anything special about the default org or not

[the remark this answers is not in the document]
moduli: > Wasn't sure if there's anything special about the default org or not
Nope. It's just a pre-created org.

testing/internal/e2e/tests/base/scope_permissions_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+	t.Cleanup(func() {
+		boundary.AuthenticateAdminCli(t, ctx)
+		output := e2e.RunCommand(ctx, "boundary",
+			e2e.WithArgs("roles", "add-grants", "-id", authenticatedUserGrantsRoleId, "-grant", "ids=*;type=scope;actions=read"),
+		)
+		require.NoError(t, output.Err, string(output.Stderr))
+	})
+
+	// Create role in global scope
```
moduli, line 134:
```suggestion
	// Create role in org and add ability to list and read a project in the org
```

testing/internal/e2e/tests/base/scope_permissions_test.go
```diff
@@ -0,0 +1,165 @@
...
+	output = e2e.RunCommand(ctx, "boundary", e2e.WithArgs("roles", "list", "-format", "json", "-scope-id", orgId))
+	err = json.Unmarshal(output.Stdout, &rolesListResult)
+	require.NoError(t, err)
+	for _, item := range rolesListResult.Items {
+		if item.Name == "Login and Default Grants" {
+			loginAndDefaultGrantsRoleId = item.Id
+		}
+	}
+
+	// Modify login and default grants role
```
moduli, line 94: Maybe we should add context here as to what we're doing with these modifications.
- In the org, prevent the user from being able to list all scopes

testing/internal/e2e/tests/base/scope_permissions_test.go
```diff
@@ -0,0 +1,165 @@
...
+	output = e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs("roles", "remove-grants", "-id", loginAndDefaultGrantsRoleId, "-grant", "ids=*;type=scope;actions=list,no-op"))
+	require.NoError(t, output.Err, string(output.Stderr))
+	output = e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs("roles", "add-grants", "-id", loginAndDefaultGrantsRoleId, "-grant", "ids=*;type=scope;actions=list"))
+	require.NoError(t, output.Err, string(output.Stderr))
+
+	// Modify login grants role
```
moduli, line 102: Maybe we should add context here as to what we're doing with these modifications.
- In global, prevent the user from being able to list all scopes

testing/internal/e2e/tests/base/scope_permissions_test.go
```diff
@@ -0,0 +1,165 @@
...
+		output := e2e.RunCommand(ctx, "boundary",
+			e2e.WithArgs("roles", "remove-grants", "-id", loginGrantsRoleId, "-grant", "ids=*;type=scope;actions=list"))
+		require.NoError(t, output.Err, string(output.Stderr))
+		output = e2e.RunCommand(ctx, "boundary",
+			e2e.WithArgs("roles", "add-grants", "-id", loginGrantsRoleId, "-grant", "ids=*;type=scope;actions=list,no-op"),
+		)
+		require.NoError(t, output.Err, string(output.Stderr))
+	})
+
+	// Modify authenticated user role
```
moduli, line 121: Maybe we should add context here as to what we're doing with these modifications.
- In global, prevent the user from being able to read all scopes

testing/internal/e2e/tests/base/scope_permissions_test.go
```diff
@@ -0,0 +1,165 @@
...
+	var scopesListResult scopes.ScopeListResult
+	err = json.Unmarshal(output.Stdout, &scopesListResult)
+	require.NoError(t, err)
+	var listedIds []string
+	for _, item := range scopesListResult.Items {
+		listedIds = append(listedIds, item.Id)
+	}
+	require.Equal(t, 1, len(listedIds))
+	require.Equal(t, projectId, listedIds[0])
+}
```
moduli, line 165: What do you think about adding a check to confirm that we can read project1, but not project 2?

moduli (commented):
Looking good!

testing/internal/e2e/tests/base/scope_permissions_test.go
dillanb-hashi: It's not specific to the original defect but shouldn't be too hard to add so I don't see why not.

testing/internal/e2e/tests/base/scope_permissions_test.go
```diff
@@ -0,0 +1,174 @@
...
+		output := e2e.RunCommand(ctx, "boundary",
+			e2e.WithArgs("roles", "remove-grants", "-id", loginGrantsRoleId, "-grant", "ids=*;type=scope;actions=list"))
+		require.NoError(t, output.Err, string(output.Stderr))
+		output = e2e.RunCommand(ctx, "boundary",
+			e2e.WithArgs("roles", "add-grants", "-id", loginGrantsRoleId, "-grant", "ids=*;type=scope;actions=list,no-op"),
+		)
+		require.NoError(t, output.Err, string(output.Stderr))
+	})
+
+	// In global, prevent the user from being able to read all specs
```
moduli, line 121: I assume this is supposed to refer to scopes?
```suggestion
	// In global, prevent the user from being able to read all scopes
```

dillanb-hashi: Correct!

moduli (approved):

PR #5781, opened 2025-05-22
dependabot: round two, group all Go modules

johanbrandhorst:
Apparently there was a `directories` key that I missed. It should hopefully group all changes into a single PR.


moduli (approved):
here we go

PR #5786, opened 2025-05-23
chore(e2e): Set auth_storage_path to a non-temp directory

moduli:
This PR updates the e2e test infra so that workers set their `auth_storage_path` to a directory that is more permanent than `/tmp/boundary`. This was causing issues, especially in the long-lived cluster, where the worker would lose registration to the cluster since the `/tmp/boundary` could get blown away (for example, when restarting/upgrading the worker). 

This PR makes the following changes
- renames the existing `create-audit-log-dir.sh` script to `create-dir.sh` to make it more generic
- sets one of the worker configs so that `auth_storage_path=/var/lib/boundary/worker`


johanbrandhorst (approved):
Thanks!

PR #5807, opened 2025-05-29
feat: grantsForUser for Global/Org/Project resources

dkanney:


internal/iam/query.go
```diff
@@ -763,6 +763,232 @@ const (
...
+           canonical_grant as grant
+      from all_roles
+  group by role_id,
+           role_scope_id,
+           role_parent_scope_id,
+           grant_scope,
+           canonical_grant;
+    `
+
+	grantsForUserGlobalOrOrgOrProjectResourcesOrgScopeRecursiveQuery = resourceRoleGrantsForUsers + `,
```
bosorawis: I think you could drop `Or` from the resource type.
Also, could you add godoc comment explaining the queries?
```suggestion
	// grantsForUserGlobalOrgProjectResourcesOrgScopeRecursiveQuery is  used to look up grants applied for 		 
        // resources that can exist in ... and when request scope is an org 
	grantsForUserGlobalOrgProjectResourcesOrgScopeRecursiveQuery = resourceRoleGrantsForUsers + `,
```

internal/iam/query.go
```diff
@@ -763,6 +763,232 @@ const (
...
+             iam_role_org.scope_id             as role_scope_id,
+             'global'                          as role_parent_scope_id,
+             iam_role_org.scope_id             as grant_scope,
+             roles_with_grants.canonical_grant as canonical_grant
+        from iam_role_org
+        join roles_with_grants
+          on roles_with_grants.role_id = iam_role_org.public_id
+       where iam_role_org.grant_this_role_scope
+         and iam_role_org.scope_id = @request_scope_id
+    ),
```
bosorawis, line 943: Do you also need `children` for global/org/project resources when reqScope is an org?

internal/iam/repository_role_grant_test.go
```diff
@@ -3868,7 +3886,537 @@ func TestGrantsForUserGlobalAndOrgResources(t *testing.T) {
...
+		{
+			name: "return grants for group resource at org2 request scope",
+			input: testInput{
+				userId:     user.PublicId,
+				reqScopeId: org2.PublicId,
+				resource:   resource.Group,
+			},
+			output: []perms.GrantTuple{
+				{
+					RoleId:            org2RoleThisAndChildren.PublicId,
```
bosorawis, line 4157: I think this result set is missing `children` from `org2RoleThisAndChildren`?

bosorawis (changes requested):
`grantsForUserGlobalOrOrgOrProjectResourcesGlobalScopeRecursiveQuery` looks good but I think `grantsForUserGlobalOrOrgOrProjectResourcesOrgScopeRecursiveQuery` is missing a `children` grants lookup. 
 
Could you also add some godoc comments on the queries to explain what they're targeting? IMHO the names are long enough that things could be unclear without additional documentation

internal/iam/query.go
dkanney: Dropped the `Or`s to shorten all these really long names 😓 

Also added clearer comments to everyting

internal/iam/query.go
dkanney: I wasn't 100% sure about this. I figured all recursive queries request everyting at the current scope + below the current scope

However, it definitely makes sense that Global `children` applies to all org scopes.. Global `descendent` would apply too in this case, right?

internal/iam/repository_role_grant_test.go
dkanney: Yeah thinking about this a lil more, it makes sense. Recursively listing from an org scope returns its projects, so `children` grant_scope would apply in this case too

internal/iam/query.go
dkanney: Yeah! If we're recursively listing projects under an org, then the `children` grant_scope would also apply.

Added a CTE: `org_roles_children_grant_scope`

internal/iam/query.go
```diff
@@ -305,6 +306,7 @@ const (
            canonical_grant;
     `
 
+	// grantsForUserOrgResourcesQuery gets a user's grants for resources only applicable to org scopes.
```
bosorawis, line 309: question: IIRC there no resources that only exists in orgs. Is this query specifically for resources that can exist in org and when request scope is an org?

bosorawis (approved):
I have 1 minor question the changes LGTM!

internal/iam/query.go
dkanney: Yes -- I realize now this comment should reflect that >.<

Updated!

PR #5815, opened 2025-05-30
Migrate ui-related scripts to use pnpm instead of yarn

hashicc:


johanbrandhorst (dismissed):
Thanks!

enos/modules/build_local/templates/build.sh
```diff
@@ -7,6 +7,7 @@ set -eux -o pipefail
 env
 
 # Requirements
+npm install --global pnpm || true
 npm install --global yarn || true
```
moduli: Would installing `yarn` not be needed anymore here?

hashicc: That's a good question and something I wanted to check in with you about. I originally had `yarn` swapped for `pnpm` but wanted to discuss if there's a situation where backwards compatibility is needed: where the boundary repo with enos would need to support a boundary-ui commit that is using either `pnpm` or `yarn`? Along with [this line that is now using pnpm](https://github.com/hashicorp/boundary/pull/5815/files#diff-46be1b2db7ec8ff2793934b332564e650a2016fc75dc1eaf7b9933b4a342c88bR241). This was a concern for local builds and so [this check](https://github.com/hashicorp/boundary/pull/5815/files#diff-d58d7eb91d367fdef556d17c890f5f7161ac29402aef6112670d784ab928a1f4R79-R103) was added

moduli: Mmmmm. I *think* the main thing is that we should only merge these changes once we update the UI commits that contain the pnpm changes? If I'm building something older, I would be using that version of boundary (with the older scripts) along with an older UI commit?

Am I thinking about that correctly?

ZedLi: That's what I was originally thinking but I also wasn't sure if the workflows for enos actually pull from `main` ever (as it looked like it was [here](https://github.com/hashicorp/boundary/blob/93f3dca8125401a15bb66f2ab3c05f0356876b40/.github/workflows/enos-run.yml#L244) since I didn't see a version). Do you know if all the enos related builds will still be using the boundary-ui `VERSION`?

moduli: I need to get rid of the boundary-ui references in the current `enos-run` workflows since it's not actually used. We'll have a separate workflow for the Admin UI tests when we get to it.

Yea, the enos related builds will build from `VERSION`.

hashicc: Thanks for the clarification @moduli and @ZedLi. Trying to support both pnpm and yarn for local builds doesn't sound like a use case worth the complexity. Like you've both mentioned, to keep things simple and consistent this PR will be held off until there's a UI commit change in `VERSION` that needs to support pnpm. I've updated the PR title, body and changed this back to a draft pr

PR #5822, opened 2025-06-05
Refactor GrantsForUser subqueries to use aggregate functions

dkanney:
This change intends to reduce the amount of rows retrieved by only querying each table once per CTE and using aggregate functions (`array_agg`). This should reduce the amount of rows returned from the DB in cases when there are many grants on a role:

Before:

```sql
-- returns one row per grant

    select role_id,
           role_scope_id,
           role_parent_scope_id,
           grant_scope,
           canonical_grant as grant
      from global_and_org_roles
  group by role_id,
           role_scope_id,
           role_parent_scope_id,
           grant_scope,
           canonical_grant;
```

After:

```sql
-- returns one row per role, with an array of distinct grants for each

    select role_id,
           role_scope_id,
           role_parent_scope_id,
           grant_scope,
           grant_this_role_scope,
           array_agg(distinct(individual_grant_scope)) as individual_grant_scopes,
           array_agg(distinct(canonical_grant))        as canonical_grants
      from global_and_org_roles
  group by role_id,
           role_scope_id,
           role_parent_scope_id,
           grant_scope,
           grant_this_role_scope;
```

### DIFF
--- a/internal/iam/repository_role_grant.go
+++ b/internal/iam/repository_role_grant.go
@@ -781,20 +781,22 @@ func (r *Repository) grantsForUserProjectResources(
 		sql.Named("request_scope_id", reqScopeId),
 	)
 
-	var grants []perms.GrantTuple
+	var grants []grantsForUserResults
 	rows, err := r.reader.Query(ctx, grantsForUserProjectResourcesQuery, args)
 	if err != nil {
 		return nil, errors.Wrap(ctx, err, op)
 	}
 	defer rows.Close()
 	for rows.Next() {
-		var g perms.GrantTuple
+		var g grantsForUserResults
 		if err := rows.Scan(
-			&g.RoleId,
-			&g.RoleScopeId,
-			&g.RoleParentScopeId,
-			&g.GrantScopeId,
-			&g.Grant,
+			&g.roleId,
+			&g.roleScopeId,
+			&g.roleParentScopeId,
+			&g.grantScope,
+			&g.grantThisRoleScope,
+			pq.Array(&g.individualGrantScopes),
+			pq.Array(&g.canonicalGrants),
 		); err != nil {
 			return nil, errors.Wrap(ctx, err, op)
 		}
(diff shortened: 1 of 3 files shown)


internal/iam/repository_role_grant.go
```diff
@@ -589,7 +591,49 @@ func (r *Repository) grantsForUserGlobalResources(
 	if err := rows.Err(); err != nil {
 		return nil, errors.Wrap(ctx, err, op)
 	}
-	return grants, nil
+	ret := make(perms.GrantTuples, 0, len(grants)*3)
+	for _, grant := range grants {
+		// If the role has grant scope set to "individual" then we need to
+		// iterate over the individual grant scopes and create a grant tuple
+		// for each one.
+		if grant.grantScope == globals.GrantScopeIndividual && len(grant.individualGrantScopes) > 0 {
```
bosorawis: No need to check for `grant.grantScope == globals.GrantScopeIndividual` here. If there's any scope in `individualGrantScopes`, they should be included. For example, a global role can have both `grant_scope = children` and individual projects.

internal/iam/repository_role_grant.go
```diff
@@ -589,7 +591,49 @@ func (r *Repository) grantsForUserGlobalResources(
...
+					gt := perms.GrantTuple{
+						RoleId:            grant.roleId,
+						RoleScopeId:       grant.roleScopeId,
+						RoleParentScopeId: grant.roleParentScopeId,
+						GrantScopeId:      grantScope,
+						Grant:             canonicalGrant,
+					}
+
+					if gt.GrantScopeId == globals.GrantScopeThis || gt.GrantScopeId == "" {
+						gt.GrantScopeId = grant.roleScopeId
```
bosorawis: can we only check for `grantThisRoleScope = true`? In the new model. you shouldn't be able to add `this` grant scope as an individual grant scope so I don't think there should be a scenario where `gt.GrantScopeId == ""` unless `gt.GrantScopeId == globals.GrantScopeThis`

dkanney: I mirrored the logic in this function with the others in anticipation of using the same function for all query results (so we just switch on the query, a la `resolveQuery(resourceType, requestScope, isRecursive) (string, error)`) in #5823

For the global query we really could just query results directly into `GrantTuples` - no need to check `grant.grantScope == globals.GrantScopeIndividual` or `grantThisRoleScope = true`

bosorawis (approved):

PR #5823, opened 2025-06-06
Support previous grants use-cases 

bosorawis:
This PR resolves broken tests found after the refactor from all packages 
- remove test cases for cases that are no longer supported
  - adding/setting duplicate grant scopes will now not return an error but will perform deduplication internally
  - switch some grants tests to use resources that's allowed in their scope-under-test
- fix issues where newly created roles have version = 2 (now new role version is 1) and fix tests 
- automatically convert role's scope ID to `this` when setting/adding/removing a role's grant_scope
- fix `authorized_action` and `authorized_collection_action` not returning for sub-type by supporting multiple resource types in permission code


internal/iam/query.go
```diff
@@ -699,6 +605,44 @@ const (
...
+             iam_role_org.scope_id             as role_scope_id,
+             'global'                          as role_parent_scope_id,
+             iam_role_org.scope_id             as grant_scope,
+             roles_with_grants.canonical_grant as canonical_grant
+        from iam_role_org
+        join roles_with_grants
+          on roles_with_grants.role_id = iam_role_org.public_id
+       where iam_role_org.grant_this_role_scope
+         and iam_role_org.scope_id = @request_scope_id
+    ),
```
dkanney: I don't think CTE belongs in this query -- It's querying for resources at the Org scope, but Project resources can't live at the Org scope

internal/iam/query.go
```diff
@@ -699,6 +605,44 @@ const (
...
+             roles_with_grants.canonical_grant as canonical_grant
+        from iam_role_global
+        join roles_with_grants
+          on roles_with_grants.role_id = iam_role_global.public_id
+        join iam_role_global_individual_project_grant_scope individual
+          on individual.role_id = iam_role_global.public_id
+        join iam_scope_project
+          on iam_scope_project.scope_id = individual.scope_id 
+       where iam_scope_project.parent_id = @request_scope_id -- only look for projects under the requested org
+    ),
```
dkanney: Wow, didn't think of this case! Nice catch

internal/iam/query.go
bosorawis: good catch!

internal/iam/repository_role_grant_test.go
```diff
@@ -2154,12 +2266,149 @@ func TestGrantsForUserGlobalOrgProjectResourcesWithRecursive(t *testing.T) {
...
 	TestRoleGrant(t, conn, proj2RoleThis.PublicId, "ids=*;type=group;actions=set-members")
 
-	// Add users to created roles
+	// Add user to created roles (some containing the list action)
 	for _, role := range roles {
 		_, err := repo.AddPrincipalRoles(ctx, role.PublicId, role.Version, []string{user.PublicId})
 		require.NoError(t, err)
 	}
 
+	// If there's a list permission any where in the scope tree, the user should be able to perform recursive list at the global scope
```
dkanney: Need to test the case where a user has grant scopes that don't contain the list action anywhere in the scope tree

dkanney:
https://github.com/hashicorp/boundary/blob/408dfcd554556a31afc1e5e2b57350fc4f42b745/internal/iam/repository_role_grant.go#L721

This function is recursive too so it needs to be removed & replaced with `grantsForUserRecursive()`

internal/daemon/controller/handlers/authmethods/authmethod_service.go
```diff
@@ -78,6 +78,11 @@ var (
 		resource.Account:      accounts.CollectionActions,
 		resource.ManagedGroup: managed_groups.CollectionActions,
 	}
+
+	additionalResourceGrants = []resource.Type{
+		resource.Account,
+		resource.ManagedGroup,
+	}
```
tmessi, line 85: Rather than having these lists defined in each service handler, it might be nice to define these as a method on `resource.Type` so that the grouping of parent and child types are all defined with the types. We already have something related to this with the [`Parent()` and `HasChildTypes()` methods](https://github.com/hashicorp/boundary/blob/f91978747b3266bf7d989d1e04682e374bbb76e3/internal/types/resource/resource.go#L144-L166)

With that, we wouldn't need the `WithFetchAdditionalResourceGrants` option, since we could just call that method on the passed in resource when fetching grants.

tmessi (changes requested):

PR #5836, opened 2025-06-11
Resolve GrantsForUser queries via resolveQuery()

dkanney:
For each grantsForUser repo function, the logic to map from DB object -> `perms.GrantTuple` was very similar.

- This PR reduces code by removing those functions & moves the mapping logic into `GrantsForUser`
- Created a new function, `resolveQuery()`, to determine which query to use based on the request scope & the resources' applicable scopes

### DIFF
--- a/internal/iam/repository_role_grant.go
+++ b/internal/iam/repository_role_grant.go
@@ -8,7 +8,6 @@ import (
 	"database/sql"
 	"fmt"
 	"slices"
-	"sort"
 	"strings"
 
 	"github.com/hashicorp/boundary/globals"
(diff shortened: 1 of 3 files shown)


mikemountain (approved):

internal/iam/repository_role_grant.go
```diff
@@ -644,334 +599,84 @@ func (r *Repository) grantsForUserGlobalResources(
...
-					RoleParentScopeId: grant.roleParentScopeId,
-					GrantScopeId:      individualGrantScope,
-					Grant:             canonicalGrant,
-				}
-				ret = append(ret, gt)
-			}
-		}
-	}
-	return ret, nil
+func (m *MultiGrantTuple) TestStableSort() {
```
bosorawis: nitpick: is this option and method still used? can we remove it?

dkanney: Yep! We can remove this

mikemountain (approved):

PR #5839, opened 2025-06-12
chore(e2e): Added build tag for enabling pprof tracing

dillanb-hashi:
Added a build tag "pprof" and a make command "make pprof" that enables tracing and exposes the pprof endpoint. 

Original tracing setup found here 
https://github.com/hashicorp/boundary/blob/main/testing/TRACING.md

This change basically removes the need to add the code snippet to a location in the code and instead puts it behind a build tag.

### DIFF
--- a/internal/cmd/base/pprof_off.go
+++ b/internal/cmd/base/pprof_off.go
@@ -11,6 +11,4 @@ import (
 )
 
 func StartPprof(_ context.Context) {
-	go func() {
-	}()
 }
--- a/internal/cmd/base/pprof_on.go
+++ b/internal/cmd/base/pprof_on.go
@@ -8,14 +8,21 @@ package base
 
 import (
 	"context"
-	"log"
+	"errors"
 	"net/http"
+
+	"github.com/hashicorp/boundary/internal/event"
 
 	_ "net/http/pprof"
 )
 
-func StartPprof(_ context.Context) {
+func StartPprof(ctx context.Context) {
+	const op = "base.StartPprof"
 	go func() {
-		log.Println(http.ListenAndServe("localhost:6060", nil))
+		const addr = "localhost:6060"
+		event.WriteSysEvent(ctx, op, "starting pprof HTTP server", "addr", addr)
+		if err := http.ListenAndServe(addr, nil); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			event.WriteSysEvent(ctx, op, "failed to serve pprof HTTP server", "error", err.Error())
+		}
 	}()
 }
--- a/internal/cmd/commands/dev/dev.go
+++ b/internal/cmd/commands/dev/dev.go
@@ -709,7 +709,6 @@ func (c *Command) Run(args []string) int {
 	}
 
 	base.StartMemProfiler(c.Context)
-	base.StartPprof(c.Context)
 
 	if err := c.SetupEventing(
 		c.Context,
@@ -722,6 +721,8 @@ func (c *Command) Run(args []string) int {
 		c.UI.Error(err.Error())
 		return base.CommandCliError
 	}
+
+	base.StartPprof(c.Context)
 
 	if c.flagRecoveryKey != "" {
 		c.Config.DevRecoveryKey = c.flagRecoveryKey
--- a/testing/TRACING.md
+++ b/testing/TRACING.md
@@ -1,26 +1,7 @@
 # Tracing in Boundary
 
 Boundary includes a small number of runtime tracing user regions, which can be used to see where Boundary spends its time during execution.
-To create a trace, we first need to expose the pprof endpoint. It is disabled by default. Exposing the pprof endpoint is as simple as importing the correct package and starting a HTTP server anywhere in Boundary:
-
-```go
-package anything
-
-import (
-    "log"
-    "net/http"
-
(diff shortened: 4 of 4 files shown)


internal/cmd/base/pprof_off.go
```diff
@@ -0,0 +1,17 @@
...
+package base
+
+import (
+	"context"
+)
+
+func StartPproff(_ context.Context) {
+	go func() {
+	}()
+
```
moduli: Note that there are some linting errors from this check: https://github.com/hashicorp/boundary/actions/runs/15618466139/job/43997106560?pr=5839

internal/cmd/base/pprof_off.go
```diff
@@ -0,0 +1,17 @@
...
+//go:build !pprof
+// +build !pprof
+
+package base
+
+import (
+	"context"
+)
+
+func StartPproff(_ context.Context) {
```
moduli: ?
```suggestion
func StartPprof(_ context.Context) {
```

Makefile
```diff
@@ -66,6 +66,12 @@ install: build
 install-no-plugins: export SKIP_PLUGIN_BUILD=1
 install-no-plugins: install
 
+.PHONY: build-pprof
+build-pprof: BUILD_TAGS+=pprof
+build-pprof:
+	@echo "==> Building Boundary with memory pprof enabled"
+	@CGO_ENABLED=$(CGO_ENABLED) BUILD_TAGS='$(BUILD_TAGS)' sh -c "'$(CURDIR)/scripts/build.sh'"
+
 .PHONY: build-memprof
```
moduli, line 75: Do we know what this `memprofiler` does? Is there ever a case we would want both `pprof` and `memprofiler` or are those separate things?

dillanb-hashi: Not sure, but if you did run into that case you can just set both build tags and just do a regular make build. The build-pprof and build mem profiler just run a build with adding in the appropriate tag.

internal/cmd/base/pprof_off.go
dillanb-hashi: fixed

internal/cmd/base/pprof_on.go
```diff
@@ -0,0 +1,22 @@
...
+	"context"
+	"log"
+
+	"net/http"
+	_ "net/http/pprof"
+)
+
+func StartPproff(_ context.Context) {
+	go func() {
+		log.Println(http.ListenAndServe("localhost:6060", nil))
```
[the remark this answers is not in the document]
dillanb-hashi: Should only be used locally so we can ignore this security issue.

Makefile
johanbrandhorst: I don't know of any recent use of this build tag, but I agree that we can just keep them separate.

internal/cmd/base/pprof_off.go
```diff
@@ -0,0 +1,16 @@
...
+
+package base
+
+import (
+	"context"
+)
+
+func StartPprof(_ context.Context) {
+	go func() {
+	}()
```
johanbrandhorst: We shouldn't need a goroutine like this, an empty function body is fine.
```suggestion
```

internal/cmd/base/pprof_on.go
```diff
@@ -0,0 +1,21 @@
...
+	"log"
+	"net/http"
+
+	_ "net/http/pprof"
+)
+
+func StartPprof(_ context.Context) {
+	go func() {
+		log.Println(http.ListenAndServe("localhost:6060", nil))
+	}()
```
johanbrandhorst: I think now that we're adding this "properly", we should use the event system to log things instead of `log.Println`. If we move the call of this function below `c.SetupEventing`, we should be able to use the `event` package to log things here instead:
```suggestion
	const op = "base.StartPprof"
	go func() {
		const addr = "localhost:6060"
		event.WriteSysEvent(ctx, op, "starting pprof HTTP server, "addr", addr)
		if err := http.ListenAndServe(addr, nil); err != nil && !errors.Is(err, http.ErrServerClosed) {
			event.WriteSysEvent(ctx, op, "failed to serve pprof HTTP server", "error", err.Error())
		}
	}()
```

What do you think?

internal/cmd/commands/dev/dev.go
```diff
@@ -709,6 +709,7 @@ func (c *Command) Run(args []string) int {
 	}
 
 	base.StartMemProfiler(c.Context)
+	base.StartPprof(c.Context)
```
johanbrandhorst: Move this below `SetupEventing` to use eventing inside call

testing/TRACING.md
```diff
@@ -20,6 +20,12 @@ import (
 ...
 ```
```
johanbrandhorst: Lets just remove this now that we have a build tag

internal/cmd/base/pprof_on.go
dillanb-hashi: Agreed, changed to this.

internal/cmd/commands/dev/dev.go
dillanb-hashi: Done.

johanbrandhorst (approved):
Thanks!

moduli (approved):

PR #5843, opened 2025-06-13
Add options to create default and admin roles

mikemountain:
When creating a new scope, by default two roles are created each with some default grants. This can be disabled by using the `-skip-admin-role-creation` or `-skip-admin-role-creation` flag/option to skip this creation. We want to replace that functionality with `-create-default-role` and `-create-admin-role` which will default to false


internal/cmd/commands/database/init.go
```diff
@@ -123,13 +123,25 @@ func (c *InitCommand) Flags() *base.FlagSets {
 	f.BoolVar(&base.BoolVar{
 		Name:   flagSkipInitialLoginRoleName,
 		Target: &c.initFlags.flagSkipInitialLoginRoleCreation,
-		Usage:  "If set, a role providing necessary grants for logging in will not be created as part of initialization. If set, the recovery KMS will be needed to perform any actions.",
+		Usage:  "Deprecated: If set, a role providing necessary grants for logging in will not be created as part of initialization. If set, the recovery KMS will be needed to perform any actions.",
```
bosorawis, line 127: question: do we mark which version this will be deprecated in?

bosorawis (changes requested):
It looks like these flags are not used by the domain layer (`internal/iam/`)

can you also add the flag support to these spots?
- https://github.com/hashicorp/boundary/blob/a5f71a648a32250a0aae2a57f6a7e6cd782e5134/internal/iam/repository_scope.go#L164
- https://github.com/hashicorp/boundary/blob/a5f71a648a32250a0aae2a57f6a7e6cd782e5134/internal/iam/repository_scope.go#L100

internal/cmd/commands/database/init.go
mikemountain: I don't think so, here's an example elsewhere https://github.com/hashicorp/boundary/blob/main/internal/cmd/commands/targetscmd/tcp_funcs.go#L118

internal/iam/repository_scope.go
```diff
@@ -93,6 +93,53 @@ func (r *Repository) CreateScope(ctx context.Context, s *Scope, userId string, o
 	var adminRole Resource
 	var adminRoleRaw any
 	switch {
+	case opts.withCreateAdminRole:
```
bosorawis: I think this clause has to take priority. 
```
userId == globals.AnonymousUserId,
userId == globals.AnyAuthenticatedUserId,
userId == globals.RecoveryUserId,
```
perhaps
```
case userId == "",
	userId == globals.AnonymousUserId,
	userId == globals.AnyAuthenticatedUserId,
	userId == globals.RecoveryUserId,
	opts.withSkipAdminRoleCreation:
...

case opts.withCreateAdminRole || !opts.withSkipAdminRoleCreation
...
```

internal/cmd/commands/database/flags.go
```diff
@@ -39,30 +47,120 @@ type initFlags struct {
...
+//
+// The initial authenticated user role is created if:
+// - The flag for the creation of the initial authenticated user role (`-create-initial-authenticated-user-role`) is provided.
+//
+// Returns:
+// - bool: True if the initial authenticated user role should be created, false otherwise.
+// - string: A reason string explaining why the role is created, or an empty string if not created.
+func (f *initFlags) CreateInitialAuthenticatedUserRole() (bool, string) {
+	if f.flagCreateInitialAuthenticatedUserRole {
+		return true, reasonFlagWasSet(flagCreateInitialAuthenticatedUserRoleName)
```
dkanney: What's the plan for these two functions? It doesn't look like they're used anywhere

internal/cmd/base/dev.go
```diff
@@ -160,10 +160,7 @@ func (b *Server) CreateDevDatabase(ctx context.Context, opt ...Option) error {
 		return nil
 	}
 
-	if _, _, err := b.CreateInitialScopes(ctx, WithIamOptions(
-		iam.WithSkipAdminRoleCreation(true),
-		iam.WithSkipDefaultRoleCreation(true),
-	)); err != nil {
```
bosorawis: This changes the `dev` database to not creating the default role when creating scope. I think we should add this back since we aren't changing the default behavior yet

PR #5863, opened 2025-06-18
chore: Update nodeenrollment

hugoghx:


johanbrandhorst (approved):

hugoghx:
@johanbrandhorst Sorry, Go was refusing to build Boundary if I didn't update the version to 1.24.4 (possibly because the nodeenrollment dep wants 1.24.4?)

api/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/hashicorp/boundary/api
 
-go 1.24.0
```
johanbrandhorst: I don't think we should do this, because it will force all users of this module to update to go 1.24.4. It's bad enough that we require 1.24.0, I think. Can we revert this change in all `go.mod` files except the top level one?

johanbrandhorst (changes requested):

api/go.mod
hugoghx: That's a good point! We need the `.go-version` one too right as that's what we use to build?

johanbrandhorst: Yeah we may as well update that since we're about to make a release.

hugoghx: Ty, fixed!

johanbrandhorst (approved):

johanbrandhorst:
No idea what happened to the enos jobs 🤔. Retry?

hugoghx:
> No idea what happened to the enos jobs 🤔. Retry?

Yeah, I'm seeing enos failures in other PRs too, so I don't think this is related to these changes. Triggered a retry now to see if things have settled.

Edit: Yep it was a flake, they just passed

Issue #5003, opened 2024-08-07
Roles are not working as expected


achetronic:
**Describe the bug**
A user is inside two different groups, for example: `admins` and `developers`. 

In the global scope the group 'admins' is linked to the GlobalAdmin role, whose permissions are:
```
ids=*;type=*;actions=*
```

And inside a project, the group `developers` which is linked to a role that has the following permissions:
```
ids=ttcp_a0r6J8Djju;type=target;actions=authorize-session
ids=*;type=target;actions=read,list
```
after setting these permissions, the mentioned user can not see everything, but only one target on that project. So I see the project permissions are overriding the global ones 

I have read permissions are additive in Boundary, is this the intended behavior? How can I manage this use case? I read about no-op but I don't know if this can help my use case

What am I doing wrong?

**Expected behavior**
What I expected is: _"if you have the permissions in the global scope, having less in the project should not limit you, as you already have them"_

What I wanted to do is trying to give permissions to use some specific targets to all the company, but some additional ones (over those general ones) to other teams

**Additional context**

moduli:
When defining grants in a `global` scope, by default, those grants only apply to items within the `global` scope and not any scopes further down the chain. 

If you want a grant to apply to additional scopes, you can go here to select your desired settings. 
![Screenshot 2024-08-09 at 12 33 44 PM](https://github.com/user-attachments/assets/0421f0b3-fcc6-456a-a28b-690351a89bac)

More information [here](https://developer.hashicorp.com/boundary/docs/common-workflows/manage-roles#add-grant-scopes)

achetronic:
Hello @moduli, thanks for your response. We are putting a lot of effort crafting our tools around Boundary

Which version are you using? I don't have those options in  v0.15.4. I have reviewed the changelog and I can not find those in messages (btw, very interesting things in latest versions about the targets aliases)

moduli:
Ah. Support for this feature in the Admin UI came in `v0.17.0`, however, you are able to utilize this feature via the CLI in `0.15.x`. 
```
boundary roles add-grant-scopes ...
```
https://developer.hashicorp.com/boundary/docs/v0.15.x/commands/roles/add-grant-scopes

achetronic:
Oh, thanks! I can update my deployment to give you some feedback about that functionality. Apart from updating the image, is it needed to execute database migrations? (I assume so, but asking just in case) :)

moduli:
[...]

PR #5877, opened 2025-06-23
bug: fix list target permissions resource ID overriding all resource grants

bosorawis:
Fix issue #5003 

When a resource ID is specified for the same scope, list permissions handling prioritizes the specific resource ID rather than the `ids=*` which grants permissions to more resources causing the query to incorrectly filter out resources that the user should have permissions to take actions on.

- **fix list target not handling all resources permissions**
- **add test**

### DIFF
--- a/internal/daemon/controller/handlers/targets/tcp/grants_test.go
+++ b/internal/daemon/controller/handlers/targets/tcp/grants_test.go
@@ -164,6 +164,36 @@ func TestGrants_ReadActions(t *testing.T) {
 				userFunc: iam.TestUserDirectGrantsFunc(t, conn, kmsCache, globals.GlobalPrefix, password.TestAuthMethodWithAccount, []iam.TestRoleGrantsRequest{}),
 				wantErr:  handlers.ForbiddenError(),
 			},
+			{
+				name: "iss 5003 less permissive grants should not override more permissive ones",
+				input: &pbs.ListTargetsRequest{
+					ScopeId:   globals.GlobalPrefix,
+					Recursive: true,
+				},
+				userFunc: iam.TestUserGroupGrantsFunc(t, conn, kmsCache, globals.GlobalPrefix, password.TestAuthMethodWithAccount, []iam.TestRoleGrantsRequest{
+					{
+						RoleScopeId: globals.GlobalPrefix,
+						Grants:      []string{"ids=*;type=*;actions=*;output_fields=id,authorized_actions,scope_id,address"},
+						GrantScopes: []string{globals.GrantScopeDescendants},
+					},
+					{
+						RoleScopeId: globals.GlobalPrefix,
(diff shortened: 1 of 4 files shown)


internal/target/repository.go
```diff
@@ -390,7 +390,7 @@ func (r *Repository) listPermissionWhereClauses() ([]string, []any) {
 		clauses = append(clauses, fmt.Sprintf("project_id = @project_id_%d", inClauseCnt))
 		args = append(args, sql.Named(fmt.Sprintf("project_id_%d", inClauseCnt), p.GrantScopeId))
 
-		if len(p.ResourceIds) > 0 {
+		if len(p.ResourceIds) > 0 && !p.All {
```
moduli, line 393: This logic seems to make sense to me?

Just curious, does this need to be applied to any other resource? Do they all have a similar kind of method?

internal/daemon/controller/handlers/targets/tcp/grants_test.go
```diff
@@ -164,6 +164,36 @@ func TestGrants_ReadActions(t *testing.T) {
...
+				input: &pbs.ListTargetsRequest{
+					ScopeId:   globals.GlobalPrefix,
+					Recursive: true,
+				},
+				userFunc: iam.TestUserGroupGrantsFunc(t, conn, kmsCache, globals.GlobalPrefix, password.TestAuthMethodWithAccount, []iam.TestRoleGrantsRequest{
+					{
+						RoleScopeId: globals.GlobalPrefix,
+						Grants:      []string{"ids=*;type=*;actions=*;output_fields=id,authorized_actions,scope_id,address"},
+						GrantScopes: []string{globals.GrantScopeDescendants},
+					},
```
moduli: Curious, what were the thoughts behind the test here? Would we want separate tests here for a different setup?

Idea:
```
[]string{"ids=*;type=*;actions=*;output_fields=id,authorized_actions,scope_id,address"},
[]string{fmt.Sprintf("ids=%s;actions=authorize-session;output_fields=id,authorized_actions,scope_id,address", target1.GetPublicId())},
```
and
```
[]string{fmt.Sprintf("ids=%s;actions=authorize-session;output_fields=id,authorized_actions,scope_id,address", target1.GetPublicId())},
[]string{"ids=*;type=target;actions=read,list;output_fields=id,authorized_actions,scope_id,address"},
```

internal/target/repository.go
bosorawis: good catch! found one more in `session` package so I fixed that one too

internal/daemon/controller/handlers/targets/tcp/grants_test.go
bosorawis: break this the test to two separate cases as suggested

moduli (approved):
I'm by no means an expert on this, but seems like this works?

PR #5878, opened 2025-06-23
Manual oss main into release/0.19.x merge

dkanney:


.github/workflows/enos-run.yml
```diff
@@ -202,7 +199,7 @@ jobs:
         run: |
           wget https://releases.hashicorp.com/vault/1.12.2/vault_1.12.2_linux_amd64.zip -O /tmp/test-deps/vault.zip
       - name: Install Vault CLI
-        if: contains(matrix.filter, 'vault') || contains(matrix.filter, 'e2e_docker') || matrix.filter == 'e2e_database' || matrix.filter == 'e2e_ui_aws builder:crt'
+        if: contains(matrix.filter, 'vault') || contains(matrix.filter, 'e2e_docker') || matrix.filter == 'e2e_database'
```
dkanney, line 202: @moduli It looks like we may want to keep the removed bit here, ` || matrix.filter == 'e2e_ui_aws builder:crt'`

.github/dependabot.yml
```diff
@@ -9,8 +9,42 @@ updates:
...
+      # Group all Go module updates into one PR for version updates
+      # and one for security updates.
+      go:
+        patterns:
+          - "*"
+        applies-to: "version-updates"
+      go-security:
+        patterns:
+          - "*"
+        applies-to: "security-updates"
```
johanbrandhorst, line 50: I expect this might trigger some dependabot updates down the road, but that's probably a good thing.

.github/workflows/enos-run.yml
moduli: We can remove it. It's not used at all.

PR #5891, opened 2025-07-01
feat: add MySQL command support PORT

bgajjala8:
## Description
Porting over a PR from an open source contribution to trigger testing suite.

## PCI review checklist

- [X] I have documented a clear reason for, and description of, the change I am making.
- [X] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.
- [X] If applicable, I've documented the impact of any changes to security controls.


bgajjala8:
@johanbrandhorst
It looks like two tests were added. I've addressed the linting issues for the tests in the `base_plus `directory and removed the one from `base` dir.

I've set the milestone for this PR to 0.20.x. Does that work?

testing/internal/e2e/tests/base_plus/target_tcp_connect_mysql_test.go
```diff
@@ -0,0 +1,125 @@
...
+	_, err = f.Write([]byte("SELECT DATABASE();\n")) // show current database
+	require.NoError(t, err)
+	_, err = f.Write([]byte("EXIT;\n")) // exit mysql session
+	require.NoError(t, err)
+	_, err = f.Write([]byte{4}) // EOT (End of Transmission - marks end of file stream)
+	require.NoError(t, err)
+
+	var buf bytes.Buffer
+	_, _ = io.Copy(&buf, f)
+	require.Contains(t, buf.String(), "mysql>", "Session did not return expected MySQL prompt")
```
johanbrandhorst: Looks like this is [failing today](https://github.com/hashicorp/boundary/actions/runs/16011194421/job/45216827194?pr=5891) in CI:
```
ERROR 2013 (HY000): Lost connection to server at 'handshake: reading initial communication packet', system error: 11\r\n\a
```
I was able to get this working locally on my machine, so we will need to try to debug why it's not working in CI. I can't tell what's going on based on the logs, so we should probably add some more debug logging to this test and the rerun CI to see if we can see what's going on. Since we're having connectivity issues, it would be nice if we could add some more logging around the attempt to establish a connection on line 50. I would try and run some of these commands through that and print them to see if that works. If so, it might be a problem with the Boundary mysql command. To speed up iteration, I would recommend trying to run it locally first and then push it up to CI. Let me know if you need help figuring out how to run the e2e tests locally.

johanbrandhorst: Also, I think we deleted the wrong file, this test should be in the `base` folder, not the `base_plus` folder.

bgajjala8: Thanks for catching that, I got it mixed up 🙃 - updated

internal/cmd/commands/connect/mysql.go
```diff
@@ -0,0 +1,120 @@
...
+
+		switch {
+		case username != "":
+			args = append(args, "-u", username)
+		case c.flagUsername != "":
+			args = append(args, "-u", c.flagUsername)
+		}
+	}
+	return
+}
```
bgajjala8, line 120:
```suggestion
}

```
New line

bgajjala8, line 120:
```suggestion
}

```

testing/internal/e2e/tests/base/target_tcp_connect_mysql_test.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+	}
+
+	// Get network alias from container name (remove leading "/")
+	networkAlias = strings.TrimPrefix(container.Name, "/")
+
+	mysqlPort = "3306"
+
+	return mysqlUser, mysqlPassword, mysqlDb, networkAlias, mysqlPort
+}
```
bgajjala8:
```suggestion
}

```
new line

internal/cmd/commands/connect/mysql.go
bgajjala8, line 120:
```suggestion
}

```
new line

bgajjala8:
@johanbrandhorst  
I went ahead and updated tests to work with our E2E suite. I also cleaned up how we gather the mysql creds in the tests, so they're less explicit now. Would love a re-review on the PR!

testing/internal/e2e/infra/docker.go
```diff
@@ -303,3 +303,44 @@ func StartOpenSshServer(t testing.TB, pool *dockertest.Pool, network *dockertest
...
+		ExposedPorts: []string{"3306/tcp"},
+		Name:         networkAlias,
+		Networks:     []*dockertest.Network{network},
+	})
+	require.NoError(t, err)
+
+	return &Container{
+		Resource:     resource,
+		UriLocalhost: "http://localhost:3306",
+		UriNetwork:   "http://e2emysql:3306",
```
johanbrandhorst: It's neat that we don't have to provide credentials to this function anymore, but this makes the container harder to use (the user has to guess to look at the environment variables). It looks like MySQL [supports using URI connection strings](https://dev.mysql.com/doc/refman/8.4/en/connecting-using-uri-or-key-value-pairs.html#connecting-using-uri) in the same way as postgres does. What do you think about putting the credentials in these returned strings instead and extract them out in the caller?
```suggestion
		UriLocalhost: fmt.Sprintf("mysql://%s:%s@localhost:3306/%s", mysqlUser, mysqlPassword, mysqlDb),
		UriNetwork:   fmt.Sprintf("mysql://%s:%s@%s:3306/%s", mysqlUser, mysqlPassword, networkAlias, mysqlDb),
```

In the test we can extract this using `url.Parse`:

```go
u, err := url.Parse(container.UriLocalhost)
if err != nil {
...
}
user := u.User.Username()
pw := u.User.Password()
```

This way it behaves just like the postgres function.

johanbrandhorst (changes requested):
Thanks for the update and for getting the test to pass, great job! I think we can make it just a little by nicer still :)

testing/internal/e2e/infra/docker.go
bgajjala8: Yep, this is quite nice! Cool Lib 💯

testing/internal/e2e/tests/base/target_tcp_connect_mysql_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+	user, hostname, port, db := u.User.Username(), u.Hostname(), u.Port(), u.Path[1:]
+	pw, pwSet := u.User.Password()
+	t.Logf("MySQL info: user=%s, db=%s, host=%s, port=%s, password-set:%t",
+		user, db, hostname, port, pwSet)
+
+	// Wait for MySQL to be ready
+	err = pool.Retry(func() error {
+		return exec.CommandContext(ctx, "docker", "exec", hostname,
+			"mysql", "-u"+user, "-p"+pw, "-e", "SELECT 1").Run()
```
johanbrandhorst, line 55: Sorry to be unclear, I meant for us to use the URI here, not the parsed values. Can you see if this works
```suggestion
			"mysql", "--uri", c.UriNetwork, "-e", "SELECT 1").Run()
```

johanbrandhorst (approved):
Just one small suggestion. Make sure to use a merge commit when merging this so the original author gets commit credit (not a squash).

testing/internal/e2e/tests/base/target_tcp_connect_mysql_test.go
RyanDerr: IIRC I don't think the mysql cli supports connecting via URI strings.

johanbrandhorst: 🤔 [The page I linked](https://dev.mysql.com/doc/refman/8.4/en/connecting-using-uri-or-key-value-pairs.html#connecting-using-uri) in my previous comment says that "The MySQL shell" uses `--uri` for these. I don't know if that means the CLI, if it doesn't then sure lets leave it this way 😂.

bgajjala8: Yeah, MySQL Shell is a different CLI than what we're using. If we wanted to go that route, we'd need to use something like `mysqlsh --uri mysql://e2eboundary:e2eboundary@localhost:3306/e2eboundarydb --sql -e "SELECT 1"`. I could try adjusting the tests to invoke that CLI, but we'd still need the traditional `mysql` client for `boundary connect mysql`. Given that, I think I'll leave it as is for now. @johanbrandhorst, would that be okay, or would you prefer I try to make those changes?

johanbrandhorst: Nah, of course, I thought it was the same thing, leave it as it is!

PR #5894, opened 2025-07-02
UpdateControllerStatus:Update Hotfix

RyanDerr:
## Description
- Update `UpdateControllerStatus` to handle empty string descriptions. When reading from the `server_controller` repo, `NULL` descriptions are cast to empty strings. To prevent violating the `wt_description_too_short` constraint in the `wt_description` domain, ensure that empty string descriptions are converted to `nil` before being passed to the appropriate prepared query [updateController](https://github.com/hashicorp/boundary/blob/d4b34b1f50f5ee1dd3a8477123cf6a18284b7f38/internal/server/query.go#L227).
- Add two additional unit tests to account for this scenario.

### DIFF
--- a/internal/server/repository_controller.go
+++ b/internal/server/repository_controller.go
@@ -95,12 +95,18 @@ func (r *Repository) UpdateControllerStatus(ctx context.Context, controller *Con
 		db.ExpBackoff{},
 		func(reader db.Reader, w db.Writer) error {
 			var err error
-			rowsUpdated, err = w.Exec(ctx, updateController,
-				[]any{
-					sql.Named("controller_address", controller.Address),
-					sql.Named("controller_description", controller.Description),
-					sql.Named("controller_private_id", controller.PrivateId),
-				})
+			params := []any{
+				sql.Named("controller_private_id", controller.PrivateId),
+				sql.Named("controller_address", controller.Address),
+			}
+
+			if controller.Description != "" {
+				params = append(params, sql.Named("controller_description", controller.Description))
+			} else {
+				params = append(params, sql.Named("controller_description", nil))
+			}
+
+			rowsUpdated, err = w.Exec(ctx, updateController, params)
 			switch {
 			case err != nil:
 				return errors.Wrap(ctx, err, op+":Update")
--- a/internal/server/repository_controller_test.go
+++ b/internal/server/repository_controller_test.go
@@ -184,6 +184,28 @@ func TestRepository_UpdateControllerStatus(t *testing.T) {
 				require.Equal(t, 1, c)
 			},
 		},
+		"valid-ipv4-controller-remove-description": {
+			originalController: NewController("test", WithAddress("127.0.0.1"), WithDescription("short name description")),
+			updatedController:  NewController("test", WithAddress("127.0.0.2")),
+			wantCount:          1,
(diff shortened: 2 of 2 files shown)


internal/server/repository_controller.go
```diff
@@ -95,12 +95,18 @@ func (r *Repository) UpdateControllerStatus(ctx context.Context, controller *Con
...
-					sql.Named("controller_address", controller.Address),
-					sql.Named("controller_description", controller.Description),
-					sql.Named("controller_private_id", controller.PrivateId),
-				})
+			params := []any{
+				sql.Named("controller_private_id", controller.PrivateId),
+				sql.Named("controller_address", controller.Address),
+			}
+
+			if controller.Description != "" {
```
bosorawis, line 103: Do we have to handle whitespace description? 
```suggestion
			if strings.TrimSpace(controller.Description) != "" {
```

Or we could create a function `func trimToNil(s string) any` to convert empty string to `nil`

internal/server/repository_controller.go
```diff
@@ -95,12 +95,18 @@ func (r *Repository) UpdateControllerStatus(ctx context.Context, controller *Con
...
+				params = append(params, sql.Named("controller_description", controller.Description))
+			} else {
+				params = append(params, sql.Named("controller_description", nil))
+			}
```
johanbrandhorst, line 107: Can we not just not add it as a parameter in this case?

internal/server/repository_controller.go
johanbrandhorst: I don't think we should mutate the user set description unless we have to

internal/server/repository_controller.go
johanbrandhorst: Ah I suppose we want to remove it if a user has set it to ""

johanbrandhorst (approved):

internal/server/repository_controller.go
RyanDerr: Right if it was changed to `NULL` which would cast to `''` we should update it such that it's treated as `NULL` in the DB.

internal/server/repository_controller.go
RyanDerr: Right I think it comes down to if a user provides a whitespace description, should we accept it and treat it as `NULL` or have the behavior today where it would get bubbled up as a constraint violation error. Sounds like the latter is desired.

johanbrandhorst: Is the behavior today that a description like `" "` triggers a constraint violation? I thought it was just `""` that would fail. Sure lets keep the behavior consistent with what we had, but I don't see why `" "` shouldn't be valid unless it already is invalid.

RyanDerr: Today it should be invalid, as the constraint [wt_description_too_short](https://github.com/hashicorp/boundary/blob/87f702e6af55e04f1538c28d6a2440882bfc4ab3/internal/db/schema/migrations/oss/postgres/1/02_domains.up.sql#L55) does a trim and only after performs an evaluation on its length. So in this case `" "` would trigger an exception to be thrown.

johanbrandhorst: Fair enough, thanks for explaining!

PR #5899, opened 2025-07-04
 fix(bug): authorized_collection_actions not returning from host-catalog

bosorawis:
## Description

When using 2 different roles to grant permissions to `host-catalog` and its child resources `host` and `host-sets`, collection_authorized_actions do not get populated


johanbrandhorst (approved):
Nice find!

PR #5904, opened 2025-07-08
Allow rollbacks of migrations if hooks fail

mikemountain:
## Description
If, during a migration, a hook fails to work correctly, the migration will currently not be rolled back which can cause Boundary to end up in a state where we do not support the current db version. This fix will rollback any transactions if a hook errors while running.

## PCI review checklist

### DIFF
--- a/internal/db/schema/manager.go
+++ b/internal/db/schema/manager.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"context"
 	"database/sql"
+	stderrors "errors"
 	"fmt"
 	"io"
 	"sync"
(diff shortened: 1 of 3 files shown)


dkanney (approved):

dkanney:
Changes look good to me - approved 👍

internal/db/schema/manager.go
```diff
@@ -244,57 +247,68 @@ func (b *Manager) runMigrations(ctx context.Context, p *provider.Provider) ([]Re
 	const op = "schema.(Manager).runMigrations"
 
 	var logEntries []RepairLog
-	var err error
```
johanbrandhorst, line 247: Sorry, I should've left a comment on this PR. The better way to do this is to use a named variable error return:
```go
func (b *Manager) runMigrations(ctx context.Context, p *provider.Provider) (_ []RepairLog, retErr error)
	...
	defer func() {
		if retErr != nil {
			...
		}
	}()
	...
}
```

This would allow us better guarantees that hoping that `errFinal` will always be assigned correctly before returning, because in deferred functions, `retErr` will always be equal to the error returned in any return statements (this is a guarantee of named return variables). So we can revert most of the `errFinal` stuff in this PR and just use a named return error (conventionally named `retErr`, there are other instances in the codebase).

PR #5908, opened 2025-07-08
fix(bug): database migration does not rollback schema version on hook failure

bosorawis:
## Description
- continuation of https://github.com/hashicorp/boundary/pull/5904
- address review comment from #5904 
- add test to `schema` package

## PCI review checklist

### DIFF
--- a/internal/db/schema/manager.go
+++ b/internal/db/schema/manager.go
@@ -254,6 +254,7 @@ func (b *Manager) runMigrations(ctx context.Context, p *provider.Provider) ([]Re
 	}
 
 	defer func() {
+		// rolling back a commited run is a no-op, so we can safely call this every time
 		if err := b.driver.RollbackRun(ctx); err != nil {
 			retErr = stderrors.Join(retErr, err)
 		}
--- a/internal/db/schema/manager.go
+++ b/internal/db/schema/manager.go
@@ -33,7 +33,7 @@ type driver interface {
 	StartRun(context.Context) error
 	// CommitRun commits a transaction, if there is an error it should rollback the transaction.
 	CommitRun(context.Context) error
-	// RollbackRun rolls back a transaction.
+	// RollbackRun rolls back a transaction. If no transaction is active, it should return nil.
 	RollbackRun(context.Context) error
 	// CheckHook is a hook that runs prior to a migration's statements.
 	// It should run in the same transaction a corresponding Run call.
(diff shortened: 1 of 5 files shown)


internal/db/schema/manager.go
```diff
@@ -243,85 +243,81 @@ func (b *Manager) ApplyMigrations(ctx context.Context) ([]RepairLog, error) {
...
 	const op = "schema.(Manager).runMigrations"
 
-	var logEntries []RepairLog
-	var errFinal error
-
 	if startErr := b.driver.StartRun(ctx); startErr != nil {
-		errFinal = errors.Wrap(ctx, startErr, op)
-		return nil, errFinal
+		retErr = errors.Wrap(ctx, startErr, op)
+		return
```
johanbrandhorst: There's no need to explicitly assign to `retErr`, you can keep this as just `err` as it was before Michael's PR. Or, just return the wrapped error directly
```suggestion
		return errors.Wrap(ctx, startErr, op)
```

internal/db/schema/manager.go
```diff
@@ -243,85 +243,81 @@ func (b *Manager) ApplyMigrations(ctx context.Context) ([]RepairLog, error) {
...
-				errFinal = stderrors.Join(errFinal, errRollback)
+				retErr = stderrors.Join(retErr, errRollback)
 			}
-			errFinal = errors.Wrap(ctx, errFinal, op)
 			return
 		}
 		if commitErr := b.driver.CommitRun(ctx); commitErr != nil {
-			errFinal = errors.Wrap(ctx, commitErr, op)
+			retErr = errors.Wrap(ctx, commitErr, op)
 		}
```
johanbrandhorst: Inside of the defer, you _should_ assign to `retErr`, so these will need to stay.

internal/db/schema/manager.go
johanbrandhorst: Also, please don't use bare returns. Even with named returns, be explicit

johanbrandhorst (commented):
I would feel that this would be easier to read without all the explicit assignments. Could you try reverting the logic mostly to what we have prior to Michael's PR instead of essentially reusing `retErr` instead of `errFinal`?

internal/db/schema/manager.go
bosorawis: I moved `commit` to the end of the function and only keep `rollback` in the defer. Rollback is called unconditionally because rolling back a committed transaction is a no-op

internal/db/schema/manager.go
```diff
@@ -247,68 +247,55 @@ func (b *Manager) runMigrations(ctx context.Context, p *provider.Provider) ([]Re
...
-			errFinal = errors.Wrap(ctx, errFinal, op)
-			return
-		}
-		if commitErr := b.driver.CommitRun(ctx); commitErr != nil {
-			errFinal = errors.Wrap(ctx, commitErr, op)
+		// rolling back a commited run is a no-op, so we can safely call this every time
+		if err := b.driver.RollbackRun(ctx); err != nil {
+			retErr = stderrors.Join(retErr, err)
 		}
+		return
```
johanbrandhorst: _Technically_ this return isn't needed anymore

johanbrandhorst (dismissed):
Thanks!

tmessi (approved):

PR #5932, opened 2025-07-14
chore(e2e): created an enos scenario for an rdp target

dillanb-hashi:
## Description

Created an enos scenario that spins up an RDP target for RDP testing.

### DIFF
--- a/enos/modules/aws_rdp_server/main.tf
+++ b/enos/modules/aws_rdp_server/main.tf
@@ -188,7 +188,3 @@ resource "time_sleep" "wait_10_minutes" {
   depends_on      = [aws_instance.rdp_target]
   create_duration = "10m"
 }
-
-output "ipv6" {
-  value = flatten(aws_instance.rdp_target.*.ipv6_addresses)
-}
--- a/enos/modules/aws_rdp_server/outputs.tf
+++ b/enos/modules/aws_rdp_server/outputs.tf
@@ -22,3 +22,7 @@ output "password" {
   description = "This is the decrypted administrator password for the EC2 instance"
   value       = nonsensitive(local.password)
 }
+
+output "ipv6" {
+  value = flatten(aws_instance.rdp_target.*.ipv6_addresses)
+}


enos/modules/aws_rdp_server/main.tf
```diff
@@ -0,0 +1,175 @@
...
+data "enos_environment" "current" {}
+
+data "aws_caller_identity" "current" {}
+
+data "aws_ami" "infra" {
+  most_recent = true
+  owners      = ["amazon"]
+  filter {
+    name     = "name"
+    values   = ["Windows_Server-${var.server_version}-English-Full-Base*"]
```
moduli: Another `enos_fmt` pr status check

To make it easier, there should be a make target in the enos directory
```
make fmt
```

enos/enos_scenario-e2e-ui-aws-rdp.hcl
```diff
@@ -0,0 +1,94 @@
+# Copyright (c) HashiCorp, Inc.
```
moduli: For files in `boundary`, we need to add the license identifier. 
```suggestion
# Copyright (c) HashiCorp, Inc.
# SPDX-License-Identifier: BUSL-1.1
```

enos/enos_scenario-e2e-ui-aws-rdp.hcl
```diff
@@ -0,0 +1,95 @@
...
+scenario "e2e_aws_rdp_target" {
+  terraform_cli = terraform_cli.default
+  terraform     = terraform.default
+  providers = [
+    provider.aws.default,
+    provider.enos.default
+  ]
+
+  matrix {
+    builder    = ["local", "crt"]
```
moduli: This looks like it's unused in this scenario. Can remove

enos/enos_scenario-e2e-ui-aws-rdp.hcl
moduli, line 1: The filename should likely be updated as I'm not sure `ui` applies here? There's also an underscore in the name as opposed to a `-`. 

Suggestion:
```
enos-scenario-e2e-aws-rdp-target.hcl
```

enos/enos_scenario-e2e-ui-aws-rdp.hcl
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    tags = merge({
+      "Project Name" : var.project_name
+      "Project" : "Enos",
+      "Environment" : "ci"
+    }, var.tags)
+
+    collocated_tag = "collocated"
+  }
```
moduli: There's likely a number of variables here that aren't used anymore given the removal of other steps.

enos/enos_scenario-e2e-ui-aws-rdp.hcl
```diff
@@ -0,0 +1,95 @@
...
+
+  output "rdp_target_admin_password" {
+    value = step.create_rdp_server.password
+  }
+
+  output "rdp_target_public_dns_address" {
+    value = step.create_rdp_server.public-dns-address
+  }
+
+  output "public_ip" {
```
moduli: May want to be more explicit in the output naming?
```suggestion
  output "rdp_target_public_ip" {
```

enos/modules/aws_rdp_server/main.tf
```diff
@@ -0,0 +1,175 @@
...
+    cidr_blocks = flatten([
+      formatlist("%s/32", data.enos_environment.current.public_ipv4_addresses),
+      join(",", data.aws_vpc.infra.cidr_block_associations.*.cidr_block),
+    ])
+  }
+}
+
+// Create an AWS security group to allow all traffic originating from the default vpc
+resource "aws_security_group" "allow_all_internal" {
+  name   = "${var.prefix}-allow-all-internal-${local.username}"
```
moduli: To ensure uniqueness, we may also want to append the vpc_id to this resource
```suggestion
  name   = "${var.prefix}-allow-all-internal-${local.username}-${var.vpc_id}"
```

enos/enos_scenario-e2e-ui-aws-rdp.hcl
dillanb-hashi: Had this in the ent pr just didn't copy over correctly. Changed.

enos/enos_scenario-e2e-ui-aws-rdp.hcl
dillanb-hashi: Removed

enos/enos_scenario-e2e-ui-aws-rdp.hcl
dillanb-hashi: Removed variables. Testing a scenario launch to make sure it still works.

enos/enos-scenario-e2e-aws-rdp-target.hcl
```diff
@@ -0,0 +1,87 @@
...
+      "crt"   = var.crt_bundle_path == null ? null : abspath(var.crt_bundle_path)
+    }
+
+    tags = merge({
+      "Project Name" : var.project_name
+      "Project" : "Enos",
+      "Environment" : "ci"
+    }, var.tags)
+
+    collocated_tag = "collocated"
```
moduli: There's still a couple of things that aren't used here
- `build_path`
- `collocated_tag`

enos/modules/aws_rdp_server/main.tf
```diff
@@ -0,0 +1,175 @@
...
+    encrypted             = true
+  }
+
+  user_data_replace_on_change = true
+
+  user_data = <<EOF
+                <powershell>
+                  $password = ConvertTo-SecureString ${random_string.DSRMPassword.result} -AsPlainText -Force
+                  Add-WindowsFeature -name ad-domain-services -IncludeManagementTools
+                  Install-ADDSForest -CreateDnsDelegation:$false -DomainMode Win2012R2 -DomainName ${var.active_directory_domain} -DomainNetbiosName ${var.active_directory_netbios_name} -ForestMode Win2012R2 -InstallDns:$true -SafeModeAdministratorPassword $password -Force:$true
```
moduli, line 167: Just noticed there were some references to `Win2012` here. I don't know what this means. Do we just leave as is?

dillanb-hashi: Not sure, that came from your code so I just left it as is...

moduli: Could you research a bit and understand what it means? We'll likely leave as-is, but we should probably get a better understanding before merging things in

dillanb-hashi: Sure thing.

dillanb-hashi: https://learn.microsoft.com/en-us/powershell/module/addsdeployment/install-addsforest?view=windowsserver2025-ps#-domainmode

It indicates the minimum server that all domain controllers have to run. So in this case all domain controllers would have to be Win2012R2 or later. Looks like the only other option would be "WinThreshold" which makes the minimum server 2016. Shouldn't be important unless there's specific domain controller features introduced in windows 2016 that we need to use.

moduli: Makes sense. Thanks for the research!

enos/enos-scenario-e2e-aws-rdp-target.hcl
```diff
@@ -0,0 +1,81 @@
...
+  terraform_cli = terraform_cli.default
+  terraform     = terraform.default
+  providers = [
+    provider.aws.default,
+    provider.enos.default
+  ]
+
+  matrix {
+    rdp_server = ["2016", "2019", "2022", "2025"]
+    ip_version = ["4", "dual"]
```
moduli: Sorry, one more thing I just thought of. Do we want to propagate this to the `aws_rdp_server` module in any way and have it potentially return an ipv6 address? (you can see the `aws_target` as an example). Otherwise, the other option is to get rid of this variable in the matrix and hardcode `create_base_infra` to just use `aws_vpc` to avoid confusion for why we're specifying `ip_version` in the scenario.

moduli (commented):
Looks good! One more question...

enos/enos-scenario-e2e-aws-rdp-target.hcl
dillanb-hashi: I would probably prefer just to hardcode dual and then return both in the output. Unless there's some usecase for just doing 4 or 6?

moduli: 👍 I like that

dillanb-hashi: Done

enos/modules/aws_rdp_server/main.tf
```diff
@@ -172,4 +187,8 @@ locals {
 resource "time_sleep" "wait_10_minutes" {
   depends_on      = [aws_instance.rdp_target]
   create_duration = "10m"
+}
+
+output "ipv6" {
+  value = flatten(aws_instance.rdp_target.*.ipv6_addresses)
```
moduli: Probably want to move this to `outputs.tf`?

dillanb-hashi: Yep. Done

moduli (approved):

Issue #4759, opened 2024-05-01
Creating multiple target connections with the same client port hangs the boundary CLI


haniffm:
**Describe the bug**
When using Boundary CLI to create a connections with the same client port to listen to, Boundary CLI gets unresponsive.

**To Reproduce**
1. Run `boundary authenticate  ...`
2. Run `boundary connect ... -listen-port 2222`. You should get a "Proxy listening information" output.
3. In a different terminal, run `boundary connect ... -listen-port 2222` (notice the same port). This time you don't get any output, and the CLI get unresponsive. You can't even `CTRL+C` out of it. I have to do `kill -9` on the last connection.

**Expected behavior**
I would expect that the boundary CLI would give an error message on that the port is already occupied and exit with non-zero.

**Additional context**
This happens often when a `default client port` is configured and the user is not sure that a connection is already establish.

PR #5939, opened 2025-07-21
bug: fix multiple connections to same listen port hanging CLI

trinhrichard:
Issue #4759 

When using Boundary CLI to establish multiple connections using the same local listen port, the CLI becomes unresponsive on subsequent connections. The process does not handle the "address already in use" error gracefully and requires a manual kill -9 to terminate.

* Added error handling when attempting to use an occupied listen port


internal/cmd/commands/connect/connect.go
```diff
@@ -732,3 +739,12 @@ func (c *Command) handleExec(clientProxy *apiproxy.ClientProxy, passthroughArgs
 	}
 	c.execCmdReturnValue.Store(0)
 }
+
+func isPortAvailable(listenPort int64, listenAddr string) bool {
```
sepehrfrgh: I would suggest returning a bool and an error in this function for handling errors gracefully

internal/cmd/commands/connect/connect.go
```diff
@@ -732,3 +739,12 @@ func (c *Command) handleExec(clientProxy *apiproxy.ClientProxy, passthroughArgs
...
+	listener, err := net.Listen("tcp", fmt.Sprintf("%s:%d", listenAddr, listenPort))
```
sepehrfrgh: nitpik just to make things more clearer
```suggestion
	address := fmt.Sprintf("%s:%d", listenAddr, listenPort)
	listener, err := net.Listen("tcp", address)
```

internal/cmd/commands/connect/connect.go
```diff
@@ -732,3 +739,12 @@ func (c *Command) handleExec(clientProxy *apiproxy.ClientProxy, passthroughArgs
...
+	if err != nil {
```
sepehrfrgh: I would suggest checking errors and returning error as well. example you can use `errors.Is(err, syscall.EADDRINUSE)` to check if the address is in use.

internal/cmd/commands/connect/connect.go
```diff
@@ -732,3 +739,12 @@ func (c *Command) handleExec(clientProxy *apiproxy.ClientProxy, passthroughArgs
...
+		return false
+	}
+	_ = listener.Close()
```
sepehrfrgh: Try to handle error closing  
```suggestion
defer func() {
		err = listener.Close()
		if err != nil {
                         error message
		}
	}()
```

testing/internal/e2e/tests/base/target_tcp_connect_ssh_test.go
```diff
@@ -106,3 +106,107 @@ func TestCliTcpTargetConnectTargetWithSsh(t *testing.T) {
...
+			"connect",
+			"-target-id", targetId,
+			"-listen-port", ListenPort, "--",
+			"-o", "UserKnownHostsFile=/dev/null",
+			"-o", "StrictHostKeyChecking=no",
+			"-o", "IdentitiesOnly=yes",
+		),
+	)
+
+	//output = e2e.RunCommand(ctx, "boundary",
```
sepehrfrgh: What is this comment? If it is unused, let's delete it

louisruch (changes requested):
Thanks for diving into this issue and working through a fix. Something felt a bit off with the approach of checking for an available port and then trying to connect to it. This still leaves a race condition where after the `isPortAvailable` check we could have two connection trying to use the same port, even though both passed the check. 

Looking at the code we are doing calling `ListenTCP` in the `clientProxy.Start`, this function should be returning the same error you are looking for in the new `isPortAvailable` check. I played with this a little bit and believe the following diff should resolve the issue:

```
@@ -455,7 +455,10 @@ func (c *Command) Run(args []string) (retCode int) {
        proxyError := new(atomic.Error)
        go func() {
                defer close(clientProxyCloseCh)
-               proxyError.Store(clientProxy.Start())
+               if err := clientProxy.Start(); err != nil {
+                       c.proxyCancel() // This will trigger the other goroutine to exit
+                       proxyError.Store(err)
+               }
        }()
        go func() {
                defer close(connCountCloseCh)
@@ -479,7 +482,7 @@ func (c *Command) Run(args []string) (retCode int) {
                // The only way a user will be able to connect to the session is by
                // connecting directly to the port and address we report to them here.

-               proxyAddr := clientProxy.ListenerAddress(context.Background())
+               proxyAddr := clientProxy.ListenerAddress(c.proxyCtx)
                var clientProxyHost, clientProxyPort string
                clientProxyHost, clientProxyPort, err = util.SplitHostPort(proxyAddr)
                if err != nil && !errors.Is(err, util.ErrMissingPort) {
```

Let me know if you wanted to talk anything through

trinhrichard:
@louisruch Good point, I never considered the race condition aspect. I have a question with this line,

> proxyAddr := clientProxy.ListenerAddress(c.proxyCtx)

Do we know if c.proxyCtx is guaranteed to be non nil at this point? I was wondering if there's a chance it could be nil, or if it's always initialized via c.Context. Just trying to understand the context lifecycle here

louisruch:
> @louisruch Good point, I never considered the race condition aspect. I have a question with this line,
> 
> > proxyAddr := clientProxy.ListenerAddress(c.proxyCtx)
> 
> Do we know if c.proxyCtx is guaranteed to be non nil at this point? I was wondering if there's a chance it could be nil, or if it's always initialized via c.Context. Just trying to understand the context lifecycle here

The proxyCtx is set on line 279 in the same function so it should always be non-nil and created with a cancel based on c.Context as the parent

 https://github.com/hashicorp/boundary/blob/main/internal/cmd/commands/connect/connect.go#L279

trinhrichard:
Gotcha, that clears things up, thanks!

testing/internal/e2e/tests/base/target_tcp_connect_ssh_test.go
```diff
@@ -106,3 +106,96 @@ func TestCliTcpTargetConnectTargetWithSsh(t *testing.T) {
...
+	)
+
+	require.NoError(t, output.Err, string(output.Stderr))
+	t.Log("Successfully connected to target with listen port 3333")
+
+	// Testing multiple connections with same listen port
+	const commandCliError = 2
+	const maxAttempts = 10
+
+	for i := 0; i < maxAttempts; i++ {
```
louisruch: Quick side note here, go 1.22 added range over types like an integer. So this could become `for i := range maxAttempts`. Since you never use the variable `i` here you can actually do `for range maxAttempts`

testing/internal/e2e/tests/base/target_tcp_connect_ssh_test.go
```diff
@@ -106,3 +106,96 @@ func TestCliTcpTargetConnectTargetWithSsh(t *testing.T) {
...
+			"-o", "IdentitiesOnly=yes", // forces the use of the provided key
+		),
+	)
+
+	require.NoError(t, output.Err, string(output.Stderr))
+	t.Log("Successfully connected to target")
+
+	// Connect to target on port 3333
+	const ListenPort = "3333"
+	output = e2e.RunCommand(ctx, "boundary",
```
louisruch: Hmm I am not sure exactly how to handle this. We seem to be getting stuck running this command and the test panics after the 10 minute timeout. This makes sense since the `boundary connect` command is a blocking call until we quit it. We might want to chat with @moduli and determine if we every test the blocking `connect` command and if now how we could. Other versions I have seen either use the `ssh` helper or run an `-exec`

testing/internal/e2e/tests/base/target_tcp_connect_ssh_test.go
```diff
@@ -106,3 +106,96 @@ func TestCliTcpTargetConnectTargetWithSsh(t *testing.T) {
...
+				"-target-id", targetId,
+				"-listen-port", ListenPort, "--",
+				"-o", "UserKnownHostsFile=/dev/null",
+				"-o", "StrictHostKeyChecking=no",
+				"-o", "IdentitiesOnly=yes",
+			),
+		)
+
+		require.Error(t, output.Err, string(output.Stderr))
+		require.Contains(t, string(output.Stderr), "Port is already in use")
```
louisruch: When the test reaches this spot we will want to update it to look for `address already in use` as returned from ListenTCP

testing/internal/e2e/tests/base/target_tcp_connect_ssh_test.go
```diff
@@ -106,3 +106,96 @@ func TestCliTcpTargetConnectTargetWithSsh(t *testing.T) {
...
+	hostSetId, err := boundary.CreateHostSetCli(t, ctx, hostCatalogId)
+	require.NoError(t, err)
+	hostId, err := boundary.CreateHostCli(t, ctx, hostCatalogId, c.TargetAddress)
+	require.NoError(t, err)
+	err = boundary.AddHostToHostSetCli(t, ctx, hostSetId, hostId)
+	require.NoError(t, err)
+	targetId, err := boundary.CreateTargetCli(t, ctx, projectId, c.TargetPort)
+	require.NoError(t, err)
+	err = boundary.AddHostSourceToTargetCli(t, ctx, targetId, hostSetId)
+	require.NoError(t, err)
```
moduli: We could simplify this a bit. We have an alternative way to set up a target where we can set the address directly on the target as opposed to using hosts. All of this could just be changed to...
```
targetId, err := boundary.CreateTargetCli(t, ctx, projectId, c.TargetPort,
	target.WithAddress(c.TargetAddress),
)
```

testing/internal/e2e/tests/base/target_tcp_connect_ssh_test.go
```diff
@@ -106,3 +106,96 @@ func TestCliTcpTargetConnectTargetWithSsh(t *testing.T) {
 	require.Equal(t, c.TargetAddress, strings.TrimSpace(string(output.Stdout)))
 	t.Log("Successfully connected to target")
 }
+
+// TestCliSshTargetConnectListenPortTest uses the boundary cli to create an ssh target,
+// attach injected credentials to the target, and connect to the target multiple times on the same listen port.
+func TestCliSshTargetConnectListenPortTest(t *testing.T) {
```
moduli: The test names have been referring to the target type that we're creating, and in this case, we're technically still creating a tcp-type target in Boundary even though we're using it for SSH purposes. By default, `CreateTargetCli` creates a tcp-type target. Boundary has a special ssh-type target, but it's only available in the enterprise version.
```suggestion
// TestCliTcpTargetConnectListenPortTest uses the boundary cli to create a target and connect to it multiple times on the same listen port. The expectation is that the cli will return an error when a port is already in use.
func TestCliTcpTargetConnectListenPortTest(t *testing.T) {
```

testing/internal/e2e/tests/base/target_tcp_connect_ssh_test.go
moduli, line 1: If you feel like this file is getting a little cluttered, I wouldn't be opposed to creating a new file for this test. Something like...
```
target_tcp_connect_listen_port_test.go
```

testing/internal/e2e/tests/base/target_tcp_connect_ssh_test.go
moduli: Also, we're technically using "brokered" credentials to the target (as opposed to injected). The end result is similar, but injected has a special meaning where Boundary handles the credentials to the target, whereas brokered passes the credentials back to the user. Injected credentials are only available in the enterprise version (and only on specific target types).

testing/internal/e2e/tests/base/target_tcp_connect_ssh_test.go
moduli: It sounds like the test needs to do the following
- start a session on a particular listen port and have it be "active"
- attempt to start a session the same listen port and expect an error

In order to accomplish the first bit, we'll want a way to start and maintain a session while allowing the test to proceed. In other tests, we've utilized `go func` to accomplish this. Here's an example: https://github.com/hashicorp/boundary/blob/main/testing/internal/e2e/tests/aws/dynamichostcatalog_host_set_test.go#L141-L160
```
// Connect to target
ctxCancel, cancel := context.WithCancel(context.Background())
go func() {
	e2e.RunCommand(ctxCancel, "boundary",
		e2e.WithArgs(
			"connect",
			"ssh",
			"-target-id", targetId,
			"-listen-port", ListenPort,                
			"-remote-command", "hostname -i; sleep 60",  # <-- executes a command on the target
			"--",
			"-o", "UserKnownHostsFile=/dev/null",
			"-o", "StrictHostKeyChecking=no",
			"-o", "IdentitiesOnly=yes", // forces the use of the provided key
		),
	)
}()
t.Cleanup(cancel)

s := boundary.WaitForSessionCli(t, ctx, projectId)
boundary.WaitForSessionStatusCli(t, ctx, s.Id, session.StatusActive.String())
```
This starts an ssh connection and runs `hostname -i; sleep 60` on that instance. 60 seconds should be enough for this test. 

Now, we can initiate additional connections to the target. We won't need the `sleep` or the `go func` in these commands as they will return immediately whether it's a success or failure.
```
output := e2e.RunCommand(ctxCancel, "boundary",
		e2e.WithArgs(
			"connect",
			"ssh",
			"-target-id", targetId,
			"-listen-port", ListenPort,                
			"-remote-command", "hostname -i,
			"--",
			"-o", "UserKnownHostsFile=/dev/null",
			"-o", "StrictHostKeyChecking=no",
			"-o", "IdentitiesOnly=yes", // forces the use of the provided key
		),
	)
```
You can check if the appropriate error was returned here.

testing/internal/e2e/tests/base/target_tcp_connect_ssh_test.go
moduli, line 1: Something I thought of... what also could be interesting is doing a similar test, but instead of specifying the `listen-port` in the `boundary connect` command, we would just define a `defaultClientPort` on the target.

testing/internal/e2e/tests/base/target_tcp_connect_ssh_test.go
louisruch: Thanks for the detailed response @moduli 
Though there are a couple things to note @trinhrichard 

Technically with the approach I recommended you will not get a CLI error without further refactoring, instead you will get a termination reason. So we may need to look at stdout for the expected string.
Also note that the `-listen-port` does not work with the `ssh` subcommand as there is no need for a specific listen port when Boundary launches the client for the user.

moduli: Ah good point. In that case, you could just do...
```
go func() {
		t.Log("Starting session...")
		sessionChannel <- e2e.RunCommand(ctxCancel, "boundary",
			e2e.WithArgs(
				"connect",
				"-target-id", targetId,
				"-listen-port", port,
				"-format", "json",
			),
		)
	}()
	t.Cleanup(cancel)

	boundary.WaitForSessionCli(t, ctx, projectId)
```

testing/internal/e2e/tests/base/target_tcp_connect_ssh_test.go
```diff
@@ -106,3 +109,135 @@ func TestCliTcpTargetConnectTargetWithSsh(t *testing.T) {
...
+	output := e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"connect",
+			"-target-id", targetId,
+			"-listen-port", ListenPort,
+		),
+	)
+
+	// Starting a proxy with same listen port already in use returns termination info instead of an error,
+	// so we should expect exit code 0 with the error message in Stdout
```
moduli: Is that expected? Should we be returning a non-zero exit code in this case?

testing/internal/e2e/tests/base/target_tcp_connect_ssh_test.go
moduli: Will we be moving these tests? (especially since these tests don't actually use `connect ssh`)

testing/internal/e2e/tests/base/target_tcp_connect_ssh_test.go
louisruch: The reason this happens is because we start the `proxyClient` in a goroutine and then display appropriate data to the user in order for them to connect to. Any proxy errors that happen after we start the client are considered a termination of the proxy while it was running instead of a CLI error, see the output below:

```
boundary connect -target-id ttcp_1234567890 -listen-port 3800

Proxy listening information:
  Address:
  Connection Limit:    -1
  Expiration:          Thu, 24 Jul 2025 16:31:43 PDT
  Port:                0
  Protocol:            tcp
  Session ID:          s_fNiymqWnmJ

Termination information:
  Reason:   Error from proxy client: unable to start listening: listen tcp 127.0.0.1:3800: bind: address already in use
```

The problem here is that this code path does not handle errors that occur **BEFORE** starting the proxy, we just assume once we call start we can moved on. 

One reason to attempt to fix this would be if you look at the `Proxy listening information` above. The displayed data is invalid, there is no address and a zero port. I did a quick check and something like this could work if we want to change this behavior (please note I have not thoroughly tested below):

```
diff --git a/internal/cmd/commands/connect/connect.go b/internal/cmd/commands/connect/connect.go
index 879320dc2..20d19f605 100644
--- a/internal/cmd/commands/connect/connect.go
+++ b/internal/cmd/commands/connect/connect.go
@@ -477,6 +477,17 @@ func (c *Command) Run(args []string) (retCode int) {
                // connecting directly to the port and address we report to them here.

                proxyAddr := clientProxy.ListenerAddress(c.proxyCtx)
+               if proxyAddr == "" {
+                       // We got an empty address, which means the proxy didn't start. Check
+                       // if we have an error to report
+                       if err := proxyError.Load(); err != nil {
+                               c.PrintCliError(fmt.Errorf("Error starting proxy: %w", err))
+                               return base.CommandCliError
+                       }
+                       // If we reach here, it means the proxy didn't start for some other reason
+                       c.PrintCliError(fmt.Errorf("Error starting proxy: no address returned"))
+                       return base.CommandCliError
+               }
                var clientProxyHost, clientProxyPort string
                clientProxyHost, clientProxyPort, err = util.SplitHostPort(proxyAddr)
                if err != nil && !errors.Is(err, util.ErrMissingPort) {
```

testing/internal/e2e/tests/base/target_tcp_connect_ssh_test.go
```diff
@@ -6,6 +6,9 @@ package base_test
 import (
 	"context"
 	"encoding/json"
+	"fmt"
+	"github.com/hashicorp/boundary/internal/session"
+	"github.com/hashicorp/boundary/internal/target"
```
louisruch: You will want to move these two into the section import group below

testing/internal/e2e/tests/base/target_tcp_connect_ssh_test.go
```diff
@@ -106,3 +109,135 @@ func TestCliTcpTargetConnectTargetWithSsh(t *testing.T) {
...
+	// Starting a proxy with same listen port already in use returns termination info instead of an error,
+	// so we should expect exit code 0 with the error message in Stdout
+	require.Contains(t, string(output.Stdout), "address already in use")
+	require.Equal(t, 0, output.ExitCode)
+}
+
+// TestCliTcpTargetConnectDefaultClientPortTest uses the boundary cli to create a target and
+// connect to it multiple times on the same client port. The test is similar to the one above,
+// but sets DefaultClientPort instead of ListenPort
+func TestCliTcpTargetConnectDefaultClientPortTest(t *testing.T) {
```
louisruch: <3

testing/internal/e2e/tests/base/target_tcp_connect_ssh_test.go
moduli: For more information, you can see the failed `make-gen-delta` PR status check below. Here is a link to it: https://github.com/hashicorp/boundary/actions/runs/16484065165/job/46605083623?pr=5939

testing/internal/e2e/tests/base/target_tcp_connect_ssh_test.go
louisruch: Ok I have looked a bit further and noticed this comment from Jeff when he originally wrote all this
https://github.com/hashicorp/boundary/blob/main/api/proxy/proxy.go#L366

Specifically he calls out `In either case, test the result to ensure it's not empty.`
I therefore believe we should update as per my latest diff or similar to validate this address is not empty.

trinhrichard: Good point, didn't realize the proxy info would be incorrect

testing/internal/e2e/tests/base/target_tcp_connect_listen_port_test.go
```diff
@@ -0,0 +1,149 @@
...
+		e2e.WithArgs(
+			"connect",
+			"-target-id", targetId,
+			"-listen-port", ListenPort,
+		),
+	)
+
+	const commandCliError = 2
+	require.Contains(t, string(output.Stderr), "address already in use")
+	require.Equal(t, commandCliError, output.ExitCode)
```
moduli: nit: Not sure if we need to define an extra const here
```suggestion
	require.Contains(t, string(output.Stderr), "address already in use")
	require.Equal(t, 2, output.ExitCode)
```

moduli: @louisruch Does exit code `2` make sense here to you? I don't know if we have any rules for what these should be. This is what happens when I do this manually
```
❯ boundary connect -target-id ttcp_6z25QI3g95 -listen-port 2323
Error starting proxy: unable to start listening: listen tcp 127.0.0.1:2323: bind: address already in use
~ ················································································ 15:54:41
❯ echo $?
2
```

testing/internal/e2e/tests/base/target_tcp_connect_listen_port_test.go
```diff
@@ -0,0 +1,149 @@
...
+				"-target-id", targetId,
+				"-listen-port", ListenPort,
+			),
+		)
+	}()
+	t.Cleanup(cancel)
+
+	s := boundary.WaitForSessionCli(t, ctx, projectId)
+	boundary.WaitForSessionStatusCli(t, ctx, s.Id, session.StatusPending.String())
+	t.Log(fmt.Sprintf("Successfully connected to target with listen port %s", ListenPort))
```
moduli: nit: `t.Logf` will accomplish the same thing 
```suggestion
	t.Logf("Successfully connected to target with listen port %s", ListenPort)
```

testing/internal/e2e/tests/base/target_tcp_connect_listen_port_test.go
```diff
@@ -0,0 +1,149 @@
...
+				"connect",
+				"-target-id", targetId,
+			),
+		)
+	}()
+	t.Cleanup(cancel)
+
+	s := boundary.WaitForSessionCli(t, ctx, projectId)
+	boundary.WaitForSessionStatusCli(t, ctx, s.Id, session.StatusPending.String())
+	t.Log(fmt.Sprintf("Successfully connected to target with client port %d", DefaultClientPort))
```
moduli: we can use `t.Logf` here

testing/internal/e2e/tests/base/target_tcp_connect_listen_port_test.go
```diff
@@ -0,0 +1,149 @@
...
+	t.Log(fmt.Sprintf("Successfully connected to target with listen port %s", ListenPort))
+
+	t.Logf("Starting new connection to target with same listen port %s", ListenPort)
+	output := e2e.RunCommand(ctx, "boundary",
+		e2e.WithArgs(
+			"connect",
+			"-target-id", targetId,
+			"-listen-port", ListenPort,
+		),
+	)
```
moduli, line 71: We probably want to check that there's an error here. Something like...
```
require.Error(t, output.Err, "Expected error when connecting to target with same listen port")
```

testing/internal/e2e/tests/base/target_tcp_connect_listen_port_test.go
```diff
@@ -0,0 +1,149 @@
...
+		target.WithAddress(c.TargetAddress),
+	)
+	require.NoError(t, err)
+	storeId, err := boundary.CreateCredentialStoreStaticCli(t, ctx, projectId)
+	require.NoError(t, err)
+	credentialId, err := boundary.CreateStaticCredentialPrivateKeyCli(t, ctx, storeId, c.TargetSshUser, c.TargetSshKeyPath)
+	require.NoError(t, err)
+	err = boundary.AddBrokeredCredentialSourceToTargetCli(t, ctx, targetId, credentialId)
+	require.NoError(t, err)
+	t.Log("Successfully attached brokered credentials")
```
moduli: We technically don't really need these steps as we're not using the credentials to ssh on to the instance in this test. This could be removed.

moduli (commented):
Nice! Getting close!

testing/internal/e2e/tests/base/target_tcp_connect_listen_port_test.go
louisruch: Yeah I think so. Our classifications of `cmd` errors stem from:
```
const (
	CommandSuccess int = iota
	CommandApiError
	CommandCliError
	CommandUserError
)
```

Of that it can either be a 2 or a 3. Since we do not know for sure its a user error I believe 2 is probably appropriate.

moduli (approved):
Awesome. Nice work! Looks good to me pending successful build/e2e tests passing.

louisruch (approved):
Great work on this @trinhrichard! :shipit:

trinhrichard:
Thank you @louisruch and @moduli! Your suggestions were very insightful!

sepehrfrgh (approved):
Thank you Richard this is great

PR #5947, opened 2025-07-22
Cassandra command e2e tests

bgajjala8:
## Description
Cassandra command e2e tests


testing/internal/e2e/infra/docker.go
```diff
@@ -344,3 +347,128 @@ func StartMysql(t testing.TB, pool *dockertest.Pool, network *dockertest.Network
...
+		UriNetwork: fmt.Sprintf("cassandra://%s:9042/%s",
+			networkAlias,
+			cassandraKeyspace,
+		),
+	}
+}
+
+// StartCassandraWithAuth starts a Cassandra database in a docker container with authentication enabled.
+// Returns information about the container
+func StartCassandraWithAuth(t testing.TB, pool *dockertest.Pool, network *dockertest.Network, repository, tag string, ctx context.Context) *Container {
```
RyanDerr: Should we expose this function? It appears for testing framework we likely always want to have a CassandraDB container that has auth and some data configured to which we perform tests against. I think we should have this be a helper function that `StartCassandra` calls (passing in only needed params) and returns only `error`.

testing/internal/e2e/infra/docker.go
```diff
@@ -344,3 +347,128 @@ func StartMysql(t testing.TB, pool *dockertest.Pool, network *dockertest.Network
...
+	err = pool.Client.RestartContainer(c.Resource.Container.ID, uint(pool.MaxWait.Seconds()))
+	require.NoError(t, err)
+
+	err = pool.Retry(func() error {
+		return exec.CommandContext(ctx, "docker", "exec", networkAlias,
+			"cqlsh", "-u", "cassandra", "-p", "cassandra", "-e", "SELECT now() FROM system.local;").Run()
+	})
+	require.NoError(t, err)
+
+	t.Log("Creating User")
```
RyanDerr: This log is not very reflected of the total operations being preformed. Additionally I recommend we group all of the database cql commands into this function rather than segmenting them if this creates data and auth which it appears to do.

testing/internal/e2e/infra/docker.go
```diff
@@ -344,3 +347,128 @@ func StartMysql(t testing.TB, pool *dockertest.Pool, network *dockertest.Network
...
+	u, err := url.Parse(c.UriNetwork)
+	require.NoError(t, err)
+
+	keyspace := u.Path[1:]
+	cassandraUser := "e2eboundary"
+	cassandraPassword := "e2eboundary"
+	networkAlias := c.Resource.Container.Name[1:]
+
+	t.Logf("Enabling Password Auth for Cassandra")
+	sedCommands := []string{
```
RyanDerr: To an outside observer it's not clear why this is happening and then we perform a container restart. Mind adding a few comments?

testing/internal/e2e/tests/base/target_tcp_connect_cassandra_test.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+// TestCliTcpTargetConnectCassandra uses the boundary cli to connect to a
+// target using `connect cassandra`
+func TestCliTcpTargetConnectCassandra(t *testing.T) {
+	e2e.MaybeSkipTest(t)
+
+	pool, err := dockertest.NewPool("")
+	require.NoError(t, err)
+	pool.MaxWait = 90 * time.Second
```
RyanDerr, line 34: Mind leaving a comment as to why we are increasing the wait time for this test?

testing/internal/e2e/tests/base/target_tcp_connect_cassandra_test.go
```diff
@@ -0,0 +1,124 @@
...
+	f, err := pty.Start(cmd)
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		err := f.Close()
+		require.NoError(t, err)
+	})
+
+	_, err = f.Write([]byte("DESCRIBE KEYSPACES;\n"))
+	require.NoError(t, err)
+	_, err = f.Write([]byte("SELECT keyspace_name FROM system_schema.keyspaces WHERE keyspace_name = '" + keyspace + "';\n"))
```
RyanDerr: Can we use a `fmt.Sprintf` to make this a little easier to read?

RyanDerr (changes requested):

testing/internal/e2e/tests/base/target_tcp_connect_cassandra_test.go
bgajjala8: Thanks!

testing/internal/e2e/tests/base/target_tcp_connect_cassandra_test.go
bgajjala8: Thanks, fixed

testing/internal/e2e/infra/docker.go
bgajjala8: Thanks, yep good call.

testing/internal/e2e/infra/docker.go
bgajjala8: Sounds good, I agree. Fixed!

testing/internal/e2e/infra/docker.go
bgajjala8: Updated

bgajjala8:
@RyanDerr Updated based on review. I'd appreciate a re-review on how I've handled the Cassandra authentication split 😄

RyanDerr:
> @RyanDerr Updated based on review. I'd appreciate a re-review on how I've handled the Cassandra authentication split 😄

Do we also need to add the cli for cqlsh to support these tests as part of enos crt?

testing/internal/e2e/tests/base/target_tcp_connect_cassandra_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+// TestCliTcpTargetConnectCassandra uses the boundary cli to connect to a
+// target using `connect cassandra`
+func TestCliTcpTargetConnectCassandra(t *testing.T) {
+	e2e.MaybeSkipTest(t)
+
+	pool, err := dockertest.NewPool("")
+	require.NoError(t, err)
+
+	// Increase timeout to accommodate Cassandra's longer startup duration
```
RyanDerr: Mind adding a comment that specifies why it takes longer to setup (just a short mention such as longer to startup do to gossip needing to settle)

testing/internal/e2e/infra/docker.go
```diff
@@ -344,3 +345,114 @@ func StartMysql(t testing.TB, pool *dockertest.Pool, network *dockertest.Network
...
+	networkAlias := "e2ecassandra"
+	cassandraKeyspace := "e2eboundarykeyspace"
+	cassandraUser := "e2eboundary"
+	cassandraPassword := "e2eboundary"
+
+	resource, err := pool.RunWithOptions(&dockertest.RunOptions{
+		Repository: fmt.Sprintf("%s/%s", c.DockerMirror, repository),
+		Tag:        tag,
+		Env: []string{
+			"CASSANDRA_CLUSTER_NAME=e2e-boundary-cluster",
```
RyanDerr, line 381: Should we hardcode this? Looking at the other DB examples we allow specification of the db name.

testing/internal/e2e/infra/docker.go
```diff
@@ -344,3 +345,114 @@ func StartMysql(t testing.TB, pool *dockertest.Pool, network *dockertest.Network
...
+			cassandraUser,
+			cassandraPassword,
+			networkAlias,
+			cassandraKeyspace,
+		),
+	}
+}
+
+// setupCassandraAuthAndUser enables authentication on a Cassandra container and creates a user with permissions.
+func setupCassandraAuthAndUser(t testing.TB, resource *dockertest.Resource, pool *dockertest.Pool, cassandraUser, cassandraPassword, keyspace, networkAlias string, cassandraKeyspace string) {
```
RyanDerr: Would it make more sense for this function to be a `t.Helper()` and return an `error` that we require not error on the caller function, rather than doing all the assertions here? Additionally would it be possible to collapse the parameter inputs?

website/content/docs/commands/connect/casandra.mdx
```diff
@@ -0,0 +1,68 @@
+---
```
RyanDerr: Does this doc contain any mention of the limitations of the connection command with regards to to the auth method enabled on the cassandra instance?

RyanDerr: Also the document is misspelled leading to a reference error when building the website in Vercel.

testing/internal/e2e/tests/base/target_tcp_connect_cassandra_test.go
bgajjala8: Done

website/content/docs/commands/connect/casandra.mdx
bgajjala8: Good catch, fixed

testing/internal/e2e/infra/docker.go
bgajjala8: I think for the scope of this, since the caller will have to parse the URI, and we are just testing the Cassandra function, it should be fine to leave it like this.

testing/internal/e2e/infra/docker.go
bgajjala8: Does it make sense to create another struct specifically for Cassandra values, or should `setupCassandraAuthAndUser` define and return the variables itself?

testing/internal/e2e/infra/docker.go
```diff
@@ -26,6 +27,14 @@ type Container struct {
 	UriNetwork   string
 }
 
+// CassandraConfig stores configuration details for the Cassandra container
+type CassandraConfig struct {
```
bgajjala8, line 31: I am not a huge fan of this. There are to many generics.... what do you think? @RyanDerr

testing/internal/e2e/infra/docker.go
```diff
@@ -372,87 +382,102 @@ func StartCassandra(t testing.TB, pool *dockertest.Pool, network *dockertest.Net
...
+			config.Password,
+			config.NetworkAlias,
+			config.Keyspace,
 		),
 	}
 }
 
 // setupCassandraAuthAndUser enables authentication on a Cassandra container and creates a user with permissions.
-func setupCassandraAuthAndUser(t testing.TB, resource *dockertest.Resource, pool *dockertest.Pool, cassandraUser, cassandraPassword, keyspace, networkAlias string, cassandraKeyspace string) {
+func setupCassandraAuthAndUser(t testing.TB, resource *dockertest.Resource, pool *dockertest.Pool, config CassandraConfig) error {
```
RyanDerr: Should we pass by value or reference here for the config?

testing/internal/e2e/infra/docker.go
```diff
@@ -372,87 +382,102 @@ func StartCassandra(t testing.TB, pool *dockertest.Pool, network *dockertest.Net
...
+		return nil
+	}); err != nil {
+		return fmt.Errorf("cassandra did not become ready: %w", err)
+	}
 
-	t.Logf("Initializing Cassandra keyspace: %s...", cassandraKeyspace)
-	cmd := fmt.Sprintf("CREATE KEYSPACE IF NOT EXISTS %s WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1};", cassandraKeyspace)
-	err = exec.Command("docker", "exec", networkAlias, "cqlsh", "-e", cmd).Run()
-	require.NoError(t, err)
+	// Create keyspace
```
RyanDerr: IMO if the operation is self explanatory or likely doesn't need background to understand why we're doing this it can be removed, as this is rather self evident for what it's doing and doesn't provide much detail for this and a few other comments.

testing/internal/e2e/infra/docker.go
RyanDerr: You could use a struct if desired for config options or required attributes. I'll let you use your best judgement for this.

testing/internal/e2e/infra/docker.go
RyanDerr: IMO it does make it cleaner though we in reality could make options for this if we wanted instead, though again its up to you. I only recommend that we follow existing patterns.

bgajjala8: Going to leave it as is. It does not make sense for a helper to return all values. It doesn't meet SRP

testing/internal/e2e/infra/docker.go
```diff
@@ -387,7 +387,7 @@ func StartCassandra(t testing.TB, pool *dockertest.Pool, network *dockertest.Net
 	})
 	require.NoError(t, err)
 
-	err = setupCassandraAuthAndUser(t, resource, pool, config)
+	err = setupCassandraAuthAndUser(t, resource, pool, &config)
```
RyanDerr, line 390: *Nitpick*: I would simply make config the pointer itself

PR #5950, opened 2025-07-24
DO NOT MERGE: Expose data warehouse tables over mcp server

dkanney:
## Description

During RND Hackweek 2025, we aimed to create an MCP server for arbitrary Boundary deployments. We created a couple MCP servers - a standalone server that can interact with the Boundary API, and a server integrated into Boundary itself (this PR).

The Boundary MCP server listens on `localhost:9203/mcp` and contains two tools:
- `get_warehouse_schemas`
    - This tool reads from a DB dump file of warehouse tables
- `run_sql_query`
    - This tools accepts arbitrary SQL and returns the results

These tools enable an LLM to read the warehouse schema and run queries against Boundary's database.

## PCI review checklist


Makefile
```diff
@@ -319,6 +319,11 @@ test-database-down:
 generate-database-dumps:
 	@$(MAKE) -C testing/dbtest/docker generate-database-dumps
 
+.PHONY: db-schema
+db-schema:
+	@echo "==> Generating database schema"
+	@pg_dump --schema-only --table='wh_*' > ./internal/warehouse/schema.sql
```
johanbrandhorst: Is this tool already installed for users? Maybe something to add to `make tools`?

internal/daemon/mcpserver/mcp.go
```diff
@@ -0,0 +1,135 @@
...
+type GetWarehouseSchemasResult struct {
+	Schemas []string `json:"schemas" jsonschema:"list of schemas"`
+}
+
+// GetWarehouseSchemas reads a premade database dump of the warehouse tables from a file and returns it as a string blob.
+func (s *MCPServer) GetWarehouseSchemas(ctx context.Context, cc *mcp.ServerSession, params *mcp.CallToolParamsFor[GetWarehouseSchemasParams]) (*mcp.CallToolResultFor[GetWarehouseSchemasResult], error) {
+	file, err := os.ReadFile(warehouseTableDumpFilePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read database dump file: %w", err)
+	}
```
johanbrandhorst: Instead of this, please use `go:embed` to embed this schema file into the binary at build time. See https://pkg.go.dev/embed

internal/daemon/mcpserver/mcp.go
```diff
@@ -0,0 +1,135 @@
...
+		StructuredContent: resp,
+	}, nil
+}
+
+type RunSQLQueryParams struct {
+	Query string `json:"query" jsonschema:"the SQL query string to run"`
+}
+
+type RunSQLQueryResult struct {
+	Results []interface{} `json:"results" jsonschema:"the query results"`
```
johanbrandhorst: Nit: use `any`
```suggestion
	Results []any `json:"results" jsonschema:"the query results"`
```

internal/warehouse/repository.go
```diff
@@ -0,0 +1,83 @@
...
+	return &Repository{
+		reader: r,
+	}, nil
+}
+
+func (r *Repository) GetWarehouseSchemas(ctx context.Context) (string, error) {
+	const op = "warehouse.Repository.GetWarehouseSchemas"
+
+	// read a file into a string and return it
+	schemas, err := os.ReadFile("internal/warehouse/schemas.sql")
```
johanbrandhorst: 🤔 this seems to be doing the same thing as the handler in `mcp.go`? I guess this is the place we should be embedding the file?

dkanney: Ah, forgot to remove this once we switched from querying the DB to reading from file.

Removed :fire:

Makefile
dkanney: Moved this `pg_dump` call to a script `/scripts/dump_db_schema.sh`

Following the pattern in `/scripts/schema-diff.sh`, the script dies if `pg_dump` isn't installed:

```sh
die() {
    echo "$@"
    exit 255
}

which pg_dump &> /dev/null || die "pg_dump must be installed"
```

johanbrandhorst: Thanks!

scripts/dump_db_schema.sh
```diff
@@ -0,0 +1,50 @@
...
+        if [[ $c -ge $max ]]; then
+            docker logs boundary-sql-tests
+            make -C internal/db/sqltest clean
+            die "timeout waiting for database, likely an error in a migration"
+        fi
+        sleep 1
+    done
+
+    echo "dumping tables beginning with 'wh_' to ${dump}"
+    PGPASSWORD=boundary pg_dump -h "${DB_HOST}" -U boundary --data-only --table="wh_*" > "${dump}"
```
johanbrandhorst: Maybe update this to be the opposite (not data-only) 😁

scripts/dump_db_schema.sh
```diff
@@ -0,0 +1,50 @@
...
+
+    make -C internal/db/sqltest clean
+}
+
+# rm -rf "${tmp_dir}"
```
johanbrandhorst:
```suggestion
```

johanbrandhorst (changes requested):

scripts/dump_db_schema.sh
johanbrandhorst: I think we can just use `psql -c \d wh_* > schema.txt` right?

PR #5952, opened 2025-07-24
[Do Not Merge]: Cassandra Connect Support

RyanDerr:
## Description
- Create `connect cassandra` command
   - Update connect for cassandra to support `keyspaces` 
 - Create enos tests for validation with offical cassandra docker image
 - Update website and docs for the new feature with limitations based on native cassandra authentication plugins.


testing/internal/e2e/infra/docker.go
```diff
@@ -26,6 +27,14 @@ type Container struct {
 	UriNetwork   string
 }
 
+// CassandraConfig stores configuration details for the Cassandra container
+type CassandraConfig struct {
```
bgajjala8, line 31: For this struct, would it be better to create a basic DB config struct, perhaps named `ContainerDBConfig`, with fields for `user`, `password`, and `dbname`? This could then be extended by creating a `CassandraConfig` struct which would embed `ContainerDBConfig` and include a `keyspace` field. This would also require me to refactor the existing DB code to use the ContainerDBConfig. Not sure what the best way was to write this...

bgajjala8 (commented):
I wanted to get some advice on the setup for Cassandra, as the container behavior is different than the other DB containers. We need to do some `sed` magic and run some `exec` commands to return a container that is valid to test.

With `MySQL`, we're able to pass in environment variables to create the user, password, and database. Unfortunately, Cassandra doesn't have a native Docker option (or direct environment variable support) to startup with a keyspace and user. In addition, Cassandra starts up with AllowAllAuthenticator by default, and we need to adjust `cassandra.yaml` (inside the container) to enable password authentication (which requires a container restart). After that, we're able to create a new user using CQL commands.

testing/internal/e2e/infra/docker.go
bgajjala8: For this struct, would it be better to create a basic DB config struct, perhaps named `ContainerDBConfig`, with fields for `user`, `password`, and `dbname`? This could then be extended by creating a `CassandraConfig` struct which would embed ContainerDBConfig and include a `keyspace` field. This would also require me to refactor the existing DB code to use the `ContainerDBConfig`. Not sure what the best way was to write this...

website/content/docs/commands/connect/cassandra.mdx
```diff
@@ -0,0 +1,69 @@
...
+> **Note:** Currently, only configurations using `PasswordAuthenticator` are supported.
+
+@include 'cmd-connect-env-vars.mdx'
+
+
+## Examples
+
+The following example shows how to connect to a target with the ID `ttcp_eTcZMueUYv` using a Cassandra helper:
+
+``shell-session
```
moduli: nit: Missing tick here?
```suggestion
```shell-session
```

testing/internal/e2e/infra/docker.go
johanbrandhorst: Nah this is plenty simple as it is, I think, no need to DRY it up IMO. Looks like we could unexport it though? Is it used outside of the package?

moduli (commented):
Unfortunately, I can't do a super thorough review right now, but a cursory look at the e2e test changes look good. Cool to see the e2e tests updated!

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ Canonical reference for changes, improvements, and bugfixes for Boundary.
   connection parameters and credentials.
 * Adds support to parse User-Agent headers and emit them in telemetry events
   ([PR](https://github.com/hashicorp/boundary/pull/5645)).
+* cli: Added `boundary connect cassandra` command for connecting to Cassandra targets.
+  This new helper command allows users to authorize sessions against Cassandra 
+  targets and automatically invoke a Cassandra client with the appropriate 
+  connection parameters and credentials. Currently only supports `PasswordAuthenticator`.
```
johanbrandhorst: What is `PasswordAuthenticator`? Is it a Cassandra thing? I might just drop it and say
```suggestion
  connection parameters and credentials. Currently only username/password credentials are automatically attached.
```

johanbrandhorst (commented):
This looks great, thank you! Some small questions

CHANGELOG.md
RyanDerr: Yep it's one of the two types of authentication that Cassandra supports out of the box.
https://cassandra.apache.org/doc/4.1/cassandra/operating/security.html#authentication

RyanDerr: Additionally I don't believe that `CassandraConfig` is used outside the `infra` package. I'll go ahead and update this to not export it.

johanbrandhorst (approved):
Thanks, this is very cool! I can't tell from the enos/e2e logs if it actually ran your new test, but I assume so? As long as you can verify that this test runs and works, LGTM!

RyanDerr:
> Thanks, this is very cool! I can't tell from the enos/e2e logs if it actually ran your new test, but I assume so? As long as you can verify that this test runs and works, LGTM!

Yeah I don't believe we can directly see the output of the base logs for enos in the GH action, though I was able to review the successful run of the tests from the artifact output [here](https://github.com/hashicorp/boundary/actions/runs/16530304691/artifacts/3618458514).
image

Needed to update a typo as a result of looking at said output 😅 .

johanbrandhorst (approved):
Great!

PR #5976, opened 2025-08-12
chore(e2e): Add module for rdp member server

moduli:
## Description
This PR adds a terraform module to create an RDP member server. This member server is added to the domain of the domain controller. 

While doing so, this PR also renames one of the existing modules to be more explicit (`rdp_target_server` > `rdp_domain_controller`) now that we're a little more intelligent about RDP setups. 

```
enos scenario launch e2e_aws_windows builder:local client:win11 rdp_server:2019
```

## PCI review checklist


enos/modules/aws_rdp_member_server/main.tf
```diff
@@ -0,0 +1,152 @@
...
+set-item wsman:localhost\client\trustedhosts *.${var.active_directory_domain} -Force
+$s = New-PSSession -ComputerName $server -Credential $credential
+Invoke-Command -Session $s -ScriptBlock { $server = Resolve-DnsName -Name _ldap._tcp.dc._msdcs.${var.active_directory_domain} -Type SRV | Where-Object {$_.Type -eq "A"} | Select -ExpandProperty Name }
+Invoke-Command -Session $s -ScriptBlock { New-ADOrganizationalUnit -Name "RDP Member Servers" -Path "DC=${local.domain_sld},DC=${local.domain_tld}" -Server $server }
+Invoke-Command -Session $s -ScriptBlock { New-GPO -Name "RDP Settings 01" }
+Invoke-Command -Session $s -ScriptBlock { $GPOGuid = Get-gpo -name "RDP Settings 01" | Select -ExpandProperty Id }
+Invoke-Command -Session $s -ScriptBlock { Set-GPRegistryValue -Guid $GPOGuid  -Key "HKEY_LOCAL_MACHINE\SOFTWARE\Policies\Microsoft\Windows NT\Terminal Services" -ValueName "fDenyTSConnections" -Value 0 -Type DWord }
+Invoke-Command -Session $s -ScriptBlock { New-GPLink -Guid $GPOGuid -Target "ou=RDP Member Servers,DC=${local.domain_sld},DC=${local.domain_tld}" -LinkEnabled Yes -Enforced Yes }
+Remove-PSSession $s
```
dillanb-hashi: We should move these commands to the domain controller otherwise the member server will be a dependency for adding other member servers to the domain (like the worker and license server)

enos/modules/aws_rdp_member_server/main.tf
```diff
@@ -0,0 +1,152 @@
...
+    delete_on_termination = "true"
+    encrypted             = true
+  }
+
+  user_data_replace_on_change = true
+
+  user_data = <<EOF
+                <powershell>
+                  # Set up SSH so we can remotely manage the instance
+                  ## Install OpenSSH Server and Client
```
dillanb-hashi, line 65: Do we need ssh on the member server?

enos/modules/aws_rdp_member_server/main.tf
```diff
@@ -0,0 +1,152 @@
...
+
+resource "time_sleep" "wait_2_minutes" {
+  depends_on      = [aws_instance.member_server]
+  create_duration = "2m"
+}
+
+# wait for the SSH service to be available on the instance. We specifically use
+# BatchMode=Yes to prevent SSH from prompting for a password to ensure that we
+# can just SSH using the private key
+resource "enos_local_exec" "wait_for_ssh" {
```
dillanb-hashi, line 156: Shouldn't need this if we're not scping over any files.

enos/modules/aws_rdp_domain_controller/outputs.tf
```diff
@@ -0,0 +1,48 @@
...
+  description = "This is the decrypted administrator password for the EC2 instance"
+  value       = nonsensitive(local.password)
+}
+
+output "ssh_private_key" {
+  description = "Private key to ssh into the windows client"
+  value       = abspath(local_sensitive_file.private_key.filename)
+}
+
+output "security_group_id_list" {
```
dillanb-hashi, line 35: Security group for the domain controller contains some ports (53, 135, 389 etc) that aren't required on the member server. Not much of a security risk since this is just a testing environment but something to maybe think if we want to apply to all the member servers as well.

enos/modules/aws_rdp_domain_controller/main.tf
```diff
@@ -164,6 +258,8 @@ resource "aws_instance" "rdp_target" {
                 <powershell>
                   $password = ConvertTo-SecureString ${random_string.DSRMPassword.result} -AsPlainText -Force
                   Add-WindowsFeature -name ad-domain-services -IncludeManagementTools
+
+                  # causes the instance to reboot
                   Install-ADDSForest -CreateDnsDelegation:$false -DomainMode Win2012R2 -DomainName ${var.active_directory_domain} -DomainNetbiosName ${var.active_directory_netbios_name} -ForestMode Win2012R2 -InstallDns:$true -SafeModeAdministratorPassword $password -Force:$true
```
dillanb-hashi: I made a change in the -Domain mode to 7 (win2016), I believe it was because server 2025 didn't work with win2012R2. I think we're only going as old as win2016 in our testing so we should be fine to bump up this to "7" for domain mode and forestmode.

enos/modules/aws_rdp_member_server/main.tf
moduli: Yea. That was the only way I could think of to get the domain hostname/computer name (`resource "enos_local_exec" "get_hostname" {`), which we will need for Kerberos auth.

The alternative would be to rename the computer based on a variable we pass in, but...
1- you need to rename the computer before adding it to the domain (to get the name right)
2- renaming it requires a reboot before you add it to the domain, which added extra complexity

enos/modules/aws_rdp_member_server/main.tf
moduli: See comment about getting the hostname

enos/modules/aws_rdp_domain_controller/main.tf
moduli: Sure. I'll add that change to this PR

enos/modules/aws_rdp_domain_controller/outputs.tf
moduli: Good point. I did wonder that and figured it was simpler to reuse it, but I think I'll update it to use its own policy. 

EDIT: We'll leave things as-is since it's working. Can clean this up later if needed. Shouldn't really impact anything.

enos/modules/aws_rdp_member_server/main.tf
moduli: Per our discussion, we are modifying the script to not need to create this Active Directory Organizational Unit and instead, just adding the computer to the domain.

PR #6002, opened 2025-08-27
Bump up the error code for InvalidListToken such that it does not conflict with the error code Paused

RyanDerr:
## Description
Increment up the `InvalidListToken` error code such that it does not conflict with the existing `Paused` error code. 

## PCI review checklist

### DIFF
--- a/internal/errors/code.go
+++ b/internal/errors/code.go
@@ -68,7 +68,7 @@ const (
 	ChecksumMismatch         = 135 // ChecksumMismatch represents an error when a checksum is mismatched
 	Paused                   = 136 // Paused represents an error when an operation cannot be completed because the thing being operated on is paused
 
-	InvalidListToken Code = 136 // InvalidListToken represents an error where the provided list token is invalid
+	InvalidListToken Code = 137 // InvalidListToken represents an error where the provided list token is invalid
 
 	AuthAttemptExpired Code = 198 // AuthAttemptExpired represents an expired authentication attempt
 	AuthMethodInactive Code = 199 // AuthMethodInactive represents an error that means the auth method is not active.


internal/errors/code.go
```diff
@@ -68,7 +68,7 @@ const (
 	ChecksumMismatch         = 135 // ChecksumMismatch represents an error when a checksum is mismatched
 	Paused                   = 136 // Paused represents an error when an operation cannot be completed because the thing being operated on is paused
 
-	InvalidListToken Code = 136 // InvalidListToken represents an error where the provided list token is invalid
+	InvalidListToken Code = 137 // InvalidListToken represents an error where the provided list token is invalid
```
louisruch: Even though this is lower on the list it was actually added first, it also has much more use and can potentially be exposed to end users. So I think it might be better to change the Paused error and move the ordering around a bit.

I assume because of the white space when Paused was added they just didnt see 136 https://github.com/hashicorp/boundary/commit/3f94c900f002788a4a8c54ac6f1102464656f567

louisruch (changes requested):

internal/errors/code.go
RyanDerr: Not sure if `Code` definition is needed to be explicitly set on the `InvalidListToken` error. I went ahead I keep it the same as the previous for safety after reordering.

louisruch: Oh that's interesting we actually probably wanna go ahead and add `Code` to all of those

louisruch: Ugh what is going on here, so there are a bunch of new codes that do not have messages set either

louisruch:
@ddebko Mind giving this a review since I added to it

RyanDerr:
@louisruch looks like it may require your review as well.

PR #6026, opened 2025-09-04
fix(rdp): delete session proxy cert on project scope deletion

irenarindos:
## Description
Added a trigger to delete the session_proxy_certificate(s) that would be affected by the deletion of a project.
This fixes a bug found where the restrict constraint on the session_proxy_certificate kms data key prevented project deletion.

## PCI review checklist

### DIFF
--- a/internal/db/schema/migrations/oss/postgres/98/06_target_credential_injection_certificates.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/98/06_target_credential_injection_certificates.up.sql
@@ -137,4 +137,22 @@ create table session_proxy_certificate(
 comment on table session_proxy_certificate is
   'session_proxy_certificate is a table where each row maps a certificate to a session id.';
 
+-- delete_session_proxy_certificate_with_null_fk is intended to be a before update trigger that
+-- deletes the session_proxy_certificate entry for a session if the project_id of the session is being set to null.
+-- If the project_id is null, this indicates that the kms key used to encrypt the session_proxy_certificate is being deleted
+-- and therefore the session_proxy_certificate entry must also be deleted.
+create or replace function delete_session_proxy_certificate_with_null_fk() returns trigger
+as $$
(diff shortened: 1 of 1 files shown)


johanbrandhorst (dismissed):
Nice, how hard would it be to add a test for this?

internal/db/schema/migrations/oss/postgres/98/06_target_credential_injection_certificates.up.sql
```diff
@@ -137,4 +137,22 @@ create table session_proxy_certificate(
...
 
+-- delete_session_proxy_certificate_with_null_fk is intended to be a before update trigger that
+-- deletes the session_proxy_certificate entry for a session if the project_id of the session is being set to null.
+-- If the project_id is null, this indicates that the kms key used to encrypt the session_proxy_certificate is being deleted
+-- and therefore the session_proxy_certificate entry must also be deleted.
+create or replace function delete_session_proxy_certificate_with_null_fk() returns trigger
+as $$
+begin
+  case
+    when new.project_id is null then
```
louisruch, line 148: This is a slight departure from how the other delete works which is based on the `canceling` or `terminated` states.  But that being said, I think this is fine and might be more clear because it depends directly on the before trigger on the `session` rather than the after trigger on the `session_state`.

Only issue is if setting a `project_id` to `null` does not cancel a session in the future so a test would be nice as Johan pointed out

louisruch (dismissed):
Thanks for this, have we manually validated this does indeed resolve the issue?

internal/db/schema/migrations/oss/postgres/98/06_target_credential_injection_certificates.up.sql
irenarindos: @louisruch I've manually validated it resolves the issue, I'll look into tests

PR #6038, opened 2025-09-08
target(addr): add validation for ipv4, ipv6, and dns addr

irenarindos:
## Description
Target address validation was only checking length and allowed DNS names with invalid symbols, leading to a bug.
This PR adds validation to ensure the address is a valid IP or DNS name

## PCI review checklist

### DIFF
--- a/internal/target/address.go
+++ b/internal/target/address.go
@@ -5,6 +5,8 @@ package target
 
 import (
 	"context"
+	"net"
+	"regexp"
 	"strings"
 
 	"github.com/hashicorp/boundary/internal/db"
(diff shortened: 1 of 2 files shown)


internal/target/address.go
```diff
@@ -31,6 +33,52 @@ var (
...
+		}
+		if strings.HasPrefix(label, "-") || strings.HasSuffix(label, "-") {
+			return false
+		}
+		if !labelRegex.MatchString(label) {
+			return false
+		}
+		// Last label cannot be all numeric
+		if i == len(labels)-1 {
+			if numbersRegex.MatchString(label) {
```
johanbrandhorst, line 66: Perhaps a bit cheeky, but it might be faster/easier/nicer to just `strconv.Itoa` and see if it _doesn't_ fail?

irenarindos: I think I'd rather be explicit here, just because I've seen some unexpected results relying on a conversion check like that :x

johanbrandhorst: OK!

johanbrandhorst (dismissed):

mikemountain (approved):

PR #6054, opened 2025-09-15
Closes any session created by `boundary connect` after command exit

kheina:
This pr fixes an issue where failing to create the client proxy within `boundary connect ...` would leave pending sessions within boundary

## Description
1. parses the listener address earlier to creating sessions without an easy way to close them
2. automatically closes any session created by `boundary connect` using the client proxy or creating a new session client after the command has exited (successfully or unsuccessfully)

## PCI review checklist

### DIFF
--- a/api/proxy/proxy.go
+++ b/api/proxy/proxy.go
@@ -350,12 +350,20 @@ func (p *ClientProxy) Start(opt ...Option) (retErr error) {
 		return nil
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), opts.withSessionTeardownTimeout)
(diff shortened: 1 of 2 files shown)


kheina:
> LGTM! Is possible to write a test for this?

at least according to the comments in our current bats test setup for `boundary connect` it looks like because we run a command directly, and therefore automatically create a connection via the client proxy, the session is already cancelled immediately by the existing logic within client proxy. so without creating a new bats test suite to specifically occupy a port and then try to connect using that port and cause a failure, we wouldn't be able to test this

https://github.com/hashicorp/boundary/blob/311525d1674f79ef10e72993d26927094f18a85a/internal/tests/cli/boundary/_connect.bash

PR #6055, opened 2025-09-16
fix(target): Skip fetching proxy server cert

AprilMay0:
## Description

Proxy server certificates are not necessary for or used with CE Targets. As a follow up to https://github.com/hashicorp/boundary/pull/6050 this PR will update LookupTargetForSessionAuthorization to use either a separate function to search for proxy server certificates. In CE this function is empty and returns `nil, nil`. 

The goal of this change is to improve performance by avoiding unnecessary trips to the database for targets that do not have proxy server certificates.

## PCI review checklist

### DIFF
--- a/internal/target/repository.go
+++ b/internal/target/repository.go
@@ -24,6 +24,7 @@ import (
 	"github.com/hashicorp/boundary/internal/types/scope"
 	"github.com/hashicorp/boundary/internal/util"
 	"github.com/hashicorp/go-dbw"
+	wrapping "github.com/hashicorp/go-kms-wrapping/v2"
 )
 
 // RepositoryFactory enables `target.Repository` object instantiation,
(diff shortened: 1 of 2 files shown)


internal/target/repository.go
```diff
@@ -51,6 +52,12 @@ type Repository struct {
 	permissions []perms.Permission
 }
 
+// getTargetProxyServerCertificateFn can be overridden for testing or extension purposes.
+// By default, it returns nil, nil.
```
irenarindos: I'm wondering if we should update this comment to mention that TCP targets do not currently use proxy server certificates, so that if they do in the future it'd be clear that this needs to change.

irenarindos (dismissed):
Thanks for this! Just one comment

internal/target/repository.go
AprilMay0: Great point! Updated this and other comment to specifically mention that TCP targets do not currently use proxy server certificates.

AprilMay0:
> Thanks! Do we intend to merge this to 0.20.0?

I am going to ask in the release channel.

PR #6125, opened 2025-10-07
Password credential migration script

bgajjala8:
## Description
The PR updates the migration changes to include updated to current tables.
 
Associated Jira [Ticket](https://hashicorp.atlassian.net/browse/ICU-17728)

### DIFF
--- a/internal/db/schema/migrations/oss/postgres/99/01_credential_static_password_credential.up.sql
+++ b/internal/db/schema/migrations/oss/postgres/99/01_credential_static_password_credential.up.sql
@@ -115,4 +115,98 @@ begin;
 
   create index credential_static_password_credential_deleted_delete_time_idx on credential_static_password_credential_deleted (delete_time);
 
+  create view credential_static_password_credential_hst_aggregate as
+    select
+      rsc.recording_id,
+      spc.public_id,
+      spc.name,
+      spc.description,
+      spc.password_hmac,
+      css.public_id as store_public_id,
+      css.project_id as store_project_id,
+      css.name as store_name,
+      css.description as store_description,
+      string_agg(distinct rsc.credential_purpose, '|') as purposes
+    from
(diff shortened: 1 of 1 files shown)


internal/db/schema/migrations/oss/postgres/99/01_credential_static_password_credential.up.sql
```diff
@@ -115,4 +115,102 @@ begin;
 
   create index credential_static_password_credential_deleted_delete_time_idx on credential_static_password_credential_deleted (delete_time);
 
+create view credential_static_password_credential_hst_agg as
```
louisruch: We should indent this line as well and then all subsiquent lines further

bgajjala8: Updated, thanks for the catch

internal/db/schema/migrations/oss/postgres/99/01_credential_static_password_credential.up.sql
```diff
@@ -115,4 +115,98 @@ begin;
 
   create index credential_static_password_credential_deleted_delete_time_idx on credential_static_password_credential_deleted (delete_time);
 
+  create view credential_static_password_credential_hst_agg as
```
laero: this should be named `credential_static_password_credential_hst_aggregate` to follow convention

`credential_static_username_password_domain_credential_hst_agg` has its name shortened because it would exceed the 63 char limit. the `hst_aggregate` table names for other credential types spell out `aggregate`

bgajjala8: Ah gotcha!

laero:
re: new table and view updates:
i'll make a separate PR for these changes

PR #6140, opened 2025-10-10
feat(credential/vault): Implement repository vault password credential

laero:
## Description
Implement repository support for `Password` credential.

internal/proto/controller/storage/credential/vault/store/v1/vault.proto
* Add `Password` support to vault credential library

internal/credential/vault/internal/password/...
* Contains package that extract `Password` from a vault secret. Used by `private_library.go`

internal/credential/vault/private_library.go
* `retrieveCredential(...)` is used to retrieve a credential from Vault. Added a case to support `Password`

internal/credential/vault/mapping_overriders.go
* Defines PasswordOverride type and methods used by repository logic in `repository_credential_library.go`

internal/credential/vault/repository_credential_library.go
* Add case to support Password in CredentialLibrary Lookup / Update. This will expose `Password` to the API

## PCI review checklist

### DIFF
--- a/internal/credential/credential.go
+++ b/internal/credential/credential.go
@@ -143,6 +143,13 @@ type UsernamePasswordDomain interface {
 	Domain() string
 }
 
+// PasswordCredential is a credential containing a username and a password.
+// Does not follow naming convention to avoid conflict with existing Password type.
+type PasswordCredential interface {
+	Credential
+	Password() Password
+}
+
 // SshPrivateKey is a credential containing a username an SSH private key and
 // an optional private key passphrase.
 type SshPrivateKey interface {
--- a/internal/credential/vault/mapping_overriders_test.go
+++ b/internal/credential/vault/mapping_overriders_test.go
@@ -46,6 +46,11 @@ func TestValidMappingOverrides(t *testing.T) {
 			want: true,
 		},
 		{
+			m:    nil,
+			ct:   globals.PasswordCredentialType,
+			want: true,
+		},
+		{
 			m:    unknownMapper(1),
 			ct:   globals.UnspecifiedCredentialType,
 			want: false,
@@ -58,6 +63,11 @@ func TestValidMappingOverrides(t *testing.T) {
 		{
 			m:    unknownMapper(1),
 			ct:   globals.UsernamePasswordDomainCredentialType,
+			want: false,
+		},
+		{
+			m:    unknownMapper(1),
+			ct:   globals.PasswordCredentialType,
 			want: false,
 		},
 		{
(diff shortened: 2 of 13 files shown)


internal/credential/credential.go
```diff
@@ -143,6 +143,13 @@ type UsernamePasswordDomain interface {
 	Domain() string
 }
 
+// PasswordCredential is a credential containing a username and a password.
+// Does not follow naming convention to avoid conflict with existing Password type.
+type PasswordCredential interface {
```
laero, line 148: this breaks our naming convention - it should be called Password but that already exists as a type in this file.

@louisruch  i am wondering if we should consider renaming the types here, rather than leaving this as a special case. my suggestion is to append `Credential` to the end of all the other credential types.

internal/credential/vault/mapping_overriders_test.go
```diff
@@ -80,6 +90,11 @@ func TestValidMappingOverrides(t *testing.T) {
 			ct:   globals.UsernamePasswordDomainCredentialType,
 			want: true,
 		},
+		{
+			m:    allocPasswordOverride(),
+			ct:   globals.PasswordCredentialType,
+			want: true,
+		},
```
bgajjala8, line 102: (Question) Do we need to test for a `globals.UnspecifiedCredentialType` ?

laero: yes we do, thanks for catching that

internal/credential/vault/repository_credential_library_test.go
```diff
@@ -1915,6 +2115,7 @@ func TestRepository_LookupCredentialLibrary(t *testing.T) {
 		}
 	}
 
+	// TODO: start looking below here tomorrow
```
bgajjala8: Might have missed removing this

bgajjala8:
Outside of the missed comment clean up, it LG2M. However, I would like to wait for @louisruch for guidance on your comment about the naming convention for `PasswordCredential`

internal/credential/vault/repository_credential_library_test.go
laero: oh yeah thanks for catching that 🤦

internal/credential/credential.go
louisruch: I think leave this as a special case for now and we can review in the llb->main merge

laero: another note on existing convention: bharath and i were looking over the static credentials code and saw it appends 'Credential' to the credential type names (such as `UsernamePasswordCredential`: https://github.com/hashicorp/boundary/blob/fb7a42671b1e45ea6d5b0130f56962c5d677d448/internal/credential/static/usernamepassword_credential.go#L25

laero: will do, thanks!

PR #6153, opened 2025-10-14
Build source override

dillanb-hashi:
## Description
Change the build script to allow the source to be modified. Used for github automation pipelines that want to test the boundary-ui repo code but still build for enterprise. 

## PCI review checklist


enos/modules/build_local/main.tf
```diff
@@ -43,6 +47,7 @@ resource "enos_local_exec" "build" {
     "BINARY_NAME"   = var.binary_name
     "BUILD_TARGET"  = var.build_target
     "EDITION"       = var.edition
+    "UI_BUILD_OVERRIDE" = var.ui_build_override
```
moduli, line 51: you will want to run to address the `enos fmt` ci check (align the `=`)
```
cd enos
make fmt
```

enos/enos-variables.hcl
```diff
@@ -271,3 +271,9 @@ variable "windows_instance_type" {
   type        = string
   default     = "m7i-flex.xlarge"
 }
+
+variable "ui_build_override" {
+  description = "Override for build for UI automation"
```
moduli, line 276: Suggestion to put in the possible values here (i.e. "oss" or "enterprise")

enos/modules/build_local/main.tf
```diff
@@ -33,6 +33,10 @@ variable "goos" {
   default = "linux"
 }
 
+variable "ui_build_override" {
+  default = null
```
moduli: suggestion to add the same desription to `build_local` and `build_boundary_docker_local`

scripts/build-ui.sh
```diff
@@ -13,7 +13,13 @@ if [ -z "$UI_VERSION_FILE" ]; then
 	echo "Must set UI_VERSION_FILE"; exit 1
 fi
 
-UI_EDITION=$(make --no-print-directory edition)
+UI_EDITION_SRC=$(make --no-print-directory edition)
+
+if [ -n "$UI_EDITION_SRC_OVERRIDE" ]; then
```
moduli: may want to put a comment here so that there's context for why this was added

example:
```
# allow override for the source of ui files
# used by end-to-end tests to build enterprise edition from community files
```

scripts/build-ui.sh
```diff
@@ -13,7 +13,16 @@ if [ -z "$UI_VERSION_FILE" ]; then
 	echo "Must set UI_VERSION_FILE"; exit 1
 fi
 
-UI_EDITION=$(make --no-print-directory edition)
+UI_EDITION_SRC=$(make --no-print-directory edition)
```
ZedLi: I think I got a little confused by the naming, just to make sure I understand, we want to build both CE and ENT versions of the the UI from the community edition of `boundary-ui`?

If that's accurate, I think I would have expected this variable to still be called `UI_EDITION` (as in the version of the UI that's being built) and `UI_EDITION_SRC` being where the actual files are being pulled from which means the variables would be swapped since it seems like `UI_EDITION_SRC_OVERRIDE` is actually overriding `UI_EDITION` as it is. 

Would `UI_SRC` maybe even be more obvious since it's really which repo we're building from that we want to override and not the edition?

dillanb-hashi: Yes that makes sense. i've changed the variable names

scripts/build-ui.sh
```diff
@@ -14,8 +14,17 @@ if [ -z "$UI_VERSION_FILE" ]; then
 fi
 
 UI_EDITION=$(make --no-print-directory edition)
+# allow override for the source of ui files
+# used by end-to-end tests to build enterprise edition from community files
+# UI_EDITION_SRC_OVERRIDE will override which edition is pulled
+# but UI_EDITION_SRC will still determine the build version
+if [ -n "$UI_EDITION_SRC_OVERRIDE" ]; then
```
moduli: Does it make more sense to call this `$UI_SRC_OVERRIDE`? (we will have to update the enos modules to use this if we do want to change it). Thoughts?

dillanb-hashi: Yes that probably makes more sense with what we're calling the other vairables

scripts/build-ui.sh
```diff
@@ -14,8 +14,17 @@ if [ -z "$UI_VERSION_FILE" ]; then
...
+# UI_SRC_OVERRIDE will override which edition is pulled
+# but UI_EDITION_SRC will still determine the build version
+if [ -n "$UI_SRC_OVERRIDE" ]; then
+  UI_SRC=$UI_SRC_OVERRIDE
+else
+  UI_SRC=$UI_EDITION
+fi
 
-if [ "$UI_EDITION" == "oss" ]; then
+if [ $UI_SRC == "oss" ]; then
```
moduli, line 27: Realized, if we're doing this, are we going to have to change all of the usages of `UI_EDITION` below (except for `EDITION=${UI_EDITION} pnpm build`) to use `UI_SRC`?

dillanb-hashi: All the usages are just log messages but I think they should be changed to UI_SRC as well.

moduli: I don't think the first one is? It's using it to get the artifact id?

ZedLi (dismissed):
Looks good!

PR #6161, opened 2025-10-15
feat(handlers): Implement CRUDL for vault generic Password credential

laero:
## Description
Implements Password credential type in API handlers.

## PCI review checklist

### DIFF
--- a/internal/credential/credential.go
+++ b/internal/credential/credential.go
@@ -143,9 +143,9 @@ type UsernamePasswordDomain interface {
 	Domain() string
 }
 
-// PasswordCredential is a credential containing a username and a password.
+// PasswordOnly is a credential containing a password.
 // Does not follow naming convention to avoid conflict with existing Password type.
-type PasswordCredential interface {
+type PasswordOnly interface {
 	Credential
 	Password() Password
 }
--- a/internal/daemon/controller/handlers/targets/tcp/target_service_test.go
+++ b/internal/daemon/controller/handlers/targets/tcp/target_service_test.go
@@ -4170,7 +4170,7 @@ func TestAuthorizeSessionTypedCredentials(t *testing.T) {
 				HttpMethod: wrapperspb.String("GET"),
 			},
 		},
-		CredentialType: "username_password",
+		CredentialType: string(globals.UsernamePasswordCredentialType),
 	}})
 	require.NoError(t, err)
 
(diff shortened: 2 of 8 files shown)


internal/credential/credential.go
```diff
@@ -143,9 +143,9 @@ type UsernamePasswordDomain interface {
 	Domain() string
 }
 
-// PasswordCredential is a credential containing a username and a password.
+// PasswordOnly is a credential containing a username and a password.
```
bgajjala8: I might have missed some context here.... any reason we are moving this to `PasswordOnly`

bgajjala8 (approved):
LG2M! Just a small query

internal/credential/credential.go
laero: there was a discussion in the proj channel, i will send you the details. tldr: we want to avoid using the package name in the types within that package

bgajjala8: Got it! Thanks 👍

internal/daemon/controller/handlers/targets/tcp/target_service_test.go
```diff
@@ -4211,6 +4211,44 @@ func TestAuthorizeSessionTypedCredentials(t *testing.T) {
...
+		Name:              wrapperspb.String("Password Library"),
+		Description:       wrapperspb.String("Password Library Description"),
+		Type:              vault.GenericLibrarySubtype.String(),
+		Attrs: &credlibpb.CredentialLibrary_VaultCredentialLibraryAttributes{
+			VaultCredentialLibraryAttributes: &credlibpb.VaultCredentialLibraryAttributes{
+				Path:       wrapperspb.String(path.Join("secret", "data", "default-password")),
+				HttpMethod: wrapperspb.String("GET"),
+			},
+		},
+		CredentialType: "password",
```
bgajjala8: (nit) Would it be a good idea to use globals const value here - `globals.PasswordCredentialType`?

bgajjala8 (commented):
LGTM! Just a small nit

bgajjala8 (approved):

internal/daemon/controller/handlers/targets/tcp/target_service_test.go
laero: yes, i'll fix that for all of the tests in this file as well

internal/credential/credential.go
louisruch: Oh I see the doc string calls out `username` still - missed that in the initial review.

laero: thanks for catching that, will get it fixed

PR #6181, opened 2025-10-23
feat(handlers): CRUDL support for static password credential

bgajjala8:
## Description
This PR adds API handling logic for the new password credentials type.

* Adding password type as an input
* Updated credential service to handle the new password credentials type

[ICU-17726](https://hashicorp.atlassian.net/browse/ICU-17726)

[ICU-17726]: https://hashicorp.atlassian.net/browse/ICU-17726?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hugoghx (commented):
Looking good! Unsure if this is in another PR but just a heads up that we'll need to add support for this credential type in the target handlers for Add/Set/Remove Credential Sources (validate functions).

bgajjala8:
> Looking good! Unsure if this is in another PR but just a heads up that we'll need to add support for this credential type in the target handlers for Add/Set/Remove Credential Sources (validate functions).

Thanks! I went ahead an added validate functions to this PR

internal/daemon/controller/handlers/credentials/credential_service_test.go
```diff
@@ -87,6 +87,24 @@ func staticPasswordCredentialToProto(cred *static.UsernamePasswordCredential, pr
 	}
 }
 
+func staticPasswordOnlyCredentialToProto(cred *static.PasswordCredential, prj *iam.Scope, hmac string) *pb.Credential {
```
laero: i think we should rename these two functions like so for clarity and consistency:
* rename this function to `staticPasswordCredentialToProto`
* rename the pre-existing function on line 71 `staticPasswordCredentialToProto` to `staticUsernamePasswordCredentialToProto` (it is actually creating a credential that contains `UsernamePassword`)

internal/daemon/controller/handlers/credentials/credential_service_test.go
```diff
@@ -1663,6 +1866,8 @@ func TestUpdate(t *testing.T) {
 				cred, cleanup = freshCredJson()
 			} else if strings.Contains(tc.name, "domainupd") {
 				cred, cleanup = freshCredUpd("user", "pass", "domain")
+			} else if strings.Contains(tc.name, "P") {
```
laero: i think something more verbose would help prevent accidents and improve readability here, if we're sticking with string comparisons like this. maybe something like `password-only`?

internal/daemon/controller/handlers/credentials/credential_service_test.go
laero, line 1: TestUpdate has a section for testing read-only fields, should we consider testing `Password` there as well? https://github.com/hashicorp/boundary/blob/7482589f38394f4bcb6be048d2c2f040f8510ec3/internal/daemon/controller/handlers/credentials/credential_service_test.go#L1953

internal/daemon/controller/handlers/credentials/credential_service_test.go
louisruch: +1

bgajjala8: Yep, good catch! Updated this one and the  `staticPasswordDomainCredentialToProto` to `staticUsernamePasswordCredentialToProto`

internal/daemon/controller/handlers/credentials/credential_service_test.go
bgajjala8: Make sense! Updated

internal/daemon/controller/handlers/credentials/credential_service_test.go
bgajjala8: Added password only cred. However, noticed test is missing for `UsernamePasswordDomain?` Would it be a good idea to add it through this PR?

hugoghx:
> Looking good! Unsure if this is in another PR but just a heads up that we'll need to add support for this credential type in the target handlers for Add/Set/Remove Credential Sources (validate functions).

I see we already have the password prefix there, however I noticed that this is also under the injected credentials (it was removed from add in this PR but not set and remove). My understanding is that this cred is brokered-only so we should remove it from all

bgajjala8:
> > Looking good! Unsure if this is in another PR but just a heads up that we'll need to add support for this credential type in the target handlers for Add/Set/Remove Credential Sources (validate functions).
> 
> I see we already have the password prefix there, however I noticed that this is also under the injected credentials (it was removed from add in this PR but not set and remove). My understanding is that this cred is brokered-only so we should remove it from all

Oh gotcha, missed those - updating (yes, this is brokered only)!

internal/daemon/controller/handlers/credentials/credential_service_test.go
laero: yup let's do that

bgajjala8: Sounds good - updated

PR #6187, opened 2025-10-23
docs: Explain storage capacity in concept topic

Dan-Heath:
## Description

From an internal Slack conversation: https://ibm-hashicorp.slack.com/archives/C09KVLZF1LM/p1761239420751909

We mention the `recording_storage_minimum_available_capacity` setting in the worker configuration topics, but it might be helpful for users if we also call out that setting as something they should think about before they configure session recording. This PR adds additional information about that setting to the Session recording concepts page in the same place where we recommend users consider how much storage space to allocate. It also adds a link to the section where we describe the storage states and how that setting impacts recording capabilities when the storage threshold is met.

[View the update in the preview deployment](https://boundary-gqjobj7qs-hashicorp.vercel.app/boundary/docs/session-recording#storage-considerations)

## PCI review checklist

### DIFF
--- a/website/content/docs/session-recording/index.mdx
+++ b/website/content/docs/session-recording/index.mdx
@@ -68,7 +68,7 @@ Determining how much storage you need to allocate on workers and the external st
 
 When you estimate worker storage requirements, consider the number of concurrent sessions that will be recorded on that worker. Boundary writes the BSR to the worker's local storage while the session is active, and then moves it to the remote storage bucket when the session is closed.
 
-You use the `recording_storage_minimum_available_capacity` setting to configure how much local disk storage space is available on the worker for the cache.
+You use the `recording_storage_minimum_available_capacity` setting to configure the minimum amount of storage space that is required for workers to perform session recording operations. If a worker is at or below the storage threshold, Boundary does not use the worker to record sessions or play back recordings.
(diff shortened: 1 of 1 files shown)


website/content/docs/session-recording/index.mdx
```diff
@@ -68,7 +68,12 @@ Determining how much storage you need to allocate on workers and the external st
 
 When you estimate worker storage requirements, consider the number of concurrent sessions that will be recorded on that worker. Boundary writes the BSR to the worker's local storage while the session is active, and then moves it to the remote storage bucket when the session is closed.
 
-When you estimate storage requirements for the external storage provider, consider your [storage policy](/boundary/docs/domain-model/storage-policy) and how long a BSR will be retained in the external storage bucket.
+You use the `recording_storage_minimum_available_capacity` setting to configure how much local disk storage space is available on the worker for the cache.
```
irenarindos: I would do something more along these lines to make it clearer what the impact of the value is... the way it is now can be read that sets aside the storage space to be used 
```suggestion
You use the `recording_storage_minimum_available_capacity` setting to configure the minimum amount of storage space that is required for workers to perform session recording operations. If a worker is at or below the storage threshold, Boundary does not use the worker to record sessions or play back recordings.
```

PR #6213, opened 2025-10-29
feat(cli): Add support for password credential in redis helper

laero:
## Description
* Parse password credential into a `UsernamePassword` for usage in brokering
* Reworked redis connect: we now prompt user for a password when no password is brokered
* Updated and cleaned up redis tests

## Sample CLI Usage
(most responses omitted for brevity)
1. `boundary scopes create -scope-id global`
2. `boundary scopes create -scope-id o_fJtwTPHZdM`
3. `boundary targets create tcp -name "test-target" -scope-id p_SdrMtBv6Zi -default-port 22 -address 127.0.0.1`
4. `boundary credential-stores create static -scope-id p_SdrMtBv6Zi`
5. `export TESTPASS="test-password"`
6. `boundary credentials create password -credential-store-id csst_POG2btczXC -password env://TESTPASS -name test-pcred`
```
Credential information:
  Created Time:          Thu, 30 Oct 2025 11:27:44 CDT
  Credential Store ID:   csst_POG2btczXC
  ID:                    credp_DlHnSSRBpu
  Name:                  test-pcred
  Type:                  password
  Updated Time:          Thu, 30 Oct 2025 11:27:44 CDT
  Version:               1

  Scope:
    ID:                  p_SdrMtBv6Zi
    Name:
    Parent Scope ID:     o_fJtwTPHZdM
    Type:                project

  Authorized Actions:
    update
    delete
    no-op
    read

  Attributes:
    Password HMAC:       eoqMVkU3PezYWtRJjMKGNOdi0xu0B1pCvT6aumdlU9o
```
7. `boundary targets add-credential-sources -id ttcp_tucp5A15x1 -brokered-credential-source credp_DlHnSSRBpu`
8. `boundary connect -target-id ttcp_tucp5A15x1`
```
Proxy listening information:
  Connection Limit:    -1
  Expiration:          Thu, 30 Oct 2025 19:29:52 CDT
  Port:                63121
  Protocol:            tcp
  Session ID:          s_5CBGrxXLJ7

  Credentials:
    Credential Source ID:   credp_DlHnSSRBpu
    Credential Source Name: test-pcred
    Credential Store ID:    csst_POG2btczXC
    Credential Store Type:  static
    Credential Type:        password
    Secret:
          password:   test-password
```


api/proxy/credentials.go
```diff
@@ -90,6 +91,18 @@ func ParseCredentials(creds []*targets.SessionCredential) (Credentials, error) {
...
 
+		case passwordCredentialType:
+			// Decode attributes from credential struct
+			if err := mapstructure.Decode(cred.Credential, &upCred); err != nil {
+				return Credentials{}, err
+			}
+
+			if upCred.Password != "" {
+				upCred.Raw = cred
+				out.UsernamePassword = append(out.UsernamePassword, upCred)
```
louisruch: So why are we creating a `UsernamePassword` credential when we get a password credential type?

api/proxy/credentials.go
```diff
@@ -90,6 +91,18 @@ func ParseCredentials(creds []*targets.SessionCredential) (Credentials, error) {
 				continue
 			}
 
+		case passwordCredentialType:
```
louisruch, line 111: Can we add some tests for this in `credentials_test.go`

louisruch (commented):
Thanks for adding e2e tests for this - great to see it coming together.
I left a comment around the what is happening `parseCredentials`

Also when working on add CLI comments I really love to see some sample CLI usage in the PR description as it clearly shows what the UX will be. Could you add CLI request and response for:
- Create a password credential
- Add it as a brokered credential source, and then should the response from `boundary connect` (not the redis connect as I don't want it consumed)

api/proxy/credentials.go
hugoghx: +1

laero: I did consider making a new `Password` struct here, but I went with the current implementation for these reasons:
* Conceptually `Password` is `UsernamePassword` without the username. As I understood it, we should attempt to broker `Password` wherever we can broker `UsernamePassword`
* Using `UsernamePassword` to represent other credential types isn't new behavior (see [this](https://github.com/hashicorp/boundary/blob/997c3cebad5c04a61a2d20d66e969cd7f7ebed90/api/proxy/credentials.go#L126)) but that only handles cases where the credential has both a username and a password (for example `UsernamePasswordDomain` would be decoded into `UsernamePassword` there). I chose to make a new case rather than alter the code there to be explicit with how we're handling `Password` (also doesn't change existing behavior for users who will be upgrading)
* If we add a new `Password` credential here instead of using `UsernamePassword`, we'd need to update each command we want to utilize it in. This would allow us to precisely control when a `Password` is brokered. However, if we want `Password` to be brokered under the same condition as `UsernamePassword` anyways and don't have special cases, it doesn't seem worth the overhead to implement `Password` in this manner. They just need to handle the empty `username` when dealing with `UsernamePassword`, which is already a basic requirement

Would love to hear your thoughts on this as I'm not 100% clear on which direction to take 😄

louisruch: Thanks for the detailed response - its good to see you put thought into this decision. Ill address each point below:

> Conceptually Password is UsernamePassword without the username. 

I get what you are saying here, but this is a slippery slope. A `UsernamePassword` credential should only be valid if it contains a non empty `username` and `password`. A `Password` credential should never contain a `username` field. One of the reason we decided to implement this `Password` credential type (instead of reusing `UsernamePassword`) was because we want the domain to be able to enforce these invariants. Now I get this is in the API proxy not the domain repositories - but conceptually this should still hold true.

> As I understood it, we should attempt to broker Password wherever we can broker UsernamePassword

I am not sure what you mean by this

> Using UsernamePassword to represent other credential types isn't new behavior 

That is a very different case, in that case the returned object did not explicitly have a `CredentialType` associated with the brokered credential. As the comment states `make a best effort attempt to parse the credential`. All we are trying to do here is allow a credential that was not explicitly typed to be consumed if it meets the criteria for that type of credential.

> If we add a new Password credential here instead of using UsernamePassword, we'd need to update each command we want to utilize it in

We should definitely be making a consensus decision about which commands we want to update to support just the password case. When we consume a credential and it does not work the command will fail, we should not just attempt a `Password` for all `UsernamePassword` connects. Going a bit further with this - our consume model attempts credentials in order - this means if you have two brokered credentials a `Password` and a `UsernamePassword` and both are treated as `UsernamePassword`, if the `Password` is first in the slice the helper functions will use that one each time which is probably not the behavior we want.

As a final note - while it may fell like more work to plumb through the `Password` credential type - it makes future developers much happier. It can become very confusing in a few years when a `UsernamePassword` credential is being used everywhere when the intent was to have a `Password` credential - being explicit in the code leads to more readable and maintainable code.

laero: Thanks for the feedback! 

EDIT: After some re-reads and more thought, I think my confusion stems from conflating the concepts of brokering a credential and consuming a credential.

> One of the reason we decided to implement this Password credential type (instead of reusing UsernamePassword) was because we want the domain to be able to enforce these invariants.

Thanks for the insight, I agree with this

> >    As I understood it, we should attempt to broker Password wherever we can broker UsernamePassword

> I am not sure what you mean by this

~~I'm voicing my assumption here that from a UX perspective we should attempt to best support `Password` brokering anywhere that we can also broker 'UsernamePassword`, because theoretically users can set up brokering any credential type to any target. And those two have a natural overlap (although as you stated they are not the same).~~

I misunderstood the difference between consuming and brokering, the initial statement doesn't really make any sense now that I've cleared up the misunderstanding.

> We should definitely be making a consensus decision about which commands we want to update to support just the password case. When we consume a credential and it does not work the command will fail, we should not just attempt a Password for all UsernamePassword connects.

I agree with this, but I think I'm still not clear on this part. There was a thread related to this, I will ping it in the channel to continue the discussion there.

> ...  if you have two brokered credentials a Password and a UsernamePassword and both are treated as UsernamePassword, if the Password is first in the slice the helper functions will use that one each time which is probably not the behavior we want.

Definitely an issue, it did cross my mind and I should have highlighted it.


I think there is enough clarity here for me to run with the explicit solution, but I will follow-up with discussion to make sure I'm clear on the details. Thanks again 😄

testing/internal/e2e/tests/base/target_tcp_connect_redis_test.go
```diff
@@ -19,14 +19,137 @@ import (
 	"github.com/stretchr/testify/require"
 )
```
bgajjala8: Nice job with the helper functions for Redis E2E test. Just a minor nit/question for the setup. Would it make sense to combine the two test functions (`TestCliTcpTargetConnectRedisWithUsernamePassword` and `TestCliTcpTargetConnectRedisWithPassword`) into a single `TestCliTcpTargetConnectRedis` function using subtests? We reuse the Redis container and Boundary resource setup, so we don't have to set it up twice since the tests are nearly identical except for the credential type. This _could_ make them easier to maintain. What do you think?

Also, just a comment on `setupBoundaryResources` seems like it could be used in other tests, but I'm unsure if we should move it to a common location since it can be reused across the `base_test` package.

laero: > We reuse the Redis container and Boundary resource setup, so we don't have to set it up twice since the tests are nearly identical except for the credential type.

i took the expedient copy/paste approach because i didn't really see us adding more tests here + simplifying isolation. i'll give this a shot anyways, it'll be a good learning experience for writing subtests

> Also, just a comment on setupBoundaryResources seems like it could be used in other tests, but I'm unsure if we should move it to a common location since it can be reused across the base_test package.

was conflicted about whether to do this or not as well, as it's just a particular series of calls to other helpers. i'll just go ahead and extract it + refine it a bit to be more generally usable

testing/internal/e2e/tests/base/target_tcp_connect_redis_test.go
```diff
@@ -11,23 +11,122 @@ import (
...
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			// Setup
+			resources := setupBoundaryCredentialStore(t, ctx, redisInfo.Hostname, redisInfo.Port)
+			c.extraSetup(t, ctx, resources, redisInfo)
+
+			// Flush redis between tests
+			t.Cleanup(func() {
+				flushRedis(t, ctx, resources.targetId)
+			})
```
bgajjala8: I see why you did this here. However, can we just flush it at the end of the test (where we are running the commands) after the `GET` command? This would save us from connecting to the target once again?

testing/internal/e2e/tests/base/target_tcp_connect_redis_test.go
```diff
@@ -11,23 +11,122 @@ import (
...
+				err = boundary.AddBrokeredCredentialSourceToTargetCli(t, ctx, resources.targetId, credentialId)
+				require.NoError(t, err)
+			},
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			// Setup
+			resources := setupBoundaryCredentialStore(t, ctx, redisInfo.Hostname, redisInfo.Port)
```
bgajjala8: (nit): can we reuse one credential store for all tests in the list instead of setting one up per test? I feel like both approaches are fine so I will leave it up to you on this one

testing/internal/e2e/tests/base/helpers_test.go
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) HashiCorp, Inc.
```
bgajjala8: Adding @moduli. Wanted to get your opinion on this addition here

testing/internal/e2e/tests/base/target_tcp_connect_redis_test.go
laero: i chose to separate the cleanup from the validation logic here for readability and maintainability. i think the trade-off of requiring a reconnect is worth it

if for some reason someone needs to update this test they don't need to pay attention to cleanup logic (they can't accidentally insert after flushing for example)

testing/internal/e2e/tests/base/helpers_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+// setupBoundaryResources sets up the following Boundary resources:
+// Org, Project, Target, Credential Store.
+// Also authenticates through CLI as admin.
+func setupBoundaryCredentialStore(t *testing.T, ctx context.Context, hostname string, port string) *boundaryResources {
+	boundary.AuthenticateAdminCli(t, ctx)
+
+	orgId, err := boundary.CreateOrgCli(t, ctx)
+	require.NoError(t, err)
+	t.Cleanup(func() {
```
laero: hmm, upon reviewing this i'm not really sure i like having the cleanup step tucked away here

testing/internal/e2e/tests/base/target_tcp_connect_redis_test.go
laero: i think it's better to have complete isolation here. it's unlikely but possible for a test to reconfigure resources, and that could cause issues if they are shared. we just avoid a lot of issues by ensuring there's a clean setup for each step

i'd consider recreating redis too, but i think that is probably overkill 🤔

testing/internal/e2e/tests/base/helpers_test.go
```diff
@@ -0,0 +1,57 @@
...
+	organizationId string
+	projectId      string
+	targetId       string
+	storeId        string
+}
+
+// setupBoundaryResources sets up the following Boundary resources:
+// Org, Project, Target, Credential Store.
+// Also authenticates through CLI as admin.
+func setupBoundaryCredentialStore(t *testing.T, ctx context.Context, hostname string, port string) *boundaryResources {
```
laero: i am still conflicted about whether this helper function is really that useful to make it available for other tests 🤔

testing/internal/e2e/boundary/credential.go
```diff
@@ -202,10 +202,10 @@ func CreateStaticCredentialPrivateKeyCli(t testing.TB, ctx context.Context, cred
 	return credentialId, nil
 }
 
-// CreateStaticCredentialPasswordCli uses the cli to create a new password credential in the
+// CreateStaticCredentialUsernamePasswordCli uses the cli to create a new password credential in the
 // provided static credential store.
 // Returns the id of the new credential
-func CreateStaticCredentialPasswordCli(t testing.TB, ctx context.Context, credentialStoreId string, user string, password string) (string, error) {
+func CreateStaticCredentialUsernamePasswordCli(t testing.TB, ctx context.Context, credentialStoreId string, user string, password string) (string, error) {
```
moduli, line 208: This change makes sense. Appreciate the cleanup/refactor here

testing/internal/e2e/tests/base/helpers_test.go
moduli: Mm I understand the intention here, but I'm hesitant to overly "don't repeat yourself" here. I see the end-to-end tests as workflows, and I like being able to read through a test and see the steps it's doing from a user's perspective (creating an org, what type of credential store, etc.)

testing/internal/e2e/tests/base/target_tcp_connect_redis_test.go
```diff
@@ -11,23 +11,122 @@ import (
...
+			require.NoError(t, err)
+			stdout, err := cmd.StdoutPipe()
+			require.NoError(t, err)
+			require.NoError(t, cmd.Start())
+
+			output, err := sendRedisCommand(stdin, stdout, "ACL WHOAMI\r\n")
+			require.NoError(t, err)
+			require.Equal(t, c.expectedUsername, output)
+
+			output, err = sendRedisCommand(stdin, stdout, "SET e2etestkey e2etestvalue\r\n")
```
moduli: Maybe we could randomize or name the key that's used here to prevent any potential conflicts? A potential idea could be
```
t.Name() + timestamp
```

That way, we could potentially remove the need for that FLUSH.

testing/internal/e2e/tests/base/target_tcp_connect_redis_test.go
```diff
@@ -11,23 +11,122 @@ import (
...
+
+			// Confirm that boundary connect has closed
+			err = cmd.Wait()
+			require.NoError(t, err)
+		})
+	}
+}
+
+// setupRedisContainer starts a Redis container and returns its connection info
+func setupRedisContainer(t *testing.T, ctx context.Context) *redisContainerInfo {
```
moduli: Feels like some of these redis methods might be better suited in another package? One idea is in `e2e/infra/redis_docker.go`?

testing/internal/e2e/tests/base/target_tcp_connect_redis_test.go
moduli: One idea I'm toying with is actually just separating these into two different tests in different files
```
target_tcp_connect_redis_username_password_test.go
target_tcp_connect_redis_password_test.go
```

That would help with readability of the test and how this file is currently organized. That would make it so you're recreating redis, but docker containers aren't too bad to spin up and down.

laero: redis starts up pretty quickly. would you choose separate files over creating two separate test functions in this file? i originally had two functions like so:
```
func TestCliTcpTargetConnectRedisWithUsernamePassword(t *testing.T) {...}
func TestCliTcpTargetConnectRedisWithPassword(t *testing.T) {...}
```

we moved to using subtests to re-use redis and reduce repeated code. i am partial to to having separate functions within the same file because it's a bit easier to browse in an IDE, and simpler to share helper functions if needed

moduli: I may be off-base in my opinion right now 😅 but here's what i'm thinking
- I'm less concerned about repeated code in e2e tests as they are limited in scope. They also help to serve as sort of like documentation for how to run certain tests. 
- I'm currently leaning towards having separate files. Part of me likes having one e2e test in a file. It keeps the file length smaller and helps to see the e2e coverage when looking at the files in a folder.

Please feel free to challenge those opinions.

testing/internal/e2e/tests/base/target_tcp_connect_redis_test.go
laero: i like this idea, i will add it 😄 

i'd prefer to keep the flush to ensure a clean slate for the next test (and to stick to the principle of isolating these tests from each other). i don't think there is much overhead to doing that here, and in the scenario someone updates this test, it will help ensure the tests don't affect each other.

although if we split the test apart like you suggested in the other comment and spin up redis per test, then we wouldn't need to flush

testing/internal/e2e/tests/base/helpers_test.go
laero: ah, okay so i originally had that helper function in the same test file. would you just prefer the setup code inside of the helper to be within the test function itself (no calls to helpers)?

moduli: I think in this particular case, with the org/project/credential store, that could can just live in the test itself without any helpers.

testing/internal/e2e/tests/base/target_tcp_connect_redis_test.go
laero: i saw your comment after i updated my comment, so i will just reiterate to make sure it's not lost:

* that's a really great point about repeated code in e2e, thanks for the insight!
* regarding the test organization:
  * having the tests in one file improves readability/organization when there are shared helper functions (although based on other comments these helpers are probably getting removed or refactored into a more appropriate place anyways)
  * just noting: some of the other e2e test files have multiple test functions already so using relying on filenames isn't giving us a full picture as it is. so you'd have to dive into those files anyways. it can still help if we stick to one test per file moving forward though
  * i think in most scenarios, people would be more interested in the coverage over a particular feature rather than overall coverage. it's easier to scroll down the outline / function list to see coverage of a feature, there is less to parse through because i'm not looking at a list containing unrelated filenames
  * i do agree that having smaller test files is a lot more readable than the large ones

i don't have a strong opinion on this, just exploring the options here. let me know if that has changed your mind or if you'd prefer to stick to one test per file 😄

testing/internal/e2e/tests/base/target_tcp_connect_redis_test.go
laero: i agree, especially if we want to consider splitting the test into separate files. your suggestion makes sense, i'll go ahead and try that

testing/internal/e2e/tests/base/target_tcp_connect_redis_test.go
laero: i went ahead and put them into separate files for now based on your original suggestion:
```
target_tcp_connect_redis_username_password_test.go
target_tcp_connect_redis_password_test.go
```

testing/internal/e2e/tests/base/target_tcp_connect_redis_test.go
laero: actually ended up keeping the existing code since we're recreating the redis containers now

bgajjala8 (approved):
LG2M! 😄

testing/internal/e2e/tests/base/target_tcp_connect_redis_username_password_test.go
```diff
@@ -6,60 +6,25 @@ package base_test
...
 	"github.com/hashicorp/boundary/testing/internal/e2e"
 	"github.com/hashicorp/boundary/testing/internal/e2e/boundary"
 	"github.com/hashicorp/boundary/testing/internal/e2e/infra"
-	"github.com/ory/dockertest/v3"
 	"github.com/stretchr/testify/require"
 )
 
-// TestCliTcpTargetConnectRedis uses the boundary cli to connect to a target using `connect redis`
-func TestCliTcpTargetConnectRedis(t *testing.T) {
+func TestCliTcpTargetConnectRedisUsernamePassword(t *testing.T) {
```
moduli, line 21: nit: perhaps a comment describing the intent of the test?
```
// TestCliTcpTargetConnectRedisUsernamePassword uses the boundary cli to connect to a target using `connect redis`. The target is configured to use a username/password credential... etc. etc.
```

same with the other test

laero: will add descriptions, thanks for catching this

PR #6219, opened 2025-11-04
refact(e2e): Rename credential methods for clarity

moduli:
## Description
This PR renames some e2e test methods for clarity. There will be a new "Password" credential type, so the existing methods should specify that it includes a username.

## PCI review checklist

### DIFF
--- a/testing/internal/e2e/boundary/credential.go
+++ b/testing/internal/e2e/boundary/credential.go
@@ -202,10 +202,10 @@ func CreateStaticCredentialPrivateKeyCli(t testing.TB, ctx context.Context, cred
 	return credentialId, nil
 }
 
-// CreateStaticCredentialPasswordCli uses the cli to create a new password credential in the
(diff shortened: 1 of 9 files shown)


testing/internal/e2e/boundary/credential.go
```diff
@@ -202,10 +202,10 @@ func CreateStaticCredentialPrivateKeyCli(t testing.TB, ctx context.Context, cred
 	return credentialId, nil
 }
 
-// CreateStaticCredentialPasswordCli uses the cli to create a new password credential in the
+// CreateStaticCredentialUsernamePasswordCli uses the cli to create a new password credential in the
```
moduli, line 205: @laero FYI: I'm making this change in this PR. Turns out we needed to also update another method name for something similar (UsernamePasswordDomain), and we're using these methods for some work I'm doing now, so I figured to just get this in.

PR #6232, opened 2025-11-07
Automatically Close `boundary connect` When Proxy is no Longer in Use

kheina:
## Description
This PR solves an issue where `boundary connect <subcommand>` remains open, even after a client has finished using the proxy tunnel.

Additionally introduces the `-inactive-timeout=<duration>` command option to `boundary connect` and all subcommands, allowing a user to optionally extend the timeout when necessary

## PCI review checklist

### DIFF
--- a/api/proxy/proxy.go
+++ b/api/proxy/proxy.go
@@ -321,7 +321,6 @@ func (p *ClientProxy) Start(opt ...Option) (retErr error) {
 			}
 		}()
 		defer p.connWg.Done()
-		defer p.cancel()
 		defer listenerCloseFunc()
 		for {
 			select {


api/proxy/option.go
```diff
@@ -142,3 +143,12 @@ func WithApiClient(with *api.Client) Option {
 		return nil
 	}
 }
+
+// WithInactivityTimeout provides an optional duration after which a session will
+// be cancelled with no active connections
```
johanbrandhorst: Nit: this reads a little easier to me
```suggestion
// WithInactivityTimeout provides an optional duration after which a session
// with no active connections will be cancelled
```

api/proxy/proxy.go
```diff
@@ -37,12 +37,16 @@ type ClientProxy struct {
 	tofuToken               string
 	cachedListenerAddress   *ua.String
 	connectionsLeft         *atomic.Int32
+	activeConns             *atomic.Int32
 	connsLeftCh             chan int32
 	callerConnectionsLeftCh chan int32
```
johanbrandhorst: I love that we already had two different ways to say "connection" 😂 . Should we double down on "conns" or "connections"?

api/proxy/proxy.go
```diff
@@ -256,6 +266,7 @@ func (p *ClientProxy) Start(opt ...Option) (retErr error) {
 					p.cancel()
 					return
 				}
+				// this is where the connections actually occur. if we want to count conns, it's here
```
johanbrandhorst: Is this comment intended to stay?

api/proxy/proxy.go
```diff
@@ -304,8 +315,8 @@ func (p *ClientProxy) Start(opt ...Option) (retErr error) {
 			}
 		}()
 		defer p.connWg.Done()
+		defer p.cancel()
```
johanbrandhorst: Why did we need to move this?

api/proxy/proxy.go
```diff
@@ -320,12 +331,37 @@ func (p *ClientProxy) Start(opt ...Option) (retErr error) {
 					// connections
 					return
 				}
+			default:
```
johanbrandhorst: Will this loop not busy loop now? The default cause will be hit almost every time. Are we waiting somewhere else where before we should wait for the context or connsLeftCh to be readable?

api/proxy/proxy.go
```diff
@@ -320,12 +331,37 @@ func (p *ClientProxy) Start(opt ...Option) (retErr error) {
...
+			}
+			switch {
+			case !p.beenActive.Load():
+				if time.Since(p.proxyStartTime) >= 10*time.Minute {
+					// Proxy has been inactive for 10 minutes, close it.
+					return
+				}
+			// nothing has happened yet, don't mark inactive
```
johanbrandhorst: Does this comment belong to the above or the below? Maybe add a newline somewhere in here to make it clear. Or move this comment above the if statement.

internal/cmd/commands/connect/connect.go
```diff
@@ -508,11 +517,31 @@ func (c *Command) Run(args []string) (retCode int) {
...
+			apiProxyOpts = append(apiProxyOpts, apiproxy.WithInactivityTimeout(3*time.Second))
+		}
+	}
+
 	proxyError := new(atomic.Error)
 	go func() {
 		defer close(clientProxyCloseCh)
-		if err = clientProxy.Start(); err != nil {
-			c.proxyCancel()
+		defer c.proxyCancel()
```
johanbrandhorst, line 543: This will now unconditionally cancel the proxy context before returning (before it was only during errors), is that intentional?

internal/cmd/commands/connect/connect.go
```diff
@@ -841,8 +869,28 @@ func (c *Command) handleExec(clientProxy *apiproxy.ClientProxy, passthroughArgs
...
+
+	go func() {
+		defer close(cmdExit)
+		if err := cmd.Start(); err != nil {
+			cmdError(err)
+		} else if err := cmd.Wait(); err != nil {
+			cmdError(err)
+		} else {
+			c.execCmdReturnValue.Store(0)
+		}
```
johanbrandhorst, line 883: This logic is fine but the if else is a bit unusual, can we use returns instead?
```suggestion
		defer close(cmdExit)
		if err := cmd.Start(); err != nil {
			cmdError(err)
			return
		}
		if err := cmd.Wait(); err != nil {
			cmdError(err)
			return
		}
		c.execCmdReturnValue.Store(0)
```

internal/tests/api/proxy/proxy_test.go
```diff
@@ -140,3 +140,95 @@ func TestConnectionsLeft(t *testing.T) {
...
+	written, err := conn.Write(echo)
+	require.NoError(err)
+	require.Equal(written, len(echo))
+	read, err := conn.Read(readBuf)
+	require.NoError(err)
+	require.Equal(read, len(echo))
+	require.NoError(conn.Close())
+
+	time.Sleep(2 * time.Second)
+	require.True(done)
```
johanbrandhorst: I see what you're doing here, but I think it'd be nicer to just have retries in both cases. In the first case we can retry dial and write until it works, in the second case you can loop with a 10ms sleep until done is `true`, or error after some timeout (5 seconds maybe?). Sleeps like this which make sense on paper are the first things that create flaky tests under heavy load (when running other tests). What do you think?

johanbrandhorst (commented):
Thanks!

internal/cmd/commands/connect/connect.go
kheina: `clientProxy.Start` is a misnomer, it does not start the proxy, it is the full lifespan of the proxy in its entirety. imo we should rename it to `clientProxy.Run`

api/proxy/proxy.go
kheina: it forces the context to be cancelled (closing the various proxy listeners) when this goroutine ends, rather than after p.connWg.Wait, which already waits until all the listeners are closed.

I think this was an edge case I ran into while prototyping, since if the proxy is being closed due to zero active conns, there shouldn't be anything else to wait for. either way I believe this is more "correct"

api/proxy/proxy.go
kheina: reworked the logic to use channels and timers rather than a busy loop. it should wait within the select now

api/proxy/proxy.go
```diff
@@ -142,7 +146,7 @@ func New(ctx context.Context, authzToken string, opt ...Option) (*ClientProxy, e
 	// We don't _rely_ on client-side timeout verification but this prevents us
 	// seeming to be ready for a connection that will immediately fail when we
 	// try to actually make it
-	p.ctx, p.cancel = context.WithDeadline(ctx, p.expiration)
+	p.ctx, p.cancel = context.WithDeadlineCause(ctx, p.expiration, fmt.Errorf("Session has expired"))
```
johanbrandhorst, line 149: :chefs-kiss:

api/proxy/proxy.go
```diff
@@ -212,6 +216,17 @@ func (p *ClientProxy) Start(opt ...Option) (retErr error) {
 	// Ensure closing the listener runs on any other return condition
 	defer listenerCloseFunc()
 
+	// automatically close the proxy when inactive
+	proxyAutoClose := time.AfterFunc(10*time.Minute, func() {
```
johanbrandhorst, line 220: Is the default of 10 minutes chosen for any particular reason? In the case where there is no inactivity timeout, we never reset this, won't it always close the proxy after 10 minutes?

internal/cmd/commands/connect/connect.go
johanbrandhorst: I see. Thanks. No need to rename the function in this PR.

internal/cmd/commands/connect/connect.go
```diff
@@ -841,8 +867,30 @@ func (c *Command) handleExec(clientProxy *apiproxy.ClientProxy, passthroughArgs
...
+			return
+		}
+		c.execCmdReturnValue.Store(0)
+	}()
+
+	for {
+		select {
+		case <-c.proxyCtx.Done():
+			// the proxy exited for some reason, end the cmd since connections are no longer possible
+			_ = cmd.Process.Signal(syscall.SIGTERM)
```
johanbrandhorst: Does this work on Windows? I thought the only signal possible [on Windows](https://pkg.go.dev/os#Signal) is `os.Interrupt`. If we don't care about windows, what do you think about moving this logic to a !windows file and adding a windows stub?

api/proxy/proxy.go
kheina: oh no you're right!! I had thought of that and addressed it in an earlier iteration but messed it up again

internal/cmd/commands/connect/connect.go
```diff
@@ -508,11 +517,31 @@ func (c *Command) Run(args []string) (retCode int) {
 	clientProxyCloseCh := make(chan struct{})
 	connCountCloseCh := make(chan struct{})
 
+	if c.flagInactiveTimeout >= 0 {
+		apiProxyOpts = append(apiProxyOpts, apiproxy.WithInactivityTimeout(c.flagInactiveTimeout))
+	} else {
```
louisruch: This does not seem like the correct approach - we are providing a `flagInactiveTimeout` that can be disabled with `-1` - but then ignore that and set random timeouts anyway. My biggest objection is to the `connect` flag as I am not clear why we need the timeout here unless provided by the user at all since we do not exec any process.

api/proxy/proxy.go
kheina: also 10 minutes was recommended by jeff, basically just a "you started this and never used it, we're gonna clean it up now"

internal/cmd/commands/connect/connect.go
kheina: the logic right now is that 0 disables the timeout, but I think it would be better to use -1 instead and have 0 be used to set defaults

louisruch: Ah I see that there is an additional check in the proxy now. But yes -1 generally indicates disable/infinite/unlimited. I would also recommend updating the flag description text to call out the expected behavior of -1 and 0

internal/cmd/commands/connect/connect.go
kheina: > On Windows, sending os.Interrupt to a process with os.Process.Signal is not implemented

this was why I actually used sigterm in the first place, but having looked into it more, idk if sigterm is honored either. I'm looking into it more.

internal/cmd/commands/connect/connect.go
kheina: updated that only `flagInactiveTimeout == 0` causes the defaults to be used, -1 timeout to be unset. also updated the flag comment to reflect this

internal/cmd/commands/connect/end_process_nonwindows.go
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+//go:build !windows
+// +build !windows
```
johanbrandhorst: I don't think we need the old style build tags anymore

internal/tests/api/proxy/proxy_test.go
```diff
@@ -140,3 +140,99 @@ func TestConnectionsLeft(t *testing.T) {
...
+	require.Equal(written, len(echo))
+	read, err := conn.Read(readBuf)
+	require.NoError(err)
+	require.Equal(read, len(echo))
+	require.NoError(conn.Close())
+
+	start := time.Now()
+	for {
+		if done == true || time.Since(start) > time.Second*2 {
+			require.True(done)
```
johanbrandhorst: Nit: if the timeout is hit, the only error the user will see is "Expected false, got true". Could we have special handling for the timeout case to make it easier to understand the test failure?

kheina:
> Is there an expected change in behavior?

~no, there shouldn't be any impact to this test. looking into it now.~
this should be fixed now

api/proxy/proxy.go
kheina: update: the error within the e2e tests was caused by this. I think cancelling the context here caused the proxy to close before copying the final bytes back to the user. I've removed this instead since p.cancel is already deferred within `ClientProxy.Start()` at the beginning of the func

johanbrandhorst (approved):
Looks good to me, thanks! We'll need to test this carefully with all the different scenarios.

PR #6235, opened 2025-11-10
Mark expired Vault token as expired when Vault is unreachable

bosorawis:
## Description

Vault credential stores have an internal job that periodically renews Vault tokens. If the renewal call returns a `403` error, the system assumes the token is no longer valid and marks it as expired.

When Vault is unreachable, the system logs an error and continues trying to renew the token past its expiration time since Vault never returns a 403 error. This change checks if the token has already expired; if so, it will mark the token as expired even when communication with Vault returns an error.

## PCI review checklist


internal/credential/vault/jobs.go
```diff
@@ -190,12 +196,16 @@ func (r *TokenRenewalJob) renewToken(ctx context.Context, s *clientStore) error
...
-	if ok := errors.As(err, &respErr); ok && respErr.StatusCode == http.StatusForbidden {
+	if err != nil {
 		// Vault returned a 403 when attempting a renew self, the token is either expired
 		// or malformed.  Set status to "expired" so credentials created with token can be
 		// cleaned up.
+		if isForbiddenError(err) ||
+			// Also check if the token has already expired based on time to avoid attempting to renew expired token
+			// against unreachable Vault server.
+			time.Now().After(token.ExpirationTime.AsTime()) {
+		}
```
irenarindos: Looks like this is an empty if statement (if X || Y{})

internal/credential/vault/jobs_test.go
```diff
@@ -500,6 +500,59 @@ func TestTokenRenewalJob_RunExpired(t *testing.T) {
...
+	require.NoError(err)
+
+	repo, err := NewRepository(ctx, rw, rw, kmsCache, sche)
+	require.NoError(err)
+	cs, err := repo.CreateCredentialStore(ctx, in)
+	require.NoError(err)
+
+	// Renewal should work here since Vault should still be reachable
+	err = r.Run(ctx, 0)
+	require.NoError(err)
```
louisruch, line 589: You should assert the token status here

internal/credential/vault/jobs_test.go
```diff
@@ -500,6 +500,59 @@ func TestTokenRenewalJob_RunExpired(t *testing.T) {
...
+
+	// Shutdown Vault server to make vault unreachable
+	v.Shutdown(t)
+	// Sleep to move clock and expire token
+	time.Sleep(time.Second * 2)
+
+	// Renewal should work here since Vault should still be reachable
```
louisruch: Your comment is wrong here, Vault is not reachable, but we still expect the job to run without error in the forbidden / expired case

louisruch (changes requested):
This will not work as you expect, the `clientStore` created from the `renewRevokeStore` does not retrieve or set the `ExpirationTime` - this change will therefore just expire all tokens on any error returned from Vault. We would need to update the store `.token()` method and update the query to return the expiration time for this to work.

We may also want to consider an alternative path here where the query we run filters out tokens that are past their expiration time - though we would still want to update these tokens to have an `expired` status somewhere.

internal/credential/vault/jobs_test.go
bosorawis: Added `token_expiration` to `credential_vault_token_renewal_revocation` view so `ExpirationTime` is set.

internal/credential/vault/private_store.go
```diff
@@ -98,7 +99,7 @@ func (ps *clientStore) token() *Token {
 		tk.Status = ps.TokenStatus
 		tk.CtToken = ps.CtToken
 		tk.KeyId = ps.TokenKeyId
-
+		tk.ExpirationTime = ps.TokenRenewalTime
```
louisruch: Auto complete typo I suspect

internal/db/schema/migrations/oss/postgres/103/01_credential_vault_token_renewal_revocation_view.up.sql
```diff
@@ -0,0 +1,52 @@
+-- Copyright (c) HashiCorp, Inc.
```
louisruch, line 1: Ugh so we have other work that is coming in at `101` either @bosorawis will need to update this and wait for password work to merge or @laero @bgajjala8 will need to update again after this merges

internal/db/schema/migrations/oss/postgres/101/01_credential_vault_token_renewal_revocation_view.up.sql
```diff
@@ -0,0 +1,52 @@
...
+           cert.certificate_key   as ct_client_key, -- encrypted
+           cert.key_id            as client_key_id
+      from credential_vault_store store
+      join tokens token
+        on store.public_id = token.store_id
+      left join credential_vault_client_certificate cert
+        on store.public_id = cert.store_id;
+   comment on view credential_vault_token_renewal_revocation is
+      'credential_vault_token_renewal_revocation is a view where each row contains a credential store and the credential store''s data needed to connect to Vault. '
+          'The view returns a separate row for each active token in Vault (current, maintaining and revoke tokens); this view should only be used for token renewal and revocation. '
```
louisruch: Not sure we want the additional tabs here

internal/db/schema/migrations/oss/postgres/101/01_credential_vault_token_renewal_revocation_view.up.sql
```diff
@@ -0,0 +1,52 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+-- public.credential_vault_token_renewal_revocation source
+
+begin;
+
+  drop view if exists credential_vault_token_renewal_revocation;
+
+  create or replace view credential_vault_token_renewal_revocation as
```
louisruch: Since we already dropped the view above we should not need the `or replace`

internal/db/schema/migrations/oss/postgres/101/01_credential_vault_token_renewal_revocation_view.up.sql
```diff
@@ -0,0 +1,52 @@
+-- Copyright (c) HashiCorp, Inc.
+-- SPDX-License-Identifier: BUSL-1.1
+
+-- public.credential_vault_token_renewal_revocation source
+
+begin;
+
+  drop view if exists credential_vault_token_renewal_revocation;
```
louisruch: I believe we moved away from adding `if exists` when dropping, as we know it exists and if it doesn't we should throw an error here

louisruch (changes requested):
Might wanna sync up with @laero or @bgajjala8 to chat through order or merges though they have some cleanup to do on the LLB already

internal/credential/vault/private_store.go
bosorawis: copilot 😡. Fixed and added a test

internal/db/schema/migrations/oss/postgres/103/01_credential_vault_token_renewal_revocation_view.up.sql
bosorawis: I moved to `102`. Happy to move this around as necessary

internal/credential/vault/testing.go
```diff
@@ -1196,6 +1196,9 @@ type TestVaultServer struct {
 	vaultContainer    any
 	ldapContainer     any
 	postgresContainer any
+
+	stopped  bool
```
irenarindos: Should we use an atomic here?

internal/credential/vault/jobs_test.go
```diff
@@ -500,6 +500,120 @@ func TestTokenRenewalJob_RunExpired(t *testing.T) {
 	require.Nil(cs)
 }
 
+// TestTokenRenewalJob_Run_VaultUnreachableTemporarily tests that tokens are not marked
+// as expired when Vault is unreachable temporarily.
+func TestTokenRenewalJob_Run_VaultUnreachableTemporarily(t *testing.T) {
+	// t.Parallel() - this was causing test failures, investigate before un-commenting
```
irenarindos: Should we just remove these at this point?

internal/credential/vault/private_store_test.go
```diff
@@ -6,15 +6,57 @@ package vault
...
 	"github.com/hashicorp/boundary/internal/db"
+	"github.com/hashicorp/boundary/internal/db/timestamp"
 	"github.com/hashicorp/boundary/internal/iam"
 	"github.com/hashicorp/boundary/internal/kms"
 	"github.com/hashicorp/boundary/internal/scheduler"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+func Test_token(t *testing.T) {
```
irenarindos: This isn't used anywhere- should this be removed?

irenarindos (changes requested):

internal/credential/vault/testing.go
louisruch: +1

internal/credential/vault/jobs_test.go
bosorawis: removed from this entire file

internal/credential/vault/private_store_test.go
bosorawis: done

irenarindos (approved):

louisruch (approved):
LGTM

PR #6239, opened 2025-11-11
chore(e2e): update vault create credential funcs

wongtonyb:
## Description
This PR is to update the e2e test helper functions `CreateKvPasswordCredential` and `CreateKvPasswordDomainCredential` to allow a password parameter to be passed in.

If a password is provided, it will be used to create the credential in vault (new behavior)
If not password is not provided, it will generate a random password to be used (old behavior)

## PCI review checklist

### DIFF
--- a/testing/internal/e2e/vault/vault.go
+++ b/testing/internal/e2e/vault/vault.go
@@ -295,7 +295,6 @@ func CreateKvPasswordCredential(t testing.TB, secretPath string, user string, pr
 	}
 
 	// Create secret
-	require.NoError(t, err)
 	output := e2e.RunCommand(context.Background(), "vault",
 		e2e.WithArgs(
 			"kv", "put",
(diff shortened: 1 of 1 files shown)


testing/internal/e2e/vault/vault.go
```diff
@@ -286,6 +286,14 @@ func CreateKvPasswordCredential(t testing.TB, secretPath string, user string) (s
 
 	policyName = WritePolicy(t, t.Context(), policyFilePath)
 
+	// Use provided password or generate random one
+	if len(providedPassword) > 0 && providedPassword[0] != "" {
+		password = providedPassword[0]
+	} else {
+		password, err = base62.Random(16)
+		require.NoError(t, err)
+	}
```
moduli, line 295: Did this work when you tested this?

It looks like `password` just gets overwritten below
```
// Create secret
password, err = base62.Random(16)
require.NoError(t, err)
```

wongtonyb: I just caught that too, removed that line. 

Was removed in the boundary-ent PR, forgot to remove it here

moduli: Do you also need to remove the `require.NoError(t, err)` that was under it?

PR #6254, opened 2025-11-17
backport: feat(clientproxy): Automatically Close `boundary connect` When Proxy is no Longer in Use

kheina:
## Description
This PR backports a fix for `boundary connect` from https://github.com/hashicorp/boundary/pull/6232 into release/0.20.

> This PR solves an issue where `boundary connect <subcommand>` remains open, even after a client has finished using the proxy tunnel.
> 
> Additionally introduces the `-inactive-timeout=<duration>` command option to `boundary connect` and all subcommands, allowing a user to optionally extend the timeout when necessary

## PCI review checklist

### DIFF
--- a/internal/cmd/commands/connect/connect.go
+++ b/internal/cmd/commands/connect/connect.go
@@ -59,15 +59,16 @@ var (
 type Command struct {
 	*base.Command
 
-	flagAuthzToken string
-	flagListenAddr string
-	flagListenPort int64
-	flagTargetId   string
-	flagTargetName string
-	flagHostId     string
-	flagExec       string
-	flagUsername   string
-	flagDbname     string
+	flagAuthzToken      string
+	flagListenAddr      string
+	flagListenPort      int64
+	flagTargetId        string
+	flagTargetName      string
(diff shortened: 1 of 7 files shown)


internal/cmd/commands/connect/connect.go
```diff
@@ -59,15 +59,17 @@ var (
...
+	flagAuthzToken                    string
+	flagListenAddr                    string
+	flagListenPort                    int64
+	flagTargetId                      string
+	flagTargetName                    string
+	flagHostId                        string
+	flagExec                          string
+	flagUsername                      string
+	flagDbname                        string
+	flagMongoDbAuthenticationDatabase string
```
kheina: I'm not 100% sure where this line exactly came from, but it's in main and seemed to jump over from the commit?

johanbrandhorst: I think we should probably drop this, we made some changes in `main` that might not have made it into 0.20.0

kheina: done

PR #6259, opened 2025-11-18
refactor: Update Password use of crypto/rand to SecureRandomReader

AbhiMgowda:
## Description
Need to unify the secure randomness in Boundary behind the use of the SecureRandomReader field, passing it in through options where we today use crypto/rand directly in [internal/auth/password.newArgon2Credential()](https://github.com/hashicorp/boundary/blob/59cea4d5b58c30ecba88c82f4e50f7b77d6c45fe/internal/auth/password/argon2.go#L182-L184).

No changes to security controls or cryptographic behavior:
* Uses the same crypto/rand.Reader underneath
* Salt generation remains cryptographically identical
* No changes to access control, logging pipelines, or audit trails
* This is a pure refactoring for API clarity; the security posture is unchanged

Testing
All existing tests pass:
* ✅ internal/auth/password package tests (12.67s)

**Files Modified**

    > internal/auth/password/argon2.go
    > internal/auth/password/argon2_test.go
    > internal/auth/password/options.go
    > internal/auth/password/options_test.go
    > internal/auth/password/repository.go
    > internal/auth/password/repository_account.go
    > internal/auth/password/repository_password.go
    > internal/auth/password/repository_test.go
    > internal/auth/password/rewrapping_test.go
    > internal/cmd/base/initial_resources.go

## PCI review checklist

### DIFF
--- a/internal/auth/password/repository_test.go
+++ b/internal/auth/password/repository_test.go
@@ -59,8 +59,10 @@ func TestRepository_New(t *testing.T) {
 				r:   rw,
 				w:   rw,
 				kms: kmsCache,
-				opts: []Option{WithLimit(5),
-					WithRandomReader(testReader)},
+				opts: []Option{
+					WithLimit(5),
+					WithRandomReader(testReader),
+				},
 			},
 			want: &Repository{
 				reader:       rw,


johanbrandhorst (changes requested):
This is not the change I expected to see. The change we need to make is to pass in the existing secure random reader into the context where random data is needed, not just create a new wrapper around `crypto/rand`. Could you take another look at the document I prepared in the JIRA ticket?

AbhiMgowda:
Presently in `base.go` we've a  `SecureRandomReader` defined in `server` struct and same is used in iam, controller and worker by importing `base` pkg.
but we can't able to import `base` package back on password or other places mentioned in the docs as it cause `import cycle error`
So I proposed this to make `SecureRandomReader` avail globally.

AbhiMgowda:
@johanbrandhorst , Thank you for reviewing the revised code changes and for providing clear direction on the initial implementation.

I have thoroughly re-examined the documentation provided in the JIRA ticket and updated the code to adhere to the requirement of passing the existing SecureRandomReader into the necessary context, rather than instantiating a new wrapper around crypto/rand.
The revised approach should now correctly integrate with our existing cryptographic setup and provide the expected behavior.
Please let me know if there are any further adjustments needed or if you have any questions regarding this revised implementation.
I appreciate your guidance on this matter.

AbhiMgowda:
> Thank you @AbhiMgowda, this is much more like what I had in mind! This passes the SecureRandomReader into the password repository, but I believe there were a few more places that need to be updated in the same way. This PR is fine for fixing the password case, but I assume you will update the rest of the cases in future PRs?

yes @johanbrandhorst, Pradeep created this as a sub-task for resolution and clarification. However, since the necessary changes span both Boundary and Boundary Enterprise, I plan to consolidate all modifications into a single Pull Request (PR). I will confirm this approach with Pradeep and provide an update.

AbhiMgowda:
@johanbrandhorst 
For this sub-task, only password changes will be included, updated, and merged. The remaining common items (Boundary and Enterprise) will be consolidated and addressed in a single, subsequent PR.
[](url)https://docs.google.com/document/d/1j-LIjrjvRtSY8dP0xuMUtJXGf6P74o-f-F-vnvfTqIA/edit?tab=t.0

PR #6269, opened 2025-11-20
Add sql schema and pgtap tests

mikemountain:
## Description
Closes ICU-17908. Add the app token database schema + pgtap tests to validate the schema.

## PCI review checklist

### DIFF
--- a/internal/db/sqltest/tests/apptoken/apptoken_project.sql
+++ b/internal/db/sqltest/tests/apptoken/apptoken_project.sql
@@ -3,7 +3,7 @@
 
 begin;
 
-select plan(15);
+select plan(16);
 select wtt_load('widgets', 'iam');
 
 -- insert app_token_project and make sure app_token has a value
@@ -86,6 +86,8 @@ select is(count(*), 1::bigint) from app_token_deleted where public_id = 'r_11111
 select is(count(*), 0::bigint) from app_token_project where public_id = 'r_1111111111';
 -- ensure app_token_permission_project is deleted
 select is(count(*), 0::bigint) from app_token_permission_project where app_token_id = 'r_1111111111';
+-- ensure app_token_cipher is deleted
+select is(count(*), 0::bigint) from app_token_cipher where app_token_id = 'r_1111111111';
 
 select * from finish();
 rollback;


internal/db/schema/migrations/oss/postgres/101/01_app_token_base.up.sql
```diff
@@ -0,0 +1,246 @@
...
+    name text primary key
+      constraint only_predefined_app_token_global_grant_scope_allowed
+        check(
+          name in (
+            'individual',
+            'children',
+            'descendants'
+          )
+        )
+  );
```
dkanney, line 18: Why create a table & FK constraint for these enums instead of creating an enum type? e.g.

```sql
create type app_token_global_grant_scope_enum as enum ('individual', 'children', 'descendants');
```

By using an enum type, we'd have one less table to `join` on

__

EDIT: I noticed we use this pattern elsewhere, so I don't expect it to change here - just curious if you knew why

internal/db/schema/migrations/oss/postgres/101/01_app_token_base.up.sql
```diff
@@ -0,0 +1,246 @@
...
+
+  -- Function to validate that revoked can only be updated from false to true
+  create or replace function validate_app_token_revocation() returns trigger
+  as $$
+  begin
+    -- For updates, check revoked field changes
+    if old.revoked is distinct from new.revoked then
+      -- Only allow change from false to true
+      if not (old.revoked = false and new.revoked = true) then
+        raise exception 'App token cannot be unrevoked. revoked value. Current: %, Attempted: %',
```
dkanney: It looks like we don't need the "revoked value." part of this exception message

```suggestion
        raise exception 'App token cannot be unrevoked. Current: %, Attempted: %',
```

internal/db/schema/migrations/oss/postgres/101/01_app_token_base.up.sql
```diff
@@ -0,0 +1,246 @@
...
+        on update cascade,
+    raw_grant text not null
+      constraint raw_grant_must_not_be_empty
+      check(
+        length(trim(raw_grant)) > 0
+      ),
+    primary key(permission_id, canonical_grant)
+  );
+  comment on table app_token_permission_grant is
+    'app_token_permission_grant contains grants assigned to app tokens in project scope';
```
dkanney: It looks like this table can apply to app tokens at any scope, not just the project scope. Should this instead say,

`'app_token_permission_grant contains grants assigned to app tokens'`

internal/db/schema/migrations/oss/postgres/101/01_app_token_base.up.sql
```diff
@@ -0,0 +1,246 @@
...
+          on update cascade,
+    key_id text not null
+      constraint kms_data_key_version_fkey
+        references kms_data_key_version (private_id)
+        on delete restrict
+        on update cascade,
+    token bytea not null unique
+  );
+  comment on table app_token_cipher is
+    'app_token_cipher is the table for application token encryption keys. '
```
dkanney: Assuming the intent here was to say "app token encryption keys" instead of "application token encryption keys":

```suggestion
    'app_token_cipher is the table for app token encryption keys. '
```

dkanney: Maybe not -- feel free to ignore this if "application tokens" was intended

internal/db/schema/migrations/oss/postgres/101/02_app_token_global.up.sql
```diff
@@ -0,0 +1,220 @@
...
+        on delete cascade
+        on update cascade,
+    description text,
+    grant_this_scope boolean not null default false,
+    grant_scope text not null
+      constraint app_token_global_grant_scope_enm_fkey
+        references app_token_global_grant_scope_enm(name)
+        on delete restrict
+        on update cascade,
+    create_time wt_timestamp,
```
dkanney: Is this field supposed to be on this table? The Whimsical DB diagram doesn't have a `create_time` field on this table

internal/db/schema/migrations/oss/postgres/101/02_app_token_global.up.sql
```diff
@@ -0,0 +1,220 @@
...
+    -- This restricts the grant_scope to be 'individual'
+    -- and since it is also a foreign key to the app_token_permission_global
+    -- grant_scope, it ensures that app_token_permission_global is set to 'individual'
+    -- if this table is populated for the corresponding permission.
+    grant_scope text not null
+       constraint only_individual_grant_scope_allowed
+         check(
+          grant_scope = 'individual'
+        ),
+    create_time wt_timestamp,
```
dkanney: This field isn't on Whimsical, but it's here -- intentional?

internal/db/schema/migrations/oss/postgres/101/02_app_token_global.up.sql
```diff
@@ -0,0 +1,220 @@
...
+       from iam_scope_org
+      where iam_scope_org.scope_id = new.scope_id;
+    if not found then
+      raise exception 'org scope_id % not found', new.scope_id;
+    end if;
+    return new;
+  end;
+  $$ language plpgsql;
+    comment on function validate_global_permission_org_scope() is
+      'validate_global_permission_org_scope is used to enforced that scope ID added to app_token_permission_global_individual_project_grant_scope'
```
dkanney: Small grammar nitpick -

```suggestion
      'validate_global_permission_org_scope is used to enforce that scope ID added to app_token_permission_global_individual_project_grant_scope'
```

internal/db/schema/migrations/oss/postgres/101/02_app_token_global.up.sql
```diff
@@ -0,0 +1,220 @@
...
+        references app_token_global(public_id)
+        on delete cascade
+        on update cascade,
+    description text,
+    grant_this_scope boolean not null default false,
+    grant_scope text not null
+      constraint app_token_global_grant_scope_enm_fkey
+        references app_token_global_grant_scope_enm(name)
+        on delete restrict
+        on update cascade,
```
dkanney, line 82: Why reference another table here instead of referncing an enum type or doing something like this:

```sql
    grant_scope text not null
       constraint only_predefined_app_token_global_grant_scopes_allowed
         check(
          grant_scope in ('individual', 'children', 'descendents')
        ),
```

internal/db/schema/migrations/oss/postgres/101/02_app_token_global.up.sql
bosorawis: good catch! `create_time` isn't needed here

internal/db/schema/migrations/oss/postgres/101/02_app_token_global.up.sql
bosorawis: We should be safe to not include `create_time` here since `create_time` isn't needed when checking the token grants. Also, token grants are immutable so we could infer this timestamp by looking at the token's `create_time` instead.

internal/db/sqltest/tests/apptoken/apptoken_global.sql
```diff
@@ -0,0 +1,226 @@
...
+  insert into app_token_global (
+    public_id,
+    scope_id,
+    revoked,
+    created_by_user_id,
+    create_time,
+    update_time,
+    approximate_last_access_time,
+    expiration_time
+  ) values ('r_1111111111', 'global', true, 'u_____walter', now(), now(), now(), now() + interval '7 day');
```
bosorawis: nitpick: ID format should be `appt_...` instead of `r_....`
```suggestion
  ) values ('appt_1111111', 'global', true, 'u_____walter', now(), now(), now(), now() + interval '7 day');
```

internal/db/sqltest/tests/apptoken/apptoken_global.sql
```diff
@@ -0,0 +1,226 @@
...
+  where public_id = 'r_1111111111';
+select lives_ok('update_approximate_last_access_time');
+-- ensure approximate_last_access_time was updated
+select is(count(*), 1::bigint) from app_token_global
+ where public_id = 'r_1111111111'
+   and approximate_last_access_time > now() + interval '1 day';
+-- ensure approximate_last_access_time was updated in app_token table as well
+select is(count(*), 1::bigint) from app_token
+ where public_id = 'r_1111111111'
+   and approximate_last_access_time > now() + interval '1 day';
```
bosorawis, line 45: Was this meant to check for 2 days?

```suggestion
   and approximate_last_access_time > now() + interval '2 day';
-- ensure approximate_last_access_time was updated in app_token table as well
select is(count(*), 1::bigint) from app_token
 where public_id = 'r_1111111111'
   and approximate_last_access_time > now() + interval '2 day';
```

internal/db/sqltest/tests/apptoken/apptoken_global.sql
```diff
@@ -0,0 +1,226 @@
...
+    permission_id,
+    scope_id,
+    grant_scope
+  ) values ('p_nonexistent_permission', 'o_____widget', 'individual');
+select throws_like('insert_app_token_pgi_org_grant_scope_invalid_permission_id',
+  'insert or update on table "app_token_permission_global_individual_org_grant_scope" violates foreign key constraint "app_token_permission_global_fkey"');
+
+-- insert app_token_permission_global_individual_org_grant_scope with:
+-- non-individual grant_scope, permission_id that exists in app_token_permission_global, valid org scope id
+-- should fail
```
bosorawis, line 159: missing tests:

- individual scopes cannot be added when `app_token_global_permissions.grant_scope = descendant`
- individual org scopes cannot be added when `app_token_global_permissions.grant_scope = children`
-

internal/db/sqltest/tests/apptoken/apptoken_global.sql
mikemountain: No, 1 day, because we're checking to make sure the last access time is greater than 1 day from now after we updated it to be 2 days (just to account for weird time wiggle room)

internal/db/sqltest/tests/apptoken/apptoken_project.sql
```diff
@@ -0,0 +1,91 @@
...
+select lives_ok('delete_app_token');
+-- ensure app_token is deleted
+select is(count(*), 0::bigint) from app_token where public_id = 'r_1111111111';
+-- ensure token is automatically entered in app_token_deleted
+select is(count(*), 1::bigint) from app_token_deleted where public_id = 'r_1111111111';
+-- ensure app_token_project is deleted
+select is(count(*), 0::bigint) from app_token_project where public_id = 'r_1111111111';
+-- ensure app_token_permission_project is deleted
+select is(count(*), 0::bigint) from app_token_permission_project where app_token_id = 'r_1111111111';
+
```
bosorawis, line 91: also check that `app_token_cipher` is deleted

PR #6315, opened 2025-12-29
chore(test) fix flaky tests

bosorawis:
## Description
An attempt to reduce flaky test failures in the CI system

- `internal/cmd/commands/server/worker_tags_reload_test.go` - switch to polling worker tags instead of sleeping for a specific amount and read once to reduce timing errors
- `internal/cmd/commands/server/worker_initial_upstreams_reload_test.go` - remove `t.Parallel()` to be consistent with other tests in this package
- **[RACE]** `internal/daemon/worker/routing_info.go` - accessing `*LastRoutingInfo` through `atomic.Load` returns the exact point which can create race data access if the returned value is modified by the reader. Make `LastRoutingInfoSuccess()` return a deep copy instead of the pointer
- `internal/tests/cluster/parallel/orphaned_connections_test.go` - raise `SuccessfulControllerRPCGracePeriodDuration` and `WorkerRPCGracePeriod` to allow the assertion to wait longer for the workers to be available.
- change port numbers in various tests to avoid startup failure
- send an exit message when Boundary server exits early and increase timeout duration for waiting for server to start
- decrease worker RPC timing across the board to speed up data-sync in tests

## PCI review checklist

### DIFF
--- a/internal/daemon/worker/routing_info.go
+++ b/internal/daemon/worker/routing_info.go
@@ -181,6 +181,8 @@ func (w *Worker) sendWorkerRoutingInfo(cancelCtx context.Context) {
 		// append initial upstreams/ cluster addr to the resolver to try
 		if pastGrace, _, _ := w.isPastGrace(); pastGrace && w.GrpcClientConn.Load().GetState() == connectivity.TransientFailure {
 			routingInfo := w.lastRoutingInfoSuccess.Load().(*LastRoutingInfo)
+			// make a deep copy to avoid race conditions from accessing underlying data
+			// since routingInfo may be modified later in this function
 			copied := *routingInfo
 			lastRoutingInfo := &copied
 			if lastRoutingInfo != nil && lastRoutingInfo.LastCalculatedUpstreams != nil {


internal/daemon/worker/routing_info.go
```diff
@@ -178,7 +180,9 @@ func (w *Worker) sendWorkerRoutingInfo(cancelCtx context.Context) {
 		// In the case that the control plane has gone down and come up with different IPs,
 		// append initial upstreams/ cluster addr to the resolver to try
 		if pastGrace, _, _ := w.isPastGrace(); pastGrace && w.GrpcClientConn.Load().GetState() == connectivity.TransientFailure {
-			lastRoutingInfo := w.lastRoutingInfoSuccess.Load().(*LastRoutingInfo)
+			routingInfo := w.lastRoutingInfoSuccess.Load().(*LastRoutingInfo)
+			copied := *routingInfo
+			lastRoutingInfo := &copied
```
AprilMay0, line 187: Could this use a small comment to describe why this data is being copied instead of directly passed to `lastRoutingInfo` to prevent Future Us refactoring it back and re-introducing the flakiness?

internal/cmd/commands/server/worker_tags_reload_test.go
```diff
@@ -98,21 +98,32 @@ func TestServer_ReloadWorkerTags(t *testing.T) {
 		t.Fatalf("timeout")
 	}
 
-	fetchWorkerTags := func(name string, key string, values []string) {
+	waitForWorkerTags := func(name string, key string, values []string) {
```
AprilMay0, line 101: Nice!

AprilMay0 (approved):
Looks good to me! Thanks for picking this up 👍🏼

internal/daemon/worker/routing_info.go
bosorawis: added a comment

moduli:
Thank you for addressing these!

PR #6318, opened 2026-01-07
feat(query): Create queries for non-recursive requests of global app token grants

dkanney:
- ICU-17922
- ICU-17923
- ICU-17924

## Description

Retrieves grants for App Tokens that live in the Global scope. Queries will be selected based on the request scope. The request scope can be either (1) the global scope, (2) an org scope, or (3) a project scope.

## PCI review checklist

### DIFF
--- a/internal/apptoken/query.go
+++ b/internal/apptoken/query.go
@@ -37,9 +37,6 @@ left join app_token_permission_global_individual_project_grant_scope project_gra
 left join iam_scope_project
        on project_grant_scope.scope_id = iam_scope_project.scope_id
  group by app_token_permission_global.private_id,
-          app_token_permission_global.description,
-          app_token_permission_global.grant_this_scope,
-          app_token_permission_global.grant_scope,
           app_token_global.public_id;
     `
 
@@ -68,9 +65,6 @@ left join app_token_permission_global_individual_org_grant_scope org_grant_scope
 left join iam_scope_org
        on org_grant_scope.scope_id = iam_scope_org.scope_id
  group by app_token_permission_global.private_id,
-          app_token_permission_global.description,
-          app_token_permission_global.grant_this_scope,
-          app_token_permission_global.grant_scope,
           app_token_global.public_id;
     `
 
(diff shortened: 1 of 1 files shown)


internal/apptoken/query.go
```diff
@@ -270,9 +255,97 @@ left join iam_scope_project
...
+        join app_token_permission_global
+          on app_token_global.public_id = app_token_permission_global.app_token_id
+        join app_token_permission_grant
+          on app_token_permission_global.private_id = app_token_permission_grant.permission_id
+        join iam_grant
+          on app_token_permission_grant.canonical_grant = iam_grant.canonical_grant
+         and iam_grant.resource = any(@resources)
+       where app_token_permission_global.grant_this_scope
+    group by app_token_permission_global.private_id,
+             app_token_permission_global.description,
```
mikemountain: Do we need to group by `description`? Is this necessary for it to show up in the query or something like that? My psql is rusty.

mikemountain (approved):

internal/apptoken/query.go
dkanney: Nope, we absolutely _don't_ need `description` in the group by clause. TIL [the reason why](https://stackoverflow.com/questions/62332982/why-group-by-is-forced-on-specific-column-but-not-on-the-others#:~:text=Postgres%20is%20one%20of%20the%20rare%20databases%20(it%20not%20the%20only%20one)%20that%20properly%20implements%20the%20concept%20of%20functionaly%2Ddependent%20column%3A%20once%20you%20put%20the%20primary%20key%20of%20a%20table%20in%20a%20group%20by%20clause%2C%20you%20don%27t%20need%20to%20add%20other%20columns%20of%20the%20same%20table%3A%20the%20primary%20key%20uniquely%20identify%20a%20row.)

To save you a click -- non-Primary Key columns aren't required in the group by clause when the table's Primary Key already exists in the group by clause: `app_token_permission_global.private_id`

On the IAM side - because we're using CTEs for those queries (and CTEs do not inherently have Primary Keys), _all_ fields are required in the group by clause.

PR #6328, opened 2026-01-12
feat(repository): Implement Create global app token functionality and encryption

mikemountain:
## Description
This PR adds the functionality for creating an apptoken based off of a provided domain object in the global scope, inserting all the pertinent information into the db, as well as the encryption and decryption of the apptoken.

[ICU-17945 - Implement encryption/decryption](ec20b706242e472c03f9cd0f8f0d4d43a686d1f8)

[ICU-17934 - Implement create functionality for global app tokens](f4b13a80fadef382c8038fb403c8dc56ac496645)

## PCI review checklist

### DIFF
--- a/internal/apptoken/repository.go
+++ b/internal/apptoken/repository.go
@@ -214,10 +214,6 @@ func (r *Repository) CreateAppToken(ctx context.Context, token *AppToken) (*AppT
 		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("creating app token in database"))
 	}
 
-	if err := r.reader.LookupByPublicId(ctx, globalAppToken); err != nil {
-		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("app token lookup"))
-	}
-
 	newAppToken := &AppToken{
 		PublicId:                  globalAppToken.PublicId,
 		ScopeId:                   globalAppToken.ScopeId,
--- a/internal/apptoken/repository_test.go
+++ b/internal/apptoken/repository_test.go
@@ -426,6 +426,7 @@ func TestRepository_CreateAppToken(t *testing.T) {
 			assert.NoError(err)
 			assert.NotNil(at.CreateTime)
 			assert.NotNil(at.ApproximateLastAccessTime)
+			assert.NotNil(at.Token)
 			assert.Equal(at.CreateTime, at.ApproximateLastAccessTime)
 			assert.GreaterOrEqual(at.ExpirationTime.AsTime().Unix(), at.CreateTime.AsTime().Unix())
 


internal/apptoken/testing.go
```diff
@@ -0,0 +1,298 @@
...
+			insert into app_token_permission_org (private_id, app_token_id, description, grant_this_scope, grant_scope, create_time)
+			values ($1, $2, $3, $4, $5, now())
+		`
+		insertGrantSQL = `
+			insert into app_token_permission_grant (permission_id, raw_grant, canonical_grant)
+			values ($1, $2, $3)
+		`
+	case strings.HasPrefix(scopeId, globals.ProjectPrefix):
+		insertTokenSQL = `
+			insert into app_token_project (public_id, scope_id, name, description, created_by_user_id,create_time, update_time)
```
AprilMay0: Nit: missing space between created_by_user_id,create_time

internal/apptoken/apptoken.go
```diff
@@ -4,9 +4,29 @@
...
+const (
+	appTokenGlobalTableName                                      = "app_token_global"
+	appTokenCipherTableName                                      = "app_token_cipher"
+	appTokenPermissionGlobalTableName                            = "app_token_permission_global"
+	appTokenPermissionGlobalIndividualOrgGrantScopeTableName     = "app_token_permission_global_individual_org_grant_scope"
+	appTokenPermissionGlobalIndividualProjectGrantScopeTableName = "app_token_permission_global_individual_project_grant_scope"
+	appTokenPermissionGrantTableName                             = "app_token_permission_grant"
+
+	// The version prefix is used to differentiate token versions just for future proofing.
+	tokenValueVersionPrefix = "0"
```
AprilMay0, line 28: Good thinking!

internal/apptoken/repository.go
```diff
@@ -0,0 +1,241 @@
...
+// CreateToken creates the provided app token in the repository.
+func (r *Repository) CreateAppToken(ctx context.Context, token *AppToken) (*AppToken, error) {
+	const op = "apptoken.(Repository).CreateToken"
+	if token == nil {
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "missing app token")
+	}
+
+	id, err := newAppTokenId(ctx)
+	if err != nil {
+		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("57"))
```
dkanney: What does this error message mean? `errors.WithMsg("57")`

internal/apptoken/repository.go
```diff
@@ -0,0 +1,241 @@
...
+	case strings.HasPrefix(token.GetScopeId(), globals.GlobalPrefix):
+		tokenToCreate = &appTokenGlobal{
+			AppTokenGlobal: &store.AppTokenGlobal{
+				PublicId:           id,
+				ScopeId:            token.ScopeId,
+				Name:               token.Name,
+				Description:        token.Description,
+				Revoked:            token.Revoked,
+				CreatedByUserId:    token.CreatedByUserId,
+				TimeToStaleSeconds: token.TimeToStaleSeconds,
```
dkanney, line 71: We don't set `ExpirationTime` here? Does that mean all Global tokens will have a hard-coded 3-year TTL

internal/apptoken/repository.go
mikemountain: 🤦‍♂️ this was for debugging a segfault, I swore I caught all of them

internal/apptoken/repository.go
mikemountain: Great catch! Missed this.

internal/apptoken/apptoken.go
```diff
@@ -65,3 +73,168 @@ func (a *AppToken) IsActive() bool {
...
+}
+
+// DeletedScope represents a scope which has been deleted from an AppTokenPermission.
+type DeletedScope struct {
+	ScopeId   string
+	TimeStamp *timestamp.Timestamp
+}
+
+// newAppTokenCipher generates a new in-memory token.
+func newAppTokenCipher(ctx context.Context) (string, error) {
```
bosorawis: nitpick: since we're returning a string and not `appTokenCipher`, this could be renamed to `newToken`

internal/apptoken/testing.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	assert.Equal(grantedPerms, wantPerms)
+	return nil
+}
+
+func testCheckPermissionGrants(t *testing.T, repo *Repository, appTokenId string, wantGrants []string) error {
+	assert := assert.New(t)
+	permGrantsQuery := `
+		select permission_id, canonical_grant, raw_grant from app_token_permission_grant where permission_id in (
+			select private_id from app_token_permission where app_token_id = $1
```
bosorawis: nitpick: formatting with our SQL style guide

internal/apptoken/repository.go
```diff
@@ -0,0 +1,244 @@
...
+		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("while creating app token"))
+	}
+	if err := r.reader.LookupByPublicId(ctx, tokenToCreate); err != nil {
+		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("app token lookup"))
+	}
+	token.PublicId = tokenToCreate.PublicId
+	token.ApproximateLastAccessTime = tokenToCreate.ApproximateLastAccessTime
+	token.CreateTime = tokenToCreate.CreateTime
+	token.ExpirationTime = tokenToCreate.ExpirationTime
+	token.Revoked = tokenToCreate.Revoked
```
bosorawis: [important] I think this mutates the input. Since we're looking to return a copy, might be more appropriate to construct a new object instead of mutating the input (or make a deep copy)

internal/apptoken/repository.go
```diff
@@ -0,0 +1,244 @@
...
+				TimeToStaleSeconds: token.TimeToStaleSeconds,
+				ExpirationTime:     token.ExpirationTime,
+			},
+		}
+	default:
+		return nil, errors.New(ctx, errors.InvalidParameter, op, "invalid scope type")
+	}
+
+	// insert into app_token_global table
+	err = r.writeToDb(ctx, tokenToCreate)
```
bosorawis: [important] These writes should happen within 1 transaction so we could easily rollback when things go wrong

internal/apptoken/repository.go
```diff
@@ -0,0 +1,244 @@
...
+		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("failed to generate cipher token"))
+	}
+	token.Token = cipherToken
+	atc := &appTokenCipher{
+		AppTokenCipher: &store.AppTokenCipher{
+			AppTokenId: id,
+			Token:      token.Token,
+		},
+	}
+	databaseWrapper, err := r.kms.GetWrapper(context.Background(), token.ScopeId, kms.KeyPurposeDatabase)
```
bosorawis:
```suggestion
	databaseWrapper, err := r.kms.GetWrapper(ctx, token.ScopeId, kms.KeyPurposeDatabase)
```

internal/apptoken/repository.go
```diff
@@ -0,0 +1,244 @@
...
+			return nil, errors.New(ctx, errors.InvalidParameter, op, "children grant scope cannot be combined with individual org grant scopes")
+		}
+
+		// generate new permission ID
+		permId, err := newAppTokenPermissionId(ctx)
+		if err != nil {
+			return nil, errors.Wrap(ctx, err, op)
+		}
+
+		grantThisScope := slices.Contains(perm.GrantedScopes, "this")
```
bosorawis: nitpick
```suggestion
		grantThisScope := slices.Contains(perm.GrantedScopes, globals.GrantScopeThis)
```

internal/apptoken/repository.go
```diff
@@ -0,0 +1,244 @@
...
+				},
+			}
+			err = r.writeToDb(ctx, permissionGrantToCreate)
+			if err != nil {
+				return nil, errors.Wrap(ctx, err, op, errors.WithMsg("while creating apptoken permission grant"))
+			}
+		}
+
+		for _, gs := range perm.GrantedScopes {
+			if gs == "this" || gs == "children" || gs == "descendants" || globalPermGrantScope == "descendants" {
```
bosorawis: nitpick: I don't think `globalPermGrantScope == "descendants"` check is necessary. If users specify `descendants` AND individual scopes, we should return an error rather than quietly skip inserting the individual scope(s)


You could also calculate the final list of individual grant scope before iterating through said list by removing the special grant scopes from the original list

internal/apptoken/repository.go
```diff
@@ -0,0 +1,244 @@
...
+						GrantScope:   globalPermGrantScope,
+						ScopeId:      gs,
+					},
+				}
+				err = r.writeToDb(ctx, individualProjGlobalPermToCreate)
+				if err != nil {
+					return nil, errors.Wrap(ctx, err, op, errors.WithMsg("while creating apptoken individual project grant scope"))
+				}
+			default:
+				return nil, errors.New(ctx, errors.InvalidParameter, op, "invalid grant scope")
```
bosorawis: nitpick
```suggestion
				return nil, errors.New(ctx, errors.InvalidParameter, op, fmt.Sprintf("invalid grant scope %s", gs))
```

internal/apptoken/repository.go
bosorawis: important - need to avoid consuming multiple DB connections with `kms.WithReader` and `kms.WithWriter`

internal/apptoken/repository.go
```diff
@@ -0,0 +1,244 @@
...
+		}
+
+		grantThisScope := slices.Contains(perm.GrantedScopes, "this")
+		// if perm.GrantedScopes contains "children", set globalPermGrantScope to "children"
+		// if perm.GrantedScopes contains "descendants", set globalPermGrantScope to "descendants"
+		// if perm.GrantedScopes contains neither but does have at least one individual org or project,
+		// set globalPermGrantScope to "individual" and create individual grant scope entries below
+		globalPermGrantScope := "individual"
+		if slices.Contains(perm.GrantedScopes, "children") {
+			globalPermGrantScope = "children"
```
bosorawis:
```suggestion
			globalPermGrantScope = globals.GrantScopeChildren
```

internal/apptoken/repository.go
```diff
@@ -0,0 +1,244 @@
...
+		} else if slices.Contains(perm.GrantedScopes, "descendants") {
```
bosorawis:
```suggestion
		} else if slices.Contains(perm.GrantedScopes, globals.GrantScopeDescendants) {
```

internal/apptoken/repository.go
```diff
@@ -0,0 +1,244 @@
...
+	err = r.writeToDb(ctx, tokenToCreate)
+	if err != nil {
+		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("while creating app token"))
+	}
+	if err := r.reader.LookupByPublicId(ctx, tokenToCreate); err != nil {
+		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("app token lookup"))
+	}
+	token.PublicId = tokenToCreate.PublicId
+	token.ApproximateLastAccessTime = tokenToCreate.ApproximateLastAccessTime
+	token.CreateTime = tokenToCreate.CreateTime
```
bosorawis: these two value shouldn't be set - `CreateTime` is set by the trigger and `ApproximateLastAccessTime` is set by using the token

internal/apptoken/repository.go
```diff
@@ -0,0 +1,244 @@
...
+			default:
+				return nil, errors.New(ctx, errors.InvalidParameter, op, "invalid grant scope")
+			}
+		}
+	}
+
+	return token, nil
+}
+
+func (r *Repository) writeToDb(ctx context.Context, tokenToCreate interface{}) error {
```
dkanney: nitpick: Since we use this function to insert token permissions too, perhaps we could rename the `tokenToCreate` param to be more generic?

internal/apptoken/repository.go
dkanney: I think this is okay beacuse we're assiging to the input struct & we already performed the insert above

dkanney: I think this is okay because we're assigning to the input struct & we already performed the insert above

internal/apptoken/repository.go
dkanney: 💡 an idea:

If we want to perform all the writes in `Create` in a single transaction (insert token & insert its permissions), we could change `tokenToCreate` into a slice/variadic param & accumulate the writes we need to perform into a slice and perform them at the end

```go
func (r *Repository) CreateAppToken(ctx context.Context, token *AppToken) (*AppToken, error) {
  ...

  stuffToInsert := make(interface{}, 0)
  defer func() {
    if err == nil && len(stuffToInsert) > 0 {
      if err = writeToDb(ctx, stuffToInsert...); err != nil { ... }
    }
  }
  ...

  return nil
}

func (r *Repository) writeToDb(ctx context.Context, stuffToCreate ...interface{}) error {
  ...
}
```

internal/apptoken/apptoken.go
mikemountain: Can do, though this may lead to some confusion about what is *really* being generated at first glance, which is why I used `newAppToken`.
`newAppTokenToken` is too wordy though, so I'm okay with using `newToken` with the comment above it

internal/apptoken/repository.go
```diff
@@ -38,6 +44,197 @@ func NewRepository(ctx context.Context, r db.Reader, w db.Writer, kms *kms.Kms,
...
+		if err != nil {
+			return nil, errors.Wrap(ctx, err, op)
+		}
+
+		grantThisScope := slices.Contains(perm.GrantedScopes, globals.GrantScopeThis)
+		// if perm.GrantedScopes contains "children", set globalPermGrantScope to "children"
+		// if perm.GrantedScopes contains "descendants", set globalPermGrantScope to "descendants"
+		// if perm.GrantedScopes contains neither but does have at least one individual org or project,
+		// set globalPermGrantScope to "individual" and create individual grant scope entries below
+		globalPermGrantScope := "individual"
```
bosorawis:
```suggestion
		globalPermGrantScope := globals.GrantScopeIndividual
```

internal/apptoken/repository_test.go
```diff
@@ -93,3 +99,345 @@ func TestNewRepository(t *testing.T) {
...
+				assert.Contains(err.Error(), tt.wantErrMsg)
+				assert.True(errors.Match(errors.T(tt.wantIsError), err))
+				return
+			}
+			assert.NoError(err)
+			assert.NotNil(at.CreateTime)
+			assert.NotNil(at.ApproximateLastAccessTime)
+			assert.Equal(at.CreateTime, at.ApproximateLastAccessTime)
+			assert.GreaterOrEqual(at.ExpirationTime.AsTime().Unix(), at.CreateTime.AsTime().Unix())
+
```
bosorawis, line 432: nitpick: you could also validate that `at.Token` isn't empty to ensure we actually return the cipher-text token

internal/apptoken/repository.go
```diff
@@ -38,6 +44,197 @@ func NewRepository(ctx context.Context, r db.Reader, w db.Writer, kms *kms.Kms,
...
+			return nil
+		},
+	)
+	if err != nil {
+		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("creating app token in database"))
+	}
+
+	if err := r.reader.LookupByPublicId(ctx, globalAppToken); err != nil {
+		return nil, errors.Wrap(ctx, err, op, errors.WithMsg("app token lookup"))
+	}
```
bosorawis: What happens when this fails? From my understanding, we would return an error to the caller which will not return the token value. This puts the user in a scenario where token creation is successful but they can never obtain the token value. Could you move this into the transaction above?

internal/apptoken/repository.go
```diff
@@ -38,6 +44,197 @@ func NewRepository(ctx context.Context, r db.Reader, w db.Writer, kms *kms.Kms,
...
+
+	// batch write all collected inserts
+	_, err = r.writer.DoTx(
+		ctx,
+		db.StdRetryCnt,
+		db.ExpBackoff{},
+		func(_ db.Reader, w db.Writer) error {
+			for _, appTokenObject := range dbInserts {
+				if err := w.Create(ctx, appTokenObject); err != nil {
```
bosorawis, line 206: nitpick: Might be worth using `CreateItems` to avoid individual round-trip to the DB when the number of objects to be created is big

bosorawis (approved):
F few non-blocking concerns but the implementation LGTM! Good work 🎉

internal/apptoken/repository.go
mikemountain: `CreateItems` only works if each item is the same, and since they're all different store objects, it returns an error

internal/apptoken/repository.go
mikemountain: So we can't really put it into the tx above, because the items don't show up in the db until after the tx has been committed, but having checked how we create resources in other Repositories, we don't actually need `LookupByPublicId` here anyway

PR #6342, opened 2026-01-14
feat(clientcache): Extend cache SearchParams and options to support sorting

sepehrfrgh:
## Description
This PR will add new sort parameters to SearchParams in cache and extend options to support sort parameters with validations
## PCI review checklist

### DIFF
--- a/internal/clientcache/internal/cache/options.go
+++ b/internal/clientcache/internal/cache/options.go
@@ -6,13 +6,19 @@ package cache
 import (
 	stderrors "errors"
 	"fmt"
-	"strings"
+	"regexp"
+	"slices"
 
 	"github.com/hashicorp/go-dbw"
 )
 
-// unsafeSortChars contains characters that could break SQL ORDER BY clauses
-const unsafeSortChars = " \t\n\r,;()'\"\\-"
+// safeSortColumnRegex contains characters that could break SQL ORDER BY clauses
+var safeSortColumnRegex = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_]*$`)
+
+var (
+	errInvalidSortColumn = stderrors.New("not allowed for this resource type")
+	errUnsafeSortColumn  = stderrors.New("contains unsafe characters")
+)
 
 type testRefreshWaitChs struct {
 	firstSempahore  chan struct{}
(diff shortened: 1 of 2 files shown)


internal/clientcache/internal/cache/options.go
```diff
@@ -139,3 +146,35 @@ func WithUseNonPagedListing(b bool) Option {
...
+		}
+
+		o.withSortBy = sortBy
+		o.withSortDirection = direction
+		return nil
+	}
+}
+
+// sliceContains checks if a value exists in a slice.
+func sliceContains[T comparable](slice []T, val T) bool {
```
emilia-grant: You can replace this with [slices.Contains](https://pkg.go.dev/slices#Contains)

internal/clientcache/internal/cache/options.go
```diff
@@ -139,3 +146,35 @@ func WithUseNonPagedListing(b bool) Option {
 		return nil
 	}
 }
+
+// WithSort configures sorting for query results.
+// Empty sortBy is silently ignored. Empty direction defaults to ascending in the repository layer.
+// Validates column against sortableColumns and rejects SQL-unsafe characters.
+func WithSort(sortBy SortBy, direction SortDirection, sortableColumns []SortBy) Option {
+	return func(o *options) error {
+		if sortBy == SortByDefault {
```
emilia-grant, line 162: Does this prevent sorting the default column by the opposite direction? e.g. sort by created_time ascending

internal/clientcache/internal/cache/options_test.go
```diff
@@ -126,4 +126,65 @@ func Test_GetOpts(t *testing.T) {
...
+		require.NoError(t, err)
+		testOpts := getDefaultOptions()
+		testOpts.withSortBy = SortByCreatedAt
+		testOpts.withSortDirection = Descending
+		assert.Equal(t, opts, testOpts)
+	})
+	t.Run("WithSort-column-not-in-sortable-list", func(t *testing.T) {
+		_, err := getOpts(WithSort(SortByName, Ascending, []SortBy{SortByCreatedAt}))
+		require.Error(t, err)
+		assert.ErrorContains(t, err, "not allowed for this resource type")
```
emilia-grant: Probably not worth changing here, but a way to make this test more durable is to use static errors in your code under test and then you can check against the error type instead of the message.

e.g.

```go
// At the top of the file where the code under test
var errMyErr = errors.New("my error message")
...
// Wherever the error is returned:
return errMyErr

// In test code:
assert.ErrorIs(t, err, errMyErr)

File: internal/clientcache/internal/cache/options.go (hunk 1 above)
Comment (sepehrfrgh) [reply internal/clientcache/internal/cache/options.go]: Good catch forgot that `slices` has this thanks

File: internal/clientcache/internal/cache/options.go (hunk 2 above)
Comment (sepehrfrgh) [reply internal/clientcache/internal/cache/options.go:162]: Good catch this check will change the default logic

File: internal/clientcache/internal/cache/options_test.go (hunk 3 above)
Comment (sepehrfrgh) [reply internal/clientcache/internal/cache/options_test.go]: Are you referring to defining static error in `options.go` or `options_test.go`?

Comment (emilia-grant) [reply internal/clientcache/internal/cache/options_test.go]: That would need to be in `options.go`

File: internal/clientcache/internal/cache/options.go [hunk 4]
```diff
@@ -5,8 +5,20 @@ package cache
...
 	"github.com/hashicorp/go-dbw"
+	"github.com/pkg/errors"
+)
+
+// unsafeSortChars contains characters that could break SQL ORDER BY clauses
+var safeSortColumnRegex = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_]*$`)
+
+var (
+	errInvalidSortColumn = errors.New("not allowed for this resource type")
+	errUnsafeSortColumn  = errors.New("contains unsafe characters")
```
irenarindos: These look like they're only used in testing so it feels weird being in the options file. Normally if we define an error like that we return it as the error.
If it's just used for string matching in the test, we should define these there instead.

PR #6346, opened 2026-01-20
feat(clientcache): pass WithSort to searchTargets

emilia-grant:
## Description
This PR implements passing the validated sorting options through the repository layer into the db query

## PCI review checklist

### DIFF
--- a/internal/clientcache/internal/cache/repository_targets.go
+++ b/internal/clientcache/internal/cache/repository_targets.go
@@ -373,8 +373,18 @@ func (r *Repository) searchTargets(ctx context.Context, condition string, search
 		searchArgs = append(searchArgs, opts.withUserId)
 	}
 
+	dbOpts := []db.Option{db.WithLimit(opts.withMaxResultSetSize + 1)}
+	if opts.withSortBy != "" {
+		sd := opts.withSortDirection
+		if sd != Ascending && sd != Descending {
+			sd = Ascending
+		}
+		orderClause := fmt.Sprintf("%s %s", opts.withSortBy, sd)
+		dbOpts = append(dbOpts, db.WithOrder(orderClause))
+	}
+
 	var cachedTargets []*Target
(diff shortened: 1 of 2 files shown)


internal/clientcache/internal/cache/repository_targets.go
```diff
@@ -373,8 +373,18 @@ func (r *Repository) searchTargets(ctx context.Context, condition string, search
 		searchArgs = append(searchArgs, opts.withUserId)
 	}
 
+	dbOpts := []db.Option{db.WithLimit(opts.withMaxResultSetSize + 1)}
+	if opts.withSortBy != "" {
+		sd := opts.withSortDirection
```
sepehrfrgh, line 378: Do you think we might need more defensive validation here?
```
// Additional validation to prevent SQL injection even if WithSort validation is bypassed
if !safeSortColumnRegex.MatchString(string(opts.withSortBy)) {
			return nil, errors.New(ctx, errors.InvalidParameter, op, "invalid sort column contains unsafe characters")
		}
```

internal/clientcache/internal/cache/repository_targets_test.go
```diff
@@ -358,6 +358,27 @@ func TestRepository_ListTargets(t *testing.T) {
 		assert.Len(t, l.Targets, len(ts))
 		assert.ElementsMatch(t, l.Targets, ts)
 	})
+
```
sepehrfrgh, line 361: we need empty sortby test as well

emilia-grant: Since `""` isn't in `sortableColumns` it's covered by the bad SortBy test. We could add a test for bad sortable columns but that isn't user input, so less worried about it

internal/clientcache/internal/cache/repository_targets.go
emilia-grant: We already limit to exact sortable columns so I think it would be overkill. You could imagine the current test as a regex that tests against exactly the sortable columns :nerd_face:

sepehrfrgh (approved):
LGTM Thanks

PR #6362, opened 2026-01-22
feat(cli): sort-by and sort-direction cli parameters

emilia-grant:
## Description
This PR adds the sort-by and sort-direction parameters for the cli.

## PCI review checklist

### DIFF
--- a/internal/clientcache/cmd/search/search.go
+++ b/internal/clientcache/cmd/search/search.go
@@ -179,7 +179,7 @@ func (c *SearchCommand) Run(args []string) int {
 		c.PrintCliError(stderrors.New("Max result set size must be greater than or equal to -1"))
 		return base.CommandUserError
 	case c.flagMaxResultSetSize > math.MaxInt:
-		c.PrintCliError(stderrors.New(fmt.Sprintf("Max result set size must be less than or equal to the %v", math.MaxInt)))
+		c.PrintCliError(fmt.Errorf("Max result set size must be less than or equal to the %v", math.MaxInt))
 		return base.CommandUserError
 	}
 
@@ -197,11 +197,11 @@ func (c *SearchCommand) Run(args []string) int {
 	if c.flagSortBy != "" {
 		sortableBy, ok := daemon.SortableColumnsForResource[cache.ToSearchableResource(c.flagResource)]
 		if !ok {
-			c.PrintCliError(stderrors.New(fmt.Sprintf("resource %q is not sortable", c.flagResource)))
+			c.PrintCliError(fmt.Errorf("resource %q is not sortable", c.flagResource))
 			return base.CommandUserError
 		}
 		if !slices.Contains(sortableBy, cache.SortBy(c.flagSortBy)) {
-			c.PrintCliError(stderrors.New(fmt.Sprintf("resource %q is not sortable by %q. %q is sortable by %v", c.flagResource, c.flagSortBy, c.flagResource, sortableBy)))
+			c.PrintCliError(fmt.Errorf("resource %q is not sortable by %q. %q is sortable by %v", c.flagResource, c.flagSortBy, c.flagResource, sortableBy))
 			return base.CommandUserError
 		}
 	}


internal/clientcache/cmd/search/search.go
```diff
@@ -128,6 +131,18 @@ func (c *SearchCommand) Flags() *base.FlagSets {
 		Usage:  `Forces a refresh to be attempted prior to performing the search`,
 		Hidden: true,
 	})
+	f.StringVar(&base.StringVar{
+		Name:       "sort-by",
+		Target:     &c.flagSortBy,
+		Usage:      `Specifies which column to sort resources by. Use sort-direction to control which direction to sort results by.`,
```
sepehrfrgh: Question: Do you think we should mention which columns are allowed?

emilia-grant: Yeah I was kinda curious if we wanna stick it there or just in the docs. What do you think?

internal/clientcache/internal/daemon/search_handler.go
```diff
@@ -59,11 +59,11 @@ const (
 	forceRefreshKey     = "force_refresh"
 	authTokenIdKey      = "auth_token_id"
 	maxResultSetSizeKey = "max_result_set_size"
-	sortByKey           = "sort_by"
-	sortDirectionKey    = "sort_direction"
+	SortByKey           = "sort_by"
```
sepehrfrgh: nitpick: Since these are the exported, they should have comments explaining their purpose and usage

internal/clientcache/internal/daemon/search_handler.go
```diff
@@ -59,11 +59,11 @@ const (
...
+	SortDirectionKey    = "sort_direction"
 )
 
-var sortableColumnsForResource = map[cache.SearchableResource][]cache.SortBy{
+var SortableColumnsForResource = map[cache.SearchableResource][]cache.SortBy{
```
sepehrfrgh, line 69: nitpick: Same as above, they should have comments explaining their purpose and usage

internal/clientcache/cmd/search/search.go
```diff
@@ -167,6 +182,11 @@ func (c *SearchCommand) Run(args []string) int {
 		return base.CommandUserError
 	}
```
sepehrfrgh, line 185: The CLI doesn't validate that the provided sort-by column is valid for the selected resource type before sending the request. While autocomplete helps, users can still provide invalid column names that will only fail at the daemon level. I think we should add it

internal/clientcache/cmd/search/search.go
```diff
@@ -167,6 +182,11 @@ func (c *SearchCommand) Run(args []string) int {
...
 
+	if c.flagSortDirection != "" && c.flagSortBy == "" {
```
sepehrfrgh: The validation only checks if sort-direction is provided without sort-by, but doesn't validate that sort-direction is one of the allowed values i.e. `asc` or `desc`, we could validate in client side before it gets to daemon

internal/clientcache/cmd/search/search.go
```diff
@@ -491,3 +519,11 @@ type filterBy struct {
 	forceRefresh     bool
 	maxResultSetSize int
 }
+
+func getSortableResources() []string {
```
sepehrfrgh, line 542: nitpick: `getSortableResources()` sounds a little misleading. we are returning sortable column names for resources, not resource types themselves. I would suggest renaming it to `getSortableColumns()` or `getSortableColumnNames()`

emilia-grant: This actually gets the sortable resources and not the columns (it's iterating over the keys in the map, not the values)

internal/clientcache/cmd/search/search.go
emilia-grant: I think this makes sense, even though it doubles up validation logic

internal/clientcache/cmd/search/search_test.go
```diff
@@ -329,12 +334,81 @@ func TestSearch(t *testing.T) {
...
 		assert.NotNil(t, resp)
 		assert.NotNil(t, r)
 		assert.Len(t, r.Sessions, 1)
 	})
+
+	t.Run("sorted targets from list", func(t *testing.T) {
+		resp, r, apiErr, err := search(ctx, srv.BaseDotDir(), filterBy{
+			authTokenId: at.Id,
+			resource:    "targets",
+		}, "name", "descending")
```
sepehrfrgh: Tests use `descending` as the sort direction value, but the actual valid values are `asc` and `desc` as defined in the `sortDirections` slice. This works because the daemon likely accepts both, but it's inconsistent with the documented valid values.

internal/clientcache/cmd/search/search.go
sepehrfrgh: I looked at all the cmd package. I think we can mention it in command line message, that which columns are allowed

```suggestion
		Usage:      `Specifies which column to sort resources by. Available columns depend on the resource type ("name" for targets, "created_time" for sessions). Use sort-direction to control which direction to sort results by (asc or desc)`,
```

PR #6373, opened 2026-01-23
test(e2e): update rdp enos modules to wait for pw

wongtonyb:
## Description
This PR is to address flakiness in the rdp enos scenarios with retrieving the server's password.

https://hashicorp.atlassian.net/browse/ICU-18331

We've identified that it is an intermittent issues with AWS windows servers, and it is recommended to add a wait delay between initializing the server and retrieving the password.

## PCI review checklist

### DIFF
--- a/enos/modules/aws_rdp_member_server_with_worker/main.tf
+++ b/enos/modules/aws_rdp_member_server_with_worker/main.tf
@@ -353,13 +353,13 @@ resource "enos_local_exec" "run_powershell_script" {
   inline = ["ssh -i ${local.private_key} -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no Administrator@${aws_instance.worker.public_ip} ${local.test_dir}/${basename(local_file.powershell_script.filename)}"]
 }
 
-resource "time_sleep" "wait_3_minutes" {
+resource "time_sleep" "wait_2_minutes" {
   depends_on      = [enos_local_exec.run_powershell_script]
-  create_duration = "3m"
+  create_duration = "2m"
 }
 
 data "aws_instance" "instance_password" {
-  depends_on        = [time_sleep.wait_3_minutes]
+  depends_on        = [time_sleep.wait_2_minutes]
   instance_id       = aws_instance.worker.id
   get_password_data = true
 }


enos/modules/aws_rdp_domain_controller/main.tf
```diff
@@ -452,13 +432,22 @@ resource "aws_instance" "domain_controller" {
...
-
   tags = {
     Name = "${var.prefix}-domain-controller-${local.username}"
   }
 }
 
+resource "time_sleep" "wait_for_domain_controller_init" {
+  depends_on      = [aws_instance.domain_controller]
+  create_duration = "3m"
+}
```
moduli: What was the reason for adding an extra sleep here? We don't need to use it for any other following resource other than the output, right? Could we just put this after the current sleep?

wongtonyb: Yeah it's just for the output. What's the current sleep you're referring to?

moduli: For example, you could put it after `resource "time_sleep" "wait_10_minutes" {`

wongtonyb: ah ok I see it now - was looking at the client module which didn't have an existing one

will make the change, shouldn't be a problem since its just for the output

moduli: The client has something like
```
resource "enos_local_exec" "wait_for_ssh" {
```

wongtonyb: the `wait_for_ssh` resource doesn't have a minimum wait time, but I think we want to explicitly wait a few minutes before trying to retrieve the password - so I'll keep the `wait_3_minute` resource in the client module

moduli: or put it as the last step. trying to avoid manual sleeps unless absolutely necessary.

enos/modules/aws_rdp_member_server_with_worker/main.tf
```diff
@@ -373,9 +353,15 @@ resource "enos_local_exec" "run_powershell_script" {
   inline = ["ssh -i ${local.private_key} -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no Administrator@${aws_instance.worker.public_ip} ${local.test_dir}/${basename(local_file.powershell_script.filename)}"]
 }
 
-resource "time_sleep" "wait_2_minutes" {
+resource "time_sleep" "wait_3_minutes" {
```
moduli: what was the reason for changing this to 3 minutes?

enos/modules/aws_windows_client/main.tf
```diff
@@ -405,3 +385,9 @@ resource "local_file" "powershell_script_output" {
   content    = enos_local_exec.run_powershell_script[0].stdout
   filename   = "${path.root}/.terraform/tmp/setup_windows_client.out"
 }
+
+data "aws_instance" "instance_password" {
+  depends_on        = [enos_local_exec.wait_for_ssh]
```
moduli: if you are concerned about the ssh timeout wait not being long enough, you could have this depend on  instead
```
  depends_on = [enos_local_exec.run_powershell_script]
```

enos/modules/aws_rdp_member_server_with_worker/main.tf
wongtonyb: based on the ticket, I thought the want was for a 3min wait - so just making it a consistent minimum 3 mins wait on all of them

enos/modules/aws_windows_client/main.tf
wongtonyb: yeah this looks like a better option

enos/modules/aws_rdp_member_server_with_worker/main.tf
moduli: Ah I see. There's no implication/requirement to do 3 minutes. It just so happened that for that example in that module, it was for 3 minutes.

wongtonyb: Ah ok, I'll change it back to 2 mins then

PR #6420, opened 2026-02-09
feat(repository): LookupAppToken() queries and repository function

dkanney:
ICU-17931

## Description
- Global, Org, and Project queries to fetch core App Token data alongside its permission data
- `LookupAppToken()` repository method utilizes these queries (based on token subtype) to fetch full app token information, including token cipher and permission data

## PCI review checklist

### DIFF
--- a/internal/apptoken/query.go
+++ b/internal/apptoken/query.go
@@ -574,7 +574,7 @@ with token_permissions as (
          app_token_project.scope_id                            as grant_scope,
          array_agg(app_token_permission_grant.canonical_grant) as canonical_grants,
          app_token_project.public_id                           as app_token_id,
-         array_agg(app_token_project.scope_id)                 as active_grant_scope_ids,
+         '{}'::text[]                                          as active_grant_scope_ids,
          '{}'::text[]                                          as deleted_grant_scope_ids,
          jsonb '[]'                                            as deleted_scope_details
        from app_token_project
@@ -584,9 +584,6 @@ with token_permissions as (
          on app_token_permission_project.private_id = app_token_permission_grant.permission_id
       where app_token_project.public_id = @app_token_id
    group by app_token_permission_project.private_id,
-            app_token_permission_project.description,
-            app_token_permission_project.grant_this_scope,
-            app_token_project.scope_id,
             app_token_project.public_id
 )
    select app_token_project.public_id,
--- a/internal/apptoken/query.go
+++ b/internal/apptoken/query.go
@@ -380,22 +380,6 @@ left join iam_scope_project
              app_token_org.public_id;
     `
 
-	// TODO: This will be properly implemented with the Create method
-	// getAppTokenByIdQuery retrieves an AppToken by its public ID
-	getAppTokenByIdQuery = `
-	 select public_id, scope_id
-	   from app_token_global
-	  where public_id = $1
-	  union all
-	 select public_id, scope_id
-	   from app_token_org
-	  where public_id = $1
-	  union all
-	 select public_id, scope_id
-	   from app_token_project
-	  where public_id = $1
-	`
-
 	// estimateCountAppTokens estimates the total number of app tokens in the three app token tables
 	estimateCountAppTokens = `
    select sum(reltuples::bigint) as estimate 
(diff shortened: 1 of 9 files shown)


internal/apptoken/repository.go
```diff
@@ -623,3 +601,265 @@ func (r *Repository) estimatedCount(ctx context.Context) (int, error) {
...
+			}
+
+			// Unpack permissions JSON from query results\
+			if err := json.Unmarshal(res.permissionsJSON, &permissionsRes); err != nil {
+				return errors.Wrap(ctx, err, op, errors.WithMsg("failed to unmarshal permissions"))
+			}
+		}
+		if err := rows.Err(); err != nil {
+			return errors.Wrap(ctx, err, op)
+		}
```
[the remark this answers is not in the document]
dkanney: Added a comment to make this clearer, but the "zero-rows" case is caught earlier in this function when we call `getAppTokenScopeType()`. The block between lines 694-696 is an additional protection against this.

internal/apptoken/query.go
```diff
@@ -380,26 +380,245 @@ left join iam_scope_project
...
+            jsonb_build_object(
+              'label', token_permissions.description,
+              'grant_this_scope', token_permissions.grant_this_scope,
+              'grants', token_permissions.canonical_grants,
+              'grant_scope', token_permissions.grant_scope,
+              'active_grant_scopes', token_permissions.active_grant_scope_ids,
+              'deleted_grant_scopes', token_permissions.deleted_grant_scope_ids,
+              'deleted_scope_details', token_permissions.deleted_scope_details
+            )
+          ) as permissions
```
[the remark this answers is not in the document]
dkanney: Simplified the permissions comparison in the test to just be:

```go
assert.ElementsMatch(want.Permissions, got.Permissions)
```

which makes the comparison regardless of order.

internal/apptoken/repository.go
```diff
@@ -623,3 +781,34 @@ func (r *Repository) estimatedCount(ctx context.Context) (int, error) {
 	}
 	return count, nil
 }
+
+// getAppTokenScopeId returns the scope id of the app token
+func getAppTokenScopeId(ctx context.Context, reader db.Reader, id string) (string, error) {
```
mikemountain, line 849: Can this be a method so that `reader` isn't being passed in? If not, can you add to the comment why it's necessary that it's not a method

mikemountain (changes requested):

internal/apptoken/repository.go
dkanney: I think we need the ability to pass around a `reader` when calling this function inside a transaction

e.g. `CreateAppToken()` -> open a new Tx with its own dedicated reader -> call `LookupAppToken()` w/ the the Tx reader -> call `getAppTokenScopeId()` with the Tx reader

mikemountain (approved):

internal/apptoken/repository.go
```diff
@@ -48,6 +50,207 @@ func NewRepository(ctx context.Context, r db.Reader, w db.Writer, kms *kms.Kms,
...
+	DeletedScopeDetails []deletedScopeResult `json:"deleted_scope_details"`
+}
+
+// deletedScopeResult represents a scope which has been deleted from an AppTokenPermission.
+type deletedScopeResult struct {
+	ScopeId   string `json:"scope_id"`
+	TimeStamp string `json:"delete_time"`
+}
+
+// LookupAppToken returns an AppToken for the id. Returns nil if no AppToken is found for id.
```
dkanney, line 87: > Returns nil if no AppToken is found for id.

But why?

dkanney: This was based on `LookupRole()` in `repository_role.go`

In that function, before looking up the full role, we query for its scope type via `getRoleScopeType()`. `getRoleScopeType()` _does_ attempt to lookup the role with the given id, returning an error when one doesn't exist. This error gets wrapped and returned by `LookupRole()`, so this comment isn't 100% accurate for `apptoken.LookupRole()`.

This comment only accurately describes the case where the scope type is a valid scope type (global/org/project). Otherwise, lookup does indeed return an error.

dkanney: Going to align with `iam.LookupRole()` for now & return nil here.

PR #6428, opened 2026-02-12
fix: sort target names case insensitively

emilia-grant:
## Description
This PR changes the order for sorting targets by name to match how the FEs previously agreed to sort stuff by name: sort case insensitive with ties broken by case.

e.g.
Descending
```
name_Target 3
name_TaRgEt 2
name_Target 2
name_TARGET 1
name_Target 1
```
Ascending
```
name_Target 1
name_TARGET 1
name_Target 2
name_TaRgEt 2
name_Target 3
```
## PCI review checklist

### DIFF
--- a/internal/clientcache/internal/cache/repository_targets.go
+++ b/internal/clientcache/internal/cache/repository_targets.go
@@ -386,7 +386,7 @@ func (r *Repository) searchTargets(ctx context.Context, condition string, search
 		orderClause := ""
 		switch opts.withSortBy {
 		case "name":
-			orderClause = fmt.Sprintf("lower(%s) %s, %s %s", opts.withSortBy, sd, opts.withSortBy, sd)
+			orderClause = fmt.Sprintf("%s collate nocase %s, %s %s", opts.withSortBy, sd, opts.withSortBy, sd)
 		default:
 			orderClause = fmt.Sprintf("%s %s", opts.withSortBy, sd)
 		}


internal/clientcache/internal/cache/repository_targets.go
```diff
@@ -379,7 +379,18 @@ func (r *Repository) searchTargets(ctx context.Context, condition string, search
...
 		}
-		orderClause := fmt.Sprintf("%s %s", opts.withSortBy, sd)
+
+		// Sorting by name requires doing case insensitive, then tie-breaking with case sensitive to put names which were
+		// originally capitalized first. To avoid weird behavior when adding new columns, break sort by name's case
+		// out separately.
+		orderClause := ""
+		switch opts.withSortBy {
+		case "name":
+			orderClause = fmt.Sprintf("lower(%s) %s, %s %s", opts.withSortBy, sd, opts.withSortBy, sd)
```
ZedLi: I don't know what the current setup of the db is but I just wanted to quickly point out that there's a small gotcha in that `lower` won't use any index if there are any on that column compared to `collate nocase` (unless you also used a `lower` as part of the index) but otherwise the same if you don't have any indexes.

emilia-grant: We have to use `lower` in this case because `collate nocase` doesn't support unicode unless sqlite is compiled with a specific flag. We could make an index for it but I'm not familiar enough with sql to know if there are any complications with that. 

I was curious if it would have a huge impact in our case, so I ran an experiment where I timed both kinds of sorting query on a db with 1,000,000 rows in a single table. It ran 200 times (10 groups of 10 attempts of 2 sorts, recreating the db for each group of attempts) and the average of averages was with within .2% (219.6ms for lower and 220ms for collate). So I'm not super inclined to worry about it here unless we see some concrete performance issues. [code](https://github.ibm.com/Emilia-Grant/dbsorttest/tree/egrant-collate-test)

ZedLi: What unicode support are you referring to in this case? I was under the impression both don't offer proper unicode support and [lower](https://sqlite.org/lang_corefunc.html#lower) would also require the ICU extension for any non ASCII characters.

I also agree it should show no difference in the case where there's no index which it looks like is what's happening in that test. It also looks like there's no index on name in the [schema](https://github.com/hashicorp/boundary/blob/egrant-dcs-target-sort-order/internal/clientcache/internal/db/schema.sql) so there shouldn't be any difference as it is.

I just wanted to bring up this easy to miss gotcha for the future since if an index is ever added, any sort that uses `lower` would not end up using that index.

emilia-grant: Aw dang, looks like I got misled by bad info on stackoverflow :sob: That's what I get for not reading the docs. I'll go ahead and swap it. Thanks for your help!

sepehrfrgh (approved):
I think this uses two-level sort (case-insensitive followed by case-sensitive), which requires SQLite to perform two comparisons per pair of elements. However, this is likely negligible for typical use cases and the trade-off for deterministic ordering is reasonable.   
Great job. Thank you

Issue #6398, opened 2026-02-03
Sessions terminating with Inactivity timeout reached


bhvishal9:
**Describe the bug**
When connecting to a TCP target with boundary, it automatically terminates the session if there is no activity. The time period after which it does so varies, sometimes it terminates after just 38 seconds but most often after 80-90 seconds. This issue started happening after we upgraded boundary version from 0.18.0 to 0.21.0.

```bash
time boundary connect -target-id ttcp_EDr1FoQeYX

Proxy listening information:
  Address:             127.0.0.1
  Connection Limit:    -1
  Expiration:          Wed, 04 Feb 2026 00:17:57 CET
  Port:                5030
  Protocol:            tcp
  Session ID:          s_QfZVA6xaXR

  Credentials:
    Credential Source ID:   clvlt_IWRrSAedKh
    Credential Source Name: api-editor
    Credential Store ID:    csvlt_KfpWq1Y1hm
    Credential Store Type:  vault-generic
    Secret:
        {
              "password": "xxxxxxxxxxxxxxxxx",
              "username": "xxxxxxxxxxxxxxxxxxxxxx"
        }


Termination information:
  Reason:   Inactivity timeout reached
boundary connect -target-id ttcp_EDr1FoQeYX  0.07s user 0.12s system 0% cpu 41.064 total
```

**To Reproduce**
1. Run `boundary authenticate`
2. Run `boundary connect`
3. See error
4. Same behaviour with boundary desktop client as well

**Expected behavior**
The session should not close after just 30 seconds of inactivity, ideally this value should be configurable but even if not perhaps after 10-15 minutes maybe?

**Additional context**
Issue happening in version 0.21.0

irenarindos:
Hi bhvishal9, 
Thanks for reporting this. 

We do have a command line parameter you can use to set the inactive timeout:
``` 
-inactive-timeout=<duration>
      How long to wait between connections before closing the session.
      Increase this value if the proxy closes during long-running processes,
      or use -1 to disable the timeout.
```

I'll follow up on having an env variable for this and increasing the inactive timeout.

bhvishal9:
@irenarindos For some reason, I overlooked that flag. Even with it, most of our developers prefer using the desktop client, and it appears that there’s no way to circumvent this issue with that.

Also, the problem is not that it terminates the session but it does so after just like a minute or less.

louisruch:
[...]

PR #6429, opened 2026-02-12
fix(cli): Remove connect inactivity timeout defaults expect for darwin+rdp

louisruch:
## Description

After a few complaints around the timeouts triggering, we have decided to move to a less aggressive approach. This PR:
- Removes the 10 minute timeout on sessions before the first connect. Note this was also causing timeouts even if the user explicitly sets `-1`
- Removes all the default timeout values, except for `boundary connect rdp` when using a Mac client
- Increases the default mac+rdp timeout from 5 seconds to 30 seconds
  - For RDP connections the behavior seen is an initial connection is created, this is expected to fail. Then before the second connection is made the user needs to accept a certificate. The default 5 second timeout begins after the first connection fails leaving the user 5 seconds to accept the certificate.

## PCI review checklist

### DIFF
--- a/api/proxy/option_test.go
+++ b/api/proxy/option_test.go
@@ -108,4 +108,13 @@ func Test_GetOpts(t *testing.T) {
 		require.NoError(t, err)
 		assert.Equal(client, opts.withApiClient)
 	})
+	t.Run("WithInactivityTimeout", func(t *testing.T) {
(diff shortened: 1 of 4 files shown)


louisruch:
Fixes #6398

irenarindos (approved):

PR #6587, opened 2026-04-13
feat(daemon): csp nonce handling

bgajjala8:
## Description

This PR adds nonce handling to support UI features. The changes include a new wrapper to apply CSP with a crypto nonce value and adjust injection of the nonce into returned HTML. This change will take place in tandem with a [UI PR](https://github.com/hashicorp/boundary-ui/pull/3230)

## PCI review checklist

### DIFF
--- a/internal/daemon/controller/handler.go
+++ b/internal/daemon/controller/handler.go
@@ -801,17 +801,11 @@ func wrapHandlerWithCsp(h http.Handler, props HandlerProperties, isUiRequest fun
 		nonceToken := fmt.Sprintf("'nonce-%s'", nonce)
 
 		csp := defaultCsp
-		switch {
-		// In the case there is style-src present in the csp then we will inject nonce
-		case strings.Contains(csp, "style-src "):
+
+		if strings.Contains(csp, "style-src ") {
 			csp = strings.Replace(csp, "style-src ", fmt.Sprintf("style-src %s ", nonceToken), 1)
-		default:
-			// If default-src exists, append a style-src directive
-			// using the same source list.
-			if _, after, ok := strings.Cut(csp, "default-src "); ok {
-				sources, _, _ := strings.Cut(after, ";")
-				csp = fmt.Sprintf("%s; style-src %s %s", csp, nonceToken, strings.TrimSpace(sources))
-			}
+		} else {
+			csp = fmt.Sprintf("%s; style-src %s 'self'", csp, nonceToken)
 		}
 
 		w.Header().Set(cspKey, csp)
--- a/internal/daemon/controller/handler.go
+++ b/internal/daemon/controller/handler.go
@@ -797,9 +797,22 @@ func wrapHandlerWithCsp(h http.Handler, props HandlerProperties, isUiRequest fun
 		}
 		nonce := base64.StdEncoding.EncodeToString(b)
 
-		// Inject nonce
+		// Inject nonce into the style-src directive of the CSP.
 		nonceToken := fmt.Sprintf("'nonce-%s'", nonce)
-		csp := strings.Replace(defaultCsp, "style-src 'self'", fmt.Sprintf("style-src %s 'self'", nonceToken), 1)
+
+		csp := defaultCsp
+		switch {
+		// In the case there is style-src present in the csp then we will inject nonce
+		case strings.Contains(csp, "style-src "):
+			csp = strings.Replace(csp, "style-src ", fmt.Sprintf("style-src %s ", nonceToken), 1)
+		default:
+			// If default-src exists, append a style-src directive
+			// using the same source list.
+			if _, after, ok := strings.Cut(csp, "default-src "); ok {
+				sources, _, _ := strings.Cut(after, ";")
+				csp = fmt.Sprintf("%s; style-src %s %s", csp, nonceToken, strings.TrimSpace(sources))
+			}
+		}
 
 		w.Header().Set(cspKey, csp)
 		// Creating a custom key so we can pull it when serving UI page
--- a/internal/daemon/controller/handler.go
+++ b/internal/daemon/controller/handler.go
@@ -6,6 +6,8 @@ package controller
 import (
 	"bytes"
 	"context"
+	"crypto/rand"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -98,7 +100,8 @@ func (c *Controller) apiHandler(props HandlerProperties) (http.Handler, error) {
 		return nil, err
 	}
 
-	corsWrappedHandler := wrapHandlerWithCors(mux, props)
+	cspWrappedHandler := wrapHandlerWithCsp(mux, props, isUiRequest)
+	corsWrappedHandler := wrapHandlerWithCors(cspWrappedHandler, props)
 	commonWrappedHandler := wrapHandlerWithCommonFuncs(corsWrappedHandler, c, props)
 	callbackInterceptingHandler := wrapHandlerWithCallbackInterceptor(commonWrappedHandler, c)
 	printablePathCheckHandler := cleanhttp.PrintablePathCheckHandler(callbackInterceptingHandler, nil)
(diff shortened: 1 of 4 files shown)


internal/daemon/controller/handler.go
```diff
@@ -759,3 +762,46 @@ func getActions(urlPath string) []string {
...
+			h.ServeHTTP(w, req)
+			return
+		}
+
+		// Static assets get the default CSP without a nonce.
+		if strings.LastIndex(req.URL.Path, ".") != -1 {
+			w.Header().Set(cspKey, defaultCsp)
+			h.ServeHTTP(w, req)
+			return
+		}
```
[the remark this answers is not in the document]
bgajjala8: Yep, I believe the edge case here is a non-ASCII extensions and something we should not need to worry about.

internal/daemon/controller/handler.go
```diff
@@ -759,3 +762,46 @@ func getActions(urlPath string) []string {
...
+		nonce := base64.StdEncoding.EncodeToString(b)
+
+		// Inject nonce
+		nonceToken := "'nonce-" + nonce + "'"
+		csp := strings.Replace(defaultCsp, "style-src 'self'", "style-src "+nonceToken+" 'self'", 1)
+
```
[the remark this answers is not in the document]
bgajjala8: I think this is okay since we control the CSP so we know the format, so I am gonna leave it as is,

internal/daemon/controller/handler_ui.go
```diff
@@ -35,6 +36,24 @@ var serveGrantSchema = func(ctx context.Context, w http.ResponseWriter) {
...
+	csp  string
+	done bool
+}
+
+func (w *cspWriter) Write(b []byte) (int, error) {
+	if !w.done {
+		w.done = true
+		b = bytes.Replace(b, []byte(cspNoncePlaceholder), []byte(w.csp), 1)
+	}
+	return w.ResponseWriter.Write(b)
```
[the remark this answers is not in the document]
bgajjala8: I thinks since, index.html is small, this is fine

internal/daemon/controller/handler_ui.go
```diff
@@ -35,6 +36,24 @@ var serveGrantSchema = func(ctx context.Context, w http.ResponseWriter) {
...
 }
 
+const cspNoncePlaceholder = "__BOUNDARY_CSP_NONCE__"
+
+// cspWriter wraps an http.ResponseWriter to replace the CSP nonce placeholder
+// in index.html with the actual Content-Security-Policy value.
+type cspWriter struct {
+	http.ResponseWriter
+	csp  string
+	done bool
```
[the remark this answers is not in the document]
bgajjala8: Nice! Good catch

irenarindos (dismissed):

internal/daemon/controller/handler_ui.go
```diff
@@ -35,6 +36,24 @@ var serveGrantSchema = func(ctx context.Context, w http.ResponseWriter) {
...
+}
+
+func (w *cspWriter) Write(b []byte) (int, error) {
+	if !w.done {
+		w.done = true
+		b = bytes.Replace(b, []byte(cspPlaceholder), []byte(w.csp), 1)
```
ZedLi, line 52: I notice for the document request on initial load, it gets blocked with `NS_ERROR_NET_PARTIAL_TRANSFER` on firefox and `ERR_CONTENT_LENGTH_MISMATCH` on chrome. Do you know if it's possible the `Content-Length` is being set and _then_ we're rewriting the content so they don't end up matching? Do we need to recalculate the `Content-Length` header? 

I'm also curious if it's possible to just pass in the nonce itself rather than the full CSP that we're pulling from the header?

internal/daemon/controller/handler.go
```diff
@@ -759,3 +762,46 @@ func getActions(urlPath string) []string {
...
+		if _, err := rand.Read(b); err != nil {
+			w.Header().Set(cspKey, defaultCsp)
+			h.ServeHTTP(w, req)
+			return
+		}
+		nonce := base64.StdEncoding.EncodeToString(b)
+
+		// Inject nonce
+		nonceToken := "'nonce-" + nonce + "'"
+		csp := strings.Replace(defaultCsp, "style-src 'self'", "style-src "+nonceToken+" 'self'", 1)
```
ZedLi, line 802: I wanted to double check since I don't know when this happens in the response pipeline but is it possible for a customer to [override](https://developer.hashicorp.com/boundary/docs/monitor/listeners/tcp#custom_ui_response_headers-parameters) the header and and `style-src 'self'` wouldn't exist or be different?

bgajjala8: Thanks for bringing this up. I didn't know this override existed 👀

internal/daemon/controller/handler_ui.go
bgajjala8: Yes, we can pass in the nonce it self. Let me verify, but I think we do need to recalculate since we are injecting the into the field after the fact.

bgajjala8: This was a great catch. I ran CURL to validate and I guess there is no  content length validation so it was passing. 🙇

internal/daemon/controller/handler.go
bgajjala8: Yes, a customer can override it. If they do not provide a CSP with `style-src` I am thinking we  can inject the nonce based on these conditions:

1. If `style-src` already exists, we prepend the nonce to it 
2. If `style-src` doesn't exist but `default-src` does, we create a new `style-src` directive using the same source list from `default-src` plus the nonce
3. If neither `style-src` nor `default-src` exists, we return the CSP unchanged 

In the 3rd case is where nonce would not be set. Now, I don't know if that is common and without either directive (Also, not sure of the implication for loading the F/E).

I am not sure if I am missing anything else in terms of F/E

Thoughts on this @ZedLi

ZedLi: I don't think in practice it matters much since if they decided to completely remove the `style-src` (i.e. no `self` policy), things would stop working from our end regardless. Either way though, is there a reason we shouldn't just always add the `style-src` with the nonce even if there's no `style-src`? 

I also think it can get wonky even in option 2 since if they have `default-src` set to `none`, wouldn't they now conflict with the addition of the nonce? We'd also have to check `default-src` is not set to `none` before combining.

bgajjala8: No specific reason not to! I think we can add `style-src` if it's not already present. And yes, copying from `default-src` would get wonky with the `'none'` case. It makes sense to keep it simple and just adding `style-src` with nonce + `'self'` when it's not there should be good for our use case.

bgajjala8: Updated!

internal/daemon/controller/handler_ui.go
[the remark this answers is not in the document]
ZedLi: Hey @bgajjala8, I just wanted to double check if this might be an issue. I think I'm noticing the response might be getting truncated. I know you mentioned the index.html is small but that's also before we do our build and we actually write and replace content in it. 

Icons weren't loading for me and I couldn't figure out why but it looks like the final large `svg` tag is getting truncated when the response gets rewritten (if I change the meta tag to the wrong placeholder everything is as expected).

When I downloaded the response documents, the original was ~1MB but the new rewritten one ends up only being around ~33KB.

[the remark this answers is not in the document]
bgajjala8: Great catch, though I don't think this is a multi chunk issue @ZedLi 

I think what was happening is that originally once I modified the buffer in `bytes.Replace` I was returning the modified length which conflicted with what the writer was expecting for its buffer management. So I think this was breaking the buffer tracking and causing it to stop sending chunks, which truncated the response. I made a change to explicitly return the original length even after the byte adjustment and this seems to fix the issue.

From the [io.Writer](https://pkg.go.dev/io#Writer)

> Write writes len(p) bytes from p to the underlying data stream. It returns the number of bytes written from p (0 <= n <= len(p)) and any error encountered that caused the write to stop early. Write must return a non-nil error if it returns n < len(p). Write must not modify the slice data, even temporarily.

It also make sense when you changed the tag it worked because `bytes.Replace` would have just made a copy of the exact same data since nothing was replaced. [[docs](https://pkg.go.dev/bytes#Replace)]

internal/daemon/controller/handler.go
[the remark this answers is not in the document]
bgajjala8: Updated!

ZedLi (approved):
Tested and works as expected for me!

irenarindos (approved):

PR #6588, opened 2026-04-13
chore(e2e): Fix worker version e2e-docker-base-with-worker-version to account for 1.0 release

dillanb-hashi:
## Description
Change the logic for getting the version to taking current version - 1 to taking a list of versions and then finding the version before.
examples
0.21.3->0.20.2
1.0.0->0.21.3 (or whatever the last non 1.x.x version is)

## PCI review checklist

### DIFF
--- a/.github/workflows/enos-run.yml
+++ b/.github/workflows/enos-run.yml
@@ -101,8 +101,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
-          fetch-depth: ${{ matrix.scenario == 'e2e_docker_base_with_worker_version' && '0' || '1' }}
-          fetch-tags: ${{ matrix.scenario == 'e2e_docker_base_with_worker_version' && 'true' || 'false' }}
+          fetch-depth: ${{ contains(matrix.filter, 'e2e_docker_base_with_worker_version') && '0' || '1' }}
+          fetch-tags: ${{ contains(matrix.filter, 'e2e_docker_base_with_worker_version') && 'true' || 'false' }}
       - name: Determine Go version
         id: get-go-version
         # We use .go-version as our source of truth for current Go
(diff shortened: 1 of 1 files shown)


.github/workflows/enos-run.yml
```diff
@@ -265,12 +267,39 @@ jobs:
...
-        # Get the Boundary version number and decrement the minor version by 1
+        # Resolve the worker version from the previous release line.
         if: contains(matrix.filter, 'e2e_docker_base_with_worker_version')
         run: |
-          version_num=$(./enos/support/boundary/boundary version | awk -F'[. ]' '/Version Number/ {print $(NF-1)}')
-          export ENOS_VAR_worker_version="$((version_num - 1))"
-          echo "ENOS_VAR_worker_version=0.$ENOS_VAR_worker_version" >> "$GITHUB_ENV"
+          current_version_raw=$(./enos/support/boundary/boundary version | awk '/Version Number/ {print $3}')
+          current_version="${current_version_raw#v}"
+          current_version="${current_version%%[-+]*}"
```
moduli: It looks like this this handles boundary versions coming from boundary-enterprise (i.e. ` 0.21.2+ent`), but I don't think we're handling this with tags?

dillanb-hashi: The current_version_raw outputs 0.21.0+ent line 275 takes out the +ent. Line 274 probably isn't needed though.

moduli: Right. That's the output of `boundary version`. But when you're looking up the tags later on, boundary-enterprise has tags with `v#.#.#+ent`

dillanb-hashi: Ah yes. I'll test this in the ent branch as well to make sure I can get the right version. We handle adding the +ent in the enos scenario I'm pretty sure so it should strip the +ent from the git tags when it's parsing through that.

moduli: I think I might be confusing myself.
- For this workflow, this probably works ok.
- We will need to update `enos-run-ent.yml` accordingly. We should specifically match against +ent tags there as there may be some versions only available in enterprise.

moduli (approved):
Reminder: There will need to be a follow-up to update this in `enos-run-ent.yml`. It will also need to account for using the `-ent` version of the tags.

PR #6626, opened 2026-04-24
Memory leak fix for worker session management by immediately removing closed connections

ChandraSGadde:
## Description

When a connection was confirmed closed by the controller, its ConnInfo entry was kept in connInfoMap indefinitely - with a CloseTime. For long-lived sessions with many short-lived connections, this caused unbounded map growth and orphaned proxy go-routines.

The changes implement a memory leak fix for worker session management by immediately removing closed connections from the local map instead of just marking them with a close time.

Added additional tests covering removal of CloseTime , cancel func .

## PCI review checklist

### DIFF
--- a/internal/daemon/worker/session/session.go
+++ b/internal/daemon/worker/session/session.go
@@ -36,10 +36,6 @@ type ConnInfo struct {
 	BytesUp func() int64
 	// The number of bytes downloaded to the client.
 	BytesDown func() int64
-
-	// The time the controller has successfully reported that this connection is
-	// closed.
-	CloseTime time.Time
 }
 
 // ConnectionCloseData encapsulates the data we need to send via CloseConnection
@@ -171,7 +167,14 @@ func (s *sess) ApplyLocalConnectionStatus(connId string, status pbs.CONNECTIONST
 
 	connInfo.Status = status
 	if connInfo.Status == pbs.CONNECTIONSTATUS_CONNECTIONSTATUS_CLOSED {
-		connInfo.CloseTime = time.Now()
+		// Cancel the proxy context so the goroutine is stopped in the
+		// controller-initiated close path, then immediately remove the
+		// entry from the map to prevent unbounded growth when a session
+		// carries a large number of short-lived connections.
+		if connInfo.connCtxCancelFunc != nil {
+			connInfo.connCtxCancelFunc()
+		}
+		delete(s.connInfoMap, connId)
 	}
 	return nil
 }
@@ -204,7 +207,6 @@ func (s *sess) GetLocalConnections() map[string]ConnInfo {
 		res[k] = ConnInfo{
 			Id:        v.Id,
 			Status:    v.Status,
-			CloseTime: v.CloseTime,
 			BytesUp:   v.BytesUp,
 			BytesDown: v.BytesDown,
 		}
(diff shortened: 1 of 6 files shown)


YashwanthaGowd (approved):

internal/daemon/worker/session/session.go
```diff
@@ -171,7 +167,14 @@ func (s *sess) ApplyLocalConnectionStatus(connId string, status pbs.CONNECTIONST
...
 	if connInfo.Status == pbs.CONNECTIONSTATUS_CONNECTIONSTATUS_CLOSED {
-		connInfo.CloseTime = time.Now()
+		// Cancel the proxy context so the goroutine is stopped in the
+		// controller-initiated close path, then immediately remove the
+		// entry from the map to prevent unbounded growth when a session
+		// carries a large number of short-lived connections.
+		if connInfo.connCtxCancelFunc != nil {
+			connInfo.connCtxCancelFunc()
+		}
+		delete(s.connInfoMap, connId)
```
irenarindos, line 177: Won't this cause data loss, since the connection can't be reported in session statistics (`func (w *Worker) sendSessionInfo`)?

irenarindos (changes requested):
I'm worried this will result in connections that do not have bytes up/down reported, since connections deleted from the session manager map cannot be reported in session statistics

internal/daemon/worker/session/session.go
ChandraSGadde: @irenarindos 
Yes, there is a risk.
For successful close RPCs, the final BytesUp/BytesDown are reported before the connection is removed locally. However, if the close RPC to the controller fails, the worker still marks the connection closed locally and deletes it from connInfoMap. 

In that failure case, the connection will no longer appear in subsequent session statistics updates, so the latest byte counters may be lost from controller-side reporting.

Sequence of events --
1) makeCloseConnectionRequest(closeInfo)— builds the RPC payload with BytesUp/BytesDown captured from the live proxy counters.
2) closeConnection(...) — sends those byte counts to the controller
3) removeClosedConnections → ApplyLocalConnectionStatus(CLOSED) → delete(s.connInfoMap, connId) — only NOW is the entry removed

The entry is deleted after the controller has already acknowledged the byte counts. By the time 'delete(s.connInfoMap, connId)' runs, the stats have already been reported.

sendSessionInfo - is a separate RPC that only reports session-level status (session IDs + active/closed state) — it never carries per-connection byte statistics at all, so it is unaffected by the delete.

ChandraSGadde:
> I'm worried this will result in connections that do not have bytes up/down reported, since connections deleted from the session manager map cannot be reported in session statistics

Follow up to my previous comment -- 
I did further analysis on data-loss scenario - failed close RPC. I could be wrong in my analysis, please advise.

Existing **tombstone** - **CloseTime** approach —
1. RPC fails → `makeFakeSessionCloseInfo` called
2. `ApplyLocalConnectionStatus(CLOSED)` → sets `connInfo.CloseTime = time.Now()` (proxy goroutine still running, bytes still accumulating)
3. On next `cleanupConnections` cycle: `CancelOpenLocalConnections` checks `CloseTime.IsZero()` — it's **false** → connection is **skipped**, never retried
4. Bytes lost from controller perspective. Proxy leaks until session ends.

Current approach - 
1. Same — RPC fails → `makeFakeSessionCloseInfo`
2. `ApplyLocalConnectionStatus(CLOSED)` → cancel proxy + `delete` from map
3. On next cycle: connection gone from map → not retried
4. Bytes lost from controller perspective. Proxy stopped cleanly.

The byte-loss outcome is **identical**. IMO, current fix is actually slightly better in the failure case: we cancel the proxy immediately (stopping the leak) rather than letting it run as a zombie accumulating bytes that can never be reported.

The existing error log already acknowledges this as a known degraded mode :
"warning", "error contacting controller, connections will be closed only on worker",

This is a pre-existing design trade-off — a retry mechanism for failed close RPCs would be a separate, larger feature.

irenarindos (approved):

Issue #4236, opened 2024-01-16
Wrong key name in log output when using Azure KMS


sym-stiller:
**Describe the bug**
During startup, the Boundary controllers and workers write a few log lines which include the names of the keys that are in use. When using Azure Key Vault KMS, there is a bug that will only output the name of the last configured key, instead of all names.

Example output from my controller:
```
==> Boundary server configuration:

Azure Environment: AzurePublicCloud
Azure Environment: AzurePublicCloud
Azure Environment: AzurePublicCloud
Azure Key Name: boundary-recovery  <--- this one should be boundary-root
Azure Key Name: boundary-recovery  <--- this one should be boundary-worker
Azure Key Name: boundary-recovery
Azure Vault Name: keyvault-name
Azure Vault Name: keyvault-name
Azure Vault Name: keyvault-name
Cgo: disabled
Controller Public Cluster Addr: boundary.exampledomain.com:9201
Listener 1: tcp (addr: "0.0.0.0:9200", cors_allowed_headers: "[]", cors_allowed_origins: "[https://boundary.exampledomain.com serve://boundary]", cors_enabled: "true", max_request_duration: "1m30s", purpose: "api")
Listener 2: tcp (addr: "0.0.0.0:9201", max_request_duration: "1m30s", purpose: "cluster")
Listener 3: tcp (addr: "0.0.0.0:9203", max_request_duration: "1m30s", purpose: "ops")
Log Level: trace
Mlock: supported: true, enabled: false
Version: Boundary v0.14.3
Version Sha: de3a3c0c382a4a394ab0d3b349ae855d66463f9f

==> Boundary server started! Log data will stream in below:
```

Despite using 3 different keys for the `root`, `worker-auth` and `recovery` purposes, only the key name that appeared last in the configuration file will appear in the logs. This issue is not reproducible with AEAD KMS. Here is an abbreviated version of my configuration file:

```
# additional arguments set via env variables:
# AZURE_ENVIRONMENT, AZURE_TENANT_ID, AZURE_CLIENT_ID, AZURE_CLIENT_SECRET, AZUREKEYVAULT_WRAPPER_VAULT_NAME
kms "azurekeyvault" {
  purpose  = "root"
  key_name = "boundary-root"
}

kms "azurekeyvault" {
  purpose  = "worker-auth"
  key_name = "boundary-worker"
}

kms "azurekeyvault" {
  purpose  = "recovery"
  key_name = "boundary-recovery"
}
```

[...]

PR #6627, opened 2026-04-24
fix(kms): Duplicate KMS startup metadata labels across multiple KMS blocks

anilvpatel:
**Summary:**
Fixes incorrect duplicate KMS metadata in Boundary server startup output when multiple KMS blocks of the same provider are configured (for example, Azure Key Vault root, worker-auth, and recovery blocks).

Address Issue: https://github.com/hashicorp/boundary/issues/4236

**Description:** 
During startup, the Boundary controllers and workers write a few log lines which include the names of the keys that are in use. When using more than one block of the same provider (expect AEAD) with different purpose. There is a bug that will only output the name of the last configured key, instead of all names.

For controller config: 
run `boundary server -config controller.hcl`
```
disable_mlock = true

controller {
   ...
}

listener "tcp" {
  ...
}

listener "tcp" {
  ...
}

kms "azurekeyvault" {
  purpose = "root"
  key_name = "boundary-root"
}

kms "azurekeyvault" {
  purpose = "worker-auth"
  key_name = "boundary-worker"
}

kms "azurekeyvault" {
  purpose = "recovery"
  key_name = "boundary-recovery"
}
````
**Example output from my controller:**

Screenshot 2026-04-23 at 3 20 51 PM <br>

**Root-Cause:**
Boundary passed shared startup info containers (b.InfoKeys and b.Info) directly into [configutil.ConfigureWrapper() ](https://github.com/hashicorp/boundary/blob/1684f729cbc506ce8b6099ebe97e1232bb6a1688/internal/cmd/base/servers.go#L598) for every configured KMS block.

For providers like Azure Key Vault, multiple KMS blocks emit the same display labels, such as: Azure Environment, Azure Vault Name, Azure Key Name. [go-secure-stdlib/configutil/kms.go](https://github.com/hashicorp/go-secure-stdlib/blob/2f652a33c02df3e74e4ca6fff6af1110cbefadda/configutil/kms.go#L410)

Because `b.Info` is a map, later KMS blocks overwrote earlier values for the same label, while b.InfoKeys continued appending duplicate keys. This caused startup output to print the same label multiple times, all resolving to the last configured value.

This was a startup info rendering bug only. It did not affect KMS purpose assignment or runtime wrapper usage. 

**Solution:** 
Collect KMS metadata per wrapper in temporary local structures, then merge it into b.InfoKeys / b.Info after all KMS blocks have been processed.

When duplicate display labels are detected, qualify them with KMS purpose (and ordinal when the same purpose appears multiple times ex: downstream-worker-auth), so the printed keys remain unique. 

For example:
[Root] Azure Key Name
[Worker-Auth] Azure Key Name
[Recovery] Azure Key Name

This keeps startup output accurate without changing runtime KMS behavior. <br>
Screenshot 2026-04-24 at 1 22 21 PM

<br>

**Scope:**
This fix is provider-agnostic and applies to any KMS wrapper type that emits repeated display labels across multiple configured KMS blocks.

**Why not fix upstream in configutil?**
This was fixed in Boundary rather than configutil because duplicate resolution is a cross-block display concern. configutil only has per-wrapper context, while Boundary has the full set of configured KMS blocks, purposes, and repeated-purpose counts needed to generate stable, unique startup labels.

**Tests**
Added regression coverage for:
- duplicate KMS metadata labels across different purposes.
- duplicate metadata labels for repeated instances of the same purpose.
- skipping purpose for KMS AEAD type as its already done in upstream. 
- preserve all existing info keys

### DIFF
--- a/internal/cmd/base/server_test.go
+++ b/internal/cmd/base/server_test.go
@@ -884,9 +884,8 @@ func TestSetupWorkerPublicAddress(t *testing.T) {
 	}
 }
 
-func TestMergeKmsInfoEntries(t *testing.T) {
+func TestMergeKmsInfoEntriesIntoInfoMap(t *testing.T) {
 	t.Parallel()
-
 	t.Run("duplicate labels are qualified by purpose", func(t *testing.T) {
 		t.Parallel()
 		assert := assert.New(t)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: internal/cmd/base/servers.go [hunk 1]
```diff
@@ -882,3 +903,28 @@ func MakeSighupCh() chan struct{} {
 	}()
 	return resultCh
 }
+
+// modifying the displayKey to include purpose information.
+func mergeKmsInfoEntries(infoKeys *[]string, info map[string]string, entries []kmsInfoEntry, purposeTotals map[string]uint) {
+	if infoKeys == nil || info == nil {
```
ChandraSGadde, line 910: I think if purposeTotals is nil, this will panic. can you check ?

internal/cmd/base/servers.go
```diff
@@ -882,3 +903,28 @@ func MakeSighupCh() chan struct{} {
 	}()
 	return resultCh
 }
+
+// modifying the displayKey to include purpose information.
```
ChandraSGadde: nitpick: 
// mergeKmsInfoEntries builds display keys for each KMS info entry, qualifying them with purpose when needed

internal/cmd/base/servers.go
```diff
@@ -882,3 +903,28 @@ func MakeSighupCh() chan struct{} {
...
+func mergeKmsInfoEntries(infoKeys *[]string, info map[string]string, entries []kmsInfoEntry, purposeTotals map[string]uint) {
+	if infoKeys == nil || info == nil {
+		return
+	}
+
+	for _, entry := range entries {
+		displayKey := entry.key
+		// Skipping KMS "aead" type as its purpose is already represented in the display key.
+		if entry.kmsType != wrapping.WrapperTypeAead.String() {
```
ChandraSGadde, line 917: suggestion : kmsInfoEntry stores the full kmsType string but it's only ever used to check whether the type is AEAD. Consider replacing it with a skipPurposePrefix bool field set at construction time — avoids leaking the 'wrapping' package dependency into 'mergeKmsInfoEntries'

internal/cmd/base/server_test.go
```diff
@@ -883,3 +883,60 @@ func TestSetupWorkerPublicAddress(t *testing.T) {
 		})
 	}
 }
+
+func TestMergeKmsInfoEntries(t *testing.T) {
```
ChandraSGadde: Suggestion :- A test covering the AEAD exemption can be added.

anilvpatel: added new testcases:
- skipping for KMS AEAD type
- preserve all existing info keys

internal/cmd/base/servers.go
anilvpatel: To preserve the type of KMS configured, we added a kmsType field to kmsInfoEntry. This provides a clearer and more flexible approach compared to the skipPurposePrefix boolean field.

internal/cmd/base/servers.go
anilvpatel: added.

internal/cmd/base/servers.go
anilvpatel: Fixed, and we now validate that purposeTotals is not nil before using it.

ChandraSGadde (approved):

PR #6644, opened 2026-04-25
feat(pprof): Enable pprof endpoints behind flag

louisruch:
## Description

A number of customers have needed to be given a custom binary in order to provide profiles to support as this is currently behind a build flag. Moving this to a exec flag allows customers to turn it on as needed without the custom binary. This is currently not be added to the config as it is meant to be a debug mode that is not part of the configuration of the controller or worker.

## PCI review checklist


irenarindos (approved):

RyanDerr (approved):
Nice job! Wondering if we should also backport the `debug` option to `n-2` versions as well? I think it's likely useful for older versions of boundary we support to also provide this option.

PR #6683, opened 2026-05-05
Add "pins" as a grant segment

jefferai:
## Description

The ability to pin grants to a specific collection has been supported for a long time, but the syntax is confusing. It reuses `ids` (originally `id`) with the thinking that you are granting, e.g., `read` on `host sets` within the collection with id `foobar`. Or, from another standpoint, in `foobar` allow `read` on `host sets`. Look, it made sense at the time.

However, it means there are different potential meanings for what can go into `ids`, and that in turn makes it confusing to explain pins and can lead to subtle misunderstandings due to e.g. mismatched types.

This change allows `pins` to be used instead, which makes it very explicit what the intention is. Internally, for backwards compatibility, the behavior is very similar to pinning with `ids` with a few extra guardrails -- something we could consider changing in the future if we want, although we'd have to do it at the API level to ensure existing grants work -- but lets the user be more specific and expressive as to intent.

## PCI review checklist

- [X] I have documented a clear reason for, and description of, the change I am making.

### DIFF
--- a/internal/perms/acl.go
+++ b/internal/perms/acl.go
@@ -173,7 +173,7 @@ func NewACL(grants ...Grant) ACL {
 		}
 
 		// Handle explicit pins grants. Pins are stored in AclGrant.Id so
-		// that Case 5 of Allowed() can match them against Resource.Pin.
+		// that Case 5 of acl.Allowed() can match them against Resource.Pin.
 		for _, pin := range grant.pins {
 			addGrant(grant, pin)
 		}
--- a/internal/perms/acl.go
+++ b/internal/perms/acl.go
@@ -147,24 +147,35 @@ func NewACL(grants ...Grant) ACL {
 		descendantsGrants: make([]AclGrant, 0, len(grants)),
 	}
 
+	addGrant := func(grant Grant, id string) {
+		switch grant.grantScopeId {
+		case globals.GrantScopeDescendants:
+			ret.descendantsGrants = append(ret.descendantsGrants, aclGrantFromGrant(grant, id))
+		case globals.GrantScopeChildren:
+			// We use the role's scope here because we're evaluating the
+			// grants themselves, not the resource, so we want to know the
+			// scope of the role that said "children"
+			ret.childrenScopeMap[grant.roleScopeId] = append(ret.childrenScopeMap[grant.roleScopeId], aclGrantFromGrant(grant, id))
+		default:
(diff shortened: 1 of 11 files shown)


internal/proto/controller/api/resources/roles/v1/role.proto
```diff
@@ -36,6 +36,10 @@ message GrantJson {
 
   // Output only. The actions.
   repeated string actions = 3; // @gotags: `class:"public"`
+
+  // Output only. The pins, if set. Pins constrain the grant to resources
+  // within the given parent resources.
+  repeated string pins = 5; // @gotags: `class:"public"`
```
irenarindos, line 42: Looks like you need to run make-gen: https://github.com/hashicorp/boundary/actions/runs/25406013733/job/74517006608?pr=6683

internal/perms/acl.go
```diff
@@ -147,24 +147,35 @@ func NewACL(grants ...Grant) ACL {
...
-				// scope of the role that said "children"
-				ret.childrenScopeMap[grant.roleScopeId] = append(ret.childrenScopeMap[grant.roleScopeId], aclGrantFromGrant(grant, id))
-			default:
-				ret.directScopeMap[grant.grantScopeId] = append(ret.directScopeMap[grant.grantScopeId], aclGrantFromGrant(grant, id))
-			}
+			addGrant(grant, id)
+		}
+
+		// Handle explicit pins grants. Pins are stored in AclGrant.Id so
+		// that Case 5 of Allowed() can match them against Resource.Pin.
```
irenarindos: to be a bit more clear:
```suggestion
		// that Case 5 of acl.Allowed() can match them against Resource.Pin.
```

internal/perms/grants.go
```diff
@@ -410,6 +452,14 @@ func (g *Grant) unmarshalText(ctx context.Context, grantString string) error {
 				}
 			}
 
+		case "pins":
+			g.pins = strings.Split(kv[1], ",")
+			for _, pin := range g.pins {
+				if pin == "" {
+					return errors.New(ctx, errors.InvalidParameter, op, "empty pin provided")
```
irenarindos, line 466: Are text pins allowed to contain a comma, semicolon, or equals sign?

internal/proto/controller/api/resources/roles/v1/role.proto
jefferai: Good catch, done!

internal/perms/grants.go
jefferai: Good catch -- the ID checks were missing that too!

irenarindos (approved):

PR #6688, opened 2026-05-06
fix(sqltest): Remove deprecated docker --link usage 

RyanDerr:
## Overview
This replaces the deprecated Docker [--link](https://docs.docker.com/engine/network/links/) flag with an explicit user-defined bridge network in the sqltest Makefile. The `--link` flag is a legacy feature that Docker recommends against and that [Podman does not implemented](https://docs.podman.io/en/latest/markdown/podman-run.1.html).
## Notes
- There should be no functional change to test behavior. Containers instead now resolve each other by name via DNS on a shared network instead of through link aliases.

## PCI review checklist

### DIFF
--- a/internal/db/sqltest/Makefile
+++ b/internal/db/sqltest/Makefile
@@ -56,14 +56,18 @@ ifeq ($(DB_LOGS), 1)
 	log_cmd = docker logs $(SQL_TEST_CONTAINER_NAME)
 endif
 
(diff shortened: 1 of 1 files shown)


moduli (approved):
Looks like the workflow that invokes this is still passing: https://github.com/hashicorp/boundary/actions/runs/25455730522/job/74684558555?pr=6688

PR #6692, opened 2026-05-08
fix(audit): populate email/name in audit events 

louisruch:
## Description

Fixes two bugs around auth data being populated in audit events:

- The first bug was making use of a `userData` local variable that was never actually populated with the user data
- The second bug was that when we added support for ldap we never updated the iam_acct_info view to grab the full name and email from the table

These bugs were never caught because the test excising this path used the default redaction rules to set `fullName` and `email` to `<redacted>`. Note that we do also redact empty strings.

## PCI review checklist

### DIFF
--- a/internal/daemon/controller/auth/auth_test.go
+++ b/internal/daemon/controller/auth/auth_test.go
@@ -14,6 +14,8 @@ import (
 	"testing"
 
 	"github.com/hashicorp/boundary/globals"
+	"github.com/hashicorp/boundary/internal/auth/ldap"
+	"github.com/hashicorp/boundary/internal/auth/oidc"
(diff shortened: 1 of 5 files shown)


internal/daemon/controller/auth/auth_test.go
```diff
@@ -266,6 +268,337 @@ func TestVerify_AuditEvent(t *testing.T) {
 	}
 }
 
+func TestVerify_UnredactedLdapAuth_AuditEvent(t *testing.T) {
```
louisruch, line 271: I did consider making a helper test to test the three paths Unredacted and then Ldap and OIDC unredacted - but I was not convinced it was any better so left it as different tests

irenarindos (approved):
🥳

PR #6757, opened 2026-07-07
chore: add set/remove alias suffix API definitions

amitsk6:
## Description
Define  set/remove alias suffix  RPC, so that it is available in API module.
## PCI review checklist

### DIFF
--- a/api/scopes/custom.go
+++ b/api/scopes/custom.go
@@ -353,3 +353,140 @@ func (c *Client) DetachStoragePolicy(ctx context.Context, scopeId string, versio
 	s.Response = resp
 	return s, nil
 }
+
+type (
+	SetAliasSuffixResult    = ScopeReadResult
+	RemoveAliasSuffixResult = ScopeReadResult
+)
+
+// SetAliasSuffix sets the alias suffix on the provided scopeId.
+func (c *Client) SetAliasSuffix(ctx context.Context, scopeId string, version uint32, aliasSuffix string, opt ...Option) (*SetAliasSuffixResult, error) {
+	if scopeId == "" {
+		return nil, errors.New("empty scopeId value passed into SetAliasSuffix request")
+	}
+	if aliasSuffix == "" {
+		return nil, errors.New("empty aliasSuffix value passed into SetAliasSuffix request")
+	}
+	if c.client == nil {
+		return nil, errors.New("nil client")
+	}
+
+	opts, apiOpts := getOpts(opt...)
+
+	if version == 0 {
+		if !opts.withAutomaticVersioning {
+			return nil, errors.New("zero version number passed into SetAliasSuffix request")
+		}
+		existingScope, existingErr := c.Read(ctx, scopeId, append([]Option{WithSkipCurlOutput(true)}, opt...)...)
+		if existingErr != nil {
+			if api.AsServerError(existingErr) != nil {
+				return nil, fmt.Errorf("error from controller when performing initial check-and-set read: %w", existingErr)
+			}
+			return nil, fmt.Errorf("error performing initial check-and-set read: %w", existingErr)
+		}
+		if existingScope == nil {
+			return nil, errors.New("nil resource response found when performing initial check-and-set read")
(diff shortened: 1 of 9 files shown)


internal/alias/target/alias_test.go
```diff
@@ -117,16 +117,16 @@ func TestCreate(t *testing.T) {
 			errContains: `destination_id_set_when_host_id_is_set constraint failed`,
 		},
 		{
-			name:        "unsupported project scope",
+			name:        "supported project scope",
 			scope:       proj.GetPublicId(),
-			value:       "unsupported.project.scope",
-			errContains: `alias_must_be_in_global_scope constraint failed`,
+			value:       "supported.project.scope",
```
RyanDerr: **Question**: I don't have much context on the alias changes but the test naming is a little confusing. This case is named as a supported project scope yet it still expects an error. Is that the intended behavior?

amitsk6: Project scope is supported but I have not backported any DB and repo changes where this has been updated. So may be it would be better to keep the test name as is, will update it.

RyanDerr: I think there may be a similar thing going on with `supported org scope` as well.

api/scopes/custom.go
```diff
@@ -353,3 +353,140 @@ func (c *Client) DetachStoragePolicy(ctx context.Context, scopeId string, versio
...
 	return s, nil
 }
+
+type (
+	SetAliasSuffixResult    = ScopeReadResult
+	RemoveAliasSuffixResult = ScopeReadResult
+)
+
+// SetAliasSuffix sets the alias suffix on the provided scopeId.
+func (c *Client) SetAliasSuffix(ctx context.Context, scopeId string, version uint32, aliasSuffix string, opt ...Option) (*SetAliasSuffixResult, error) {
```
ddebko, line 363: Is there a reason this needs to live in a custom.go file rather than a generated Go file? We have similar "mutation" type actions against target resource types, e.g. [SetCredentialSource](https://github.com/hashicorp/boundary/blob/b495fe653bb04a02ca31120422055fa3c1acd949/api/targets/target.gen.go#L698), so it seems like this could follow the same pattern.

api/scopes/custom.go
```diff
@@ -353,3 +353,140 @@ func (c *Client) DetachStoragePolicy(ctx context.Context, scopeId string, versio
 	s.Response = resp
 	return s, nil
 }
+
+type (
+	SetAliasSuffixResult    = ScopeReadResult
```
ddebko, line 358: I don't think we need to create custom types that alias ScopeReadResult for each mutation action. Looking at [SetCredentialSources](https://github.com/hashicorp/boundary/blob/b495fe653bb04a02ca31120422055fa3c1acd949/api/targets/target.gen.go#L698), it returns the TargetUpdateResult type instead, and the same pattern is used across several other mutations: AddCredentialSources, AddHostSources, SetHostSources, RemoveCredentialSources, and RemoveHostSources. We should probably follow that same convention here.

ddebko (changes requested):

api/scopes/custom.go
ddebko: Ignore this comment, I forgot this was a change being backported.

api/scopes/custom.go
ddebko: Ignore this comment, I forgot this was a backported change.

RyanDerr (dismissed):
Besides the [minor comment](https://github.com/hashicorp/boundary/pull/6757#discussion_r3539712855), looks good!

ddebko (dismissed):

internal/alias/target/alias_test.go
amitsk6: Good catch, I had missed that. I have reverted changes to alias_test.go to avoid any confusion.

RyanDerr (approved):

Mohit0233 (approved):
LGTM

PR #6764, opened 2026-07-27
chore: Update secret usage in github actions

moduli:
## Description
This PR makes the following changes to the use of secrets in github actions workflows
- Remove the use of `ENOS_GPG*` secrets as we can generate the required info on the fly instead of relying on hard-coded secrets
- Remove use of cloud providers in e2e tests so we're not storing credentials
- Remove use of slack bot

## PCI review checklist

### DIFF
--- a/.github/workflows/enos-run.yml
+++ b/.github/workflows/enos-run.yml
@@ -67,12 +67,9 @@ jobs:
       fail-fast: false # don't fail as that can skip required cleanup steps for jobs
       matrix:
         include:
-          - filter: 'e2e_aws builder:crt ip_version:4'
-          - filter: 'e2e_database'
           - filter: 'e2e_docker_base builder:crt'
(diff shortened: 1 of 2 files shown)


.github/workflows/enos-run.yml
```diff
@@ -396,24 +390,3 @@ jobs:
         run: |
           env
           find ./enos -name "scenario.tf" -exec cat {} \;
-      - name: Send Slack message if Run and Retry fails (or if something else went wrong)
```
dillanb-hashi, line 399: What's the reasoning for getting rid of slack messages? Is there somewhere else we'll be routinely looking for failures?

irenarindos: Same question ^

irenarindos (approved):

